PR #1, opened 2020-04-20
Add f16 support.

bzm3r:
Handling f16 requires special work, compared to other scalars, as the minimum conversion operation for u32->f16 in GLSL (unpackHalf2x16) loads two f16s from one u32. This means that in order to minimize unnecessary calls to unpackHalf2x16, we should look-ahead to see if the current f16 has already been extracted in the process of dealing with the last f16. Similar considerations exist for write operations, where we want to pack, when possible, two f16s in one go (using packHalf2x16).


piet-gpu-derive/src/glsl.rs
```diff
@@ -210,8 +258,47 @@ fn extract_ibits(offset: usize, nbytes: usize) -> String {
...
+fn is_f16(ty: &GpuType) -> bool {
+    match ty {
+        GpuType::Scalar(s) => match s {
+            GpuScalar::F16 => true,
+            _ => false,
+        },
+        GpuType::Vector(s, _) => match s {
+            GpuScalar::F16 => true,
+            _ => false,
+        },
```
raphlinus, line 289:
```suggestion
        GpuType::Scalar(GpuScalar::F16) => true,
        GpuType::Vector(GpuScalar::F16, _) => true,
```

piet-gpu-types/src/lib.rs
```diff
@@ -1,3 +1,12 @@
 pub mod encoder;
 pub mod ptcl;
 pub mod scene;
+pub mod test;
+
+pub struct f16(half::f16);
```
raphlinus, line 6: So as written this has the problem that there's no way to construct a value. That could be fixed by changing it to `pub struct f16(pub half::f16);` but I'm also thinking of a different approach. That mostly involves adding this code to the module (see line 16 of piet-gpu-derive/derive.rs):

```rust
            trait HalfToLeBytes {
                fn to_le_bytes(&self) -> [u8; 2];
            }
            
            impl HalfToLeBytes for half::f16 {
                fn to_le_bytes(&self) -> [u8; 2] {
                    self.to_bits().to_le_bytes()
                }
            }
```

piet-gpu-derive/src/derive.rs
```diff
@@ -121,6 +121,7 @@ fn gen_derive_ty(ty: &GpuType) -> proc_macro2::TokenStream {
 
 fn gen_derive_scalar_ty(ty: &GpuScalar) -> proc_macro2::TokenStream {
     match ty {
+        GpuScalar::F16 => quote!(crate::f16),
```
raphlinus, line 124: If you take my "different approach" (see below), then this changes to `half::f16`.

PR #2, opened 2020-04-29
Set up validation layers and debug reporting.

bzm3r:
This PR sets up validation layers and debug reporting, when running in debug mode. In release mode, these features are not turned on. 

[x] ran `cargo fmt`
[x] ran `collatz` example of `piet-gpu-hal`
[ ] ran `piet-gpu`

I tried running `piet-gpu` but got an error unrelated to the debug layers:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TIMEOUT', piet-gpu\src\main.rs:215:9
```

which refers to the following timing out in `main.rs`:

`device.run_cmd_buf(&cmd_buf).unwrap();`

Any ideas?

(Note that this PR does not try to fix the validation layer messages. That will be a separate PR.)

### DIFF
--- a/piet-gpu-hal/src/vulkan.rs
+++ b/piet-gpu-hal/src/vulkan.rs
@@ -1,10 +1,13 @@
 //! Vulkan implemenation of HAL trait.
 
-use std::ffi::CString;
+use std::borrow::Cow;
+use std::ffi::{CStr, CString};
 use std::sync::Arc;
 
+use ash::extensions::ext::DebugUtils;
 use ash::version::{DeviceV1_0, EntryV1_0, InstanceV1_0};
 use ash::{vk, Device, Entry, Instance};
+use once_cell::sync::Lazy;
 
 use crate::Error;
 
@@ -12,8 +15,9 @@ pub struct VkInstance {
     /// Retain the dynamic lib.
     #[allow(unused)]
     entry: Entry,
-
     instance: Instance,
+    _dbg_loader: Option<DebugUtils>,
+    _dbg_callbk: Option<vk::DebugUtilsMessengerEXT>,
 }
 
 pub struct VkDevice {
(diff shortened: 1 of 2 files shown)


raphlinus:
The timeout is set to 100ms. If you have a very slow graphics chip and a fairly challenging scene, it might be simply exceeding it. That can be changed (to, say, 1s) in line 353 of vulkan.rs.

Timeouts are also likely when buffer contents are corrupted for whatever reason, as the shaders loop over whatever inputs are available.

raphlinus:
One other quick feedback: I don't have this layer installed (my Vulkan SDK is old, so I have `VK_LAYER_LUNARG_standard_validation` instead; I'm installing a new one). If the layer is not present, it should gracefully warn and continue, rather than panicking as it does now:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: VkError(ERROR_LAYER_NOT_PRESENT)', piet-gpu\src\main.rs:91:20
```

piet-gpu-hal/src/vulkan.rs
```diff
@@ -70,18 +124,74 @@ impl VkInstance {
...
+                .enumerate_instance_layer_properties()?;
+            let lyrs = LAYERS.iter().filter_map(|&lyr| {
+                exist_lyrs
+                    .iter()
+                    .find(|x|
+                        CStr::from_ptr(x.layer_name.as_ptr()) == lyr
+                    )
+                    .map(|_| lyr.as_ptr())
+                    .or_else(|| {
+                        println!("Unable to layer: {}, have you installed the Vulkan SDK?", lyr.to_string_lossy());
```
raphlinus, line 138: "Unable to find layer" perhaps?

piet-gpu-hal/src/vulkan.rs
```diff
@@ -70,18 +124,74 @@ impl VkInstance {
...
+                        None
+                    })
+            }).collect::<Vec<*const i8>>();
```
raphlinus, line 141: The type is `std::os::raw::c_char`, which is not necessarily the same as `i8` - this may cause a type error on platforms that have unsigned char. In this particular case you can write `collect::<Vec<_>>()` which I think is idiomatic.

piet-gpu-hal/src/vulkan.rs
```diff
@@ -70,18 +124,74 @@ impl VkInstance {
...
+                    .iter()
+                    .find(|x|
+                        CStr::from_ptr(x.extension_name.as_ptr()) == ext
+                    )
+                    .map(|_| ext.as_ptr())
+                    .or_else(|| {
+                        println!("Unable to find extension: {}, have you installed the Vulkan SDK?", ext.to_string_lossy());
+                        None
+                    })
+            }).collect::<Vec<*const i8>>();
```
raphlinus, line 156: as above

piet-gpu-hal/src/vulkan.rs
```diff
@@ -61,6 +65,56 @@ pub struct QueryPool {
...
+    } else {
+        CStr::from_ptr(callback_data.p_message).to_string_lossy()
+    };
+
+    println!(
+        "{:?}:\n{:?} [{} ({})] : {}\n",
+        message_severity,
+        message_type,
+        message_id_name,
+        &message_id_number.to_string(),
```
raphlinus, line 95: Can be just `message_id_number`, doesn't need to be converted into a string (or have its reference taken).

piet-gpu-hal/src/vulkan.rs
```diff
@@ -61,6 +65,56 @@ pub struct QueryPool {
...
 pub struct MemFlags(vk::MemoryPropertyFlags);
 
+#[cfg(debug_assertions)]
+unsafe extern "system" fn vulkan_debug_callback(
+    message_severity: vk::DebugUtilsMessageSeverityFlagsEXT,
+    message_type: vk::DebugUtilsMessageTypeFlagsEXT,
+    p_callback_data: *const vk::DebugUtilsMessengerCallbackDataEXT,
+    _user_data: *mut std::os::raw::c_void,
+) -> vk::Bool32 {
+    let callback_data = *p_callback_data;
```
raphlinus, line 75: Personally I'd write this as `&*p_callback_data`, to avoid making a copy. But that makes very very little difference :)

PR #6, opened 2020-05-01
Swapchain support

msiglreith:
Add support for printing the resulting image in a window - currently winit.

## Implementation details
- Most of the WSI related implementation is not abstracted as it's difficult to come up with a good model here, all APIs do it completely different.
- Kernel 4 writes out the image to a separate image as compute write access to swapchain image is not universally supported
- Using a classic ringbuffer approach to avoid stalls between frames
- Swapchain is half-res of the created image as it's currently too large to fit on a non-4k screen -> blitted with a linear filter


msiglreith:
Split of `main.rs` into `cli` and `winit` binaries for different execution modes

piet-gpu-hal/src/vulkan.rs
```diff
@@ -168,7 +208,7 @@ impl VkInstance {
                 None,
             )?;
 
-            let (_dbg_loader, _dbg_callbk) = if cfg!(debug_assertions) {
+            let (_dbg_loader, _dbg_callbk) = if false {
```
raphlinus, line 211: Is this intentional?

piet-gpu-hal/src/vulkan.rs
```diff
@@ -281,25 +412,127 @@ impl crate::Device for VkDevice {
...
+    }
+
     /// This creates a pipeline that runs over the buffer.
     ///
     /// The descriptor set layout is just some number of buffers (this will change).
     unsafe fn create_simple_compute_pipeline(
         &self,
         code: &[u8],
         n_buffers: u32,
+        n_images: u32,
```
raphlinus, line 512: Docstring could be updated.

msiglreith:
Thanks for the review, resolved the mentioned points!

PR #15, opened 2020-05-28
Checkpoint of old piet-gpu architecture

raphlinus:
This is the best version of the old piet-gpu architecture. For reference, it renders the tiger in 744μs on a laptop GTX 1060. This is pretty good. Performance is not as good on other inputs though, such as paris-30k and paper-1.

There may be some things in here worth preserving. For one, there's an extension to piet-gpu-hal to allow setting the subgroup size. The head doesn't actually use this (it's gone back to threadgroup shared local memory).

Likely this PR will be closed rather than merged, in favor of the sort_middle branch, which appears to be more promising.


raphlinus:
I was keeping this open because I wasn't sure the tree would still be accessible when the PR is closed, but apparently it is. Or, if not, no huge loss at this point.

PR #18, opened 2020-06-04
Format code using 'cargo fmt'

Atul9:


raphlinus:
Thanks for this, but I'm going to close. I'm not actively developing the master branch, development is currently in "sort_middle" and "sorta." Locally I have a *lot* of branches with various experiments, some of which are quite valuable (one example is generating trace data which can then be visualized with chromium tracing). Having formatting changes would make it more difficult to juggle these. I might create an archive version with all these local branches.

When I am no longer in experiment mode and am reasonably satisfied with the code, I'll do a cleanup which will include rustfmt. Again, thanks for your interest.

PR #24, opened 2020-06-28
Shader documentation and a slight cleanup

msiglreith:
A bit of documentation while reading through most of the shader code, I hope this helps others as well (:


piet-gpu/shader/binning.comp
```diff
@@ -72,17 +77,17 @@ void main() {
     }
     int x0 = 0, y0 = 0, x1 = 0, y1 = 0;
     float my_right_edge = INFINITY;
-    bool crosses_edge = false;
+    // bool crosses_edge = false;
```
msiglreith, line 80: the right_edge code doesn't seem to used right now, therefore I commented out the corresponding parts

raphlinus: Yes, thanks. Probably makes sense to do more cleanup to actually remove this.

PR #27, opened 2020-08-09
Clean up bits of right edge tracking logic left over from sort-middle.

bzm3r:
Changes made:

1) removed `min_fill` related logic in `elements.comp`
2) removed `state` related logic in `binning.comp`
3) renamed `tile_ix` to `part_ix` in `elements.comp`.

Performance regressions/improvements: none that I have noticed on my machine. Both element kernel and binning kernel have the same performance. 

Maybe there will be an improvement to be gained by reducing the size and stride of the state buffer, now that we do not need memory for tracking right edge data? I have left that alone for now, since the structure of the state buffer is not totally clear to me yet.


piet-gpu/shader/binning.comp
```diff
@@ -15,11 +15,6 @@ layout(set = 0, binding = 0) buffer AnnotatedBuf {
     uint[] annotated;
 };
 
-// This is for scanning forward for right_edge data.
```
raphlinus, line 18: It should be possible to remove the descriptor, as we no longer need to give the binning kernel access to the state buffer. Since this is a cleanup, interested in doing that as well?

bzm3r: This should be done now!

bzm3r:
Yes, I do. I'll hit the button.

PR #31, opened 2020-08-29
compute IMAGE_WIDTH and IMAGE_HEIGHT; remove dead code from setup.h

eliasnaur:


bzm3r:
This works for me, and I'm merging it after cross-checking with Raph.

PR #32, opened 2020-09-13
Clean up, add maximum workgroup size control

eliasnaur:
First commit deletes unused code, the second adds a control for maximum workgroup size.


raphlinus (approved):
Yes, this looks good, thanks. I was slightly surprised to see some of the cleanup of unused code because I thought it had happened in an earlier pass, but it's good it's happening now.

Issue #30, opened 2020-08-29
Path intersection (aka SVG clipPath)


eliasnaur:
I'd like to generalize path+fill to stack-of-paths+fills such that fills only cover the intersection of the paths in a stack. Would that be feasible with the current architecture? Would you be interested in having that work in piet-gpu?

An efficient algorithm that converts intersecting paths into a single path would be a nice alternative, in which case clipPath would be a pre-process step.

raphlinus:
Yes, I'd very much like to have this functionality; it's in the Piet imaging model and I would consider it fundamental to 2D graphics.

There are a few ways to do this. Doing it the second way using path ops is appealing, but has two major challenges. First is that robust, performant path ops are *very* difficult, largely because of numerical robustness issues. (They make #23 look like child's play). I'd estimate that standing up a new path op library from scratch is about a one year project (take a look at the commit history of Skia path ops if you're having trouble believing me). The second is that the preprocessing might have to be redone. For example, take a filled path inside a rounded rect, but the inner path is translating (for example because of a scrolling gesture). The path computation would need to be redone. If those paths are text, that's a serious performance killer. I'm also not sure if it's practical do to path ops GPU-side at all.

So I prefer doing it with blending. Another major reason to prefer blending is that the same infrastructure can be applied to do a bunch of other blending effects, such as group alpha and the non Porter-Duff blending modes.

Within blending, there are a few other choices. Pathfinder relatively recently solved this problem, and I think did it a little differently than what I have in mind (I discussed it with Patrick a bit but didn't review it carefully and don't remember all details). I *strongly* recommend reviewing that solution, so that there's a conscious choice which way to go.

Here's a rough outline of how I was envisioning doing it.

Instead of working on a single pixel, the fine rasterizer maintains a stack of pixels, and an alpha mask value. Having a clip in the scene translates into a push, rasterization of the clip path into the alpha mask channel, then the contents of the clip, and finally a pop. The pop operation does Porter-Duff compositing. A challenge here is storage space for the stack. Also what happens the stack depth is exceeded?

To be consistent with the existing scene representation (a sequence of fixed-size elements), I'm imagining extending that alphabet with similar operations: push, render path to alpha, pop. If it weren't for the pesky problem of bounding boxes, the coarse raster pipeline could just pass those right through.

[...]

PR #34, opened 2020-10-09
implement clip paths

eliasnaur:
Tested in a local branch of Gio. I tried to adding a simple clip seen below, but got stuck on the RenderContext trait being defined in another repository (crate?). It's my first attempt at writing Rust and figured you don't want this monkey-patched in anyway.

```
diff --git piet-gpu/src/lib.rs piet-gpu/src/lib.rs
index 6177b3e..01c76a9 100644
--- piet-gpu/src/lib.rs
+++ piet-gpu/src/lib.rs
@@ -49,14 +49,12 @@ pub fn render_scene(rc: &mut impl RenderContext) {
         let circle = Circle::new(center, radius);
         rc.fill(circle, &color);
     }
-    /*
     let mut path = BezPath::new();
     path.move_to((100.0, 1150.0));
     path.line_to((200.0, 1200.0));
     path.line_to((150.0, 1250.0));
     path.close_path();
-    rc.fill(path, &Color::rgb8(128, 0, 128));
-    */
+    rc.fill_mask_inv(path, 1.0);
     rc.stroke(
         Line::new((100.0, 100.0), (200.0, 150.0)),
         &Color::WHITE,
diff --git piet-gpu/src/render_ctx.rs piet-gpu/src/render_ctx.rs
index c3e1dd7..7c38992 100644
--- piet-gpu/src/render_ctx.rs
+++ piet-gpu/src/render_ctx.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 
 use piet_gpu_types::encoder::{Encode, Encoder};
 
-use piet_gpu_types::scene::{CubicSeg, Element, Fill, LineSeg, QuadSeg, SetLineWidth, Stroke};
+use piet_gpu_types::scene::{CubicSeg, Element, Fill, LineSeg, QuadSeg, SetLineWidth, Stroke, FillMask};
 
 use piet::kurbo::{Affine, PathEl, Point, Rect, Shape};
 
@@ -138,6 +138,14 @@ impl RenderContext for PietGpuRenderContext {
         }
     }
 
+    fn fill_mask_inv(&mut self, shape: impl Shape, mask: f64) {
+        let path = shape.to_bez_path(TOLERANCE);
+        self.encode_path(path, true);
+        let mask = mask as f32;
+        self.elements.push(Element::FillMaskInv(FillMask { mask }));
+        self.path_count += 1;
+    }
+
     fn fill_even_odd(&mut self, _shape: impl Shape, _brush: &impl IntoBrush<Self>) {}
 
     fn clip(&mut self, _shape: impl Shape) {}
```


raphlinus:
Don't worry too much about the piet implementation; if the gpu side is good, I can take care of that. I've skimmed this and have some idea what's going on, but it would help me a lot to get answers to these basic questions, if for no other reason so I can understand your intent better:

* When there's a clip, what's the scope of it? Specifically, when does it end?

* It looks like on the GPU side there's only one clip at a time (but I'm not sure about this, as I see some mixing functionality in the fine rasterizer). Is the intent to support a more limited imaging model that doesn't do nested clips, or perhaps the idea is to intersect paths CPU-side?

* I'm a little confused by fill_mask and fill_mask_inv; I don't remember seeing anything quite like this in any graphics API (though it's reminding me just a bit of knockout groups in the PDF 1.4 imaging model). It seems like fill_mask_inv with alpha = 0.0 is closest to the traditional clip operation - the mask becomes 0.0 outside the clip path and 1.0 inside it (with aa at the edges). What's the use case for fill_mask?

* I see wiring up of "backdrop" plus the inverse stuff, but I also have in mind another potential optimization (which is not applicable in the ordinary fill case) - when the entire tile becomes 0.0, as would be especially be the case for fill_mask_inv when there are no path segments and backdrop = 0, then drawing operations can be suppressed while that mask is in effect. It's totally fine for correctness not to do this optimization, so easy for me to imagine it being applied in the future.

I have some questions/thoughts about bounding boxes, but those can wait until later. The workaround of treading inverse mask fills as having infinite bounding box is a reasonable interim stage. I'd *like* to do bbox computation on GPU, but it might be more pragmatic to do it on CPU and pass it along during encoding.

It's gratifying to see this move forward, even if the approach might not be exactly the way I was imagining doing it.

eliasnaur:
On Sun Oct 11, 2020 at 17:27, Raph Levien wrote:
> * When there's a clip, what's the scope of it? Specifically, when does it end?
>

The clip operations (fill_mask and fill_mask_inv) are both tied to a path, just
like the fill operation. In fact, fill_mask is like fill except that fill_mask
fills the pixel mask, not the pixel color. A fill_mask has the same area of effect
(and thus bounding box) as the fill: it only touches pixels covered by path, with
AA at the edges.

The fill_mask_inv operation is the same as fill_mask except it fills everything
*not* inside its path. A fill_mask_inv conceptually takes up the entire screen
except for areas inside its path. So its bounding box is infinite, but of course
that can be tightened to the union of the bounding boxes of affected operations
as you hinted at in #30.

> * It looks like on the GPU side there's only one clip at a time (but I'm not sure about this, as I see some mixing functionality in the fine rasterizer). Is the intent to support a more limited imaging model that doesn't do nested clips, or perhaps the idea is to intersect paths CPU-side?
>

The main idea is to support SVG clip paths. A clip path can be converted to a
fill_mask_inv with alpha = 0.0. Nested clip paths work as you expect: each
fill_mask_inv(0.0) acts like set intersection, by setting its outside area to
0.0 and leaving the inside alone.

> * I'm a little confused by fill_mask and fill_mask_inv; I don't remember seeing anything quite like this in any graphics API (though it's reminding me just a bit of knockout groups in the PDF 1.4 imaging model). It seems like fill_mask_inv with alpha = 0.0 is closest to the traditional clip operation - the mask becomes 0.0 outside the clip path and 1.0 inside it (with aa at the edges). What's the use case for fill_mask?
>

I didn't model an existing graphics API, sorry. My focus was to fit in the
functionality of clip paths with the existing concepts of piet-gpu (fills,
paths).

The use for fill_mask is (at least) twofold

- For resetting the mask when "popping" a set of clip paths. To clear the
mask, use a fullscreen rectangular path with fill_mask(1.0). It's efficient
because it is translated to a solid fill per tile.
- For clipping to the union of paths, unlike the intersecting SVG clip paths.
To clip to a union of paths, (1) clear mask, (2) use a fill_mask_inv(0.0) for the
first path in the set, (3) use fill_mask(1.0) for the rest of the paths.

I considered using an explicit stack of pixels as you suggested in #30, but
decided it wasn't worth the extra complexity, at least for my use cases. A
stack allows you to "pop" individual clip paths, whereas my approach only
allows you to clear the mask, and redo the clip path stack up to the point
you need it.

> * I see wiring up of "backdrop" plus the inverse stuff, but I also have in mind another potential optimization (which is not applicable in the ordinary fill case) - when the entire tile becomes 0.0, as would be especially be the case for fill_mask_inv when there are no path segments and backdrop = 0, then drawing operations can be suppressed while that mask is in effect. It's totally fine for correctness not to do this optimization, so easy for me to imagine it being applied in the future.
>

Yes, that's indeed an obvious optimization. I left it out until after the approach
was settled, and more mundane features are added to piet-gpu (texture support, dynamic
width/height etc).

> I have some questions/thoughts about bounding boxes, but those can wait until later. The workaround of treading inverse mask fills as having infinite bounding box is a reasonable interim stage. I'd *like* to do bbox computation on GPU, but it might be more pragmatic to do it on CPU and pass it along during encoding.
>

I'm not sure it's worth it to do on the CPU, except if the application has
special knowledge of the geometry. Wouldn't CPU bounding box calculations need
to duplicate the work of elements.comp? On the other hand, infinite bounding
boxes only waste a solid fill per tile per fill_mask_inv.

Bounding box intersection on the GPU would be nice, of course.

raphlinus (approved):
Right, got it. I'm going to approve this so you can keep developing, but the long term plan is for nesting and bbox calculation to be handled on gpu. I have ideas how to do that, but it's a lot more involved, so I don't want to block the clipping you need on that.

Thanks!

PR #35, opened 2020-10-14
Limit spinning in elements.comp

eliasnaur:
The first commit simplifies map_element, the second implements the progress guarantee.


raphlinus (approved):
Fantastic! This fixes a real correctness issue with the prefix sum approach.

And apologies for not reviewing earlier, I'm juggling too many things and sometimes drop a ball.

Issue #38, opened 2020-11-16
bitmap and gradients support


eliasnaur:
The pipeline supports uniform colors. I'd like to work on expanding that to accept bitmaps, and later on gradients. If you have any thoughts on the design, let me know. If not, I'll come up with something to start the discussion.

raphlinus:
As it turns out, I have been thinking about this quite a bit, and did some explorations over the weekend. I'll write up my detailed thoughts later today.

raphlinus:
As promised, here are my thoughts on adding images to the piet-gpu pipeline.

A few general thoughts, then I'll get into specifics. One characteristic of a retained-mode pipeline like piet-gpu is that *all* resources have to be resident at the time rendering starts. This is in some contrast to using rasterization to draw images; you can load an image, do the draw call, then the image no longer has to be resident after that. The images can only be freed after the pipeline completes (ie after the fence for the command buffer is signaled). This will require some lifetime tracking on CPU-side objects that hold handles to GPU-side texture resources.

Also, because piet-gpu is using the equivalent of an ubershader, the descriptor set provided to the fine rasterizer has to be able to access all images. There are two fundamental approaches to this. One is to use an array of textures, and build the descriptor set provided to the fine rasterizer with this array. There's a [good blog](http://kylehalladay.com/blog/tutorial/vulkan/2018/01/28/Textue-Arrays-Vulkan.html) on that technique. The other is to use atlasing techniques, so there is a smaller (likely bounded) number of textures, the images are copied into subregions of those textures, then the description of an image consists of a uv rectangle within one of those textures.

[...]

PR #43, opened 2020-11-20
Start adding support for creating images

raphlinus:
This is still WIP, focused on creating image resources and making them
available GPU-side.

Progress toward #38


raphlinus:
I have a local branch where I tested this, if people are curious to see. It just adds a single additional image for use by k4; it doesn't do the descriptor array magic as described in the linked issue, and is also wired as a storage image, not a sampler, so bilinear filtering isn't available. But it does validate that the created image is available for use in the kernel.

Issue #36, opened 2020-11-14
Proper nested clipping


raphlinus:
This issue contains a detailed design of how to properly add clipping to piet-gpu. It is a followup to #30, which represents a more limited form of clipping which might be useful in some UI contexts, but doesn't efficiently capture the general clipping problem. The PostScript-derived imaging model entails clips being nested to arbitrary depth, with the nesting defined by `save` and `restore` operations. The arbitrary depth is challenging for a GPU, which is generally optimized for fixed-sized grains of work. Thus, this design contains profound changes across the entire pipeline.

A central concern is bounding box tracking. The invariant associated with a bounding box of a drawing operation is that pixels outside the bounding box are not affected. In the case of a clip operation, it is tempting to take the bounding box of the clip mask, but this is inadequate. Pixels outside the bounding box of the clip mask *are* affected, in particular all drawing operations inside the clip become no-ops.

The invariant is restored when the bounding box of a clip operation is taken as the union of the bounding boxes of the drawing operations inside.

## Encoding

A long-term goal of piet-gpu is to make the CPU-side encoding process as lightweight as possible. To this end, there is a nontrivial bbox tracking mechanism in the encoding stage of the pipeline, eliminating the burden to track bboxes for stroke and fill elements during CPU-side encoding. An additional potential benefit of delegating bbox tracking to the GPU is that display lists can be assembled by binary string concatenation of encoded display list fragments.

However, maintaining those properties in the face of nested clips is nontrivial, therefore we propose to regress the light weight of CPU encoding, with the hope of restoring it in a later cycle. In this design, we *will* track the bbox of each drawing element. When encoding a clip, we include the clip's bbox (which is the union of drawing elements contained in the clip) in a "push clip" element, and also in the corresponding "pop" element, which is encoded at the time of the `restore` method is called on the render context.

Note that bbox tracking also entails tracking of transforms, another task that was entirely subsumed by element processing in the previous design.

## Element processing

Element processing is barely affected by this design; the main requirement is passing through the encoded bbox.

Element processing also potentially computes the bbox of the clip path. I *believe* it makes sense to simply discard that, and use the encoded bbox.

## Binning

Binning is barely affected; the main requirement is to use the clip's bbox for binning purposes. (as opposed to the clip path's bbox, if both bboxes are in fact preserved by the element processing stage).

## Coarse rasterization

[...]

Issue #40, opened 2020-11-18
Dynamic GPU memory management


eliasnaur:
piet-gpu allocates large GPU memory for its working memory. I'd like for memory use to be proportional to scene complexity, or even a constant low amount with several passes over complex scenes. I'm particularly interested in a multi-pass scheme because I want to run piet-gpu shaders on the CPU at some point, where caching unchanged scenery is vital.

Again, if you have thoughts in this area, I'm all ears; otherwise I'll see if I can work something out.

raphlinus:
As it turns out, yes, I have been thinking about this. I think there are several overlapping issues here:

* Estimation of required memory usage for each buffer
* Readback from GPU to allocate/schedule future stages
* Retry / partial progress on failure

Each buffer is a little different in how these should be handled. I'll start with the ptcl as it has one of the best stories. I think it makes sense for that to have a modest sized allocation (possibly using scene complexity and available GPU memory as inputs to a heuristic), so it's certainly possible the entire ptcl doesn't fit.

The way I'd handle that is to have a scoreboard indicating the progress at each tile - the representation I have in mind is 1 + the pass number at which the tile completed, or 0 if it hasn't. The ptcl writer simply bails if the atomic bump allocation fails, leaving the scoreboard untouched. The fine raster only proceeds if the pass number matches. Therefore, you get the whole surface rendered by repeating the coarse raster (ptcl generation) and fine raster steps, until all tiles complete.

That in turn requires more sophistication in dispatching the compute kernels, as you have to know how many times to iterate. Basically, when dispatching the coarse+fine raster part of the pipeline, it needs to wait on the fence for the cmd buffer to complete, then do a readback of info indicating whether another pass is needed (the easiest way to do this is the value of the bump allocator after coarse raster).

How should it wait? There are two choices, one is to integrate an async runtime, the other is to dedicate a thread to the renderer and do a blocking wait on the fence. The Rust community is increasingly enamored with the former approach, but I think the latter might be better all around. Another minor point while I'm thinking about synchronization is that you can improve CPU/GPU overlap for this by doing two separate cmdbuf submissions, one for the coarse raster and one for the fine, then wait on the fence for the *first* one before deciding whether to schedule another pass. If it's complete, it can submit an empty cmdlist that signals the semaphore for presentation.

Btw [A High-Performance Software Graphics Pipeline Architecture
for the GPU
[...]

Issue #23, opened 2020-06-15
Rendering glitches on path filling


oleid:
With great interest, I read your blog post on the new code in `piet-gpu` and gave it a try on my Radeon RX 470 GPU:

```
08:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] (rev cf)
```

I got the example images from:
[1] http://w3.impa.br/~diego/projects/GanEtAl14/

It would seem there are a few rendering glitches, most notably:

1. Some images are mirrored
2. A few font glitches

Please note, I used the default `radav` driver on ArchLinux. There is an experimental new shader backend, which is faster, but has the same output (or at least I didn't notice any difference). It can be activated by exporting `export RADV_PERFTEST=aco`.

Also, there is another vulkan driver for AMD, namely `amdvlk`. That seems to be a little slower for most operations, but the output seems to be the same as well.

How where the benchmarks from the blog post performed? Merely run the CLI tool without any additional options? Given the tiger svg from this repo, my old RX 470 seems to be faster than a 1060, which seems odd.

paper-1
======
```
cargo run --release --bin=cli --  inputs/svg/paper-1.svg
    Finished release [optimized] target(s) in 0.06s
     Running `target/release/cli inputs/svg/paper-1.svg`
parsing time: 44.684545ms
flattening and encoding time: 5.586778ms
scene: 104322 elements
Element kernel time: 0.258ms
Tile allocation kernel time: 0.016ms
Coarse path kernel time: 0.199ms
Backdrop kernel time: 0.074ms
Binning kernel time: 0.106ms
Coarse raster kernel time: 0.070ms
Render kernel time: 0.220ms
```
![paper-1](https://user-images.githubusercontent.com/3369899/84707742-8a950f80-af5f-11ea-8d07-c5d7fffcae29.png)

At scale=2
---------

```
cargo run --release --bin=cli -- --scale 2  inputs/svg/paper-1.svg
    Finished release [optimized] target(s) in 0.06s
     Running `target/release/cli --scale 2 inputs/svg/paper-1.svg`
parsing time: 43.184138ms
flattening and encoding time: 3.919508ms
scene: 104322 elements
Element kernel time: 0.259ms
Tile allocation kernel time: 0.014ms
Coarse path kernel time: 0.157ms
Backdrop kernel time: 0.055ms
Binning kernel time: 0.024ms
[...]

PR #44, opened 2020-11-21
First cut at clipping

raphlinus:
This is a significant amount of progress towards #36. Bounding boxes for clips are accumulated (in world coordinates, see that issue for more details) CPU-side and encoded in the BeginClip and EndClip elements. These are carried through the pipeline in a straightforward way, then there is a blend stack in the fine rasterizer.

Here are the cleanups not yet done:

* The blend stack has finite size. Deal with the overflow (by writing to a scratch buffer in global memory)
* Clean up the older FillMask mechanism.
* Optimize all-clear and all-solid clip tiles.

This PR also encodes transforms (they have always worked, just weren't encoded).


piet-gpu/shader/kernel4.comp
```diff
@@ -27,6 +27,8 @@ layout(rgba8, set = 0, binding = 2) uniform writeonly image2D image;
 #include "ptcl.h"
 #include "tile.h"
 
+#define BLEND_STACK_SIZE 4
```
eliasnaur, line 30: Will you lift this restriction before replacing FillMask?

raphlinus: I think so, but I'm not 100% sure when. On my AMD 5700 XT card, the static limit can be increased quite a bit without seriously impacting performance (utilization goes down as register pressure goes up). Using a scratch buffer in global memory can be done in two stages - malloc-only (using an atomic bump counter, as is currently done for dynamic allocations), and a second malloc/free stage.

eliasnaur: Can the FillMask stopgap co-exist with the new clipping? I'd hoped not to have hard limits such as this when adding piet-gpu as a Gio backend. I'll probably take a stab at #40, but I think this limit is worse.

piet-gpu/shader/coarse.comp
```diff
@@ -224,20 +220,23 @@ void main() {
...
             uint width = sh_tile_width[el_ix];
             uint x = sh_tile_x0[el_ix] + seq_ix % width;
             uint y = sh_tile_y0[el_ix] + seq_ix / width;
-            Tile tile = Tile_read(TileRef(sh_tile_base[el_ix] + (sh_tile_stride[el_ix] * y + x) * Tile_size));
-            // Include the path in the tile if
-            // - the tile contains at least a segment (tile offset non-zero)
-            // - the tile is completely covered (backdrop non-zero)
-            // - the tile is not covered and we're filling everything outside the path (backdrop zero, inverse fills).
-            bool inside = tile.backdrop != 0;
-            bool fill = tag != Annotated_FillMaskInv;
```
eliasnaur, line 239: I think this check needs restoring.

piet-gpu/shader/kernel4.comp
eliasnaur: Thank you very much. If you decide to leave it, I'll take a look after #23, #38, #40 after which Gio has all it needs for now. The existing Gio renderer tests can be re-used and expanded to cover piet-gpu, and I'll be much more comfortable working on enhancements and bugs.

eliasnaur:
Confirmed that FillMask still works with this branch.

raphlinus:
Good to know. I'll take that as a go-ahead to commit when I feel it's appropriate. (FYI, generally the strategy I use is that my local branch is one commit ahead. When I'm ready to PR from that, I merge the one behind it into master). I've started the scratch buffer (malloc-only) and do not think it will be very hard. I'll also comment on the dynamic allocation issue.

PR #45, opened 2020-11-23
Scratch buffer for clip stack

raphlinus:
We keep a small window of the clip stack in registers in the fine
rasterization kernel, and when that window is exceeded, spill to global
memory, so the clip stack can be unbounded.


eliasnaur:
Thanks for working on this. Feel free to delete FillMask et al, or leave it to me.

PR #46, opened 2020-11-25
Progress towards images

raphlinus:
This patch passes a dynamically sized array of textures to the fine
rasterizer.

A bunch of the low level Vulkan stuff is done, but only enough of the
shaders and encoders to do minimal testing. We'll want to switch from
storage images to sampled images, track the actual array of textures
during encoding, use that to build the descriptor set (which will need
to be more dynamic), and of course run image elements through the
pipeline.

As part of the work, I've moved the creation of pipelines and descriptor
sets to a builder pattern, providing convenience methods for very simple
pipelines.

Progress towards #38


raphlinus:
There are a few things to say here.

I'm using runtimeDescriptorArray and related features, which will definitely cause compatibility issues on lower tier devices. However, I think it's clearly the right thing to do on newer devices, and makes life easier. I don't think the use of texture arrays is a total dead end on older devices. There, I think the thing to do is instantiate a shader with the number of textures (I'm pretty sure this can be done with a specialization constant, something that we're currently lacking infrastructure for but we'll want to do). This can be rounded up some, I don't think it needs to be exact (that might reduce the amount of work rebuilding the pipeline when changing the specialization constant), in which case we'd want to fill in the unused slots in the descriptor set with some dummy value. I've seen a [blog](http://roar11.com/2019/06/vulkan-textures-unbound/) which suggests duplicating one of the textures.

Of course, on very low tier devices, with a limit on descriptor set size that might be exceeded by the number of images in the scene (or dx11, which doesn't do arrays), we'll need to do atlasing, but I think we can put that off a while.

PR #50, opened 2020-12-01
Avoid inconsistent line-tile intersections

eliasnaur:


raphlinus (approved):
I haven't reviewed this carefully, but from what I see it seems reasonable. The gold standard would be accompanying reasoning *why* it's more robust, but I haven't been setting a good example at all with that. I have tested it and it seems to work, with no measurable performance penalty on my system.

If need be, there can be another pass, which might also take a wrench to performance. One thing to explore is whether we can reduce memory bandwidth by using f16 for the tile values (this would only work if everything is relative to tile origin, otherwise loss of precision would be a problem).

But this looks like good progress for now, and thanks! I think we'll get a chance to exercise it with glyph outlines.

Btw, I'll merge my 2 pending CL's after this goes in (I expect only simple merge conflicts, largely due to spv recompilation).

eliasnaur:
I tried to reduce the loop to something I could reason about, but gave up, sorry. Reducing to f16 values would be a great practical test for the robustness; glyph outlines as well.

PR #52, opened 2020-12-09
Fetch correct query pool

raphlinus:
It should fetch the last query pool, and was off by one. That worked on
my machine (windows), but did cause panics.


zjiekai:
Merged into master, tested on ubuntu.  It works now.

PR #55, opened 2020-12-10
elements.comp: tighten state size calculations

eliasnaur:


raphlinus (approved):
Yup, this looks good. Ideally there would be a comment explaining the layout of the state buffer. Alternatively, it might be nice to autogenerate all of the array access logic, but I'm not sure that's a win for accesses entirely within one file.

PR #56, opened 2020-12-11
add SPDX license headers

eliasnaur:


raphlinus (approved):
Looks good. Also, kinda astonishing just how much cleanup there is from the various architectural changes. Makes me wonder if we've still got it all.

PR #59, opened 2020-12-12
path_coarse.comp: fix intersection inconsistencies, take 2

eliasnaur:
The previous attempt to fix inconsistent intersections because of floating
point inaccuracy[0] missed two cases.

The first case is that for top intersections with the very first row would fail
the test

tag == PathSeg_FillCubic && y > y0 && xbackdrop < bbox.z

In particular, y is not larger than y0 when y0 has been clipped to 0.

Fix that by re-introducing the min(p0.y, p1.y) < tile_y0 check that does work
and is just as consistent.

The second case is that the tracking left intersections in the [xray, next_xray]
range of tiles may fail when next_xray is forced to last_xray, the final xray value.

Fix that case by computing next_xray explicitly, before looping over the
x tiles. The code is now much simpler.

Updates #23

[0] https://github.com/linebender/piet-gpu/commit/29cfb8b63edc28517b16e3ba1da7790e360ed557

Signed-off-by: Elias Naur <mail@eliasnaur.com>


eliasnaur:
Replaced by https://github.com/linebender/piet-gpu/pull/61.

PR #61, opened 2020-12-24
Work branch

eliasnaur:
It's become unwieldy to maintain separate branches for each change. In particular, separate branches interacts badly with GitHub's PR work-flow. Going forward, I'm having one open PR for work not yet upstreamed. Each commit serves as a separate change, motivated by the commit description.

I try to order commits by controversy, but feel free to review and approve them in any order; I'll take care of rebasing and re-ordering around not-yet approved or rejected changes.


raphlinus:
Sorry for the delays in reviewing this. I took time off and now am having difficulty juggling things, especially finding a large chunk of time to review the whole thing. I think I'll try doing one commit at a time, in oldest-to-newest commit order.

raphlinus:
I'll accept a2a2d12c5da24ee245512eabd3ba508eb73f02aa, and appreciate the detailed description in the changelog. Hopefully this is reliable now, but to consider the logic bulletproof I would want either extensive automated testing or a proof. I haven't gone over the logic line by line, but it's in line with what I'd expect.

Thanks, this stuff is quite tricky!

eliasnaur:
Thanks. I've added https://github.com/linebender/piet-gpu/commit/a2a2d12c5da24ee245512eabd3ba508eb73f02aa to master (not sure why this pull request is not updated automatically to exclude the commit).

If it's any consolation, the large changes are mostly automatic, dur to much smaller changes in the piet-gpu header generator (consolidating shader buffers, rewriting buffer access with functions with optional bounds checks).

The good news is that this PR implements everything needed for Gio to use piet-gpu as a backend. I'm especially happy with how simple the [driver](https://git.sr.ht/~eliasnaur/gio/tree/main/item/gpu/compute.go) is.

raphlinus:
An update on this, as I've let it slide, apologies for that. I've reviewed the changes up to FillImage and find them generally good. I have notes locally, which I should dust off and post soon.

I haven't dug into the change to image yet but it sounds scary. From a process perspective, it might make sense to land the changes up to, say, 1e9d699b24a372a0b6aa4cd4290f551797198d1c into master, and consider the images as a separate PR.

eliasnaur:
Thanks. I've merged everything except the FillImage change.

raphlinus:
First, sorry for the delayed response. I've meant to get back to you much earlier, but things keep popping up.

As I mentioned, I don't see any serious issues, so am happy to go forward from this patch. A few notes I made:

The sRGB commit ( was 15de5690ad04588fc9dccb50117bf7839a6405c2, seems to have a different commit id now) is a significant slowdown on my machine (AMD 5700 XT). I did a little investigation, and most of the extra time seems to be spent in textureGrad in kernel 4. I haven't dug to the bottom of it, but think it may be relatively benign, that the scene is simply doing more work after the commit. It would be very surprising if the sRGB math itself was a major time sink, and light experimentation suggests that it isn't. Next time I do careful quantitative measurement I would like to get some understanding of what's going on. Until then, it's not urgent.

The calculation of config_buf_host size could use a comment or some method of automatically caclulating it, as currently it's a magic number.

The use of a single memory allocation pool gave me lots to think about. Originally I thought there might be some gains (caching etc) from marking input buffers as readonly. But I don't have any *quantitative* data to back that up, and the ability to run the stages in series without blocking CPU-side on a success/failure condition is likely a much bigger performance issue. So I'm very happy to go forward with the current state.

Thanks again for this work, it goes very deep and I appreciate it. I've put time aside and will be reviewing the other patches carefully in the next day or two, as well as wiring up the Piet drawing trait to the new functionality.

eliasnaur:
> First, sorry for the delayed response. I've meant to get back to you much earlier, but things keep popping up.
> 
> As I mentioned, I don't see any serious issues, so am happy to go forward from this patch. A few notes I made:
> 

I've closed this PR because I've abandoned the FillTexture approach, and believe [FillImage](https://github.com/linebender/piet-gpu/pull/63) is better for several reasons listed in the commit message. I merged the sRGB support with FillImage for my convenience. Let me know if you'd like to review them separately.

> The sRGB commit ( was [15de569](https://github.com/linebender/piet-gpu/commit/15de5690ad04588fc9dccb50117bf7839a6405c2), seems to have a different commit id now) is a significant slowdown on my machine (AMD 5700 XT). I did a little investigation, and most of the extra time seems to be spent in textureGrad in kernel 4. I haven't dug to the bottom of it, but think it may be relatively benign, that the scene is simply doing more work after the commit. It would be very surprising if the sRGB math itself was a major time sink, and light experimentation suggests that it isn't. Next time I do careful quantitative measurement I would like to get some understanding of what's going on. Until then, it's not urgent.
> 

Indeed, textureGrad increased the kernel4 runtime on a simple Gio scene threefold on my Pixel phone, which is one of the reasons I abandoned FillTexture. Another is CPU fallback; without texture sampler, a slightly modified SwiftShader can output a self-contained ELF object file of the entire piet-gpu pipeline.

> The calculation of config_buf_host size could use a comment or some method of automatically caclulating it, as currently it's a magic number.
> 

Got it. The change is merged, but I'll create another change with a comment.

> The use of a single memory allocation pool gave me lots to think about. Originally I thought there might be some gains (caching etc) from marking input buffers as readonly. But I don't have any _quantitative_ data to back that up, and the ability to run the stages in series without blocking CPU-side on a success/failure condition is likely a much bigger performance issue. So I'm very happy to go forward with the current state.
> 

Indeed, the ability to run the entire pipeline to completion without blocking was the main reason I merged the buffers. Another great reason is the lessened buffer space waste.

> Thanks again for this work, it goes very deep and I appreciate it. I've put time aside and will be reviewing the other patches carefully in the next day or two, as well as wiring up the Piet drawing trait to the new functionality.

Sounds great. As I mentioned, I don't plan to merge FillTexture, not even as an addition to FillImage because of the significant drawbacks.

If you have the time, please advice on https://github.com/linebender/piet-gpu/pull/65, which needs Rust-side changes to support the new end-of-path flag. If you approve of the approach, we may want to fold the stroke/fill flag into the tag as well.

Issue #62, opened 2021-02-13
elements.comp scene traversal may leave gaps in paths


eliasnaur:
The fill algorithm implemented by path_coarse.comp assumes that closed paths contain no gaps. Unfortunately, the non-deterministic scene tracersal in elements.comp may introduce gaps in otherwise gap-less paths.

Consider two fill path segments through endpoints `A`, `B`, `C` in a scene fragment with three preceding transforms. The segments are initialized with the identity transform, `I`.

`T₁·T₂·T₃·S(I, A, B)·S(I, B, C)`

Since scene element operations are associative, elements.comp evaluates them in a GPU-friendly but non-deterministic order. In particular, it may use the ordering

`S((T₁·T₂)·T₃, A, B)·S(T₁·(T₂·T₃), B, C)`

Because of numerical imprecision, the path segments may end up with slightly different transformations,

`S(M₁, A, B)·S(M₂, B, C)`

Finally, `M₁·B` is not in general equal to `M₂·B`, leading to a gap in the path that includes the two segments.

To verify my claim, I made the following change,

```
diff --git gpu/shaders/elements.comp gpu/shaders/elements.comp
index a43c270..b5f2342 100644
--- gpu/shaders/elements.comp
+++ gpu/shaders/elements.comp
@@ -259,34 +259,8 @@ void main() {
                     State their_prefix = State_read(state_prefix_ref(look_back_ix));
                     exclusive = combine_state(their_prefix, exclusive);
                     break;
-                } else if (flag == FLAG_AGGREGATE_READY) {
-                    their_agg = State_read(state_aggregate_ref(look_back_ix));
-                    exclusive = combine_state(their_agg, exclusive);
-                    look_back_ix--;
-                    their_ix = 0;
-                    continue;
-                }
+                } 
                 // else spin
-
-                // Unfortunately there's no guarantee of forward progress of other
-                // workgroups, so compute a bit of the aggregate before trying again.
-                // In the worst case, spinning stops when the aggregate is complete.
-                ElementRef ref = ElementRef((look_back_ix * PARTITION_SIZE + their_ix) * Element_size);
-                State s = map_element(ref);
-                if (their_ix == 0) {
-                    their_agg = s;
-                } else {
-                    their_agg = combine_state(their_agg, s);
-                }
-                their_ix++;
-                if (their_ix == PARTITION_SIZE) {
-                    exclusive = combine_state(their_agg, exclusive);
-                    if (look_back_ix == 0) {
-                        break;
[...]

PR #65, opened 2021-02-17
close gaps left by non-deterministic elements.comp monoid

eliasnaur:
This change fixes the gaps described in #62 for Gio.

However, I haven't figured out how to squeeze the FLAG_END_PATH flag into the scene.rs tag encoding. There is already the is_fill flag, but it's encoded as separate Rust enums, and I don't want to add another variant dimension (not-end-of-path vs end-of-path).

Please advice?


raphlinus (commented):
This can't be merged as-is because it breaks Rust rendering. It is possible to pack more flags into the tag word, but this would require nontrivial work. I had started work on it but ripped it out. Alternatively I could hack the code to set the bit. Perhaps as an intermediate stage we could have a separate flags word, then an issue to make the encoding more efficient.

One question I have that's unresolved is how an open stroke is to be encoded.

Longer term I'm considering alternate solutions. One idea is to have a separate pass to generate the transforms, so there is one resolved transform per "transform" element in the input. This may have the desirable side effect of reducing the cost of computing the transform "monoid", as this is likely a large part of the total cost of element transform. I considered fixed point math for the transforms, but this doesn't feel right, especially as I'd imagine fixed point math may be much slower than floating.

I'm still chewing on this, but wanted to get a snapshot of my thoughts across.

raphlinus:
Ok, I've done #69 so I can encode bits into the tag word, something that I think will generally be useful. I also locally patched this to set the flag for the last element on `close_path`, and I get this result:

image

There are two issues here. One is stray segments from stroked paths that aren't closed. This is as expected from reading the patch. Also, there are *more* artifacts related to closed paths not being watertight. I haven't tracked these down either.

So I have questions about how carefully this patch was tested. Were you able to get the tiger to render artifact free?

eliasnaur:
Ugh, that's bad. No, I didn't test the tiger because I didn't figure out a way to pass the end path flag in Rust. This change does work great for all of Gio's tests, even with non-trivial transformations and lots of text. I don't use strokes yet, though, which is why I didn't notice the open path flaw.

Now, I can dig into the tiger rendering glitches if you like, but it sounds to me you've found a better fix than this change. If so, let's just abandon this change and not waste more time on it.

raphlinus:
I don't have a good solution complete, but understand the problem space well enough I think the next major iteration of work on path filling robustness can take on this issue as well as #68. I agree with closing this PR for now, with appreciation for having discovered the problem and explored one avenue to solving it.

PR #67, opened 2021-02-22
kernel4.comp: remove unused command

eliasnaur:


ishitatsuyuki:
Isn't circle and line also unused?

eliasnaur:
> Isn't circle and line also unused?

Removed circle command as well. What line are you referring to?

ishitatsuyuki:
`CmdLine`, apparently it's only in `ptcl.rs`.

eliasnaur:
Thanks, done.

raphlinus:
I'll probably want to bring circle and line back later as optimizations of special shapes, but they're not a priority at all and it's worth measuring how much of a speedup they represent over the general path mechanism.

The one that I *do* expect will have a significant impact is rectangle and probably rounded rectangle. But those can be brought back later.

raphlinus (approved):

PR #69, opened 2021-03-02
Add encoding of flags into tag

raphlinus:
Add a `TagFlags` type. You can write enum variants as `Variant(TagFlags, Content)` and this will encode a u16 into the top 16 bits of the tag word.

Related to #65.


eliasnaur:
Great, I assume this can be used to merge the Stroke* and Fill* variants?

raphlinus:
Yes, we can do that merge, and another thing I'm hoping this can be used for is adding a non-zero/even-odd winding number option.

Issue #70, opened 2021-03-12
Add command for setting fill mode


eliasnaur:
The split between stroking and filling is awkward because there are two versions of many scene commands (StrokeLine/FillLine, StrokeQuad/FillQuad, StrokeCubic/FillCubic, Fill/Stroke), and some don't even support strokes even though they easily could (BeginClip, EndClip, [FillImage](https://github.com/linebender/piet-gpu/pull/63)).

There's now support for [command flags](https://github.com/linebender/piet-gpu/pull/69), that can be used to make stroke vs fill a flag for every supported command.

However, I still find it somewhat awkward to select the correct command (or flag) when constructing scenes. One issue is that you don't specify a stroke/fill flag when constructing paths in Gio; rather the mode is given from the *use* of the path, and there is no reason you can't re-use a path for both a stroke and a fill. Another is that Gio separates clipping from filling, so that even strokes are merely a form of clipping. That model is easier to implement when all commands automatically support both stroking and filling.

So I'd like to propose a SetMode command that sets the filling mode for subsequent commands similar to how SetLineWidth sets the line width for subsequent strokes. In particular, SetMode has a single argument specifying the mode to be either stroking or nonzero-fill. A future even-odd mode naturally fits into SetMode as a third option.

The result is that there will no longer be duplicate commands for stroking and filling, and the user won't have to set a mode flag in every command that needs it. Instead, the elements.comp monoid will automatically fill the mode flags, and subsequent pipeline stages can use the flags as they see fit.

Issue #66, opened 2021-02-20
Investigate tight ptcl allocation


ishitatsuyuki:
The fine raster stage reads commands (ptcl) sequentially, so there should be no penalty from sizing the Cmd enum dynamically. If we add a piet-gpu-types feature where each Cmd only takes as much space as whatever the variant occupies, then it probably can slightly reduce memory traffic.

That said, in the current state most Cmd variants are similarly sized so this probably isn't going to make a big difference.

PR #72, opened 2021-03-15
Work branch

eliasnaur:
Work branch, containing

- Cleanup
- Inconsistent transformation fix (#62).
- sRGB and FillImage support (#38).
- Merging stroke vs fill variants into one variant with flags (preparation for #70)

As usual, my commits are self-contained and you may review and accept or reject commits in any order; I'll handle the rebase and reordering before submitting.


raphlinus:
This looks very interesting, thanks! I'll try to review soon, and apologize for the continued difficulty I'm having in prioritizing this.

eliasnaur:
I've pushed a bunch of commits preparing for #70. They add tag flags support to shader headers, and merge stroke vs fill commands to one command with flags specifying the fill mode.

eliasnaur:
#70 is about generalizing fill mode so that all relevant commands (BeginClip/EndClip/FillImage) support strokes and fills.

A first step i merging CmdStroke and CmdFill:

```
          struct CmdStroke {
              // This is really a Ref<Tile>, but we don't have cross-module
              // references.
              tile_ref: u32,
              half_width: f32,
              rgba_color: u32,
          }
          struct CmdFill {
              // As above, really Ref<Tile>
              tile_ref: u32,
              backdrop: i32,
              rgba_color: u32,
          }
```

Unfortunately, that's not possible without making Cmd_size larger. I'd like to split them into something like

```
          struct CmdFillColor {
              // This is really a Ref<Tile>, but we don't have cross-module
              // references.
              tile_ref: u32,
              rgba_color: u32,
          }
          struct CmdStroke {
              half_width: f32,
          }
          struct CmdNonzero {
              backdrop: i32,
          }
```

And then read either a CmdStroke or CmdNonzero, depending on the flags.

@raphlinus you may have an idea on how to represent this in the generator.

raphlinus:
One quick observation as I start to review in more detail. I get this validation error from running the `cli` binary:

```
VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated (1345731725)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated ] Object 0: handle = 0xe2dc30000000046, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x50363c8d | VkDescriptorSet 0xe2dc30000000046[] encountered the following validation error at vkCmdDispatch() time: Descriptor in binding #3 index 1 is being used in draw but has never been updated via vkUpdateDescriptorSets() or a similar call.
```

I haven't investigated in more detail. It doesn't seem to affect rendering of the tiger (but I haven't tried anything involving images yet).

eliasnaur:
> One quick observation as I start to review in more detail. I get this validation error from running the `cli` binary:
> 
> ```
> VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated (1345731725)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated ] Object 0: handle = 0xe2dc30000000046, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x50363c8d | VkDescriptorSet 0xe2dc30000000046[] encountered the following validation error at vkCmdDispatch() time: Descriptor in binding #3 index 1 is being used in draw but has never been updated via vkUpdateDescriptorSets() or a similar call.
> ```
> 
> I haven't investigated in more detail. It doesn't seem to affect rendering of the tiger (but I haven't tried anything involving images yet).

I believe that's a variant of https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/1998, a false positive caused by us not initializing all image descriptors in the texture array. The warning disappears if you apply

```
diff --git piet-gpu/src/lib.rs piet-gpu/src/lib.rs
index fd4d0f9..eb41445 100644
--- piet-gpu/src/lib.rs
+++ piet-gpu/src/lib.rs
@@ -302,7 +302,7 @@ impl Renderer {
         // images encoded (I believe there's an cost when allocating descriptor pools). If
         // it can't be satisfied, then for compatibility we'll probably want to fall back
         // to an atlasing approach.
-        let max_textures = 256;
+        let max_textures = 1;
         let k4_pipeline = session
             .pipeline_builder()
             .add_buffers(2)
```

because we only assign one texture below:

```
          let k4_ds = session
              .descriptor_set_builder()
              .add_buffers(&[&memory_buf_dev, &config_buf_dev])
              .add_images(&[&image_dev])
              .add_textures(&[&bg_image])
              .build(&session, &k4_pipeline)?;
  
```

eliasnaur:
I've just added a few more preliminary commits for #70, fixing #66 in the process. I plan to address #70 tomorrow with an SetMode command, replacing the tag flags during encoding for elements.comp.

raphlinus (approved):
This is really impressive work, thanks. I went into the review with all kinds of concern (like - why are the path transforms not affecting binning), but going through the code I had my fears allayed.

Ah ok, I see the reference to the validation errors in the FillImage/sRGB commit. Sorry for not seeing that earlier, and thanks for being on top of it. I still haven't got my mind totally around it, trying to think about the path changes first.

I am totally in agreement with the approach to fix the transform inconsistencies. I also think there's a potential for future speedup by splitting up the work of computing the transform monoid(-ish, thanks floating point!) on a separate compacted stream. But I also think that's a lower priority.

Another general thought - I wonder if the blend stack can be made much smaller, perhaps a single level, and more likely to spill to memory. Part of the reason that comes to mind is that I wonder about the conversions back and forth to sRGB - if it's just one level, it might make sense to store as f32 or possibly f16. But I also consider that a minor detail suitable for further exploration.

I am not 100% sure about doing image/texture sampling as a separate preprocessor step, but do see your reasoning behind it. One thing that immediately comes to mind is that it makes sense to retain the resampled image, for example when scrolling by integer pixels (which I expect to be common). I am a bit surprised that bilinear texture samples are slow, as I imagined they were heavily optimized (and latency hidden), but that said, I haven't done any benchmarks. I feel like this is something we can come back to later if it does make sense to add texture sampling to k4. I was also planning on using texture sampling for gradients, but this is also something to consider.

I did a bit of testing of the cli app and didn't find any problems, and no dramatic performance changes, so am happy with this going into the main branch. Thanks again.

raphlinus:
I spoke a little too soon regarding the Rust version working. The `cli` command does, but the `winit` command fails with this validation error. That's also true when max_textures is set to 1. I 

```
VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated (1345731725)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated ] Object 0: handle = 0xbaa068000000004d, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x50363c8d | VkDescriptorSet 0xbaa068000000004d[] encountered the following validation error at vkCmdDispatch() time: Descriptor in binding #3 index 0 is using imageView VkImageView 0x404f600000000045[] that is invalid or has been destroyed.
```

That gives some clue how to track it down, but pasting it in case you have a clearer idea what might have gone wrong.

ishitatsuyuki:
> I wonder if the blend stack can be made much smaller, perhaps a single level, and more likely to spill to memory.

I suggest that we get rid of the manual allocation logic altogether and simply rely on the shader compiler to do the spills. I do that in my [work branch](https://github.com/ishitatsuyuki/piet-gpu/blob/2e2b938ea6888e9a1d39c0f2d4b022be3167a3a4/piet-gpu/shader/kernel4.comp#L136), although I have the composite layers removed and instead put the stencil buffers there (but the idea is same).

I think we can count on the compiler to spill allocations like this, because registers are not indexable ([reference](https://developer.download.nvidia.com/CUDA/training/register_spilling.pdf), page 5) and compilers cannot put such arrays in registers in the first place. We do need to do a bit of validation though, because there are shader compilers that are poorly implemented.

Putting that concern aside, this approach is much simpler to write, a bit more hardware accelerated, and also likely reduces memory consumption.

eliasnaur:
> I spoke a little too soon regarding the Rust version working. The `cli` command does, but the `winit` command fails with this validation error. That's also true when max_textures is set to 1.
> 
> ```
> VALIDATION [UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated (1345731725)] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-DescriptorSetNotUpdated ] Object 0: handle = 0xbaa068000000004d, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x50363c8d | VkDescriptorSet 0xbaa068000000004d[] encountered the following validation error at vkCmdDispatch() time: Descriptor in binding #3 index 0 is using imageView VkImageView 0x404f600000000045[] that is invalid or has been destroyed.
> ```
> 
> That gives some clue how to track it down, but pasting it in case you have a clearer idea what might have gone wrong.

Yeah, winit didn't work for me, so I forgot about it, sorry. I've added another change to the stack that fixes the winit Vulkan errors on my system.

The validation error you see is caused by my deleting the otherwise unused bg_image field in Renderer; I was tempted by the compiler warning, and didn't realize its purpose: to keep the image alive. I've restored that field along with a comment and a warning-squashing rename in a fixup:

```
--- piet-gpu/src/lib.rs
+++ piet-gpu/src/lib.rs
@@ -191,6 +191,9 @@ pub struct Renderer {
     n_elements: usize,
     n_paths: usize,
     n_pathseg: usize,
+
+    // Keep a reference to the image so that is not destroyed.
+    _bg_image: hub::Image,
 }
 
 impl Renderer {
@@ -342,6 +345,7 @@ impl Renderer {
             n_elements,
             n_paths,
             n_pathseg,
+            _bg_image: bg_image,
         })
     }
 ```

I also changed the number of image descriptors to match the actual number of images we use, to get rid of the other (harmless) warning:

```
--- piet-gpu/src/lib.rs
+++ piet-gpu/src/lib.rs
@@ -305,7 +305,10 @@ impl Renderer {
         // images encoded (I believe there's an cost when allocating descriptor pools). If
         // it can't be satisfied, then for compatibility we'll probably want to fall back
         // to an atlasing approach.
-        let max_textures = 256;
+        //
+        // However, we're adding only one texture for now. Avoid a harmless Vulkan validation
+        // error by using a tight bound.
+        let max_textures = 1;
         let k4_pipeline = session
             .pipeline_builder()
             .add_buffers(2)
```

I'll submit the changes below the FillImage change, and wait a day or so before submitting the rest in case you have objections. I'd also appreciate your approval of the winit Vulkan fixes.

raphlinus:
Thanks, that was the cause, and `winit` now works, with no validation warning. I'm fine with this patch going in as-is, and figuring out the rendering of piet images and gradients on top of it.

eliasnaur:
I've merged everything. Thanks for the reviews.

PR #73, opened 2021-03-16
Always close fill paths, fix #68

### LINKED ISSUE #68: Garbled output with reschart
For some reason, piet-gpu seems to give garbled rendering for the reschart test. The test svg is attached below, and it consists of only typical path and strokes, which piet-gpu should have no problem handling. (It does use a transform to flip the image upside down, but piet-gpu simply ignore so it's unrelated to the issue.)

(File taken from MPVG test suite, renamed to txt for uploading to GitHub)

[reschart.svg.txt](https://github.com/linebender/piet-gpu/files/6056713/reschart.svg.txt)

piet-gpu rendering (`cargo run --release --bin cli -- reschart.svg --scale 2`):

image

Looks like backdrop calculation is all over the place, but I'm not sure what is causing that.

### LINKED ISSUE #74: Minor glitches with reschart
This is a continuation of #68 since the auto-close of path has been implemented.

After the fix, a few glitches can still be seen in the rendering. They are the most visible around the crossing-like structures.

rendering

ishitatsuyuki:
rendering

There are still a few minor glitches visible (even after applying #72), so we should revisit them at a later time. For now, I think this is enough for closing the issue. The glitches are primarily around the edge of the crossing-like structures.

Close #68 
Close #74

### DIFF
--- a/piet-gpu/src/render_ctx.rs
+++ b/piet-gpu/src/render_ctx.rs
@@ -303,6 +303,21 @@ impl PietGpuRenderContext {
     }
 
     fn encode_path(&mut self, path: impl Iterator<Item = PathEl>, is_fill: bool) {
+        if is_fill {
+            self.encode_path_inner(path.flat_map(|el| {
+                match el {
+                    PathEl::MoveTo(..) => {
+                        Some(PathEl::ClosePath)
+                    }
+                    _ => None
+                }.into_iter().chain(Some(el))
+            }).chain(Some(PathEl::ClosePath)), is_fill)
+        } else {
+            self.encode_path_inner(path, is_fill)
+        }
+    }
+
+    fn encode_path_inner(&mut self, path: impl Iterator<Item = PathEl>, is_fill: bool) {
         let flatten = false;
         if flatten {
             let mut start_pt = None;


eliasnaur:
Please do leave #68 open, updating the screenshot, or open a new issue so we don't forget.

ishitatsuyuki:
I opened #74.

raphlinus (commented):
This will certainly fix the svg examples, but a better place to fix it would be in render_ctx in the piet backend, as the implicit closepath is implemented by the other piet backends. I'm ok merging this and having an issue to fix it there, if you'd prefer to get a fix in quickly.

ishitatsuyuki:
You're right, I was concerned as there was no method to mutably close a `Shape`, but I now see that this could be implemented efficiently within `encode_path`. I'll push the alternative tomorrow.

ishitatsuyuki:
I just pushed the alternative.

piet-gpu/src/render_ctx.rs
```diff
@@ -303,6 +303,14 @@ impl PietGpuRenderContext {
     }
 
     fn encode_path(&mut self, path: impl Iterator<Item = PathEl>, is_fill: bool) {
+        if is_fill {
+            self.encode_path_inner(path.chain(std::iter::once(PathEl::ClosePath)), is_fill)
```
raphlinus, line 307: This is also not quite right (and makes me realize a similar issue in the original version) - if there are multiple open subpaths, each one needs an implicit close, not just at the end. Sorry for not communicating this earlier!

ishitatsuyuki: Good catch! Working on this.

ishitatsuyuki:
I went with a somewhat weird iterator trick so that it's allocation-free. Hope it's fine.

Apparently the remaining cases of glitches are also fixed with the updated logic. I'll put close #74 in the PR description.

image

raphlinus (approved):
Thanks, that looks like a proper fix! I've invited to to commit access to the repo, so hopefully you should be able to merge now.

raphlinus:
Sweet, I'm glad my desk-checking caught the additional artifacts without having to investigate. Thanks for iterating til we got it right. I was a bit surprised by the iterator trick, but it seems reasonable to me, especially as it doesn't allocate. (I probably would have done it using lower level logic, the way I would write it in C, but this looks clean and likely just as efficient)

PR #76, opened 2021-03-23
kernel4: Create chunks over the x axis in addition to y axis

ishitatsuyuki:
This allows more coalescing with image loads/stores, since all of our images are stored with a tiled layout.

The tiling layout is vendor dependent so it probably takes tuning to figure out which of 2x4 or 4x2 is faster (I just arbitrarily picked one that is more similar to the one we used to have). BTW My Radeon 5700 XT favors wave64 (local size = 64), so during performance testing I'm using 2x2 chunks. I can change the default to that if you think it's a good idea.

This change was originally made for #64 where the coverage computation had a branch that would be much less divergent with square chunks. It doesn't benefit the current implementation because it branches based on only y values. Though I think the regression on that part would be negligible.


raphlinus (commented):
This seems pretty reasonable to me, though I personally am not prioritizing microoptimizations until I feel like we have the basic architecture for a complete renderer right.

@eliasnaur any opinions, or fears this might conflict with other work? (It's possibly a mechanical merge conflict with the alpha changes in #75)

eliasnaur:
Looks good to me. I'll rebase my work branch after merge.

ishitatsuyuki:
Thanks for the review. I agree that micro optimizations can be looked into later, and hopefully I will focus on finishing #64 from now.

Issue #78, opened 2021-04-01
Android port


raphlinus:
I have a local branch which displays tiger on a Pixel 4, but it needs a fair amount of cleanup before it can be merged. This issue is to track that happening and also document some of the things I ran into.

I was not able to get winit to work on my device - for reasons I still do not understand, it never seems to get a `WindowCreated` event which is properly plumbed through to the user. I used [ndk-glue](https://crates.io/crates/ndk-glue) instead and was able to get it to work. I have some concerns about that code as well, which I should raise as issues on that repo, but none of it is blocking the work at hand, which is to get it running well enough to do performance measurement. One concern is that at least one [callback](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#struct_a_native_activity_callbacks_1ae99c9ef9edf260f2ff423207eb9615f3) is intended to be synchronous, but the ndk-glue code just asynchronously sends an event (over a Unix fd pipe, for a reason I do not understand) to the user thread.

The Pixel 4 does not debug validation layers, and the current code depends on those existing in debug builds. It should sense it at runtime.

The Pixel 4 also does not have descriptor indexing. Again, this should be queried at runtime. This issue is considerably more complex, as it requires multiple compilations of k4, and also CPU-side code to deal with the non descriptor indexing case. The local branch just disables images.

For some reason, I'm getting `ERROR_SURFACE_LOST_KHR` when acquiring the swapchain after two presentations. This is probably something simple, I just haven't figured it out yet.

As part of the cleanup, I want to reduce code duplication between the existing two binaries (cli and winit) which are now three.

If anyone is really eager to try this at home, I can upload my local branch as a draft PR. But I plan on getting the real PR uploaded before long also.

eliasnaur:
> The Pixel 4 also does not have descriptor indexing. Again, this should be queried at runtime. This issue is considerably more complex, as it requires multiple compilations of k4, and also CPU-side code to deal with the non descriptor indexing case. The local branch just disables images.
> 

In case you haven't seen the option, there's a [header define](https://github.com/linebender/piet-gpu/blob/678bfedfca949b54854f47319de6c6075fec3494/piet-gpu/shader/setup.h#L18) to disable use of descriptor indexing. I use that for Gio.

raphlinus:
[...]

PR #79, opened 2021-04-03
Query extensions at runtime

raphlinus:
Don't run extensions unless they're available. This includes querying
for descriptor indexing, and running one of two versions of kernel4
depending on whether it's enabled.

Part of the support needed for #78


raphlinus:
Note: some of the size of this diff is because I ran `cargo fmt`. I don't enforce that in CI in this repo, but we do in most of the other linebender repos - maybe we should set that up.

One change that might be risky is that I changed all the conditional logic in kernel4 to key off `ENABLE_IMAGE_INDICES`. That way, the spv for the non descriptor indexing case doesn't have a `GL_EXT_nonuniform_qualifier` in it, which I think could cause problems for low-spec drivers.

This has been needed for a while, but trying to run the code on my Pixel 4 somewhat forced the issue. In my local branch, I just disabled everything that was causing problems, but that was pretty hacky, so hopefully this is the right solution.

raphlinus:
One potential criticism: I'm not explicitly checking the device extensions against [enumerate_device_extension_properties](https://docs.rs/ash/0.32.1/ash/version/trait.InstanceV1_0.html#method.enumerate_device_extension_properties), but rather gating them on other information. In the case of descriptor indexing, that's a query of more fine grained device descriptor features through `get_physical_device_features2_khr`. It's hard to imagine this failing. In any case, it would be reasonable easy to change, basically just use the `Extensions` type for device extensions as well as instance extensions. I'm interested what Vulkan experts might have to say on this.

Also, I've now rebased by Android branch on top of this and it seems to be working smoothly.

piet-gpu/shader/kernel4.comp
```diff
@@ -128,6 +130,10 @@ void main() {
         if (xy_uint.x < 1024 && xy_uint.y < 1024) {
             rgb[i] = imageLoad(images[gl_WorkGroupID.x / 64], ivec2(xy_uint + chunk_offset(i))/4).rgb;
         }
+#else
+        if (xy_uint.x < 1024 && xy_uint.y < 1024) {
+            rgb[i] = imageLoad(images[0], ivec2(xy_uint + chunk_offset(i))/4).rgb;
+        }
```
eliasnaur, line 136: This unconditionally copies images[0] to the output. This used to not affect me, because I'm not using ENABLE_IMAGE_INDICES. Can we replace this debug output of images with a FillImage command from Rust? Or more expedient, hide the copying behind a DEBUG flag?

Also, this needs rebasing on top of the recently merged RGBA output support.

eliasnaur (commented):
Looks good to me, although I don't know Rust enough to comment on the finer details. There's one issue in kernel4 which will affect me.

> Note: some of the size of this diff is because I ran `cargo fmt`. I don't enforce that in CI in this repo, but we do in most of the other linebender repos - maybe we should set that up.
> 

I'll try to remember `cargo fmt` for my future work.

> One change that might be risky is that I changed all the conditional logic in kernel4 to key off `ENABLE_IMAGE_INDICES`. That way, the spv for the non descriptor indexing case doesn't have a `GL_EXT_nonuniform_qualifier` in it, which I think could cause problems for low-spec drivers.
> 

Is there a specific issue you're thinking of here? When not ENABLE_IMAGE_INDICES, the images array is always constantly indexed with 0.

piet-gpu/shader/kernel4.comp
raphlinus: Moved it behind a debug.

raphlinus:
> Is there a specific issue you're thinking of here? When not ENABLE_IMAGE_INDICES, the images array is always constantly indexed with 0.

If this is working for you now, I wouldn't worry about it. I consider it a cleanup in the generated spir-v that should not affect semantics.

PR #80, opened 2021-04-08
Start DX12 backend

raphlinus:
Very early so far, but uploading the progress I made.

Much of the code is adapted from Brian Merchant's https://github.com/bzm3r/piet-dx12 codebase.


eliasnaur:
Is there a reason you're using D3D12 and not D3D11? Since last we spoke, I figured out why the shaders wouldn't compile for the D3D11 cs_5_0 profile (I had introduced non-uniform barriers) and now they compile fine. Note that I haven't actually run the shaders in D3D11 yet.

raphlinus:
It's a good question, ultimately I want to support both. I go back and forth about whether I want to target "advanced features" including subgroups and descriptor indexing, and those are DX12 only. I think the current plan is to get SM5 working first, then explore those advanced features later (with runtime query).

I had this code lying around for a while, based on Brian's piet-dx12 prototype, so it was handy.

raphlinus:
Merging this now even though it's in rough state, as I'm working concurrently on a number of subproblems (I've had to change the signatures of a few Device methods, mostly because outside Vulkan it's not valid to assume Fence and Semaphore are Copy). I'll continue to refine this as part of the #95 work.

PR #81, opened 2021-04-08
elements.comp: don't modify BeginClip bounding box

eliasnaur:
The BeginClip and EndClip bounding boxes are absolute and must pairwise
match. I mistakenly modified the BeginClip bounding box for stroked
clips.

Signed-off-by: Elias Naur <mail@eliasnaur.com>


raphlinus (approved):
Seems reasonable, though I can't say I've fully audited the logic here.

PR #82, opened 2021-04-10
Add example Android apk

raphlinus:
This is a draft of the current Android support, just to be reproduceable. There will be a proper PR later.


raphlinus:
I uploaded this as a draft PR, but I think it should be possible to merge (I've been keeping it up to date). There are a few things to keep in mind. One is that the default setting of CHUNK is 8, which is one subgroup per workgroup for a subgroup size of 32. On GPUs with a subgroup size of 64 (which includes Pixel 4), CHUNK should be set to 4 at most.

Also, the default size of the render context doesn't match the screen. The right solution of course is to query the window size and set up the renderer to match that, but in the meantime, I've been patching piet-gpu/src/lib.rs with this:

```rust
const WIDTH_IN_TILES: usize = 68;
const HEIGHT_IN_TILES: usize = 143;
```

Clearly, there's a bunch more work to make it performant, but for the most part that's being tracked. There's also lots of potential for cleanup of the duplication between android.rs and winit.rs, among other things. But I think this would be useful to have in tree for experimentation.

eliasnaur:
> I uploaded this as a draft PR, but I think it should be possible to merge (I've been keeping it up to date). There are a few things to keep in mind. One is that the default setting of CHUNK is 8, which is one subgroup per workgroup for a subgroup size of 32. On GPUs with a subgroup size of 64 (which includes Pixel 4), CHUNK should be set to 4 at most.
> 

FWIW, the current CHUNK_X/Y seems like the sweet spot for Pixel 1. X = 2, Y = 2 makes kitchen.apk go from ~15ms to ~22ms in k4, and X = 4, Y = 4 degenerates to ~47ms.

raphlinus:
Interesting. I don't have a good bead on the subgroup size of the Adreno 530. I *thought* it was architecturally pretty similar to the 640 that's in the Pixel 4, but [gpuinfo](https://vulkan.gpuinfo.org/displayreport.php?id=6560#properties_core_11) reports 32. That said, without the subgroup size extension (missing from pixel 1), I've found the reported subgroup not very reliable. It certainly is possible to do other experiments though.

Tuning tile and subgroup sizes is definitely one of the things we should be doing later in the game, but I'm personally not planning on spending too much time on it until the overall architecture looks a bit more solid.

PR #84, opened 2021-04-10
winit: fix n_trans count

eliasnaur:
Missed winit this when doing the deferred transformation fix (only cli worked for me at the time).


raphlinus (approved):
Yes, fixes something that was clearly wrong, thanks.

PR #85, opened 2021-04-11
Encode premultiplied alpha in render_ctx.rs

ishitatsuyuki:
With this change the premultiplied alpha is encoded in the backend, instead of as a part of the API. I expect most of the piet backends to use non-premultiplied alpha so this brings consistency. This also makes alpha in svgs working properly again.

I tested this and it seems to be fine, although I'm not totally confident.


eliasnaur:
LGTM. Thanks for doing this.

PR #86, opened 2021-04-11
kernel4: replace continue in switch to support D3D11 shader model 5.0

eliasnaur:
kernel4: replace continue in switch to support D3D11 shader model 5.0

Without this change, the fxc.exe compiler complains

error X3708: continue cannot be used in a switch


raphlinus (approved):
I think this was needed in the past but no longer, so is a good simplification.

PR #89, opened 2021-05-05
Start text rendering

raphlinus:
This commit puts in basic integration with ttf-parser and starts
populating the various piet text objects. The font is currently
hard-coded.


raphlinus:
I'm inclined to merge this. It's very rough, and in many ways doesn't represent the long term direction of where I want this to go, but it gets some text rendering in place, and most importantly for my present goals, allows us to start running performance evaluation for font rendering workloads.

ishitatsuyuki:
Sounds good, the render_ctx refactoring also look nice.

Issue #88, opened 2021-05-01
A bit of clip optimization


raphlinus:
I was curious about whether the clip stack really needs to store both rgba and another alpha channel, so had another look at that logic. I think it is possible to get rid of the extra channel, with just a bit of cleverness.

The tricky bit is that the coarse rasterizer needs the path at BeginClip, so it can do the optimizations, but the fine rasterizer only needs it at EndClip. The basic idea is to make it available to both.

Encoding: the EndClip element is annotated with the number of paths inside the clip, ie the path count at BeginClip plus this quantity equals the path count at EndClip. This will require just a bit of accounting, but hopefully not too bad. Also note that by storing the difference (as opposed to an absolute path count) the encoding is no less "relocatable" than before.

Coarse rasterization, input stage: for EndClip, the `path_ix` is not just the `element_ix` (coarse.comp line 236), but the delta encoded above is subtracted. Thus, we actually read the BeginClip path, and the path associated with EndClip is ignored.

Coarse rasterization, output stage: BeginClip has the same optimization logic, but does not write the fill. EndClip writes the fill right before writing the EndClip command.

Fine rasterization: BeginClip does not push an alpha value. EndClip does not pop the alpha value, but is otherwise basically unchanged; it already composites using area[k], which at present is 1.0 because coarse rasterization always sends a Solid command before the EndClip.

A couple other notes. The do-while in kernel4 Cmd_Fill should be a while loop, as there are cases when coarse rasterization can send a tile with no path segments (when the nesting depth exceeds 32; this is unusual but possible).

Lastly, I think using the same path (and thus the same bbox) for begin and end of clips might address the problem I was talking about in [this comment](https://github.com/linebender/piet-gpu/issues/36#issuecomment-731288938), and allow the use of relative bounding boxes again, which would increase the compositionality of encoded scene subtrees.

I'm not 100% sure this will work, but I think so. I also reviewed the current separation between path alpha and color source, and that's looking good for gradients; thanks Elias!

PR #90, opened 2021-05-08
Add stack monoid implementation

raphlinus:
This should be a complete implementation of the core logic of a fast
implementation of the stack monoid. It has been tested somewhat but
needs more work to make performance measurement meaningful.


piet-gpu-hal/examples/shader/stack.comp
```diff
@@ -0,0 +1,217 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
eliasnaur, line 3: The Apache license is incompatible with the piet-gpu shaders available under the MIT or UNLICENSE.  Consider making the licensing clear by replacing this section with

// SPDX-License-Identifier: Apache-2.0 OR MIT OR Unlicense

piet-gpu-hal/examples/shader/stack.comp
```diff
@@ -0,0 +1,217 @@
...
+        uint pop_last = sh_npop[last];
+        uint m = max(push_mid, pop_last);
+        uint push_last = sh_npush[last];
+        if (last - th_ix >= push_last) {
+            uint iprev = last - th_ix - push_last;
+            if (iprev < m - pop_last) {
+                stack = sh_stack[mid - pop_last - iprev];
+            } else {
+                // TODO: not needed
+                stack = 99;
```
eliasnaur, line 104: If you plan to leave this, a comment saying how it can be removed would be nice.

piet-gpu-hal/examples/shader/stack.comp
```diff
@@ -0,0 +1,217 @@
...
+
+    // Do lookback
+    // Note: unlike prefix sum example, we need all threads active to do stack reduction.
+    if (part_ix != 0) {
+        bool backlink = false;
+        uint look_back_ix = part_ix - 1;
+        while (true) {
+            if (th_ix == PART_SIZE - 1) {
+                // This is for evaluation and will be removed or conditional
+                //atomicAdd(state[part_ix * STATE_SIZE + 5], 1);
```
eliasnaur, line 145: Remove or expand comment. STATE_SIZE depends on this as well, right? If so, it too should be conditional.

piet-gpu-hal/examples/shader/stack.comp
```diff
@@ -0,0 +1,217 @@
...
+
+layout(set = 0, binding = 0) buffer DataBuf {
+    uint[] data;
+};
+
+layout(set = 0, binding = 1) volatile buffer StateBuf {
+    uint[] state;
+};
+
+#define STATE_COUNTER 0
```
eliasnaur, line 32: Is STATE_COUNTER global? If so, I suggest replacing it with a `state_counter` field in StateBuf.

piet-gpu-hal/examples/stack.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
eliasnaur, line 3: Replace with

// SPDX-License-Identifier: Apache-2.0 OR MIT OR Unlicense

?

eliasnaur (commented):
Exciting!

My comment are janitorial in nature; I haven't studied the algorithm enough to understand it.

A few questions to please the audience:

- What is the plan to get this work integrated into piet-gpu?
- Can this implementation tolerate GPUs with primitive scheduling? Or will a change similar to https://github.com/linebender/piet-gpu/blob/f6c2558743937f1e5d0eeb9ae1998a3746133349/piet-gpu/shader/elements.comp#L251 be needed?
- How much does this work overlap with https://github.com/linebender/piet-gpu/issues/88? I assume that annotating EndClip with the path count will not be necessary?

raphlinus:
Thanks for the comments. This is still work in progress, and part of that is that some debug scaffolding is still in place. I'll address the specific issues in the next commit. Right now though, I'm tracking down what appears to be a pretty grungy correctness issue, which seems to involve memory coherence.

It does need the scheduling fix as you describe, also one reason this is work in progress.

Regarding integration into piet-gpu for the clip stack, yes, I think so. I'm still working out the details, but so far it seems encouraging. The direction I'm leaning now is to treat element processing less as an "ubershader", and do more segregation of the streams for transforms, clips, and drawing ops. That already seems to be a trend with your transform consistency fixes. I also think it's not a special burden on CPU-side encoding; the idea that there's a single flatbuffer-style flattened representation seems worth letting go. This is a more major change, but yes, it would have the effect of not needing to add the path count to EndClip. I think doing this right would enable #88 and further optimizations (including accounting for blends, which is very similar to clips but subtly different).

I also spaced out on the licensing; I'm trying to follow the letter of Google's [releasing guidelines](https://opensource.google/docs/releasing/preparing/) but those are mostly set up for Apache-2 only projects, and this is multiply licensed. That's navigable but will require a bit of care on my part. Thanks for reminding me.

raphlinus:
Closing this in favor of a new approach which is not based on decoupled lookback, and which is potentially more optimized.

PR #92, opened 2021-05-08
Initial commit of gpu test utility

raphlinus:
Basic test of memory coherence. We run a simple test of acquire / release semantics under four conditions:

    * Pure array access, no expectation of any memory semantics
    * Same but with buffer annotated as "volatile"
    * All access to memory is through atomics
    * Use Vulkan memory model to specify acquire / release semantics


raphlinus:
This is now subsumed by the work in the tests/ subdirectory, particularly #127.

PR #99, opened 2021-05-27
Boost backdrop parallelism for the prefix sums

ishitatsuyuki:
The backdrop stage can become quite a lot slower when there are objects covering a large area; this PR alleviates the issue by giving the prefix sum a little more parallelism when there are a lot of rows to process. For paris-70k scaled at 4x, this cuts down the backdrop stage from 340us to 292us (a very slight boost). This doesn't boost the overall performance much since the bottleneck is coarse and fine stages anyway, but I think backdrop's scalability issues might arise at some point and it's better to keep an eye on it.

One potential compatibility issue is that the workgroup size have been increased to 1024. I believe this is supported on most GPUs, but it needs further testing.

Profiling also shows further work we can do to improve the scalability. Right now, the backdrop stage is bounded by the maximum wavefront duration; in other words, the longest running thread dominates. (That's probably why there isn't significant perf boost with this change.) We can further parallelize on the column level through [Segmented Reduction](https://moderngpu.github.io/segreduce.html) to alleviate this.

We might also try to sparsify the backdrop (this is done by Li et al), i.e. merging tiles that are horizontally adjacent and does not intersect with the outline into a single rectangular tile. I'm still not sure how to do it in piet-gpu's architecture though.


eliasnaur:
A workgroup size of 1024 will break some mobile GPUs such as the one in my Pixel 1. Its maximum workgroup size is the Vulkan minimum of 128.

raphlinus:
We should probably beef up runtime querying of [this limit](http://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxComputeWorkGroupInvocations&platform=android) and allow for permutations of the shader.

ishitatsuyuki:
> A workgroup size of 1024 will break some mobile GPUs such as the one in my Pixel 1. Its maximum workgroup size is the Vulkan minimum of 128.

@eliasnaur In that case, do you modify the shaders to run on the Pixel? The current workgroup size is 256, which is above that. Maybe the driver can transform the shader code to make it compatible though.

eliasnaur:
> > A workgroup size of 1024 will break some mobile GPUs such as the one in my Pixel 1. Its maximum workgroup size is the Vulkan minimum of 128.
> 
> @eliasnaur In that case, do you modify the shaders to run on the Pixel? The current workgroup size is 256, which is above that. Maybe the driver can transform the shader code to make it compatible though.

Good question. I added the [LG_WG_FACTOR](https://github.com/linebender/piet-gpu/blob/acb4e5f8affae5cfaaab880d963db86264f00e9f/piet-gpu/shader/setup.h#L8) setting to work around this problem.

ishitatsuyuki:
I see. I will look into the runtime permutations once I have the exam sorted.

ishitatsuyuki:
I have implemented the runtime querying of threadgroup size. Please take a look.

ishitatsuyuki:
Friendly ping!

piet-gpu-hal/src/metal.rs
```diff
@@ -164,6 +164,10 @@ impl MtlInstance {
                 has_descriptor_indexing: false,
                 has_subgroups: false,
                 subgroup_size: None,
+                workgroup_limits: WorkgroupLimits {
```
raphlinus, line 167: This is conservative, as it seems to be 1024 on all desktops (we're not officially supporting iOS yet, although it would be nice to do so in the future). The official documentation is https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf , and there a [Stack Overflow thread](https://stackoverflow.com/questions/56677031/availability-of-older-metal-feature-set-table-documents) with links to older documents.

Conservative is ok given the TODO above, but I would prefer 1024 here and a note saying that it should be feature-detected when iOS is supported.

raphlinus (approved):
Looks good, thanks! Having the workgroup limits runtime detected is definitely an important addition to the hal layer.

PR #104, opened 2021-06-26
Add command line parameters to winit

ishitatsuyuki:
So that I don't need to modify lib.rs every time I want to benchmark...

Code copied over from the cli example.


raphlinus (approved):
Yes, this is definitely useful, thanks.

At some point I'll want to clean up the cut'n'pasting, but it's fine for now.

PR #105, opened 2021-07-11
Linguist highlighting

DJMcNab:
I was getting annoyed by following links to this code in zulip and it not being highlighted. See [an arbitrary file in this branch](https://github.com/DJMcNab/piet-gpu/blob/98862ab1c7c4b02c3c3d01e2384e299e890ee50b/piet-gpu/shader/annotated.h) for what the highlighting looks like.


ishitatsuyuki (approved):
Looks good to me.

PR #106, opened 2021-07-12
Fix shader hang in coarse path

raphlinus:
I'm observing a hang in the coarse path rasterizer on Nvidia 1060 (not
on other GPUs, and I think only after a more recent driver update).
Investigation shows that the hang happens when the loop to output line
segments from flattening does not terminate. I believe this is a
floating point inconsistency of some kind - the "target" should always
be at least the sum of the "val" for all quadratics, but if it
undershoots the loop may continue when it should not.

It's *possible* this patch is masking a deeper logic error. Ideally, I'd
instrument the shader and capture a trace of when exactly this goes
wrong. But in any case this patch shouldn't make things any worse.


eliasnaur:
I have reports of GPU hangs ([1](https://todo.sr.ht/~eliasnaur/gio/214#event-80820), [2](https://todo.sr.ht/~eliasnaur/gio/219#event-80819)), and had hoped this change might fix it. However, I'm seeing corruption that looks similar to when I debugged paths that weren't airtightly closed.

2021-07-13-095218_grim
2021-07-13-095216_grim

raphlinus:
Thanks Elias, that's good feedback. It sounds like there is indeed a logic bug, and my proposed fix looks like it's truncating paths so the coverage isn't all the way from one end to the other. Fortunately, I now have a reproducible case (at least on 1060 hardware, if not my other GPUs), so I should be able to add instrumentation to track down in more detail what's really happening. I'll not merge this PR in its current state.

ishitatsuyuki:
This is probably superseded by #108. Would you want to close this?

raphlinus:
Yes, thanks for the reminder.

PR #108, opened 2021-07-15
Retain subdivision results

### LINKED ISSUE #107: Hangs on coarse path on Nvidia 1060
This is branched from #106 , which observed the issue and attempted to fix it, but the patch didn't contain a full analysis.

The problem is that on some hardware (including Nvidia 1060, at least with recent driver), the `n_out` loop continues. I've got instrumentation on the shader now, and one of the things that's most obviously wrong is a NaN val coming out of `estimate_subdiv`. The idea that can happen when `cross` is zero (or nearly so) is not surprising, looking at the code.

I've tried a simple patch to that function that early-outs when cross is small, but that's not fixing all the looping. I'll do a deeper analysis and then propose a patch, but I did want to update based on my results so far.

raphlinus:
Don't recompute the parameters from quadratic subdivision, but rather
retain them across the two phases (summing the subdivision estimate, and
generating the subdivisions). The motivation for this is that the values
were subtly different (differing by 1 or 2 least signficant bits) across
the two phases. It *might* also be faster depending on ALU/memory
relative performance.

Fixes #107


piet-gpu/shader/path_coarse.comp
```diff
@@ -69,6 +70,9 @@ SubdivResult estimate_subdiv(vec2 p0, vec2 p1, vec2 p2, float sqrt_tol) {
     float x0 = (d01.x * dd.x + d01.y * dd.y) / cross;
     float x2 = (d12.x * dd.x + d12.y * dd.y) / cross;
     float scale = abs(cross / (length(dd) * (x2 - x0)));
+    if (abs(cross) < 1e-9) {
+        //return SubdivResult(0.0, 0.0, 0.0);
+    }
```
eliasnaur, line 75: Delete?

piet-gpu/shader/path_coarse.comp
```diff
@@ -144,12 +151,13 @@ void main() {
...
             vec2 qp1 = eval_cubic(cubic.p0, cubic.p1, cubic.p2, cubic.p3, t - 0.5 * step);
             qp1 = 2.0 * qp1 - 0.5 * (qp0 + qp2);
-            SubdivResult params = estimate_subdiv(qp0, qp1, qp2, sqrt(REM_ACCURACY));
+            SubdivResult params = keep_params[i];
             float u0 = approx_parabola_inv_integral(params.a0);
             float u2 = approx_parabola_inv_integral(params.a2);
             float uscale = 1.0 / (u2 - u0);
             float target = float(n_out) * v_step;
             while (n_out == n || target < val_sum + params.val) {
+                if (n_out == 1000) break;
```
eliasnaur, line 160: Delete?

piet-gpu/shader/path_coarse.comp
```diff
@@ -30,6 +30,7 @@ layout(set = 0, binding = 1) readonly buffer ConfigBuf {
 #define Q_ACCURACY (ACCURACY * 0.1)
 #define REM_ACCURACY (ACCURACY - Q_ACCURACY)
 #define MAX_HYPOT2 (432.0 * Q_ACCURACY * Q_ACCURACY)
+#define MAX_QUADS 16
```
eliasnaur, line 33: Seems a bit low?

raphlinus: I think it's adequate, but it probably deserves deeper analysis. In the worst case, the error scales as n^3, so this means the error is 1/4096 what it would be for a single quadratic bezier. The worst case is when the third derivative is large (which happens for S shapes like y=x^3, and probably also cusps). I think it will be fine for basically non-adversarial cases, but it might be worth taking a closer look at adversarial ones.

eliasnaur (approved):
Thanks!

PR #111, opened 2021-08-20
piet-gpu/shader: treat memoryBarrierBuffer as a control barrier

eliasnaur:
With this change, piet-gpu successfully runs Gio on Metal.


eliasnaur:
With this change, piet-gpu successfully runs Gio on Metal.

raphlinus (approved):
Thanks! I'm sure this would have taken a long time for me to track down. I'm curious: is there any place where this is written down? It's one of the reasons I'm looking forward to WebGPU - it will force GPU drivers to conform to an actual spec.

eliasnaur:
I don't know if this is written down anywhere. I had to bisect a GPU hang, seeing memoryBarrierBuffer mapped to threadgroup_barrier by spirv-cross and finally skim the Metal language spec to determine that spirv-cross couldn't have done any better.

It didn't help that my mid-2014 macbook pro GPU hangs running piet-gpu no matter what I do; I had to find a newer macbook pro to make it run :)

eliasnaur:
FWIW, I have not been able to successfully run piet-gpu on any of my iPhone devices. I've tried an iPhone SE 2, iPhone SE, iPhone 6. Only the SE 2 can run for a bit with garbled output before crashing, the other two crash during shader compilation.

PR #112, opened 2021-09-06
Render COLRv0 emoji; animate

raphlinus:
This PR is a bit of a grab bag of various work I've been doing. The biggest change is use of swash for conversion of font outlines (replacing ttf-parser), including the ability to render COLRv0 layers.

There is a bit of refactoring to support animation (the scene can be uploaded to the renderer). The scene buffer is also padded with nop elements, as the ability to read corrupt elements after the actual end of the scene was a bug, just much more likely to get triggered when animating.

This PR also reinstates the reuse of command buffers, as the current state is to leak them, again a significant problem when animating. That works on Vulkan but temporarily breaks DX12 and Metal. The most straightforward and portable approach to that would be to destroy command buffers; I'm currently working on that and it will be a followup. If there's a feeling against committing a PR that regresses on that front, I can remove those changes and roll them up into the followup.


jrmuizel:
Out of curiosity, why did you switch to swash from ttf-parser?

PR #114, opened 2021-10-21
Coarser grain ScaleContext

raphlinus:
This patch sets up a ScaleContext for the duration of a draw_text call,
where it was previously per-glyph.

### DIFF
--- a/piet-gpu/src/text.rs
+++ b/piet-gpu/src/text.rs
@@ -1,6 +1,6 @@
 use std::ops::RangeBounds;
 
-use swash::scale::ScaleContext;
+use swash::scale::{ScaleContext, Scaler};
 use swash::zeno::{Vector, Verb};
 use swash::{FontRef, GlyphId};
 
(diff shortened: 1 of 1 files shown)


raphlinus:
@dfrg You might want to take a look at this.

I experimented with retaining a singleton ScaleContext (held behind a mutex) across multiple draw calls, but it was a bit slower than this. I started to do a little profiling but didn't get very far. Maybe you have a bit of insight why that might be? In any case, I'm fairly satisfied with the performance now, including on Android.

dfrg:
Ah, I see the split the struct to satisfy the borrow checker trick. If the ultimate goal is to move to multiple threads then this is probably unnecessary as we’ll want a different approach anyway. 

Honestly, aside from my allergy to malloc overuse, the biggest performance boost from retaining a ScaleContext is going to come from the hinter cache. We definitely don’t want to be executing fpgm/prep per glyph but doing it per thread/pass/font may be negligible.

I’m going to fork this and hook up the layout code later this evening so that’ll give me a chance to get better acquainted with the code base and hopefully give me some ideas about how this will all fit in a multithreaded model.

dfrg:
That sounds good. Layout will add a lot of irrelevant noise so I suspected that you’d prefer to keep it separate while you’re focused on raw performance. Basic integration should be trivial so I’m not concerned about divergence.

PR #120, opened 2021-11-02
Add memory barrier to elements shader

raphlinus:
The flag read needs acquire semantics. There are a number of ways that
could be expressed, but a generally portable way is to have a barrier
after. However, in the translation to Metal, that barrier needs to be in
uniform control flow. This patch does some workarounds to ensure that.


raphlinus:
The [element_barrier_ifdef](https://github.com/linebender/piet-gpu/tree/element_barrier_ifdef) branch has an additional commit that generates the Metal workaround version only when the #define is set. The non-Metal version compiles successfully with FXC using cs_5_0.

eliasnaur:
FWIW, tried patch on iPhone 5s, but it didn't make any noticeable difference (corrupted output, crash after a few frames).

PR #121, opened 2021-11-05
Alternative strategies for elements barrier

raphlinus:
This patch provides an #ifdef METAL for switching between two different
strategies for the barriers in the elements shader. With it enabled, the
barriers are in workgroup uniform control flow, which is compatible with
translation to Metal, but fails to compile in FXC, as FXC's uniformity
analysis fails. With it disabled, the lookback logic (including
barriers) runs only in a single thread, which compiles in FXC (though
not tested), but creates problems in Metal.

In testing on Android, the METAL version seems slightly faster. On AMD
5700 XT, there is no measurable difference.

I'm inclined not to commit this, but it's potentially useful if we want
to explore cs_5_0 compatibility.


raphlinus:
As discussed, this doesn't feel like it's worth the complexity because we're not targeting cs_5_0. For DX12, we'll probably pursue DXC instead of FXC. Also, we may end up with tree reduction instead of decoupled look-back for more robust compatibility.

But if we ever *do* want to try to port decoupled look-back to FXC, this is probably the way to do it.

eliasnaur:
For completeness, I'll note that you can run D3D11 on top of D3D12 through official Microsoft API:

https://github.com/microsoft/D3D11On12

For Gio, I'll probably end up expanding the D3D11 backend to attempt a D3D11On12 context and only use piet sm 6 shaders if that succeeds. If that works, we won't have to maintain two D3D renderers.

PR #123, opened 2021-11-06
 Start testing framework, with prefix sum

raphlinus:
This adds a new testing directory, and starts populating it with prefix sum implementations, both decoupled look-back and tree reduction. The intent is to build out more tests, and more options for configuring and reporting, but for now it's pretty bare-bones.

This will subsume #92; I also plan to get updated versions of the atomic litmus tests in there and have a bunch locally that I'm experimenting with.

Shaders are built for 3 languages, but the code only runs on Vulkan for now. I have gotten the tree reduction working on both Metal and DX12 with some tweaks, though timer queries are not supported on either yet. I'll do some portability improvements as a follow-on (and will use these tests to push that work forward).


raphlinus:
Going to merge this now even though it feels a bit incomplete, as I'm doing some of the DX12 portability (#125) work on top of it.

Issue #119, opened 2021-11-01
New element processing pipeline


raphlinus:
It's time to redo the element processing pipeline. The proposal in this issue fixes the following problems (of which some but not all clip related issues are addressed in #88):

* Accounting of clip bbox needs to be done by CPU, and is in global coordinates. This seriously impacts composability of encoded scene subgraphs.
* Clip bbox is accounted as the *union* of enclosed drawing commands, rather than intersection of clips in the stack.
* Clip mask is computed at BeginClip but is not needed for alpha composition until EndClip, requiring more memory bandwidth in fine rasterization.
* Bounding boxes are not tight in the presence of non axis aligned transforms.
* Element processing is slower than it might be.
* Adding blends makes many of the above issues worse.

The main theme of this proposal is to divide the scene buffer into *streams,* rather than having a single flattened encoding of the scene tree. Later streams can reference earlier streams, and use a prefix sum of the count to track the reference. This idea is a continuation of the transform work done by Elias (#62, #71), but takes it farther.

In this proposal, there are 3 streams encoded by the CPU: transform, path, and draw object. From the draw object stream is derived a clip stream, so there is a pipeline of 4 stream processing shaders, each implemented as a single compute dispatch implementing a monoid scan. It should be emphasized, the size of all these streams is easily determined during encoding and is used to allocate buffer space and determine the number of workgroups in each such dispatch. We're already doing a lot of that.

Thus, an encoded scene (or subgraph of a scene) is a tuple of the 3 streams, as well as some counts of number of transform and clip elements. The idea is that it's still straightforward and efficient to stitch together encoded subgraphs, which is done by concatenating each of the streams and adding the counts.

## Transform stream

The transform stream is very simple, each element is just an affine transform, and the monoid is affine matrix multiplication.

As a potential later refinement, we might encode push and pop elements and use a stack monoid, rather than the current approach of representing a pop by pushing the inverse transform. That would allow degenerate transforms and reduce roundoff error (because floating point, we might not end up with the identical transform), but it's not clear it's worth the effort at this point.

## Path stream

The path stream has the following elements:

* Transform()
* SetLineWidth(line width)
* LineSegment(p0, p1)
* QuadSegment(p0, p1, p2)
* CubicSegment(p0, p1, p2, p3)
* EndPath()

(Alternatively, we might SetLineWidth as is done now. I don't have a strong feeling about this, but encoding this in the End element seems like it might be slightly more efficient)

[...]

PR #128, opened 2021-11-14
Fix write-after-read in prefix test

raphlinus:
Thanks to Jeff Bolz for spotting the write-after-read hazard on the
sh_flag accesses. This fixes observed failures on Nvidia Turing and
Ampere on DX12.


raphlinus:
I should note, this bug also exists in elements.comp, but is not associated with any failures I'm aware of. I don't think it's worth the effort to try to fix that, especially as it's untestable; #119 is the way forward, which involves getting reliable, tested versions of the primitives first, then stitching them together into the new element processing pipeline.

raphlinus:
This also fixes the observed failures on Intel HD 630, both Vulkan and DX12.

PR #130, opened 2021-11-18
Atomic load coherence test

raphlinus:
This is a rough draft of a test for coherence of relaxed atomic loads. At this point, I'm not 100% convinced it's testing something real, though it does seem promising - it correlates pretty well with failures of the prefix sum example, even when the message passing litmus test passes.

My goal right now is to foster discussion. Does this test catch anything new, ie that would not be caught by existing litmus tests? Is it suitable to add to a test suite? It's very different in flavor from most atomic consistency tests, as it explicitly invokes time.


DJMcNab:
Closing, as it mainly uses piet-gpu-hal, and it seems unlikely to foster more discussion

PR #131, opened 2021-11-20
Prefix

raphlinus:
This is the version of the prefix sum algorithm as described in my [Prefix sum on Vulkan](https://raphlinus.github.io/gpu/2020/04/30/prefix-sum.html). That sadly referred to the tree (a bad habit of mine), so to avoid link rot it might make sense to keep it. Probably a better solution is to change the link to refer to the PR.


DJMcNab:
Closing; we now use multi-dispatch parallel scan extensively in Vello. Linking to the PR doesn't care if its closed.

PR #136, opened 2021-12-01
Path stream processing

raphlinus:
This patch contains the core of the path stream processing, though some
integration bits are missing. The core logic is tested, though
combinations of path types, transforms, and line widths are not (yet).

Progress towards #119


raphlinus:
Now passes on a bunch of test GPUs, including mac M1.

piet-gpu/shader/bbox_clear.comp
```diff
@@ -0,0 +1,29 @@
...
+#version 450
+#extension GL_GOOGLE_include_directive : enable
+
+#include "mem.h"
+#include "setup.h"
+
+#define LG_WG_SIZE 9
+#define WG_SIZE (1 << LG_WG_SIZE)
+
+layout(local_size_x = WG_SIZE, local_size_y = 1) in;
```
eliasnaur, line 14: Some older Android phones limit workgroup sizes to 128, and the old pipeline has a setting to support that. Do you plan to have a similar facility in the new pipeline?

raphlinus: Thanks for the reminder. Yes, I'll definitely make sure to make this compatible. It's mostly just an issue of more repetition in the ninja build file and the various code to instantiate the stages.

PR #138, opened 2021-12-03
Beginnings of new element pipeline

raphlinus:
This successfully renders the tiger; fills and strokes are supported.
Other parts of the imaging model, not yet.

Progress toward #119


raphlinus:
An update on the current state. As of the most recent commit, it should basically be at feature parity with the previous version. I think it would be reasonable to merge, but want to think through a few loose ends first.

One is the tuning of workgroup size. In the current state, some workgroups are 256 and some are 512. That's going to hurt compatibility with low-spec devices. @eliasnaur mentioned this above. The easiest fix is to change all to 256 max and have them all respect `LG_WG_FACTOR` in setup.h so they will be 128 when this is 0. We don't currently compile or run that case from Rust, but it seems reasonable.

There are two reasons the workgroups are large. One is performance tuning (done on 5700 XT only so far). The other is that the monoid scan stages are currently limited to a height of 2, meaning that the total problem size is limited to workgroup^2. With the current large workgroups, that's not a serious limitation. With smaller ones, it may be. The code for that is not conceptually difficult but is tedious. The [prefix_tree](https://github.com/linebender/piet-gpu/blob/master/tests/src/prefix_tree.rs) test is a model of scaling to arbitrary height. A limit of 3 might be reasonable.

This brings us to another related question. Currently I have N_SEQ at 8 for many of the monoid stages. This means that each thread processes 8 items sequentially. That's great for large problems, but, combined with the large number of dispatches (because we aren't doing a single-pass approach) means that you get very low utilization when there are small problems. It's not clear to me whether this is even worth optimizing - it's arguably not a problem in real usage, but might make benchmarks look bad. One approach is to dynamically tune, but that makes the number of shader permutations even more out of control. I might look into whether I can do specialization constants portably to try to mitigate that.

There are still a lot of clip improvements to be done, but I'd like to land this and schedule that later.

Lastly, this work is triggering some validation errors on DX12, but those were already present. Those should also be captured in a separate issue.

PR #140, opened 2021-12-17
New stack monoid implementation

raphlinus:
This is another major attempt at fast implementation of parentheses matching. It's also intended to be more portable, not depending on decoupled look-back.

As of this snapshot, there is a working implementation, but only partly optimized. The first dispatch has a sequential iteration at the leaves, which makes it go faster, but the second dispatch processes only one element per thread so far. Also, only the SPV shader artifacts are currently built, though no portability concerns are anticipated.

The ideas in this code are fairly deep and my current intent is to write a paper to explain them. Even so, I'm hoping that the code is more readable than the last time around (#90, also the [stack monoid revisited](https://raphlinus.github.io/gpu/2021/05/13/stack-monoid-revisited.html) blog post).


raphlinus:
Closing this as it's not suitable for merging to main. The linked blog post is *still* in the pipeline, but that can reference this closed PR. It's also the case that the stack monoid *is* in main as part of the clip implementation, but that doesn't quite match the paper. In particular, intersection is implemented there, but not union, which might be useful in the future for improving the blend implementation; right now it depends on the application to provide a bounding rectangle for each blend.

PR #144, opened 2022-01-19
First cut at glyph rendering API

raphlinus:
This exposes interfaces to render glyphs into a texture atlas. The main changes are:

* Methods to plumb raw Metal GPU resources (device, texture, etc) into piet-gpu-hal objects.
    
* A new glyph_render API specialized to rendering glyphs. This is basically the same as just painting to a canvas, but will allow better caching (and has more direct access to fonts, bypassing the Piet font type which is underdeveloped).
    
* Ability to render to A8 target in addition to RGBA.
    
WIP, there are some rough edges, not least of which is that the image format changes are only on mac and cause compile errors elsewhere.


piet-gpu-hal/src/hub.rs
```diff
@@ -312,8 +312,13 @@ impl Session {
     ///
     /// Currently this creates only a 2D image in RGBA8 format, with usage
```
ratmice, line 313: Comment could use an update.

raphlinus:
mtl-text should be updated to point to the SHA when this is merged. @dfrg any concerns with this PR or do you think it's good to go?

dfrg:
This looks good to me. I'll send a new PR for the proposed changes to add_glyph/make_glyph.

PR #145, opened 2022-01-20
piet-gpu: Add phony targets for spv, dxil and msl

ishitatsuyuki:
Makes building without dxc or spirv-cross easier.

As a random note, I use `cargo watch -w piet-gpu/shader -- ninja -C piet-gpu/shader spv` (from https://github.com/watchexec/cargo-watch) during the development process. Might want to document this somewhere.


raphlinus (approved):
I see the value of this, having been in a similar position (and it's going to get worse when we generate AIR). But it does add to the boilerplate, I'm thinking it's going to be necessary to autogen the ninja file before too long.

Thanks!

PR #146, opened 2022-01-21
piet-gpu, hal: Add Vulkan debug markers

ishitatsuyuki:
Other backends are left as stub for now.

Makes it easier to identify stages within Radeon GPU Profiler and others.

The markers are only enabled for `debug-assertions` builds for now, we probably should revisit this in the future.


raphlinus (approved):
Looks great, thanks!

PR #147, opened 2022-01-30
shaders: Add .clang-format and reformat

ishitatsuyuki:
Helps keeping the code tidy.

Style is chosen to minimize diff, but contains a slight bit of personal taste.


raphlinus (approved):
Thanks, looks good. This also reminds me, I want to set up CI with cargo check, rustfmt, etc.

PR #148, opened 2022-02-09
Hierarchical fine raster

ishitatsuyuki:
This commit introduces a hierarchical approach to exploit sparsity in fine rasterization.

A cheap and non-antialiased is performed first for each pixel, then a sparse AA pass is performed for only the tiles that are touched by the line segment.

The performance is slightly worse (~10% slower k4) than the previous approach, unfortunately. While this approach doesn't give much benefit as-is, it can be effective in bringing down the cost of a bitmask-based supersampling approach (#64), which notably solves conflation artifacts.

There are detailed code comments on various numerical correctness topics as well as some non-obvious optimizations.


eliasnaur:
> The performance is slightly worse (~10% slower k4) than the previous approach, unfortunately.

Sorry for being the worrywart, but is 10% representative? Also on weak(er) mobile GPUs?

ishitatsuyuki:
> Sorry for being the worrywart, but is 10% representative? Also on weak(er) mobile GPUs?

It's not representative, it's an optimistic number, measured after spending my best effort to optimize for 5700 XT. There are 3 factors to consider:
- Subgroup size. The current approach favors architectures with a subgroup size of 32 or below. Old AMD GPUs and qcom GPUs have 64+ subgroup size which results in redundant cycles.
- Shared memory atomics. Using shared memory atomics can result in fallback code generation using cmpxchg if proper atomics is unsupported, and on Mali GPUs shared memory use is lowered to global memory access. How much exactly does this impact performance is unknown.
- Compiler quality. I developed on RADV ACO, which is a pretty reliable and good quality compiler. Some mobile compilers are infamous for not being able to optimize properly. This might cause issue with them, but in that case we can just hand unroll the optimizations.

So there are reasons that mobile might be slower. But I don't have an idea how much until we get a benchmark.

raphlinus:
I just want to say, I want to review this soon, but am currently slightly stuck on getting the new clip implementation done. Thanks for doing it, it looks very interesting.

AlexVestin:
There seems to be some issues at with the rendering at the left border (top is hierarchical2, bot is master)
hier

the svg file rendered:
https://user-images.githubusercontent.com/31164322/154847582-8154d127-f663-4da6-bd79-9b63d228d353.svg

raphlinus:
Closed as this is not mergeable, and with thanks for the exploration.

While my current focus is on conflation artifacts and I am satisfied with the performance of the current fine stage for the analytical antialiasing case (and find its simplicity and code size appealing), I've also been thinking about an approach similar to that sketched in #270, where individual pixels sparsely calculate the delta of area (a[x + 1] - a[x], so the existing area calculation could be done twice, though I think it's possible to simplify), then a prefix sum stage. So I'm looking forward to the exploration continuing.

PR #150, opened 2022-02-18
New clip implementation

### LINKED ISSUE #88: A bit of clip optimization

raphlinus:
This PR reworks the clip implementation. The highlight is that clip bounding box accounting is now done on GPU rather than CPU. The clip mask is also rasterized on EndClip rather than BeginClip, which decreases memory traffic needed for the clip stack.

This is a pretty good working state, but not all cleanup has been applied. An important next step is to remove the CPU clip accounting (it is computed and encoded, but that result is not used). Another step is to remove the Annotated structure entirely.

Fixes #88. Also relevant to #119


piet-gpu/src/lib.rs
```diff
@@ -401,6 +412,19 @@ impl Renderer {
         cmd_buf.end_debug_label();
         cmd_buf.write_timestamp(&query_pool, 1);
         cmd_buf.memory_barrier();
+        cmd_buf.begin_debug_label("Clip bounding box calculation");
+        self.clip_binding
+            .record(cmd_buf, &self.clip_code, self.n_clip as u32);
+        cmd_buf.end_debug_label();
```
ishitatsuyuki, line 418: There is no memory barrier after the refactor. Probably OK, but might want to fix when accounting for the timestamp queries.

raphlinus: The lack of memory barrier is deliberate, though maybe could be highlighted in a comment. The convention for the clip dispatch is "barrier in, barrier out." The reason that's significant is that if there are no clips in the scene, there are no dispatches, hence no barriers.

I'll want to redo the timer queries somewhat, as the operations are reordered and there's at least one, perhaps two, additional stages that should get counted. That can happen in this PR or later (I plan to do a bunch of cleanup after the main architectural changes settle).

raphlinus:
Merging now even though there is cleanup needed; that can be done in a separate PR.

PR #157, opened 2022-03-07
Initial implementation of blend modes

dfrg:
* Add blend and composition mode enums to API
* Mirror these in the shaders
* Add new public blend function to PietGpuRenderContext that mirrors clip
* Plumb the modes through the pipeline from scene to kernel4


dfrg:
This is certainly caused by my incorrect alpha computation in `mix_compose` (https://github.com/dfrg/piet-gpu/blob/90774f1f46305a91ed99b770b301aedf6688dbe0/piet-gpu/shader/blend.h#L257). This generates `a^2` alpha components which are obviously wrong. If you replace it with the correct `vec4(cs, as) * fa + vec4(cb, ab) * fb` then the cli test works as expected, but some of the blend modes render with artifacts.

raphlinus:
Got it, thanks. I'm interested in coming back to this, but don't think it's critical path. Since my current work is based on this rev, what we might do is merge this, get variable encoding in, and develop a "more correct" blending computation as a separate PR. It sounds like that should be localized to fine rasterization, so unlikely to cause serious merge conflict.

raphlinus (approved):
Let's get this in now, and fix the alpha and other issues as a followon.

PR #158, opened 2022-03-14
Glyph api

dfrg:


dfrg:
This PR breaks the glyph rendering API into separate GlyphProvider and GlyphRenderer types. GlyphProvider produces owned pre-encoded outlines that can be generated on multiple threads and cached for later submission to GlyphRenderer for rasterization. The encoded outlines can also be used to calculate bounding boxes for computing accurate side-bearings and allocating slots in a texture atlas.

dfrg:
These changes are made obsolete by the new fragment API.

PR #159, opened 2022-03-14
Variable size encoding of draw objects

### LINKED ISSUE #152: Variable size scene encoding
Right now, the main scene encoding is a sequence of fixed size draw objects. This has been the case for a while, but more recently path segments and a couple other things moved out of the scene encoding. In addition, there's a stream of `Annotated` objects, also fixed size, with a 1:1 correspondence to the `Element` objects in the scene encoding.

It would be better for it to be fully variable size, I think. That would allow larger objects (gradients come to mind) without concern for bloating the encoding of other objects that may be more compact. Here's roughly what's involved.

The main step to unlock variable size encoding is to add two more prefix sums to the draw monoid: these count byte (or possibly u32) offsets within the element stream, and also a new variable size object to take the place of "annotated." Most of the functionality of Annotated moves to other places. The bounding box gets its own stream (this is generally a move towards SoA). anything currently just copied from Element is replaced by a read from the scene buffer (there's no value in copying it; this was originally seen as a simplification, so the memory read patterns were more uniform). The main thing that remains is the line equation for linear gradients (and a similar computation for radial gradients, when that's done).

There's nothing particularly deep about these changes. They follow the same general pattern as the element pipeline changes. Also there should be cleanup involved, the Rust-side encoding will I think no longer use piet-gpu-derive at all, and in general that could be removed.

raphlinus:
This patch switches to a variable size encoding of draw objects.

In addition to the CPU-side scene encoding, it changes the representation of intermediate per draw object state from the `Annotated` struct to a variable "info" encoding. In addition, the bounding boxes are moved to a separate array (for a more "structure of "arrays" approach). Data that's unchanged from the scene encoding is not copied. Rather, downstream stages can access the data from the scene buffer (reducing allocation and copying).

Prefix sums, computed in `DrawMonoid` track the offset of both scene and intermediate data. The tags for the CPU-side encoding have been split into their own stream (again a change from AoS to SoA style).

This is not necessarily the final form. There's some stuff (including at least one piet-gpu-derive type) that can be deleted. In addition, the linewidth field should probably move from the info to path-specific. Also, the 1:1 correspondence between draw object and path has not yet been broken.

Closes #152


piet-gpu/shader/coarse.comp
```diff
@@ -283,15 +289,23 @@ void main() {
...
+                if (is_clip) {
+                    uint drawmonoid_base = drawmonoid_start + 4 * element_ix;
+                    uint scene_offset = memory[drawmonoid_base + 2];
+                    uint dd = drawdata_start + (scene_offset >> 2);
+                    uint blend = scene[dd];
+                    is_blend = (blend != BlendComp_default);
+                }
                 include_tile = tile.tile.offset != 0 || (tile.backdrop == 0) == is_clip
-                    || (is_clip && (anno_tag.flags & 0x2) != 0);
+                    || (is_clip && is_blend);
```
dfrg, line 308: Minor nit: is_blend is only set if is_clip is already true.

piet-gpu/shader/coarse.comp
```diff
@@ -146,6 +151,10 @@ void main() {
     uint part_start_ix = 0;
     uint ready_ix = 0;
 
+    uint drawmonoid_start = conf.drawmonoid_alloc.offset >> 2;
+    uint drawtag_start = conf.drawtag_offset >> 2;
+    uint drawdata_start = conf.drawdata_offset >> 2;
+    uint drawinfo_start = conf.drawinfo_alloc.offset >> 2;
```
dfrg, line 157: Is it worth storing these as indices rather than byte offsets? I don't think this is a perf concern, but it might improve clarity if all memory references are at 32-bit granularity.

dfrg (approved):
A few small points above but nothing major. Dropping annotated feels like a big win. My only concern is that the concentration of magic numbers seems to be growing. We should consider adding abstractions for reading/writing various bits of the streams. Still, this looks good to me to merge as is.

piet-gpu/shader/coarse.comp
raphlinus: This is a good question and worth reconsidering. Basically this code started in Metal where pointers are byte indices, and more or less the same is true in HLSL. Similarly offsets for CPU-side encoding are more naturally at byte granularity. But I agree this might be worth reconsidering now that we're writing more low-level GLSL.

raphlinus:
Agree we should improve the magic constants and consolidate types and reader/writer functions into header files. Will try to roll that up into future updates.

PR #160, opened 2022-03-31
Radial gradients

raphlinus:
This patch adds radial gradients, including both the piet API and some
new methods specifically to support COLRv1, including the ability to
transform the gradient separately from the path.


dfrg (approved):
Aside from the API extensions to accommodate COLRv1 radial gradients, this PR looks really clean to me and I’m happy with the (relative) simplicity of the math in k4.

PR #162, opened 2022-04-04
Factored out file loading on render tick for winit bin

### LINKED ISSUE #161: Duplicate reparsing in SVG Rendering native example
render_svg() in the example is loading and reparsing the file every frame, which was definitely not intended... I'll make a PR tomorrow or soon when I have time to fix that.

simbleau:
Closes #161


simbleau:
Just wanted to leave a note that I had to `pub use PicoSvg`. That is about the only potential side effect.

PR #164, opened 2022-04-15
Timer queries on more platforms, particularly Metal

raphlinus:
Change the API of compute dispatches to support explicit compute passes, an API more similar to WebGPU than the previous Vulkan inspiration. This change is to support timer queries on Apple Silicon. This will also make Metal more efficient, as previously each dispatch got its own compute pass. This change also aggregates multiple blit (copy) commands into a single encoder; no API change was made for that, partly because it's not part of the WebGPU API, and partly because timer queries there are not very interesting.

Current status: the piet-gpu-hal module (including the collatz example) have the new API (with queries set on compute pass) implemented. The other uses have not yet been updated.

On Metal, only M1 is tested. The "command" counter style is partly implemented, but not fully wired up.

As a next step, as we're updating clients, it's probably best to just return n time stamps rather than n - 1 (subtracting off the first one). That logic has been moved to fetch_query_pool in hub (out of the individual back-ends).


piet-gpu-hal/src/hub.rs
```diff
@@ -692,6 +722,32 @@ impl Drop for SubmittedCmdBuf {
...
+        descriptor_set: &DescriptorSet,
+        workgroup_count: (u32, u32, u32),
+        workgroup_size: (u32, u32, u32),
+    ) {
+        self.cmd_buf
+            .cmd_buf()
+            .dispatch(pipeline, descriptor_set, workgroup_count, workgroup_size);
+    }
+
+    pub unsafe fn end(&mut self) {
```
dfrg, line 746: Should this consume self? It seems like the right choice semantically but there might be ergonomic issues of which I'm unaware.

raphlinus: Good question. The underlying issue is that you can't easily have an encoder embedded in some other struct, otherwise you get errors like "cannot move out of." It's possible to address those in turn by writing `Option<Encoder>` and doing a `take`, but that's clunky. I ran into similar problems in resource (image, buffer) destructors but now that I think about it, we might not have the same problem. It's likely that `end` will be called in the same scope as encoder creation. I'll make the change and we can revert it if it turns out to cause problems.

I'll also highlight some ergonomics issues. For a lot of the "stage" stuff in piet-gpu, we'll need to pass in a compute encoder, perhaps instead of a command buffer, perhaps in addition to it. Things get more complicated if any of those stages need to do blit or clear operations; I haven't carefully audited the code. If so, then there will need to be more than one compute encoder. It should still be doable to have results from multiple encoders into a single query pool, but I still would like to make the change of returning timestamps directly rather than subtracting off the first one.

dfrg:
This code currently renders nothing on my 12" Intel MacBook (HD Graphics 615), but I get the same result with master so the problem is not directly related to this PR. The ComputePass refactoring gives expected results on M1 so I went ahead and pushed that commit. 

I'll setup a rust toolchain on my 16" MBP is see if it exhibits the same problem. We've been developing primarily on M1 for a while and it's possible we broke something on Intel along the way.

dfrg:
This should now be working on both Intel and M1. There is additional work to do on accounting along with potentially returning the raw timestamps, but the functionality for sampling is in place for both styles and all code has been refactored to use ComputePass.

The Intel rendering bug has to do with formats. Copying between textures requires both to have the same pixel format. On Intel, BGRA8Unorm is selected for the swapchain which doesn't match the RGBA8Unorm format that is used in create_image2d for constructing the target image. There are a few different approaches to fixing this, so I didn't include it in this PR.

piet-gpu-hal/src/hub.rs
dfrg: > I'll also highlight some ergonomics issues. For a lot of the "stage" stuff in piet-gpu, we'll need to pass in a compute encoder, perhaps instead of a command buffer, perhaps in addition to it. Things get more complicated if any of those stages need to do blit or clear operations; I haven't carefully audited the code.

Followup: all uses were clean with the stages requiring only dispatches and memory barriers so the conversion to ComputePass (with consuming self) was fairly trivial.

piet-gpu-hal/src/backend.rs
```diff
@@ -159,14 +160,32 @@ pub trait Device: Sized {
...
     unsafe fn reset(&mut self) -> bool;
 
+    /// Begin a compute pass.
+    ///
+    /// State: ready -> in_compute_pass
+    unsafe fn begin_compute_pass(&mut self, desc: &ComputePassDescriptor);
+
+    /// Dispatch
+    ///
+    /// State: in_compute_pass
```
raphlinus, line 188: Note (partly to self): I started writing these state machine transitions in the docs but ran out of steam. In some cases, the implementation won't be strict, but it probably should be. Probably the best end state for this is debug asserts or some other kind of opt-in validation.

raphlinus (commented):
I can't approve this because I am the original PR author, but it looks good. Please feel free to approve and merge.

I didn't analyze everything in detail, but I'm happy now to get this merged and sort out problems as they arise.

piet-gpu-hal/src/backend.rs
dfrg: We can use the borrow checker to enforce this with a bit of refactoring.

dfrg (approved):

PR #165, opened 2022-04-24
Bbox union

raphlinus:
This is a rough draft of the compute shaders for computing bounding box union, just to checkpoint the current status.


raphlinus:
Closing as not suitable for merging into main. There is likely future work that will refer to this, but that can be done against a closed PR.

PR #168, opened 2022-05-10
New scene/resource API 

dfrg:
This is a rough but functioning cut of a new scene API supporting both retained, composable fragments and the initial foundation for a resource framework. In addition, this includes a C API that is (for now) designed specifically for rendering glyphs on Metal. 

There are no significant changes to existing code, but the PR is large and will definitely require some (perhaps not so) fine tuning before merging.


pgpu-render/pgpu.h
```diff
@@ -0,0 +1,140 @@
+#include <cstdarg>
```
raphlinus, line 1: I think this file is auto-generated? It's actually a bit hard to tell as some of the comments look hand-written. In any case, would it be possible to add a line at or near the top explaining what's going on?

piet-scene/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+license = "MIT/Apache-2.0"
+edition = "2021"
+
+[dependencies]
+bytemuck = { version = "1.7.2", features = ["derive"] }
+smallvec = "1.8.0"
+pinot = "0.1.5"
+moscato = "0.1.2"
+
+# remove these and move demo to another directory
```
raphlinus, line 13: Should this comment be labeled "TODO"?

piet-scene/src/brush/color.rs
```diff
@@ -0,0 +1,58 @@
...
+        Self { r, g, b, a: 255 }
+    }
+
+    pub const fn rgba8(r: u8, g: u8, b: u8, a: u8) -> Self {
+        Self { r, g, b, a }
+    }
+
+    pub fn rgb<C: Into<f64>>(r: C, g: C, b: C) -> Self {
+        Self::rgb8(
+            (r.into() / 255.0) as u8,
```
raphlinus, line 36: It might be worth looking at the piet sources for this. This conversion neither rounds nor clamps to (0..=255) so I don't think it's quite ideal. Given that it's low level, it might not be important surfacing the float versions here.

piet-scene/src/brush/color.rs
```diff
@@ -0,0 +1,58 @@
...
+        Self::rgba8(
+            (r.into() / 255.0) as u8,
+            (g.into() / 255.0) as u8,
+            (b.into() / 255.0) as u8,
+            (a.into() / 255.0) as u8,
+        )
+    }
+
+    pub fn to_premul_u32(self) -> u32 {
+        let a = self.a as f64 / 255.0;
```
raphlinus, line 52: I usually write this as a multiplication by (1.0 / 255.0), as this generates `divsd` (checked in godbolt). Also I think it would be better with rounding.

piet-scene/src/geometry.rs
```diff
@@ -0,0 +1,225 @@
+// Copyright 2022 The piet-gpu authors.
```
raphlinus, line 1: Some of this code seems pretty familiar. That's ok, but maybe it should say something like "based in part on kurbo."

Similarly, a pretty common pattern would be to have kurbo as an optional dependency, and have conversions in place. I'm not sure that should be done this round, perhaps in future. I think the Rust traits for approximate conversions (f64->f32 and friends) were still a bit controversial last time I took a serious look at this, not sure if that's been settled by now.

raphlinus (approved):
Overall looks good to me, and quite ambitious the way it subsumes so much of the encoding task. I have some suggestions for improvements, but none blocking - I'd be happy to build future stuff on top of this, and continue to iterate quality-of-life features such as kurbo conversions.

pgpu-render/pgpu.h
dfrg: This is auto-generated. I'll configure the generator to add a comment at the top marking it as such.

piet-scene/Cargo.toml
dfrg: Oops. Good catch. I'm happy to just remove the binary altogether along with its dependencies. I believe we discussed adding a more structured example framework to the project and it seems like "soon" might be a good time to do that.

piet-scene/src/brush/color.rs
dfrg: Thanks for not pointing out that all of these divisions should be multiplications anyway :) Clearly I never made use of the float versions. I'll go ahead and drop them.

piet-scene/src/brush/color.rs
dfrg: Always nice to facilitate better codegen. Will make both of these changes.

piet-scene/src/geometry.rs
dfrg: I did pull most of the Affine functions from kurbo. I'll add an attribution. 

Happy to prep another PR adding the optional dependency along with the conversions and changing some signatures to take `impl Into<T>` for the appropriate types.

PR #172, opened 2022-05-19
Fix query pool size

raphlinus:
This was causing harmless warnings on mac but causing rendering to fail on Pixel 6.

### DIFF
--- a/piet-gpu/bin/cli.rs
+++ b/piet-gpu/bin/cli.rs
@@ -232,7 +232,7 @@ fn main() -> Result<(), Error> {
         let session = Session::new(device);
 
         let mut cmd_buf = session.cmd_buf()?;
-        let query_pool = session.create_query_pool(8)?;
+        let query_pool = session.create_query_pool(12)?;
 
         let mut ctx = PietGpuRenderContext::new();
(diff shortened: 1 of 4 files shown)


dfrg:
While this is in PR form, should we just make this an associated constant... Renderer::QUERY_POOL_SIZE?

piet-gpu/bin/cli.rs
```diff
@@ -232,7 +232,7 @@ fn main() -> Result<(), Error> {
         let session = Session::new(device);
 
         let mut cmd_buf = session.cmd_buf()?;
-        let query_pool = session.create_query_pool(8)?;
+        let query_pool = session.create_query_pool(12)?;
```
dfrg, line 235:
```suggestion
        let query_pool = session.create_query_pool(Renderer::QUERY_POOL_SIZE)?;
```

dfrg: missed one :)

raphlinus: Gah, I had it in my buffer, just didn't save to the PR. Good catch thanks.

dfrg (approved):
lgtm!

PR #173, opened 2022-05-19
Blend mem

### LINKED ISSUE #156: Memory for blend state
In this issue is the design that will hopefully resolve issues with the blend state, mostly #83 but it's also come up in #155 and possibly other issues.

There are several parts. I'll work backward from fine rasterization, as that may be a clearer presentation.

In fine rasterization, the blend stack is split into two segments. The first segment is a *small* array in vector registers (ie function scope). The exact size of that array is a tunable parameter, small enough it doesn't cause a spill to scratch, large enough to accommodate "most" blending. Also note, this parameter interacts with CHUNK (the number of pixels rendered by a single thread). My gut feeling is that reducing CHUNK to 4 and an array size of 4 will be close to the sweet spot.

The second segment is a region of memory in a separate read/write blend stack binding. This way, the memory (ptcl) binding can be read-only, which appears to be important on at least some hardware (Pixel 4, see #83). The offset of that region is written into the ptcl by coarse rasterization.

The logic to select between the two is a simple `clip_depth < BLEND_STACK_SIZE` predicate. I've considered more complex logic, something like a stack window, but I think keeping it simple is a win.

In coarse rasterization, we reinstate the logic by Elias to compute maximum blend stack depth. However, we subtract off the small array size and only allocate if it's above that. Note that this lifts a hard limit on blend depth, but at the same time introduces a new requirement for memory allocation. Obviously the blend stack buffer needs to be large enough to hold all allocations. We'll have another architecture issue coming up soon with a more detailed plan for that. For the time being, it's come up with a number for the size of that buffer, cross fingers and hope it's big enough.

One more point, addressing approaches considered and rejected. It's disappointing that memory used by the blend stack can't be recovered after a workgroup does its rendering. I thought about a malloc/free approach, but there are a number of problems. One is that free requires release semantics and malloc requires acquire, if memory is to be reused, and this barrier is not available on current Metal. Another is that while it's very likely that the number of concurrently resident workgroups is much smaller than the total, it is very difficult to quantify that in a reliable, portable way. Thus, a conservative or worst-case estimate for the required allocation may not be significantly smaller than adding up the total allocation, and the difference in complexity is significant.

raphlinus:
Split the blend stack into register and memory segments. Do blending in registers up to that size, then spill to memory if needed.

This version regresses performance on Pixel 4 (as determined by experimentation), as it uses common memory for the blend stack, rather than keeping that memory read-only in fine rasterization, and using a separate buffer for blend stack. We'll want to address this in a followup. A bit of experimentation suggests that either having a separate buffer binding for blending, or using scratch memory (a large array for the second segment) addresses the performance issue. The first of these approaches will work across pretty much all devices, but is a bit ugly and will require logic to determine the size of the buffer allocation, as well as (at some point) dynamically deal with overflow.

Also a flaw in this version: there is no checking of memory overflow.

For understanding code history: this commit largely reverts #77, but there were some intervening changes to blending, and this commit also implements the split so some of the stack is in registers.

Closes #156


piet-gpu/shader/kernel4.comp
```diff
@@ -236,24 +239,38 @@ void main() {
...
             clip_depth++;
             cmd_ref.offset += 4;
             break;
         case Cmd_EndClip:
             CmdEndClip end_clip = Cmd_EndClip_read(cmd_alloc, cmd_ref);
-            uint blend_mode = uint(end_clip.blend >> 8);
-            uint comp_mode = uint(end_clip.blend & 0xFF);
             clip_depth--;
+            uint base_ix;
+            if (clip_depth < BLEND_STACK_SPLIT) {
```
dfrg, line 262: I believe this should be `>=`

dfrg:
All of the logic in place so far looks good.

I assume we want to fill in the allocation logic before merging? Right now, spills seem to access random locations in memory.

piet-gpu/shader/coarse.comp
```diff
@@ -451,5 +459,8 @@ void main() {
     }
     if (bin_tile_x + tile_x < conf.width_in_tiles && bin_tile_y + tile_y < conf.height_in_tiles) {
         Cmd_End_write(cmd_alloc, cmd_ref);
+        if (max_blend_depth > BLEND_STACK_SPLIT) {
```
raphlinus, line 462: Oops. I am surprised my tests worked with this still in place. I guess I was lucky that the memory happened to be accessible.

piet-gpu/shader/kernel4.comp
raphlinus: Yes.

raphlinus (commented):
Will do another round to address the flaws, and now I'm concerned the clip test doesn't adequately exercise the logic.

raphlinus:
Ok this is updated now. I also validated it a bit more carefully, but locally patching BLEND_STACK_SPLIT to 1 and observing that there are rendering artifacts in the version of the previous commit, and no artifacts in the proposed version. Obviously testing could be more rigorous.

dfrg (approved):
I had just assumed you were deferring the spill allocation case for later. This all looks good to me now.

PR #178, opened 2022-06-17
Shader compilation in GitHub Actions

### LINKED ISSUE #177: Shader compilation plan
It's super annoying and error-prone to compile shaders and check the generated code into the repo. Yet there are advantages to having the generated code checked in, particularly not requiring extra tools to build and run, and also pinning those tools to a specific version, for repeatability.

The proposed solution is for the source of truth to be GitHub Actions that compile the shaders. Here is a somewhat more detailed plan.

All commits (and PRs) go to a newly created `dev` branch. This branch has no shader/gen directories. Thus, the PR contains changes to the shader source but not any compiled artifacts. The `main` branch tracks `dev` closely but has the gen subdirectories. A GitHub action runs on pushes to the dev branch and creates a commit to `main` updating the gen subdirectories.

This action runs on Windows, which gives us access to both the official DXIL signing DLL and also the Windows port of [Metal tools](https://developer.apple.com/documentation/metal/shader_libraries/building_a_library_with_metal_s_command-line_tools) (subject to figuring out how to download the latter). In future, we are likely to add to the sophistication of shader compilation in various ways, including getting metadata programmatically (binding renumbering on MSL/HLSL, workgroup size, etc), spinning more permutations. Instead of shell/ninja, we may move to a Rust tool, possibly using [hassle-rs](https://docs.rs/hassle-rs/latest/hassle_rs/) and related tools.

raphlinus:
This patch sets up very basic CI (right now just cargo fmt) but more importantly compiles shaders in a GitHub Action.

Any PR to branches other than main will run shader compilation. Any push to the dev branch will run shader compilation and then merge to main.

Closes #177

### DIFF
--- a/.github/workflows/shader.yml
+++ b/.github/workflows/shader.yml
@@ -1,0 +1,13 @@
+on:
+  pull_request:
+    branches-ignore:
+     - main
+
+jobs:
+  push-shaders:
+    runs-on: windows-latest
+    name: compile shaders
+    steps:
+      - uses: actions/checkout@v3
+      - run: mkdir piet-gpu/shader/gen
+      - uses: ./.github/actions/shader
--- a/.github/workflows/shader.yml
+++ b/.github/workflows/shader.yml
@@ -9,5 +9,4 @@ jobs:
     name: compile shaders
     steps:
       - uses: actions/checkout@v3
-      - run: mkdir piet-gpu/shader/gen
       - uses: ./.github/actions/shader
--- a/.github/actions/shader/action.yml
+++ b/.github/actions/shader/action.yml
@@ -16,7 +16,15 @@ runs:
     - name: install DXC
       uses: napokue/setup-dxc@v1.0.0
 
-    - name: run shader compilers
-      run: ninja
+    - name: 'run shader compilers: piet-gpu'
+      run: mkdir gen && ninja
       shell: pwsh
       working-directory: piet-gpu/shader
+    - name: 'run shader compilers: tests'
+      run: mkdir gen && ninja
+      shell: pwsh
+      working-directory: tests/shader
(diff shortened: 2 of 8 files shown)


raphlinus:
The CI failed for a technical reason (it doesn't run shader compilation on `main` but the PR is against `master`; I want to do the rename but haven't yet). Possibly I could do more fine-tuning to make this more robust, or in any case produce better error messages.

.github/workflows/shader.yml
```diff
@@ -0,0 +1,13 @@
...
+    branches-ignore:
+     - main
+
+jobs:
+  push-shaders:
+    runs-on: windows-latest
+    name: compile shaders
+    steps:
+      - uses: actions/checkout@v3
+      - run: mkdir piet-gpu/shader/gen
```
dfrg, line 12: Adding `-force` here should suppress the error if the directory already exists. It might be worth not doing this though and just rejecting any PRs that contain the gen directory?

dfrg:
I'm not the most qualified to review GitHub actions, but nothing here stands out as problematic to me. 

I assume that after this lands, all PRs should be directed to dev?

raphlinus:
Yes. I think something that should happen in this PR is that the README gets updated with clear language describing what's going on. People will want `main` if they're building, but `dev` if they're either making updates or wanting to look at clean PRs without the noise of generated compiled shaders. There's obviously a ton more that can be done in CI as well, but my current thinking is to land this with minimal changes (aside from the README) and add more as followup PRs.

One other question is that we have a few more gen directories (tests, examples for piet-gpu-hal). Should we do all of these so it's consistent?

dfrg (approved):
This all looks good to me. Now seems as good a time as any to make the change.

Issue #175, opened 2022-06-15
Robust dynamic memory


raphlinus:
This issue outlines our plans for robust dynamic memory in piet-gpu. Right now, we essentially hardcode "sufficiently large" buffers for the intermediate results such as path segments, per-tile command lists, and so on. The actual size needed is highly dependent on the scene, and difficult to predict without doing significant processing. This is also an area where support in modern GPUs is sadly lacking and might be expected to evolve. Thus, a satisfactory design involves some tradeoffs.

The goals are to (a) reliably render correct results, except in extreme resource-constrained environments relative to the scene (ie when the scene is adversarial in its resource requirements), (b) use modest amounts of memory when more is not required, and (c) not impact performance too much. These goals are in tension.

The general strategy is similar to what's already partially supported in the code, but currently lacking a full implementation CPU-side. A small number of atomic counters (currently 1, but this will increase, as described below) provide bump allocation, then all memory writes are conditional on the offset being less than the allocated buffer size. Currently, we have logic to early-out when memory is exceeded, but we may replace that with logic to proceed when allocation was successful at the *input* stage, so that the atomic counter accurately reflects how much memory would be needed for the successive stage to succeed. (Fine implementation detail: if the number of stages is no more than 32, then atomicOr a bit corresponding to the failed stage, and for early-out do a relaxed atomic load testing bits for input dependencies. If greater than 32, atomicMax the stage number, which is assigned in *decreasing* dependency order (so if A depends on B, A < B), and again do relaxed atomic load confirming this value is less than the minimum of the input dependencies)

Currently there is one command buffer submission all the way from the input scene to the presentation of the rendered frame buffer. In this proposal, we split that in half and fence back to the CPU after the first submission. The first submission is everything up to fine rasterization. On fence back, the CPU checks the value of the atomic counter, and if it's less than or equal to the buffer size, submits a command buffer for fine rasterization.

If it's greater than buffer size, it reallocates the buffer, rebinds the descriptor sets, and tries again. One reasonable choice for the new size of the buffer is the value of the atomic counter, perhaps rounded up a bit. A heuristic could refine this, for example if the error is at an earlier pipeline stage, a multiplier could be applied on the reasonable assumption that later stages will also require additional memory.

[...]

PR #179, opened 2022-06-23
Add HAL methods to update descriptor set

raphlinus:
This patch adds the capability to the HAL to update descriptor sets.

Part of the work for #175 but might be useful for other purposes (such as window resize) and is fairly self-contained.

Note that there are rustfmt changes in unrelated areas, due to rustfmt not being reliably applied through macros. I'd like to figure out a better solution to that in CI.

### DIFF
--- a/piet-gpu-hal/src/backend.rs
+++ b/piet-gpu-hal/src/backend.rs
@@ -98,6 +98,36 @@ pub trait Device: Sized {
         builder.build(self, pipeline)
     }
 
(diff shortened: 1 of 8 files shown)


dfrg (approved):
This all looks simple, clean and correct to me.

One thought on related work: we will need some support for (non-storage) image and sampler descriptors for image rendering but my inclination is to special case and encapsulate the whole render pipeline state in HAL for now and expose the detailed bits if/when it becomes necessary.

Good to merge!

PR #180, opened 2022-06-25
Make `pgpu-render`  compile on non apple platforms

DJMcNab:
This makes use of a cfg to disable the metal dependencies and the functions they are used in on ios or macos.
This is also reflected in the generated `cbindgen` code, using the `with_define` feature.

Also add `resolver="2"` to the workspace root, primarily in case we gain a `wgpu` dependency.

I've also added a fix for the android example

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 
 members = [
     "pgpu-render",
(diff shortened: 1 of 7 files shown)


raphlinus (commented):
This seems good to me. @dfrg can you confirm, as this is mostly in your code? Also this might conflict with some of the CI changes I'm making.

raphlinus (approved):

PR #181, opened 2022-07-08
Implement robust dynamic memory

### LINKED ISSUE #83: Slow fine rasterization (k4) on Adreno 640
I'm getting piet-gpu running on Android (see #82 for snapshot), and running into fine rasterization being considerably slower than expected. The other stages in the pipeline seem fine. I've done some investigation but the fundamental cause remains a mystery.

Info so far: the Adreno 640 (Pixel 4) has a default subgroup size of 64 (though it can also be set to 128 using the vk subgroup size control). That should be fine, as it means that the memory read operations from the per-tile command list are amortized over a significant number of pixels even if CHUNK (the number of pixels written per invocation of kernel4.comp) is 1 or 2. If CHUNK is 4, then the workgroup and subgroup sizes are the same; any larger value results in a partially filled subgroup. There's more detail about the [Adreno 6xx](https://gitlab.freedesktop.org/freedreno/freedreno/-/wikis/A6xx-SP) on the freedreno wiki.

There are some experiments that move the needle. Perhaps the most interesting is that commenting out the body of `Cmd_BeginClip` and `Cmd_EndClip` at least doubles the speed. This is true even when the clip commands are completely absent from the workload. My working hypothesis is that register pressure from accommodating the clip stack and other state is reducing the occupancy.

Another interesting set of experiments involves adding a per-pixel ALU load. The amount of time taken increases significantly with CHUNK. I'm not sure how to interpret that yet. Tweaking synthetic workloads like this may will be the best way to move forward, though I'd *love* to be able to see the asm from the shader compilation. I'm looking into the possibility of running this workload on the same (or similar hardware) but a free driver such as freedreno so that I might be able to gain more insight that way.

I've been using [Android GPU Inspector](https://github.com/google/agi) (make sure to use at least 1.1) but so far it only gives me fairly crude metrics - things like % ALU capacity and write bandwidth scale with how much work gets done, and other metrics like % shaders busy and GPU % Utilization are high in all cases.

There are other things I've been able to rule out: a failure of loop unrolling by the shader compiler. Failure to account ptcl reads as dynamically uniform (I manually had one invocation read and broadcast the results, yielding no change).

I do have some ideas how to make things faster (including moving as much of the math as possible to f16), but the first order of business is understanding *why* it's slow, especially when we don't seem to be seeing similar problems on desktop GPUs.

### LINKED ISSUE #175: Robust dynamic memory

raphlinus:
This is the core logic for robust dynamic memory. There are changes to both shaders and the driver logic.

On the shader side, failure information is more useful and fine grained. In particular, it now reports which stage failed and how much memory would have been required to make that stage succeed.

On the driver side, there is a new RenderDriver abstraction which owns command buffers (and associated query pools) and runs the logic to retry and reallocate buffers when necessary. There's also a fairly significant rework of the logic to produce the config block, as that overlaps the robust memory.

The RenderDriver abstraction may not stay. It was done this way to minimize code disruption, but arguably it should just be combined with Renderer.

Another change: the GLSL length() method on a buffer requires additional infrastructure (at least on Metal, where it needs a binding of its own), so we now pass that in as a field in the config.

This also moves blend memory to its own buffer. This worked out well because coarse rasterization can simply report the size of the blend buffer and it can be reallocated without needing to rerun the pipeline. In the previous state, blend allocations and ptcl writes were interleaved in coarse rasterization, so a failure of the former would require rerunning coarse. This should fix #83 (finally!)

There are a few loose ends. The binaries haven't (yet) been updated (I've been testing using a hand-written test program). Gradients weren't touched so still have a fixed size allocation. And the logic to calculate the new buffer size on allocation failure could be smarter.

Closes #175


raphlinus:
Also note: this commit doesn't include generated shaders. We should land #178 and #179 then PR this against dev. But I wanted to get a draft out there to look at.

raphlinus:
Marking this as ready for review, as the command line drivers are updated. No changes have been made to pgpu_render, so hopefully that'll continue as before (requiring pre-allocation of adequate buffers).

dfrg (approved):
I've gone through this several times now and aside from some potential resource management improvements that have already been discussed (and that are marked in the code with comments), nothing stands out as needing attention. This is a huge step forward for the project-- let's get it merged :)

eliasnaur:
Was [shader/image.png](https://github.com/linebender/piet-gpu/pull/181/files#diff-f5fb01c848038f6cca0c9f64dcfaee65266a2209c1f4ea7ad5f27a1f8dec0301) intentionally checked in? I couldn't find a reference to it in the PR.

raphlinus:
Nope, that was by mistake. We'll remove it in a cleanup, thanks.

PR #183, opened 2022-07-13
Update winit to 0.26.1

raphlinus:
This is mostly to silence security warnings about a nix vulnerability.


dfrg (approved):
This should probably target `dev` rather than `main`. Otherwise, lgtm.

raphlinus:
Duh on the base. Maybe we want to have a warning generated for that.

PR #184, opened 2022-07-14
Separate Instance and Surface creation in HAL

dfrg:
This allows surfaces to be created independently from instances and devices. The intention is primarily to support dynamic window management, particularly in the context of UI.

The Metal and DX12 changes are straightforward as there are no dependencies between surface and instance/device creation.

Vulkan is a bit tricky. In this case, `InstanceFlags` has grown a `PRESENT` bit that directs instance creation to attempt to add the appropriate surface extensions for the target platform. This replaces `ash_window::enumerate_required_extensions`. Physical device and queue family selection has been modified to require both graphics and compute support. This is necessary to handle images anyway, and the current code operates on the assumption that such a device will also support presentation. This is likely to hold in practice on any device that has a display, but is probably not guaranteed by the spec.

We'll probably want a more sophisticated device selection API at some point, but that is out of scope for this work.

This has been tested on Mac and Windows (both Vulkan and DX12). The Android example has been updated, but not compiled or tested.


piet-gpu-hal/src/vulkan.rs
```diff
@@ -175,10 +173,33 @@ impl VkInstance {
             if cfg!(debug_assertions) {
                 has_debug_ext = exts.try_add(DebugUtils::name());
             }
-            if let Some(ref handle) = window_handle {
-                for ext in ash_window::enumerate_required_extensions(*handle)? {
-                    exts.try_add(ext);
+
+            // Enable platform specific surface extensions if presentation
+            // support is requested.
+            if support_present {
```
raphlinus, line 179: Not a blocker, but I'm curious what it would take to add fuchsia to this list.

piet-gpu-hal/src/vulkan.rs
```diff
@@ -1421,26 +1450,35 @@ impl Layers {
     }
 }
 
-unsafe fn choose_compute_device(
+unsafe fn choose_device(
     instance: &Instance,
     devices: &[vk::PhysicalDevice],
-    surface: Option<&VkSurface>,
+    support_graphics: bool,
```
raphlinus, line 1456: It's slightly confusing to follow this because `support_present` is used as the variable name in the caller, which is not quite consistent. I recommend picking one name then optionally adding comments to discuss.

piet-gpu-hal/src/vulkan.rs
```diff
@@ -1421,26 +1450,35 @@ impl Layers {
...
-            if let Some(surface) = surface {
-                if !surface
-                    .surface_fn
-                    .get_physical_device_surface_support(*pdevice, ix as u32, surface.surface)
-                    .unwrap()
-                {
-                    continue;
-                }
-            }
+            // TODO: is this strictly necessary? We'll need a queue supporting graphics
```
raphlinus, line 1465: Ordinarily I am not a fan of commented out code, but see why it's done here. If I were doing it, I would replace it with a comment, something like this:

We probably want to make the support for graphics non-optional, as we will require it for render pipelines to support image scaling. In addition, when rendering to a surface, to be more technically correct we should check whether the physical device has support for that surface. This probably won't be a problem in practice, though.

piet-gpu/bin/android.rs
```diff
@@ -54,9 +54,9 @@ fn my_main() -> Result<(), Error> {
                         let width = window.width() as usize;
                         let height = window.height() as usize;
                         let handle = get_handle(window);
-                        let (instance, surface) = Instance::new(Some(&handle), Default::default())?;
-                        gfx_state =
-                            Some(GfxState::new(&instance, surface.as_ref(), width, height)?);
+                        let instance = Instance::new(InstanceFlags::PRESENT)?;
+                        let surface = instance.surface(&handle)?;
```
raphlinus, line 58: This is missing an unsafe, but otherwise is tested to work on Pixel 4.
```suggestion
                        let surface = unsafe { instance.surface(&handle)? };
```

raphlinus (commented):
Overall looks pretty good. Needs a small fix on Android, and I have some other suggestions. Thanks!

raphlinus:
As a simplification, I would be in favor of making graphics support non-optional. I don't think it really buys us anything.

piet-gpu-hal/src/vulkan.rs
dfrg: Yes, since we need graphics support regardless, I think this parameter should just be removed. More comments on this on the next item.

piet-gpu/bin/android.rs
dfrg: Thanks! I should have caught that at least :)

dfrg:
> As a simplification, I would be in favor of making graphics support non-optional. I don't think it really buys us anything.

This works for me. Vulkan is the outlier here and I assume that speculatively adding the surface extensions doesn't have a significant cost, if any.

raphlinus (approved):
Looks good. Feel free to merge, and thanks.

PR #186, opened 2022-07-25
[pgpu-render] Fixes to PgpuRenderer::render

armansito:
- Fixed the `is_color` check to look for MTLPixelFormat::R8Unorm instead
  of MTLPixelFormat::A8Unorm since the backing texture format always
  uses the former.
- Added a call to metal::CommandBufferRef::finish() to clean up the blit
  command encoder to free the command buffer for new encoders.


DJMcNab:
The version of the pipeline this is targetting no longer exists. I suggest we close this

raphlinus:
Closing as it's pretty far out of date. I am hoping to pick up the integration work again, but it would be based on the new wgsl pipeline.

PR #188, opened 2022-08-08
Expose path rendering in C API

dfrg:
This PR exposes path rendering functionality through the C API. It also contains a minimal representation of brushes (supporting only solid colors) as a foundation for extension in the future. 

There are also a few bug fixes in the renderer and HAL. Specifically, a better heuristic for calculating the initial buffer size based on target dimensions and an additional flush method on command buffers that ensures any open encoder is committed.

### DIFF
--- a/pgpu-render/src/lib.rs
+++ b/pgpu-render/src/lib.rs
@@ -26,6 +26,9 @@
 
 mod render;
 
+use piet_scene::brush::{Brush, Color};
+use piet_scene::path::Element;
+use piet_scene::scene::Fill;
 use render::*;
 use std::ffi::c_void;
 use std::mem::transmute;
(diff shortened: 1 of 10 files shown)


pgpu-render/src/lib.rs
```diff
@@ -98,24 +101,239 @@ pub unsafe extern "C" fn pgpu_scene_destroy(scene: *mut PgpuScene) {
...
+    }
+}
+
+/// Sets the current absolute transform for the scene builder.
+#[no_mangle]
+pub unsafe extern "C" fn pgpu_scene_builder_transform(
+    builder: *mut PgpuSceneBuilder<'static>,
+    transform: *const PgpuTransform,
+) {
+    if !transform.is_null() {
```
raphlinus, line 280: I'm wondering if [as_ref](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref) here and in a few other places might be more concise than the is_null check.

raphlinus (approved):
Looks good! I wasn't quite expecting dyn iterator style crossing the ffi boundary for the paths, but I can see it being effective and reducing unneeded allocation.

pgpu-render/src/lib.rs
dfrg: Oh nice. I wasn't aware of that method. I'm happy to clean those up by I'd like to land this as is to avoid merge conflicts with #189. I can either add these changes as part of that PR or, probably better, do an FFI cleanup as a follow up to avoid making that one any larger.

PR #189, opened 2022-08-08
Remove piet API and replace with scene fragments

dfrg:
This PR removes the dependency on the piet crate and replaces it with piet-scene. It also includes some renaming of types and reorganization of the scene crate to provide a more ergonomic API. Some more work is probably necessary here, certainly with regard to documentation. 

The swash dependency is also removed and very simple text support has been added (based on piet-scene/moscato) to support rendering info strings.

One open question is whether to retain the ResourceContext as part of the public API. This must be kept in sync with scene construction and rendering and is a potential point of failure. The original intent was to support persistent (retained) brushes in addition to caching gradient ramps, but persistent brushes seem unnecessary and I'm inclined to late bind ramps as well and move the cache back to Renderer. This makes patching ramp ids more complicated, but support for such patching is necessary for images anyway, so it makes sense to reuse it and simplify the public API.


dfrg:
This is layered on top of #188 to enable a clean merge, so that needs to land first.

dfrg:
Closing this to resubmit a clean PR without overlapping C API commits.

PR #190, opened 2022-08-12
Remove piet API and replace with scene fragments

dfrg:
This PR removes the dependency on the piet crate and replaces it with piet-scene. It also includes some renaming of types and reorganization of the scene crate to provide a more ergonomic API. Some more work is probably necessary here, certainly with regard to documentation.

The swash dependency is also removed and very simple text support has been added (based on piet-scene/moscato) to support rendering info strings.

Finally, gradient ramp caching is moved back to piet-gpu and handled during scene upload. This has the side effect of making the Scene and SceneFragment types nearly identical. It's worth discussing whether it makes sense to retain the distinct types.


raphlinus (approved):
I took a quick pass over this. We should definitely get it merged, as there's about to be a bunch of activity, and already there are merge conflicts (hopefully will be simple to resolve).

bzm3r:
Nice!

PR #194, opened 2022-10-26
Use 'C style' preprocessing in piet-wgsl

DJMcNab:
This lets us use https://github.com/wgsl-analyzer/wgsl-analyzer for writing the wgsl files.
The imports (for `wgsl-analyzer`) have to be machine specific at the moment - to use this you need to configure `.vscode/settings.json` yourself.  The alternative is to point them at static files on GitHub, which is tempting to make things easier, but would potentially go out of sync with what is actually used.

There are a few issues with `wgsl-analyzer` which we run into at the moment:

1. https://github.com/wgsl-analyzer/wgsl-analyzer/pull/32 hasn't landed in a release (now in [0.5.1](https://github.com/wgsl-analyzer/wgsl-analyzer/releases/tag/v0.5.1))
2. `while` isn't supported - done in https://github.com/wgsl-analyzer/wgsl-analyzer/pull/37
3. The imports have to be machine specific - see above
4. `bitcast`'s use of 'generics' is broken
5. It doesn't "like" `array<Segment>` being used in a `storage` var, cascading errors. This *seems* resolved, and the cascading errors come from `vec2(x, y)` without a specific type being currently unsupported
6. `firstTrailingBit` is not supported - done in https://github.com/wgsl-analyzer/wgsl-analyzer/pull/38

I'm planning on working through these issues upstream and resolving where feasible. Hopefully a new release will come soon.
The use of `wgsl-analyzer` is of course entirely optional until these are resolved. 1 and 2 are the most annoying, but the others are frustrating on their own.


raphlinus (approved):
Ok, I've played with it and it seems to do what it says on the tin. It will definitely be annoying to work with until the problems with wgsl-analyzer are resolved, but if that happens quickly it should be a nice quality of life improvement.

Suggesting we merge rather quickly to minimize merge conflicts, as there's a whole bunch of other stuff in flight.

Thanks!

PR #196, opened 2022-10-29
WGSL: Allow imports anywhere

DJMcNab:
Permit using an import anywhere within the file.

Note that only certain locations are supported by wgsl-analyzer. However, this is better than requiring them also to be on newlines. If you need a new place, we can probably add it to wgsl-analyzer.

We intentionally only support `ifdef` etc. at the start of a line, to avoid having to think too hard about it.

The preprocessor also doesn't have any tests, so if something here looks wrong it probably is.

### DIFF
--- a/piet-wgsl/src/shaders/preprocess.rs
+++ b/piet-wgsl/src/shaders/preprocess.rs
@@ -41,7 +41,7 @@ pub fn preprocess(
 ) -> String {
     let mut output = String::with_capacity(input.len());
     let mut stack = vec![];
-    'outer: for (line_number, mut line) in input.lines().enumerate() {
+    'all_lines: for (line_number, mut line) in input.lines().enumerate() {
         loop {
             if line.is_empty() {
                 break;
@@ -51,14 +51,16 @@ pub fn preprocess(
             let hash_index = match (hash_index, comment_index) {
                 (Some(hash_index), None) => hash_index,
                 (Some(hash_index), Some(comment_index)) if hash_index < comment_index => hash_index,
+                // Add this line to the output - all directives are commented out or there are no directives
                 _ => break,
             };
             let directive_start = &line[hash_index + '#'.len_utf8()..];
             let directive_len = directive_start
-                .chars()
-                .take_while(|char| char.is_alphanumeric())
-                .map(char::len_utf8)
-                .sum();
+                // The first character which can't be part of the directive name marks the end of the directive
+                // In practise this should always be whitespace, but in theory a 'unit' directive
+                // could be added
+                .find(|c: char| !c.is_alphanumeric())
+                .unwrap_or(directive_start.len());
             let directive = &directive_start[..directive_len];
             let directive_is_at_start = line.trim_start().starts_with('#');
 
@@ -75,7 +77,8 @@ pub fn preprocess(
                         active: mode == exists,
                         else_passed: false,
                     });
-                    continue 'outer;
+                    // Don't add this line to the output; instead process the next line
+                    continue 'all_lines;
                 }
                 "else" => {
                     let item = stack.last_mut();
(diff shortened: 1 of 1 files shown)


piet-wgsl/src/shaders/preprocess.rs
```diff
@@ -71,29 +87,53 @@ pub fn preprocess(
...
+                        directive_end.find(|c: char| !c.is_whitespace())
+                    {
+                        import_name_start
+                    } else {
+                        eprintln!("#import needs a non_whitespace argument (line {line_number})");
+                        continue 'outer;
+                    };
+                    let import_name_start = &directive_end[import_name_start..];
+                    let import_name_end_index = import_name_start
+                        .find(|c| !(matches!(c, '_') || c.is_alphanumeric()))
```
raphlinus, line 115: Any special reason `matches!` rather than just `c == '_'`? Looking at the code I'm surprised this compiles without `c` having an explicit type, maybe matches is forcing type equality while PartialEq would be looser.

piet-wgsl/src/shaders/preprocess.rs
```diff
@@ -41,25 +41,41 @@ pub fn preprocess(
 ) -> String {
     let mut output = String::with_capacity(input.len());
     let mut stack = vec![];
-
-    for (line_number, line) in input.lines().enumerate() {
-        let trimmed = line.trim();
-        if trimmed.starts_with("#") {
-            let trimmed = &trimmed[1..];
-            let val_idx = trimmed
+    'outer: for (line_number, mut line) in input.lines().enumerate() {
```
raphlinus, line 44: Oooh, fancy, labeled loops. I found the flow a little hard to follow, but don't have a suggestion how to improve it.

piet-wgsl/src/shaders/preprocess.rs
```diff
@@ -71,29 +87,53 @@ pub fn preprocess(
...
+                    eprintln!("Unknown preprocessor directive `{val}` (line {line_number})");
                 }
             }
-        } else {
-            if stack.last().map(|x| x.active).unwrap_or(true) {
-                output.push_str(line);
-                output.push('\n');
-            }
+        }
+        if stack.last().map(|x| x.active).unwrap_or(true) {
```
raphlinus, line 134: Is this actually correct? It seems like a sequence of `#ifdef resolves_false`, `#ifdef resolves_true`, `inner`, `#endif`, `#endif` will actually cause inner to be included. My intuition (without having experimented) is that we want an `all` on the `active` field of all entries on the stack.

piet-wgsl/src/shaders/preprocess.rs
```diff
@@ -71,29 +87,53 @@ pub fn preprocess(
...
+                    let import_name_start = if let Some(import_name_start) =
+                        directive_end.find(|c: char| !c.is_whitespace())
+                    {
+                        import_name_start
+                    } else {
+                        eprintln!("#import needs a non_whitespace argument (line {line_number})");
+                        continue 'outer;
+                    };
+                    let import_name_start = &directive_end[import_name_start..];
+                    let import_name_end_index = import_name_start
```
raphlinus, line 114: It seems to me this is doing the same logic as the calculation of directive_len but in a different way. I find this version slightly more intuitive and less fragile. Either way, unless there's a compelling reason it seems like the two should be the same.

raphlinus (approved):
Generally looks good but I had a bunch of nits :)

piet-wgsl/src/shaders/preprocess.rs
DJMcNab: I think orginally there were several characters I wanted to allow there (e.g. `-`, I think)

It does also make the types work out, although obviously it's reasonable to be consistent.

piet-wgsl/src/shaders/preprocess.rs
DJMcNab: Yeah. In https://github.com/linebender/piet-gpu/pull/196/commits/40f808b366b4a7087b4da97f756e480c5265c859 I tried a different way, but that was really messy.

piet-wgsl/src/shaders/preprocess.rs
DJMcNab: Good point. Thanks

piet-wgsl/src/shaders/preprocess.rs
DJMcNab: Well spotted, I've used the same technique for `directive_len ` now

DJMcNab:
> Can we expect wgsl-analyzer to also be able to handle #import in the middle of a line?

In the sense that it's theoretically possible, yes. At the moment, the actual places are quite limited, but for any other places you need, adding such support should be *relatively* feasible

raphlinus (approved):
It occurs to me `all` is a monoid, so all the usual optimizations are possible. But I agree, let's optimize for code clarity, O(n) is fine if n is really small.

DJMcNab:
> Hmm, that means actually using imports for constants will result in a red squiggle? 

Yes exactly, but if you poke me on the [zulip](https://xi.zulipchat.com/) I'd guess it should take maybe 30 minutes per syntactic location.

I can only think that we need some attributes support (potentially just top level attribute to avoid having to think about nesting). Apparently parameter declarations was also needed - that supports already exists, but I'm not sure why.

You should already be able to import global constants, which is equivalent to having it be in any other place you can refer to constants, just with slightly more indirection.

I think everything else should be equivalently constructable from these cases.

I'll merge now though

PR #197, opened 2022-10-30
Don't encode empty glyph

raphlinus:
This fixes artifacts we are seeing in porting the "shello" demo to glazier, which was caused by appending glyphs with empty paths to the scene fragment. It's not obvious to me this is the best fix, as it could be done higher up by special casing glyphs with empty outlines, or lower down by allowing correct encoding of empty paths.

So partly it's a discussion of how to best fix the problem. I think a case can be made we shouldn't include empty paths in the encoding (though they can be made valid and evaluated correctly). It's likely that similar problems remain with empty paths made through the path building mechanism.


raphlinus:
Closing this because #200 does it better.

PR #199, opened 2022-11-05
Prepare example to isolate shader miscompilation

raphlinus:
When tile_alloc uses a workgroup shared variable to broadcase `tile_offset`, it gives wrong answers seemingly consistent with a nonexistent barrier on mac.


raphlinus:
Uploading not to merge but to help with investigating a potential shader miscompilation. When run on mac (just `cargo run` in the piet-wgsl subdir), it gives this output:

```
0: 6
1: 6
2: 20
3: 11
4: 3741d830 0.0000115540315
8: 9
9: 9
10: 23
11: 14
12: 3741d94e 0.000011554292
16: c
17: c
18: 26
19: 17
20: 3741da6c 0.000011554552
```

With exact values varying from run to run.

The correct output is:
```
0: 6
1: 6
2: 20
3: 11
8: 9
9: 9
10: 23
11: 14
12: 11e
16: c
17: c
18: 26
19: 17
20: 23c
```

The values at offsets 4, 12, and 20 are instances of `tile_offset + tile_subix` being written by tile_alloc.wgsl. The behavior is consistent with a missing barrier right before the `let tile_offset = sh_tile_offset;` statement; it seems to be reading garbage values, while threads in the same subgroup do appear to be getting correct values (this particular example doesn't display any values read from the same subgroup, but it should be reasonably straightforward to modify it so).

I looked at the naga output and the `threadgroup_barrier` is present, which casts suspicion on the Metal compiler itself. I think it's worth trying to isolate this further, as it's a worrisome miscompilation.

raphlinus:
I have indeed reduced the example a lot. Currently it's just running tile_alloc.wgsl, which is now a very simple shader that just does an atomic add (on one thread), followed by a barrier and writing the result of that add.

raphlinus:
Not intended to merge, so closing. Working around the miscompilation is still a relevant issue, though, and this could be done at many possible layers in the stack: in the shader (as is currently the case), disabling buffer robustness (or possibly changing the configuration to use clamping rather than conditionalization), or in naga to avoid patterns (such as passing threadgroup shared variables by reference to the main entry point) that trigger miscompilation. Probably an issue should be opened to track that work.

PR #200, opened 2022-11-08
More robust path encoding

dfrg:
This fixes the problem with desync when encoding empty paths. It also relaxes the restrictions on element sequences, inserting move-to elements where required.


raphlinus (approved):
This looks good, thanks!

One of the things I noticed while doing the wgsl port is that n_pathseg *might* be misused. It is not a suitable basis for scheduling workgroups and the like, as there can be pathtags that don't encode path segments (transform and the like). I'd like to audit those uses and possibly remove it (at least from export from scene data), but that's orthogonal to these fixes, which look right to me.

dfrg:
Interesting. I’m curious to see what the audit reveals.

PR #201, opened 2022-11-08
Piet wgsl wasm

armansito:
Run `cargo run-wasm piet-wgsl` from the root of the repository and navigate to http://localhost:8000 on Chrome Canary to view the demo.

There are a few rough edges that I would like to clean up in a follow-up eventually, in particular:
- Reorganize the wasm vs non-wasm example code
- Make the shader assembly code that has to use `include_str!` more compact. Shared and non-shared versions of the shaders currently need to be enumerated in two different places. The code could be made tighter using a procedural macro
- Add a texture output alternative to the current `out_buf` with a buffer usage flag. The blit from the output buffer to the swapchain texture currently involves some hand-coded resizing for the ImageCopyBuffer layout and extent to align with the output buffer dimensions and row stride.


piet-wgsl/src/main.rs
```diff
@@ -54,23 +76,186 @@ async fn do_render(
...
+        let bytes_per_pixel = std::mem::size_of::<u32>() as u32;
+        wgpu::Extent3d {
+            width: self.width,
+            height: self.height,
+            depth_or_array_layers: 1,
+        }
+    }
+}
+
+async fn run_wasm(event_loop: EventLoop<()>, window: Window) {
```
DJMcNab, line 126: It's not clear to me why the winit usage is locked to wasm.

Additionally, perhaps this should just be an example?

piet-wgsl/src/shaders/preprocess.rs
```diff
@@ -7,24 +7,26 @@ use std::{
 
 pub fn get_imports(shader_dir: &Path) -> HashMap<String, String> {
     let mut imports = HashMap::new();
-    let imports_dir = shader_dir.join("shared");
-    for entry in imports_dir
-        .read_dir()
-        .expect("Can read shader import directory")
+    // When targeting WASM, we need to directly bundle shader source strings at compile-time since
+    // we cannot read from the local build directory at runtime.
```
DJMcNab, line 11: This justification is true, however I'm not sure I follow how the inference was made. Another viable option may be to host the shaders on the web server, and then `fetch` them at runtime. That alternative doesn't appear to have been recognised here.

I think including them verbatim is probably the right tradeoff either way; however I suspect that will be the end state on all platforms, not just native.

Although at some point we should probably allow hot reload of the shaders for development (although these aren't exactly the kind of shaders that work very well with hot reload, it can't hurt either).

DJMcNab (commented):
The code looks reasonable; just a thought on organisation. Generally, a winit example isn't *too* useful when we don't support different sizes of canvas, but we can disable resize in winit on native for now.

I'm not clear on why you think a procedural macro would be needed over a declarative one.

raphlinus:
I'll close this with thanks, as it's subsumed by #220 (which adapted a lot of code in here) and pretty much working in the browser.

PR #198, opened 2022-11-05
Hacky tiger

raphlinus:
This is a snapshot of my prototype work on building out more of the full rendering pipeline in wgsl. It renders the tiger, but no other guarantees apply. I'll clean this up and do a real PR later.

PR #203, opened 2022-11-16
Extend pipeline

dfrg:
This PR pulls in the work to extend the WGSL port to the full pipeline.


DJMcNab:
This builds on top of https://github.com/linebender/piet-gpu/pull/198, right?

raphlinus:
> This builds on top of #198, right?

Yes. That has known flaws, but if we want, we can commit that (either with changes, which will force some rebasing here), or as-is, with the intent of fixing the flaws going forward.

dfrg:
I support anything that avoids rebasing. It looks like this completely subsumes that PR?

raphlinus:
Happy to close 198 and do all the iteration on this one.

DJMcNab (approved):
I can't afford the time to meaningfully review this, but I don't think there's any use. As I expressed in Zulip; since the base isn't useful, there's little to gain from not landing things quickly. Obviously if we think it's worth waiting for Raph to review your new code, that's also reasonable.

The CPU side code is pretty far from ideal, but it's clear to me that fixing that isn't an immediate term priority, and shouldn't block this PR. I would suggest removing the outdated infrastructure for the minimal rendering (i.e. `struct Shaders` and `fn render`), but other improvements and arranging can wait

raphlinus:
I haven't had time to actually review the code, but am fine with it being merged and we'll fix things going forward.

I'm of two minds on deleting the 5-stage pipeline. For one, we do have it running (in a branch) in browser. For two, keeping that pipeline running may be of value for rendering fonts for Skia. The argument against is that it's inadequately tested and may represent a burden to continue to maintain.

In view of our current theme of moving forward with maximal velocity and minimal friction, I would lean toward deleting it now and bringing it back later if and when it's needed. But I do not have a strong feeling about this.

dfrg:
Happy to merge this as is or wait for the completed work on clips. It’s a huge chunk of code so my expectation was more future debugging and iteration and less review.

dfrg:
Now renders the mozilla blend grid. Ship it!

PR #206, opened 2022-11-21
Allow arbitrary output sizes

DJMcNab:
This just took some simple wiring through to pass the dynamic size down into `render_full`, and then blit the image into a new buffer of the output size.

The command can now be ran as `cargo run -p piet-wgsl -- [width] [height]`, which each default to `1024`.

### DIFF
--- a/piet-wgsl/src/render.rs
+++ b/piet-wgsl/src/render.rs
@@ -6,6 +6,7 @@ use piet_scene::Scene;
 use crate::{
     engine::{BufProxy, Recording, ResourceProxy},
     shaders::{self, FullShaders, Shaders},
+    Dimensions,
 };
 
 const TAG_MONOID_SIZE: u64 = 12;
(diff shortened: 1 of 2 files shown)


piet-wgsl/src/render.rs
```diff
@@ -52,7 +53,14 @@ fn size_to_words(byte_size: usize) -> u32 {
     (byte_size / std::mem::size_of::<u32>()) as u32
 }
 
-pub fn render(scene: &Scene, shaders: &Shaders) -> (Recording, BufProxy) {
+pub const fn next_multiple_of(val: u32, rhs: u32) -> u32 {
+    match val % rhs {
```
raphlinus, line 57: One other way to write this is `val + rhs - 1 - (val + rhs - 1) % rhs`

raphlinus (approved):
Thanks! This is definitely a useful step toward making it dynamically sized for window-integration.

piet-wgsl/src/render.rs
DJMcNab: I copied this from the unstable `std` implementation, and this is not hot

I'd be inclined to leave it as it.

DJMcNab:
Stupid. I targeted main again

DJMcNab:
Fixed

PR #208, opened 2022-11-22
Update piet-scene to depend on peniko

dfrg:
This adds a new dependency on peniko, reintroduces kurbo for geometry (!), removes the now defunct types from piet-scene and updates all the test scenes to use the new types.

There is unfortunately a lot of churn here, but most of the changes should be relatively simple.

Depends on linebender/peniko#1


DJMcNab:
`piet-scene` is depended on by both the new and the old editions, to add to the context.

dfrg:
> `piet-scene` is depended on by both the new and the old editions, to add to the context.

Right, thanks. To expand on this a bit, piet-scene currently contains a set of types for defining rendering and composition state (brushes, blend modes, stroke/fill styles, etc) along with it's own geometry types (point, matrix). This lifts the former set of types to the peniko crate which now depends on kurbo for the latter set.

The goal of all this is to both reintroduce kurbo to piet-gpu and to provide a separate crate for these "styling" types that can be used across the ecosystem. At the very least, parley (text layout), the font enumeration library, and the oxidize glyph loader will make use of this.

dfrg:
This is now ready to go. linebender/xilem#1 will need changes after this lands. Also depends on linebender/glazier#33 to sync kurbo versions.

piet-gpu/Cargo.toml
```diff
@@ -28,7 +28,6 @@ path = "../piet-gpu-types"
 
 [dependencies.piet-scene]
 path = "../piet-scene"
-features = ["kurbo"]
```
raphlinus, line 31: 🚀

piet-scene/src/scene/builder.rs
```diff
@@ -14,17 +14,17 @@
...
-use core::borrow::Borrow;
+use peniko::kurbo::{Affine, PathEl, Shape};
+use peniko::{BlendMode, BrushRef, ColorStop, Fill, Stroke};
 use smallvec::SmallVec;
 
 /// Builder for constructing a scene or scene fragment.
 pub struct SceneBuilder<'a> {
     scene: &'a mut SceneData,
-    layers: SmallVec<[BlendMode; 8]>,
+    layers: SmallVec<[(BlendMode, bool); 8]>,
```
raphlinus, line 27: I personally would find it useful to have a comment what the bool means.

piet-scene/src/scene/builder.rs
```diff
@@ -134,30 +131,35 @@ impl<'a> SceneBuilder<'a> {
...
     }
 }
 
 impl<'a> SceneBuilder<'a> {
-    fn encode_path<E>(&mut self, elements: E, is_fill: bool) -> bool
-    where
-        E: Iterator,
-        E::Item: Borrow<PathElement>,
-    {
+    fn encode_path(&mut self, shape: &impl Shape, is_fill: bool) -> bool {
```
raphlinus, line 141: This is another place where I'd like a comment explaining the bool result.

piet-scene/src/scene/builder.rs
```diff
@@ -51,44 +51,47 @@ impl<'a> SceneBuilder<'a> {
...
         self.linewidth(-1.0);
-        let elements = elements.into_iter();
-        self.encode_path(elements, true);
-        self.begin_clip(Some(blend));
-        self.layers.push(blend);
+        if self.encode_path(shape, true) {
+            self.begin_clip(blend);
+            self.layers.push((blend, true));
+        } else {
+            // When the layer has an empty path, record an entry to prevent
```
raphlinus, line 67: I think I'm confused what the semantics should be for an empty path. Putting blends aside, clipping to an empty path should suppress rendering of the children altogether (not super useful, but it should do the right thing). Here it seems it renders the children without a clip.

For blends, the path given should enclose the content, so an empty path with nonempty content is a client bug. I'm less concerned about correct semantics in that case.

raphlinus (commented):
Mostly looks good, but I want to understand empty clip paths.

piet-scene/src/scene/builder.rs
dfrg: Will do.

piet-wgsl/shader/shared/blend.wgsl
```diff
@@ -45,11 +45,11 @@ fn color_burn(cb: f32, cs: f32) -> f32 {
 }
 
 fn hard_light(cb: vec3<f32>, cs: vec3<f32>) -> vec3<f32> {
-	return mix(
-		screen(cb, 2.0 * cs - 1.0),
-		cb * 2.0 * cs,
+    return mix(
```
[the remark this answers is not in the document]
dfrg: Makes sense. This is mostly a direct port of the GLSL where ?: doesn’t work like that.

piet-gpu/src/samples.rs
```diff
@@ -170,28 +167,28 @@ fn render_alpha_test(sb: &mut SceneBuilder) {
     sb.fill(
         Fill::NonZero,
         Affine::IDENTITY,
-        &Color::rgb8(255, 0, 0).into(),
+        Color::rgb8(255, 0, 0),
         None,
-        make_diamond(1024.0, 100.0),
+        &&make_diamond(1024.0, 100.0)[..],
```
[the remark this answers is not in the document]
dfrg: Oops. Missed those when I was updating.

piet-gpu/src/samples.rs
```diff
@@ -101,63 +101,60 @@ pub fn render_scene(sb: &mut SceneBuilder) {
...
     }
     sb.stroke(
-        &simple_stroke(2.0),
+        &Stroke::new(2.0),
         Affine::IDENTITY,
-        &Brush::Solid(Color::rgb8(0, 0, 0)),
+        Color::rgb8(0, 0, 0),
         None,
-        &path,
+        &&path[..],
```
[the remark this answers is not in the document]
dfrg: We can impl Shape for `Vec<PathEl>` in kurbo to fix this.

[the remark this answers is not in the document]
raphlinus: Hmmm. The way I've dealt with this in the path is use `BezPath` which is basically just a newtype for that, and I think I wanted to encourage people to use that rather than the raw vec. So I think the most idiomatic way to write this in this function is `path = BezPath::new()`.

I don't want to rathole on this too much, but I do find that sample code often has a disproportionate impact, as people often use it as a starting point.

raphlinus:
Very happy for this to go in once my review comments are addressed. I'll be around to do an approve quickly.

piet-gpu/src/samples.rs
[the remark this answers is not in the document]
dfrg: Fair enough! I believe it’s just a drop in replacement anyway.

piet-scene/src/scene/builder.rs
dfrg: A further thought on this: when reading through kurbo code, I noticed that it treats a line/quad/curve at the start of a path as a move. This code currently assumes an implicit move to (0,0) in that case. Should I update this to be consistent with kurbo?

raphlinus: I'm slightly confused what you're asking. In `BezPath` builder methods, there's a debug assert against that (I consider it an invalid state). I think a reason to have an implicit move of (0, 0) is so you don't get an invalidly encoded path as output even when the input is wrong, but the exact value of what you get is not too important, people shouldn't be relying on it.

But perhaps I'm just misunderstanding.

dfrg: It specifically noted for the Shape impl for slices here: https://github.com/linebender/kurbo/blob/fd839c25ea0c98576c7ce5789305822675a89938/src/bezpath.rs#L1177

The logic is implemented in the Segments iterator. FWIW, I think SVG does the same but I’d have to check the spec.

raphlinus: Ah yes, this behavior was changed in https://github.com/linebender/kurbo/pull/98. That particular use case doesn't seem too important here, but probably having fewer different behaviors is better.

dfrg:
This should be ready now.

For the invalid layer shape case, I decided to encode an empty path regardless of whether it's a clip or blend. This makes sense since blends are also clips in the valid case.

In the future, I'd also like to consider moving the blend mode syncing between BeginClip/EndClip to the clip_leaf shader where path_ix is copied over. This would remove the need for accounting on the encoding side.

raphlinus (approved):
Looks good, thanks!

I think moving the blend mode syncing into clip_leaf makes sense but haven't thought it through in detail.

PR #209, opened 2022-11-23
Remove the old pipeline

DJMcNab:
The only changed file contents should be removing the deleted items from the workspace, and updating the README slightly.


raphlinus:
We do need to update readme. I'm in favor of this but don't think we can merge until xilem moves its pin to wgpu.

dfrg:
I think in this case we actually want to target main :)

DJMcNab:
Not quite

We need to target `dev` then force push `dev` onto main

`main`'s history has all of the big pushes and pulls of the old shaders for every change to `dev`, which I don't think we care about.

It could be worthwhile rewriting history to only contain the contents of `piet-scene` and `piet-wgsl` (obviously whilst still keeping old `dev` and `main` around somewhere) to reduce cloning time; our history is full of `spirv` shaders from before the new branch setup for example.

I'm not sure if there's a good tool for visualising how necessary that is (e.g. "how much of our cloning time is each file type")

DJMcNab:
I've made two branches: https://github.com/linebender/piet-gpu/tree/custom-hal-archive and https://github.com/linebender/piet-gpu/tree/custom-hal-archive-with-shaders which contain the old versions, with and without the compiled shaders committed.

README.md
```diff
@@ -4,35 +4,30 @@ This repo contains the new prototype for a new compute-centric 2D GPU renderer.
...
-
-The hal layer in this repo is strongly inspired by gfx-hal, but with some differences. One is that we're shooting for a compile-time pipeline to generate GPU IR on DX12 and Metal, while gfx-hal ships [SPIRV-Cross] in the runtime. To access [Shader Model 6], that would also require bundling [DXC] at runtime, which is not yet implemented (though it's certainly possible).
-
-### Why not wgpu?
-
-The case for wgpu is also strong, but it's even less mature. I'd love to see it become a solid foundation, at which point I'd use it as the main integration with [Druid].
+## History
 
-In short, the goal is to facilitate the research now, collect the data, and then use that to choose a best path for shipping later.
+A prior incarnation used a custom cross-API hal. An archive of this version can be found in the branches [`custom-hal-archive-with-shaders`] and [`custom-hal-archive`].
```
raphlinus, line 30: I'm not sure we should keep branches around, it clutters things up. What I've taken to doing is creating a PR and then closing it. The branch is still accessible through the PR. That said, it does make us more dependent on GitHub specifically rather than Git.

raphlinus (approved):
I think we can do this now. Agree the README could use more work, but this is definitely an acceptable intermediate state.

README.md
DJMcNab: I've put some thoughts in Zulip; I think we can resolve this outside of this PR.

PR #210, opened 2022-11-25
Remove all wgsl-analyzer errors

DJMcNab:
The only remaning cause was a missing import from the settings

I have also applied wgsl-analyzer formatting to all shaders (apart from `draw_leaf`, where the formatting is a bit messy for the long `||` condition)


raphlinus:
Target `dev`, then this is good to go. I think there are probably more changes, and will experiment with that.

DJMcNab:
We really need to fix these ridiculous branches

raphlinus (approved):
Yes, moving to the wgsl port will be my priority, then we can nuke the branches. I underestimated how much friction it would cause.

PR #211, opened 2022-11-25
Clean up unneeded type annotations in vec

raphlinus:
Now that wgsl-analyzer 0.6 is released, most explicit type annotations on vec can be dropped (the exception being when it is a type conversion).

Also changes mix to select when the selector is actually boolean.

### DIFF
--- a/piet-wgsl/shader/shared/blend.wgsl
+++ b/piet-wgsl/shader/shared/blend.wgsl
@@ -45,23 +45,23 @@ fn color_burn(cb: f32, cs: f32) -> f32 {
 }
 
 fn hard_light(cb: vec3<f32>, cs: vec3<f32>) -> vec3<f32> {
-    return mix(
+    return select(
         screen(cb, 2.0 * cs - 1.0),
         cb * 2.0 * cs,
-        vec3<f32>(cs <= vec3<f32>(0.5))
+        cs <= vec3(0.5)
     );
 }
 
 fn soft_light(cb: vec3<f32>, cs: vec3<f32>) -> vec3<f32> {
-    let d = mix(
+    let d = select(
         sqrt(cb),
-        ((16.0 * cb - vec3(12.0)) * cb + vec3(4.0)) * cb,
-        vec3<f32>(cb <= vec3<f32>(0.25))
+        ((16.0 * cb - 12.0) * cb + 4.0) * cb,
+        cb <= vec3(0.25)
     );
-    return mix(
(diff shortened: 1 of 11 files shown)


piet-wgsl/shader/shared/blend.wgsl
```diff
@@ -45,23 +45,23 @@ fn color_burn(cb: f32, cs: f32) -> f32 {
...
-    let d = mix(
+    let d = select(
         sqrt(cb),
-        ((16.0 * cb - vec3(12.0)) * cb + vec3(4.0)) * cb,
-        vec3<f32>(cb <= vec3<f32>(0.25))
+        ((16.0 * cb - 12.0) * cb + 4.0) * cb,
+        cb <= vec3(0.25)
     );
-    return mix(
-        cb + (2.0 * cs - vec3(1.0)) * (d - cb),
```
DJMcNab, line 62: I don't like that `vec3<f32>-1.0` works

I guess it's fine here, but in general it's a bit suspect imo

piet-wgsl/shader/shared/blend.wgsl
```diff
@@ -45,23 +45,23 @@ fn color_burn(cb: f32, cs: f32) -> f32 {
...
-        cb - (vec3(1.0) - 2.0 * cs) * cb * (vec3(1.0) - cb),
-        vec3<f32>(cs <= vec3<f32>(0.5))
+    return select(
+        cb + (2.0 * cs - 1.0) * (d - cb),
+        cb - (1.0 - 2.0 * cs) * cb * (1.0 - cb),
+        (cs <= vec3(0.5))
```
DJMcNab, line 64: If this doesn't already have inlay hints, it probably needs them

It's not immediately clear if this is choosing the first argument or the second from (mix has a similar issue, fwiw)

DJMcNab (approved):
Yay!

piet-wgsl/shader/shared/blend.wgsl
raphlinus: It's super common in shader code, and WGSL made an explicit effort to include it, so I think it's worth going with the flow. I guess from a math perspective the multiply is fine but addition is wonky?

DJMcNab: Yeah, exactly

Both multiplies are a bit dodgy, but I can understand the ergonomics of the vec3 * vec3 component wise multiplication.

It's not sure important

PR #212, opened 2022-11-25
Make the WGSL port interactive

dfrg:
This adds the infrastructure to support writeable images and also hacks in a method for supplying external resources while running a recording. There are better designs for this, but I suspect that there will be a lot of refactoring around that code so I didn't want to invest a ton of effort here right now.

Changes fine to output to a texture instead of a buffer and also hacks in a small render pass to blit this texture to swapchain texture.

Finally, runs it all in a window with the original piet-gpu test scenes. There are some artifacts on wide strokes, but otherwise seems to be about right.

As usual with me, seems to be a lot of churn here :)

### DIFF
--- a/piet-wgsl/shader/fine.wgsl
+++ b/piet-wgsl/shader/fine.wgsl
@@ -21,10 +21,6 @@ var<storage> tiles: array<Tile>;
 @group(0) @binding(2)
 var<storage> segments: array<Segment>;
 
-// This will become a texture, but keeping things simple for now
-@group(0) @binding(3)
-var<storage, read_write> output: array<u32>;
-
 #ifdef full
 
 #import blend
@@ -32,6 +28,9 @@ var<storage, read_write> output: array<u32>;
 
 let GRADIENT_WIDTH = 512;
 let BLEND_STACK_SPLIT = 4u;
+
+@group(0) @binding(3)
+var output: texture_storage_2d<rgba8unorm, write>;
 
 @group(0) @binding(4)
 var<storage> ptcl: array<u32>;
@@ -77,6 +76,11 @@ fn read_rad_grad(cmd_ix: u32) -> CmdRadGrad {
     let roff = bitcast<f32>(ptcl[cmd_ix + 11u]);
     return CmdRadGrad(index, matrx, xlat, c1, ra, roff);
 }
+
+#else
+
+@group(0) @binding(3)
+var output: texture_storage_2d<r8, write>;
 
 #endif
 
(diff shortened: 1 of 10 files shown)


piet-wgsl/shader/fine.wgsl
```diff
@@ -273,20 +277,23 @@ fn main(
...
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
-        let fg = rgba[i];
-        let a_inv = 1.0 / (fg.a + 1e-6);
-        let rgba_sep = vec4(fg.r * a_inv, fg.g * a_inv, fg.b * a_inv, fg.a);
-        let bytes = pack4x8unorm(rgba_sep);
-        output[out_ix + i] = bytes;
-    }
+        let coords = xy_uint + vec2(i, 0u);
+        if coords.x < config.target_width && coords.y < config.target_height {
+            textureStore(output, vec2<i32>(coords), rgba[i]);
```
raphlinus, line 284: According to the WGSL spec, coords can be u32, but naga complains. This might be a naga issue.

piet-wgsl/src/simple_text.rs
```diff
@@ -0,0 +1,81 @@
+// Copyright 2022 The piet-gpu authors.
```
raphlinus, line 1: Is this here just because the demo hasn't been ported over to using parley? I can't imagine we'll want to maintain this long term.

raphlinus (approved):
Looks good. The line count had me a bit scared, but it's not too bad, a good chunk of it is just bringing other things in.

We want to lib'ize the repo pretty soon, but I'm fine with that being another PR.

raphlinus:
FWIW, I think my original idea for external resources was to have them the same as regular resources, but have a method to register them with the `Engine`. But I didn't work it out in any detail and don't have strong feelings about the way it's done; this seems Good Enough.

piet-wgsl/shader/fine.wgsl
dfrg: Yeah, I had tried u32 first, but it yelled at me. I'm sure it'll be fixed in time.

piet-wgsl/src/simple_text.rs
dfrg: Exactly. My goal was to bring the wgsl crate up to parity with piet-gpu so we can drop the old code.

I intend to submit another PR to peniko soon that will add a collection of font types (specifically the resource type itself, but also the styling properties) after which I'll update parley to depend on that and we'll have much better cohesiveness among these crates.

dfrg:
Agree 100% with lib'izing but I felt that was just additional churn for this PR. I'll put something together for that next.

As for external resources, your idea is much better and probably what we'll want to go with, but I threw this in as a stopgap to get it working. My assumption is that there are going to be a lot of changes to the structure as we figure out optimal resource allocation/reuse, handling the robust memory cases, and effectively intermixing CPU stages.

PR #213, opened 2022-11-26
Libify piet-wgsl

dfrg:
This creates a new Renderer type that offers fairly simple render_to_texture and render_to_surface methods, the latter of which handles the final blit internally. Also adds a util module with some helpers for device and surface creation.

There is a new winit example in piet-wgsl/examples/winit that shows how to make use of it all. This should be fairly trivial to adapt to glazier/xilem.


dfrg:
Also, should we rename piet-wgsl to piet-wgpu?

raphlinus:
https://github.com/lapce/piet-wgpu exists but hasn't been pushed to crates.io. It's a better name, but potentially *really* confusing, which is why I went with "wgsl". Another thing I'm considering is just calling it "piet-gpu".

dfrg:
I'm thinking toward the future when we have multiple backends, but I think just calling it piet-gpu and having the backends selected by features is probably the best idea. It would also mean that we never really have to expose the internal "recording" stuff as public API.

raphlinus (approved):
Looks good, thanks! As mentioned on Zulip, I wasn't a huge fan of `AnyScene` so am ok with going back to the old style, but if other people have a strong preference I'm happy to go with that.

raphlinus:
> It would also mean that we never really have to expose the internal "recording" stuff as public API.

Hmm, I think for Skia integration we're seriously considering doing just that. But it's a discussion we can have later after things have stabilized a bit.

PR #214, opened 2022-11-27
Fix rendering artifacts with strokes

raphlinus:
A number of things weren't being computed correctly for strokes, including bounding boxes and also incorrectly applying y_edge logic (which is only correct for fills).

### DIFF
--- a/piet-wgsl/shader/pathseg.wgsl
+++ b/piet-wgsl/shader/pathseg.wgsl
@@ -129,9 +129,8 @@ fn main(
     var tag_byte = (tag_word >> shift) & 0xffu;
 
     let out = &path_bboxes[tm.path_ix];
-    var linewidth: f32;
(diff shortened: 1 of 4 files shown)


piet-wgsl/shader/pathseg.wgsl
```diff
@@ -182,14 +181,16 @@ fn main(
                 p1 = mix(p1, p0, 1.0 / 3.0);
             }
         }
+        var stroke = vec2<f32>(0.0, 0.0);
```
dfrg, line 184: Probably don’t want to backtrack on these :)

dfrg (approved):
Looks good! Just one very minor nit. I think this now actually brings us to parity in terms of visual output.

piet-wgsl/shader/pathseg.wgsl
raphlinus: Heh, I originally had `vec2<f32>()` but decided to put in the actual numbers.

PR #215, opened 2022-11-28
Hodgepodge of small fixes

dfrg:
This just a collection of minor fixes to various things I noticed while looking over the code.

Adds a `Device::poll` call to the winit example.
Removes some lingering unnecessary `<f32>`s in the blit shaders.
Removes a duplicated bbox entry in the shared shader inclusion code and collapses nested `concat!`s.

Finally, it restores the computation of separated alpha in the output of fine which I removed in a previous PR. This allows the `render_to_texture` mode to be useful for generating images that can be properly blended over other content. The blit shader has been changed to accommodate this.

### DIFF
--- a/piet-wgsl/shader/fine.wgsl
+++ b/piet-wgsl/shader/fine.wgsl
@@ -282,8 +282,9 @@ fn main(
         let coords = xy_uint + vec2(i, 0u);
         if coords.x < config.target_width && coords.y < config.target_height {
             let fg = rgba[i];
-            let a_inv = 1.0 / (fg.a + 1e-6);
-            let rgba_sep = vec4(fg.r * a_inv, fg.g * a_inv, fg.b * a_inv, fg.a);            
+            // Max with a small epsilon to avoid NaNs
+            let a_inv = 1.0 / max(fg.a, 1e-6);
+            let rgba_sep = vec4(fg.rgb * a_inv, fg.a);            
             textureStore(output, vec2<i32>(coords), rgba_sep);
         }
     } 
--- a/piet-wgsl/shader/fine.wgsl
+++ b/piet-wgsl/shader/fine.wgsl
@@ -281,7 +281,10 @@ fn main(
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
         let coords = xy_uint + vec2(i, 0u);
         if coords.x < config.target_width && coords.y < config.target_height {
-            textureStore(output, vec2<i32>(coords), rgba[i]);
+            let fg = rgba[i];
+            let a_inv = 1.0 / (fg.a + 1e-6);
+            let rgba_sep = vec4(fg.r * a_inv, fg.g * a_inv, fg.b * a_inv, fg.a);            
+            textureStore(output, vec2<i32>(coords), rgba_sep);
         }
     } 
 #else
--- a/piet-wgsl/src/lib.rs
+++ b/piet-wgsl/src/lib.rs
@@ -205,7 +205,7 @@ impl BlitPipeline {
             @fragment
             fn fs_main(@builtin(position) pos: vec4<f32>) -> @location(0) vec4<f32> {
                 let rgba_sep = textureLoad(fine_output, vec2<i32>(pos.xy), 0);
-                return vec4(rgba_sep.rgb * rgba_sep.a, 1.0);
+                return vec4(rgba_sep.rgb * rgba_sep.a, rgba_sep.a);
             }
         "#;
 
--- a/piet-wgsl/src/lib.rs
+++ b/piet-wgsl/src/lib.rs
@@ -143,7 +143,6 @@ impl Renderer {
 }
 
 struct TargetTexture {
-    texture: Texture,
     view: TextureView,
     width: u32,
     height: u32,
@@ -166,7 +165,6 @@ impl TargetTexture {
         });
         let view = texture.create_view(&wgpu::TextureViewDescriptor::default());
         Self {
-            texture,
             view,
             width,
             height,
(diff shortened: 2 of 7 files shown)


piet-wgsl/shader/fine.wgsl
```diff
@@ -281,7 +281,10 @@ fn main(
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
         let coords = xy_uint + vec2(i, 0u);
         if coords.x < config.target_width && coords.y < config.target_height {
-            textureStore(output, vec2<i32>(coords), rgba[i]);
+            let fg = rgba[i];
+            let a_inv = 1.0 / (fg.a + 1e-6);
+            let rgba_sep = vec4(fg.r * a_inv, fg.g * a_inv, fg.b * a_inv, fg.a);
```
DJMcNab, line 286: This probably isn't a meaningful question, but why does an earlier step do the inverse of this multiplication?

It feels like it would be more meaningful to just avoid having that done too early into what is stored?

piet-wgsl/shader/fine.wgsl
```diff
@@ -281,7 +281,10 @@ fn main(
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
         let coords = xy_uint + vec2(i, 0u);
         if coords.x < config.target_width && coords.y < config.target_height {
-            textureStore(output, vec2<i32>(coords), rgba[i]);
+            let fg = rgba[i];
+            let a_inv = 1.0 / (fg.a + 1e-6);
```
DJMcNab, line 285: Is there any firm reasoning behind the use of specifically $10^{-6}$ here?
It probably needs a comment explaining the choice of magic number.

DJMcNab (approved):
Makes sense to me

piet-wgsl/shader/fine.wgsl
dfrg: Raph and I had a long conversation about this at some point and the upshot is that storing in premultiplied form is a few ALU ops shorter for basic source over and thus faster in the common case. We have to handle the conversion in blends on either input or output regardless, so this was chosen as the optimal strategy.

piet-wgsl/shader/fine.wgsl
dfrg: This exists simply to avoid a divide by zero without adding a branch. Happy to add a comment noting that, similar to the one in blend.wgsl.

dfrg:
Thanks Daniel. I’ll wait to merge pending Raph’s thoughts on supporting both premul and separate output from fine. I believe the only question is whether this is a flag in config with a dynamically uniform branch or another permutation of the shader.

piet-wgsl/shader/fine.wgsl
raphlinus: This is to avoid NaN. The idea is that it's a number small enough that it won't affect visual rendering, but not so small you might get overflow. 1e-6 is a pretty big "epsilon" as far as f32's go, but this code was originally ported from GLSL which had "mediump" and we do want to support f16 (yet another permutation, sigh).

One thought that occurs to me is that min() might be better than +, as it would leave pretty much all nonzero values untouched.

piet-wgsl/shader/fine.wgsl
raphlinus: It's not just one earlier step, it's an arbitrary amount of blending and compositing, which will be Porter-Duff "over" operations in the common case. It's both more efficient and more correct to do this operation premultiplied space. Best source for that is probably [Tom Forsyth's blog](https://tomforsyth1000.github.io/blog.wiki.html) (search for "premultiplied alpha, internal links seem to be broken). There are a few more links in my [2d.graphics page](https://2d.graphics/book/compositing.html), which I'd really like to flesh out some time.

As we've been discussing, it makes sense for this shader to produce either premultiplied or separated alpha depending on how its output will be used. For most use cases where it is consumed by the GPU, premultiplied alpha is better. For storing in a PNG, separated alpha is better. The *simplest* thing to do is to output separated here (basically at the cost of that divide instruction), then convert to premultiplied alpha downstream, and I'm ok with leaving it in that state.

The premultiplied alpha question also has a complex interaction with gamma. That's out of scope for now but is something we will want to address at some time.

piet-wgsl/src/lib.rs
```diff
@@ -185,28 +183,29 @@ impl BlitPipeline {
...
             }
             
             @group(0) @binding(0)
             var fine_output: texture_2d<f32>;
             
             @fragment
             fn fs_main(@builtin(position) pos: vec4<f32>) -> @location(0) vec4<f32> {
-                return textureLoad(fine_output, vec2<i32>(pos.xy), 0);
+                let rgba_sep = textureLoad(fine_output, vec2<i32>(pos.xy), 0);
+                return vec4(rgba_sep.rgb * rgba_sep.a, 1.0);
```
raphlinus, line 208: Shouldn't the last component be rgba_sep.a instead of 1.0? This is effectively compositing over opaque black, and we should be able to generate semi-transparent windows.

raphlinus (approved):
Looks good, thanks! One question in the fragment shader; this sorta depends on the presentation infrastructure being able to handle semi-transparent windows, which I think should be a goal but not necessarily a high priority.

piet-wgsl/shader/fine.wgsl
dfrg: I think you mean max() rather than min()? I'll make that change (in blend as well)

raphlinus: Yes max, duh.

piet-wgsl/src/lib.rs
dfrg: Yes, I hadn't considered semi-transparent windows. I'll change this to emit the alpha component.

raphlinus (approved):
Looks good, this addresses my concerns.

dfrg:
Thanks! Happy to address generating either output from fine in a future PR if that's something we want to do.

piet-wgsl/shader/fine.wgsl
DJMcNab: I don't think I made myself clear here; I knew why this addition was done; it's quite clear it's to avoid divide by zero.
My question was more about why $10^-6$ as opposed to say $10^-7$ or $10^-5$. It's fine if it's basically arbitrary (or in this case, other historical reasons?)

PR #216, opened 2022-11-28
Make the top level package be piet-wgsl

DJMcNab:
This also cleans up a number of remaining evidence of the previous version, such as the `.gitattributes` file, and much of the `.gitignore`.


DJMcNab:
I still think we should do this, but this kind of PR needs pre-approval and to land quickly, to minimise merge conflicts.

PR #217, opened 2022-11-28
Read gradient data from info buffer in fine

dfrg:
This changes coarse to write only index + info_offset into the ptcl for gradients. Per tile cost of gradients is reduced by 8 bytes and 36 bytes for linear and radial gradients respectively. We could save an additional 4 bytes per tile per gradient by copying the ramp index into the info buffer in draw_leaf but I'm not sure it's worth it.


raphlinus (approved):
Looks good, very much what I expected. Thanks!

Issue #202, opened 2022-11-09
Get tiger to run on Chrome Canary


raphlinus:
The simple path rendering demo runs on Web (#201) without too much trouble, and we should get that merged. An intermediate version that runs the tiger runs on wgpu native (current dev/main), but there are some problems running on Web. This issue lists those problems and outlines what needs to be done. There are two blockers: workgroup shared memory exceeded and bindings exceeded, plus a number of paper cuts. Some of these can be fixed in the piet-gpu codebase, perhaps others are better addressed elsewhere.

## Workgroup shared memory

The binning stage has 17408 bytes of workgroup shared memory, while the standard only mandates 16k. Most modern hardware can do 32k, so this is partly a case being more restrictive than need be (this is why it runs on native), but in this case the best mitigation is to optimize the shader, as trimming workgroup shared memory usage can improve occupancy. The best candidate is `sh_count`, which is currently u32[N_TILE =256][N_SLICE = 8]. These counts never exceed 256, so can be packed as u16x2, or, with a bit of careful logic, u8x4 (we only use the exclusive sum from shared memory, so the max value is 224.

## Number of buffer bindings

The coarse stage takes 10 bindings, and Chrome complains that the maximum is 8. The WebGPU spec only requires 8, but I consider this an unreasonable limitation, suitable only for low-spec hardware. I recommend the implementation raise it (which might require work in Chrome, but it's also possible this is being limited in the wgpu layer).

The reason for this is that the piet-wgpu reimplementation now uses a separate binding for each major data type. In some cases, that's mandated by the spec (an unsized array of atomics cannot share a binding with an unsized array of nonatomics, a limitation imposed by WGSL rather than the underlying APIs). In other cases, it seems like a good idea for performance reasons (input buffers can be bound readonly, which may have better cache behavior) and also code clarity.

One of the buffers (Config) can be moved from storage to uniform. That just requires a bit of additional plumbing in the engine implementation (I'm pretty sure uniforms don't count against the limit).

That leaves one extra binding. Basically, we need to do suballocation by hand. The most direct way to do this is to blit, which helps keeps things modular but at some cost (I consider this acceptable if we think the limits will be raised). To combine A and B, allocate the A buffer with size A+B, then before running coarse, blit B to after A. Good candidates include DrawMonoid + BinHeader, but Path and info are also plausible, as the sizes of these are all known at dispatch time.

## Paper cuts

[...]

PR #219, opened 2022-11-30
Reduce resource requirements of shaders

raphlinus:
Three changes were made to reduce the requirements of the shaders to fit in the "minimum maximum" parameters defined by WebGPU, apparently needed to get it to run in the browser.

* Config moved to uniform binding from storage_readonly
* workgroup memory requirements for binning reduced to < 16k
* info and bin_data buffers combined so coarse only needs 8 storage bindings

Progress to #202


piet-wgsl/src/render.rs
```diff
@@ -209,6 +210,7 @@ pub fn render_full(
     // TODO: calculate for real when we do rectangles
     let n_drawobj = n_path;
     let n_clip = data.n_clip;
+    let bin_data_start = n_drawobj * MAX_DRAWINFO_SIZE as u32;
```
dfrg, line 213: Now that I’ve internalized the drawtag encoding, it seems like we can compute this exact size for free in SceneBuilder. Consider this more of a note to self rather than a suggestion for this PR.

dfrg (approved):
Looks great! It will be exciting to see the full pipeline running on the web.

One small comment but no suggestions for changes.

PR #220, opened 2022-11-30
Full pipeline running in browser

raphlinus:
This is based on #201 and probably supersedes it. It's still hacky but does demonstrate the full pipeline running in a browser.

This is the command line I use:

```
RUSTFLAGS=--cfg=web_sys_unstable_apis cargo run --release --package run-wasm -- winit-demo
```

Then navigate Chrome Canary (with WebGPU flag enabled) to http://localhost:8000/

Fails on Windows because FXC, but runs on mac M1.

Significant progress toward #202


piet-wgsl/Cargo.toml
```diff
@@ -7,8 +7,144 @@ edition = "2021"
...
 [dependencies]
 wgpu = "0.14"
+wasm-bindgen = "0.2"
 raw-window-handle = "0.5"
 futures-intrusive = "0.5.0"
 parking_lot = "0.12"
 bytemuck = { version = "1.12.1", features = ["derive"] }
 piet-scene = { path = "../piet-scene" }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
DJMcNab, line 17: Why are all of these dependencies needed? I can't see where they're used

dfrg: Good catch. Looks like we don't need any of these on the main crate. They're only needed for the example.

run-wasm/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "run-wasm"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+
```
DJMcNab, line 6:
```suggestion
publish = false
```

We should also ensure this is consistent with the other example (i.e. changing the winit example to use `edition.workspace`)

piet-wgsl/shader/coarse.wgsl
```diff
@@ -235,8 +234,7 @@ fn main(
             let x1 = clamp(i32(path.bbox.z) - i32(bin_tile_x), 0, i32(N_TILE_X));
             let y1 = clamp(i32(path.bbox.w) - i32(bin_tile_y), 0, i32(N_TILE_Y));
             sh_tile_width[local_id.x] = u32(x1 - x0);
-            sh_tile_x0[local_id.x] = u32(x0);
-            sh_tile_y0[local_id.x] = u32(y0);
+            sh_tile_x0y0[local_id.x] = u32(x0) | u32(y0 << 16u);
```
DJMcNab, line 237: Should this `16` be a constant defined somewhere?

It's probably fine as-is.

piet-wgsl/shader/coarse.wgsl
```diff
@@ -46,7 +46,7 @@ var<storage, read_write> ptcl: array<u32>;
 // Much of this code assumes WG_SIZE == N_TILE. If these diverge, then
 // a fair amount of fixup is needed.
 let WG_SIZE = 256u;
-//let N_SLICE = WG_SIZE / 32u;
+//const N_SLICE = WG_SIZE / 32u;
```
DJMcNab, line 49:
```suggestion
//let N_SLICE = WG_SIZE / 32u;
```
we should be consistent I think?

piet-wgsl/shader/binning.wgsl
```diff
@@ -45,7 +45,7 @@ let SY = 0.00390625;
 
 let WG_SIZE = 256u;
 let N_SLICE = 8u;
-//let N_SLICE = WG_SIZE / 32u;
+//const N_SLICE = WG_SIZE / 32u;
```
DJMcNab, line 48:
```suggestion
//let N_SLICE = WG_SIZE / 32u;
```
and here

piet-wgsl/examples/winit/src/test_scene.rs
```diff
@@ -50,7 +50,7 @@ const N_CIRCLES: usize = 0;
 #[allow(unused)]
 pub fn render_svg(sb: &mut SceneBuilder, svg: &PicoSvg, print_stats: bool) {
     use crate::pico_svg::*;
-    let start = std::time::Instant::now();
+    //let start = std::time::Instant::now();
```
DJMcNab, line 53: I'd be tempted to just remove this line entirely.

piet-wgsl/examples/winit/src/main.rs
```diff
@@ -103,5 +104,28 @@ async fn run() -> Result<()> {
...
+        let event_loop = EventLoop::new();
+        let window = winit::window::Window::new(&event_loop).unwrap();
+
+        std::panic::set_hook(Box::new(console_error_panic_hook::hook));
+        console_log::init().expect("could not initialize logger");
+        use winit::platform::web::WindowExtWebSys;
+
+        // On wasm, append the canvas to the document body
+        let canvas = window.canvas();
+        canvas.set_width(1044);
```
DJMcNab, line 122: I see this size is pre-existing, so I don't think we need to change it, but `1044` is not a very round number

piet-wgsl/examples/winit/src/main.rs
```diff
@@ -19,25 +19,26 @@ mod simple_text;
...
     let window = WindowBuilder::new()
         .with_inner_size(LogicalSize::new(1044, 800))
         .with_resizable(true)
         .build(&event_loop)
         .unwrap();
-    let render_cx = RenderContext::new().await?;
+    run_wasm(event_loop, window).await
+}
+
+async fn run_wasm(event_loop: EventLoop<()>, window: Window) {
```
DJMcNab, line 36: This function name isn't quite right. I'd be tempted here to factor out the wasm specific code from `main` into the equivalent place in `run`, and then not create the second function. 

AFAIK, `resizable` just does nothing on web, rather than causing issues.

piet-wgsl/examples/winit/Cargo.toml
```diff
@@ -14,3 +14,138 @@ winit = "0.27.5"
...
 # for picosvg
 roxmltree = "0.13"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+console_error_panic_hook = "0.1.7"
+console_log = "0.2"
+wasm-bindgen-futures = "0.4.33"
+js-sys = "0.3.60"
+web-sys = { version = "0.3.60", features = [
+    "Document",
```
DJMcNab, line 24: Why do we need to specify all the GPU specific things? Does `wgpu` not enable those features for us?

piet-wgsl/examples/winit/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "winit"
+name = "winit-demo"
```
DJMcNab, line 2: Ha, I've been meaning to fix this. I presume you ran into the issue where it couldn't work out which `winit` you meant in `cargo run -p winit`?

DJMcNab (commented):
I can't approve as I haven't run it, but this looks good.

piet-wgsl/shader/coarse.wgsl
dfrg: Seems like a common enough pattern to not require a definition.

piet-wgsl/shader/coarse.wgsl
dfrg: Done

piet-wgsl/examples/winit/src/main.rs
dfrg: This was specifically not very round to test an output that didn't align to a tile boundary.

piet-wgsl/examples/winit/Cargo.toml
dfrg: Removed these. Works fine without them.

piet-wgsl/examples/winit/src/main.rs
dfrg: Refactored this a bit: `run_wasm` -> `run` and the native code is simply pulled into `main`

piet-wgsl/examples/winit/src/test_scene.rs
dfrg: Done

dfrg (approved):
Since this is a joint PR, I'll add an approval. Current state works here (both web and native) on M1.

raphlinus:
This works on Canary Chrome (but is still missing some strokes). I'm comfortable with it going in based on the review it's gotten, but can also take a closer look if people feel that's needed.

dfrg:
I feel pretty good about it as is. Feel free to merge

DJMcNab (commented):
I forgot to submit these comments. They're not too important though

PR #221, opened 2022-12-01
Don't encode anything for "empty" brushes

dfrg:
I was crashing on ramp generation when a gradient contained an empty color stop collection (bug in the lottie importer). This changes `SceneBuilder` to check for that case and prevent encoding anything at all. Does the same for solid brushes with alpha of 0 because might as well.

Also adds a new method for creating a surface with a specific presentation mode to `util::RenderContext`.

### DIFF
--- a/piet-scene/src/scene/builder.rs
+++ b/piet-scene/src/scene/builder.rs
@@ -19,7 +19,6 @@ use crate::ResourcePatch;
 use bytemuck::{Pod, Zeroable};
 use peniko::kurbo::{Affine, PathEl, Rect, Shape};
 use peniko::{BlendMode, BrushRef, ColorStop, Fill, Stroke};
-use smallvec::SmallVec;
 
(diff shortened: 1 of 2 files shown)


piet-scene/src/scene/builder.rs
```diff
@@ -273,6 +280,16 @@ impl<'a> SceneBuilder<'a> {
     }
 }
 
+/// Returns false if a brush would be guaranteed to render nothing.
+fn check_brush(brush: &BrushRef) -> bool {
```
HalfVoxel, line 284: I'd suggest `is_empty` or `renders_nothing` (together with inverting the return value) since that is less ambiguous than `check_brush`.

dfrg: Makes sense— I’ll give it a better name. Thanks!

dfrg:
Closing old PRs. Will address this in the future.

PR #223, opened 2022-12-05
Increase sizes of binning buffers

### LINKED ISSUE #222: Missing elements when drawing >256 elements
When encoding more then 256 elements in scene, elements start to disappear at the screen.
This seems to be caused by the `binning` step which allocates a too small `bin_header` buffer to fit all the data.

The [buffer size](https://github.com/linebender/piet-gpu/blob/main/piet-wgsl/src/render.rs#L338) assumes a packed bin layout, while in the shader the [access pattern](https://github.com/linebender/piet-gpu/blob/main/piet-wgsl/shader/binning.wgsl#L131) differs. One workgroup of 256 threads executes binning for 256 elements and each threads writes the output for one bin. Therefore the number of buffer size should be able to hold `drawobj_wgs` * 256 elements or the shader adjusted.

raphlinus:
Increase the size of the bin header and path buffers to accommodate the actual extents accessed by the shaders.

Also add a missing workgroup barrier, not essential to fixing this bug but found while auditing the code.

Fixes #222

### DIFF
--- a/piet-wgsl/shader/binning.wgsl
+++ b/piet-wgsl/shader/binning.wgsl
@@ -62,6 +62,7 @@ fn main(
     for (var i = 0u; i < N_SLICE; i += 1u) {
         atomicStore(&sh_bitmaps[i][local_id.x], 0u);
     }
+    workgroupBarrier();
 
     // Read inputs and determine coverage of bins
     let element_ix = global_id.x;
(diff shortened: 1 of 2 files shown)


piet-wgsl/shader/binning.wgsl
```diff
@@ -62,6 +62,7 @@ fn main(
     for (var i = 0u; i < N_SLICE; i += 1u) {
         atomicStore(&sh_bitmaps[i][local_id.x], 0u);
     }
+    workgroupBarrier();
```
DJMcNab, line 65: This is for the read on (new) line 110, right?

Just want to check my understanding of barriers to be honest.

raphlinus: It protects the `atomicOr` on what's now line 110. This bug existed in the GLSL version too, but for some reason we never triggered it. And it *really* makes me want a race-catching tool, as that is conceptually not difficult at all, and I've been bitten so many times now.

msiglreith (approved):

PR #225, opened 2022-12-09
Rename piet-wgsl to vello!

dfrg:
This renames `piet-wgsl` to `vello` and also merges the code from the now deleted `piet-scene` crate. There is a bit of churn, but structural and code changes have been kept to a minimum, mostly just patching up various imports and paths.

The `glyph` module carries over from `piet-scene` but I think "very soon" is a good time to drop that and expose a preliminary glyph run API in `SceneBuilder`.

We'll likely want to coordinate on how we're handling the repo (rename or fork+archive?) and then push an immediate patch to xilem to limit breakage.


README.md
```diff
@@ -1,4 +1,4 @@
-# piet-gpu
+# vello (formerly piet-gpu)
```
DJMcNab, line 1: I'd rather we had the reference to `piet-gpu` within the first paragraph, but not directly in the title.

If we do want it in the title, I think we should set an end-date on it being the title; say 2023-03-09 as that is when Rust 1.68 is released; it also happens to be exactly 3 calendar months from now.

vello/src/scene.rs
```diff
@@ -469,3 +618,30 @@ impl<'a> PathBuilder<'a> {
         }
     }
 }
+
+fn affine_to_f32(affine: &Affine) -> [f32; 6] {
+    let c = affine.as_coeffs();
+    [
+        c[0] as f32,
```
DJMcNab, line 625: This code can use [`[T;N]::map()`](https://doc.rust-lang.org/std/primitive.array.html#method.map). That's not important for this PR

DJMcNab (approved):
🚀

vello/src/scene.rs
dfrg: oh, const generics ftw!

PR #226, opened 2022-12-10
Add an experimental bevy example

DJMcNab:
This is not an idiomatic combination of the two because we just ignore bevy's render graph abstraction (in particular, we don't have any way to support rendering something in bevy, then putting that in a texture, using that in vello, then rendering that texture in a different bevy camera at the moment).

This was however remarkably painless.

https://user-images.githubusercontent.com/36049421/206856331-a082197b-1773-4d43-841c-f88fc5fc1b05.mp4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,22 @@
 version = 3
 
 [[package]]
+name = "ab_glyph"
+version = "0.2.18"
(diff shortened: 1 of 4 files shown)


raphlinus:
A little bit of merge conflict fixing is needed because of the rename, but then I'd be happy to approve.

raphlinus (approved):

Issue #224, opened 2022-12-08
wgpu Adapter and Surface might not be compatible


rosefromthedead:
Since `RequestAdapterOptions` field `compatible_surface` is not specified ([see here](https://github.com/linebender/piet-gpu/blob/a7dbeb3b9be0573c4aae2b2703a33d391178ae44/piet-wgsl/src/util.rs#L34)), wgpu may not give an adapter that can render to the display. But the current API means the Instance and Adapter are constructed at the same time and the Surface is constructed later, so without changing the API (which admittedly seems pretty natural how it is), it's not possible to fix this. Maybe it's worth separating Instance and Adapter creation? Or making the user provide Instance and Surface themselves?

FWIW I ran into this incompatibility on my intel+nvidia laptop, but the default `PowerPreference` should select intel, so I might just not have *any* compatible combinations. Still might be worth allowing this for other machines.

dfrg:
Thanks for reporting this. The intent behind separating surface and device creation is to support multiple windows (which may come and go, particularly in UI contexts) that can share resources on a single device. If the current design is failing on dual GPU setups, it’s definitely something that we need to address even if that makes the API less pleasant. 

I am curious about the error message. Is it general surface incompatibility or is the format (we unconditionally choose Bgra8Unorm) incompatible?

rosefromthedead:
The error message is `Error in Surface::configure: surface does not support the adapter's queue family`, so it seems not to be pixel format, but idk how to check.

rosefromthedead:
IMO it would be good to have error handling at surface creation time, to request an adapter that is compatible with the new surface if the current adapter is not, and then continue using both adapters for their respective surfaces. This means that if I have 2 cards with 1 display each then my app can draw to surfaces on both displays simultaneously without incompatibility issues, which might not be possible if vello only keeps one adapter around. Helpfully, it also solves the original problem :P 

Here's hoping it's not too hard to add multi-adapter support?

dhardy:
Just a little note, I had a [similar report recently](https://github.com/kas-gui/kas/issues/367#issuecomment-1357956379), and while at first glance use of a `compatible_surface` to ensure compatibility may appear a good idea, my conclusion was that this only really does anything for users with multiple GPUs (in the reporter's case, Intel iGPU + NVidia discrete graphics). The better solution, for a general purpose GUI, is to make the surface compatible with as many adapters as possible (including old Intel iGPUs).

rosefromthedead:
[...]

PR #227, opened 2022-12-14
Multi-adapter support

rosefromthedead:
Now Adapter, Device, Queue are bundled into DeviceHandle and created alongside Surface, or reused if we already created a compatible handle.

I'm open to design complaints, just wanted to prove that the logic would solve my problem from #224, and it does :). Now on x11 (through xwayland) the winit example works fine, and on wayland I get a different error message than before 🥳

All of this comes at the cost of a little extra user-side juggling, but in the winit example it's not much extra effort.

### DIFF
--- a/examples/with_winit/src/main.rs
+++ b/examples/with_winit/src/main.rs
@@ -23,10 +23,13 @@ use winit::{event_loop::EventLoop, window::Window};
 
 async fn run(event_loop: EventLoop<()>, window: Window) {
     use winit::{event::*, event_loop::ControlFlow};
-    let render_cx = RenderContext::new().await.unwrap();
+    let mut render_cx = RenderContext::new().unwrap();
     let size = window.inner_size();
-    let mut surface = render_cx.create_surface(&window, size.width, size.height);
-    let mut renderer = Renderer::new(&render_cx.device).unwrap();
+    let mut surface = render_cx
+        .create_surface(&window, size.width, size.height)
+        .await;
+    let device_handle = &render_cx.devices[surface.dev_id];
+    let mut renderer = Renderer::new(&device_handle.device).unwrap();
     let mut simple_text = simple_text::SimpleText::new();
     let mut current_frame = 0usize;
     let mut scene_ix = 0usize;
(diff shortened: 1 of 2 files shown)


lambda:
Would just like to note that this branch fixes the issue for me as well; I am running on X11 (since I still can't screenshare using Zoom on Wayland), and have both integrated and discrete graphics. [Some more detail on Zulip](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Xilem.2FVello.20Examples.20fail.20to.20run/near/318924499).

rosefromthedead:
Re: [the discussion on Zulip](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Xilem.2FVello.20Examples.20fail.20to.20run/near/319011610) - for what it's worth, I think this is complexity that has to happen somewhere. Picking an adapter with no criteria won't work, as we've seen for nvidia laptops. I tried to balance vello complexity and API consumer complexity, by exposing the fact that multiple adapters are in use but leaving the selection and reuse logic on our side. That logic could be left to the user, or the whole business could be hidden away, but whatever happens, there's gonna be more work somewhere.

dfrg:
To add some context, I agree that we’ll need extra complexity somewhere. The open question for me is whether it’s actually necessary to maintain a collection of devices or just choose the appropriate one. In the former case, we’ll also need to duplicate the Renderer state and all associated resources per device. I also wonder what happens if you drag a window to another screen that may be attached to a different device. For the latter case, I’d rather add the logic to select a single device that supports the surface.

dfrg:
I do appreciate the fix and this does address the immediate problem allowing people with dual GPUs to actually run the code so if you fix the conflicts I’ll be happy to approve this and get it merged. We can revisit the full implications later.

rosefromthedead:
That's weird, the merge conflicts weren't showing up locally, so I just rebased and hoped for the best.

I'd always imagined that dragging windows across screens should force apps to make a new surface for each new screen the window is on; maybe because I've been reading too many discussions about fractional scaling. In the absence of this, I see that things could get messy with my code.

Good point about duplication. Looking at wgpu again, I think it's possible to minimise the number of devices in use, but it would mean that a surface needs to switch device if a new surface is created. Room for improvement, I guess

dfrg (approved):
Looks good, thanks!

PR #231, opened 2023-01-04
Move the vello crate to the workspace root

DJMcNab:
Move the vello crate to the root, as discussed in this week's office hours.

Available [to browse](https://github.com/DJMcNab/vello/tree/vello_to_root).

I've chosen to move `run_wasm` to `examples/`.

I've also renamed the examples to both match their folder names, and have a consistent naming scheme (`with_winit` and `with_bevy`). Note that this can be changed to taste; I don't know what the standard thing in the ecosystem is. Note that we can't call them `bevy` or `winit` directly, because the `bevy` and `winit` both already exist in the dependency.

I've also added a note to the readme that it's work in progress; I've merged in `vello/README.md` by sticking it at the end, so that writing isn't lost for when we do the rewrite.

I'm tempted to also move `UNLICENSE` to 'shaders/UNLICENSE', it's a bit misleading having it at the root.


raphlinus:
Overall looks good. One thing I can't quite figure out is that `cargo run --bin with_winit` doesn't seem to work from the root, while the corresponding command does seem to work in main. Is this a limitation or just something I'm missing?

DJMcNab:
That would be due to cargo's package selection. When you run on this branch, by default the `vello` package will be used, and since there is no binary crate with that name in this package, it fails. On `main`, it searches the entire workspace for binary crates, because there isn't a prefered package (we could have also had a preferred package if we set the `workspace.default-members` key).

I'd class it a cargo bug that it doesn't inform you how to correct your case; the correct command would be `cargo run --package with_winit --bin with_winit` (or `cargo run -p with_winit` for short, since the package only has one binary crate).

raphlinus (approved):
Yep, let's go ahead and do this. Perhaps feedback to cargo about not finding the example might be useful, but I'm not too bothered by the change.

Thanks!

PR #234, opened 2023-01-05
Improve the gitattributes and gitignore files

DJMcNab:
As fallout from #231, our language stats are all wrong. Try to fix that.

This was already done in #216, but that was superceded after the rename.


raphlinus (approved):
Agree, good cleanup. Thanks!

PR #235, opened 2023-01-05
Support for larger pathtags

raphlinus:
Previously there was a limit of 256k pathtags in a scene, due to the need for multi-dispatch prefix sum for the pathtag monoid. This patch increases the limit to 64M, which ought to be enough for most applications.

It works by having 4 dispatches for the pathtag prefix sum: 2 to reduce, then 2 to scan.

### DIFF
--- a/shader/pathtag_reduce2.wgsl
+++ b/shader/pathtag_reduce2.wgsl
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT OR Unlicense
+
+// This shader is the second stage of reduction for the pathtag
+// monoid scan, needed when the number of tags is large.
+
+#import config
+#import pathtag
+
+@group(0) @binding(0)
+var<storage> reduced_in: array<TagMonoid>;
+
+@group(0) @binding(1)
+var<storage, read_write> reduced: array<TagMonoid>;
+
+let LG_WG_SIZE = 8u;
+let WG_SIZE = 256u;
+
+var<workgroup> sh_scratch: array<TagMonoid, WG_SIZE>;
+
+@compute @workgroup_size(256)
+fn main(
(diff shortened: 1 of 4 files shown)


raphlinus:
Some things to discuss.

First, I ended up doing a lot of copy-paste on the shader code. It might be better to either use `#ifdef` more extensively or put some of the common bits of logic in shared.

Second, I find myself wanting to reconsider the dispatch logic for the pathtag scan. In the "small" case (<= 256k path tags) it does two dispatches, but the second dispatch does the scan of the reduced monoids in a pretty brute-force way (each dispatch does its own reduction). In the large case, the last scan does less work, as the next-to-last scan prepares the prefix. I suspect the large case might actually be faster. Obviously that's something that can should be measured and tuned. I'm reasonably satisfied with the logic in this PR though, as it's a fairly modest diff on the existing state.

With this change, it runs paris-30k with a couple caveats. One is that the hardcoded buffer sizes in render.rs need to be increased. The proper fix for this is robust dynamic memory. Second, the current logic in test_scene loads the XML file each frame, which is very slow. That can be improved fairly straightforwardly though.

lemmih:
`pathtag_reduce2` is numbered 2 because it runs after `pathtag_reduce`? And `pathtag_scan1` is numbered 1 because it runs before `pathtag_scan`?

raphlinus:
I didn't make this explicit (it could be a comment), but a systematic way to think about the number is the size of the slice that it represents. The output of reduce2 is the reduction of wg^2 elements, while the output of scan1 is the scan over reductions by wg^1. The natural or default value for reduce is 1, and for scan is 0.

dfrg (approved):
This has had sufficient testing so I’m going to +1 so it can be merged when ready.

PR #239, opened 2023-01-06
Public encoding API

dfrg:
This exposes significantly more of the previously internal encoding guts.

This is beneficial for its own sake as there always seem to be new and interesting things that are supported by the encoding and GPU pipeline that are not directly exposed by the scene builder.

The secondary purpose is laying the groundwork for scene *decoding* and processing. This will enable us to implement various stages of the pipeline on the CPU and hopefully allow development of better debugging tools.

### DIFF
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -138,7 +138,7 @@ fn fill_path(tile: Tile, xy: vec2<f32>) -> array<f32, PIXELS_PER_THREAD> {
     }
     // nonzero winding rule
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
-        area[i] = abs(area[i]);
+        area[i] = min(abs(area[i]), 1.0);
     }
     return area;
 }
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -138,7 +138,7 @@ fn fill_path(tile: Tile, xy: vec2<f32>) -> array<f32, PIXELS_PER_THREAD> {
     }
     // nonzero winding rule
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
-        area[i] = min(abs(area[i]), 1.0);
+        area[i] = abs(area[i]);
     }
     return area;
 }
--- a/src/encoding/draw.rs
+++ b/src/encoding/draw.rs
@@ -1,0 +1,164 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Also licensed under MIT license, at your choice.
+
+use bytemuck::{Pod, Zeroable};
+use peniko::{BlendMode, Color};
+
+use super::Monoid;
+
+/// Draw tag representation.
+#[derive(Copy, Clone, PartialEq, Eq, Pod, Zeroable)]
+#[repr(C)]
+pub struct DrawTag(pub u32);
+
+impl DrawTag {
+    /// No operation.
+    pub const NOP: Self = Self(0);
+
+    /// Color fill.
+    pub const COLOR: Self = Self(0x44);
+
+    /// Linear gradient fill.
+    pub const LINEAR_GRADIENT: Self = Self(0x114);
+
+    /// Radial gradient fill.
+    pub const RADIAL_GRADIENT: Self = Self(0x2dc);
+
+    /// Image fill.
+    pub const IMAGE: Self = Self(0x48);
+
+    /// Begin layer/clip.
+    pub const BEGIN_CLIP: Self = Self(0x9);
+
+    /// End layer/clip.
+    pub const END_CLIP: Self = Self(0x21);
+}
+
+impl DrawTag {
+    /// Returns the size of the info buffer used by this tag.
+    pub fn info_size(self) -> u32 {
+        (self.0 >> 6) & 0xf
+    }
+}
+
+/// Draw data for a solid color.
+#[derive(Clone, Copy, Debug, Default, Zeroable, Pod)]
+#[repr(C)]
+pub struct DrawColor {
+    /// Packed RGBA color.
+    pub rgba: u32,
+}
+
+impl DrawColor {
+    /// Creates new solid color draw data.
+    pub fn new(color: Color) -> Self {
+        Self {
+            rgba: color.to_premul_u32(),
+        }
+    }
+}
+
+/// Draw data for a linear gradient.
+#[derive(Clone, Copy, Debug, Default, Zeroable, Pod)]
+#[repr(C)]
+pub struct DrawLinearGradient {
+    /// Ramp index.
+    pub index: u32,
+    /// Start point.
+    pub p0: [f32; 2],
+    /// End point.
+    pub p1: [f32; 2],
+}
+
+/// Draw data for a radial gradient.
+#[derive(Clone, Copy, Debug, Default, Zeroable, Pod)]
+#[repr(C)]
+pub struct DrawRadialGradient {
+    /// Ramp index.
+    pub index: u32,
+    /// Start point.
+    pub p0: [f32; 2],
+    /// End point.
+    pub p1: [f32; 2],
+    /// Start radius.
+    pub r0: f32,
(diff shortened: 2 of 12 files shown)


src/encoding/draw.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Begin layer/clip.
+    pub const BEGIN_CLIP: Self = Self(0x9);
+
+    /// End layer/clip.
+    pub const END_CLIP: Self = Self(0x21);
+}
+
+impl DrawTag {
+    /// Returns the size of the info buffer used by this tag.
+    pub fn info_size(self) -> u32 {
```
raphlinus, line 52: This might be a candidate for being marked const fn?

src/encoding/draw.rs
```diff
@@ -0,0 +1,164 @@
...
+        (self.0 >> 6) & 0xf
+    }
+}
+
+/// Draw data for a solid color.
+#[derive(Clone, Copy, Debug, Default, Zeroable, Pod)]
+#[repr(C)]
+pub struct DrawColor {
+    /// Packed RGBA color.
```
raphlinus, line 61: This would probably be a good place to document the byte order.

src/encoding/encoding.rs
```diff
@@ -0,0 +1,235 @@
...
+
+impl Encoding {
+    /// Creates a new encoding.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns true if the encoding is empty.
+    pub fn is_empty(&self) -> bool {
+        self.path_tags.is_empty()
```
raphlinus, line 57: Partly a note to self, this will change when there are rectangles.

src/encoding/encoding.rs
```diff
@@ -0,0 +1,235 @@
...
+            self.color_stops.extend_from_slice(stops);
+        }
+        self.patches.push(Patch::Ramp {
+            offset,
+            stops: stops_start..self.color_stops.len(),
+        });
+    }
+}
+
+fn color_with_alpha(mut color: Color, alpha: f32) -> Color {
```
raphlinus, line 232: This feels like it should be a method on Color?

src/encoding/monoid.rs
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Also licensed under MIT license, at your choice.
+
+/// Interface for a monoid.
```
raphlinus, line 17: Probably good to mention that default is the identity of the monoid.

I'm also curious why T is a generic parameter rather than an associated type.

The existence of the `new` method makes this a monoid homomorphism rather than a vanilla monoid, but for the purposes of this module we don't have to go full math nerd.

src/encoding/path.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /// Returns true if this is a 32-bit floating point segment.
+    pub fn is_f32(self) -> bool {
+        self.0 & Self::F32_BIT != 0
+    }
+
+    /// Returns true if this segment ends a subpath.
+    pub fn is_subpath_end(self) -> bool {
+        (self.0 & Self::SUBPATH_END_BIT) != 0
```
raphlinus, line 97: Parentheses are inconsistent with previous fn.

raphlinus (approved):
I like it! It seems clean, and in particular I like the way the tag values are encapsulated, it's a big improvement over having magic numbers scattered in the source. It will be very interesting to see how people use this.

Most of the comments are minor documentation nits.

shader/fine.wgsl
```diff
@@ -138,7 +138,7 @@ fn fill_path(tile: Tile, xy: vec2<f32>) -> array<f32, PIXELS_PER_THREAD> {
     }
     // nonzero winding rule
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
-        area[i] = abs(area[i]);
+        area[i] = min(abs(area[i]), 1.0);
```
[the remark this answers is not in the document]
dfrg: Sounds good.

src/encoding/draw.rs
dfrg: Will do.

src/encoding/draw.rs
dfrg: Ah, good point since it’s technically ABGR and also premultiplied.

src/encoding/encoding.rs
dfrg: I think this was originally a hack to detect empty glyphs when that would break the path encoding but I think it’s probably useful regardless.

src/encoding/encoding.rs
dfrg: Yes, I’ll update add it to my pending peniko patch. For context, lottie has an additional opacity modifier and I wanted to be able to encode this without rebuilding brushes.

src/encoding/monoid.rs
dfrg: Associated type seems reasonable to me. Also, always happy to be educated in category theory :)

src/encoding/path.rs
dfrg: Will make these consistent.

src/encoding/draw.rs
```diff
@@ -0,0 +1,164 @@
...
+
+use bytemuck::{Pod, Zeroable};
+use peniko::{BlendMode, Color};
+
+use super::Monoid;
+
+/// Draw tag representation.
+#[derive(Copy, Clone, PartialEq, Eq, Pod, Zeroable)]
+#[repr(C)]
+pub struct DrawTag(pub u32);
```
DJMcNab, line 25: Is it meaningful for this to be public (rather than having a getter?)

That is, is there any valid case where this could be set to a value other than one of the below constants?

I suppose it being `Pod` is equivalent to this being `pub`?

src/encoding/draw.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl DrawTag {
+    /// No operation.
+    pub const NOP: Self = Self(0);
+
+    /// Color fill.
+    pub const COLOR: Self = Self(0x44);
```
DJMcNab, line 32: The values seem very arbitrary. I see that 4 bits of the value is `info_size`. 6 is an awkward number of bits for that to be down, if these are being written out as (non-binary) literals.

src/encoding/draw.rs
dfrg: With the full encoding exposed, any number of invariants can be broken so I see little value in hiding some arbitrary bits of data behind getters. For those wanting a “safe” interface, SceneBuilder should be used.

src/encoding/draw.rs
dfrg: Raph can give a more informed rationale for the specific packing but various bits are used on the GPU side for computing offsets and more efficient branching on clips.

src/encoding/path.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+/// Path tag representation.
+#[derive(Copy, Clone, PartialEq, Eq, Pod, Zeroable)]
+#[repr(C)]
+pub struct PathTag(pub u8);
+
+impl PathTag {
+    /// 32-bit floating point line segment.
+    pub const LINE_TO_F32: Self = Self(0x9);
```
lemmih, line 55: As an outsider, a comment explaining that `0x9 = 0b1001 = Self::F32_BIT | PathSegmentType::LINE_TO` would help me understand these seemingly magic numbers.

src/encoding/draw.rs
raphlinus: This is actually a good point. Right now it's tightly packed rightward, but there's probably no compelling reason for that to be the case. If all our shifts were multiples of 4 then it would read as a hex value. One minor consideration is whether to count bytes or u32 words; the latter is more 

Another approach to this would be to have a const fn builder that takes is_clip, scene_size (maybe "scene_words"), and info_size as args. That feels Rust-like and less magical, though unfortunately the current state of naga is that it might be clearer to use hex values as those would match CPU and GPU sides.

dfrg: I had considered the const fn approach as well but felt it better if these values just matched those in the shader for now. I expect these to change again when we actually handle extend modes for gradients (scene size will change) so it may be worth reconsidering the packing at 4-bit granularity then.

src/encoding/path.rs
dfrg: Good point. I'll add a comment.

src/encoding/draw.rs
raphlinus: FWIW, the encoding will also change with rectangles.

dfrg:
This should address all of the review feedback. I'll fix `color_with_alpha` in another PR after the required method is added to peniko. Also adds some additional low level encoding methods for the three supported brush types as I think it could be useful to encode those as well without constructing full brushes.

PR #241, opened 2023-01-08
Fixes to fine rasterization

raphlinus:
The area bug was found and fixed by @dfrg, and is adapted from #239. I wanted to move it to a separate PR so that one would be more focused on API.

The other bug is currently silent because the two quantities swapped are both 4, but it is triggered when experimenting with tuning for performance.

### DIFF
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -138,7 +138,7 @@ fn fill_path(tile: Tile, xy: vec2<f32>) -> array<f32, PIXELS_PER_THREAD> {
     }
     // nonzero winding rule
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
-        area[i] = abs(area[i]);
+        area[i] = min(abs(area[i]), 1.0);
     }
     return area;
 }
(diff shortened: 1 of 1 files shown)


dfrg (approved):
Looks good!

shader/fine.wgsl
```diff
@@ -138,7 +138,7 @@ fn fill_path(tile: Tile, xy: vec2<f32>) -> array<f32, PIXELS_PER_THREAD> {
     }
     // nonzero winding rule
     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
-        area[i] = abs(area[i]);
+        area[i] = min(abs(area[i]), 1.0);
```
DJMcNab, line 141: Without comments in the vicinity, I can't easily work out the reason for this change.

I will note that this is a significant change if the value is highly negative; whether that makes sense is a different question.

This does at least agree with the `nonzero` comment

dfrg: This ensures consistent coverage between overlapping and non-overlapping regions in self-intersecting paths. When `alpha < 1` this results in increased brightness in overlapping pixels. Note that this clamp was in the original GLSL code so it’s not technically a change.

raphlinus: It's basically the definition of the nonzero fill rule: 0 maps to 0, any other value maps to 1, and this particular formulation gets fractional values for antialiasing basically correct.

DJMcNab: Oh, my mistake. I didn't see that `abs` wasn't removed

PR #242, opened 2023-01-10
Support even-odd fill rule

dfrg:
Add logic for the even-odd fill rule to `SceneBuilder` and the coarse and fine shaders.

There is some new unfortunate branching in coarse around `write_path` to handle the case where entire solid tiles are omitted. It might be possible to include this with the other checks for `include_tile` but I'm not sure if that is a win.


raphlinus (approved):
Right, we have to handle even-backdrop tiles with no path segments. It is possible we could refine the branching, but this seems good enough.

PR #243, opened 2023-01-11
Change SceneBuilder to be an owning and cloneable structure that is consumed to produce fragments

IntrepidPig:
This PR implements an API change that allows partially built scenes to be reused by making the `SceneBuilder` an owning and cloneable data structure.

It also makes it possible to store a partially constructed scene in a struct which I think would be impossible with the current API because storing both a `SceneFragment` and a `SceneBuilder<'a>` in a struct makes it self-referential.

The way this allows scene construction work to be reused is demonstrated in the contrived example below:
```
let mut builder = FragmentBuilder::new();
for i in 0..10 {
    builder.line_to(i * a, i * b);
    let fragment = builder.clone().finish();
    scene.append(&fragment);
    renderer.render(&scene);
}
```
In fact, this could be simplified even further by removing `SceneFragments` and allowing a `&SceneBuilder` to be appended directly to a `Scene`. I think that would most accurately convey the performance characteristics of the operation in the API as well.

I would be unsurprised to learn that the API is designed as it is right now with some future changes in mind that make what I suggest in this PR unfeasible. I'm posting this just as a discussion regarding how some scene-in-progress data could be reused out of efficiency, and to bring up the self-referential restrictions I see the current API imposing.

### DIFF
--- a/src/encoding/encoding.rs
+++ b/src/encoding/encoding.rs
@@ -22,7 +22,7 @@ use super::{
 use peniko::{kurbo::Shape, BlendMode, BrushRef, Color, ColorStop, Extend};
 
 /// Encoded data streams for a scene.
-#[derive(Default)]
+#[derive(Default, Clone)]
 pub struct Encoding {
     /// The path tag stream.
     pub path_tags: Vec<PathTag>,
--- a/src/glyph.rs
+++ b/src/glyph.rs
@@ -18,7 +18,7 @@
 
 pub use moscato::pinot;
 
-use crate::scene::{SceneBuilder, SceneFragment};
+use crate::scene::{FragmentBuilder, SceneFragment};
 use peniko::kurbo::{Affine, Rect};
 use peniko::{Brush, Color, Fill, Mix};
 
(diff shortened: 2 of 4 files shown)


dfrg:
Apologies for allowing this to fall by the wayside.

As you've noticed, the current design lacks some clarity of intent although this is historical rather than forward thinking. Originally, `Scene` and `SceneFragment` were distinct in that the latter contained resource "patches" that were resolved when appended to the former. Resource resolution, which involves deduplication, is naturally a sync point so this was moved to render time. `SceneBuilder` also contained state in the form of transform and clip stacks which have since been made obsolete due to changes in the encoding. 

I'm not very enthusiastic about changing `SceneBuilder` to be an owning type. That said, I do think that the ability to store and append to partially constructed scenes is worthwhile. Given the current state, it's worth considering whether `SceneBuilder` is necessary at all. It would be possible to move all methods to `Scene` allowing for incremental construction. We could also get rid of `SceneFragment` entirely although that would require some subtle changes when packing the scene for the GPU (the one remaining distinction is that `Scene` has an initial identity transform and -1.0 line width).

In its current form, I don't think this PR is the direction we want to go, but the problems it raises are definitely worth addressing so I appreciate the effort. If you'd like, feel free to file an issue about incremental scene construction and we can discuss API design and gather feedback before implementation.

dfrg:
After a very long delay, #432 has landed which should hopefully address the issues raised by this PR.

PR #244, opened 2023-01-11
Support click and drag for svgs, as well as scene fragment caching

DJMcNab:
Reopen of #236 after it was accidentally closed by GitHub UX.

The new paris-30k example, in `with_winit`, now caches the `SceneFragment`, and allows dragging and scaling the svg (click and drag, and mouse wheel).

The scaling is centred on the top left, which isn't ideal, but it works for demonstration. Doing the correct maths centered on the cursor should be reasonably easy, but it takes some thinking

I also added support for closing the window with `Escape`, whilst I was editing it.

Note that to try this yourself, you need to source the paris-30k svg from somewhere, and make the other changes discussed in #235, namely to the buffer sizes. 

The intention is to make a CLI flag to select the svg, but this is not yet implemented.

For reference, I found this diff sufficiently increases the buffer sizes for things to render, although I do still get random seeming artifacts (not consistent frame-to-frame). These numbers were chosen as arbitrary increases on the prior values, which it turns out are sufficient.

```diff
diff --git a/src/render.rs b/src/render.rs
index eee3d54..8e9c7c5 100644
--- a/src/render.rs
+++ b/src/render.rs
@@ -308,7 +308,7 @@ pub fn render_full(
         [config_buf, scene_buf, draw_reduced_buf],
     );
     let draw_monoid_buf = ResourceProxy::new_buf(n_drawobj as u64 * DRAWMONOID_SIZE);
-    let info_bin_data_buf = ResourceProxy::new_buf(1 << 20);
+    let info_bin_data_buf = ResourceProxy::new_buf(1 << 24);
     let clip_inp_buf = ResourceProxy::new_buf(data.n_clip as u64 * CLIP_INP_SIZE);
     recording.dispatch(
         shaders.draw_leaf,
@@ -382,7 +382,7 @@ pub fn render_full(
     // in storage rather than workgroup memory.
     let n_path_aligned = align_up(n_path as usize, 256);
     let path_buf = ResourceProxy::new_buf(n_path_aligned as u64 * PATH_SIZE);
-    let tile_buf = ResourceProxy::new_buf(1 << 20);
+    let tile_buf = ResourceProxy::new_buf(1 << 23);
     let path_wgs = (n_path + shaders::PATH_BBOX_WG - 1) / shaders::PATH_BBOX_WG;
     recording.dispatch(
         shaders.tile_alloc,
@@ -397,7 +397,7 @@ pub fn render_full(
         ],
     );
 
-    let segments_buf = ResourceProxy::new_buf(1 << 24);
+    let segments_buf = ResourceProxy::new_buf(1 << 27);
     recording.dispatch(
         shaders.path_coarse,
         (path_coarse_wgs, 1, 1),
@@ -417,7 +417,7 @@ pub fn render_full(
         (path_wgs, 1, 1),
         [config_buf, path_buf, tile_buf],
     );
-    let ptcl_buf = ResourceProxy::new_buf(1 << 24);
+    let ptcl_buf = ResourceProxy::new_buf(1 << 26);
     recording.dispatch(
         shaders.coarse,
         (width_in_bins, height_in_bins, 1),

```

### DIFF
--- a/examples/with_winit/src/main.rs
+++ b/examples/with_winit/src/main.rs
@@ -18,7 +18,11 @@ mod pico_svg;
 mod simple_text;
 mod test_scene;
 
-use vello::{util::RenderContext, Renderer, Scene, SceneBuilder};
+use vello::{
+    kurbo::{Affine, Vec2},
+    util::RenderContext,
+    Renderer, Scene, SceneBuilder,
+};
 use winit::{event_loop::EventLoop, window::Window};
 
 async fn run(event_loop: EventLoop<()>, window: Window) {
@@ -34,6 +38,11 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
     let mut current_frame = 0usize;
     let mut scene_ix = 0usize;
     let mut scene = Scene::new();
+    let mut paris_scene = None;
+    let mut drag = Vec2::default();
+    let mut scale = 1f64;
+    let mut mouse_down = false;
+    let mut prior_position = None;
     event_loop.run(move |event, _, control_flow| match event {
         Event::WindowEvent {
             ref event,
@@ -45,6 +54,9 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
                     match input.virtual_keycode {
                         Some(VirtualKeyCode::Left) => scene_ix = scene_ix.saturating_sub(1),
                         Some(VirtualKeyCode::Right) => scene_ix = scene_ix.saturating_add(1),
+                        Some(VirtualKeyCode::Escape) => {
+                            *control_flow = ControlFlow::Exit;
+                        }
                         _ => {}
                     }
                 }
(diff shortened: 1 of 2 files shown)


examples/with_winit/src/main.rs
```diff
@@ -32,8 +54,34 @@ async fn run(event_loop: EventLoop<()>, window: Window) {
     let mut renderer = Renderer::new(&device_handle.device).unwrap();
     let mut simple_text = simple_text::SimpleText::new();
     let mut current_frame = 0usize;
-    let mut scene_ix = 0usize;
     let mut scene = Scene::new();
+    let mut paris_scene = None;
```
raphlinus, line 58: I think maybe this should be called `svg_scene`.

examples/with_winit/src/test_scene.rs
```diff
@@ -75,13 +73,24 @@ pub fn render_svg(sb: &mut SceneBuilder, svg: &PicoSvg) {
...
+    sb: &mut SceneBuilder,
+    scene: &mut Option<SceneFragment>,
+    xform: Affine,
+    svg: &str,
+) {
+    if scene.is_none() {
+        use super::pico_svg::*;
+        let start = Instant::now();
+        eprintln!("Starting to parse svg");
+        let svg = PicoSvg::load(svg, 6.0).unwrap();
```
raphlinus, line 86: This could be a command line argument (it should be -1.5 for paris), but a more principled approach is to parse transforms in `<g>` elements.

examples/with_winit/src/test_scene.rs
```diff
@@ -75,13 +73,24 @@ pub fn render_svg(sb: &mut SceneBuilder, svg: &PicoSvg) {
...
-        std::str::from_utf8(include_bytes!("../../assets/Ghostscript_Tiger.svg")).unwrap();
-    let svg = PicoSvg::load(xml_str, 6.0).unwrap();
-    render_svg(sb, &svg);
+pub fn render_svg_scene(
+    sb: &mut SceneBuilder,
+    scene: &mut Option<SceneFragment>,
+    xform: Affine,
+    svg: &str,
+) {
+    if scene.is_none() {
```
raphlinus, line 82: Consider `Option::get_or_insert_with()` here, it avoids the unwrap below.

raphlinus (approved):
Confirm this works, and the code looks good. Minor nits in comments.

One thing to mention: because pico_svg currently drops `path` elements with no `fill` specified, it's not a full exercise of the paris-30k test. I've locally patched it to just add a default color, but the real solution is to parse attributes in `g` tags and pass them down so they're inherited.

examples/with_winit/src/test_scene.rs
DJMcNab: I made it a command line flag for now. Properly parsing svg is quite a task.

Some of the work in https://github.com/google/forma/blob/main/demo/src/demos/svg.rs might be useful to us, although `forma` is Apache only - no MIT option; so we need to be careful that we use it correctly if we do.

examples/with_winit/src/test_scene.rs
DJMcNab: Nice, I wasn't aware of that API.

PR #245, opened 2023-01-13
Prototype of buffer reuse

raphlinus:
This helps performance but not all performance issues have been resolved. Nontrivial CPU goes into write_buffer, and it's also possible that there isn't enough overlapping between CPU and GPU work.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -26,7 +26,7 @@ use parking_lot::RawMutex;
 use wgpu::{
     util::DeviceExt, BindGroup, BindGroupLayout, Buffer, BufferAsyncError, BufferSlice, BufferView,
     ComputePipeline, Device, Queue, Texture, TextureAspect, TextureFormat, TextureUsages,
-    TextureView, TextureViewDimension,
+    TextureView, TextureViewDimension, BufferUsages,
 };
 
 pub type Error = Box<dyn std::error::Error>;
@@ -41,6 +41,7 @@ static ID_COUNTER: AtomicU64 = AtomicU64::new(0);
 
 pub struct Engine {
     shaders: Vec<Shader>,
+    pool: ResourcePool,
 }
 
 struct Shader {
@@ -123,9 +124,17 @@ struct BindMap {
     image_map: HashMap<Id, (Texture, TextureView)>,
 }
 
+#[derive(Default)]
+struct ResourcePool {
+    bufs: HashMap<(u64, BufferUsages), Vec<Buffer>>,
+}
+
 impl Engine {
     pub fn new() -> Engine {
-        Engine { shaders: vec![] }
+        Engine {
+            shaders: vec![],
+            pool: Default::default(),
+        }
     }
 
     /// Add a shader.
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -648,3 +652,42 @@ impl<'a> DownloadsMapped<'a> {
...
+        }
+        device.create_buffer(&wgpu::BufferDescriptor {
+            label: None,
+            size: rounded_size,
+            usage,
+            mapped_at_creation: false,
+        })
+    }
+
+    fn reap_bindmap(&mut self, bind_map: BindMap) {
```
DJMcNab, line 675: Any consideration to reusing the BindMap allocation here?

Realistically, it's only a handful of bytes, so it's probably easier not to.

src/engine.rs
```diff
@@ -648,3 +652,42 @@ impl<'a> DownloadsMapped<'a> {
         Ok(slice.get_mapped_range())
     }
 }
+
+const SIZE_CLASS_BITS: u32 = 1;
+
+impl ResourcePool {
+    /// Get a buffer from the pool or create one.
+    fn get_buf(&mut self, size: u64, usage: BufferUsages, device: &Device) -> Buffer {
```
DJMcNab, line 660: As the scene grows across frames, this currently becomes unbounded. We should have a garbage collection scheme of some time (e.g. a time last used counter on each buffer, clearing when they're more than 3 seconds old or something)

src/engine.rs
```diff
@@ -648,3 +652,42 @@ impl<'a> DownloadsMapped<'a> {
...
+    fn reap_bindmap(&mut self, bind_map: BindMap) {
+        for (_id, buf) in bind_map.buf_map {
+            let size = buf.size();
+            let usage = buf.usage();
+            self.bufs.entry((size, usage)).or_default().push(buf);
+        }
+    }
+
+    /// Quantize a size up to the nearest size class.
+    fn size_class(x: u64, bits: u32) -> u64 {
```
DJMcNab, line 684: For very small values of `x`, this is very stingy, likely requiring buffer recreation for small variations in size.

Now it is true that most small buffers will be fixed size anyway, for uniforms and such.
So it's probably that this doesn't matter.

DJMcNab (commented):
I haven't run it yet, but it looks good.

src/engine.rs
raphlinus: Right now the goal is to have the code as clear and simple as possible, and only optimize the things that show up on profiles. We've got plenty of that. But I feel you, there's a case to be made for writing "idiomatically performant" code.

src/engine.rs
raphlinus: If you do the math, in the case of a scene growing, the actual total memory allocation is a reasonably small constant multiple of the working set. The constant depends on the number of bits used for size_class. The story also gets better if there's one big buffer and a bunch of smaller buffers that can reuse the allocations left over from the big ones.

That said, yes we should have eviction, but the heuristic to do it *well* is tricky, and I don't think it's blocking anything that we're doing right now. If we wanted to land this PR, I would feel comfortable having an issue.

src/engine.rs
raphlinus: Yeah again the thinking here is to use simple logic, and fine-tuned based on measurements. There's no evidence yet that this kind of change would make any actual difference.

src/engine.rs
DJMcNab: I think, as we do currently, using the front two bits, the 'lost' buffer is 75% the size of the current[^1]. That means the cost is at most a factor of:

$$\frac{1}{1 - 0.75} = 4$$

That's not that huge of a cost to eat, I agree, when our biggest supported buffer is on the order of 16MB (I think, anyway). It's certainly not ideal though. I'm happy landing this as-is, at least for now.

[^1]: Half the time, it's actually 66% the size, but I'm ignoring that because it makes my calculation easier. Also ignoring the case of size 0 and 1 as negligible

DJMcNab (approved):
I've tested this, and it all works as expected. 

It shouldn't conflict with #244 either, luckily.

raphlinus:
Thanks for the review! I'll merge this soon but in this particular case would appreciate a glance from @dfrg, as I know he's spent some time thinking the problem.

dfrg (approved):
Seems like a remarkably simple solution that solves the immediate problem really well. I share most of Daniel’s concerns but am satisfied that we can address the outstanding ones in the future. LGTM!

PR #246, opened 2023-01-13
Fix even-odd rule

raphlinus:
This works with winding numbers even larger than 2.


raphlinus:
See https://www.desmos.com/calculator/cemy6mrxmq for justification of the math.

dfrg (approved):
Doh! Isn’t two big enough for everyone?

PR #247, opened 2023-01-13
Improve the svg parser

DJMcNab:
Now it almost correctly handles nesting. We don't properly handle `opacity` nesting, but luckily this doesn't matter for paris-30k


raphlinus (approved):
Looks good on quick skim. Again, the purpose of this is to run a specific set of test files, not to be a general purpose SVG implementation. Any flaws have very limited impact.

Thanks!

PR #249, opened 2023-01-14
Clippy fixes

simbleau:
All clippy smells were fixed except for `dead_code` warnings. I assumed there are plans for dead_code.


raphlinus (approved):
Yup, these are all improvements, thanks. You should have an invite for commit access.

PR #250, opened 2023-01-14
clippy CI

simbleau:
I noticed we have fmt as a CI step but not clippy.

This MR would add clippy, but that's an opinion I don't want to enforce, so I'll add it here as a suggestion.

I chose `actions-rs/clippy-check` - pretty standard. Also nicely formats the clippy errors for the reviewer/reviewee through annotations.[ See here](https://github.com/linebender/vello/pull/250/files)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,22 +6,28 @@ on:
   pull_request:
 
 jobs:
-  rustfmt:
+  lint:
     runs-on: ubuntu-latest
     name: cargo fmt
     steps:
       - uses: actions/checkout@v2
 
-      - name: install stable toolchain
+      - name: Setup | Toolchain
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
           profile: minimal
-          components: rustfmt
+          components: rustfmt, clippy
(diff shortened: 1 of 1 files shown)


DJMcNab:
I've been avoiding adding different CI because of the dead code warnings which are blocking this.

I would love to get those resolved, but it doesn't seem to be a priority.

raphlinus:
I've been using dead code as a lazy todo tracker, as a lot of it represents things that should be wired up soon. But I think we're at the point where that's not good for collaboration, we should either add TODO comments or just strip things out that are low value.

nicoburns:
I'm in favour of a clippy CI check, but I think anything from `actions-rs` should be avoided as their actions are unmaintained. I would suggest the `dtolnay/rust-toolchain` action to install the rust toolchain and then calling `cargo clippy` manually in a script action

DJMcNab:
I think https://github.com/linebender/vello/pull/325 supercedes this PR, although it's currently a draft pr

DJMcNab:
Thanks for the PR! Unfortunately, it was blocked until we resolved #325 , which integrated similar changes

PR #251, opened 2023-01-15
Some WASM changes - including a run_wasm alias

DJMcNab:
Blocked on rukai/cargo-run-wasm#19. This PR should be able to move over to `0.4` and then come out of draft when/if that is merged.

Supercedes https://github.com/linebender/vello/pull/248

I've renamed the `run-wasm` package to the more idiomatic crate name `run_wasm`, which also aligns with our other examples.
However, I have added aliases for both forms.

I was hesistant to add aliases, because it makes the workspace root messier, and disallows setting custom `.cargo/config.toml` files.
However, writing out the full command is messy, and hopefully other tools will allow their config files in `.cargo` too, so it's not a special cost just for run-wasm.

### DIFF
--- a/examples/run_wasm/src/main.rs
+++ b/examples/run_wasm/src/main.rs
@@ -1,4 +1,17 @@
+/// Use [cargo-run-wasm](https://github.com/rukai/cargo-run-wasm) to build an example for web
+///
+/// Usage:
+/// ```
+/// cargo run -p run_wasm -- --package [example_name]
+/// ```
+/// Generally:
+/// ```
+/// cargo run -p run_wasm -- --package with_winit
+/// ```
+
 fn main() {
+    // HACK: We rely heavily on compute shaders; which means we need WebGPU to be supported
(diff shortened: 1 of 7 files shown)


examples/run_wasm/src/main.rs
```diff
@@ -1,3 +1,21 @@
...
+/// ```
+/// cargo run_wasm --package [example_name]
+/// ```
+/// Generally:
+/// ```
+/// cargo run_wasm --package with_winit
+/// ```
+
 fn main() {
+    // HACK: We rely heavily on compute shaders; which means we need WebGPU to be supported
```
raphlinus, line 13: I'm not going to block the PR on this, but I'd word the comment differently. It enables an unstable web feature, and it's basically a TODO to remove the flag when it gets stabilized and we bump the dependency on web_sys to a version that includes WebGPU. That's not exactly a hack, at least in my opinion.

raphlinus (approved):
Thanks!

PR #252, opened 2023-01-15
Add hot reloading of shaders to the winit example

DJMcNab:
In theory, this should be useful for debugging. 

Obviously, a good proportion of changes can't be hot-reloaded. But this should make those which can be a bit easier.

I don't believe this can cause memory leaks, since everything created in `reload_shaders` should be properly dropped.

### DIFF
--- a/examples/with_winit/src/hot_reload.rs
+++ b/examples/with_winit/src/hot_reload.rs
@@ -1,0 +1,29 @@
+use std::{path::Path, time::Duration};
+
+use notify_debouncer_mini::{new_debouncer, notify::*, DebounceEventResult};
+
+pub(crate) fn hot_reload(mut f: impl FnMut() -> Option<()> + Send + 'static) -> impl Sized {
+    let mut debouncer = new_debouncer(
+        Duration::from_millis(500),
+        None,
+        move |res: DebounceEventResult| match res {
+            Ok(_) => f().unwrap(),
+            Err(errors) => errors.iter().for_each(|e| println!("Error {:?}", e)),
+        },
+    )
+    .unwrap();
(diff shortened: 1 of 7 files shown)


examples/with_winit/src/hot_reload.rs
```diff
@@ -0,0 +1,28 @@
...
+    .unwrap();
+
+    debouncer
+        .watcher()
+        .watch(
+            dbg!(&Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("../../shader")
+                .canonicalize()
+                .unwrap()),
+            // We currently don't support hot reloading the imports, so don't recurse inot there
```
derekdreery, line 23: `inot -> if not` (I assume)

derekdreery (approved):
I've reviewed the code and it looks good. I don't think we need a wide approval for this because it's small and easy to revert if there are any issues.

DJMcNab:
I'd prefer to get confirmation from @raphlinus that we do want this, and ideally confirmation that it works on other platforms too. I don't think anything I've done is platform dependendant, but filesystems can be finicky sometimes.

raphlinus (approved):
I approve making the developer experience smoother, and this leverages the benefits of runtime shader compilation. The edit-compile-test cycle time hasn't affected me much (partly because I have a fast machine), but I can see cases where it would help.

It seems this also improves the error reporting from wgpu? That's also a good thing.

DJMcNab:
I don't think this makes any difference to the error reporting, apart from when hot-reloading? But yes, when hot-reloading, the errors are shown, rather than just crashing and burning.

PR #254, opened 2023-01-16
Allow giving the buffers names

DJMcNab:
I also now unconditionally name shaders and compute pipelines.

Naming buffers however is an additional feature, as it potentially has consequences for buffer reuse efficiency, since we specify buffers with the given names.

I can confirm that this works in renderdoc.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -121,6 +121,7 @@ pub enum BindType {
 
 struct BindMapBuffer {
     buffer: Buffer,
+    #[cfg_attr(not(feature = "buffer_labels"), allow(unused))]
     label: &'static str,
 }
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -496,7 +496,6 @@ impl BindMap {
             proxy.id,
             BindMapBuffer {
                 buffer,
-                #[cfg(feature = "buffer_labels")]
                 label: proxy.name,
             },
         );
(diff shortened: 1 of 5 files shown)


src/engine.rs
```diff
@@ -475,8 +491,15 @@ impl Id {
...
 
 impl BindMap {
-    fn insert_buf(&mut self, id: Id, buf: Buffer) {
-        self.buf_map.insert(id, buf);
+    fn insert_buf(&mut self, proxy: &BufProxy, buffer: Buffer) {
+        self.buf_map.insert(
+            proxy.id,
+            BindMapBuffer {
+                buffer,
+                #[cfg(feature = "buffer_labels")]
```
raphlinus, line 499: This does need a None case (or taking the field out of BindMapBuffer, which might make more sense) if the cfg is not enabled.

DJMcNab: I originally intended to not have the field, but that meant the `buffer_labels` feature propogated very far.

The only real cost to having the names is the buffer reuse, as far as I know; the binary size for the string literals is negligible. I therefore chose to keep the names in `BufProxy` and `BindMapBuffer` in all cases, for ease of maintenance

raphlinus:
The Xcode crashes are happening on main, it doesn't seem like this patch is responsible. It might just be that my machine is in a bad state.

src/engine.rs
raphlinus: If the cost is low, I don't think it justifies having the feature, I'd be ok with having it on.

PR #256, opened 2023-01-17
Update for peniko changes and pin git rev

dfrg:
This applies updates for the gradient API in peniko and pins the git dependency so prevent further breakage. Also removes Cargo.lock.


Cargo.toml
```diff
@@ -26,7 +26,7 @@ parking_lot = "0.12"
 bytemuck = { version = "1.12.1", features = ["derive"] }
 smallvec = "1.8.0"
 moscato = { git = "https://github.com/dfrg/pinot" }
-peniko = { git = "https://github.com/linebender/peniko" }
+peniko = { git = "https://github.com/linebender/peniko", rev = "ba96b90" }
```
raphlinus, line 29: Do we want to pin the one above as well?

.gitignore
```diff
@@ -1 +1,2 @@
 /target
+Cargo.lock
```
raphlinus, line 2: I'm not 100% sure what to do with this - having it for purely library use is not useful, but a nontrivial amount of development is the examples. I think ok to drop it as you've done here, but I'm curious if other people have thoughts.

raphlinus (approved):
The gradient changes look fine, thanks!

Cargo.toml
dfrg: We can but I don't plan any further updates to moscato and it will go away as soon as googlefonts/fontations#191 lands.

dfrg: Added a pin just for consistency. Doesn't hurt!

.gitignore
dfrg: I don't have strong feelings on this, but I think I'll merge as to move forward and ask for commentary on zulip.

Speykious: The Cargo book has an answer as to what to do with Cargo.lock:
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
> If you’re building a non-end product, such as a rust library that other rust [packages](https://doc.rust-lang.org/cargo/appendix/glossary.html#package) will depend on, put `Cargo.lock` in your `.gitignore`. If you’re building an end product, which are executable like command-line tool or an application, or a system library with crate-type of `staticlib` or `cdylib`, check `Cargo.lock` into `git`.
It then links to [this section of their FAQ](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) which explains the reasoning behind this.

Thus (and I'm quite surprised myself as I believed the contrary until I searched it up just now) it seems like ignoring Cargo.lock for Xilem is actually a good idea.

PR #257, opened 2023-01-17
Robust memory redux (GPU side only)

dfrg:
Initial cut of robust memory for the WGSL port. This currently only handles the GPU side but is missing reading/writing the spilled blend stack in fine.

### DIFF
--- a/shader/binning.wgsl
+++ b/shader/binning.wgsl
@@ -127,7 +127,11 @@ fn main(
         sh_count[i][local_id.x] = element_count_packed;
     }
     // element_count is the number of draw objects covering this thread's bin
-    let chunk_offset = atomicAdd(&bump.binning, element_count);
+    var chunk_offset = atomicAdd(&bump.binning, element_count);
+    if chunk_offset > bump.binning_size {
+        chunk_offset = 0u;
+        atomicOr(&bump.failed, STAGE_BINNING);
+    }    
     sh_chunk_offset[local_id.x] = chunk_offset;
     bin_header[global_id.x].element_count = element_count;
     bin_header[global_id.x].chunk_offset = chunk_offset;
--- a/shader/binning.wgsl
+++ b/shader/binning.wgsl
@@ -128,7 +128,7 @@ fn main(
     }
     // element_count is the number of draw objects covering this thread's bin
     var chunk_offset = atomicAdd(&bump.binning, element_count);
-    if chunk_offset > bump.binning_size {
+    if chunk_offset + element_count > config.binning_size {
         chunk_offset = 0u;
         atomicOr(&bump.failed, STAGE_BINNING);
     }    
--- a/shader/coarse.wgsl
+++ b/shader/coarse.wgsl
@@ -70,8 +70,11 @@ fn alloc_cmd(size: u32) {
         // We might be able to save a little bit of computation here
         // by setting the initial value of the bump allocator.
         let ptcl_dyn_start = config.width_in_tiles * config.height_in_tiles * PTCL_INITIAL_ALLOC;
-        let new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
-        // TODO: robust memory
+        var new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
+        if new_cmd > bump.ptcl_size {
+            new_cmd = 0u;
+            atomicOr(&bump.failed, STAGE_COARSE);
+        }
         ptcl[cmd_offset] = CMD_JUMP;
         ptcl[cmd_offset + 1u] = new_cmd;
         cmd_offset = new_cmd;
@@ -142,6 +145,9 @@ fn main(
     @builtin(local_invocation_id) local_id: vec3<u32>,
     @builtin(workgroup_id) wg_id: vec3<u32>,
 ) {
+    if (atomicLoad(&bump.failed) & (STAGE_BINNING | STAGE_TILE_ALLOC | STAGE_PATH_COARSE)) != 0u {
+        return;
+    }     
     let width_in_bins = (config.width_in_tiles + N_TILE_X - 1u) / N_TILE_X;
     let bin_ix = width_in_bins * wg_id.y + wg_id.x;
     let n_partitions = (config.n_drawobj + N_TILE - 1u) / N_TILE;
@@ -169,6 +175,10 @@ fn main(
     // blend state
     var render_blend_depth = 0u;
     var max_blend_depth = 0u;
+
+    let blend_offset = cmd_offset;
+    cmd_offset += 1u;
+    cmd_limit -= 1u;
 
     while true {
         for (var i = 0u; i < N_SLICE; i += 1u) {
@@ -401,6 +411,9 @@ fn main(
     }
     if bin_tile_x + tile_x < config.width_in_tiles && bin_tile_y + tile_y < config.height_in_tiles {
         ptcl[cmd_offset] = CMD_END;
-        // TODO: blend stack allocation
+        if max_blend_depth > BLEND_STACK_SPLIT {
+            let scratch_size = max_blend_depth * TILE_WIDTH * TILE_HEIGHT * 4u;
+            ptcl[blend_offset] = atomicAdd(&bump.blend, scratch_size);
+        }
     }
 }
--- a/shader/coarse.wgsl
+++ b/shader/coarse.wgsl
@@ -71,7 +71,7 @@ fn alloc_cmd(size: u32) {
         // by setting the initial value of the bump allocator.
         let ptcl_dyn_start = config.width_in_tiles * config.height_in_tiles * PTCL_INITIAL_ALLOC;
         var new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
-        if new_cmd > bump.ptcl_size {
+        if new_cmd + PTCL_INCREMENT > config.ptcl_size {
             new_cmd = 0u;
             atomicOr(&bump.failed, STAGE_COARSE);
         }
@@ -137,7 +137,7 @@ fn write_end_clip(end_clip: CmdEndClip) {
     ptcl[cmd_offset] = CMD_END_CLIP;
     ptcl[cmd_offset + 1u] = end_clip.blend;
     ptcl[cmd_offset + 2u] = bitcast<u32>(end_clip.alpha);
-    cmd_offset += 3u;            
+    cmd_offset += 3u;
 }
 
 @compute @workgroup_size(256)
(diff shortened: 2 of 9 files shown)


shader/coarse.wgsl
```diff
@@ -142,6 +145,9 @@ fn main(
     @builtin(local_invocation_id) local_id: vec3<u32>,
     @builtin(workgroup_id) wg_id: vec3<u32>,
 ) {
+    if (atomicLoad(&bump.failed) & (STAGE_BINNING | STAGE_TILE_ALLOC | STAGE_PATH_COARSE)) != 0u {
```
DJMcNab, line 148: Is there any advantage here to checking only whether previous stages failed (vs just doing a general non-zero check). Is it a case of branching paths in the shaders which reconverge?

As it stands, since we don't check for prior failure in `fine`, we seem to saying that if `coarse` fails, we should still run `fine` on a best effort basis? That seems fairly reasonable, although I don't think there's any guarantee that what we got out of fine is still usable?

src/render.rs
```diff
@@ -54,6 +56,22 @@ pub const fn next_multiple_of(val: u32, rhs: u32) -> u32 {
     }
 }
 
+// This must be kept in sync with the struct in shaders/shared/bump.wgsl
```
DJMcNab, line 59: I think the "inverse" of this in `shader/shared/bump.wgsl` is missing 

Also:
```suggestion
// This must be kept in sync with the struct in shader/shared/bump.wgsl
```

src/render.rs
```diff
@@ -185,6 +203,22 @@ pub fn render_encoding_full(
     let scene_buf = ResourceProxy::Buf(recording.upload("scene", packed.data));
     let config_buf =
         ResourceProxy::Buf(recording.upload_uniform("config", bytemuck::bytes_of(&config)));
+    let info_size = config.layout.bin_data_start;
+    let bump = BumpAllocators {
+        binning_size: ((1 << 20) / 4) - info_size,
+        ptcl_size: (1 << 25) / 4,
```
DJMcNab, line 209: I'd be tempted to do `size_of::<u32>()` here and above, just to make things really obvious

My initial reaction when seeing this was "why not just not shift left as far". I see that this version shows in-source the exact sizes of the buffers, whereas the version which doesn't shift as far then hides the buffer sizes.

src/render.rs
```diff
@@ -54,6 +56,22 @@ pub const fn next_multiple_of(val: u32, rhs: u32) -> u32 {
...
 
+// This must be kept in sync with the struct in shaders/shared/bump.wgsl
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, Zeroable, Pod)]
+struct BumpAllocators {
+    failed: u32,
+    binning_size: u32,
+    ptcl_size: u32,
+    tiles_size: u32,
+    segments_size: u32,
```
DJMcNab, line 67: Also, is there a reason these aren't just measured using `arrayLength` in the shaders instead? I guess it's useful to keep the data around CPU side.

```suggestion
    // Sizes of the provided buffers
    binning_size: u32,
    ptcl_size: u32,
    tiles_size: u32,
    segments_size: u32,
    // Final needed dynamic size of the buffers. If any of these are larger than the corresponding `_size` element
    // reallocation needs to occur
```

shader/coarse.wgsl
DJMcNab: Ah got it, we need to dry-run all the possible allocations to get the actual final needed size. That still doesn't explain why it's not a `bump.failed & ~STAGE_COARSE` though, or why we don't check in fine.

src/render.rs
DJMcNab: I'll add a note here, hopefully one of us will remember at the time it's relevant:

It would be short-sighted to set the corresponding `_size` to exactly this requested value, because we round buffer sizes up in `engine`; this should be set based on the the new rounded sizes instead.

DJMcNab (commented):
Looks like a good start. Not approving as obviously not in a final state.

Some of my questions/confusions can potentially be turned into comments, so it's clearer for future readers.

shader/coarse.wgsl
dfrg: We could use the complement but I think the current code makes the dependencies explicit which is nice for the reader. 

Fine doesn’t contain a check because the CPU side (which isn’t implemented yet) is meant to read back the bump buffer after coarse and rerun the stages that failed until they succeed. We could also speculatively run fine and that may prove to be a reasonable choice.

src/render.rs
dfrg: Yes, the size calculation code is nasty across the board. We will eventually replace this with better heuristics for initial sizes and typed buffer handles that handle element sizing automatically.

src/render.rs
dfrg: I’m reluctant to use `arrayLength` because it’s not supported on Metal without contrivances and I’d prefer not to have to deal with those in a future native backend.

src/render.rs
dfrg: It is missing. Will fix.

shader/coarse.wgsl
DJMcNab: My intuition would be that reading back the bump buffer to the CPU within every frame would be extremely expensive, when we expect only a miniscule proportion of frames to require it.

But idk, GPUs are a bit unintuitive sometimes.

dfrg: Raph and I spent a great deal of time discussing that before it was implemented in piet-gpu. It ended up having a negligible impact on performance.

dfrg:
This is incomplete but functional so pending review from @raphlinus, I think we should actually land this to avoid potential merge conflicts.

DJMcNab:
I think these bail out checks could do with an explanatory comment; something along the lines of

```
// Exit early if prior stages failed, as we can't run this stage
// We need to check only prior stages, as if this stage has failed in another workgroup, 
// we still want to know this workgroup's memory requirement
```

shader/binning.wgsl
```diff
@@ -127,7 +127,11 @@ fn main(
         sh_count[i][local_id.x] = element_count_packed;
     }
     // element_count is the number of draw objects covering this thread's bin
-    let chunk_offset = atomicAdd(&bump.binning, element_count);
+    var chunk_offset = atomicAdd(&bump.binning, element_count);
+    if chunk_offset > bump.binning_size {
```
raphlinus, line 131: I think this needs to be `chunk_offset + element_count`. What you get back is the previous value.

shader/coarse.wgsl
```diff
@@ -70,8 +70,11 @@ fn alloc_cmd(size: u32) {
         // We might be able to save a little bit of computation here
         // by setting the initial value of the bump allocator.
         let ptcl_dyn_start = config.width_in_tiles * config.height_in_tiles * PTCL_INITIAL_ALLOC;
-        let new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
-        // TODO: robust memory
+        var new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
+        if new_cmd > bump.ptcl_size {
```
raphlinus, line 74: Same, need to add bump.

shader/coarse.wgsl
```diff
@@ -170,6 +176,9 @@ fn main(
     var render_blend_depth = 0u;
     var max_blend_depth = 0u;
 
+    let blend_offset = cmd_offset;
+    cmd_offset += 1u;
```
raphlinus, line 180: BTW we can use `++` in WGSL. I'm just used to not having it in Rust.

shader/coarse.wgsl
```diff
@@ -401,6 +410,9 @@ fn main(
     }
     if bin_tile_x + tile_x < config.width_in_tiles && bin_tile_y + tile_y < config.height_in_tiles {
         ptcl[cmd_offset] = CMD_END;
-        // TODO: blend stack allocation
+        if max_blend_depth > BLEND_STACK_SPLIT {
+            let scratch_size = max_blend_depth * TILE_WIDTH * TILE_HEIGHT * 4u;
```
raphlinus, line 414: What's the `4u` for? Atomic bump allocation is in elements (u32 here) in Vello but was bytes in piet-gpu.

I guess there's no other code that depends on this yet, so it could go either way, but generally prefer consistency.

shader/coarse.wgsl
```diff
@@ -401,6 +410,9 @@ fn main(
...
+            ptcl[blend_offset] = atomicAdd(&bump.blend, scratch_size);
```
raphlinus, line 415: It seems like this needs to set the failure flag?

shader/path_coarse_full.wgsl
```diff
@@ -93,7 +93,12 @@ fn eval_cubic(p0: vec2<f32>, p1: vec2<f32>, p2: vec2<f32>, p3: vec2<f32>, t: f32
 }
 
 fn alloc_segment() -> u32 {
-    return atomicAdd(&bump.segments, 1u) + 1u;
+    var offset = atomicAdd(&bump.segments, 1u) + 1u;
```
raphlinus, line 96: Similar as before.

shader/shared/bump.wgsl
```diff
@@ -1,9 +1,22 @@
...
+let STAGE_TILE_ALLOC: u32 = 0x2u;
+let STAGE_PATH_COARSE: u32 = 0x4u;
+let STAGE_COARSE: u32 = 0x8u;
+
+// This must be kept in sync with the struct in src/render.rs
 struct BumpAllocators {
+    // Bitmask of stages that have failed allocation.
+    failed: atomic<u32>,
+    binning_size: u32,
+    ptcl_size: u32,
```
raphlinus, line 14: I prefer the size fields be in `Config`, I'm fairly confident the caching will be better reading from uniforms than here, as the cache line will be constantly bouncing.

raphlinus (commented):
Overall looks pretty close to what I was expecting. I have some concerns, including what might be a confusion about what `atomicAdd` returns.

shader/binning.wgsl
dfrg: Oops, yes. I think I had that in there originally and removed it for some reason. I’ll adjust them all.

shader/coarse.wgsl
dfrg: Good catch, ty.

shader/coarse.wgsl
dfrg: Should it? This can’t really fail as we don’t actually use the allocation until fine. We’re just doing the accounting. But I could be missing something.

shader/shared/bump.wgsl
dfrg: Ok, will change. My thought here was that we’ll only need to update one buffer on rerun of course but your concern makes sense to me.

dfrg: I meant coarse, of course :)

shader/coarse.wgsl
dfrg: Me too and I've come to prefer it.

dfrg:
This should be good to go now.

raphlinus (approved):
Looks good, thanks!

PR #258, opened 2023-01-18
Do another pass on the README

DJMcNab:
As discussed in the office hours, this updates references to Vello to treat it as a proper noun (except when referring to a crate, when `vello` is used instead).

I also added a badge [wgpu version](https://crates.io/crates/wgpu) to show the WGPU version. That's because we expect people to use vello in wgpu applications, for which they need to have the same wgpu as us.

Additionally, I added a link in prose to the Zulip, in addition to the badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <div align="center">
 
-# vello
+# Vello
 
 **An experimental GPU compute-centric 2D renderer**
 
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
Thanks, looks good. Clear communication here is important, especially for newcomers to the project, so I definitely appreciate improvements.

PR #261, opened 2023-01-21
Allow selecting GPU from env

jneem:
Running the `with_winit` example fails on my machine because it chooses the "wrong" gpu and then fails with `[destroyed object]: error 7: importing the supplied dmabufs failed`.

I'm not sure what's the *best* fix for this (i509VCB suggested on the wgpu helix that newer mesas may do this better), but this pr makes it possible to at least override the desired adapter from the environment. So after this pr, `cargo run -p with_winit` still fails for me but

`WGPU_ADAPTER_NAME="AMD RADV DIMGREY_CAVEFISH" cargo run -p with_winit`

succeeds.

### DIFF
--- a/examples/with_winit/Cargo.toml
+++ b/examples/with_winit/Cargo.toml
@@ -15,6 +15,7 @@ pollster = "0.2.5"
 # for picosvg
 roxmltree = "0.13"
 clap = { version = "4.1.0", features = ["derive"] }
+env_logger = "0.10.0"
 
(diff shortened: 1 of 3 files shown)


DJMcNab (approved):
Looks good to me. I can confirm it runs locally, althouhg I only have one GPU.

Logging properly does make sense, although I'm getting some Vulkan validation issues coming through, about swapchain handling.

PR #262, opened 2023-01-21
Split the examples into frontends with a shared scene repository

DJMcNab:
Follow up from #260.

Merges the with_winit and usvg_viewer examples

Also librarifies the new svg handling from #260, to allow using it seperately (e.g. in Xilem applications)


raphlinus (approved):
I decided to go ahead with this. I have reservations about removing pico_svg as I think we may have some use for it, but we can always bring the code back if and when that happens. In the meantime, it'll be cleaner to have one supported SVG implementation. Many thanks!

PR #263, opened 2023-01-24
Update to wgpu 0.15

DJMcNab:
This is just the lowest friction path to wgpu main.

The biggest issue is naga has changed global variables from let to const, in line with the spec. This is not yet supported by wgsl-analyzer.


raphlinus (approved):
I assume we'll have another PR to move the shaders to `const` when wgsl-analyzer catches up? I imagine this would be a pretty big paper cut now.

DJMcNab:
Yeah, I imagine so. It's unclear when wgsl-analyzer will update - there's only one maintainer.

Issue #237, opened 2023-01-06
Resolve uniformity analysis warnings


raphlinus:
The coarse shader gives warnings in WebGPU about failing uniformity analysis. The `workgroupUniformLoad` intrinsic is the principled way to resolve this, and is now in the spec. An implementation is expected soon in tint, and we'd like to (a) verify that this does solve the problem as expected, and (b) be robust to those warnings being turned back into errors, as barriers in non-uniform control flow are potentially a security problem.

Because we expect this feature won't land in naga for a while, we'll make the use of the intrinsic gated by an `#ifdef`, and only turn it on in web targets for now.

raphlinus:
Should be working warning-free in main now.

PR #267, opened 2023-01-26
Fixes to get example running in wasm

raphlinus:
A number of things were wrong:

* The args were missing to `run`
* The robust memory changes introduced uniformity errors
* `clear_buffer` is a todo for wgpu on wasm
* Some more time calls crept in
* Initializing both env_logger and console_logger fails

In addition, we conditionally opt the shaders into `workgroupUniformLoad`, as that's available on wasm but not yet native.

Some of the things (args, uniformity errors) are important fixes. Other things (clear_buffer, wUL being optional) are workarounds for wgpu limitations and have TODO items to be removed when wgpu catches up.

### DIFF
--- a/examples/with_winit/src/main.rs
+++ b/examples/with_winit/src/main.rs
@@ -225,6 +225,9 @@ enum UserEvent {
 
 fn main() {
     let args = Args::parse();
+    // TODO: initializing both env_logger and console_logger fails on wasm.
+    // Figure out a more principled approach.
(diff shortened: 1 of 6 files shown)


raphlinus:
This will close #237, as it uses `workgroupUniformLoad` on wasm targets.

dfrg (approved):
I see what’s going on but the changes here are… unfortunate :(

PR #271, opened 2023-02-03
Introduce test scenes that demonstrate conflation artifacts

armansito:
Added two scenes that demonstrate the conflation artifact problem as described in https://github.com/linebender/vello/issues/49. The first scene demonstrates adjacent triangles and rects that belong to the same path and use opposite winding.

The second scene demonstrates strokes with overlapping square caps (these strokes are currently expressed as rects painted with the NonZero fill rule).

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -414,11 +414,14 @@ fn conflation_artifacts(sb: &mut SceneBuilder, _: &mut SceneParams) {
     let x = N + 0.5; // Fractional pixel offset reveals the problem on axis-aligned edges.
     let mut y = N;
 
+    let bg_color = Color::rgb8(255, 194, 19);
+    let fg_color = Color::rgb8(12, 165, 255);
+
     // Two adjacent triangles touching at diagonal edge with opposing winding numbers
     sb.fill(
         Fill::NonZero,
         Affine::translate((x, y)) * scale,
-        Color::RED,
+        fg_color,
         None,
         &[
             // triangle 1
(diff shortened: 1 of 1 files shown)


DJMcNab:
Scene 1 `conflation_artifacts` (with feasibly less colourblind-unfriendly colours):
scene_1

Presumably, all three squares should be solid colour. Note that the second and third squares have an underlaid red `#ff0000` background  (or well `#0000ff` blue), whereas the first has no such underlay, so the background black is seeping through.

Scene 2 `labyrinth`: 
image

I'm not quite sure which part of this is interesting, presumably the extensions at the intersections

armansito:
@DJMcNab I'd be happy to change the colors for the test case if you prefer that. What you described is correct, in the `conflation_artifacts` scene, all 3 squares should have a solid color. This demonstrates when opposite winding at adjacent edges "cancel each other out".

The `labyrinth` scene, if rendered correctly, should not have the "bulging" effect where the rectangles overlap. You can see what it _should_ look like by changing the translation offset of the labyrinth from `Affine::translate((20.5, 20.5))` to `Affine::translate((20., 20.))` which will cause the edges of the shapes to fall on exact pixel boundaries, leading to 100% coverage of covered pixels and will only show solid lines. Whereas the fractional offsets in the test scene lead to partial pixel coverage, which makes the artifact visible where coverage is over-counted on overlapping regions which affects the alpha mask (I believe in this case you get 100% alpha instead of 50% as with the non-overlapping edges).

DJMcNab:
My only experience with colourblind-friendliness is from watching [Cracking The Cryptic](https://www.youtube.com/@CrackingTheCryptic), where they prefer to use their blue and orange colours when colouring.

I strongly suspect that red and green aren't a good set of colours to use, but I'm not certain what a good set would be.

I don't have colourblindness, so without using external tools, I can't say if the scene in this PR is accessible.

armansito:
I changed the colors to the following. These foreground/background colors should maintain a high contrast across the types of colorblindness I'm aware of based on [this tool](https://davidmathlogic.com/colorblind/#%23E1C213-%230C7BDC-%23156bf1).

Screenshot 2023-02-03 at 12 29 10 PM

raphlinus (approved):
Thanks, these are useful for showing conflation artifacts and will be helpful for verifying that attempts to remove them actually work.

PR #272, opened 2023-02-04
Update dependencies, and add more metadata

DJMcNab:
I've also made sure to use workspace inheritance where it makes sense - this includes any dependencies used in multiple crates.

The only real dependency update is `winit 0.28`, which appears to work fine. I'm intending to update our winit example so it can be used on Android properly, but I need to find a good example of how to do that.

I've left `rust-version` as commented out because we don't have an MSRV policy. We should probably make one, but it's not urgent.

Because of these changes, the root `vello` crate would now be published using `cargo publish`. Because of this, I've added `publish=false` to it, until we're ready.

I've also changed the inherited version to `0.0.1`, as we're still e.g. printing debug information each frame, have no CI, etc.
I've also removed the `roxmltree` dependency from `with_winit`, although it is still used in the tree through usvg anyway

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ repository = "https://github.com/linebender/vello"
 name = "vello"
 description = "An experimental GPU compute-centric 2D renderer"
 categories = ["rendering", "graphics"]
-keywords = ["2d", "vector-graphics"]                            # , "xilem"?
+keywords = ["2d", "vector-graphics"]
 
 # This crate is intended for publishing, but not ready yet
 publish = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,16 +12,29 @@ members = [
 
 [workspace.package]
 edition = "2021"
-version = "0.1.0"
-
-[workspace.dependencies]
-wgpu = "0.15"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -12,16 +12,29 @@ members = [
...
+repository = "https://github.com/linebender/vello"
 
 [package]
 name = "vello"
-version = "0.1.0"
-license = "MIT/Apache-2.0"
-edition = "2021"
+description = "An experimental GPU compute-centric 2D renderer"
+categories = ["rendering", "graphics"]
+keywords = ["2d", "vector-graphics"]                            # , "xilem"?
```
raphlinus, line 25: I don't think "xilem" is an appropriate keyword. Otherwise, all this is fine.

raphlinus (approved):
Looks good, thanks!

Cargo.toml
DJMcNab: My logic was that it's the "backend" for Xilem, and it seemed useful to have all our "new stack" crates unified under one keyword umbrella

I'll remove this comment though; I'm not married to the idea.

PR #273, opened 2023-02-04
Make the `with_winit` example run on android

### LINKED ISSUE #78: Android port

DJMcNab:
To run the example on android, use `cargo apk run --package with_winit`.

I believe all other platforms should still work. The one which is more awkward is web, for which I don't have a testing set up.

Fixes #78


m-hugo:
haven't looked at the rust code but changing test_scenes.rs "../../assets/Ghostscript_Tiger.svg"  to "/../assets/Ghostscript_Tiger.svg" made the 'file not found' when targeting android disappear and desktop still works

and I'm not sure how you ran it i know 3 ways:

- cargo apk, gave an error
- cargo ndk + gradlew, no gradlew
- xbuild, had to change name = "with_winit_bin"
 to name = "with_winit" to make it work

now it runs but, both in debug and release, it shows a black screen and I get the log below followed by many other 'Failed to read shader'
after it stops 2min later I finally get a blue triangle I can manipulate

```
Starting: Intent { act=android.intent.action.MAIN cmp=com.example.with_winit/android.app.NativeActivity }
pid of com.example.with_winit is 22383
--------- beginning of main

22383 22383 I mple.with_wini: Late-enabling -Xcheck:jni
22383 22383 I mple.with_wini: Unquickening 17 vdex files!
22383 22383 W System  : ClassLoader referenced unknown path:
22383 22383 I Perf    : Connecting to perf service.
22383 22383 D NetworkSecurityConfig: No Network Security Config specified, using platform default
22383 22383 D NetworkSecurityConfig: No Network Security Config specified, using platform default
22383 22422 I RustStdoutStderr: Creating: 0xb400007cc95db610, saved_state = 0x0, save_state_size = 0
22383 22422 I RustStdoutStderr: Config: Configuration {
22383 22422 I RustStdoutStderr:     mcc: 208,
22383 22422 I RustStdoutStderr:     mnc: Some(
22383 22422 I RustStdoutStderr:         1,
22383 22422 I RustStdoutStderr:     ),
22383 22422 I RustStdoutStderr:     lang: Some(
22383 22422 I RustStdoutStderr:         "en",
22383 22422 I RustStdoutStderr:     ),
22383 22422 I RustStdoutStderr:     country: Some(
22383 22422 I RustStdoutStderr:         "GB",
22383 22422 I RustStdoutStderr:     ),
22383 22422 I RustStdoutStderr:     orientation: Port,
22383 22422 I RustStdoutStderr:     touchscreen: Finger,
22383 22422 I RustStdoutStderr:     density: Some(
22383 22422 I RustStdoutStderr:         289,
22383 22422 I RustStdoutStderr:     ),
22383 22422 I RustStdoutStderr:     keyboard: NoKeys,
22383 22422 I RustStdoutStderr:     navigation: NoNav,
22383 22422 I RustStdoutStderr:     keys_hidden: Soft,
22383 22422 I RustStdoutStderr:     nav_hidden: Any,
22383 22422 I RustStdoutStderr:     sdk_version: 30,
22383 22422 I RustStdoutStderr:     screen_size: Large,
22383 22422 I RustStdoutStderr:     screen_long: Yes,
22383 22422 I RustStdoutStderr:     ui_mode_type: Normal,
22383 22422 I RustStdoutStderr:     ui_mode_night: Yes,
22383 22422 I RustStdoutStderr: }
22383 22423 D vulkan  : searching for layers in '/data/app/~~eTd8vsk--8GzkkpgFQy4pQ==/com.example.with_winit-QSotrvY5mA5hgxNFoT8xLg==/lib/arm64'
22383 22423 D vulkan  : searching for layers in '/data/app/~~eTd8vsk--8GzkkpgFQy4pQ==/com.example.with_winit-QSotrvY5mA5hgxNFoT8xLg==/base.apk!/lib/arm64-v8a'
22383 22423 W wgpu_hal::vulkan::ins..: Unable to find layer: VK_LAYER_KHRONOS_validation
22383 22423 I AdrenoVK-0: ===== BEGIN DUMP OF OVERRIDDEN SETTINGS =====
22383 22423 I AdrenoVK-0: ===== END DUMP OF OVERRIDDEN SETTINGS =====
22383 22423 I AdrenoVK-0: QUALCOMM build          : 559e78c, Ib9d997268a
22383 22423 I AdrenoVK-0: Build Date              : 10/13/20
22383 22423 I AdrenoVK-0: Shader Compiler Version : EV031.32.02.02
22383 22423 I AdrenoVK-0: Local Branch            : mybrancheb0d76cb-f27a-f1ff-9cdc-1458d0b52ae8
22383 22423 I AdrenoVK-0: Remote Branch           : quic/gfx-adreno.lnx.1.0.r99-rel
22383 22423 I AdrenoVK-0: Remote Branch           : NONE
22383 22423 I AdrenoVK-0: Reconstruct Branch      : NOTHING
22383 22423 I AdrenoVK-0: Build Config            : S P 10.0.7 AArch64
22383 22423 I AdrenoVK-0: Driver Path             : /vendor/lib64/hw/vulkan.adreno.so
22383 22423 W winit::platform_impl:..: TODO: forward onStart notification to application
22383 22423 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_DEMAND_REFRESH
22383 22423 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_CONTINUOUS_REFRESH
22383 22423 W wgpu_core::instance: Missing downlevel flags: SURFACE_VIEW_FORMATS
22383 22423 W wgpu_core::instance: The underlying API or device in use does not support enough features to be a fully compliant implementation of WebGPU. A subset of the features canstill be used. If you are running this program on native and not in a browser and wish to limit the features you use to the supported subset, call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the current platform supports.
22383 22423 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_DEMAND_REFRESH
22383 22423 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_CONTINUOUS_REFRESH
22383 22423 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_DEMAND_REFRESH
22383 22423 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_CONTINUOUS_REFRESH
22383 22422 I RustStdoutStderr: Failed to read shader pathtag_reduce, error falling back toversion at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
22383 22422 I RustStdoutStderr: Failed to read shader pathtag_reduce2, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
22383 22422 I RustStdoutStderr: Failed to read shader pathtag_scan1, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file ordirectory" }
22383 22396 I mple.with_wini: Thread[6,tid=22396,WaitingInMainSignalCatcherLoop,Thread*=0xb400007d795ea630,peer=0x12f80318,"Signal Catcher"]: reacting to signal 3
22383 22396 I mple.with_wini:
22383 22396 I mple.with_wini: Wrote stack traces to tombstoned
22383 22422 I RustStdoutStderr: Failed to read shader pathtag_scan, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
22383 22422 I RustStdoutStderr: Failed to read shader pathtag_scan, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
22383 22422 I RustStdoutStderr: Failed to read shader bbox_clear, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
22383 22422 I RustStdoutStderr: Failed to read shader pathseg, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
22383 22422 I RustStdoutStderr: Failed to read shader draw_reduce, error falling back to version at compilation time. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

DJMcNab:
The way to run it is `cargo apk run --package with_winit`.

It taking a while to load is somewhat expected if you're using debug mode, it's doing on-device shader compilation.

I should probably disable the hot reload feature on android too, which would get rid of the spam. Are you quite sure it's taking two minutes, and that's not an exaggeration? On my device it takes on the order of 10 seconds, which is still too long, but manageable

I'll try and understand the `/` change. I have a theory, but need to check it.

I don't understand why you'd need to change the binary name for `xbuild`; it shouldn't be running the binary, as it should be using the android entry point in `lib.rs`. I haven't looked into it though, as it's not the supported way to run this example.

m-hugo:
I updated cargo-apk and the error disappeared
debug 1m17 between app launch and triangle
release 1m8
for both 'with_winit isn't responding'
each new 'Failed to read shader' takes 1-10s to appear

m-hugo:
minimizing and reopening it starts it all again except the reappearing triangle can't be moved anymore

m-hugo:
that was on xperia 5, on honor play it's only 5s and the triangle can still be manipulated after a reload

m-hugo:
tried the new changes, still takes 1m10 to load on my xperia5 (at the 'Creating renderer 0' line) but after that all examples work well (the paris map is a bit slow)

my phone has never had any problem with any task i threw at it so not sure why it's 1m slower than the otherwise-slower honor

I also had to add the paris-30k svg manually

```
02-06 21:39:01.874  7184  7246 W winit::platform_impl:..: TODO: forward onStart notification to application
02-06 21:39:01.918  7184  7246 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_DEMAND_REFRESH
02-06 21:39:01.918  7184  7246 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_CONTINUOUS_REFRESH
02-06 21:39:01.919  7184  7246 W wgpu_core::instance: Missing downlevel flags: SURFACE_VIEW_FORMATS
02-06 21:39:01.919  7184  7246 W wgpu_core::instance: The underlying API or device in use does not support enough features to be a fully compliant implementation of WebGPU. A subset of the features can still be used. If you are running this program on native and not in a browser and wish to limit the features you use to the supported subset, call Adapter::downlevel_properties or Device::downlevel_properties to get a listing of the features the currentplatform supports.
02-06 21:39:01.925  7184  7246 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_DEMAND_REFRESH
02-06 21:39:01.925  7184  7246 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_CONTINUOUS_REFRESH
02-06 21:39:01.926  7184  7246 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_DEMAND_REFRESH
02-06 21:39:01.926  7184  7246 W wgpu_hal::vulkan::conv: Unrecognized present mode SHARED_CONTINUOUS_REFRESH
02-06 21:39:01.939  7184  7245 I RustStdoutStderr: Creating renderer 0
02-06 21:39:36.620  7184  7195 I rust.with_wini: Thread[3,tid=7195,WaitingInMainSignalCatcherLoop,Thread*=0xb400007d795e6e90,peer=0x12f80238,"Signal Catcher"]: reacting to signal 3
02-06 21:39:36.621  7184  7195 I rust.with_wini:
02-06 21:39:36.807  7184  7195 I rust.with_wini: Wrote stack traces to tombstoned
02-06 21:40:26.708  7184  7245 I RustStdoutStderr: Rendered svg in 10.781787808s
```

DJMcNab:
@m-hugo is that the logs from the xperia or the honor?

m-hugo:
xperia, I didn't keep logs from the honor
I tried a 3rd phone but it crashes at startup  in the ndk from whichever print statement is responsible for 'RustStdoutStderr: Config: Configuration { ...'

DJMcNab:
Interesting. No idea why the ndk might fail halfway through such a print.

So loading all the shaders takes a minute? That's surprising, although not outside the realm of possibility. I'm not really sure how we can debug that though.

m-hugo:
the xiaomi phone, with ndk patched, it's not your fault it's xiaomi's, runs very well after a 4s startup but spams this between
RustStdoutStderr: Config: Configuration { 
and
RustStdoutStderr: Creating renderer 0

```
21304 21333 E gralloc4: ERROR: Format allocation info not found for format: 38
21304 21333 E gralloc4: ERROR: Format allocation info not found for format: 0
21304 21333 E gralloc4: Invalid base format! req_base_format = 0x0, req_format = 0x38, type= 0x0
21304 21333 E gralloc4: ERROR: Unrecognized and/or unsupported format 0x38 and usage 0xb00
21304 21333 E Gralloc4: isSupported(1, 1, 56, 1, ...) failed with 5
21304 21333 E GraphicBufferAllocator: Failed to allocate (4 x 4) layerCount 1 format 56 usage b00: 5
21304 21333 E AHardwareBuffer: GraphicBuffer(w=4, h=4, lc=1) failed (Unknown error -5), handle=0x0
```

examples/with_winit/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+                                if touch.location.x < render_state.surface.config.width as f64 / 3.
+                                {
+                                    scene_ix = scene_ix.saturating_sub(1);
+                                } else if touch.location.x
+                                    > 2. * render_state.surface.config.width as f64 / 3.
+                                {
+                                    scene_ix = scene_ix.saturating_add(1);
+                                }
+                            }
+                            TouchPhase::Ended | TouchPhase::Cancelled => {
```
DJMcNab, line 140: This needs to be unconditional, i.e. not inside the:
```rust
if touch.location.y > render_state.surface.config.height as f64 - 400. { .. }
```

src/lib.rs
```diff
@@ -54,10 +54,10 @@ pub struct Renderer {
 
 impl Renderer {
     /// Creates a new renderer for the specified device.
-    pub fn new(device: &Device) -> Result<Self> {
+    pub fn new(device: &Device, surface_format: TextureFormat) -> Result<Self> {
```
armansito, line 57: Consider introducing a `RendererOptions` struct and adding `surface_format` as a field there. That would allow us to extend the parameters/options in the future without breaking the interface.

armansito (commented):
The changes overall look good. I only left some minor suggestions though I suggest you rebase the work on top of main.

I haven't been able to get things running on my Android phone but that's because the current Java SDK on macOS story is a pain. I'll try to get set up and test this on my Linux machine again when I get back home later today.

examples/with_winit/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+                                *control_flow = ControlFlow::Exit;
+                            }
+                            _ => {}
+                        }
+                    }
+                }
+                WindowEvent::Touch(touch) => {
+                    match touch.phase {
+                        TouchPhase::Started => {
+                            if touch.location.y > render_state.surface.config.height as f64 - 400. {
```
armansito, line 129: Should this be based on a fraction of the height instead of a hardcoded number?

examples/with_winit/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+                                navigation_fingers.insert(touch.id);
```
armansito, line 130: Please document here that `navigation_fingers` prevents touches to be considered for gesture handling and gestures are intentionally disabled at the bottom margin of the screen as that's reserved for scene navigation.

armansito (commented):
This PR works great on my Pixel 6. Just a couple of additional thoughts:

1. I had to add a `[package.metadata.android.signing.release]` entry to the `Cargo.toml` file with a reference to `$HOME/.android/debug.keystore` for `cargo apk` to work with `--release`. I wonder if the Android build instructions should mention `cargo-apk` manifest metadata as potentially being necessary. Alternately, should the `Cargo.toml` contain some boilerplate `package.metadata.android` entries (for example for the minimum Vulkan version)?

2. Does `cargo apk` support package-specific CLI arguments? The usual `--` delimiter doesn't seem to work for me as it does with `cargo run`. If there is no good way for this that's fine but IWBN to have the ability set an SVG via the CLI. Alternately, maybe we should consider bundling downloaded assets in the APK and add UI to navigate among them. This is more of a suggestion for the future, no need to block this PR on it.

DJMcNab:
On your point 1, cargo apk's intentional design seems to be to not allow optimised builds in the default setup, for unknown reasons. I don't think there's a compatible way to point to the debug keystore path it generates for release mode.

I presume that there's a security risk to just committing a debug keystore here.

I'm not sure what other metadata we can use.


I don't think cargo apk supports cli arguments.

armansito (approved):
I left one small suggestion related to error handling, otherwise looks great. Thanks!

src/lib.rs
```diff
@@ -130,12 +138,16 @@ impl Renderer {
         let mut encoder =
             device.create_command_encoder(&wgpu::CommandEncoderDescriptor { label: None });
         {
+            let blit = self
+                .blit
+                .as_ref()
+                .expect("renderer should have configured surface_format to use on a surface");
```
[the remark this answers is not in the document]
DJMcNab: We could do? I think this is fine panicking, as there's no scenario where this is recoverable.  

The only reasonble way for code to handle this error condition is by panicking themselves. Additionally, there are no suitable implementors of `Error` in scope, so I'd rather defer this until #285 is resolved.

Doing the check early fails the borrow checker, because of the mutable borrow of all of self in `render_to_texture`.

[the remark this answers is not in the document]
armansito: > We could do? I think this is fine panicking, as there's no scenario where this is recoverable.
>
> The only reasonble way for code to handle this error condition is by panicking themselves. Additionally, there are no suitable implementors of Error in scope, so I'd rather defer this until https://github.com/linebender/vello/issues/285 is resolved.

I personally find that libraries should not assert on errors that arise from user/client input and only panic in situations when an internal invariant is violated (i.e. there is some programmer error that we made that led to the panic). I think the client code can panic on this if they wish so (this is also my stance on the overall error handling philosophy within the context of https://github.com/linebender/vello/issues/285). I'm happy to table this until we have a better defined `Error` story and not block this PR on this if you disagree.

> Doing the check early fails the borrow checker, because of the mutable borrow of all of self in render_to_texture, unless we check 'twice'

Ah, that makes sense. Please ignore my suggestion.

armansito (approved):

PR #278, opened 2023-02-10
Lottie animation

dfrg:
This patch adds the `velato` crate which implements a Lottie animation player on top of vello.

The implementation is incomplete but it is capable of rendering a large number of animations. Notable missing features that are on the roadmap:
* Solid color layers
* Trim paths
* Merge paths (pending kurbo pathops!)
* Cubic time mapping and interpolation
* Stars/polygons
* Blur effects (pending support in vello)
* Probably more...

Currently depends on a modified version of the [bodymovin-rs](https://github.com/BrainiumLLC/bodymovin-rs) parser. I believe we'll eventually want a different solution to parsing that is robust in the context of an evolving file format, but this works for now.

To avoid churn and additional merge conflicts, I haven't updated any of the example code to exercise this, though there is example usage in the readme. It's only a few lines of code to get a lottie file into a scene builder.


Cargo.toml
```diff
@@ -2,6 +2,8 @@
 resolver = "2"
 
 members = [
+    "velato",
```
DJMcNab, line 5: Obviously it's open for discussion, but I'd be inclined to put this in the `integrations` folder

velato/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "velato"
+version = "0.1.0"
+edition = "2021"
```
DJMcNab, line 4: This should use the same `workspace.* = true` as the other crates.

DJMcNab (commented):
I haven't been able to run this myself yet, and haven't really looked at the code.

I don't think there's any conflicts likely with `scenes`. That is, I'd like to have a runnable example from this PR. I'd be inclined to add parsing/running lottie files in the positional arguments

velato/src/model/value.rs
```diff
@@ -0,0 +1,160 @@
...
+        self.evaluate_inner(frame).unwrap_or_default()
+    }
+
+    fn evaluate_inner(&self, frame: f32) -> Option<T> {
+        let ([ix0, ix1], t) = Time::frames_and_weight(&self.times, frame)?;
+        Some(self.values.get(ix0)?.lerp(self.values.get(ix1)?, t))
+    }
+}
+
+pub trait Lerp: Clone + Default {
```
derekdreery, line 115: Might be off topic, but could interpolation support be added to kurbo at some point in the future? IIRC it's complicated though.

derekdreery:
Might be nice to include a simple example, so you can do `cargo run --example some_example_name`?

Cargo.toml
dfrg: To me, integration implies glue code for a third party crate which makes sense for svg but not so much for this. I’ll rename the directory to _crates_ and move both there as that seems to be standard practice for this sort of thing.

velato/Cargo.toml
dfrg: Will do, thanks.

velato/src/model/value.rs
dfrg: That’s a good question. I can see animation primitives being in scope for kurbo but as you say, it’s complicated. Lottie in particular makes assumptions about certain things that may not be generally applicable.

For example, this code currently uses Rect and Ellipse from kurbo to represent those shapes but that will have to change when trim path support is added because the expectation is that those paths originate at top center (and can be cw or ccw ordered) which is not currently the case in kurbo.

dfrg:
I changed the catch all `svgs` command line argument in the winit demo to accept both svg and lottie json files and added an `ExampleScene` generator for the latter so `cargo run -p winit -- path/to/lottie.json` will now work.

Cargo.toml
DJMcNab: `crates` to me suggests the opposite dependency flow, i.e. that `crates` are dependencies of the top level crate, not the other way around.

I chose the name integrations for this directory as that maintains the feeling of the crates being dependent on the top level crate. (I'd also claim that this is an integration with `lottie`)

Maybe there's a third name which captures that?

crates/velato/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+categories = ["rendering", "graphics"]
+keywords = ["2d", "vector-graphics", "animation", "lottie"]
+
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+vello = { path = "../../" }
+bodymovin = { git = "https://github.com/dfrg/bodymovin-rs", rev= "c156dda9" }
```
lemmih, line 14: Did you fork `bodymovin-rs` because the upstream repository is unmaintained? Do you plan on taking over the `bodymovin` crate?

lemmih: Ah, missed your explanation in the PR description.

derekdreery:
> I changed the catch all `svgs` command line argument in the winit demo to accept both svg and lottie json files and added an `ExampleScene` generator for the latter so `cargo run -p winit -- path/to/lottie.json` will now work.

Could you include a very basic lottie json, so the user can run an example without having to find their own?

DJMcNab:
I've really struggled to find suitably licensed lottie files; however, the files in https://github.com/CommunityToolkit/Lottie-Windows/tree/main/LottieSamples/Assets *appear* to be MIT licensed.

I tried https://github.com/CommunityToolkit/Lottie-Windows/blob/main/LottieSamples/Assets/LottieLogo1.json, which gave incorrect rendering

derekdreery:
Aah I see. If there are licensing issues then I don't think it's super important. Perhaps someone could make an animation later (explaining vello 👀) and add it to the project.

DJMcNab:
There are other files in the examples repository; I might have also tried the light bulb document, which might have worked, although it is a bit dull. It would be good to have a thematic example, although I'm not the person to make such a thing.

derekdreery:
I don't think having an example lottie file should block this PR. Let's merge this, then perhaps open an issue for someone to find or create a lottie file with the right licensing.

DJMcNab:
We've chosen to move this to it's own repository, i.e. [linebender/velato](https://github.com/linebender/velato). Go give it some love 🎉 

This allows greater flexibility in making changes there, without compromising the review standards of the Vello repo.

PR #279, opened 2023-02-10
Fix include_str! compile error in examples/scenes.

jasondavies:
I don't think `CARGO_MANIFEST_DIR` should ever end in "/", though I've only checked on Linux and macOS.

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -243,7 +243,7 @@ fn included_tiger() -> impl FnMut(&mut SceneBuilder, &mut SceneParams) {
         let (scene_frag, resolution) = cached_scene.get_or_insert_with(|| {
             let contents = include_str!(concat!(
                 env!("CARGO_MANIFEST_DIR"),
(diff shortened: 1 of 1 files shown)


jasondavies:
Full error trace, when running `cargo run_wasm --package with_winit`:

```error: couldn't read /Users/jason/src/vello/examples/scenes../../assets/Ghostscript_Tiger.svg: No such file or directory (os error 2)
   --> examples/scenes/src/test_scenes.rs:244:28
    |
244 |               let contents = include_str!(concat!(
    |  ____________________________^
245 | |                 env!("CARGO_MANIFEST_DIR"),
246 | |                 "../../assets/Ghostscript_Tiger.svg"
247 | |             ));
    | |______________^
    |
    = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `scenes` due to previous error```

DJMcNab (approved):
This is already fixed in vello#273, but I'm not sure how long that will take to be reviewed.

Thanks!

Issue #281, opened 2023-02-21
Weird behaviour with rotating gradients


DJMcNab:
At the moment, implementing transformations on scenes with certain gradients has strange behaviour, where the gradient appears to rotate in the opposite direction to the rest of the scene.

At the moment, this is most easily reproduced in #273

PR #282, opened 2023-02-22
Add a keyboard key to rotate the scene

DJMcNab:
<kbd>Q</kbd> rotates the scene anticlockwise, <kbd>E</kbd> rotates the scene clockwise, about the cursor position

This is useful for testing #281

It's intentionally non-framerate independent, both to keep the results consistent, and to keep the code simple.


armansito (approved):
Rotating around the cursor position is a nice touch!

PR #283, opened 2023-02-23
Fix brush transforms

### LINKED ISSUE #281: Weird behaviour with rotating gradients

dfrg:
This fixes an incorrect application of the inverse transform for radial gradients in fine.

Also fixes an edge case in `SceneBuilder` where supplying a brush transform that is identical to the path transform generates a corrupt encoding.

Discovery of both done by @DJMcNab on [this zulip thread](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Radial.20Blending.20Behaviour.20-.20vello.23281).

Should fix #281

### DIFF
--- a/shader/draw_leaf.wgsl
+++ b/shader/draw_leaf.wgsl
@@ -151,11 +151,7 @@ fn main(
                 let r1 = bitcast<f32>(scene[dd + 6u]);
                 let inv_det = 1.0 / (matrx.x * matrx.w - matrx.y * matrx.z);
                 let inv_mat = inv_det * vec4(matrx.w, -matrx.y, -matrx.z, matrx.x);
-                var inv_tr = inv_det * vec2(
-                    matrx.z * translate.y - matrx.w * translate.x,
-                    matrx.y * translate.x - matrx.x * translate.y,
-                );
-                inv_tr -= p0;
+                let inv_tr = mat2x2(inv_mat.xy, inv_mat.zw) * -translate - p0;
                 let center1 = p1 - p0;
                 let rr = r1 / (r1 - r0);
                 let ra_inv = rr / (r1 * r1 - dot(center1, center1));
--- a/shader/draw_leaf.wgsl
+++ b/shader/draw_leaf.wgsl
@@ -151,9 +151,9 @@ fn main(
                 let r1 = bitcast<f32>(scene[dd + 6u]);
                 let inv_det = 1.0 / (matrx.x * matrx.w - matrx.y * matrx.z);
                 let inv_mat = inv_det * vec4(matrx.w, -matrx.y, -matrx.z, matrx.x);
-                var inv_tr = vec2(
-                    inv_det * (matrx.z * translate.y - matrx.w * translate.x),
-                    inv_det * (matrx.y * translate.x - matrx.x * translate.y),
+                var inv_tr = inv_det * vec2(
+                    matrx.z * translate.y - matrx.w * translate.x,
+                    matrx.y * translate.x - matrx.x * translate.y,
                 );
                 inv_tr -= p0;
                 let center1 = p1 - p0;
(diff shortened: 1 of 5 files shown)


src/encoding/encoding.rs
```diff
@@ -120,10 +120,16 @@ impl Encoding {
     }
 
     /// Encodes a transform.
-    pub fn encode_transform(&mut self, transform: Transform) {
+    ///
+    /// If the given transform is different from the current one, encodes it an
```
armansito, line 124: encodes it an -> encodes it *and*

dfrg: Oops. Fixed, ty.

shader/fine.wgsl
```diff
@@ -251,7 +251,7 @@ fn main(
                 for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
                     let my_xy = vec2(xy.x + f32(i), xy.y);
                     // TODO: can hoist y, but for now stick to the GLSL version
-                    let xy_xformed = rad.matrx.xz * my_xy.x + rad.matrx.yw * my_xy.y - rad.xlat;
+                    let xy_xformed = rad.matrx.xy * my_xy.x + rad.matrx.zw * my_xy.y - rad.xlat;
```
armansito, line 254: When I tested this change locally as is it didn't fix the broken rotation for me on the `blend_grid` test (you can reproduce it with the new Q/E controls that @DJMcNab added).

However I was able to fix it by reverting your change to fine.wgsl back to the original and applying @DJMcNab's suggestion on zulip to change [shaders/draw_leaf.wgsl#153](https://github.com/linebender/vello/blob/main/shader/draw_leaf.wgsl#L153) to:
```
let inv_mat = inv_det * vec4(matrx.w, matrx.y, matrx.z, matrx.x);
```

dfrg: The fix in this PR comes from experimenting with radial gradient rotation in general.

I used this codepen as reference: https://codepen.io/dfrgy/pen/KKxMpWj and added a corresponding example to the brush transform test scene. Without these changes, the transform is very wonky.

The original inverse calculation is algebraically correct so I'm assuming the problem is elsewhere but I haven't been able to track down the source of the reversed rotation direction in the blend grid. I'll keep investigating.

dfrg: Looks like the inverse of the translation components was incorrect. This seems to be working for all cases now.

dfrg: Or not...

dfrg: Ok, good to go :)

shader/draw_leaf.wgsl
```diff
@@ -151,8 +151,11 @@ fn main(
                 let r1 = bitcast<f32>(scene[dd + 6u]);
                 let inv_det = 1.0 / (matrx.x * matrx.w - matrx.y * matrx.z);
                 let inv_mat = inv_det * vec4(matrx.w, -matrx.y, -matrx.z, matrx.x);
-                var inv_tr = inv_mat.xz * translate.x + inv_mat.yw * translate.y;
-                inv_tr += p0;
+                var inv_tr = inv_det * vec2(
```
armansito, line 154: Just to confirm my understanding, the 2x2 matrix is stored in column-major order in the vector, correct?

A slightly shorter formulation that reuses `inv_mat` and built-in matrix multiplication could look like:
```
let inv_tr = mat2x2<f32>(inv_mat.xy, inv_mat.zw) * (-translate) - p0;
```
which has the added benefit of declaring `inv_tr` as immutable.

Feel free to push back if you disagree (or if there's a performance assumption that I'm missing here). Otherwise the PR overall lgtm.

armansito (commented):
I tested the changes locally and it all looks good! I left one question/suggestion for readability in `shader/draw_leaf.wgsl`. Feel free to push back if you don't agree.

Otherwise lgtm.

shader/draw_leaf.wgsl
dfrg: I think the lack of matrix types in the codebase is simply a tradition at this point. I like this formulation and we should probably do more of it moving forward. Updated!

armansito:
Thank you for the back and forth!

armansito (approved):

PR #284, opened 2023-02-24
Glyph run API

dfrg:
This patch implements a comprehensive glyph run API to replace the ad hoc fragment construction we currently use.

My initial intent was to build the front end only, loading glyphs inline in `SceneBuilder` but I had the motivation to go further and wanted to implement something that could potentially be used by Lapce without serious performance regressions due to the loss of all caching. 

In this implementation, the glyph run data is packed into several new arrays in the `Encoding` structure while also making use of the same "patching" strategy as gradient ramps. In the `Resolver` (previously `ResourceCache`), glyph outlines are extracted from the font, encoded and cached. During this process, a sequence of resolved patches is collected and size diffs for all of the streams are calculated. Following this, the full packed buffer size is computed and the streams are processed in order handling all resolved patches by taking data from the glyph cache and splicing it in at the appropriate places. This ensures that data is written contiguously and in order supporting the eventual use case of packing directly into a mapped buffer.

This diff is significantly larger than I intended (apologies!), but the consolation prize is that it lays the groundwork for rectangles, images and glyph atlases, all of which require more sophisticated splicing of the packed scene.

For those purely interested in the API, there is a gist [here](https://gist.github.com/dfrg/59cfbfea9a6a7a85e7ca4f26cf33ccf0) that contains the broad strokes and an example of how you might make use of it with a parley layout.


src/encoding/encoding.rs
```diff
@@ -248,15 +298,19 @@ impl Encoding {
...
-        self.n_paths += 1;
-        self.n_clips += 1;
+        if self.n_open_clips > 0 {
+            self.draw_tags.push(DrawTag::END_CLIP);
+            // This is a dummy path, and will go away with the new clip impl.
+            self.path_tags.push(PathTag::PATH);
+            self.n_paths += 1;
+            self.n_clips += 1;
+            self.n_open_clips -= 1;
+        }
```
DJMcNab, line 313: Should this else panic?

examples/scenes/src/simple_text.rs
```diff
@@ -31,12 +34,75 @@ const FONT_DATA: &[u8] = include_bytes!("../../assets/roboto/Roboto-Regular.ttf"
...
+                    .unwrap_or(0);
+                let mut pen_x = 0f64;
+                builder
+                    .draw_glyphs(&self)
+                    .font_size(size)
+                    .transform(transform)
+                    .glyph_transform(Some(Affine::new([
+                        1.,
+                        0.,
+                        20f64.to_radians().tan(),
```
DJMcNab, line 78: This could do with a comment; I'm struggling to understand the context.

Unless it's just a testing change to make all text faux-italicised?

DJMcNab (commented):
I haven't run the code, but it all looks reasonable to me.

The data flows aren't especially clear to me, but it does seem that the timings are as expected, i.e. Glyphs are resolved as late as possible (unless you manually ask for them to be resolved earlier?)

src/encoding/encoding.rs
dfrg: In general, I don’t think libraries should ever intentionally panic though I would be open to returning an error here.

examples/scenes/src/simple_text.rs
dfrg: Yes, it’s just a manual skew to test that the per glyph transform works correctly. When I get a chance, I’ll write up a patch to add this to kurbo. In the meantime, I’ll tack a comment on here.

src/encoding/encoding.rs
DJMcNab: I don't think this is a recoverable error, which is why I suggest a panic. Your use of the API is wrong if you ever reach this case. Maybe I'm misunderstanding

src/encoding/resolve.rs
```diff
@@ -0,0 +1,463 @@
...
+                    let mut transform = run.transform;
+                    if hint {
+                        // If hinting was requested and our transform matrix is just a uniform
+                        // scale, then adjust our font size and cancel out the matrix. Otherwise,
+                        // disable hinting entirely.
+                        if transform.matrix[0] == transform.matrix[3]
+                            && transform.matrix[1] == 0.0
+                            && transform.matrix[2] == 0.0
+                        {
+                            font_size *= transform.matrix[0];
```
DJMcNab, line 377: Can this also occur if the transform is a scale + rotation (i.e. `transform.matrix[1]==-transform.matrix[2]`?). Hinting (likely?) still needs to be disabled in that case.

src/encoding/encoding.rs
dfrg: It’s recoverable in the sense that we can avoid producing a corrupt encoding albeit likely with incorrect rendering.

I think it’s less misunderstanding and more difference of perspective. Since vello is likely to sit at a lower level of the stack, my preference is that malformed (potentially user generated) content should not kill the whole process without warning.

src/encoding/resolve.rs
dfrg: I believe it already rejects that case? Hinting should only be enabled if the diagonal components are equal and the others are zero. If the logic doesn’t capture that then it’s a bug.

We can potentially handle non-uniform scales as well if we’re only hinting in the Y direction (which is common), but I’ll leave that discussion for later.

DJMcNab: We've clarified this on Zulip; my understanding is that this translation would be valid; however it's equivalent to what we have now, so doesn't matter.

src/encoding/encoding.rs
DJMcNab: I think the error should be reported somehow; if your user controlled data allows this case to happen, that's a bug in your translation of the user data.

Returning an error here would be awkward for consumers; it's the only place such an error can occur - maybe poisoning the `Scene/SceneBuilder` is good enough; panicing if there's such an error on dropping the builder, unless it was handled properly?

dfrg: Looking at other implementations: piet (and old piet-gpu) return an error in this case, but allow you to continue. Skia, notably, just [ignores](https://github.com/google/skia/blob/ad6b08ad714c22adbcee9f92f03e75949aaa925e/src/core/SkCanvas.cpp#L651) a restore operation if there are no saved layers on the stack.

Pragmatically, there are potential consumers of vello for which a bug in content translation causing a crash is not an acceptable result and we are subject to that constraint even if I conceptually agree with you on ensuring correctness.

Broader in scope, general robustness and error handling are things we want to tackle in a principled way. I think you mentioned this on the zulip and it's probably worth filing an issue to continue discussion.

examples/scenes/src/simple_text.rs
dfrg: Oh, I didn't realize this was in `simple_text.rs` and not in the test scene! Good catch. I'll shuffle it around.

src/encoding/resolve.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    /// Returns the path tag stream in chunks of 4.
+    pub fn path_tags_chunked<'a>(&self, data: &'a [u8]) -> &'a [u32] {
+        let start = self.path_tag_base as usize * 4;
+        let end = self.path_data_base as usize * 4;
+        bytemuck::cast_slice(&data[start..end])
+    }
+
+    /// Returns the path data stream.
```
raphlinus, line 76: This feels like it needs a caveat, it's only valid if all the paths encoded are in f32 format.

src/encoding/encoding.rs
```diff
@@ -14,15 +14,15 @@
...
+    resolve::Patch, DrawColor, DrawLinearGradient, DrawRadialGradient, DrawTag, Glyph, GlyphRun,
+    PathEncoder, PathTag, Transform,
 };
 
-use peniko::{kurbo::Shape, BlendMode, BrushRef, Color, ColorStop, Extend, GradientKind};
+use peniko::{kurbo::Shape, BlendMode, BrushRef, ColorStop, Extend, GradientKind};
 
 /// Encoded data streams for a scene.
-#[derive(Default)]
+#[derive(Clone, Default)]
```
raphlinus, line 25: I've got my head pretty much around this now, but it's reminding me that we should have a writeup of how these pieces fit together. There's a higher level theory about how glyph runs contain splice points referencing the main encodings, and also lower level details that are significant: for example, conceptually each glyph run contains a sequence of glyphs, but in practice the storage for those is flattened into a single vec. Perhaps this is the proper scope for the big Vello writeup, perhaps we should do something before then.

I'm also aware that the details will change when we get images done.

raphlinus (approved):
This looks like a significant step forward. I can see nontrivial work ahead of us to get other resources (images and cached glyphs) added, but I also don't see any way around that. Also we should keep an eye on the performance costs in the resolve step, but I'm satisfied with the current implementation - it seems like probably the simplest thing that implements the desired semantics.

src/encoding/glyph_cache.rs
```diff
@@ -0,0 +1,87 @@
...
+        &mut self,
+        key: GlyphKey,
+        style: &Style,
+        scaler: &mut GlyphProvider,
+    ) -> Option<CachedRange> {
+        // For now, only cache non-zero filled glyphs so we don't need to keep style
+        // as part of the key.
+        let is_nz_fill = matches!(style, Style::Fill(Fill::NonZero));
+        if is_nz_fill {
+            if let Some(range) = self.glyphs.get(&key) {
```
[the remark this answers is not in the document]
dfrg: Updated to use the entry API, but I expect this method and the caching strategy will see some changes soon. I'd like to do a better job on this all around, but since this state is currently just dropped at the end of render, for now I focused on doing the minimum to cache encoded glyphs per frame.

src/encoding/resolve.rs
dfrg: Oops. Good catch. I just changed this to return bytes as that's the best we can do without further context. I guess we'll need to pair this up with path tags to actually extract useful data.

src/encoding/encoding.rs
dfrg: This makes sense. I have a bit of a document queue right now, but I'll try to at least put together some loose notes about this to capture what's happening.

Issue #276, opened 2023-02-07
WASM crashes


simbleau:
So I made sure I was on wgpu 15 and using the most recent version.

I cannot seem to make the run_wasm work. On the vello example, I see a problem with adapter.features() from webgpu causing a crash. Doesn't work on canary or nightly.

<img width="572" alt="image" src="https://user-images.githubusercontent.com/48108917/217147591-6b799d48-68f4-4ecc-bfab-44643164a776.png">

In my personal project I get a bug saying there's no way to allocate a storage buffer. Unrelated, but just letting you know. Perhaps this could be a 2 birds with 1 stone issue.

<img width="551" alt="image" src="https://user-images.githubusercontent.com/48108917/217133107-56447471-15fa-42b9-8b5f-7fdebf7c970b.png">

raphlinus:
That particular error message makes it look like it's getting `Limits::downlevel_defaults` which will not work, at least without doing considerable additional compatibility work.

armansito:
> So I made sure I was on wgpu 15 and using the most recent version.
> 
> I cannot seem to make the run_wasm work. On the vello example, I see a problem with adapter.features() from webgpu causing a crash. Doesn't work on canary or nightly.
> 

The WASM crash is due to a known issue in wgpu which should get fixed by https://github.com/gfx-rs/wgpu/pull/3428.

simbleau:
> That particular error message makes it look like it's getting `Limits::downlevel_defaults` which will not work, at least without doing considerable additional compatibility work.

I am getting `Limits::default`. What limits should I need for Vello?

My code:
```

    /// Creates a compatible device handle id.
    async fn new_device(&mut self, compatible_surface: Option<&Surface>) -> Option<usize> {
        let adapter = wgpu::util::initialize_adapter_from_env_or_default(
            &self.instance,
            wgpu::Backends::PRIMARY,
            compatible_surface,
        )
        .await?;
        let features = adapter.features();
        let limits = Limits::default();
        let (device, queue) = adapter
            .request_device(
                &wgpu::DeviceDescriptor {
                    label: None,
                    features: features
                        & (wgpu::Features::TIMESTAMP_QUERY | wgpu::Features::CLEAR_TEXTURE),
                    limits,
                },
                None,
            )
            .await
            .ok()?;
        let device_handle = DeviceHandle {
            adapter,
            device,
            queue,
        };
        self.devices.push(device_handle);
        Some(self.devices.len() - 1)
    }
```

[...]

PR #287, opened 2023-03-02
Add a base-color uniform

armansito:
This PR introduces a clear-color to apply as the base color for blends in the fine rasterizer stage. This uniform only has an effect in the full fine pipeline.

This PR also includes a minor reorganization of render-time parameters (the new clear color, target image widht/height) in a new `RenderParams` struct and updates the examples accordingly.

### DIFF
--- a/examples/headless/src/main.rs
+++ b/examples/headless/src/main.rs
@@ -96,6 +96,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
         time: args.time.unwrap_or(0.),
         text: &mut text,
         resolution: None,
+        base_color: None,
     };
     (example_scene.function)(&mut builder, &mut scene_params);
     builder.finish();
@@ -130,6 +131,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
         base_color: args
             .args
             .get_base_color()?
+            .or(scene_params.base_color)
             .unwrap_or(vello::peniko::Color::BLACK),
         width,
         height,
--- a/examples/headless/src/main.rs
+++ b/examples/headless/src/main.rs
@@ -130,7 +130,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
     let render_params = vello::RenderParams {
         base_color: args
             .args
-            .get_base_color()?
+            .base_color
             .or(scene_params.base_color)
             .unwrap_or(vello::peniko::Color::BLACK),
         width,
--- a/examples/headless/src/main.rs
+++ b/examples/headless/src/main.rs
@@ -126,6 +126,11 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
             (Some(x), Some(y)) => (x.try_into()?, y.try_into()?),
         }
     };
+    let params = vello::RenderParams {
+        clear_color: vello::peniko::Color::BLACK,
+        width,
+        height,
+    };
     let mut scene = Scene::new();
     let mut builder = SceneBuilder::for_scene(&mut scene);
     builder.append(&fragment, Some(transform));
@@ -147,7 +152,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
     });
     let view = target.create_view(&wgpu::TextureViewDescriptor::default());
     renderer
-        .render_to_texture(&device, &queue, &scene, &view, width, height)
+        .render_to_texture(&device, &queue, &scene, &view, &params)
         .or_else(|_| bail!("Got non-Send/Sync error from rendering"))?;
     // (width * 4).next_multiple_of(256)
     let padded_byte_width = {
--- a/examples/headless/src/main.rs
+++ b/examples/headless/src/main.rs
@@ -92,15 +92,15 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
     let mut builder = SceneBuilder::for_fragment(&mut fragment);
     let example_scene = &mut scenes.scenes[index];
     let mut text = SimpleText::new();
-    let mut params = SceneParams {
+    let mut scene_params = SceneParams {
         time: args.time.unwrap_or(0.),
         text: &mut text,
         resolution: None,
     };
-    (example_scene.function)(&mut builder, &mut params);
+    (example_scene.function)(&mut builder, &mut scene_params);
     builder.finish();
     let mut transform = Affine::IDENTITY;
-    let (width, height) = if let Some(resolution) = params.resolution {
+    let (width, height) = if let Some(resolution) = scene_params.resolution {
         let ratio = resolution.x / resolution.y;
         let (new_width, new_height) = match (args.x_resolution, args.y_resolution) {
             (None, None) => (resolution.x.ceil() as u32, resolution.y.ceil() as u32),
@@ -126,8 +126,11 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
             (Some(x), Some(y)) => (x.try_into()?, y.try_into()?),
         }
     };
-    let params = vello::RenderParams {
-        clear_color: vello::peniko::Color::BLACK,
+    let render_params = vello::RenderParams {
+        base_color: args
+            .args
+            .get_base_color()?
+            .unwrap_or(vello::peniko::Color::BLACK),
         width,
         height,
     };
@@ -152,7 +155,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
     });
     let view = target.create_view(&wgpu::TextureViewDescriptor::default());
     renderer
-        .render_to_texture(&device, &queue, &scene, &view, &params)
+        .render_to_texture(&device, &queue, &scene, &view, &render_params)
         .or_else(|_| bail!("Got non-Send/Sync error from rendering"))?;
     // (width * 4).next_multiple_of(256)
     let padded_byte_width = {
--- a/examples/scenes/src/lib.rs
+++ b/examples/scenes/src/lib.rs
@@ -19,6 +19,7 @@ pub struct SceneParams<'a> {
     pub time: f64,
     pub text: &'a mut SimpleText,
     pub resolution: Option<Vec2>,
+    pub base_color: Option<vello::peniko::Color>,
 }
 
 pub struct SceneConfig {
--- a/examples/scenes/src/lib.rs
+++ b/examples/scenes/src/lib.rs
@@ -5,7 +5,7 @@ mod svg;
 mod test_scenes;
 use std::path::PathBuf;
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use clap::{Args, Subcommand};
 use download::Download;
 pub use simple_text::SimpleText;
@@ -13,7 +13,7 @@ pub use simple_text::SimpleText;
 pub use svg::{default_scene, scene_from_files};
 pub use test_scenes::test_scenes;
 
-use vello::{kurbo::Vec2, SceneBuilder};
+use vello::{kurbo::Vec2, peniko::Color, SceneBuilder};
 
 pub struct SceneParams<'a> {
     pub time: f64,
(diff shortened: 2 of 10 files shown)


examples/with_winit/src/main.rs
```diff
@@ -49,10 +53,51 @@ struct Args {
...
+        value_parser = PossibleValuesParser::new(["black", "white", "aquamarine", "crimson"])
+            .map(|s| s.parse::<ClearColor>().unwrap())
+    )]
+    clear_color: ClearColor,
     #[command(flatten)]
     args: scenes::Arguments,
 }
 
+#[derive(Debug, Clone)]
+enum ClearColor {
```
dfrg, line 68: The peniko `Color` type has a [parse](https://github.com/linebender/peniko/blob/9526fb1afa85493cf537afac76c1e38dc7c76d23/src/color.rs#L128) method that supports hex values and svg color names. Can we make use of that here instead of a new enum with a limited set of colors?

dfrg (commented):
The plumbing and GPU side all looks good! Added a note inline about potentially reusing peniko color parsing for the CLI arg.

src/lib.rs
```diff
@@ -52,8 +52,19 @@ pub struct Renderer {
...
+    pub clear_color: peniko::Color,
+
+    /// Dimensions of the rasterization target
+    pub width: u32,
+    pub height: u32,
+}
+
 impl Renderer {
-    /// Creates a new renderer for the specified device.
+    /// Creates a new renderer for the specified device with default options.
```
DJMcNab, line 67: `Renderer` is never created without 'default options' as far as I can see.

src/render.rs
```diff
@@ -61,6 +62,7 @@ struct Config {
     height_in_tiles: u32,
     target_width: u32,
     target_height: u32,
+    clear_color: u32,
```
DJMcNab, line 65: `struct Config` (Rust) and `struct Config` (wgsl) need comments specifying that they must be kept in sync in the correct order.

This is as good a time as any to add them

examples/with_winit/src/main.rs
```diff
@@ -49,10 +53,51 @@ struct Args {
     /// Switch between scenes with left and right arrow keys
     #[arg(long)]
     scene: Option<i32>,
+    #[arg(
+        long,
+        default_value_t = ClearColor::Black,
+        value_parser = PossibleValuesParser::new(["black", "white", "aquamarine", "crimson"])
+            .map(|s| s.parse::<ClearColor>().unwrap())
+    )]
+    clear_color: ClearColor,
```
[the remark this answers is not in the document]
DJMcNab: A scene which cycles through all clear colours of a specific luminance/saturation would be cool, although that seems equivalent to just doing the same thing with a normal fill.

examples/with_bevy/src/main.rs
```diff
@@ -46,16 +46,19 @@ fn render_scenes(
 ) {
     for scene in &mut scenes {
         let gpu_image = gpu_images.get(&scene.1).unwrap();
-
+        let params = vello::RenderParams {
+            clear_color: vello::peniko::Color::AQUAMARINE,
```
DJMcNab, line 50: Is it a clear colour or a base colour? For me, clear colour suggests an extra step in the pipeline, running a clearing step.

It's probably a distinction without a difference

examples/with_winit/src/main.rs
```diff
@@ -49,10 +53,51 @@ struct Args {
     /// Switch between scenes with left and right arrow keys
     #[arg(long)]
     scene: Option<i32>,
+    #[arg(
```
DJMcNab, line 56: This is missing some docs

It can probably also reasonably move to `scenes`. It seems equally applicable to `headless`

DJMcNab (commented):
Seems like a useful little feature, just some peripheral concerns.

examples/with_bevy/src/main.rs
armansito: I certainly had the same question as I settled on clear color, as it is slightly misleading if you think of it as the clear blit in a render pass.  Base color sounds good to me.

examples/with_winit/src/main.rs
armansito: Great! Done.

examples/with_winit/src/main.rs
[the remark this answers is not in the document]
armansito: > This should be an Option, because scenes should be able to set the clear colour if not specified by the user.

Good idea! I made this an optional parameter and made it the responsibility of the examples to supply a default color.

> A scene which cycles through all clear colours of a specific luminance/saturation would be cool, although that seems equivalent to just doing the same thing with a normal fill.

I thought of this, though since I defined this as a render parameter (rather than a scene parameter) this would require changing the test_scenes plumbing a little bit so that render options can be modified and animated by the test scenes. Which is OK, and I'll think about it as a follow up.

I considered making this a scene parameter as well but I wasn't sure whether this belongs in the scene or as a renderer option. Maybe there is value in specifying a background color at the scene level; I don't feel strongly either way.

@dfrg @raphlinus Any thoughts?

examples/with_winit/src/main.rs
armansito: Good idea! Done.

src/lib.rs
armansito: Ah this comment is leftover from the first patch where I had the options at the Renderer level. Removed.

src/render.rs
armansito: Good point! Done.

armansito:
Thank you for the feedback everyone. I pushed a new commit and responded inline to all comments. Let me know if I missed anything!

src/encoding/packed.rs
```diff
@@ -48,7 +48,8 @@ pub struct Layout {
     pub linewidth_base: u32,
 }
 
-/// Scene configuration.
+/// Scene configuration. This data structure must be kept in sync with the definition in
+/// shaders/shared/config.wgsl.
```
DJMcNab, line 52: I think we could do with a "backlink" here, i.e. an equivalent comment referring to this file in `shaders/shared/config.wgsl`

src/render.rs
```diff
@@ -54,13 +55,16 @@ const BIN_HEADER_SIZE: u64 = 8;
 const TILE_SIZE: u64 = 8;
 const SEGMENT_SIZE: u64 = 24;
 
+// This data structure must be kept in sync with encoding::Config and the definition in
+// shaders/shared/config.wgsl.
```
DJMcNab, line 59: Oh, ouch. The synchronisation is in three places?

DJMcNab (commented):
Looks reasonable to me, presuming it runs correctly on web (which I can't test)

src/render.rs
dfrg: The one in render.rs should go away. I’ll do a cleanup pass soon.

dfrg (approved):
This looks good to me. Thanks!

src/render.rs
dfrg: The config struct in packed.rs will also move in the glyph run PR so I’ll fix up the comment there. I think we should land this as is.

examples/with_winit/src/main.rs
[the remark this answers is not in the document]
DJMcNab: > Anything that can be parameterized wrt to scene probably should so I prefer the current state.

My proposal would be that the priority is:
1. Provided by user on command line
2. Provided by scene
3. Default for renderer (effectively always black)

The option in 2 is mostly useful for a theoretical future where we use `test_scenes` for e.g. CI tests.
In that case, we'd want to validate that base colours works correctly (e.g. correct colour space/alpha sent through, etc.). So a scene which changes the base colour would therefore make sense.

OTOH, the current state is also fine, as implementing this scene parameter in such a case would be trivial.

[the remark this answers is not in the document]
armansito: I just pushed a new commit that adds a new scene parameter and a test scene that animates a CIE HCL color (`peniko::Color` had a constructor for this, so I just went with it). I implemented the precedence that you specified. Please try it out.

I'll wait for your approval before I merge the change.

src/encoding/packed.rs
armansito: I'll leave things as is per @dfrg since these types are going to be moving around.

src/render.rs
armansito: sgtm

armansito:
> Looks reasonable to me, presuming it runs correctly on web (which I can't test)

Unfortunately I can't test this on the web either on the current HEAD due to the issues in https://github.com/linebender/vello/issues/276  😕

src/lib.rs
```diff
@@ -115,7 +128,7 @@ impl Renderer {
         if target.width != width || target.height != height {
             target = TargetTexture::new(device, width, height);
         }
-        self.render_to_texture(device, queue, scene, &target.view, width, height)?;
+        self.render_to_texture(device, queue, scene, &target.view, &params);
```
DJMcNab, line 131:
```suggestion
        self.render_to_texture(device, queue, scene, &target.view, &params)?;
```

examples/headless/src/main.rs
```diff
@@ -92,15 +92,15 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
     let mut builder = SceneBuilder::for_fragment(&mut fragment);
     let example_scene = &mut scenes.scenes[index];
     let mut text = SimpleText::new();
-    let mut params = SceneParams {
+    let mut scene_params = SceneParams {
         time: args.time.unwrap_or(0.),
         text: &mut text,
         resolution: None,
```
DJMcNab, line 98: This doesn't compile. That is, `headless` also needs to handle the case where this is set in the scene.

Hopefully we can get #250 sorted soon.

src/lib.rs
```diff
@@ -103,9 +113,12 @@ impl Renderer {
         queue: &Queue,
         scene: &Scene,
         surface: &SurfaceTexture,
+        params: &RenderParams,
         width: u32,
         height: u32,
```
DJMcNab, line 118:
```suggestion
```

DJMcNab (commented):
I'm happy for this landing once `headless` compiles and the new warnings are dealt with

examples/headless/src/main.rs
armansito: Oops! Done.

src/lib.rs
armansito: Done.

msiglreith:
Would it make sense to rather read the values from the framebuffer on the beginning of the fine renderer stage? The initial color would be then set via standard clear operations or could be also some already rendered content.

examples/scenes/src/lib.rs
```diff
@@ -46,6 +47,12 @@ pub struct Arguments {
     #[arg(help_heading = "Scene Selection", global(false))]
     /// The svg files paths to render
     svgs: Option<Vec<PathBuf>>,
+    #[arg(help_heading = "Render Parameters")]
+    #[arg(long, global(false))]
+    /// The base color applied as the blend background to the rasterizer.
+    /// Format is CSS style hexidecimal (#RGB, #RGBA, #RRGGBB, #RRGGBBAA) or
```
DJMcNab, line 53: The one issue with this is that in powershell, `#` starts a comment. For example:
```powershell
cargo run ... --base-color #bed
```
needs to be written as (with the `` ` `` as the escape character)
```powershell
cargo run ... --base-color `#bed
```

That's not really the fault of this PR though. I don't think we need to add a comment here

DJMcNab (approved):
Nothing blocking, but some nice things to clean up. Note that we follow the policy 'documented' in https://github.com/linebender/druid/pull/2357#issuecomment-1426814447 here, so you should be the one to merge this PR.

examples/with_winit/src/main.rs
```diff
@@ -245,6 +262,7 @@ fn main() -> Result<()> {
     env_logger::init();
     let args = Args::parse();
     let scenes = args.args.select_scene_set(|| Args::command())?;
+    let base_color = args.args.get_base_color()?;
```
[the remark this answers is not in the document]
armansito: I wasn't too happy with this. The only reason I put it here was to propagate the error but it's probably not egregious to just unwrap the result inside `run`.

armansito:
> Would it make sense to rather read the values from the framebuffer on the beginning of the fine renderer stage? The initial color would be then set via standard clear operations or could be also some already rendered content.

This is a neat idea and something I have considered. I personally find that the fine rasterization stage having its own base blend color is the right abstraction and generally works for the headless case as well (where there may be no clear color or the client maybe simply starts with an empty texture).

What would be nice as a future improvement is the ability to set the base color for fine with a 0 alpha value and have the blit pass perform an alpha blend over the existing contents of the target texture.

examples/scenes/src/lib.rs
armansito: I would agree that this is a shell-specific concern.

examples/with_winit/src/main.rs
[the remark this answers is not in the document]
armansito: > I just want to check ,clap doesn't handle Option<Color> automatically, does it? I think they use FromStr, so it might just work.

Actually this does work pretty well. I used `value_parser` instead of FromStr though I imagine that would work too. The output looks even nicer now. PTAL.

Issue #266, opened 2023-01-25
Winit example crashes on Linux with AMD GPU


BillyDM:
The `with_winit` example crashes on my system when I try to run it. The `with_bevy` example works fine though.

OS: Arch Linux
GPU: AMD RX 570

backtrace:
```
[examples/with_winit/src/hot_reload.rs:19] &Path::new(env!("CARGO_MANIFEST_DIR")).join("../../shader").canonicalize().unwrap() = "/home/billydm/Dev/vello/shader"
thread 'main' panicked at 'failed to get surface texture: Timeout', examples/with_winit/src/main.rs:189:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/panicking.rs:65:14
   2: core::result::unwrap_failed
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/result.rs:1791:5
   3: core::result::Result<T,E>::expect
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/result.rs:1070:23
   4: with_winit::run::{{closure}}::{{closure}}
             at ./examples/with_winit/src/main.rs:186:35
   5: winit::platform_impl::platform::sticky_exit_callback
             at /home/billydm/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.5/src/platform_impl/linux/mod.rs:849:9
   6: winit::platform_impl::platform::x11::EventLoop<T>::run_return::single_iteration
             at /home/billydm/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.5/src/platform_impl/linux/x11/mod.rs:380:21
   7: winit::platform_impl::platform::x11::EventLoop<T>::run_return
             at /home/billydm/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.5/src/platform_impl/linux/x11/mod.rs:488:27
   8: winit::platform_impl::platform::x11::EventLoop<T>::run
             at /home/billydm/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.5/src/platform_impl/linux/x11/mod.rs:503:25
   9: winit::platform_impl::platform::EventLoop<T>::run
[...]

PR #288, opened 2023-03-04
Add frame statistics UI to with_winit example

armansito:
Added a module for frame time statistics and a UI layer that displays the average, minimum, and maximum frame time alongside FPS. The UI can be toggled by pressing the `S` key.

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -305,7 +305,12 @@ fn run(
             }
             builder.append(&fragment, Some(transform));
             if stats_toggle {
-                snapshot.draw_layer(&mut builder, &mut scene_params.text, width as f64);
+                snapshot.draw_layer(
+                    &mut builder,
+                    &mut scene_params.text,
+                    width as f64,
+                    height as f64,
+                );
             }
             let surface_texture = render_state
                 .surface
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -368,10 +368,11 @@ fn run(
             surface_texture.present();
             device_handle.device.poll(wgpu::Maintain::Poll);
 
+            let new_time = Instant::now();
             stats.add_sample(stats::Sample {
-                frame_time_us: frame_start_time.elapsed().as_micros() as u64,
+                frame_time_us: (new_time - frame_start_time).as_micros() as u64,
             });
-            frame_start_time = Instant::now();
+            frame_start_time = new_time;
         }
         Event::UserEvent(event) => match event {
             #[cfg(not(any(target_arch = "wasm32", target_os = "android")))]
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -158,9 +158,9 @@ fn run(
                                 render_cx.set_present_mode(
                                     &mut render_state.surface,
                                     if vsync_on {
-                                        wgpu::PresentMode::Fifo
+                                        wgpu::PresentMode::AutoVsync
                                     } else {
-                                        wgpu::PresentMode::Immediate
+                                        wgpu::PresentMode::AutoNoVsync
                                     },
                                 );
                             }
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -101,6 +101,7 @@ fn run(
     let mut stats = stats::Stats::new();
     let mut stats_shown = true;
     let mut vsync_on = true;
+    let mut frame_start_time = Instant::now();
     let start = Instant::now();
 
     let mut touch_state = multi_touch::TouchState::new();
@@ -275,7 +276,6 @@ fn run(
             let height = render_state.surface.config.height;
             let device_handle = &render_cx.devices[render_state.surface.dev_id];
             let snapshot = stats.snapshot();
-            let frame_start_time = Instant::now();
 
             // Allow looping forever
             scene_ix = scene_ix.rem_euclid(scenes.scenes.len() as i32);
@@ -371,6 +371,7 @@ fn run(
             stats.add_sample(stats::Sample {
                 frame_time_us: frame_start_time.elapsed().as_micros() as u64,
             });
+            frame_start_time = Instant::now();
         }
         Event::UserEvent(event) => match event {
             #[cfg(not(any(target_arch = "wasm32", target_os = "android")))]
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -260,7 +260,7 @@ fn run(
             let height = render_state.surface.config.height;
             let device_handle = &render_cx.devices[render_state.surface.dev_id];
             let snapshot = stats.snapshot();
-            let _stats_frame = stats.frame_scope();
+            let frame_start_time = Instant::now();
 
             // Allow looping forever
             scene_ix = scene_ix.rem_euclid(scenes.scenes.len() as i32);
@@ -310,6 +310,7 @@ fn run(
                     &mut scene_params.text,
                     width as f64,
                     height as f64,
+                    stats.samples(),
                 );
             }
             let surface_texture = render_state
@@ -350,6 +351,10 @@ fn run(
                 .expect("failed to render to surface");
             surface_texture.present();
             device_handle.device.poll(wgpu::Maintain::Poll);
+
+            stats.add_sample(stats::Sample {
+                frame_time_us: frame_start_time.elapsed().as_micros() as u64,
+            });
         }
         Event::UserEvent(event) => match event {
             #[cfg(not(any(target_arch = "wasm32", target_os = "android")))]
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -99,7 +99,7 @@ fn run(
     let mut simple_text = SimpleText::new();
     let mut images = ImageCache::new();
     let mut stats = stats::Stats::new();
-    let mut stats_toggle = true;
+    let mut stats_shown = true;
     let start = Instant::now();
 
     let mut touch_state = multi_touch::TouchState::new();
@@ -146,7 +146,10 @@ fn run(
                                 transform = Affine::IDENTITY;
                             }
                             Some(VirtualKeyCode::S) => {
-                                stats_toggle = !stats_toggle;
+                                stats_shown = !stats_shown;
+                            }
+                            Some(VirtualKeyCode::C) => {
+                                stats.clear_min_and_max();
                             }
                             Some(VirtualKeyCode::Escape) => {
                                 *control_flow = ControlFlow::Exit;
@@ -304,7 +307,7 @@ fn run(
                 transform = transform * Affine::scale(scale_factor);
             }
             builder.append(&fragment, Some(transform));
-            if stats_toggle {
+            if stats_shown {
                 snapshot.draw_layer(
                     &mut builder,
                     &mut scene_params.text,
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -100,6 +100,7 @@ fn run(
     let mut images = ImageCache::new();
     let mut stats = stats::Stats::new();
     let mut stats_shown = true;
+    let mut vsync_on = true;
     let start = Instant::now();
 
     let mut touch_state = multi_touch::TouchState::new();
@@ -150,6 +151,17 @@ fn run(
                             }
                             Some(VirtualKeyCode::C) => {
                                 stats.clear_min_and_max();
+                            }
+                            Some(VirtualKeyCode::V) => {
+                                vsync_on = !vsync_on;
+                                render_cx.set_present_mode(
+                                    &mut render_state.surface,
+                                    if vsync_on {
+                                        wgpu::PresentMode::Fifo
+                                    } else {
+                                        wgpu::PresentMode::Immediate
+                                    },
+                                );
                             }
                             Some(VirtualKeyCode::Escape) => {
                                 *control_flow = ControlFlow::Exit;
@@ -314,6 +326,7 @@ fn run(
                     width as f64,
                     height as f64,
                     stats.samples(),
+                    vsync_on,
                 );
             }
             let surface_texture = render_state
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -37,6 +37,7 @@ use winit::{
 #[cfg(not(any(target_arch = "wasm32", target_os = "android")))]
 mod hot_reload;
 mod multi_touch;
+mod stats;
 
 #[derive(Parser, Debug)]
 #[command(about, long_about = None, bin_name="cargo run -p with_winit --")]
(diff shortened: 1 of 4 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -249,6 +255,8 @@ fn run(
             let width = render_state.surface.config.width;
             let height = render_state.surface.config.height;
             let device_handle = &render_cx.devices[render_state.surface.dev_id];
+            let snapshot = stats.snapshot();
+            let _stats_frame = stats.frame_scope();
```
DJMcNab, line 259: This is unfortunate, because (at least on my machine), this seems to be vsync locked. On the Tiger demo, I get frame times of nearly exactly `16.62ms`, i.e. slightly less than one vsync interval (other comments cover the gap)

examples/with_winit/src/lib.rs
```diff
@@ -96,6 +97,8 @@ fn run(
     let mut scene = Scene::new();
     let mut fragment = SceneFragment::new();
     let mut simple_text = SimpleText::new();
+    let mut stats = stats::Stats::new();
+    let mut stats_toggle = true;
```
DJMcNab, line 101: Can we rename this to `stats_shown`?

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,169 @@
...
+
+pub struct FrameScope<'a> {
+    stats: &'a mut Stats,
+    start: Instant,
+}
+
+impl<'a> Drop for FrameScope<'a> {
+    fn drop(&mut self) {
+        self.stats
+            .add_sample(self.start.elapsed().as_micros() as u64);
```
DJMcNab, line 117: To ensure that this is measuring time between the same point, rather than rendering time, I think we need to inline the call to `elapsed`, store the new `Instant`, and use that as the start time in `frame_scope`. At the moment, we ignore input processing time, for example, which I don't think is right. 

Actually working out the best measure of FPS has always eluded me.

We also should measure GPU time (e.g. using https://github.com/Wumpf/wgpu-profiler), but wiring that up is non-trivial.

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            self.samples.pop_front()
+        };
+        self.sum += micros;
+        self.samples.push_back(micros);
+        if let Some(oldest) = oldest {
+            self.sum -= oldest;
+        }
+        if micros < self.min {
+            self.min = micros;
```
DJMcNab, line 163: This could also be written as:
```rust
self.min = self.min.min(micros);
```
Which admittedly does produce semantic satiation. I have no real preference there.

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+        if micros > self.max {
+            self.max = micros;
```
DJMcNab, line 166: It's unfortunate that this isn't tracked over only the time in `samples`, but instead is global. I think we need a reset keybinding too, maybe `c` for clear?

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,169 @@
...
+        );
+        text.add(
+            sb,
+            None,
+            text_size,
+            Some(&Brush::Solid(Color::WHITE)),
+            offset * Affine::translate((width * 0.67, text_height)),
+            &format!("FPS: {:.2}", self.fps),
+        );
+    }
```
DJMcNab, line 98: Could we also render a frame time bar graph, showing how stable we are, maybe with some neat colour coding?

DJMcNab (commented):
Nothing blocking, some thoughts on potential future work.

examples/with_winit/src/lib.rs
armansito: I agree. In my local benchmarks we were vsync limited (I've been running mostly on 120Hz/90Hz) displays so I don't always hit the limit but vsync does impact the measurement methodology.

Is it possible to configure the winit event loop to not be locked by vsync? Maybe we should use a different (perhaps headless) environment for benchmarks although I really like the interactive set up (I have already identify several adversarial cases by playing with the scene transform alone).

examples/with_winit/src/stats.rs
armansito: > We also should measure GPU time (e.g. using https://github.com/Wumpf/wgpu-profiler), but wiring that up is non-trivial.

Oh, the GPU time measurement looks neat. I haven't looked into `wgpu-profiler` very closely but supporting GPU timers out of the box would be great (so far I've had to rely on external tools, such as the Metal GPU Performance monitor). I'll investigate this.

examples/with_winit/src/stats.rs
armansito: This is absolutely on my TODO list. I'll spend some time this week to incorporate this into this PR (should be easy to add since the code to track samples for the bar graph is already there).

examples/with_winit/src/stats.rs
DJMcNab: That is to say, I'm proposing that we allow <kbd>C</kbd> to reset the timing, and don't concern ourselves with making the minimum being across the set of samples

DJMcNab: That is to say, I'm proposing that we allow <kbd>C</kbd> to reset the extremes, and don't concern ourselves with making the minimum be across the set of samples

examples/with_winit/src/lib.rs
DJMcNab: I know it's possible - we need to set the wgpu present_mode to *something*

I'd rather stay vsyncy by default through. Maybe another toggle?

DJMcNab:
A couple other thoughts; we should record startup time (ideally both time spent in wgpu `create_shader_module` and seperately, time spent in our pre-processing). And we should have a way to open this on Android. What that looks like, I'm not sure

examples/with_winit/src/stats.rs
armansito: Done. For now all bars are drawn as green but I can look into some color coding next.

examples/with_winit/src/stats.rs
armansito: Done.

examples/with_winit/src/stats.rs
armansito: I like the single line version. Done.

examples/with_winit/src/lib.rs
armansito: Done.

examples/with_winit/src/lib.rs
armansito: Done.

examples/with_winit/src/stats.rs
armansito: I made the elapsed time measurement tighter so it should closely match the presentation time. I haven't really had time to investigate wgpu-profiler yet but I plan to spend some time on wiring it up soon.

examples/with_winit/src/lib.rs
```diff
@@ -142,6 +147,23 @@ fn run(
...
+                                stats.clear_min_and_max();
+                            }
+                            Some(VirtualKeyCode::V) => {
+                                vsync_on = !vsync_on;
+                                render_cx.set_present_mode(
+                                    &mut render_state.surface,
+                                    if vsync_on {
+                                        wgpu::PresentMode::Fifo
+                                    } else {
+                                        wgpu::PresentMode::Immediate
```
DJMcNab, line 163:
```suggestion
                                        wgpu::PresentMode::AutoNoVsync
```
Should be supported on all platforms. There's also a question of doing `PresentMode::AutoVsync` for vsync. I don't have a machine which it would make a difference on; apparently it's only Amd Vulkan.

The alternative might also resolve #266

examples/with_winit/src/lib.rs
```diff
@@ -333,6 +366,11 @@ fn run(
                 .expect("failed to render to surface");
             surface_texture.present();
             device_handle.device.poll(wgpu::Maintain::Poll);
+
+            stats.add_sample(stats::Sample {
+                frame_time_us: frame_start_time.elapsed().as_micros() as u64,
+            });
+            frame_start_time = Instant::now();
```
DJMcNab, line 373:
```suggestion
            let new_time = Instant::now();
            stats.add_sample(stats::Sample {
                frame_time_us: (new_time - frame_start_time).as_micros() as u64,
            });
            frame_start_time = new_time;
```
Should be slightly more correct. In practise, it probably doesn't matter

DJMcNab (commented):
Looks good, just a few thoughts on cross-platform support.

I haven't properly reviewed the code for the frame time graph, but it seems to work OK.
We can get overflow in cases of <kbd>C</kbd> being pressed, but that's unavoidable.

I'd also like to get some way to enable this on Android in (without plugging in a USB keyboard, which might just work), but I'm not sure what that would be.

armansito:
> Looks good, just a few thoughts on cross-platform support.
> 
> 
> 
> I haven't properly reviewed the code for the frame time graph, but it seems to work OK.
> 
> We can get overflow in cases of <kbd>C</kbd> being pressed, but that's unavoidable.
> 
> 
> 
> I'd also like to get some way to enable this on Android in (without plugging in a USB keyboard, which might just work), but I'm not sure what that would be.

I made the stats display on by default and I can see it on my Android device. Does it show up on your device? 

For controls to hide it I thought of adding a "double tap" gesture or recognizing taps over some visible rect which could serve as a button.

DJMcNab:
Ah yes, I hadn't realised the consequences of that; I hadn't tried it on my device. I'm not certain it being displayed by default is what we want, but I don't have a strong objection to it either.

examples/with_winit/src/lib.rs
armansito: Done.

examples/with_winit/src/lib.rs
armansito: Done, I change the code to always use the `Auto*` modes. On web this likely means that "vsync off" will be incorrect so maybe the stats UI shouldn't just blindly display "VSync: off", but it's probably not a huge deal.

armansito:
> Ah yes, I hadn't realised the consequences of that; I hadn't tried it on my device. I'm not certain it being displayed by default is what we want, but I don't have a strong objection to it either.
 
Also there is currently no way to clear min/max on mobile, so we'll want some UI for it. I think can make some simple "double-tap" or some other gesture could work, maybe in another PR.

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,183 @@
...
+            LineTo((bar_width, graph_max_height).into()),
+        ];
+        for (i, sample) in samples.enumerate() {
+            let t = offset * Affine::translate((i as f64 * bar_extent, graph_max_height));
+            // The height of each sample is based on its ratio to the maximum observed frame time.
+            // Currently this maximum scale is sticky and a high temporary spike will permanently
+            // shrink the draw size of the overall average sample, so scale the size non-linearly to
+            // emphasize smaller samples.
+            let h = (*sample as f64) * 0.001 / self.frame_time_max_ms;
+            let s = Affine::scale_non_uniform(1., -h.sqrt());
```
DJMcNab, line 112: This logic is non-obvious to me; I suspect a maximum height based on the maximum seen last render might be better. I'd much rather keep things linear.

DJMcNab (commented):
Code mostly looks good to me, although I'm not convinced by the non-linearity.

So to confirm,  the current status is:
- https://github.com/linebender/vello/pull/288#issuecomment-1472488546 (mobile buttons) can be done in a different PR
- https://github.com/linebender/vello/pull/288#issuecomment-1458553615 (shader startup time) should be reasonably straightforward, but isn't critical
- https://github.com/linebender/vello/pull/288#discussion_r1138265798 `wgpu_profiler` - doesn't need to be this PR
- https://github.com/linebender/vello/pull/288#discussion_r1138259806 (Colour coding) - would be nice to get in this PR - we should also have dashed lines for (at least) 16.6ms and 33.3 ms. It would also be good to use accessible colours here, rather than red and green, although the colour coding is not functional.

examples/with_winit/src/stats.rs
armansito: I switched this to linear now. I thought using sqrt here added value initially but now that we can reset the max/min frame times with a keypress (which resets the scale) this type of visualization is not necessary.

I'll follow up with the dashed lines for 16.6ms/33.3ms as you suggested, which should provide a good relative visual anchor.

armansito:
> Code mostly looks good to me, although I'm not convinced by the non-linearity.
> 
> So to confirm, the current status is:
> 
> * [#288 (comment)](https://github.com/linebender/vello/pull/288#issuecomment-1472488546) (mobile buttons) can be done in a different PR
> * [#288 (comment)](https://github.com/linebender/vello/pull/288#issuecomment-1458553615) (shader startup time) should be reasonably straightforward, but isn't critical
> * [#288 (comment)](https://github.com/linebender/vello/pull/288#discussion_r1138265798) `wgpu_profiler` - doesn't need to be this PR

Working on these in follow up PRs sounds good to me.

> * [#288 (comment)](https://github.com/linebender/vello/pull/288#discussion_r1138259806) (Colour coding) - would be nice to get in this PR - we should also have dashed lines for (at least) 16.6ms and 33.3 ms. It would also be good to use accessible colours here, rather than red and green, although the colour coding is not functional.

I added color coding and lines and picked accessible colors for the coding. I added markers for 8.33, 16.66, and 30.33. I used a solid stroke instead of a dashed stroke as the dashed stroke style didn't work.

DJMcNab (approved):
I'm happy for this to land as is. I've added some more thoughts on Zulip about the scale, so they might be worth considering. We might discuss tomorrow

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,221 @@
...
+            // The height of each sample is based on its ratio to the maximum observed frame time.
+            let h = (*sample as f64) * 0.001 / self.frame_time_max_ms;
+            let s = Affine::scale_non_uniform(1., -h);
+            sb.fill(
+                Fill::NonZero,
+                t * Affine::translate((
+                    left_margin_padding,
+                    (1 + labels.len()) as f64 * text_height,
+                )) * s,
+                if *sample < 16667 {
```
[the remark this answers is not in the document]
armansito: Done.

examples/with_winit/src/stats.rs
```diff
@@ -0,0 +1,221 @@
...
+                sb,
+                None,
+                thres_text_height as f32,
+                Some(&Brush::Solid(Color::WHITE)),
+                offset
+                    * Affine::translate((
+                        left_margin,
+                        (2. - y) * graph_max_height + thres_text_height_2,
+                    )),
+                &format!("{}", t),
```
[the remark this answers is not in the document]
armansito: I think that could be neat, though it's a minor visual improvement given how small the label is rendered. @dfrg Is this straightforward to do this today with the existing code?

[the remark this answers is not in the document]
dfrg: We don’t currently do any sort of alignment but we could probably hack something in.

PR #275, opened 2023-02-06
Migrate the bevy example to stageless

DJMcNab:
The changes required for this were mercifully minimal

PR #290, opened 2023-03-07
Update for Bevy 0.10 and other dependencies

DJMcNab:
Very simple changes, to fix [dependency status](https://deps.rs/repo/github/linebender/vello)

No code changes required, since #275 handled that. I have found that headless isn't quite working right, see [#gpu>Headless](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Headless), but that's pre-existing


dfrg (approved):
I haven’t tested, but this all looks reasonable to me. Thanks!

PR #294, opened 2023-03-10
Add space key for transform reset

power-unsealed:
This adds the missing `SPACE` input handler for resetting the view in the `with_winit` example.

That was an easy fix 🙂

I have no clue about doing pull request, so please tell me if I am doing something wrong...

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -138,6 +138,9 @@ fn run(
                                         * transform;
                                 }
                             }
+                            Some(VirtualKeyCode::Space) => {
+                                transform = Affine::IDENTITY;
+                            }
                             Some(VirtualKeyCode::Escape) => {
                                 *control_flow = ControlFlow::Exit;
                             }


examples/with_winit/src/lib.rs
```diff
@@ -138,6 +138,9 @@ fn run(
                                         * transform;
                                 }
                             }
+                            Some(VirtualKeyCode::Space) => {
```
DJMcNab, line 141: I wonder if <kbd>R</kbd> would be a better choice here. The use of <kbd>Space</kbd> is a remnant from #260 , and this wasn't discussed in there.

I don't have a strong preference any way, but I'd like to leave this open for a bit to let anyone else comment.

(If this is changed, the README would also need to be updated)

DJMcNab (approved):
Looks good to me! Works when I test it locally. Thanks

Note for future readers; this is adding the behaviour to match [examples/with_winit/README.md](https://github.com/linebender/vello/blob/main/examples/with_winit/README.md).

I've just added one thought on the choice of key.

DJMcNab:
Thanks for the PR! After coverage in the office hours, I've decided to merge as-is

PR #295, opened 2023-03-15
Fix handling of interactivity

DJMcNab:
We should just block when making SVGs in the headless example, so do that

### DIFF
--- a/examples/headless/src/main.rs
+++ b/examples/headless/src/main.rs
@@ -104,6 +104,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
         images: &mut images,
         resolution: None,
         base_color: None,
+        interactive: true,
     };
     (example_scene.function)(&mut builder, &mut scene_params);
     let mut transform = Affine::IDENTITY;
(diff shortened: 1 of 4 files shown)


examples/headless/src/main.rs
```diff
@@ -104,6 +104,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
         images: &mut images,
         resolution: None,
         base_color: None,
+        interactive: true,
```
dfrg, line 107: Should `interactive` be false for headless?

examples/with_winit/src/lib.rs
```diff
@@ -268,6 +268,7 @@ fn run(
                 images: &mut images,
                 resolution: None,
                 base_color: None,
+                interactive: false,
```
dfrg, line 271: … and true here? That’s what I would expect semantically and seems to align with the actual use of the value but I may be missing something.

DJMcNab: Thanks! :sweat_smile:

dfrg (approved):
This looks good, thanks!

PR #297, opened 2023-03-21
Replace font backend

dfrg:
This replaces the old moscato font backend with one that has proper support for variable fonts.


raphlinus (approved):
I've only done a light scan through the changes, but they seem reasonable, and I believe we're getting more production quality font parsing through these changes. I was able to run the test scene and it looks good - I might tweak it some for the demo, mostly to have more text.

PR #298, opened 2023-03-22
Change image to flower

raphlinus:
Using vector graphics for the image doesn't make a huge amount of sense. The flower photo is by Raph and happily licensed to anyone who wants to use it.


dfrg (approved):
This is a beautiful photo and a much better demonstration of image rendering!

PR #301, opened 2023-04-05
Update `cargo-run-wasm`

DJMcNab:
`cargo-run-wasm` 0.3.2 is now released, which fixes most of the challenges we were facing.

As I understand it (see [`#gpu > WASM issues`](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/WASM.20issues)), there are still challenges running on web.

This at least doesn't make things worse


DJMcNab:
Since WASM is still fairly broken, we've chosen to defer this

PR #304, opened 2023-04-07
Add GPU profiling

DJMcNab:
Press `P` to write tracing to a file.
```json
{
"traceEvents": [
{ "pid":1, "tid":1, "ts":770178020.7280753, "dur":42696.14479096617, "ph":"X", "name":"render_to_texture_async_coarse" },
{ "pid":1, "tid":1, "ts":770178021.7176586, "dur":354.3749912751082, "ph":"X", "name":"pathtag_reduce" },
{ "pid":1, "tid":1, "ts":770178377.1863999, "dur":105.5208307434441, "ph":"X", "name":"pathtag_reduce2" },
{ "pid":1, "tid":1, "ts":770178483.7488973, "dur":21.718749394494807, "ph":"X", "name":"pathtag_scan1" },
{ "pid":1, "tid":1, "ts":770178506.5613967, "dur":386.66665716391435, "ph":"X", "name":"pathtag_scan" },
{ "pid":1, "tid":1, "ts":770178894.3738872, "dur":74.63541476226965, "ph":"X", "name":"bbox_clear" },
{ "pid":1, "tid":1, "ts":770178970.415552, "dur":8231.510215637172, "ph":"X", "name":"pathseg" },
{ "pid":1, "tid":1, "ts":770187204.321601, "dur":37.23958241153014, "ph":"X", "name":"draw_reduce" },
{ "pid":1, "tid":1, "ts":770187242.5507668, "dur":98.5937475661558, "ph":"X", "name":"draw_leaf" },
{ "pid":1, "tid":1, "ts":770187345.3632643, "dur":64.2187483208545, "ph":"X", "name":"binning" },
{ "pid":1, "tid":1, "ts":770187410.5715959, "dur":124.37499697171006, "ph":"X", "name":"tile_alloc" },
{ "pid":1, "tid":1, "ts":770187535.9361762, "dur":32892.238780277694, "ph":"X", "name":"path_coarse_full" },
{ "pid":1, "tid":1, "ts":770220429.1645398, "dur":154.47916291577712, "ph":"X", "name":"backdrop_dyn" },
{ "pid":1, "tid":1, "ts":770220584.6332861, "dur":127.34374683986971, "ph":"X", "name":"coarse" },
{ "pid":1, "tid":1, "ts":770221705.3103421, "dur":417.1354064510524, "ph":"X", "name":"render_to_texture_async_fine" },
{ "pid":1, "tid":1, "ts":770221706.2478421, "dur":415.2083231474535, "ph":"X", "name":"fine" }
]
}
```

This trace is from zooming in quite far into paris-30k, and shows some interesting results.

This can be used in e.g. https://ui.perfetto.dev/

Headless example is currently untested, but should still work.


src/lib.rs
```diff
@@ -53,6 +55,10 @@ pub struct Renderer {
     shaders: FullShaders,
     blit: Option<BlitPipeline>,
     target: Option<TargetTexture>,
+    #[cfg(feature = "wgpu-profiler")]
+    profiler: GpuProfiler,
+    #[cfg(feature = "wgpu-profiler")]
+    pub profile_result: Option<Vec<wgpu_profiler::GpuTimerScopeResult>>,
```
armansito, line 61: Should this be called `profile_results` as it can store multiple records?

armansito (commented):
Thank you for putting this together! The comments I left are mostly nits, the code looks great.

I ran this locally on mac but the profile UI didn't show up and the captured trace.json file is empty. Is there a special way I need to invoke the with_winit example or am I hitting the mac timer query issues?

src/lib.rs
DJMcNab: I don't think so; this is the results from a single frame.

The resulting value is only incidentally a vector

DJMcNab:
I'm not sure how to diagnose what you're seeing.
I think the wgpu metal backend probably shouldn't enable `TIMESTAMP_QUERY`, which means there should be no trace results.

However, `trace.json` shouldn't be created at all; if `TIMESTAMP_QUERY` isn't created, `profile_result` should always be None.

DJMcNab:
Yeah, it appears [that wgpu-profiler doesn't](https://github.com/Wumpf/wgpu-profiler/blob/d0fae65e9187a64914dca262b309c46c90acb813/src/lib.rs#L279-L283) gate adding frames on whether the frames have any content.

@Wumpf is this an intentional choice? (By the way, `wgpu-profiler` is an otherwise fantastic library, and made the data collection side of this PR trivial)

armansito (approved):
Apologies for the late turnaround, I was out sick most of yesterday. I tried your branch on Linux and it works great. I have a couple of final requests for you to consider:

1. In addition to the current display of pipeline stages by execution order, IWBN to add an option (e.g. with a key binding) to sort the pipeline stage labels by execution time, so that the longest one appears on top.
2. The text labels for `render_to_texture_async_coarse` and `render_to_texture_async_fine` spill past the background rect. I wonder if they should perhaps get clipped or maybe the background rect could get sized to fit the longest text label. This isn't a huge issue but it does make the text hard to read when the background has bright colors in it.
3. The macOS weirdness probably has to do with limited support for the platform on the wgpu-profiler end. IWBN if we didn't create an empty trace.json file but it's not a huge issue. I or someone with access to a Mac can take a look at this later if a macOS device isn't available to you.

Wumpf:
> Yeah, it appears [that wgpu-profiler doesn't](https://github.com/Wumpf/wgpu-profiler/blob/d0fae65e9187a64914dca262b309c46c90acb813/src/lib.rs#L279-L283) gate adding frames on whether the frames have any content.

First, glad you like the crate 😄! The linked line should not care if a frame is empty and only check if the previous purging of exceeding the max pending frame count worked out as expected, i.e. hitting this assert implies a bug in wgpu-profiler if I'm not mistaken.
But sure could be that there's something in there that doesn't handle empty frames as nicely as it should, contributions welcome :)

DJMcNab:
Basically the question is around what the behaviour should be if it's not *possible* for there to be any profiling results, as is the case in the wgpu metal backend (and presumably either web backend too)

For our purposes l, it might make more sense if neither feature is enabled to make the pending frames be none. We can easily enough work around this on our side though

Fwiw, in terms of idiomatic APIs, write_chrometrace should probably accept an `impl Write` rather than a Path.

JMS55:
I worked on a similar feature for bevy, built on top of wgpu-profiler as well. One of the things I ran into is that GPUs tend to clock memory and core speeds up and down, making it hard to get stable timing results. 

You can get around this by locking the GPU clocks using vendor-specific tooling. I included a script for NVIDIA to lock and then unlock the clocks on exit, although I couldn't figure out the equivalent commands for AMD/Intel. You may find it useful.

https://github.com/bevyengine/bevy/pull/8067

DJMcNab:
I've made some changes to hopefully make this better.
Picture's worth a thousand words.

image

I've chosen not to add a sorting by pipeline length, because I think we don't need that with the "slow" stages highlighted directly.

Instead of working out the text width/making it fit or clipping, I just shortened the overall labels.

PR #307, opened 2023-04-14
vello_encoding crate

armansito:
This PR introduces a companion crate to `vello_shaders` that contains the core scene encoding logic as well utilities for renderers to calculate workgroup counts and buffer sizes required to render an encoded scene on the GPU. This PR also updates the vello crate to use this new crate in its own renderer.

This crate is intended to be a step towards the [Semi-stable encoding format](https://github.com/linebender/vello/blob/main/doc/roadmap_2023.md#semi-stable-encoding-format) goal outlined in the 2023 roadmap.


crates/encoding/src/config.rs
```diff
@@ -0,0 +1,327 @@
...
+        let bin_headers = BufferSize::new(draw_object_wgs * 256);
+        let n_paths_aligned = align_up(n_paths, 256);
+        let paths = BufferSize::new(n_paths_aligned);
+
+        // The following buffer sizes have been hand picked to accommodate the vello test scenes as
+        // well as paris-30k. These should instead get derived from the scene layout using
+        // reasonable heuristics.
+        let bin_data = BufferSize::from_size_in_bytes(1 << 20);
+        let tiles = BufferSize::from_size_in_bytes(1 << 24);
+        let segments = BufferSize::from_size_in_bytes(1 << 26);
```
dfrg, line 290: In the original code these, were `/ TILE_SIZE` and `/ SEGMENT_SIZE` respectively. Not too important, but we should probably use `new` instead of `from_size_in_bytes` here to keep the sizes the same.

crates/encoding/src/path.rs
```diff
@@ -381,3 +402,25 @@ impl<'a> PathEncoder<'a> {
         self.n_encoded_segments
     }
 }
+
+impl fello::scale::Pen for PathEncoder<'_> {
```
dfrg, line 406: Nice!

Cargo.toml
```diff
@@ -40,17 +41,20 @@ hot_reload = []
...
-bytemuck = { version = "1.12.1", features = ["derive"] }
 smallvec = "1.8.0"
-fello = { git = "https://github.com/dfrg/fount", rev = "a8c0686ca9da236420c04d1f476c41cf7fe6d2ba" }
-peniko = { git = "https://github.com/linebender/peniko", rev = "cafdac9a211a0fb2fec5656bd663d1ac770bcc81" }
-guillotiere = "0.6.2"
+vello_encoding = { path = "crates/encoding" }
 
 [workspace.dependencies]
+bytemuck = { version = "1.12.1", features = ["derive"] }
+fello = { git = "https://github.com/dfrg/fount", rev = "a8c0686ca9da236420c04d1f476c41cf7fe6d2ba" }
```
dfrg, line 56: This will need to be updated to the match the current rev in main to resolve the merge conflict.

dfrg (approved):
This looks great and I'm happy to see it actually working. Thanks for doing the extra step of getting it fully integrated!

crates/encoding/src/config.rs
armansito: In the original code these size were multiplied by`TILE_SIZE` and `SEGMENT_SIZE` later for buffer allocation, so I think they were given in bytes. That said there is no reason why these couldn't use `::new`. I changed these numbers to be in terms of element counts instead of bytes. The segments buffer is now smaller than this number but it seems to still accommodate paris-30k on my machine.

PR #310, opened 2023-04-22
[svg] Log parse and render times separately

armansito:
The examples logged the total processing time for an SVG including both parse and GPU buffer encoding times. Times for these two operations are now logged separately.

### DIFF
--- a/examples/scenes/src/svg.rs
+++ b/examples/scenes/src/svg.rs
@@ -93,11 +93,13 @@ pub fn svg_function_of<R: AsRef<str>>(
         let start = Instant::now();
         let svg = usvg::Tree::from_str(&contents, &usvg::Options::default())
             .expect("failed to parse svg file");
+        eprintln!("Parsed svg {name} in {:?}", start.elapsed());
+        let start = Instant::now();
         let mut new_scene = SceneFragment::new();
         let mut builder = SceneBuilder::for_fragment(&mut new_scene);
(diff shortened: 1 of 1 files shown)


examples/scenes/src/svg.rs
```diff
@@ -93,11 +93,13 @@ pub fn svg_function_of<R: AsRef<str>>(
         let start = Instant::now();
         let svg = usvg::Tree::from_str(&contents, &usvg::Options::default())
             .expect("failed to parse svg file");
+        eprintln!("Parsed svg {name} in {:?}", start.elapsed());
+        let start = Instant::now();
```
DJMcNab, line 97: This has the same concern as #288 - that is, this excludes the time taken to perform the printing (as `elapsed` creates a new ephemeral `Instant`).

However, I think that's fine, because this isn't trying to determine complete time - it's just to give a flavour of performance.

DJMcNab (approved):
Looks good to me.

To give a flavour of the results, encoding takes approximately 1 order of magnitude less time than parsing.

examples/scenes/src/svg.rs
armansito: Indeed, I don't think including the time to print the message itself is not interesting (and I doubt it will show up even when parsing a modest SVG). As it stands this should print out a fine-grained measurement of parse and encoding times as opposed to a thoroughly inclusive total time.

PR #312, opened 2023-04-23
Fix WASM build

### LINKED ISSUE #276: WASM crashes

armansito:
- Rolled wgpu to 0.16.
- Incorprated the [instant crate](https://crates.io/crates/instant) in lieu of `std::time::Instant` which works on WASM and native builds.
- Fixed the issue with window scaling by setting the canvas size based on winit Window dimensions.

This resolves #276

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ cargo run -p with_winit -- download
 
 ### Bevy
 
-> **Warning**
+> **Warning**  
 > This example currently does not compile. We expect to resolve this as soon as Bevy updates to wgpu 0.16
 
 The [Bevy] example ([examples/with_bevy](examples/with_bevy)) demonstrates using Vello within a [Bevy] application.
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ cargo run -p with_winit -- download
 
 ### Bevy
 
+> **Warning**
+> This example currently does not compile. We expect to resolve this as soon as Bevy updates to wgpu 0.16
+
 The [Bevy] example ([examples/with_bevy](examples/with_bevy)) demonstrates using Vello within a [Bevy] application.
 This currently draws to a [`wgpu`] `Texture` using `vello`, then uses that texture as the faces of a cube.
 
--- a/README.md
+++ b/README.md
@@ -86,10 +86,10 @@ Until browser support becomes widespread, it will probably be necessary to use d
 The following command builds and runs a web version of the [winit demo](#winit). 
 This uses [`cargo-run-wasm`](https://github.com/rukai/cargo-run-wasm) to build the example for web, and host a local server for it
 
-Other examples use the `-p` shorthand, but `cargo-run-wasm` requires the full `--package` and binary name to be specified
-
 ```shell
-cargo run_wasm --package with_winit --bin with_winit_bin
+# Make sure the Rust toolchain supports the wasm32 target
+rustup target add wasm32-unknown-unknown
+cargo run_wasm -p with_winit
 ```
 
 > **Warning**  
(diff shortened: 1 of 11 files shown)


examples/scenes/Cargo.toml
```diff
@@ -16,6 +16,7 @@ vello_svg = { path = "../../integrations/vello_svg" }
 anyhow = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
 image = "0.24.5"
+instant = "0.1.12"
```
DJMcNab, line 19: According to `instant`'s documentation, one of the `stdweb` or `wasm-bindgen` features need to be updated.

I still don't have a working browser with WebGPU, so maybe this works? Perhaps `instant`'s docs need updating

src/engine.rs
```diff
@@ -311,8 +311,8 @@ impl Engine {
                         bytes,
                         wgpu::ImageDataLayout {
                             offset: 0,
-                            bytes_per_row: NonZeroU32::new(
-                                image_proxy.width * format.describe().block_size as u32,
+                            bytes_per_row: Some(
+                                image_proxy.width * format.block_size(None).unwrap(),
```
DJMcNab, line 315:
```suggestion
                                image_proxy.width * format.block_size(None).expect("All supported ImageFormats have a valid block size"),
```
or words to that effect

DJMcNab (commented):
Code seems reasonable to me, although I can't test it without a WebGPU compatible browser.

examples/scenes/Cargo.toml
armansito: Good catch. I went ahead and enabled `wasm-bindgen` though AFAICT all the animated scenes still worked without it.

src/engine.rs
armansito: Done.

armansito:
Unfortunately this PR breaks the bevy example since bevy_render currently depends on wgpu 0.15.

dfrg (approved):

armansito:
> All the changes here are good, there are just a few additions I think should be included in this PR.
> 
> I think this can probably incorporate the changes from #301. In particular, that's an update of `cargo-run-wasm`, including the new support for the `-p` shorthand; adding the WASM target to the instructions is optional but would be nice.
> 
> Also, I'd like us to add a Warning to the Bevy section of the README that the example temporarily doesn't compile. (For context, we decided on Zulip that the Bevy example shouldn't block needed changes, so long as there's a reasonable expectation it won't be broken for long)
> 
> > Warning
> > This example currently does not compile. We expect to resolve this as soon as bevy's has updated to wgpu 0.16
> 
> Additionally, we specify the wgpu version in a badge in README.md. This also needs to be updated. There should probably also be a comment to this effect in the root Cargo.toml.

Done. I cherry-picked your change from #301 into the PR and updated the WASM build instructions to say `cargo run_wasm -p with_winit --bin with_winit_bin` (as the `--bin` part is still necessary for it to work correcfly).

README.md
```diff
@@ -65,6 +65,9 @@ cargo run -p with_winit -- download
 
 ### Bevy
 
+> **Warning**
```
DJMcNab, line 68:
```suggestion
> **Warning**  
```

DJMcNab (approved):
So long as the most recent changes have been tested (i.e. in a browser), this seems all good

PR #313, opened 2023-04-26
Update the bevy example to wgpu 0.16

DJMcNab:
After https://github.com/bevyengine/bevy/pull/8446


armansito (approved):
Thank you for fixing the Bevy example. The code looks good to me.

Please also remove the warning we added earlier to the README as the example compiles and runs now.

PR #314, opened 2023-04-28
A bit of polishing on the demo

raphlinus:
Detect the case where creating the WebGPU surface fails, and insert a message explaning the failure into the DOM, rather than panicking.


examples/with_winit/src/lib.rs
```diff
@@ -453,6 +453,23 @@ enum UserEvent {
...
+    let window = web_sys::window()?;
+    let document = window.document()?;
+    let elements = document.get_elements_by_tag_name("body");
+    let body = elements.item(0)?;
+    let canvas = body.first_child()?;
+    // TODO: style the notice at least a little, maybe link?
+    let text = document.create_text_node(
+        "WebGPU is not enabled. Make sure your browser is updated to
+        Chrome M113 or another browser compatible with WebGPU.",
+    );
```
nicoburns, line 467: You can (presumably) use `innerHTML` here to include an entire HTML string inline.

nicoburns (approved):
Looks good. Would it make sense to also have an error path outside of WASM for the case where WASM itself isn't supported? Or even a `<noscript>` tag for users with JavaScript disabled?

examples/with_winit/src/lib.rs
raphlinus: Ah, good point. For some reason, I was going down the path of creating it using DOM manipulation, but that would be better.

nicoburns:
```rust
let div = document.create_element("div");
div.set_inner_html("HTML here...");
```
if you're wondering about the exact incantation.

raphlinus:
I do think it would be worthwhile supporting the non-wasm case, but that involves figuring out how to manipulate the HTML file served by run-wasm, which I'm not sure how to do.

raphlinus:
I've changed the name of the PR as its scope has expanded. I've made a splash screen that includes instructions and always displays the tiger.

examples/scenes/src/mmark.rs
```diff
@@ -0,0 +1,191 @@
+//! A benchmark based on MotionMark 1.2's path benchmark.
```
DJMcNab, line 1: Should we include a sharing link (e.g. https://browserbench.org/MotionMark1.2/developer.html?warmup-length=2000&warmup-frame-count=30&first-frame-minimum-length=0&test-interval=15&display=minimal&tiles=big&controller=adaptive&frame-rate=50&time-measurement=performance&suite-name=MotionMark&test-name=Paths&complexity=1)

Caveating it as discussed with numbers only approximately comparable

examples/scenes/src/mmark.rs
```diff
@@ -0,0 +1,191 @@
...
+struct Element {
+    seg: PathSeg,
+    color: Color,
+    width: f64,
+    is_split: bool,
+    grid_point: GridPoint,
+}
+
+#[derive(Clone, Copy)]
+struct GridPoint((i64, i64));
```
DJMcNab, line 32: I can't obviously see a reason for this to nest the tuples

DJMcNab (approved):
A few nits/points of discussion, nothing blocking

examples/scenes/src/mmark.rs
```diff
@@ -0,0 +1,191 @@
...
+            self.elements.extend((old_n..n).map(|_| {
+                let element = Element::new_rand(last);
+                last = element.grid_point;
+                element
+            }));
+        }
+    }
+}
+
+impl TestScene for MMark {
```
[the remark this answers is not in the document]
raphlinus: Given that there is state (the elements), I do feel it's cleaner to represent it as a trait implementation than a closure. But I don't have super strong feelings about this, if people want to make further changes (for example, to make things more consistent).

examples/scenes/src/mmark.rs
```diff
@@ -0,0 +1,201 @@
...
+                // This gets color and width from the last element, original
+                // gets it from the first, but this should not matter.
+                sb.stroke(
+                    &Stroke::new(element.width as f32),
+                    Affine::IDENTITY,
+                    element.color,
+                    None,
+                    &path,
+                );
+                path.truncate(0); // Should have clear method, to avoid allocations.
```
DJMcNab, line 101: This increases our kurbo requirement (effectively requiring a `cargo update`)

That's probably fine, but it might be worth telling cargo directly about this (that is, we'd currently fail under `-Z minimal-versions`)

DJMcNab (approved):
I've not gone through with a fine-toothed comb, but it runs OK on my machine, and looks good

WASM doesn't work for me because of #315

examples/scenes/src/mmark.rs
raphlinus: I'm going to leave this. We get our kurbo dep through peniko, so it's not trivial to express and would need to be reworked later. We'll certainly be bumping our kurbo dep because of the stroke work. But thanks for the heads up!

PR #317, opened 2023-05-02
[encoding] Add feature gate for full pipeline support

dfrg:
~~Adds a new `resolve_simple` function that doesn't handle late bound resources (gradients, images and glyph runs).~~

Adds a new feature called "full" that gates encoding support for the full pipeline. This is on by default.

Also adds a new `resolve_solid_paths_only` function that resolves and packs encodings that don't contain late bound resources. 

The primary purpose is to reduce code size and remove unnecessary dependencies for users that are only interested in rendering masks for paths.

### DIFF
--- a/crates/encoding/Cargo.toml
+++ b/crates/encoding/Cargo.toml
@@ -3,8 +3,14 @@ name = "vello_encoding"
 version = "0.1.0"
 edition = "2021"
 
+[features]
+default = ["full"]
+# Enables support for the full pipeline including late-bound
+# resources (gradients, images and glyph runs)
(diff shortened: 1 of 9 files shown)


dfrg:
@armansito this should be ready for review now!

dfrg:
>LGTM! With these changes (plus some additional changes I have planned for vello_shaders) we shave off ~350KB from the total binary size which is substantial.

🎉 

I'm not completely satisfied with `full` either, but it does accurately capture the concept and there is precedent for this feature name in the `tokio` crate. We can revisit if something better pops up.

PR #320, opened 2023-05-10
[shaders] Revise access to backend-agnostic metadata

armansito:
Previously the generated shader data structures were rooted in backend-specific top-level mods (`mod wgsl`, `mod msl`, etc). This made access to per-shader information that is common to all backends (e.g. workgroup sizes, shader name etc) awkward to access from backend agnostic code, especially when feature-gated conditional compilation is used on the client side.

The data structures have been rearranged such that there is a top-level `ComputeShader` declaration for each stage under a `gen` mod. The `ComputeShader` struct declares feature-gated fields for backend shader sources, such that backend specific data is now a leaf node in the structure rather than the root. This has some additional benefits:

1. Common data doesn't have to be redeclared, saving on code size when multiple backends are enabled.

2. The backend specific source code was previously encoded as a `[u8]`. We can now use types that more closely match the expected format, for example `&str` for WGSL and MSL, `[u32]` for SPIR-V, etc.

3. If we ever need to expose additional backend-specific metadata in the future, we can bundle them alongside the source code in a backend-specific data structure at this level of the tree.

### DIFF
--- a/crates/shaders/build.rs
+++ b/crates/shaders/build.rs
@@ -21,18 +21,7 @@ fn main() {
     shaders.sort_by(|x, y| x.0.cmp(&y.0));
     let mut buf = String::default();
     write_types(&mut buf, &shaders).unwrap();
-    if cfg!(feature = "wgsl") {
-        write_shaders(&mut buf, "wgsl", &shaders, |info| {
-            info.source.as_bytes().to_owned()
-        })
-        .unwrap();
-    }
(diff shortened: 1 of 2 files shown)


dfrg (approved):
This restructuring makes a lot of sense to me and I’m particularly happy to see correct typing for shader code for the various backends. Next step for me is to actually hook this up to vello.

Thanks!

PR #321, opened 2023-05-12
Add a CI check that everything compiles

DJMcNab:
Complements #250, but proposes a more likely path to landing in the short term.

This should catch the compilation failure of #314

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,14 +24,21 @@ jobs:
       # --exclude with_bevy # for when bevy has an outdated wgpu version
       # -Dwarnings # for when we have fixed unused code warnings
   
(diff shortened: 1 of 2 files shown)


DJMcNab:
I think that to fully test the WASM, we would also need to manage wasm-bindgen.

Is it valid to just do a `cargo check --target wasm32-unknown-unknown`?

simbleau:
Cargo check will ensure it can be compiled, at the least.

WASM-bindgen provides the #[wasm-bindgen-test] macro for testing on WASM

I'm not sure if it will try to run your current tests on WASM, that might be an issue and I see what you mean.

I still think a compile check would be helpful :)

raphlinus (approved):
This looks good, thanks!

PR #322, opened 2023-05-13
Add a stats display for scene complexity

DJMcNab:
Press <kbd>D</kbd> to enable. It's off by default, as the numbers aren't given in context - accessing the actual maxima requires reading the allocating code. 

image

There are a few hacks to transfer the data around - we will probably want to have a longer term consideration of what form exporting "vello_encoding" takes, for example.


src/lib.rs
```diff
@@ -196,20 +198,25 @@ impl Renderer {
     /// The texture is assumed to be of the specified dimensions and have been created with
     /// the [wgpu::TextureFormat::Rgba8Unorm] format and the [wgpu::TextureUsages::STORAGE_BINDING]
     /// flag set.
+    ///
+    /// The return value is the value of the BumpAllocators in this rendering, which is currently used
```
raphlinus, line 202: Returning BumpAllocators is exposing too much low-level detail, though obviously useful for development for the moment. I'd add a comment suggesting that the exact type may change, for example to more of an opaque stats structure.

raphlinus (approved):
This looks useful, thanks!

PR #323, opened 2023-05-16
[shaders] Shader path look up fixes for hermetic builds

armansito:
* Bazel builds seem to fail to open relative paths to parent directories due to hermetic sandboxing of third-party repositories. This adds a WORKSPACE_MANIFEST_FILE environment variable that allows the caller to optionally provide an absolute path to the workspace root manifest file.
* The existing code processed a shader file only if `FileType::is_file` returns true for it. This is not the case when sources are accessed via symbolic links, which is possible in a Bazel sandbox. The code now filters for the ".wgsl" file extension instead of the file type which should generally be safe.

### DIFF
--- a/crates/shaders/build.rs
+++ b/crates/shaders/build.rs
@@ -8,14 +8,26 @@ mod types;
 
 use std::env;
 use std::fmt::Write;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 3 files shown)


dfrg (approved):
Glad to see this falls back to the relative path when the environment variable is not present. LGTM!

PR #325, opened 2023-05-18
Quiet warnings

raphlinus:
Fixes the warnings remaining in the code, which should in turn let us check that in CI.

Also checks the optional features, which are easy enough to break.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,14 @@ jobs:
       - run: cargo check --features=hot_reload,buffer_labels
       # --exclude with_bevy # for when bevy has an outdated wgpu version
       # -Dwarnings # for when we have fixed unused code warnings
+
+  clippy:
+    runs-on: ubuntu-latest
+    name: Enforce clippy lints
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dtolnay/rust-toolchain@stable
+      - run: cargo clippy --all-targets --workspace -- -D warnings
   
   wasm:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,8 @@ jobs:
       - name: Install native dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
       - run: cargo check --workspace
-      - run: cargo check --features=hot_reload,buffer_labels
+      # Check vello (the default crate) without the features used by `with_winit` for debugging
+      - run: cargo check 
       # --exclude with_bevy # for when bevy has an outdated wgpu version
       # -Dwarnings # for when we have fixed unused code warnings
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,8 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - name: Install native dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
-      - run: cargo check --workspace 
+      - run: cargo check --workspace
+      - run: cargo check --features=hot_reload,buffer_labels
       # --exclude with_bevy # for when bevy has an outdated wgpu version
       # -Dwarnings # for when we have fixed unused code warnings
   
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -20,9 +20,20 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - name: Install native dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
-      - run: cargo check --workspace 
+      - run: cargo check --workspace
+      - run: cargo check --features=hot_reload,buffer_labels
       # --exclude with_bevy # for when bevy has an outdated wgpu version
       # -Dwarnings # for when we have fixed unused code warnings
```
DJMcNab, line 26:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -20,9 +20,20 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - name: Install native dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
-      - run: cargo check --workspace 
+      - run: cargo check --workspace
+      - run: cargo check --features=hot_reload,buffer_labels
```
DJMcNab, line 24: I *think* this should already be checked by `--workspace`, as that implies a unified set of features.

However, I'm not sure that the inverse is being checked (base `vello` with no features). That's unfortunate, as that is the state actual users should be accessing.
I'm not sure how to properly test that

.github/workflows/ci.yml
```diff
@@ -20,9 +20,20 @@ jobs:
...
+
+  clippy:
+    runs-on: ubuntu-latest
+    name: Enforce clippy lints
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Install native dependencies
+        run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
+      - run: cargo clippy --all-targets --workspace -- -D warnings
```
DJMcNab, line 36: My understanding is that `cargo clippy` is strictly more strict than `cargo check`, so we don't need a job for both.
I think we can merge both these jobs (maybe renaming the job?)

If we do rename, we'd need to change the branch protection rules to allow this to land.

src/engine.rs
```diff
@@ -81,6 +82,7 @@ pub enum ResourceProxy {
 }
 
 pub enum ExternalResource<'a> {
+    #[allow(unused)]
```
DJMcNab, line 85: The unfortunate thing about allow unused is that there's no unused allow unused lint.
We just need to hope that we'll be vigilant enough to catch it when these variants do become used.

DJMcNab (commented):
Seems reasonable - I think the workflow file has some unnecessary work.

Not approving as it's still a draft, but most of the changes look good.

.github/workflows/ci.yml
raphlinus: I don't believe `--workspace` covers features in any way. Also, we're not setting default features, so the run without features set should cover that.

.github/workflows/ci.yml
raphlinus: Right. But there is still some discussion to be had about the exact build matrix. Arguably we should check build success in a variety of configurations, but perhaps be strict about warnings and lints only in a smaller number (this may allow us to reduce the `[#allow]` instances that are ascribed to cfg changes.

raphlinus:
I'm marking as ready for review as it's pretty clear some form of this should be landed. I'm not clear on Daniel's latest review - I'm happy to improve it but it's not obvious what the next step is. I'd personally be fine with landing this and then having another PR to refine it.

.github/workflows/ci.yml
DJMcNab: My reason for suspecting it would check all features is that `with_winit` requires `buffer_labels` and `hot_reload`.

However, you are correct that a "root" `cargo check` only checks Vello without features, so I believe this should be:
```suggestion
      # Check vello (the default crate) without the features used by `with_winit` for debugging
      - run: cargo check 
```

dfrg (approved):
This transitions us to a warning and clippy clean state and seems to be sufficient in ensuring we stay there. It would be nice if our CI rules were principled, well understood and consistent across the org so I think it’d be useful to have a broader conversation about that as a group but this PR shouldn’t block on that so this lgtm.

raphlinus:
Agree with @dfrg. I'm not going to claim this is the final state, and we *should* normalize our config across the repos (including figuring out the correct toolchain script, as discussed in last week's office hours), but this gets us from a bunch of warnings to clean clippy enforced, and that's progress.

PR #329, opened 2023-06-14
[draw_leaf] Don't write past the end of the draw_monoids buffer

armansito:
The number of global invocations for draw_leaf can exceed the size of the draw_monoids buffer which gets conservatively set to the number of draw objects. Added an explicit bounds check to prevent the invalid write.

This is not an issue when targeting wgpu as the WGSL compiler emits implicit bounds checking. When targeting Metal, we disable implicit bounds checks as that requires an extra buffer binding containing buffer sizes.

This was caught by Xcode's Metal shader validation and resulted in visual artifacts in native rendering.


DJMcNab:
There is the concern that we're doing the same bounds check twice here when using `wgpu`

Not sure how to resolve it though - gating this check would be ridiculously noisy

raphlinus (approved):
Approved, as it fixes the problem. I'm not 100% sure this is the best long-term approach, for two reasons:

* The scene read is not gated, so *potentially* can exceed the buffer. Almost all the time, we expect this to read whatever is placed in the scene after the draw tags, but it doesn't feel clean.
* The cost of the branch may not be zero, and could be avoided if the buffer was padded up (as is done for some of the other buffers).

I'm not too worried about the cost of the branch. I was going to write (hopefully) that a good shader compiler could take the two branches and remove one, but now I realize that won't work as one limit is read from config, the other is metadata, managed by the API. It's *possible* we could use the `arrayLength` WGSL built-in function, but I've been avoiding that because it requires an extra binding for the array sizes on Metal. All this to say, I don't consider it worth micro-optimizing right now, but it's something we might get back to later, especially if we try to do pre-compiled shaders.

armansito:
I agree that I don't think the cost of the extra branch is important. I also agree with Raph's comment that I don't think the additional branch can be optimized away.

In the long term I think it's probably best to have the robustness transform turned on even if the extra metadata buffer binding is annoying when targeting Metal (when targeting Vulkan, the codegen uses the SPIR-V  `OpArrayLength` built-in). For now I'm going to move forward with the branch conditional here as it fixes the bug and satisfies shader validation but I'll file an issue about robustness in native backends and mention it in a comment that this needs work.

---
P.S.:
The robustness transform turns any unsized buffer access expression of the form `buffer[index]` into
```
buffer[min(index, arrayLength(buffer) - 1)]
```
When targeting Metal, `arrayLength` is implemented by storing buffer sizes in a special uniform (i.e. `constant`) buffer binding. Example output from Tint looks like this:
```
(*(buffer_ptr))[min(index, (((*(buffer_sizes_uniform)).buffer_size[0u][1u] / 4u) - 1u))]
```
When targeting SPIR-V you get the following (where %20 is the `index` and `%19 = OpExtInstImport "GLSL.std.450"`):
```
%21 = OpArrayLength %uint %buffer 0
%23 = OpISub %uint %21 %uint_1
%18 = OpExtInst %uint %19 UMin %20 %23
%25 = OpAccessChain %_ptr_StorageBuffer_uint %buffer %uint_0 %18
```
Regardless of whether the access is gated with a branch conditional, or with an explicitly authored call to `min`, if the array size comes from a uniform binding other than `arrayLength` it's safe to assume that it won't get optimized away.

Interestingly, spirv-opt doesn't seem to eliminate an extra call to min when I actually gate the write on the result of `arrayLength`, such that the following WGSL
```
buffer[min(index, arrayLength(&buffer) - 1)] = 5;
```
gets reduced by `spirv-opt` to the following (where %36 corresponds to `index`):
```
%37 = OpArrayLength %uint %buffer 0
%38 = OpISub %uint %37 %uint_1
%39 = OpExtInst %uint %19 UMin %36 %38
%42 = OpExtInst %uint %19 UMin %39 %38
%43 = OpAccessChain %_ptr_StorageBuffer_uint %buffer %uint_0 %42
      OpStore %43 %uint_5
```
in which `UMin` is chained with the same identical upper bound (%38). It could be possible that the driver compiler could eliminate the additional `UMin` but I haven't looked at an ISA to confirm this (though it would be interesting to do so).

armansito:
I decided to fix several other shader validation errors by clamping the buffer indices by the object counts stored in `config`. I actually think that these additional checks are useful even when the robustness transform is enabled during WGSL translation (such as when running on wgpu or Dawn), because the implicit checks are against the buffer object sizes which won't catch overruns that can occur if the buffer bindings are sub-allocations from a buffer object.

Vello doesn't run into this since the wgpu renderer allocates a new buffer object for each binding, whereas Skia sub-allocates from a buffer object pool (though I was triggering shader validation errors because the buffers in my test run happened to be large enough to span an entire object).

There could be more invalid accesses that I missed. These two changes fix several visual artifacts and I haven't run into any new validation errors (yet).

shader/draw_leaf.wgsl
```diff
@@ -73,7 +73,7 @@ fn main(
     workgroupBarrier();
     var m = sh_scratch[0];
     workgroupBarrier();
-    let tag_word = scene[config.drawtag_base + ix];
+    let tag_word = scene[min(config.drawtag_base + ix, config.drawtag_base + config.n_drawobj - 1u)];
```
raphlinus, line 76: I think a better approach (and a pattern followed elsewhere) is to conditionalize the read so it becomes `DRAWTAG_NOP`.

shader/draw_leaf.wgsl
```diff
@@ -90,14 +90,16 @@ fn main(
         m = combine_draw_monoid(m, sh_scratch[local_id.x - 1u]);
     }
     // m now contains exclusive prefix sum of draw monoid
-    draw_monoid[ix] = m;
+    if ix < config.n_drawobj {
```
raphlinus, line 93: +1

shader/draw_leaf.wgsl
```diff
@@ -90,14 +90,16 @@ fn main(
...
+        draw_monoid[ix] = m;
+    }
     let dd = config.drawdata_base + m.scene_offset;
     let di = m.info_offset;
     if tag_word == DRAWTAG_FILL_COLOR || tag_word == DRAWTAG_FILL_LIN_GRADIENT ||
         tag_word == DRAWTAG_FILL_RAD_GRADIENT || tag_word == DRAWTAG_FILL_IMAGE ||
         tag_word == DRAWTAG_BEGIN_CLIP
     {
-        let bbox = path_bbox[m.path_ix];
+        let bbox = path_bbox[min(m.path_ix, config.n_path - 1u)];
```
raphlinus, line 102: This will already be gated by the tag being NOP.

shader/draw_leaf.wgsl
```diff
@@ -242,7 +244,7 @@ fn main(
         if tag_word == DRAWTAG_BEGIN_CLIP {
             path_ix = m.path_ix;
         }
-        clip_inp[m.clip_ix] = ClipInp(ix, i32(path_ix));
+        clip_inp[min(m.clip_ix, config.n_clip - 1u)] = ClipInp(ix, i32(path_ix));
```
raphlinus, line 247: This will be gated by the tag being NOP. Further, it's the locus of a potentially more severe set of problems. In the likely case that you get a whole bunch of END_CLIP tags, they'll all pile on to the last element in the `clip_inp` array, which will leave it with a likely incorrect value, and also be undefined behavior due to racy writes.

shader/binning.wgsl
```diff
@@ -74,7 +74,7 @@ fn main(
         let draw_monoid = draw_monoids[element_ix];
         var clip_bbox = vec4(-1e9, -1e9, 1e9, 1e9);
         if draw_monoid.clip_ix > 0u {
-            clip_bbox = clip_bbox_buf[draw_monoid.clip_ix - 1u];
+            clip_bbox = clip_bbox_buf[min(draw_monoid.clip_ix - 1u, config.n_clip - 1u)];
```
raphlinus, line 77: This should not ever trigger (the code is only called for draw objects in range, which should in turn have valid monoid values), but I suppose is harmless. See "debug assert" comment in review.

shader/clip_leaf.wgsl
```diff
@@ -187,8 +187,8 @@ fn main(
         // Fix up drawmonoid so path_ix of EndClip matches BeginClip
         let parent_clip = clip_inp[parent];
         let path_ix = parent_clip.path_ix;
-        let parent_ix = parent_clip.ix;
-        let ix = ~inp;
+        let parent_ix = min(parent_clip.ix, config.n_drawobj - 1u);
```
raphlinus, line 190: I'm not a fan of this change, as if it triggers it will create UB and incorrect output.

raphlinus (commented):
There are comments associated with individual changes, but I have some thoughts.

There is *definitely* a problem with the existing state of the code, which is that it will potentially read junk draw tags past the end, and do a bunch of additional stuff downstream based on them, especially clip-related. I suggest we gate the read of the draw tag. Another approach is to pad the draw tag stream to workgroup size with NOP, but that's an additional burden on the encode side so I'm fine with fixing it in the shader.

There's a deeper issue worth discussing, which I think of as "debug asserts." If the logic is correct, then there should be no buffer reads or writes past the end. Further, out of bounds memory accesses are a pretty good indicator of logic errors. Doing clamping or conditionalizing should have no effect if the logic is correct, and the cost is modest-ish (actually that's an empirical question, with very complicated overlaps with existing robustness mechanisms in the GPU infrastructure). It can avoid things falling over, but doesn't avoid UB in the write case, and I'm worried can mask deeper problems.

A direction I'd be interested in going is having a "debug assert" mechanism, gated by an ifdef, so that a bounds check hits an atomic (in bump). We actually had something much like this in piet-gpu, but I ripped it out because we didn't have a systematic way of exercising it on the Rust side (it was developed for Gio), and it was causing nontrivial friction as I was making changes. I think it would be especially useful in conjunction with automated testing.

raphlinus (approved):
Looks good, thanks!

PR #331, opened 2023-06-18
Encoding.encode_radial_gradient double-draw for degenerate gradients

jonathanolson:
Beforehand, if p0=p1 and r0=r1, it would encode the transparent color and continue, encoding ANOTHER draw for the gradient afterward (either the encode_color, or pushing the gradient).

This would cascade into further draws NOT corresponding to the desired paths.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,8 +11,8 @@
     "ptcl": "${workspaceFolder}/shader/shared/ptcl.wgsl",
     "segment": "${workspaceFolder}/shader/shared/segment.wgsl",
     "tile": "${workspaceFolder}/shader/shared/tile.wgsl",
-    "transform": "${workspaceFolder}/shader/shared/transform.wgsl"
-    "util": "${workspaceFolder}/shader/shared/util.wgsl",
+    "transform": "${workspaceFolder}/shader/shared/transform.wgsl",
+    "util": "${workspaceFolder}/shader/shared/util.wgsl"
   },
   "wgsl-analyzer.diagnostics.nagaVersion": "main",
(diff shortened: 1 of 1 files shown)


crates/encoding/src/encoding.rs
```diff
@@ -306,13 +306,15 @@ impl Encoding {
         if gradient.p0 == gradient.p1 && (gradient.r0 - gradient.r1).abs() < SKIA_EPSILON {
             self.encode_color(DrawColor::new(Color::TRANSPARENT));
```
DJMcNab, line 307: I'd be tempted to just make this condition return

Either as `return self.encode_color`, or by adding a `return;` statement.
That matches the semantics of this being an early-exit condition

DJMcNab (commented):
Good catch! I'll let one of the experts review it properly

Do you think you can add a test scene which meets this case? (Or an item into an existing test scene)

crates/encoding/src/encoding.rs
jonathanolson: Sounds good, added here I believe (and fixed formatting)

jonathanolson:
Added a test case above in test_scenes.rs

Without the patch:
image

With the patch:
image

It just adds a degenerate gradient fill AFTER the red square, but before the green/blue squares. When unpatched, the double-draw shifts the 2nd and 3rd (supposed to be green/blue) squares, with the 3rd square incorrectly showing up as green.

dfrg (approved):
Nice catch indeed, and the fix looks good to me. Thanks!

dfrg:
Looks like a subsequent PR created a conflict. Happy to merge this as soon as it’s resolved.

Issue #333, opened 2023-06-27
Incorrect clipping


jonathanolson:
In https://github.com/jonathanolson/vello/commit/6eea43d8d77576e36b15bd578f549a04676e6e1e, I have a test scene that is rendering content outside of a defined clip area:

<img width="52" alt="image" src="https://github.com/linebender/vello/assets/480878/341be5cb-a4bf-4142-96d2-6d0170d2909c">

Where a rectangular clipping area is applied (where the blue rectangle is visible, it covers the entire region). The red rectangle should not appear. The green rectangle and blue rectangle are correctly clipped.

```rust
let large_background_rect = kurbo::Rect::new(
    -1000.0, -1000.0,
    2000.0, 2000.0
);
let inside_clip_rect = kurbo::Rect::new(
    11.0, 13.399999999999999,
    59.0, 56.6
);
let outside_clip_rect = kurbo::Rect::new(
    12.599999999999998, 12.599999999999998,
    57.400000000000006, 57.400000000000006
);
let clip_rect = kurbo::Rect::new(
    0.0, 0.0,
    74.4, 339.20000000000005
);
let scale = 0.8876953125;

sb.push_layer( BlendMode { mix: peniko::Mix::Normal, compose: peniko::Compose::SrcOver}, 1.0, Affine::new([scale, 0.0, 0.0, scale, 27.07470703125, 176.40660533027858]), &clip_rect);

sb.fill( peniko::Fill::NonZero, kurbo::Affine::new([scale, 0.0, 0.0, scale, 27.07470703125, 176.40660533027858]), peniko::Color::rgb8(0, 0, 255), None, &large_background_rect);
sb.fill( peniko::Fill::NonZero, kurbo::Affine::new([scale, 0.0, 0.0, scale, 29.027636718750003, 182.9755506427786]), peniko::Color::rgb8(0, 255, 0), None, &inside_clip_rect);
sb.fill( peniko::Fill::NonZero, kurbo::Affine::new([scale, 0.0, 0.0, scale, 29.027636718750003, 559.3583631427786]), peniko::Color::rgb8(255, 0, 0), None, &outside_clip_rect);

sb.pop_layer();
```

This came up in the clip area used for our carousel component, where clipped content shows below the carousel:
<img width="130" alt="image" src="https://github.com/linebender/vello/assets/480878/4a625c75-84e4-4312-968b-e62e0ae8aee6">

jonathanolson:
[...]

PR #335, opened 2023-06-29
Correctly handle disjoint bounding-box intersections in binning

### LINKED ISSUE #286: Text clipping artifacts
It seems as though text overflowing a clipping layer doesn't work properly. I've added the following to the `animated_text` test scene, around the second "Animated text!" element

```rust
const X0: f64 = 50.0;
const Y0: f64 = 450.0;
const X1: f64 = 200.0;
const Y1: f64 = 950.0;

let path = [
    PathEl::MoveTo((X0, Y0).into()),
    PathEl::LineTo((X1, Y0).into()),
    PathEl::LineTo((X1, Y0 + t * (Y1 - Y0)).into()),
    PathEl::LineTo((X1 + t * (X0 - X1), Y1).into()),
    PathEl::LineTo((X0, Y1).into()),
    PathEl::ClosePath,
];

sb.push_layer(Mix::Clip, 1.0, Affine::IDENTITY, &path);
{
    params.text.add(
        sb,
        None,
        text_size,
        None,
        Affine::translate((110.0, 700.0)),
        s,
    );
}
sb.pop_layer();
```

Which results in many vertical bars clipping the text, instead of the remaining text being clipped in its entirety:

image

armansito:
When the bounding boxes of a path and its clip are disjoint (i.e. they do not intersect) the result of their intersection is a negative rectangle.

When calculating the intersection of the bboxes, the binning stage ensures that the bbox is non-negative. It then normalizes the coordinates to bin dimensions and rounds the top-left corner down to the neareast integer and the bottom-right corner up.

However this rounding causes zero-area bounding boxes to have a non-zero area and sends the bottom-right corner to be placed in the next bin. This causes fully clipped out draw objects to be included in binning, with an incorrect clip bounding box that causes them to be erroneously drawn with partial clipping.

`binning` now takes care around this logic to make sure that a zero-area intersected-bbox gets skipped and clipped out. `tile_alloc`, also takes care in its own similar logic over intersected-bbox contents.

Fixes #286 and #333

### DIFF
--- a/shader/binning.wgsl
+++ b/shader/binning.wgsl
@@ -86,15 +86,20 @@ fn main(
 
         let path_bbox = path_bbox_buf[draw_monoid.path_ix];
         let pb = vec4<f32>(vec4(path_bbox.x0, path_bbox.y0, path_bbox.x1, path_bbox.y1));
-        let bbox_raw = bbox_intersect(clip_bbox, pb);
-        // TODO(naga): clunky expression a workaround for broken lhs swizzle
-        let bbox = vec4(bbox_raw.xy, max(bbox_raw.xy, bbox_raw.zw));
+        let bbox = bbox_intersect(clip_bbox, pb);
 
         intersected_bbox[element_ix] = bbox;
-        x0 = i32(floor(bbox.x * SX));
-        y0 = i32(floor(bbox.y * SY));
-        x1 = i32(ceil(bbox.z * SX));
-        y1 = i32(ceil(bbox.w * SY));
+
+        // `bbox_intersect` can result in a zero or negative area intersection if the path bbox lies
+        // outside the clip bbox. If that is the case, Don't round up the bottom-right corner of the
+        // and leave the coordinates at 0. This way the path will get clipped out and won't get
+        // assigned to a bin.
+        if bbox.x < bbox.z && bbox.y < bbox.w {
+            x0 = i32(floor(bbox.x * SX));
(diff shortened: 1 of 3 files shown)


shader/binning.wgsl
```diff
@@ -89,12 +89,18 @@ fn main(
         let bbox_raw = bbox_intersect(clip_bbox, pb);
         // TODO(naga): clunky expression a workaround for broken lhs swizzle
         let bbox = vec4(bbox_raw.xy, max(bbox_raw.xy, bbox_raw.zw));
```
raphlinus, line 91: You probably no longer need the max here, as everything downstream should be able to interpret x1 < x0 or y1 < y0 as an empty box.

raphlinus (approved):
Looks great, and exciting to see a fix for this long-standing issue.

dfrg:
Yes! This has been in my queue for a long time and I’ve been dreading hunting it down. I’m excited to test this on some Lottie scenes with known clipping artifacts.

shader/binning.wgsl
armansito: Done.

armansito:
> Yes! This has been in my queue for a long time and I’ve been dreading hunting it down. I’m excited to test this on some Lottie scenes with known clipping artifacts.

@dfrg Yes, please do! Let me know if this fixes the artifacts you've been seeing with lottie scenes.

PR #338, opened 2023-07-03
Remove direct dep on parking_lot, smallvec.

waywardmonkeys:
Both of these are used transitively, but not directly by `vello`, so we don't need to list them here.


DJMcNab (approved):
Thanks!

PR #339, opened 2023-07-03
headless: Use pollster from workspace.

waywardmonkeys:
This prevents having 2 versions of `pollster` pulled into the workspace. The workspace provided `pollster` `0.3`, while `headless` was still using `0.2.5`.


DJMcNab (approved):
Looks good!

PR #340, opened 2023-07-05
vello_shaders: Use naga 0.12, not git rev.

waywardmonkeys:
The git revision being used was from `0.11` while the current release is `0.12`. `0.12` is used in `wgpu`, so this lets us have a single `naga` crate being built.


DJMcNab:
The git revision was pinned, without comment that I can find, and appears to be included in the 0.12 release

DJMcNab (approved):
We think this shouldn't be a problem

PR #343, opened 2023-07-10
with_bevy: Update to bevy 0.11 release.

waywardmonkeys:
* Use `add_plugins` as `add_plugin` is deprecated.
* Use `world.resource` instead of unwrapping `world.get_resource` as it gives better error messages.
* Add the `VelloRenderer` to the render app in the `finish` method as the `Renderer` is async and may not have initialized by the time that the Vello plugin is being set up.

### DIFF
--- a/examples/with_bevy/src/main.rs
+++ b/examples/with_bevy/src/main.rs
@@ -21,8 +21,8 @@ struct VelloRenderer(Renderer);
 
 impl FromWorld for VelloRenderer {
     fn from_world(world: &mut World) -> Self {
-        let device = world.get_resource::<RenderDevice>().unwrap();
-        let queue = world.get_resource::<RenderQueue>().unwrap();
+        let device = world.resource::<RenderDevice>();
+        let queue = world.resource::<RenderQueue>();
 
         VelloRenderer(
             Renderer::new(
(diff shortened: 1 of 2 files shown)


examples/with_bevy/src/main.rs
```diff
@@ -78,11 +82,11 @@ fn render_scenes(
...
     App::new()
         .add_plugins(DefaultPlugins)
-        .add_plugin(VelloPlugin)
+        .add_plugins(VelloPlugin)
         .add_systems(Startup, setup)
         .add_systems(Update, bevy::window::close_on_esc)
         .add_systems(Update, cube_rotator_system)
-        .add_plugin(ExtractComponentPlugin::<VelloScene>::default())
+        .add_plugins(ExtractComponentPlugin::<VelloScene>::default())
         .add_systems(Update, render_fragment)
```
DJMcNab, line 90: Should these disjoint calls be combined? I.e. `add_plugins((DefaultPlugins, VelloPlugin, Extract...))`. I'm not sure what's currently idiomatic.

Same with the (non-startup) `add_systems` calls, fwiw

DJMcNab (approved):
Looks OK to me, just one thought about being idiomatic

Runs alright on my machine

examples/with_bevy/src/main.rs
waywardmonkeys: I think in samples that I saw, it seemed to go either way ... we should reach out to the people doing the bevy /vello integration and suggest that they look at how we can improve this as well. :)

Going to land it as-is for now though.

PR #344, opened 2023-07-17
deps: with_winit: Update notify-debouncer-mini and console-log.

waywardmonkeys:


DJMcNab (commented):
Hot reloading still works, although I haven't tested the console logging.

It still compiles though, so it seems reasonable

PR #345, opened 2023-07-18
deps: Update usvg to 0.33 from 0.29

waywardmonkeys:
This is not the latest version currently available, but the changes to update to 0.34 are more involved, so we'll do this intermediate step now.


DJMcNab (approved):
Works on my machine!

PR #347, opened 2023-07-20
 Add bevy_vello in README integrations

simbleau:


DJMcNab (changes requested):
I'd be happy to mention this crate, but I don't think this is the right stop. I believe integrations should only be code under linebender, which this isn't

If you wanted to add a single pointer sentence to the bevy example instructions, that would be fine.
E.g.
```md
A community integration as a library can be found at...
```

simbleau:
@DJMcNab Added it to the examples, per your request. Thanks!

DJMcNab (approved):
I'm happy enough with this now.

Thanks!

PR #352, opened 2023-08-07
Update fello version

dfrg:
This brings in experimental CFF/CFF2 support.


waywardmonkeys (approved):
Not sure what to verify here other than it passing CI … so this is a rubber stamp unless there is something else I can help to validate.

dfrg:
There was a fairly small change to fello to pull in code from the google fonts repo for CFF support. This shouldn't break anything that was already working. If there are problems with the CFF code, maybe we'll catch them in xilem. Thanks!

PR #359, opened 2023-09-11
📀 Make `wgpu` an optional dependency

Jasper-Bekkers:
This turns `wgpu` into an optional dependency that's default enabled.

In our project we're not using `wgpu` as the renderer so interop-ing with this library is a bit trick, but I think I've find a way that would work for us. We can take the pre-recorded command buffer `Recording` and convert it to our own abstraction.

As for the shaders, we can use naga and patch files to make them work in our environment.

@raphlinus We disccussed various ways of going about this problem some time ago, but I think this wouldn't be too intrusive on either of our projects. If it is, let me know.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
       - run: cargo check --workspace
       # Check vello (the default crate) without the features used by `with_winit` for debugging
       - run: cargo check 
+      # Check vello (the default crate) without wgpu
+      - run: cargo check --no-default-features
       # --exclude with_bevy # for when bevy has an outdated wgpu version
       # -Dwarnings # for when we have fixed unused code warnings
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -21,6 +21,7 @@ use std::{
     sync::atomic::{AtomicU64, Ordering},
 };
 
+#[cfg(feature = "wgpu")]
 use wgpu::{
     BindGroup, BindGroupLayout, Buffer, BufferUsages, CommandEncoderDescriptor, ComputePipeline,
     Device, Queue, Texture, TextureAspect, TextureUsages, TextureView, TextureViewDimension,
@@ -28,14 +29,15 @@ use wgpu::{
 
 pub type Error = Box<dyn std::error::Error>;
 
-#[derive(Clone, Copy)]
-pub struct ShaderId(usize);
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct ShaderId(pub usize);
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash)]
-pub struct Id(NonZeroU64);
+pub struct Id(pub NonZeroU64);
 
 static ID_COUNTER: AtomicU64 = AtomicU64::new(0);
 
+#[cfg(feature = "wgpu")]
 pub struct Engine {
     shaders: Vec<Shader>,
     pool: ResourcePool,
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -23,6 +23,8 @@ jobs:
       - run: cargo check --workspace
       # Check vello (the default crate) without the features used by `with_winit` for debugging
       - run: cargo check 
+      # Check vello (the default crate) without wgpu
+      - run: cargo check --no-default-features
```
DJMcNab, line 27: How long does this check take?

src/engine.rs
```diff
@@ -29,20 +30,22 @@ use wgpu::{
...
+pub struct ShaderId(pub usize);
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash)]
-pub struct Id(NonZeroU64);
+pub struct Id(pub NonZeroU64);
 
 static ID_COUNTER: AtomicU64 = AtomicU64::new(0);
 
+#[cfg(feature = "wgpu")]
 pub struct Engine {
```
DJMcNab, line 41: These sections together are a bit strange

This suggests the engine module is too entangled with wgpu/that these IDs shouldn't be in this module.
If this is the wgpu engine implementation, it should maybe be more separated?

src/util.rs
```diff
@@ -175,6 +175,7 @@ impl std::task::Wake for NullWake {
 /// Block on a future, polling the device as needed.
 ///
 /// This will deadlock if the future is awaiting anything other than GPU progress.
+#[cfg(feature = "wgpu")]
```
DJMcNab, line 178: Why is this double `cfg(wgpu)`ed?

DJMcNab (commented):
I really don't like quite how many `cfg(feature="wgpu")` there are here. I can already foresee a significant number of changes running into "oh, this needs wpgu".

I think the recommended approach for other renderers might be to use the shaders crate? (That might only be for non-Rust languages though, I've not looked into it too much)

src/engine.rs
Jasper-Bekkers: It seems like at least there was an intention behind splitting `Engine` and `Recording` up, maybe moving all of the `Engine` stuff to a `wgpu.rs` file is better going forward?

src/util.rs
Jasper-Bekkers: My bad - I later realized this entire file was just wgpu helpers.

.github/workflows/ci.yml
Jasper-Bekkers: The CI run for that step (on this PR) was ~3 seconds.

src/engine.rs
Jasper-Bekkers: Fwiw, anybody implementing another `Engine` type thing would need access to these internals to be able to translate between `Recording` and it's own internal representation since the `Engine` is responsible for handing these out.

.github/workflows/ci.yml
DJMcNab: Fair enough! Useful to have it documented here though. I've got no objection to adding this check on that metric then.

raphlinus:
Very interesting to see this!

The `Engine` abstraction was designed to support multiple GPU execution engines, and we are doing some work with just using the shaders, and a separate crate for encoding (completely reimplementing the dispatching.

The direction I'd like to see is that `Engine` is factored so that `Recording` is in its own module that would always be present, but all of `Engine` and its dependencies would be gated by a single feature gate. An even sharper separation would be to have that as a separate crate, but that might increase friction for making further changes to the engine (see #360, as well as changes needed to support WGPU extensions, including f16, subgroups, and descriptor indexing). But I can see this as a reasonable first step; slightly messy but not a massive impact on the code.

raphlinus (approved):
Yes, let's land this, as it's a fairly minimal change so won't have much conflict with other work in flight. As a followup, I think we should move the wgpu backend into a separate rs file, to make the granularity of the cfg coarser. I've done enough experimentation to be confident that will be straightforward.

Thanks!

PR #360, opened 2023-09-11
Add indirect dispatch and CPU shaders

raphlinus:
A first cut at adding indirect dispatch and CPU shaders. This version just has one (very simple shader), and allows its selection on the command line.

### DIFF
--- a/src/cpu_shader/pathtag_reduce.rs
+++ b/src/cpu_shader/pathtag_reduce.rs
@@ -1,0 +1,35 @@
+// Copyright 2023 The Vello authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use vello_encoding::{ConfigUniform, Monoid, PathMonoid};
+
+use crate::cpu_dispatch::CpuBinding;
+
+const WG_SIZE: usize = 256;
+
+fn pathtag_reduce_main(
+    n_wg: u32,
+    config: &ConfigUniform,
+    scene: &[u32],
+    reduced: &mut [PathMonoid],
+) {
+    let pathtag_base = config.layout.path_tag_base;
+    for i in 0..n_wg {
+        let mut m = PathMonoid::default();
+        for j in 0..WG_SIZE {
+            let tag = scene[(pathtag_base + i * WG_SIZE as u32) as usize + j];
+            m = m.combine(&PathMonoid::new(tag));
+        }
+        reduced[i as usize] = m;
+    }
+}
+
+pub fn pathtag_reduce(n_wg: u32, resources: &[CpuBinding]) {
+    let r0 = resources[0].as_buf();
+    let r1 = resources[1].as_buf();
+    let mut r2 = resources[2].as_buf();
+    let config = bytemuck::from_bytes(&r0);
+    let scene = bytemuck::cast_slice(&r1);
+    let reduced = bytemuck::cast_slice_mut(r2.as_mut());
+    pathtag_reduce_main(n_wg, config, scene, reduced);
+}
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -16,15 +16,19 @@
 
 use std::{
     borrow::Cow,
+    cell::RefCell,
     collections::{hash_map::Entry, HashMap, HashSet},
     num::NonZeroU64,
     sync::atomic::{AtomicU64, Ordering},
 };
 
 use wgpu::{
-    BindGroup, BindGroupLayout, Buffer, BufferUsages, CommandEncoderDescriptor, ComputePipeline,
-    Device, Queue, Texture, TextureAspect, TextureUsages, TextureView, TextureViewDimension,
+    BindGroup, BindGroupLayout, Buffer, BufferUsages, CommandEncoder, CommandEncoderDescriptor,
+    ComputePipeline, Device, Queue, Texture, TextureAspect, TextureUsages, TextureView,
+    TextureViewDimension,
 };
+
+use crate::cpu_dispatch::CpuBinding;
 
 pub type Error = Box<dyn std::error::Error>;
 
@@ -47,6 +51,7 @@ struct Shader {
     pipeline: ComputePipeline,
     bind_group_layout: BindGroupLayout,
     label: &'static str,
+    cpu_shader: Option<fn(u32, &[CpuBinding])>,
 }
 
 #[derive(Default)]
@@ -97,6 +102,7 @@ pub enum Command {
     // Maybe use tricks to make more ergonomic?
     // Alternative: provide bufs & images as separate sequences
     Dispatch(ShaderId, (u32, u32, u32), Vec<ResourceProxy>),
+    DispatchIndirect(ShaderId, BufProxy, u64, Vec<ResourceProxy>),
     Download(BufProxy),
     Clear(BufProxy, u64, Option<NonZeroU64>),
     FreeBuf(BufProxy),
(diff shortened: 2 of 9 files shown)


src/shaders.rs
```diff
@@ -88,6 +95,9 @@ pub fn full_shaders(device: &Device, engine: &mut Engine) -> Result<FullShaders,
         preprocess::preprocess(shader!("pathtag_reduce"), &full_config, &imports).into(),
         &[BindType::Uniform, BindType::BufReadOnly, BindType::Buffer],
     )?;
+    if use_cpu {
+        engine.set_cpu_shader(pathtag_reduce, cpu_shader::pathtag_reduce);
+    }
```
DJMcNab, line 100: This feels quite noisy, but I don't know if avoiding that is worth the hassle

src/cpu_shader/pathtag_reduce.rs
```diff
@@ -0,0 +1,35 @@
...
+    n_wg: u32,
+    config: &ConfigUniform,
+    scene: &[u32],
+    reduced: &mut [PathMonoid],
+) {
+    let pathtag_base = config.layout.path_tag_base;
+    for i in 0..n_wg {
+        let mut m = PathMonoid::default();
+        for j in 0..WG_SIZE {
+            let tag = scene[(pathtag_base + i * WG_SIZE as u32) as usize + j];
```
DJMcNab, line 20: `(pathtag_base + i * WG_SIZE as u32) as usize` is a constant per loop iteration, right?

Obviously, this should get hoisted anyway, but it might be clearer to split it out

src/cpu_shader/pathtag_reduce.rs
```diff
@@ -0,0 +1,35 @@
...
+            m = m.combine(&PathMonoid::new(tag));
```
DJMcNab, line 21: <details>
<summary>Musings on bad and premature optimisations</summary>
I wonder if these operations could be reordered to improve pipelining. That is, since operation 2 depends on operation 1 (according to the CPU), could we start operation 3/4 whilst 1/2 are ongoing.
My (likely unfounded) intuition around CPU implementations is that loop steps depending on the previous iteration is slow, because the operation must fully complete before the next can begin.

Is there some way to signal to LLVM that these operations are order-invariant?
</details>

DJMcNab (commented):
I don't want to approve, as it's still over my head, and haven't run the code. But this looks really good.

I appreciate the simplifying choice to not support doing `GPU->CPU` results dynamically (even ignoring the extreme performance costs). I hadn't appreciated that choice had been made before, but it makes a lot of sense. It would be good to document that somewhere, possibly in `cpu_shader/mod.rs`?

I have added some thoughts on developer experience

src/engine.rs
```diff
@@ -365,49 +408,101 @@ impl Engine {
...
-                    profiler.begin_scope(shader.label, &mut cpass, device);
-                    cpass.set_pipeline(&shader.pipeline);
-                    cpass.set_bind_group(0, &bind_group, &[]);
-                    cpass.dispatch_workgroups(wg_size.0, wg_size.1, wg_size.2);
-                    #[cfg(feature = "wgpu-profiler")]
-                    profiler.end_scope(&mut cpass);
+                    if let Some(cpu_shader) = shader.cpu_shader {
+                        let resources =
+                            transient_map.create_cpu_resources(&mut self.bind_map, bindings);
+                        cpu_shader(wg_size.0, &resources);
```
armansito, line 414: I recommend adding a brief comment here that explains that the stage will be executed on the CPU synchronously right here. I think it's worth explaining, in a comment, how you foresee this system will work when CPU and GPU execution of pipelines can be mixed in the future.

For example, AIUI as long as no CPU stage depends on a preceding GPU stage, all buffers that are used by downstream GPU pipelines are guaranteed to be fully initialized at the time of dispatch. I would add a note that it is (currently) forbidden for a CPU stage to depend on the output of a GPU stage because the necessary synchronization mechanism is not supported. I would add to the comment that this is not an issue because (currently) the CPU flag is all or nothing and prevents mixing of stages.

src/engine.rs
```diff
@@ -365,49 +408,101 @@ impl Engine {
...
+                        if let CpuBinding::BufferRW(b) = self.bind_map.get_cpu_buf(proxy.id) {
+                            let slice = b.borrow();
+                            let indirect: &[u32] = bytemuck::cast_slice(&slice);
+                            n_wg = indirect[0];
+                        } else {
+                            panic!("indirect buffer missing from bind map");
+                        }
+                        let resources =
+                            transient_map.create_cpu_resources(&mut self.bind_map, bindings);
+                        cpu_shader(n_wg, &resources);
```
armansito, line 448: Similar comment as above. In particular, I would mention that the indirect buffer is expected to have been initialized by a preceding CPU stage.

A question that comes to mind is why you would need indirect dispatch at all when all prior stages execute on the CPU, which means the workgroup count is known at this stage. I think it's worth clarifying in a comment that this system exists primarily for testing.

src/engine.rs
```diff
@@ -853,3 +858,194 @@ impl ResourcePool {
         }
     }
 }
+
+impl BindMapBuffer {
+    fn upload_if_needed(
```
armansito, line 863: Am I understanding this correctly, that once this gets called a materialized CPU buffer gets converted to a GPU buffer and can no longer be used in a CPU pipeline (since `cpu_buf` gets overwritten in the assignment below to `self.buffer = MaterializedBuffer::Gpu(buf);`?

I think the function name should reflect this conversion (or "re-materialization") as it's more than just uploading data from the CPU to the GPU.

src/engine.rs
raphlinus: The reasoning is to make CPU stages have exactly the same interfaces as the GPU counterparts. Obviously if you just want to run the CPU shaders to get results, indirect dispatch is a [karotousen](https://www.tokyoweekender.com/art_and_culture/japanese-culture/weekly-japanese-idiom-karotousen-useless-things/).

I'm in favor of clearly expressing goals, but am less sure that a comment in the code is the best way to do it.

src/engine.rs
raphlinus: I added a comment but was not inspired with a better name. It's not part of the public API.

src/engine.rs
```diff
@@ -853,3 +858,194 @@ impl ResourcePool {
...
+        &self,
+        bind_map: &'a mut BindMap,
+        bindings: &[ResourceProxy],
+    ) -> Vec<CpuBinding> {
+        // First pass is mutable; create buffers as needed
+        for resource in bindings {
+            match resource {
+                ResourceProxy::Buf(buf) => match self.bufs.get(&buf.id) {
+                    Some(TransientBuf::Cpu(_)) => (),
+                    _ => bind_map.materialize_cpu_buf(buf),
```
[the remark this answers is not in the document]
raphlinus: Grr, I remember looking at this case and feeling like I had some reason not to panic, but can't remember. I'll put the panic in, and if it triggers, that will likely jog my memory :)

src/lib.rs
```diff
@@ -204,7 +207,7 @@ impl Renderer {
     pub async fn reload_shaders(&mut self, device: &Device) -> Result<()> {
         device.push_error_scope(wgpu::ErrorFilter::Validation);
         let mut engine = Engine::new();
-        let shaders = shaders::full_shaders(device, &mut engine)?;
+        let shaders = shaders::full_shaders(device, &mut engine, false)?;
```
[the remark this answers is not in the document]
raphlinus: Agreed, having thought about potential use cases. If you've got a mixed cpu/gpu pipeline, then running the previous stages CPU and the downstream "shader under test" on GPU, it makes sense to hot-reload the latter.

src/engine.rs
armansito: I think the comment looks good and explains the side-effect that I wanted to captured.

armansito (approved):

src/engine.rs
armansito: The comments you added look great. The broader design could probably use more documentation in a README.md or module-level comments but that can all wait until this is more hashed out.

PR #362, opened 2023-09-14
[shaders] Expose a binding index set per target language 

### LINKED ISSUE #330: Potentially inaccurate clip_reduce bindings in shader build
I'm tinkering with a few things, and noticed a discrepancy in the bindings for the clip_reduce shader. I'm curious if this is intended, or I'm missing something (hopefully it's a helpful report).

The ShaderInfo and build output from `crate build` under `crates/shaders/` seems to be missing the Uniform binding.

Steps to reproduce:

1. Cloned fresh checkout of vello (probably unnecessary)
2. `cd vello/crates/shaders/`
3. `cargo build` (also similar with --release)
4. The generated shaders.rs file (under target/debug/build/vello_shaders-0f98b3bcd5fa0642/out/shaders.rs, although the hash seems to be randomly generated, it will be one of the two most recently generated directories starting with `vello_shaders-`) shows only 4 bindings, with Uniform missing:

```rust
        clip_reduce: ComputeShader {
            name: Cow::Borrowed("clip_reduce"),
            workgroup_size: [256, 1, 1],
            bindings: Cow::Borrowed(&[BufReadOnly, BufReadOnly, Buffer, Buffer]),
```

However in `src/shaders.rs`, clip_reduce is noted has having a uniform binding:

```rust
    let clip_reduce = engine.add_shader(
        device,
        "clip_reduce",
        preprocess::preprocess(shader!("clip_reduce"), &empty, &imports).into(),
        &[
            BindType::Uniform,
            BindType::BufReadOnly,
            BindType::BufReadOnly,
            BindType::Buffer,
            BindType::Buffer,
        ],
    )?;
```

This is also backed up by `clip_reduce.wgsl` having the config uniform in binding 0, and additionally I'm getting WebGPU createComputePipeline issues without the additional uniform:

```
The buffer type in the shader (BufferBindingType::ReadOnlyStorage) is not compatible with the type in the layout (BufferBindingType::Storage).
 - While validating that the entry-point's declaration for @group(0) @binding(2) matches [BindGroupLayout "clip_reduce bindGroupLayout"]
 - While validating the entry-point's compatibility for group 0 with [BindGroupLayout "clip_reduce bindGroupLayout"]
 - While validating compute stage ([ShaderModule "clip_reduce"], entryPoint: main).
 - While calling [Device].CreateComputePipeline([ComputePipelineDescriptor "clip_reduce pipeline"]).
 ```

I can't find the reason for the discrepancy yet, I didn't look into whether `wgsl::parse_str(&source)?.global_variables` was missing the entry.

armansito:
vello_shaders exposes a list of resource binding types for each
compute stage. This list contains only the bindings that are
actually reachable (obeying WebGPU WGSL validation rules). The
list also contains only the types and no index information is
provided; instead the elements are provided in their order of
declaration in the shader source and the client is expected to
infer the index from the declaration order.

This leads to some issues that are unique to each target language:

1. When targeting WGSL, this scheme is fragile if the shader source
   doesn't keep the index declarations contiguous. Any gaps in the
   indices will lead to incorrect results. This can happen due to a
   programmer error. This can also happen due to bindings that are
   unreachable by the entry point function, for example as a result
   of some refactoring.

2. When targeting MSL, the client needs to know the vello_shader's
   crate's internal binding re-assignment scheme and implement the
   same logic on their end to compute the indices. This stems from
   the fact that WGSL bindings *where indices are scoped to a
   bind group) and MSL bindings (where indices are scoped separately
   by resource type) get assigned from different ranges and
   numerically not the same.

To resolve this, the crate now exposes a list of indices alongside each
backend source.

Additionally, as a cleanup, this PR removes all unused bindings and
rearranges the binding indices to leave no gaps.

Resolves #330


raphlinus (approved):
This looks good to me. I'm curious to see how it will go for something with more complex bindings, like HLSL, but it seems reasonably straightforward so far.

armansito:
> This looks good to me. I'm curious to see how it will go for something with more complex bindings, like HLSL, but it seems reasonably straightforward so far.

HLSL is different but not drastically. Part of the idea with this change is that binding index representation is now separate for each backend language, so it's a (small) step towards supporting more of them. Not using multiple bind groups (thus descriptor sets/tables) definitely simplifies things for now.

PR #363, opened 2023-09-19
[shaders][msl] Enable `zero_initialize_workgroup_memory` setting

armansito:
Zero-initializing workgroup memory works around a Metal driver bug that can cause the tile_alloc and coarse stages to freeze up the system when running natively on Metal due to their use of `workgroupUniformLoad`. See gfx-rs/naga#2482 for more details about this bug.

Zero-initializing workgroup memory also makes the native Metal shaders match the behavior of the current wgpu runner and the invariants of the original WGSL that the shaders were authored in.

In addition, this PR enables the `CONTROL_FLOW_UNIFORMITY` validation check.


raphlinus (approved):
Looks good. And enabling the control flow uniformity check is the right thing to do now that workgroupUniformLoad has landed. Thanks!

PR #364, opened 2023-09-19
[shaders] Option to force read_write access mode on all storage bindings

armansito:
Introduced the `force_rw_storage` feature which applies a transformation that converts all read-only storage bindings to have the `read_write` access mode.

This is being provided as an optional preprocessor transformation to work around the WebGPU standard's limitation on mixed `storage-read` and `storage` usages by a GPU program over the same buffer object (see https://www.w3.org/TR/webgpu/#programming-model-resource-usages).

This limitation makes the WGSL shaders incompatible with (Skia) Graphite's sub-allocating buffer manager when they are run on its Dawn backend. This workaround will be removed once Graphite's buffer management system supports whole buffer bindings.

### DIFF
--- a/crates/shaders/src/compile/preprocess.rs
+++ b/crates/shaders/src/compile/preprocess.rs
@@ -69,7 +69,10 @@ pub fn preprocess(
 
             match directive {
                 if_item @ ("ifdef" | "ifndef" | "else" | "endif") if !directive_is_at_start => {
-                    eprintln!("#{if_item} directives must be the first non_whitespace items on their line, ignoring (line {line_number})");
+                    eprintln!(
+                        "#{if_item} directives must be the first non_whitespace items on \
+                               their line, ignoring (line {line_number})"
+                    );
                     break;
                 }
                 def_test @ ("ifdef" | "ifndef") => {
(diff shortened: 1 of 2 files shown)


crates/shaders/src/compile/preprocess.rs
```diff
@@ -153,6 +166,14 @@ pub fn preprocess(
             if line.starts_with("let ") {
                 output.push_str("const");
                 output.push_str(&line[3..]);
+            } else if let Some(idx) = line.find("var<storage>") {
+                if cfg!(feature = "force_rw_storage") {
+                    let mut line = line.to_string();
+                    line.replace_range(idx..12, "var<storage, read_write>");
```
raphlinus, line 172: This should either be `idx..idx+12` or we should use `starts_with` rather than `find`.

raphlinus (approved):
It feels hacky, so I'm glad there's a plan to remove it, but if it unblocks integration work I'm ok with it.

raphlinus:
The CI problem was fixed by the #360 merge. If you merge or rebase against that, it should clear it up.

crates/shaders/src/compile/preprocess.rs
armansito: Ah good catch. I was getting lucky because the shaders keep the binding index declarations in a separate line. I'll fix this.

PR #367, opened 2023-10-02
Move WGPU engine to its own module

raphlinus:
Disentangle the wgpu implementation of Engine from the mechanisms to create recordings, as the former can in theory be used with multiple back-ends.

Also change the name to WgpuEngine to emphasize that it is one choice among possibly many. Most clients are currently using the `Renderer` abstraction, which is not changed, so the refactor is fairly well hidden.


armansito (approved):
I think this is a good step towards modularization and unblocks other backends/engines.

One thing I would like to see going forward is for the CPU execution logic to be decoupled from the GPU backend-specific engines; I would like to come up with a modular design that allows CPU shaders to interoperate with GPU stages on any backend without having to duplicate concepts like `TransientBindMap` and `TransientBuf` for every engine.

Issue #303, opened 2023-04-06
Stroke rework


raphlinus:
We have quite a bit of work planned on redoing strokes to make them richer and more correct, while retaining performance. There's a considerable amount of discussion in the [Zulip thread](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Stroke.20rework), but I also wanted to snapshot the current state of the plans here.

At a very high level, the plan is to compute offset curve + flattening in the compute shader, expanding the stroke into a fill, then using the same mechanisms downstream to render the fills. This will replace the current distance-field approach. The latter has some nice properties, but also limitations including not handling stroke styles other than round-cap/round-join, and also not handling general (non angle preserving) affine transformations.

The following capabilities are in scope for the compute shader:
* Handling butt/square/round line caps
* Handling miter/bevel/round line joins
* Handling arbitrary affine transformations
* Being correct and robust even for challenging curves (cusps etc)

The following capabilities are, for the moment, out of scope:
* Dashing in the compute shader
* Evolutes as described in [Nehab 2020]

Dashing will be computed during encoding, and will be handled by the [inv_arclen] method.

Evolutes are part of an extremely rigorous approach to spec compliance, where the marked area is defined as a line swept along the source curve, normal to the curve. Very few renderers correctly implement them, and they come at a complexity and performance cost. A serious attempt will be made to ensure results equivalent to a [Minkowski sum] of the path and the disc (or ellipse, under affine transform) of the stroking pen, when the stroke style is round-cap, round-join, but not an absolute guarantee for all adversarial inputs.

The main flattening method will be a variant of the fractional subdivision method currently used, but with Euler spirals as the base curve rather than quadratic Beziers. Offset curve will be done Euler spiral to Euler spiral, using the algorithms in [Cleaner parallel curves with Euler spirals]. The main motivation for Euler spiral is to make the cusp detection and subdivision simple, as this is a tricky recursive subdivision with Bezier-based representations. Another advantage of Euler spiral is that the round caps and joins have an exact representation. Cubic-to-cubic offsetting may also bring challenging issues with numerical stability of the quartic solver; it's only been tested with f64 arithmetic.

The proposed pipeline is:

* pathtag scan, as now. output is `TagMonoid` (scan of 4x reduction).
* mega stage which decodes path segments, performs offset, flattening, generation of caps and joins. Output is: (a) line soup, (b) writing bounding boxes for intra-workgroup bbox, (c) wg-reduced bbox monoid
[...]

PR #368, opened 2023-10-04
Switch to CPU-side stroke encoding

raphlinus:
Don't use GPU-side signed distance field rendering for strokes, but rather expand them on the CPU (using kurbo) during encoding, then render using fill.

This is a significant performance regression, but is on the critical path path for multisampled path rendering and stroke rework (#303).


raphlinus:
One cleanup that probably should be considered is getting rid of the `is_fill` parameter, as that will always be `true` now.

dfrg (approved):
Looks great! Thanks for doing the work to keep stroked glyph outlines functional.

Deferring is_fill removal makes sense for now. We’ll want to update path encoding to match whatever we do for linebender/kurbo#303 anyway so let’s do the work then.

raphlinus:
Thanks for the review. Agree on the is_fill question, I was thinking about that on my morning run. I'll merge the peniko PR then update this to point there.

PR #370, opened 2023-10-06
[test_scenes] Port longpathdash test case from Skia

armansito:
This is a single very long path that results in a large number of line segments when dashing is applied. This demonstrates rendering artifacts.

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -1,5 +1,5 @@
 use crate::{ExampleScene, SceneConfig, SceneParams, SceneSet};
-use vello::kurbo::{Affine, BezPath, Ellipse, PathEl, Point, Rect, Stroke};
+use vello::kurbo::{Affine, BezPath, Cap, Ellipse, PathEl, Point, Rect, Stroke};
 use vello::peniko::*;
 use vello::*;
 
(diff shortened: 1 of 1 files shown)


examples/scenes/src/test_scenes.rs
```diff
@@ -21,26 +21,21 @@ macro_rules! scene {
             function: Box::new($name),
```
DJMcNab, line 21: I'd probably express this case (lines 16-22) in terms of the new addition

DJMcNab (commented):
I've not had a chance to run this yet - can you include a screenshot?

armansito:
> I've not had a chance to run this yet - can you include a screenshot?

Yes of course! 

This is the expected shape with butt caps:
Correct butt cap

This is the same butt cap render showing artifacts when it's zoomed out:
Butt cap - zoomed out artifacts

This is the same shape with round caps (which increases the line segment count) showing artifacts at a moderate zoom level:
Screenshot 2023-10-11 at 10 18 42 AM

The latter can be fixed by increasing the bump buffer sizes. I'm treating this as a pathological case with a very high memory requirement. The zoomed out case appears regardless of the size of the bump buffers so that's probably a separate issue.

examples/scenes/src/test_scenes.rs
armansito: Done.

raphlinus (approved):
Thanks, and also for the cleanup of the scene creation macros. It will be useful to have more examples that stress the buffer sizes, as we make our way through the dynamic memory question.

PR #375, opened 2023-10-08
Implement a workaround for "Map callback was leaked"

DJMcNab:
See https://github.com/gfx-rs/wgpu/issues/4214
and https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Map.20Callbacks.20Leak


armansito (approved):
Thank you for investigating this issue and unblocking folks with a workaround.

PR #377, opened 2023-10-11
Add multisampled antialiasing

raphlinus:
This is ported from the multi branch.

Configuration of antialiasing mode is currently set statically, but could become more dynamic. In addition, the mask LUT is computed and uploaded every frame, rather than being persistent.

### DIFF
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -2,8 +2,10 @@
 
 // Fine rasterizer. This can run in simple (just path rendering) and full
 // modes, controllable by #define.
+//
+// To enable multisampled rendering, turn on both the msaa ifdef and one of msaa8
+// or msaa16.
 
-// This is a cut'n'paste w/ backdrop.
 struct Tile {
     backdrop: i32,
     segments: u32,
@@ -17,8 +19,6 @@ var<uniform> config: Config;
 
 @group(0) @binding(1)
 var<storage> segments: array<Segment>;
-
-#ifdef full
 
 #import blend
 #import ptcl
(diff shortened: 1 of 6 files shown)


shader/fine.wgsl
```diff
@@ -40,6 +40,309 @@ var gradients: texture_2d<f32>;
...
+            let xy0 = select(xy1_in, xy0_in, is_down);
+            let xy1 = select(xy0_in, xy1_in, is_down);
+
+            // Set up data for line rasterization
+            // Note: this is duplicated work if total count exceeds a workgroup.
+            // One alternative is to compute it in a separate dispatch.
+            let dx = abs(xy1.x - xy0.x);
+            let dy = xy1.y - xy0.y;
+            let dy_dxdy = dy / (dx + dy);
+            let a = dx / (dx + dy);
```
armansito, line 184: You may want to define and reuse `idxdy = 1.0 / (dx + dy)` to avoid computing it twice. An optimizing shader compiler could reduce the divisor in both lines into one computation but I don't know if that's guaranteed.

shader/fine.wgsl
```diff
@@ -40,6 +40,309 @@ var gradients: texture_2d<f32>;
...
+            let is_positive_slope = xy1.x >= xy0.x;
+            let sign = select(-1.0, 1.0, is_positive_slope);
+            let xt0 = floor(xy0.x * sign);
+            let c = xy0.x * sign - xt0;
+            // This has a special case in the JS code, but we should just not render
```
armansito, line 189: Maybe link to the JS code or provide some background about it (or remove the comment)?

armansito:
I'm seeing artifacts on cartioid with both msaa8 and msaa16. 

I don't know if I want to necessarily block this PR on this as it can be investigated in a follow up. Unless you want to fix that in this PR, consider setting the default static AA config to `AaConfig::Area` until msaa is stabilized.

Screenshot 2023-10-11 at 4 10 38 PM

shader/fine.wgsl
```diff
@@ -40,6 +40,309 @@ var gradients: texture_2d<f32>;
...
+            }
+            workgroupBarrier();
+            sh_count[th_ix] = count;
+        }
+#ifdef have_uniform
+        let total = workgroupUniformLoad(&sh_count[slice_size - 1u]);
+#else
+        workgroupBarrier();
+        let total = sh_count[slice_size - 1u];
+#endif
```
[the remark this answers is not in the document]
raphlinus: I think we need it to pass uniformity analysis. In any case, this was a holdover from the original multi branch being on an older wgpu that didn't have it, so good catch.

shader/fine.wgsl
raphlinus: I've added a TODO here. When #378 is fixed, we'll want to apply it here as well.

An optimizing compiler could do it, but one that satisfied IEEE arithmetic could not.

shader/fine.wgsl
raphlinus: I've removed the special case, as pixel-aligned horizontal lines are now discarded upstream (see comment around line 127). The comment suggests that it was something I was struggling with :)

raphlinus:
I think the artifacts you're seeing are because of only 4 bits for winding number accumulation, and that test case is a good one for slamming lots of winding number increments into a small area. I'm not 100% sure though. An obvious way to validate that would be to try an 8 bit accumulator.

In any case, agree with keeping area as the default and considering this a followup.

armansito (approved):

PR #382, opened 2023-10-12
Pass the return of `fill_path` in a pointer

Zoxc:
This is just a workaround for https://github.com/gfx-rs/wgpu/issues/4393.


raphlinus (approved):
Thanks, this seems like a reasonable workaround, and not entirely unidiomatic. It reminds me of programming K&R-style C before returning structs was allowed, back when I was a kid.

PR #386, opened 2023-10-13
Get the `fine` CPU shader running

Zoxc:
This hooks up the `fine` CPU shader to the rest of the infrastructure.

- `TransientBindMap` is moved outside of `run_recording` to allow passing CPU buffers from coarse to fine.
- There's partial support for CPU image buffer added.
- The transient map can now hold a `TransientImage::CpuOwned` CPU buffer for `WriteOnce` images. This is used to hold the result of `fine`.
- There's new command `Command::Writeback` which writes `TransientImage::CpuOwned` back into a GPU texture view using the new `writeback.wgsl` shader.

This is based on https://github.com/linebender/vello/pull/383.


shader/writeback.wgsl
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT OR Unlicense
```
armansito, line 1: Why is this shader necessary? Can't we use [`wgpu::CommandEncoder::copy_buffer_to_texture`](https://docs.rs/wgpu/latest/wgpu/struct.CommandEncoder.html#method.copy_buffer_to_texture) or `Command::UploadImage` to perform the upload?

src/engine.rs
```diff
@@ -48,12 +48,28 @@ pub enum ImageFormat {
...
 pub struct ImageProxy {
     pub width: u32,
     pub height: u32,
     pub format: ImageFormat,
     pub id: Id,
+    pub access: ImageAccess,
+}
+
+impl ImageProxy {
+    pub fn byte_size(&self) -> usize {
```
armansito, line 68: Is this used anywhere?

src/cpu_shader/fine.rs
```diff
@@ -154,7 +153,7 @@ fn fine_main(
                     for a in &mut area {
                         *a = 1.0;
                     }
-                    cmd_ix += 2;
+                    cmd_ix += 1;
```
armansito, line 156: Good catch!

src/cpu_shader/fine.rs
```diff
@@ -1,9 +1,9 @@
 // Copyright 2023 The Vello authors
```
armansito, line 1: Much of the PTCL is still not supported and several test scenes will trigger an assertion when `--use-cpu` is enabled. Maybe we should set `let force_gpu_from = Some("fine");` for now until full support is there?

shader/writeback.wgsl
Zoxc: That can't be used with just a texture view.

src/engine.rs
Zoxc: No, I ended up removing its usage.

src/cpu_shader/fine.rs
Zoxc: Done.

shader/writeback.wgsl
xStrom, line 1: Since #456 we have standard copyright headers. Please update this shader file to be consistent with the others.

```suggestion
// Copyright 2023 the Vello Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT OR Unlicense
```

PR #387, opened 2023-10-14
Fix shader compilation on Chrome

Zoxc:
This has some fixes for shader compilation on Chrome.

The `sign` variable shadows the `sign` function, so it's renamed.

Atomics must have a `read_write` binding so `BumpAllocators` usage is changed.

### DIFF
--- a/shader/path_tiling.wgsl
+++ b/shader/path_tiling.wgsl
@@ -8,7 +8,7 @@
 #import tile
 
 @group(0) @binding(0)
-var<storage> bump: BumpAllocators;
+var<storage, read_write> bump: BumpAllocators;
 
 @group(0) @binding(1)
 var<storage> seg_counts: array<SegmentCount>;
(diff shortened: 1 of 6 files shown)


shader/path_tiling.wgsl
```diff
@@ -62,19 +62,19 @@ fn main(
         let idxdy = 1.0 / (dx + dy);
         var a = dx * idxdy;
         let is_positive_slope = s1.x >= s0.x;
-        let sign = select(-1.0, 1.0, is_positive_slope);
-        let xt0 = floor(s0.x * sign);
-        let c = s0.x * sign - xt0;
+        let sign_ = select(-1.0, 1.0, is_positive_slope);
```
raphlinus, line 65: Makes sense but I'm not in love. I think I might prefer `xsign` (or possibly `x_sign` as it represents the sign of dx.

raphlinus (approved):
Thanks, these are useful fixes, and it reminds me we should do more testing on the web backend.

PR #388, opened 2023-10-17
[test_scenes] Tests scenes for strokes

armansito:
Added two new test scenes that exercise stroke styles and some edge conditions:

1. stroke_styles: Renders a combination of cap and join styles

2. tricky_strokes: Renders cubic strokes that demonstrate certain edge conditions, such as cusps and co-linear control points. Several of these tests are commented out as they trigger an infinite recursion in the kurbo curve fitting logic and crash the example.

These should probably get copied as unit tests in kurbo as part of a fix for the infinite recursions.


raphlinus (approved):
Useful cases, thanks!

All the broken_cubic cases are fixed by kurbo#318, and I've included those in a test. Additionally, the artifact that appears is a numerical robustness issue that was fixed in msaa fine but not yet ported to the area aa. This can be done as an isolated patch, but I'm also tempted to switch the segment storage from 24 to 8 bytes and fold the change into that; the msaa is already written in that style and does not read the y_edge value, so making that change would get rid of some duplicated logic overlapping the robustness concern.

PR #389, opened 2023-10-21
New "style" encoding to replace "linewidth"

armansito:
This PR replaces the existing floating-point "linewidth" stream with a new "style" stream based on the proposal in #303. 

- All pipelines up to `flatten` have been wired up to extract the fill style using the new encoding.
- The encoding crate now encodes fills using the new scheme. Strokes still get converted to fills on the CPU but will be updated to optionally use the new encoding in follow-up changes.
- Added a new test scene to specifically exercise even-odd vs non-zero fill rules.

### DIFF
--- a/crates/encoding/src/path.rs
+++ b/crates/encoding/src/path.rs
@@ -267,8 +267,8 @@ impl PathTag {
     /// Path marker.
     pub const PATH: Self = Self(0x10);
 
-    /// Line width setting.
-    pub const LINEWIDTH: Self = Self(0x40);
+    /// Style setting.
+    pub const STYLE: Self = Self(0x40);
 
     /// Bit for path segments that are represented as f32 values. If unset
     /// they are represented as i16.
@@ -316,8 +316,8 @@ pub struct PathMonoid {
     pub pathseg_ix: u32,
     /// Offset into path segment stream.
     pub pathseg_offset: u32,
-    /// Index into linewidth stream.
-    pub linewidth_ix: u32,
+    /// Index into style stream.
+    pub style_ix: u32,
     /// Index of containing path.
     pub path_ix: u32,
 }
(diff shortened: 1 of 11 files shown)


crates/encoding/src/path.rs
```diff
@@ -2,10 +2,184 @@
...
+    /// - flags: u16 - encodes fill vs stroke, even-odd vs non-zero fill mode for fills and cap
+    ///                and join style for strokes. See the FLAGS_* constants below for more
+    ///                information.
+    /// ```text
+    /// flags: |style|fill|join|start cap|end cap|reserved|
+    ///  bits:  0     1    2-3  4-5       6-7     8-15
+    /// ```
+    ///
+    /// - miter_limit: u16 - The miter limit for a stroke, encoded in binary16 (half) floating
+    ///                      point representation. This field is on meaningful for the `Join::Miter`
```
dfrg, line 28: on -> only

crates/encoding/src/path.rs
```diff
@@ -2,10 +2,184 @@
...
+    pub const FLAGS_START_CAP_BITS_ROUND: u32 = 0x0800_0000;
+    pub const FLAGS_END_CAP_BITS_BUTT: u32 = 0;
+    pub const FLAGS_END_CAP_BITS_SQUARE: u32 = 0x0100_0000;
+    pub const FLAGS_END_CAP_BITS_ROUND: u32 = 0x0200_0000;
+
+    pub const FLAGS_START_CAP_MASK: u32 = 0x0C00_0000;
+    pub const FLAGS_END_CAP_MASK: u32 = 0x0300_0000;
+    pub const MITER_LIMIT_MASK: u32 = 0xFFFF;
+
+    pub fn from_fill(fill: &Fill) -> Self {
```
dfrg, line 69: I think it’d be slightly cleaner to take this parameter by value.

crates/encoding/src/path.rs
```diff
@@ -2,10 +2,184 @@
...
+        };
+        let miter_limit = crate::math::f32_to_f16(stroke.miter_limit as f32) as u32;
+        Self {
+            flags_and_miter_limit: style | join | start_cap | end_cap | miter_limit,
+            line_width: stroke.width as f32,
+        }
+    }
+
+    #[cfg(test)]
+    fn get_fill(&self) -> Option<Fill> {
```
dfrg, line 105: Small style nit: I know these are only for tests now but it seems likely they’ll become generally useful enough to be exposed at some point and the `get_` prefix is generally elided for accessors.

dfrg (approved):
A few tiny style nits but otherwise lgtm, thanks!

crates/encoding/src/path.rs
armansito: Done.

crates/encoding/src/path.rs
armansito: Done.

crates/encoding/src/path.rs
armansito: Done.

Issue #390, opened 2023-10-21
[MSAA] Blocky artifacts with the even-odd fill rule


armansito:
I added a new test scene for fill rules in #389. When I tried it with both 16xMSAA and 8xMSAA modes I observed some visual artifacts.

This is how it looks with Analytic AA:
<img width="316" alt="Screenshot 2023-10-21 at 1 35 49 PM" src="https://github.com/linebender/vello/assets/6933700/6e3dcfb5-1140-4730-91a6-5437fa5ad7a0">

This is with MSAA:
<img width="312" alt="Screenshot 2023-10-21 at 1 33 16 PM" src="https://github.com/linebender/vello/assets/6933700/ae026137-c9a1-43b5-9b83-1eea2796e2d6">

raphlinus:
This is consistent with the even-odd rule just not being handled correctly in the msaa fine rasterizer, especially as it looks like the logic in coarse is doing the right thing for solid blocks. It wouldn't be hard to add a case, but the best solution is to write a specialized version that uses 1 bit per sample to accumulate winding numbers, using xor. The performance of that should be excellent, with no extra cost to using 16 bits over 8.

PR #391, opened 2023-10-24
Fixes to winding number accumulation

### LINKED ISSUE #380: [MSAA] Artifacts in cardioid with 8xMSAA and 16xMSAA
The msaa8 and msaa16 AA config's introduced in #377 render with artifacts:

From @raphlinus: "I think the artifacts you're seeing are because of only 4 bits for winding number accumulation, and that test case is a good one for slamming lots of winding number increments into a small area. I'm not 100% sure though. An obvious way to validate that would be to try an 8 bit accumulator."

raphlinus:
This patch fixes rendering artifacts related to winding number accumulation. The first problem is that 4 bits per pixel is not adequate for all test scenes (the cardioid one in particular exercises this). The second problem is that we didn't have a case for even-odd.

This patch is expected to regress performance for the nonzero fill rule, but improve it for even-odd. Some of the performance loss is decreased occupancy because of a larger shared memory footprint for the accumulation, some is because of increased atomic traffic. We should do some performance work as a followup, first characterizing the performance impact, then doing some mitigation. One promising idea is to specialize further based on the number of segments in the slice, using 4 bit accumulation when we can guarantee that won't overflow.

This will have a (small) merge conflict with #382. If that's landed first, I'll happily update this one.

Fixes #380 and #390.

### DIFF
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -186,9 +186,9 @@ fn fill_path_ms(fill: CmdFill, wg_id: vec2<u32>, local_id: vec2<u32>) -> array<f
             let idxdy = 1.0 / (dx + dy);
             var a = dx * idxdy;
             let is_positive_slope = xy1.x >= xy0.x;
-            let sign = select(-1.0, 1.0, is_positive_slope);
-            let xt0 = floor(xy0.x * sign);
-            let c = xy0.x * sign - xt0;
+            let x_sign = select(-1.0, 1.0, is_positive_slope);
+            let xt0 = floor(xy0.x * x_sign);
+            let c = xy0.x * x_sign - xt0;
             let y0i = floor(xy0.y);
             let ytop = y0i + 1.0;
             let b = min((dy * c + dx * (ytop - xy0.y)) * idxdy, ONE_MINUS_ULP);
@@ -198,12 +198,12 @@ fn fill_path_ms(fill: CmdFill, wg_id: vec2<u32>, local_id: vec2<u32>) -> array<f
             if robust_err != 0.0 {
                 a -= ROBUST_EPSILON * sign(robust_err);
             }
-            let x0i = i32(xt0 * sign + 0.5 * (sign - 1.0));
+            let x0i = i32(xt0 * x_sign + 0.5 * (x_sign - 1.0));
             // Use line equation to plot pixel coordinates
 
             let zf = a * f32(sub_ix) + b;
             let z = floor(zf);
-            let x = x0i + i32(sign * z);
+            let x = x0i + i32(x_sign * z);
             let y = i32(y0i) + i32(sub_ix) - i32(z);
             var is_delta: bool;
             // We need to adjust winding number if slope is positive and there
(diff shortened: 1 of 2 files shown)


shader/fine.wgsl
```diff
@@ -63,12 +63,13 @@ var<storage> mask_lut: array<u32, 2048u>;
 let WG_SIZE = 64u;
 var<workgroup> sh_count: array<u32, WG_SIZE>;
 
-// This is 8 winding numbers packed to a u32, 4 bits per sample
-var<workgroup> sh_winding: array<atomic<u32>, 32u>;
+// This is 4 winding numbers packed to a u32, 8 bits per sample
+var<workgroup> sh_winding: array<atomic<u32>, 64u>;
```
armansito, line 67: As we discussed today, please document not only the packing pattern but what each shared buffer semantically represents in the hierarchy of counts.

shader/fine.wgsl
```diff
@@ -260,88 +269,315 @@ fn fill_path_ms(fill: CmdFill, wg_id: vec2<u32>, local_id: vec2<u32>) -> array<f
...
-            let xored1 = (expected_zero * 0x11111111u) ^ samples1;
+            let xored1 = (expected_zero * 0x1010101u) ^ samples1;
             let xored1_2 = xored1 | (xored1 >> 1u);
             let xored2 = (xored0_2 & 0xAAAAAAAAu) | (xored1_2 & 0x55555555u);
             let xored4 = xored2 | (xored2 * 4u);
-            area[i] = f32(countOneBits(xored4 & 0xCCCCCCCCu)) * 0.0625;
+            let xored16 = xored4 | (xored4 * 16u);
+            area[i] = f32(countOneBits(xored16 & 0xC0C0C0C0u)) * 0.125;
 #endif
+#ifdef msaa16
```
armansito, line 353: Please try to document your invariants around the SWAR math as best as you can. Perhaps even add a note that refers the reader to the even-odd version as it's easier to understand.

src/lib.rs
```diff
@@ -73,7 +73,7 @@ enum AaConfig {
 
 /// Configuration of antialiasing. Currently this is static, but could be switched to
 /// a launch option or even finer-grained.
-const ANTIALIASING: AaConfig = AaConfig::Area;
+const ANTIALIASING: AaConfig = AaConfig::Msaa16;
```
armansito, line 76: Do we want to make `Msaa16` the default yet? Perhaps this should remain as set to area for now as we may not want to get hit by the performance cost of increased workgroup memory in Msaa16 mode.

That said, I think I would like to make this configurable at runtime. I think a reasonable approach is to introduce two levels of settings: a `RendererOptions` setting that tells the library which pipeline variants to compile/enable and a `RenderParams` setting that selects a particular AA mode (which can eventually have a "dynamic" option if we decide to support that). That way we can run the test suite in various modes without having to re-compile the code.

If you agree with that abstraction, I'd be happy to wire that up in a follow-up.

armansito (commented):
Thank you for the fixes. I think the new approach makes sense but as we discussed today, a lot of the logic is very dense/involved and could use some clarification.

LGTM otherwise.

src/lib.rs
raphlinus: I think it's good making `Msaa16` the default at least temporarily, if for no other reason than to exercise it. I also agree that it should be configurable, but the mechanism to do that is nontrivial (thanks for agreeing to take it on!) and seems like it can be done separately from this PR.

shader/fine.wgsl
```diff
@@ -63,13 +63,38 @@ var<storage> mask_lut: array<u32, 2048u>;
...
+// relative to the pixel immediately to the left), then expanded using
+// prefix sum and reusing the same storage.
+//
+// The encoding and packing is the same as `sh_winding_y`. For the even-odd
+// case, only the first 16 values are used, and each word stores packed
+// parity values for one row of pixels.
+var<workgroup> sh_winding: array<atomic<u32>, 64u>;
+// This array contains winding numbers of multiple sample points within
+// a pixel, relative to the winding number of the top left corner of the
+// pixels. The encoding and packing is the same as `sh_winding_y`.
```
armansito, line 96: > pixels. The encoding...

of the pixel (singular) or is this relative to a group of pixels?

armansito (approved):
Thank you very much for the documentation!

shader/fine.wgsl
raphlinus: I went back and forth between singular and plural. Fixed.

PR #392, opened 2023-10-24
Work around broken JS deps

raphlinus:
CI is failing because of some updated ecosystem crates that we don't care about at all. This patch is a hammer to get it to stop failing, but is clearly not the right thing to do.

### DIFF
--- a/examples/with_winit/Cargo.toml
+++ b/examples/with_winit/Cargo.toml
@@ -48,3 +48,6 @@ console_log = "1"
 wasm-bindgen-futures = "0.4.33"
 web-sys = { version = "0.3.60", features = [ "HtmlCollection", "Text" ] }
 getrandom = { version = "0.2.10", features = ["js"] }
+rustls-webpki = "=0.101.4"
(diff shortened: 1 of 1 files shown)


examples/with_winit/Cargo.toml
```diff
@@ -48,3 +48,6 @@ console_log = "1"
 wasm-bindgen-futures = "0.4.33"
 web-sys = { version = "0.3.60", features = [ "HtmlCollection", "Text" ] }
 getrandom = { version = "0.2.10", features = ["js"] }
+rustls-webpki = "=0.101.4"
+sct = "=0.7.0"
+ureq = "=2.7.1"
```
DJMcNab, line 53:
```rust
ring = { version = "0.17.5", features = ["wasm32_unknown_unknown_js"] }
```
works for me. I suspect this is the correct behaviour. I'd also add a comment like:
`# Tell our dependencies that we use wasm-bindgen and are running in a JavaScript host`

PR #395, opened 2023-10-25
Disable download feature on wasm

### LINKED ISSUE #394: CI is broken because dependencies upgraded
This command is failing and can be replicated locally after `cargo update`:

```rust
RUSTFLAGS=--cfg=web_sys_unstable_apis cargo check -p with_winit --target wasm32-unknown-unknown
```

The error is in the compilation of ring 0.17.5

```
warning: error: unable to create target: 'No available targets are compatible with triple "wasm32-unknown-unknown"'
warning: 1 error generated.

error: failed to run custom build command for `ring v0.17.5`
```

That in turn is depended on by rustls 0.21.8 and rustls-webpki 0.101.7, and sct 0.7.1, which come in through ureq, which is a dependency in our `scenes` crate to support the download feature.

I posted #392 yesterday, but I'm now thinking the proper fix is to remove the ureq dependency on wasm.

raphlinus:
The download feature caused CI failures on wasm, due to dependencies breaking in ways difficult to diagnose. I searched the issues for the relevant dependencies and didn't find other people running into the same problem.

Since I worry that these deps may continue to be unstable, I think the best thing is to disable them on wasm builds. This shouldn't affect any functionality we actually care about.

Longer term, we probably want to move these more sophisticated examples out of the vello crate proper.

Fixes #394

### DIFF
--- a/examples/scenes/Cargo.toml
+++ b/examples/scenes/Cargo.toml
@@ -20,6 +20,7 @@ rand = "0.8.5"
 instant = { workspace = true }
 
 # Used for the `download` command
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 byte-unit = "4.0"
 dialoguer = "0.10"
 ureq = "2.6"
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
I don't think it's even possible to provide the command line arguments on wasm targets we support, so :+1:

Ideally this wouldn't be broken, but that doesn't matter

PR #396, opened 2023-10-25
Fix strokes by updating kurbo dep

raphlinus:
Version 0.10.4 of kurbo fixes numerical robustness problems with strokes. This patch bumps to that version, and also moves `_broken_cubics` into `tricky_cubics` so we can exercise those fixes.


dfrg (approved):
Looks good, thanks!

PR #398, opened 2023-10-26
Update to wgpu 0.18

DJMcNab:
All examples seem to work as expected

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
 
     "examples/headless",
     "examples/with_winit",
-    "examples/with_bevy",
+    # "examples/with_bevy", # Disable for now until bevy is using wpgu 0.18
     "examples/run_wasm",
     "examples/scenes",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ fello = { git = "https://github.com/dfrg/fount", rev = "dadbcf75695f035ca46766bf
 peniko = { git = "https://github.com/linebender/peniko", rev = "629fc3325b016a8c98b1cd6204cb4ddf1c6b3daa" }
 
 # NOTE: Make sure to keep this in sync with the version badge in README.md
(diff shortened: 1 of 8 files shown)


waywardmonkeys (approved):
So happy to see that `timestamp_period` going away.

raphlinus (changes requested):
I am sadly still getting "map callback was leaked" errors, plus "failed to get surface texture: Timeout" on my Windows 10, AMD 5700 XT system. That may be some kind of problem with my setup, though it did work with wgpu 0.16 as I recall.

Further, it is breaking Metal.

```
Caused by:
    In Device::create_compute_pipeline
      note: label = `fine`
    Internal error: Metal: program_source:454:16: error: use of undeclared identifier '_e7'
    float fb = _e7;
```

Doing a little bit of digging, that's this this fragment of [blend.wgsl](https://github.com/linebender/vello/blob/c9faf2dbb79ae652c19302a913bbb4526b889052/shader/shared/blend.wgsl#L228-L239):

```wgsl
// Apply general compositing operation.
// Inputs are separated colors and alpha, output is premultiplied.
fn blend_compose(
    cb: vec3<f32>,
    cs: vec3<f32>,
    ab: f32,
    as_: f32,
    mode: u32
) -> vec4<f32> {
    var fa = 0.0;
    var fb = 0.0;
    switch mode {
```

being compiled to this:

```msl
metal::float4 blend_compose(
    metal::float3 cb_6,
    metal::float3 cs_6,
    float ab,
    float as,
    uint mode_1
) {
    float fa = 0.0;
    float fb = _e7;
```

Seems like some kind of constant folding gone wrong. So I don't think we should merge this until we have a workaround.

raphlinus (approved):
I verified that this is working on mac with naga 0.14.1. I didn't try to resolve the merge conflicts myself, just ran the branch. Hopefully those won't be too difficult.

It'll be nice to be on wgpu 0.18, as there are a number of things we want to investigate, including slow shader compilation on D3D12.

PR #399, opened 2023-10-27
Transition all pipelines to new `draw_flags` field for fill rule

armansito:
This migrates all pipelines downstream of `flatten` away from the old `linewidth` encoding which was only being used for the fill rule, which now gets stored in a `draw_flags` field. I also cleaned up some leftover linewidth transform logic which was no longer used and wired up the new fill rule flag to the CPU version of the `coarse` stage.

This continues the work outlined in #303. For more details please refer to the individual commit messages.

### DIFF
--- a/shader/coarse.wgsl
+++ b/shader/coarse.wgsl
@@ -315,11 +315,8 @@ fn main(
             // If this draw object represents an even-odd fill and we know that no line segment
             // crosses this tile and then this draw object should not contribute to the tile if its
             // backdrop (i.e. the winding number of its top-left corner) is even.
-            //
-            // NOTE: A clip should never get encoded with an even-odd fill.
-            let even_odd_discard = n_segs == 0u && even_odd && (abs(tile.backdrop) & 1) == 0;
-            let include_tile = !even_odd_discard
-                    && (n_segs != 0u || (tile.backdrop == 0) == is_clip || is_blend);
+            let backdrop_clear = select(tile.backdrop, abs(tile.backdrop) & 1, even_odd) == 0;
+            let include_tile = n_segs != 0u || (backdrop_clear == is_clip) || is_blend;
             if include_tile {
                 let el_slice = el_ix / 32u;
                 let el_mask = 1u << (el_ix & 31u);
--- a/shader/coarse.wgsl
+++ b/shader/coarse.wgsl
@@ -82,8 +82,8 @@ fn alloc_cmd(size: u32) {
     }
 }
 
-fn write_path(tile: Tile, tile_ix: u32, linewidth: f32) -> bool {
-    let even_odd = linewidth < -1.0;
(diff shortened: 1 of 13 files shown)


shader/coarse.wgsl
```diff
@@ -310,7 +306,20 @@ fn main(
...
+            let draw_flags = info_bin_data[di];
+            let even_odd = (draw_flags & DRAW_INFO_FLAGS_FILL_RULE_BIT) != 0u;
+            let n_segs = tile.segment_count_or_ix;
+
+            // If this draw object represents an even-odd fill and we know that no line segment
+            // crosses this tile and then this draw object should not contribute to the tile if its
+            // backdrop (i.e. the winding number of its top-left corner) is even.
+            //
+            // NOTE: A clip should never get encoded with an even-odd fill.
+            let even_odd_discard = n_segs == 0u && even_odd && (abs(tile.backdrop) & 1) == 0;
```
raphlinus, line 320: In my first pass, I was worried that this would be incorrect for even-odd clips. But I now see this can't be encoded, so my concern is more one of style than correctness.

If we ever do support even-odd clips (which is `eoclip` in PostScript and Core Graphics, `W*` in PDF), then we'd want to make the change. I think a version that gets this correct is just as concise, and potentially clearer, so suggest something along these lines:

```rust
    let backdrop_clear = select(tile.backdrop, tile.backdrop & 1, even_odd) == 0;
    let include_tile = n_segs != 0u || (backdrop_clear == is_clip) || is_blend;
```

But I wouldn't block the PR on this, it is correct as written.

src/cpu_shader/coarse.rs
```diff
@@ -238,22 +242,34 @@ fn coarse_main(
                         let blend = scene[dd as usize];
                         is_blend = blend != BLEND_CLIP;
                     }
+
+                    let draw_flags = info_bin_data[di as usize];
+                    let even_odd = (draw_flags & DRAW_INFO_FLAGS_FILL_RULE_BIT) != 0;
                     let n_segs = tile.segment_count_or_ix;
-                    let include_tile = n_segs != 0 || (tile.backdrop == 0) == is_clip || is_blend;
+
+                    // If this draw object represents an even-odd fill and we know that no line segment
```
raphlinus, line 250: See above comment; if the change is made there, it should be consistent here.

raphlinus (approved):
This all looks good except for a possible tweak to the `include_tile` logic, see inline. I especially appreciate the code simplification and reduced divergence in coarse. I think divergence could be further improved by grouping the write_paths for the different draw object types, but I'm not sure how much of a real performance improvement that would be.

shader/coarse.wgsl
armansito: I like your formulation better than what I have as it is concise and doesn't rule out even-odd clips (which I wasn't aware of until now).

PR #400, opened 2023-10-31
Support setting AA method dynamically

armansito:
This replaces the static anti-aliasing setting with a dynamic option in the form of two new settings (one used during pipeline creation and one used during a render):

- The `FullShaders` collection now maintains up to 3 `fine` stage pipeline variants. This can be driven using a new optional `RendererOptions` field called `preferred_antialiasing_method`, which determines which fine stage pipeline variants should get  instantiated at start up.

- `RenderParams` now requires an `AaConfig` which selects which `fine` stage variant to use.

- Added a new key binding (`M`) to the `with_winit` example to dynamically cycle through all 3 AA methods.

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -114,7 +114,8 @@ fn run(
     let mut vsync_on = true;
 
     const AA_CONFIGS: [AaConfig; 3] = [AaConfig::Area, AaConfig::Msaa8, AaConfig::Msaa16];
-    let mut aa_config_ix = 0;
+    // We allow cycling through AA configs in either direction, so use a signed index
+    let mut aa_config_ix: i32 = 0;
 
     let mut frame_start_time = Instant::now();
     let start = Instant::now();
@@ -134,8 +135,8 @@ fn run(
     }
     let mut profile_stored = None;
     let mut prev_scene_ix = scene_ix - 1;
-    let mut prev_aa_config_ix = aa_config_ix;
     let mut profile_taken = Instant::now();
+    let mut modifiers = ModifiersState::default();
     // _event_loop is used on non-wasm platforms to create new windows
     event_loop.run(move |event, _event_loop, control_flow| match event {
         Event::WindowEvent {
@@ -150,6 +151,7 @@ fn run(
             }
             match event {
                 WindowEvent::CloseRequested => *control_flow = ControlFlow::Exit,
+                WindowEvent::ModifiersChanged(m) => modifiers = *m,
                 WindowEvent::KeyboardInput { input, .. } => {
                     if input.state == ElementState::Pressed {
                         match input.virtual_keycode {
@@ -180,7 +182,11 @@ fn run(
                                 stats.clear_min_and_max();
                             }
                             Some(VirtualKeyCode::M) => {
-                                aa_config_ix = (aa_config_ix + 1) % AA_CONFIGS.len();
+                                aa_config_ix = if modifiers.shift() {
+                                    aa_config_ix.saturating_sub(1)
+                                } else {
+                                    aa_config_ix.saturating_add(1)
+                                };
                             }
                             Some(VirtualKeyCode::P) => {
                                 if let Some(renderer) = &renderers[render_state.surface.dev_id] {
@@ -333,27 +339,15 @@ fn run(
 
             // Allow looping forever
             scene_ix = scene_ix.rem_euclid(scenes.scenes.len() as i32);
+            aa_config_ix = aa_config_ix.rem_euclid(AA_CONFIGS.len() as i32);
+
             let example_scene = &mut scenes.scenes[scene_ix as usize];
-            let mut reset_title = false;
             if prev_scene_ix != scene_ix {
                 transform = Affine::IDENTITY;
                 prev_scene_ix = scene_ix;
-                reset_title = true;
-            }
-            if prev_aa_config_ix != aa_config_ix {
-                prev_aa_config_ix = aa_config_ix;
-                reset_title = true;
-            }
-            if reset_title {
-                let aa_str = match AA_CONFIGS[aa_config_ix] {
-                    AaConfig::Area => "Analytic Area",
-                    AaConfig::Msaa16 => "16xMSAA",
-                    AaConfig::Msaa8 => "8xMSAA",
-                };
-                render_state.window.set_title(&format!(
-                    "Vello demo - {} - AA method: {}",
-                    example_scene.config.name, aa_str
-                ));
+                render_state
+                    .window
+                    .set_title(&format!("Vello demo - {}", example_scene.config.name));
             }
             let mut builder = SceneBuilder::for_fragment(&mut fragment);
             let mut scene_params = SceneParams {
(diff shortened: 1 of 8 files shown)


DJMcNab (commented):
Mostly some style considerations

src/lib.rs
```diff
@@ -99,43 +93,56 @@ pub struct RenderParams {
...
     pub timestamp_period: f32,
+
+    /// If true, run all stages up to fine rasterization on the CPU.
+    /// TODO: Consider evolving this so that the CPU stages can be configured dynamically via
+    /// `RenderParams`.
     pub use_cpu: bool,
+
+    /// The anti-aliasing specialization that should be used when creating the pipelines. `None`
+    /// initializes all variants.
+    pub preferred_antialiasing_method: Option<AaConfig>,
```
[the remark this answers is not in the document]
raphlinus: A partial answer to this is that we think we'll probably want to end up rendering glyphs with area AA and general paths with MSAA. But that speaks to a setting that's even more dynamic than per-render.

src/render.rs
```diff
@@ -414,10 +417,10 @@ impl Render {
     pub fn record_fine(&mut self, shaders: &FullShaders, recording: &mut Recording) {
         let fine_wg_count = self.fine_wg_count.take().unwrap();
         let fine = self.fine_resources.take().unwrap();
-        match ANTIALIASING {
+        match fine.aa_config {
             AaConfig::Area => {
                 recording.dispatch(
-                    shaders.fine,
+                    shaders.fine_area.expect("unsupported AA mode: area"),
```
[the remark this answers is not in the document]
raphlinus: Yeah, I think "shaders not configured to support AA mode" would be a better string.

src/render.rs
```diff
@@ -432,16 +435,21 @@ impl Render {
...
-                    let mask_lut = match ANTIALIASING {
+                    let mask_lut = match fine.aa_config {
                         AaConfig::Msaa16 => crate::mask::make_mask_lut_16(),
                         AaConfig::Msaa8 => crate::mask::make_mask_lut(),
                         _ => unreachable!(),
                     };
                     let buf = recording.upload("mask lut", mask_lut);
                     self.mask_buf = Some(buf.into());
                 }
+                let shader = match fine.aa_config {
```
raphlinus, line 446: Personally I'd call this `fine_shader`.

src/shaders.rs
```diff
@@ -305,38 +299,51 @@ pub fn full_shaders(device: &Device, engine: &mut WgpuEngine) -> Result<FullShad
...
+        const MSAA_INFO: [Option<(&str, &str)>; 3] = [
+            None,
+            Some(("fine_msaa8", "msaa8")),
+            Some(("fine_msaa16", "msaa16")),
+        ];
+        let mut pipelines = [None, None, None];
+        for (i, aa_mode) in AA_MODES.iter().enumerate() {
+            if options
+                .preferred_antialiasing_method
+                .as_ref()
```
raphlinus, line 325: Could the `as_ref` be eliminated by comparing `m != *aa_mode`? It's `Copy` for this reason.

raphlinus (approved):
Looks good. I agree with most of Daniel's comments, and just had some golfing to make the logic hopefully a little simpler - it feels just a bit clunky and I'm slightly worried about what will happen as the number of permutations grows. But this is definitely useful to have!

Of course, we might reduce permutations by having multiple aa methods present in the same shader, but one nice thing about this setup is that we can experiment with the effect of shared memory size (with effect on occupancy but also likely a nontrivial effect from just clearing the memory, especially as this isn't always optimized well).

armansito:
> Looks good. I agree with most of Daniel's comments, and just had some golfing to make the logic hopefully a little simpler - it feels just a bit clunky and I'm slightly worried about what will happen as the number of permutations grows. But this is definitely useful to have!
> 
> Of course, we might reduce permutations by having multiple aa methods present in the same shader, but one nice thing about this setup is that we can experiment with the effect of shared memory size (with effect on occupancy but also likely a nontrivial effect from just clearing the memory, especially as this isn't always optimized well).

I agree that this is a little clunky and it would be good to figure out a better way to specify the desired pipeline permutations. For now, this is scoped to 3 AA settings and I imagine it wouldn't be too hard to extend this to an additional "dynamic" mode if we were to explore that.

examples/with_winit/src/lib.rs
```diff
@@ -356,6 +374,7 @@ fn run(
                     .unwrap_or(Color::BLACK),
                 width,
                 height,
+                antialiasing_method: aa_config,
```
DJMcNab, line 377: I'd probably inline this (so that the comment above about base_colour logic is in the right place)

src/shaders.rs
```diff
@@ -305,38 +298,47 @@ pub fn full_shaders(device: &Device, engine: &mut WgpuEngine) -> Result<FullShad
...
+            (aa_support.msaa16, Some(("fine_msaa16", "msaa16"))),
+        ];
+        let mut pipelines = [None, None, None];
+        for (i, (enabled, msaa_info)) in aa_modes.iter().enumerate() {
+            if !enabled {
+                continue;
+            }
+            let (range_end_offset, label, aa_config) = match *msaa_info {
+                Some((label, config)) => (0, label, Some(config)),
+                None => (1, "fine_area", None),
```
DJMcNab, line 327: (I had a big comment written out here, which GitHub mobile helpfully deleted)

But broadly, I was wondering if all of this should be inlined into aa_mode, i.e. make it a `(bool, usize, &str, Option<&str>)`.

If we were doing that, I'd then replace the usize offset with the pipeline resources, so fine_area would be:
`(support.area, &fine_resources[..7], "fine_area", None)`

(Where 7 might not be the right number, but trying to check it lead to the deletion of the previous version of this comment)

DJMcNab (commented):
Not approving as I haven't had a chance to run this yet, but the code looks good.

There are a couple of small style points, one of which isn't necessarily an improvement (especially if it makes rustfmt break up the aa_modes across 20 lines, which it might accidentally do)

examples/with_winit/src/lib.rs
```diff
@@ -180,7 +182,11 @@ fn run(
                                 stats.clear_min_and_max();
                             }
                             Some(VirtualKeyCode::M) => {
-                                aa_config_ix = (aa_config_ix + 1) % AA_CONFIGS.len();
+                                aa_config_ix = if modifiers.shift() {
```
raphlinus, line 185: Personally I would keep the mod n behavior (as the reverse is less discoverable), but that's a style preference. Note though that we do have +/- and mod n for cycling through the test scenes.

raphlinus (approved):
This needs some merge conflict fixing, but hopefully that should be easy. I haven't run this version but have run a previous version, so I'm ok as long as things didn't get broken. And if they did, that can be a followup PR :)

examples/with_winit/src/lib.rs
armansito: `aa_config` gets reused below so I wanted to have a single binding for it. I moved things around so that the comment is positioned better.

src/shaders.rs
armansito: Done. I just folded the offset into the tuple to keep the column width manageable.

examples/with_winit/src/lib.rs
armansito: That's fair, the scene cycling logic is actually structured this way (though bound to arrow keys instead of using a modifier) so I just followed that here for consistency.

PR #401, opened 2023-11-01
Improve numerical robustness in path tiling

raphlinus:
This PR fixes a "blocky artifact" bug in area antialiasing, makes changes in the way path segments are represented, and introduces a test scene crafted to expose numerical robustness artifacts.

The artifact was due to a discrepancy in the way numerical robustness was handled between msaa and area antialiasing. This patch moves that logic into tiling, which should help performance as well as consistency. In so doing, it changes the representation from a viewport-relative coordinate of point0 and (point1 - point0) to one where both points are relative to the top left of the tile. That in turn is preparation for a change to f16 coordinates (not done in this PR).

Here is an informal description of the numerical robustness logic for path tiling.

A grid-aligned vertical line (x0 = x1 = 0 in tile coordinates) is handled specially. If y0 = 0, then the crossing is counted for the entire tile. Otherwise, the line is discarded. The "entire tile" case is modeled in the code as drawing a vertical line offset slightly from the left edge of the tile, which has the same effect, and fine code never has to deal with the case where x0 = x1 = 0.

Non-vertical lines touching the left edge of the tile (x0 = 0 xor x1 = 0) are *potentially* a y-edge crossing. If the line touches the top left of the tile, it is not a crossing, otherwise it is. The former case is modeled in the code by displacing the point slightly to the right, so that x0 = 0 or x1 = 0 in fine is a reliable indicator of y-edge crossing. Currently a y-edge value is computed and stored in `Segment`, but it is redundant with zero comparisons of x values, and is intended to go away when those values move to f16.

There are two good ways to understand the numerical robustness logic. One is formally, by counting intersections between the path segments (represented as floating point) against "reference points" which are determined with tiny epsilon values. The reference point corresponding to backdrop is -ε², ε). The reference point for the tile is (ε², ε), which is why a vertical line from the tile origin counts as a crossing for the tile. The reference point for the pixel row y is (ε², ε + y), which is why lines touching the top left corner are not counted as a crossing, while otherwise lines touching the left edge are.

The other way to understand the logic is by cases, and a good way to present those are by illustrations. Unfortunately that's beyond the scope of this PR description, but we will write this up properly.

The added test is very good at triggering numerical robustness issues, and in fact surface errors in msaa that haven't been detected before. Addressing those will be a followup PR. It is believed that area aa is now completely correct.

### DIFF
--- a/crates/encoding/src/path.rs
+++ b/crates/encoding/src/path.rs
@@ -207,6 +207,7 @@ pub struct SegmentCount {
 #[derive(Clone, Copy, Debug, Zeroable, Pod, Default)]
 #[repr(C)]
 pub struct PathSegment {
+    // Points are relative to tile origin
     pub point0: [f32; 2],
     pub point1: [f32; 2],
     pub y_edge: f32,
--- a/crates/encoding/src/path.rs
+++ b/crates/encoding/src/path.rs
@@ -207,8 +207,8 @@ pub struct SegmentCount {
 #[derive(Clone, Copy, Debug, Zeroable, Pod, Default)]
 #[repr(C)]
 pub struct PathSegment {
-    pub origin: [f32; 2],
-    pub delta: [f32; 2],
+    pub point0: [f32; 2],
+    pub point1: [f32; 2],
     pub y_edge: f32,
     pub _padding: u32,
 }
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -251,7 +251,7 @@ fn fill_path_ms(fill: CmdFill, local_id: vec2<u32>, result: ptr<function, array<
             let zp = floor(a * f32(sub_ix - 1u) + b);
             if sub_ix == 0u {
                 is_delta = y0i == xy0.y && y0i != xy1.y;
-                is_bump = xy0.x == 0.0;
+                is_bump = xy0.x == 0.0 && y0i != xy0.y;
             } else {
                 is_delta = z == zp;
                 is_bump = is_positive_slope && !is_delta;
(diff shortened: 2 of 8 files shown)


crates/encoding/src/path.rs
```diff
@@ -207,8 +207,8 @@ pub struct SegmentCount {
 #[derive(Clone, Copy, Debug, Zeroable, Pod, Default)]
 #[repr(C)]
 pub struct PathSegment {
-    pub origin: [f32; 2],
-    pub delta: [f32; 2],
+    pub point0: [f32; 2],
```
armansito, line 210: I would repeat the comment here about `point0` and `point1` being relative to tile origin.

shader/fine.wgsl
```diff
@@ -465,7 +452,7 @@ fn fill_path_ms(fill: CmdFill, wg_id: vec2<u32>, local_id: vec2<u32>, result: pt
             // bits 4 * k + 2 and 4 * k + 3 contain 4-reductions
             let xored01_4 = xored01 | (xored01 * 4u);
             let xored2 = (expected_zero * 0x1010101u) ^ samples2;
-            let xored2_2 = xored0 | (xored0 * 2u);
+            let xored2_2 = xored2 | (xored2 * 2u);
```
armansito, line 455: Good catch!

shader/fine.wgsl
```diff
@@ -264,7 +251,7 @@ fn fill_path_ms(fill: CmdFill, wg_id: vec2<u32>, local_id: vec2<u32>, result: pt
             let zp = floor(a * f32(sub_ix - 1u) + b);
             if sub_ix == 0u {
                 is_delta = y0i == xy0.y && y0i != xy1.y;
-                is_bump = xy0.x == 0.0;
+                is_bump = xy0.x == 0.0 && y0i != xy0.y;
```
armansito, line 254: This looks good but it would be very nice to capture in the write up the nuances between `is_bump` vs `is_delta` vs `is_positive_slope` and what I believe are the vertical pixel-grid alignment checks (i.e. `y0i == xy0.y` and `y0i != xy0.y`). Maybe add a TODO here so we remember to reference it when the write up is done?

armansito (approved):
Thank you for working on these improvements.

shader/fine.wgsl
raphlinus: I did a bit of explanation in comments. The logic is very, very subtle, and I think needs a real writeup, but hopefully having some explanation will be helpful until then.

PR #405, opened 2023-11-04
Downloads: Handle an updated content length and add better error checking

DJMcNab:
I tried running the download script today, which failed.

In case a similar issue occurs in future, I have added some more graceful error handling.


armansito (approved):
Approved % merge conflict

PR #410, opened 2023-11-10
Refactor transform handling in flatten & support miter join

armansito:
This PR implements the miter join style and adds additional test scenes for stroking under non-uniform scale and skew transforms. This includes some changes to transform handling in the `flatten` stage to keep stroking correct under these transforms:

  * Flattening (including for fills, offset curves, caps, and joins) is now performed in a curve's local coordinate space (pre-transform) and the transform is applied at the time a line gets output (i.e. LineSoup time). This may be unnecessary for fills but the shader currently uses the same code for flattening fills and strokes.

* The line count estimate for subdivision uses local control point coordinates and makes use of the scale factor which is decomposed from the transform matrix.

* The bounding box computation is now based on the union of output line segments, which get accumulated at LineSoup time.

This PR is based on #408


shader/flatten.wgsl
```diff
@@ -407,55 +488,49 @@ fn main(
...
         if is_stroke {
             let linewidth = bitcast<f32>(scene[config.style_base + style_ix + 1u]);
+            let offset = 0.5 * linewidth;
+
             // See https://www.iquilezles.org/www/articles/ellipses/ellipses.htm
             // This is the correct bounding box, but we're not handling rendering
             // in the isotropic case, so it may mismatch.
-            stroke = 0.5 * linewidth * vec2(length(transform.mat.xz), length(transform.mat.yw));
-            bbox += vec4(-stroke, stroke);
+            stroke = offset * vec2(length(transform.mat.xz), length(transform.mat.yw));
```
armansito, line 501: `stroke` is now unused. I've been thinking about removing this field from the `cubic` struct or possibly replacing `Cubic` with `CubicPoints` + just the `path_ix`.

shader/flatten.wgsl
```diff
@@ -227,7 +225,59 @@ fn flatten_cubic(cubic: Cubic) {
...
+            output_two_lines_with_transform(path_ix,
+                                            p0 + n_prev, p0 + n_next,
+                                            p0 - n_next, p0 - n_prev,
+                                            transform);
+        }
+        case /*STYLE_FLAGS_JOIN_BITS_MITER*/0x10000000u: {
+            let c = tan_prev.x * tan_next.y - tan_prev.y * tan_next.x;
+            let d = dot(tan_prev, tan_next);
+            let hypot = length(vec2f(c, d));
+            let miter_limit = unpack2x16float(style_flags & STYLE_MITER_LIMIT_MASK).x;
```
DJMcNab, line 245: I'd *probably* choose to use the indexing form of vector access (i.e. `unpack2x16float(...)[0]`) here, as this isn't semantically an `x` point but just the first 16 bits.

But that's not based on any checking what the style is elsewhere

shader/flatten.wgsl
```diff
@@ -333,33 +383,57 @@ fn read_path_segment(tag: PathTagData, transform: Transform, is_stroke: bool) ->
...
+    lines[line_ix] = LineSoup(path_ix, tp0, tp1);
+}
+
+fn output_two_lines_with_transform(
+    path_ix: u32,
+    p00: vec2f, p01: vec2f,
+    p10: vec2f, p11: vec2f,
+    transform: Transform
+) {
+    let line_ix = atomicAdd(&bump.lines, 2u);
```
DJMcNab, line 425: Is this semantically equivalent to `output_line_with_transform` called twice, and only written this way to reduce atomic traffic?

DJMcNab (commented):
Does this also need to be reflected in the CPU shaders, or will those continue to use the kurbo based stroke expansion?

Otherwise, a couple of tiny style questions

shader/flatten.wgsl
armansito: We have some uses of the `unpack` intrinsics in the fine shader that perform a swizzle using the `xyzw` accessors, though I suppose changing those to use the color (`rgba`) accessors would be semantically more correct.

Using `[0]` here sounds good to me.

armansito:
> Does this also need to be reflected in the CPU shaders, or will those continue to use the kurbo based stroke expansion?
> 
Yes, I'm planning to make the changes to the CPU shaders in a follow-up PR very soon (I'm planning to get that done over the next couple weeks). The kurbo-based stroke expansion will only be used in the "CPU stroke mode" (for lack of a better term) and for dashing in the short term.

The CPU shaders themselves are a CPU incarnation of the GPU shaders with a better test environment, so I'm planning to mirror these changes there pretty much 1:1.

shader/flatten.wgsl
DJMcNab: Thanks, that makes sense.

I'd probably add a comment to that effect, or possible add a function (`output_line_into_allocated_space`?) which these both call

I did notice one case where four lines were handled in a row (when handling miter joins) . Handling that case all in one bump could make sense (and could additionally avoid creating some `var`iables). But that also adds would seem to add quite a lot of noise for tiny gains

shader/flatten.wgsl
```diff
@@ -139,7 +139,9 @@ fn flatten_cubic(cubic: Cubic) {
     let Q_ACCURACY = ACCURACY * 0.1;
     let REM_ACCURACY = ACCURACY - Q_ACCURACY;
     let MAX_HYPOT2 = 432.0 * Q_ACCURACY * Q_ACCURACY;
-    var n_quads = max(u32(ceil(pow(err * (1.0 / MAX_HYPOT2), 1.0 / 6.0))), 1u);
+    let scale = vec2(length(transform.mat.xz), length(transform.mat.yw));
```
raphlinus, line 142: The correct math for this is `0.5 * length(vec2(mat.x + mat.w, mat.y - mat.z)) + length(mat.x - mat.w, mat.y + mat.z))` (using a bit of shorthand). This is not substantially more complicated or inefficient than what you have, so I'd like to see the real thing.

To see the difference, consider the skew matrix [[1 0] [1 1]]. The correct answer is the golden mean ≈1.618, the formula given is sqrt(2) ≈1.414.

The formula given is accurate for uniform scaling and rotation, and also for nonuniform axis-aligned scaling, so it's not hard to see why the incorrect equation persists.

shader/flatten.wgsl
```diff
@@ -151,20 +153,16 @@ fn flatten_cubic(cubic: Cubic) {
...
 
-        // HACK: this increases subdivision count as a function of the stroke width for shitty
-        // strokes. This isn't systematic or correct and shouldn't be relied on in the long term.
-        var tol = sqrt(REM_ACCURACY);
-        if cubic.flags == CUBIC_IS_STROKE {
-            tol *= min(1000., dot(cubic.stroke, cubic.stroke));
-        }
-        let params = estimate_subdiv(qp0, qp1, qp2, tol);
+        // TODO: Estimate an accurate subdivision count for strokes, handling cusps.
+        let tol = sqrt(REM_ACCURACY);
```
raphlinus, line 157: I believe the correct way to do scaling is:

```wgsl
    let scaled_sqrt_tol = (REM_ACCURACY / scale_factor);
    ... apply fudge factor here, which will later be reverted...
    let params = estimate_subdiv(qp0, qp1, qp2, scaled_sqrt_tol);
...
    let n = max(u32(ceil(val * (0.5 / scaled_sqrt_tol))), 1u);
```

raphlinus (commented):
I'm commenting on the transforms only, as I've done a deep dive. I'll dig into the rest a bit later.

In addition to what's inline:

The handling of `scale_factor` and transform should be different in the fill and stroke case. In the fill case, the scale_factor should be 1 and the cubic should be transformed before going into flatten_cubic. That gets near-optimal flattening even in the presence of non-angle-preserving transform.

In the stroke case, scale_factor should be computed as stated, and the transform applied post-stroking. That gets correct computation of miters, elliptical pen for round, etc. It's also not optimal flattening (more so the more eccentric the transform), but we've decided the extra complexity for optimal flattening in that case is not worth it.

Let me know if any of this is not clear.

shader/flatten.wgsl
armansito: That makes sense. I had this initially written similarly to the correct version but the existing formula looked right under my test cases (as backed up by your explanation) and this slipped through the cracks. Fixed this now to do the correct thing.

shader/flatten.wgsl
```diff
@@ -52,15 +52,15 @@ fn approx_parabola_inv_integral(x: f32) -> f32 {
     return x * sqrt(1.0 - B + (B * B + 0.5 * x * x));
 }
 
-fn estimate_subdiv(p0: vec2f, p1: vec2f, p2: vec2f, sqrt_tol: f32) -> SubdivResult {
+fn estimate_subdiv(p0: vec2f, p1: vec2f, p2: vec2f, sqrt_tol: f32, transform_scale: f32) -> SubdivResult {
```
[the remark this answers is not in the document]
armansito: > I don't think this needs an additional `transform_scale` argument, and the scaling is subtly incorrect with it added.
> 
> The `val` return has this meaning: an optimal subdivision of the quadratic into `val` / 2 subdivisions (which may be fractional) will have an error sqrt_tol^2 (ie the desired tolerance). There are two cases. In the non-cusp case, the error scales as the inverse square of val (doubling val causes the error to be one fourth), so the tolerance is actually not needed for the calculation (which is why it's applied in the caller). In the cusp case, this scaling breaks down and the tolerance parameter is needed to compute the correct result.
> 
Thank you, this is very illuminating as to how `val` gets computed (as I didn't understand before why the tolerance was only considered when the signs are different).

I copied your comment and included it in the code as it's a useful read.

shader/flatten.wgsl
armansito: I think this is a good idea. I'm not sure yet how significant the impact of the additional atomic here is but I suspect it might matter. At any rate, I like the idea of adding a `output_line_into_allocated_space` helper as I need something to that effect for round caps and joins where the line count is typically a lot higher.

dfrg (approved):
Daniel and Raph already looked at this so I just did an additional sanity check and this all looks reasonable to me. Let's land it!

armansito:
> Daniel and Raph already looked at this so I just did an additional sanity check and this all looks reasonable to me. Let's land it!

Thanks!

PR #412, opened 2023-11-15
CPU-side dashing for GPU strokes & encoding-time filtering of zero-length segments

armansito:
- When GPU-side stroking is enabled, a stroke with a dash style now gets converted to a series of regular strokes which then get encoded separately and in sequence.

- Zero-length segments lead to numerical errors and lead to extra complexity and wasted work (in the form of thread 
   allocation) for GPU flattening. We now detect and drop them at encoding time.
   
   Currently, a path segment is considered zero-length if all of its control points (in local coordinates) are within EPSILON (1e-12) of each other. This may not be the desired behavior under transform.

   Also since the length here isn't in terms of the actual arc length, this won't detect all curves that are within an EPSILON sized bounding box (which may have a larger distance between their control points).

   For stroking purposes, the distance between the control points is what matters most as that's what's used to compute a curve's tangent.

This PR is based on #410 

#303

### DIFF
--- a/crates/encoding/src/path.rs
+++ b/crates/encoding/src/path.rs
@@ -523,9 +523,8 @@ impl<'a> PathEncoder<'a> {
             self.move_to(self.first_point[0], self.first_point[1]);
         }
         if self.state == PathState::MoveTo {
-            let p0 = (self.first_point[0], self.first_point[1]);
             // Ensure that we don't end up with a zero-length start tangent.
-            let Some((x, y)) = start_tangent_for_curve(p0, (x, y), None, None) else {
+            let Some((x, y)) = self.start_tangent_for_curve((x, y), None, None) else {
                 return;
             };
             self.first_start_tangent_end = [x, y];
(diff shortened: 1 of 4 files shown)


crates/encoding/src/path.rs
```diff
@@ -723,31 +763,25 @@ impl fello::scale::Pen for PathEncoder<'_> {
...
 // represent a line, a quadratic Bezier, or a cubic Bezier. Lines and quadratic Beziers can be
 // passed to this function by simply setting the invalid control point degrees equal to `(x0, y0)`.
 fn start_tangent_for_curve(
     p0: (f32, f32),
     p1: (f32, f32),
-    p2: (f32, f32),
-    p3: (f32, f32),
+    p2: Option<(f32, f32)>,
+    p3: Option<(f32, f32)>,
 ) -> Option<(f32, f32)> {
```
dfrg, line 777: Small style suggestion: making this a method of `PathEncoder` and dropping the `p0` parameter (since it always appears to be `PathEncoder::first_point`?) should simplify the call sites a bit.

examples/scenes/src/test_scenes.rs
```diff
@@ -70,6 +70,7 @@ pub fn test_scenes() -> SceneSet {
 fn funky_paths(sb: &mut SceneBuilder, _: &mut SceneParams) {
     use PathEl::*;
     let missing_movetos = [
+        MoveTo((0., 0.).into()),
```
dfrg, line 73: I find this more amusing than is reasonable :)

dfrg (approved):
LGTM modulo small style suggestion. Thanks!

PR #414, opened 2023-11-16
Round join and all cap styles

armansito:
Implemented the logic for round join and cap styles. The logic generates lines directly from a circular arc. The arc gets flattened in the curve's local coordinate space and gets transformed to device-space coordinates post-flattening.

This completes the initial implementation of all stroke styles. This PR is based on #412.

#303

### DIFF
--- a/shader/flatten.wgsl
+++ b/shader/flatten.wgsl
@@ -294,7 +294,8 @@ fn flatten_arc(
     let radius = max(tol, length(p0 - transform_apply(transform, center)));
     let x = 1. - tol / radius;
     let theta = acos(clamp(2. * x * x - 1., -1., 1.));
-    let n_lines = select(u32(ceil(6.2831853 / theta)), 1u, theta <= EPS);
+    let MAX_LINES = 1000u;
+    let n_lines = select(min(MAX_LINES, u32(ceil(6.2831853 / theta))), MAX_LINES, theta <= EPS);
 
     let th = angle / f32(n_lines);
     let c = cos(th);
--- a/shader/flatten.wgsl
+++ b/shader/flatten.wgsl
@@ -290,11 +290,11 @@ fn flatten_arc(
     var r = begin - center;
 
     let EPS = 1e-9;
-    let tol = 0.1;
+    let tol = 0.5;
     let radius = max(tol, length(p0 - transform_apply(transform, center)));
     let x = 1. - tol / radius;
     let theta = acos(clamp(2. * x * x - 1., -1., 1.));
-    let n_lines = select(u32(ceil(6.2831854 / theta)), 1u, theta <= EPS);
+    let n_lines = select(u32(ceil(6.2831853 / theta)), 1u, theta <= EPS);
 
     let th = angle / f32(n_lines);
     let c = cos(th);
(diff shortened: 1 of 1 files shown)


shader/flatten.wgsl
```diff
@@ -280,39 +280,95 @@ fn flatten_cubic(cubic: Cubic, local_to_device: Transform, offset: f32) {
     }
 }
 
+// Flattens the circular arc that subtends the angle begin-center-end. It is assumed that
```
DJMcNab, line 283: It's not obvious from this comment how the direction is chosen. It must be based on the value of `angle`, but it feels like 'pacman' style arcs will be difficult.

I worry that if n_lines is big, the arc would draw the full way around, but if n_lines is small, it would end up drawing the closing part, rather than the full way around

I'm not sure that this is coherent - my last maths education was many years ago.

Maybe this would only be an issue if n_lines is 1, which is impossible?

armansito: The direction of the arc is always a counter-clockwise rotation starting from `begin`, towards `end`, centered at `center`, and will be subtended by `angle` (which is assumed to be positive). If `angle` terminates the arc before it smoothly reaches `end`, a line will be drawn from the end of the arc towards `end`. I will document this.

`n_lines` only depends on the radius of the arc and the chosen tolerance.  For `n_lines` to be $1$, `theta` must be close to $6.2831853$ ($2\pi$). `acos` can't return that [by definition](https://www.w3.org/TR/WGSL/#acos-builtin). _If_ you somehow got `n_lines == 1`, basically no circle would be drawn and you'd get a line connecting `begin` and `end`.

Realistically, `n_lines` could be as low as $2$ if the absolute value of `x` is very small, which can happen if `tol` and `radius` are close to each other (the code prevents `radius` from being smaller than `tol`, so that case isn't possible). You'd only get into that situation with a sub-pixel radius. At that scale, the coarseness of the approximation won't be easily noticeable. If you increased the tolerance to make this possible at larger radii, with `n_lines == 2` you'd get a shape resembling a wedge/triangle (I tested this locally, which works as you'd expect).

That said, this function is meant for drawing caps and joins where `angle`$\le\pi$ and the `begin`, `center`, and `end` points are chosen carefully (so that the angle and winding make sense). 

Though there is something that should get fixed: I think selecting `1u` when `theta <= EPS` is the wrong thing to do and this is a bug. That happens when the subdivision count tends to infinity. It's probably better to define a `MAX_LINES` and return that instead.

armansito: To illustrate what I mean, here is what the code draws with a center directly below the join and an angle of $343.77\degree$:

- `n_lines == 1`:
Screenshot 2023-11-17 at 12 10 01 PM

- `n_lines == 2`:
Screenshot 2023-11-17 at 12 10 44 PM

- `n_lines == 3`:
Screenshot 2023-11-17 at 12 11 29 PM

- `n_lines == 6`:
Screenshot 2023-11-17 at 12 12 18 PM

- `n_lines == 100`:
Screenshot 2023-11-17 at 12 12 58 PM

DJMcNab: Thanks for the very detailed reply!

I hadn't clocked the additional invariants being maintained by the callers

Those pictures are very cool as well!

dfrg (approved):
LGTM. Exciting to see this all coming together!

shader/flatten.wgsl
armansito: I'm glad they were helpful 🙂. I updated the comment above `flatten_arc` in [latest commit](https://github.com/linebender/vello/pull/414/commits/6e6d02a69e4df6b297bedde16370053e05d6e207).

Issue #366, opened 2023-09-28
Strategy for robust dynamic memory, readback, and async


raphlinus:
One of the stickier points is how to handle robust dynamic memory. The fundamental problem is that the pipeline creates intermediate data structures (grids of tiles containing coarse winding numbers and path segments, per-tile command lists) whose size depends dynamically on the scene being rendered. For example, just changing a transform can significantly affect the number of tiles covered by a path, and thus the size of these data structures.

The standard GPU compute shader execution model cannot express what we need. At the time a command buffer is submitted, all buffers have pre-determined size. There is no way to dynamically allocate memory based on computation done inside a compute shader (note: CUDA doesn't have this limitation, shaders can simply call malloc or invoke C++ `new`).

Another potential way to address the fundamental problem is to divide work into chunks so that intermediate results fit in fixed size buffers. This would be especially appealing in resource constrained environments where calling malloc may not be guaranteed to succeed, or may cause undesirable resource contention. However, that requires the ability to launch work dependent on computation done in a shader. Again, CUDA can do this (for example, with [device graph launch](https://developer.nvidia.com/blog/enabling-dynamic-control-flow-in-cuda-graphs-with-device-graph-launch/)) but it is not a common capability of compute shaders, much less WebGPU.

The previous incarnation, piet-gpu, had *a* solution (see #175), but with drawbacks. Basically, rendering a scene required a fence back to the GPU to read a buffer with a success/failure indication, with reallocation and retry on failure. However, this requires the ability to do blocking readback (which is missing in WebGPU), and also blocks the calling thread (usually the UI runloop) until the result of the computation is available (which is the reason why it's missing in WebGPU).

There's no good solution to this problem, only a set of tradeoffs. We need to decide what to implement. The best choice will depend on the details of what's being rendered and which application it's integrated with. In no particular order, here are some of the choices:

1. When the maximum complexity of the scenes being rendered is known in advance, then the buffer sizes can simply be determined in advance. On failure, the scene would fail to render. This may well be the best choice for games and applications in which the UI is not rendering user-provided content. It allows the entire rendering pipeline to be launched as "fire and forget" with no blocking.

2. We could do analysis CPU-side to determine the memory usage, before launching the render. This is simple and poses no integration challenges, but such analysis is slow. In fact, it's probably comparable to running the compute pipeline on CPU and just using the GPU for fine rasterization, a modality we're considering as a compatibility fallback. It may be a viable choice when the scene complexity is low.

[...]

PR #415, opened 2023-11-22
Port recent GPU stroke expansion changes in `flatten` to CPU version

armansito:
This replicates the GPU-side stroke expansion changes that were recently made to the `flatten` stage to its CPU version.

Note on buffer sizes:
---------------------
The `longpathdash (round caps)` and `mmark` scenes assert (due to out-of-bounds buffer access) if zoomed in (the former crashes immediately). The asserts can occur in the `path_count` and `flatten` stages. The root cause is that the `segments`, `seg_counts`, and `lines` buffers are not large enough to accommodate the required number of line segments. When run on the GPU, this manifests itself as artifacts rather than a crash.

This PR does not adjust the bump buffer sizes in order to avoid hiding the problem. I would like to fix this as part of #366. Unfortunately, even an accurate buffer size estimate will require more memory than what's permitted by the WebGPU standard (the default limit for [maxStorageBufferBindingSize](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxstoragebufferbindingsize) is 128 MiB and implementations aren't guaranteed to support a higher value). We have been considering culling lines that fall outside the viewport which can alleviate the memory usage in zoomed-in scenarios.


crates/encoding/src/math.rs
```diff
@@ -172,4 +196,11 @@ mod tests {
         let output: u16 = f32_to_f16(input);
         assert_eq!(0xFC00, output); // -inf
     }
+
+    #[test]
+    fn test_f16_to_f32() {
+        const EPS: f32 = 0.001;
+        let input: f32 = std::f32::consts::PI;
+        assert!((input - f16_to_f32(f32_to_f16(input))).abs() < EPS);
```
DJMcNab, line 204: It would be nice to have:
1) an exported bit pattern of a 16 bit float, which we check gets the right number (i.e. test that both directions aren't wrong in the same way)
2) test that values in the extreme ranges (i.e. values of the order `60_000` and `1/60_000` ± a factor of 2, if required - and 0) are appropriately roundtripped
3) special values (i.e. NaN, infinity) roundtrip as expected. This is less important, as we don't really want to see them - maybe we should `warn!` in those cases?

armansito: Done. I added tests for max and min value roundtrip, nan/inf roundtrip, and an explicit conversion. Perhaps `test_f16_to_f32_roundtrip` should become a proptest t but I think this is good enough for now.

src/cpu_shader/flatten.rs
```diff
@@ -151,12 +311,176 @@ fn flatten_cubic(cubic: Cubic, line_ix: &mut usize, lines: &mut [LineSoup]) {
...
+    let x = 1. - tol / radius;
+    let theta = (2. * x * x - 1.).clamp(-1., 1.).acos();
+    const MAX_LINES: u32 = 1000;
+    let n_lines = if theta <= ROBUST_EPSILON {
+        MAX_LINES
+    } else {
+        MAX_LINES.min((std::f32::consts::TAU / theta).ceil() as u32)
+    };
+
+    let th = angle / (n_lines as f32);
```
DJMcNab, line 338: Why does the number of lines depend only on the radius, rather than including the angle being swept over?

I would naïvely expect `th` to be the same for all circles with the same radius, as that should make equal length lines.
It feels quite likely that I'm missing something obvious here

src/cpu_shader/flatten.rs
```diff
@@ -151,12 +311,176 @@ fn flatten_cubic(cubic: Cubic, line_ix: &mut usize, lines: &mut [LineSoup]) {
...
+    line_ix: &mut usize,
+    lines: &mut [LineSoup],
+    bbox: &mut IntBbox,
+) {
+    let mut p0 = transform.apply(begin);
+    let mut r = begin - center;
+    let tol: f32 = 0.5;
+    let radius = tol.max((p0 - transform.apply(center)).length());
+    let x = 1. - tol / radius;
+    let theta = (2. * x * x - 1.).clamp(-1., 1.).acos();
```
raphlinus, line 330: This is the same as sqrt(8 tol / radius) for small tol. I wonder if the extra math is worth it. I haven't investigated in detail.

src/cpu_shader/flatten.rs
raphlinus: I also have this concern, and wonder if it might have something to do with the high observed memory usage.

src/cpu_shader/flatten.rs
```diff
@@ -151,12 +311,176 @@ fn flatten_cubic(cubic: Cubic, line_ix: &mut usize, lines: &mut [LineSoup]) {
...
+    center: Vec2,
+    angle: f32,
+    transform: &Transform,
+    line_ix: &mut usize,
+    lines: &mut [LineSoup],
+    bbox: &mut IntBbox,
+) {
+    let mut p0 = transform.apply(begin);
+    let mut r = begin - center;
+    let tol: f32 = 0.5;
```
raphlinus, line 327: This seems a little loose to me, I've generally been using 0.25. I'm also concerned that it doesn't match the flattening tolerance for curves in general.

This might be a good time for a heads up: I am not confident that the tolerance value in the existing code is accurately calibrated against the actual Fréchet distance between flattened and source curve; I may have let a constant factor of error enter. But the current work I'm doing *does* calibrate that, so when I land that I will be systematic.

raphlinus (commented):
I have concerns about the number of subdivisions in the arc flattening logic. I didn't wrap my head around it fully - perhaps it's right and there's something I'm missing, in which case I'd happily accept an argument.

Other than that, this looks like a good foundation for the Euler spiral flattening, and I'm looking forward to getting that fully wired.

src/cpu_shader/flatten.rs
armansito: Ah yes, there is a bug here and thank you for taking a close look and catching that. `n_lines` is the number for a full circle. This is directly transcribed from the GPU flatten code where I was getting the count for a full circle and fitting that into any sweep angle. I also noticed that the derivation of `theta` is also not quite right. Given a radius $r$ and the length $l$ of the line segment from the center of the circle to the middle of the chord (i.e. a flattened line segment) given by $l = r - tol$, the derivation should look like:

$$ tol = r - r \times cos(\frac{\theta}{2}) $$

$$ \theta = 2 \times arccos(1 - \frac{tol}{r}) $$

And `n_lines = (angle / theta).ceil()`. We still need to guard against a division by zero / over-subdividing. I went ahead and picked an arbitrary minimum angle limit of `0.0001` radians but this could probably get fine tuned.

src/cpu_shader/flatten.rs
armansito: The new expression should be correct now and is simpler.

src/cpu_shader/flatten.rs
armansito: With the new derivation, 0.25 leads to visible line segments. 0.1 looks decent enough to me. I'm worried that anything lower will blow up the line count but you should probably take a look and make a judgement as to whether this tolerance is visually satisfactory.

src/cpu_shader/flatten.rs
```diff
@@ -151,12 +311,176 @@ fn flatten_cubic(cubic: Cubic, line_ix: &mut usize, lines: &mut [LineSoup]) {
...
+    let radius = tol.max((p0 - transform.apply(center)).length());
+    let x = 1. - tol / radius;
+    let theta = (2. * x * x - 1.).clamp(-1., 1.).acos();
+    const MAX_LINES: u32 = 1000;
+    let n_lines = if theta <= ROBUST_EPSILON {
+        MAX_LINES
+    } else {
+        MAX_LINES.min((std::f32::consts::TAU / theta).ceil() as u32)
```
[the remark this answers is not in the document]
armansito: > Is it likely that most of this maximum will be used on paths which are off-screen?

Yes, the really high line segment counts occur at really high zoom levels (for example if you zoom in directly to a stroke's cap). There is actually currently no culling and it's highly likely to wastefully generate off-screen line segments at high zoom levels. longpathdash with round caps is a good example of this, where zooming in actually exceeds the memory budget for LineSoup.

We've been talking about implementing at the very least viewport culling next year to handle the zoomed-in cases.

crates/encoding/src/math.rs
```diff
@@ -123,9 +122,34 @@ pub(crate) fn f32_to_f16(val: f32) -> u16 {
     output | (sign >> 16) as u16
 }
 
+/// Convertes a 16-bit precision IEEE-754 binary16 float to a f32.
+/// This implementation was adapted from Fabian Giesen's `half_to_float`()
```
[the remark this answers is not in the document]
armansito: Done.

armansito:
PTAL. I think the arc flattening logic should be correct now barring any adjustments to `MIN_THETA` and the error tolerance. The current values still exceed the LineSoup and SegmentCount buffer sizes with the mmark and longpathdash (round) scenes.

shader/flatten.wgsl
```diff
@@ -580,7 +580,6 @@ fn output_two_lines_with_transform(
 
 struct NeighboringSegment {
     do_join: bool,
-    p0: vec2f,
```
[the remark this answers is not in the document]
armansito: \*\*suspense\*\*

raphlinus (approved):
Right, this makes more sense. I expect to take a closer look (hopefully visualizing the flattening output) as I adapt the Euler stuff, so I didn't go through with a fine-tooth comb, but this definitely gets us to a better place. Thanks!

PR #416, opened 2023-12-13
Debug and Validation Layers

armansito:
We've been talking about various ways to perform CPU-side validation/testing over the outputs of Vello pipeline stages. It's also generally useful to be able to visualize the contents of some of these intermediate data structures (such as bounding boxes, the line soup, etc) and to be able to visually interpret any errors that are surfaced from the validation tests.

I implemented the beginnings of this in a new `debug_layers` feature. I'm putting this up as a Draft PR as there are a few unresolved aspects that I'd like to get feedback on first (more on this below).

## Running Instructions

To try this out run the `with_winit` example with `--features debug_layers` and use the number keys (1-4) to toggle the individual layers.

## Summary
This PR introduces the concept of "debug layers" that are rendered directly to the surface texture over the fine rasterizer output. There are currently 4 different layers:

- `BOUNDING_BOXES`: Visualizes the edges of path bounding boxes as green lines
- `LINESOUP_SEGMENTS`: Visualizes LineSoup segments as orange lines
- `LINESOUP_POINTS`: Visualizes the LineSoup endpoints as cyan circles.
- `VALIDATION`: Runs a collection of validation tests on intermediate Vello GPU data structures. This is currently limited to a watertightness test on line segments. Following the test run, the layer visualizes the positions of detected errors as red circles.

These layers can be individually toggled using a new `DebugLayers` field in `vello::RenderParams`. The following is an example output with all 4 layers enabled:

Screenshot 2023-12-12 at 3 13 51 PM

Each layer is implemented as an individual render pass. The first 3 layers are simple visualizations of intermediate GPU buffers. The `VALIDATION` layer is special since it runs CPU-side validation steps (currently a watertightness test over the LineSoup buffer), which requires read-back. The general idea is that `VALIDATION` can grow to encompass additional sanity checks.

### Overview of Changes
- Engine support for render pipeline creation and draw commands. In particular, the existing `blit` pipeline can now be expressed as a `Recording`. The debug layer render passes get recorded to this `Recording`. All render passes share the same render encoder and target the same surface texture.
- A simple mechanism to extend the lifetime of GPU buffers beyond their original `Recording` to allow them to be used in a subsequent render pass. Currently this separation into multiple recordings is necessary since the visualizations require GPU->CPU read-back. This is partially encapsulated by the new `CapturedBuffers` data structure.
- The `debug` module and the `DebugLayers` and `DebugRenderer` data structures. `DebugRenderer` is an encapsulation of the various render pipelines used to visualize the layers that are requested via `DebugLayers`. `DebugRenderer` is currently also responsible for execution the validation tests when the `VALIDATION` layer is enabled.
- The `with_winit` example now has key bindings (the number keys 1-4) to toggle the individual layers.

## Open Questions
1. It probably makes sense to have a better separation between running validation tests and visualizing their output. Currently both are performed by `DebugRenderer::render`.

2. `CapturedBuffers` doesn't handle buffer clean up well. The current `engine` abstractions require that a buffer be returned to the underlying engine's pool via a `Recording` command. This creates and submits a command buffer to simply free a buffer, which is a bit too heavy-weight. This whole mechanism could use some rethinking.

    Currently, these buffers get conditionally freed in various places in the code and it would be nice to consolidate that logic. 

3. The `VALIDATION` layer currently doesn't work with the `--use-cpu` flag since the buffer download command isn't supported for CPU-only buffers. Currently, it's the job of `src/render.rs` to know which buffers need to get downloaded for validation purposes. It currently simply records a download command. It would be nice for the engine to make the download command seamless across both CPU and GPU buffers rather than having the `src/render.rs` code do something different across the CPU vs GPU modalities.

4. Currently all layers require read-back. The debug layers (`BOUNDING_BOXES`, `LINESOUP_SEGMENTS`, `LINESOUP_POINTS`) read back the `BumpAllocators` buffers to obtain instance counts used in their draw commands. This read-back could be avoided by instead issuing indirect draws for the debug layers. I think this could be implemented with a relatively simple scheme: a new compute pipeline stage is introduced (gated by `#[cfg(feature = "debug_layers")]`, which can inspect any number of the vello intermediate buffers (such as the `bump` buffer) and populate an indirect draw buffer. The indirect draw buffer would be laid out with multiple [`DrawIndirect`](https://docs.rs/wgpu/latest/wgpu/util/struct.DrawIndirect.html) entries, each assigned to a different pre-defined instance type (the `DebugRenderer` only issues instanced draws). `DebugRenderer` would then issue an indirect draw with the appropriate indirect buffer offset for each render pipeline.

   The read-back would still be necessary for CPU-side validation stages and their visualization can't really take advantage of the indirect draw. Then again, the exact ordering of the draw submission and the read-backs implemented in this PR is likely to change following the proposal in #366.


waywardmonkeys (commented):
A nit, but could this and the `sin_cos` change be done separately from this?

armansito:
> A nit, but could this and the `sin_cos` change be done separately from this?

Thanks for taking a look. I don't mind sending those as a a separate PR.

armansito:
I decided to move this out of Draft status to kick off the discussion. I think some of the questions I raised in the original comment above are still valid and I would love to hear people's thoughts on them.

PoignardAzur:
To be clear, this would be *visual* validation for debugging, right? This doesn't include a way to say eg "assert that this list of command doesn't trigger a validation error"?

DJMcNab (commented):
Overall, having some more debug code is extremely useful, thanks.

I've not done a very thoughtful/thorough review yet, but think getting initial comments in is useful.

I was unable to see any watertightness failures in tiger as in your screenshot. Were those intentionally added for an example?

armansito:
> To be clear, this would be _visual_ validation for debugging, right? This doesn't include a way to say eg "assert that this list of command doesn't trigger a validation error"?

This PR is mostly about visual validation, yes. That said the watertightness test could be run as part of a test runner that actually asserts or reports the failures in a textual form (e.g. by incorporating into what's proposed in #439)

armansito:
> Overall, having some more debug code is extremely useful, thanks.
> 
> I've not done a very thoughtful/thorough review yet, but think getting initial comments in is useful.
> 
Thanks, I'll address them as soon as I get time.

> I was unable to see any watertightness failures in tiger as in your screenshot. Were those intentionally added for an example?

No, these are real errors that are present in the flattening logic (in both the CPU and GPU versions of `flatten`). I'm surprised you're not seeing them. Did you try enabling that layer by pressing "4"?

I'm also intrigued by #439 and I'm wondering if the watertightness test can be run as part of that as well.

DJMcNab:
> No, these are real errors that are present in the flattening logic (in both the CPU and GPU versions of flatten). I'm surprised you're not seeing them. Did you try enabling that layer by pressing "4"?

Hmm, I still don't see them. The layer appears to be enabled, as the screen goes dark, and the frame time increases massively

I can't test the CPU version, as that instantly panics (as expected/as you documented) due to the download issue

armansito:
> > No, these are real errors that are present in the flattening logic (in both the CPU and GPU versions of flatten). I'm surprised you're not seeing them. Did you try enabling that layer by pressing "4"?
> 
> Hmm, I still don't see them. The layer appears to be enabled, as the screen goes dark, and the frame time increases massively
> 
Ah, you're likely not seeing the errors because the GPU-side stroke expansion is disabled by default (look for `GPU_STROKES` in src/scene.rs) and the stroke encoding converts strokes to fills using kurbo. These errors are only present due to a bug in the new GPU stroke expansion code, so try setting `GPU_STROKES` to `true` before trying again.

src/debug.rs
```diff
@@ -0,0 +1,542 @@
+// Copyright 2023 The Vello authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
xStrom, line 2:
```suggestion
// Copyright 2023 the Vello Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT
```

src/debug/validate.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2023 The Vello authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
xStrom, line 2:
```suggestion
// Copyright 2023 the Vello Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT
```

xStrom (commented):
Since #456 we have standard copyright headers. Please update the two files to have consistent capitalization with the standard.

DJMcNab:
I'm sorry this hasn't been reviewed for so long - it keeps on being pushed down my todo list.

In my mind, the major concern is that this doesn't compose with the CPU shaders. It feels like it should be trivial to do readback on a CPU buffer, given that we already have it in memory; is it only an API issue?

vello/src/wgpu_engine.rs
```diff
@@ -360,8 +386,11 @@ impl WgpuEngine {
                     transient_map
                         .bufs
                         .insert(buf_proxy.id, TransientBuf::Cpu(bytes));
-                    let usage =
-                        BufferUsages::COPY_SRC | BufferUsages::COPY_DST | BufferUsages::STORAGE;
+                    // TODO: restrict VERTEX usage to "debug_layers" feature?
```
DJMcNab, line 389: Yeah, we have lots of issues with implicit usages; I think we might need to either make this explicit at the recording level, or else loop through the recording and determine what capabilities we need. I think I'd prefer to just make it explicit.

But this isn't a task for this PR.

vello/src/lib.rs
```diff
@@ -211,6 +216,11 @@ pub enum Error {
     #[cfg(feature = "wgpu")]
     #[error("Failed to async map a buffer")]
     BufferAsyncError(#[from] wgpu::BufferAsyncError),
+    /// Failed to download an internal buffer for debug visualization.
+    #[cfg(feature = "wgpu")]
+    #[cfg(feature = "debug_layers")]
+    #[error("Failed to download internal buffer for visualization")]
+    DownloadError(&'static str),
```
DJMcNab, line 223: I think this should use `{0}` somewhere in the message?

vello/src/debug/validate.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Point")
+            .field("path_ix", &self.path_ix)
+            .field("x", &f32::from_bits(self.x))
+            .field("y", &f32::from_bits(self.y))
+            .finish()
+    }
+}
+
+pub(crate) fn validate_line_soup(lines: &[LineSoup]) -> Vec<LineEndpoint> {
```
DJMcNab, line 45: If we did want to do this on the GPU, I suppose we'd want to do a 2d sort, so that all points at the same position are next to each other, then check that the points all pair up (maybe in a cleverer way than a single-threaded scan)

examples/with_winit/src/lib.rs
```diff
@@ -444,6 +464,8 @@ fn run(
                             width,
                             height,
                             antialiasing_method,
+                            #[cfg(feature = "debug_layers")]
```
DJMcNab, line 467: We should probably warn if the debug_layers feature is enabled, but `args.async_pipeline` isn't set.

vello/src/lib.rs
```diff
@@ -551,53 +629,73 @@ impl Renderer {
...
+                );
+
+                // TODO: this sucks. better to release everything in a helper
+                // TODO: it would be much better to have a way to safely destroy a buffer.
+                self.engine.free_download(captured.lines);
+                captured.release_buffers(&mut recording);
+            }
         }
-        #[cfg(feature = "wgpu-profiler")]
-        self.profiler.resolve_queries(&mut encoder);
```
DJMcNab, line 598: These lines are important

vello/src/lib.rs
```diff
@@ -392,45 +420,39 @@ impl Renderer {
...
+        let external_resources = [
+            ExternalResource::Image(target_proxy, &target.view),
+            ExternalResource::Image(surface_proxy, &surface_view),
+        ];
+        self.engine.run_recording(
+            device,
+            queue,
+            &recording,
+            &external_resources,
+            "blit (render_to_surface_async)",
```
DJMcNab, line 452:
```suggestion
            "blit (render_to_surface)",
```

vello/src/lib.rs
```diff
@@ -392,45 +420,39 @@ impl Renderer {
...
             #[cfg(feature = "wgpu-profiler")]
-            let mut render_pass = self
-                .profiler
-                .scope("blit to surface", &mut render_pass, device);
-            render_pass.set_pipeline(&blit.pipeline);
-            render_pass.set_bind_group(0, &bind_group, &[]);
-            render_pass.draw(0..6, 0..1);
-        }
-        #[cfg(feature = "wgpu-profiler")]
-        self.profiler.resolve_queries(&mut encoder);
```
DJMcNab, line 432: These lines are important

DJMcNab (approved):
I'm slightly conflicted here; the actual debug layer code looks good and works fine, but it's very clear that the code here has outgrown our abstractions.

Given that:
1) We recommend that users use the `render_to_surface` method (rather than `render_to_surface_async`)
2) All the new complexity is inside `render_to_surface_async`

I think it's *fine* to ship this. It's no secret that this code is going to need a really careful rethink at some point, but that isn't the job of this PR, and doing that rethink will be easier if we have this use case in mind.
That's also why I don't mind the CPU shaders not being supported here; clearly it's pretty far from ideal, but it doesn't break anything too badly there

Sorry again for not reviewing this earlier; my personal philosophy on code review has shifted in the intermediate time, to more favour landing things and improving them in-situ.

This PR is however definitely blocked on restoring the wgpu-profiler functionality, i.e. the `resolve_queries` calls.

vello/src/debug/renderer.rs
```diff
@@ -0,0 +1,497 @@
...
+    height: u32,
+}
+
+#[derive(Copy, Clone, Zeroable, Pod)]
+#[repr(C)]
+struct LinepointsUniforms {
+    point_color: [f32; 3],
+    point_size: f32,
+    // Uniform parameters for individual SDF point draws are stored in a single buffer.
+    // This 240 byte padding is here to bring the element ffset alignment of 256 bytes.
```
waywardmonkeys, line 347: `ffset` -> `offset`

vello/src/debug/validate.rs
```diff
@@ -0,0 +1,62 @@
...
+    std::{collections::BTreeSet, fmt},
+    vello_encoding::LineSoup,
+};
+
+#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Pod, Zeroable)]
+#[repr(C)]
+pub struct LineEndpoint {
+    pub path_ix: u32,
+
+    // Coordinates in IEEE-754 32-bit float representation
```
waywardmonkeys, line 15: why not just store them as `f32` ? (This should at least be explained here if it stays this way.)

vello/src/lib.rs
```diff
@@ -268,6 +277,13 @@ pub struct RenderParams {
     /// The anti-aliasing algorithm. The selected algorithm must have been initialized while
     /// constructing the `Renderer`.
     pub antialiasing_method: AaConfig,
+
+    /// Options for debug layer rendering.
+    ///
+    /// This only has an effect when the `debug_layers` feature is enabled.
+    // This is exposed publicly as a least-effort to avoid changing the API when features change.
+    // We expect the API to change here in the near future.
+    pub debug: DebugLayers,
```
waywardmonkeys, line 286: Since this is here, if this is set to anything but the feature is not enabled, then a warning should be printed.

waywardmonkeys (commented):
A first pass....

vello/src/lib.rs
DJMcNab: This isn't really intended to be used by end-user consumers, but I've added this message anyway

vello/src/debug/validate.rs
DJMcNab: Added my best-guess explanation

waywardmonkeys (approved):
I don't know about the actual visualization parts, but I'm okay with the integration parts!

PR #418, opened 2023-12-20
Make fine.wgsl integer literals explicitly signed

Pengiie:
The previous literals were being interpreted as an [AbstractInt](https://www.w3.org/TR/WGSL/#abstractint), which according to the [u32 constructor](https://www.w3.org/TR/WGSL/#u32-builtin):
> If T is [i32](https://www.w3.org/TR/WGSL/#i32), this is a reinterpretation of bits (i.e. the result is the unique value in [u32](https://www.w3.org/TR/WGSL/#u32) that has the same bit pattern as e).

> If T is [AbstractInt](https://www.w3.org/TR/WGSL/#abstractint), this is an identity operation if the e can be represented in [u32](https://www.w3.org/TR/WGSL/#u32), otherwise it produces a shader-creation error.

I made the literals explicit signed integers so the correct constructor overload is used.

### DIFF
--- a/shader/fine.wgsl
+++ b/shader/fine.wgsl
@@ -276,7 +276,7 @@ fn fill_path_ms(fill: CmdFill, local_id: vec2<u32>, result: ptr<function, array<
             if u32(x) < TILE_WIDTH - 1u && u32(y) < TILE_HEIGHT {
                 let delta_pix = pix_ix + 1u;
                 if is_delta {
-                    let delta = select(u32(-1), 1u, is_down) << ((delta_pix & 3u) << 3u);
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
I'm not going to merge, as I haven't run this change.

But if someone else can confirm this still executes as expected, that would be helpful.

Is there a corresponding Naga issue? i.e. tracking that the abstractint doesn't get maintained through unary minus (?)

simbleau (approved):

simbleau:
Related: https://github.com/vectorgameexperts/bevy-vello/pull/17

DJMcNab:
Thanks for the PR!

PR #421, opened 2024-01-16
Incorporate remaining stroke rework and MSAA changes into the encoding and shaders crates

armansito:
This PR makes a number of changes to make the stroke rework and MSAA pipeline features available via the `vello_encoding` and `vello_shaders` crates:

- Fixed all new and removed resource bindings and update the `WorkgroupSizes` and `BufferSizes` structures.
- Moved the sample mask LUT generation functions to `vello_encoding`.
- The exported shader data structures now include separate `fine_area`, `fine_msaa8`, and `fine_msaa16` instances.
- A couple other fixes to make the new pipelines compile successfully in both the generated WGSL and MSL.


raphlinus (approved):
Looks good, thanks for cleaning this up. I'm looking forward to when we have one source of truth for the shaders so we won't see divergence like this.

PR #422, opened 2024-01-19
feat: update vello_svg to usvg 0.37, fix some artifacts.

simbleau:
- Ignores DS_Store
- Updates to usvg 0.37
- Update the docs on integrations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target
 Cargo.lock
 /trace.json
+.DS_Store
 
 examples/assets/downloads/*
 !examples/assets/downloads/.tracked
--- a/README.md
+++ b/README.md
@@ -30,8 +30,7 @@ cargo run -p with_winit
 
 ### SVG
 
-This repository also includes [`vello_svg`](./integrations/vello_svg/), which supports converting 
-a [`usvg`](https://crates.io/crates/usvg) `Tree` into a Vello scene.
+This repository also includes [`vello_svg`](./integrations/vello_svg/), which supports converting a [`usvg`](https://crates.io/crates/usvg) `Tree` into a Vello scene.
 
 This is currently incomplete; see its crate level documentation for more information.
 
@@ -40,6 +39,11 @@ This is used in the [winit](#winit) example for the SVG rendering.
 ### Lottie
 
 A separate integration for playing Lottie animations is available through the [`velato`](https://github.com/linebender/velato) crate.
+
+### Community
+
+- [bevy-vello](https://github.com/vectorgameexperts/bevy-vello): A bevy integration to render lottie or svg files
+- [vellottie](https://github.com/vectorgameexperts/vellottie): A non-serde lottie stepthrough parser and runtime for lottie files, with a mirror'ed runtime of velato
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -40,11 +40,6 @@ This is used in the [winit](#winit) example for the SVG rendering.
 
 A separate integration for playing Lottie animations is available through the [`velato`](https://github.com/linebender/velato) crate.
 
-### Community
-
-- [bevy-vello](https://github.com/vectorgameexperts/bevy-vello): A bevy integration to render lottie or svg files
-- [vellottie](https://github.com/vectorgameexperts/vellottie): A non-serde lottie stepthrough parser and runtime for lottie files, with a mirror'ed runtime of velato
-
 ## Examples
 
 Our examples are provided in separate packages in the [`examples`](examples) folder.
--- a/integrations/vello_svg/src/lib.rs
+++ b/integrations/vello_svg/src/lib.rs
@@ -53,8 +53,8 @@ pub use usvg;
 ///
 /// See the [module level documentation](crate#unsupported-features) for a list
 /// of some unsupported svg features
-pub fn render_tree(sb: &mut SceneBuilder, svg: &usvg::Tree, transform: Option<Affine>) {
-    render_tree_with(sb, svg, transform, default_error_handler).unwrap_or_else(|e| match e {})
+pub fn render_tree(sb: &mut SceneBuilder, svg: &usvg::Tree) {
+    render_tree_with(sb, svg, default_error_handler).unwrap_or_else(|e| match e {})
 }
 
 /// Append a [`usvg::Tree`] into a Vello [`SceneBuilder`].
@@ -67,7 +67,6 @@ pub fn render_tree(sb: &mut SceneBuilder, svg: &usvg::Tree, transform: Option<Af
 pub fn render_tree_with<F: FnMut(&mut SceneBuilder, &usvg::Node) -> Result<(), E>, E>(
     sb: &mut SceneBuilder,
     svg: &usvg::Tree,
-    transform2: Option<Affine>,
     mut on_err: F,
 ) -> Result<(), E> {
     for elt in svg.root.descendants() {
@@ -80,14 +79,7 @@ pub fn render_tree_with<F: FnMut(&mut SceneBuilder, &usvg::Node) -> Result<(), E
                 tx,
                 ty,
             } = elt.abs_transform();
-            let (a, b, c, d, e, f) = (
-                sx as f64, ky as f64, kx as f64, sy as f64, tx as f64, ty as f64,
-            );
-            if let Some(t2) = transform2 {
-                t2 * Affine::new([a, b, c, d, e, f])
-            } else {
-                Affine::new([a, b, c, d, e, f])
-            }
+            Affine::new([sx, kx, ky, sy, tx, ty].map(f64::from))
         };
         match &*elt.borrow() {
             usvg::NodeKind::Group(_) => {}
@@ -110,7 +102,7 @@ pub fn render_tree_with<F: FnMut(&mut SceneBuilder, &usvg::Node) -> Result<(), E
                             if std::mem::take(&mut just_closed) {
                                 local_path.move_to(most_recent_initial);
                             }
-                            local_path.line_to::<Point>((p.x as f64, p.y as f64).into())
+                            local_path.line_to(Point::new(p.x as f64, p.y as f64))
                         }
                         usvg::tiny_skia_path::PathSegment::QuadTo(p1, p2) => {
                             if std::mem::take(&mut just_closed) {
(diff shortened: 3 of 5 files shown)


integrations/vello_svg/src/lib.rs
```diff
@@ -197,20 +229,21 @@ fn paint_to_brush(paint: &usvg::Paint, opacity: usvg::Opacity) -> Option<(Brush,
...
-            let end: vello::kurbo::Point = (gr.x2, gr.y2).into();
-            let transform = Affine::new([
-                gr.transform.a,
-                gr.transform.b,
-                gr.transform.c,
-                gr.transform.d,
-                gr.transform.e,
-                gr.transform.f,
-            ]);
+            let start: vello::kurbo::Point = (gr.x1 as f64, gr.y1 as f64).into();
```
DJMcNab, line 236: I think I'd use `Point::new` or `Point::from` here

integrations/vello_svg/src/lib.rs
```diff
@@ -115,9 +144,11 @@ pub fn render_tree_with<F: FnMut(&mut SceneBuilder, &usvg::Node) -> Result<(), E
                     if let Some((brush, brush_transform)) =
                         paint_to_brush(&fill.paint, fill.opacity)
                     {
-                        // FIXME: Set the fill rule
                         sb.fill(
-                            Fill::NonZero,
+                            match fill.rule {
+                                usvg::FillRule::NonZero => Fill::NonZero,
+                                usvg::FillRule::EvenOdd => Fill::EvenOdd,
+                            },
```
DJMcNab, line 151: 👍

integrations/vello_svg/src/lib.rs
```diff
@@ -33,78 +33,107 @@
...
+/// Re-export vello.
+pub use vello;
 
+/// Re-export usvg.
 pub use usvg;
 
-/// Append a [`usvg::Tree`] into a Vello [`SceneBuilder`], with default error handling
-/// This will draw a red box over (some) unsupported elements
+/// Append a [`usvg::Tree`] into a Vello [`SceneBuilder`], with default error
+/// handling This will draw a red box over (some) unsupported elements
```
DJMcNab, line 50: This is an odd change.

I must have failed to put two spaces after `handling`, previously. I'd propose at least putting a full stop between the two

integrations/vello_svg/src/lib.rs
```diff
@@ -33,78 +33,107 @@
...
-            Affine::new([a, b, c, d, e, f])
+            let usvg::Transform {
+                sx,
+                kx,
+                ky,
+                sy,
+                tx,
+                ty,
+            } = elt.abs_transform();
+            let (a, b, c, d, e, f) = (
```
DJMcNab, line 83: I'd be tempted to make a single array, then map it through `f64::from` and pass that straight to the constructor

integrations/vello_svg/src/lib.rs
```diff
@@ -33,78 +33,107 @@
...
+/// handling This will draw a red box over (some) unsupported elements
 ///
 /// Calls [`render_tree_with`] with an error handler implementing the above.
 ///
-/// See the [module level documentation](crate#unsupported-features) for a list of some unsupported svg features
-pub fn render_tree(sb: &mut SceneBuilder, svg: &usvg::Tree) {
-    render_tree_with(sb, svg, default_error_handler).unwrap_or_else(|e| match e {});
+/// See the [module level documentation](crate#unsupported-features) for a list
+/// of some unsupported svg features
+pub fn render_tree(sb: &mut SceneBuilder, svg: &usvg::Tree, transform: Option<Affine>) {
```
DJMcNab, line 56: Why is this transform needed? You can append a fragment created using this to a scene with an applied transform

integrations/vello_svg/src/lib.rs
```diff
@@ -33,78 +33,107 @@
...
                 let mut local_path = BezPath::new();
-                // The semantics of SVG paths don't line up with `BezPath`; we must manually track initial points
+                // The semantics of SVG paths don't line up with `BezPath`; we
+                // must manually track initial points
                 let mut just_closed = false;
                 let mut most_recent_initial = (0., 0.);
                 for elt in path.data.segments() {
                     match elt {
-                        usvg::PathSegment::MoveTo { x, y } => {
+                        usvg::tiny_skia_path::PathSegment::MoveTo(p) => {
```
DJMcNab, line 102: This tiny_skia_path identifier is interesting. Do you know the reason it was moved to this path?

integrations/vello_svg/src/lib.rs
```diff
@@ -197,20 +229,21 @@ fn paint_to_brush(paint: &usvg::Paint, opacity: usvg::Opacity) -> Option<(Brush,
...
-                gr.transform.a,
-                gr.transform.b,
-                gr.transform.c,
-                gr.transform.d,
-                gr.transform.e,
-                gr.transform.f,
-            ]);
+            let start: vello::kurbo::Point = (gr.x1 as f64, gr.y1 as f64).into();
+            let end: vello::kurbo::Point = (gr.x2 as f64, gr.y2 as f64).into();
+            let (a, b, c, d, e, f) = (
```
DJMcNab, line 238: Is this exactly the same code as before, but in a different style, or is it doing something different?

I'd expect both instances to be factored out if they're the same thing, although I see that the repetition is pre-existing

The new need to use `as f64` complicates things

integrations/vello_svg/src/lib.rs
```diff
@@ -225,23 +258,24 @@ fn paint_to_brush(paint: &usvg::Paint, opacity: usvg::Opacity) -> Option<(Brush,
...
                     cstop.color.a = (stop.opacity * opacity).to_u8();
-                    cstop.offset = stop.offset.get() as f32;
+                    cstop.offset = stop.offset.get();
                     cstop
                 })
                 .collect();
 
-            let start_center: vello::kurbo::Point = (gr.fx, gr.fy).into();
-            let end_center: vello::kurbo::Point = (gr.cx, gr.cy).into();
+            let start_center: vello::kurbo::Point = Point::new(gr.cx as f64, gr.cy as f64);
```
DJMcNab, line 266: Indeed, like these

integrations/vello_svg/src/lib.rs
DJMcNab: Ah, I see you're copying the extant style

In that case, I guess either is fine

integrations/vello_svg/src/lib.rs
```diff
@@ -33,78 +33,107 @@
...
-use vello::peniko::{Brush, Color, Fill};
-use vello::SceneBuilder;
+use vello::{
+    kurbo::{Affine, BezPath, Point, Rect, Stroke},
+    peniko::{Brush, Color, Fill},
+    SceneBuilder,
+};
+
+/// Re-export vello.
+pub use vello;
```
DJMcNab, line 44: Hmm, I'm less sure if we should do this, although I can't put my finger on why

integrations/vello_svg/src/lib.rs
```diff
@@ -33,78 +33,107 @@
...
 
+/// Re-export usvg.
```
DJMcNab, line 46: Thanks for adding docs 🙂

README.md
```diff
@@ -30,8 +30,7 @@ cargo run -p with_winit
 
 ### SVG
 
-This repository also includes [`vello_svg`](./integrations/vello_svg/), which supports converting 
-a [`usvg`](https://crates.io/crates/usvg) `Tree` into a Vello scene.
+This repository also includes [`vello_svg`](./integrations/vello_svg/), which supports converting a [`usvg`](https://crates.io/crates/usvg) `Tree` into a Vello scene.
```
DJMcNab, line 33: Can someone help explain why GitHub thinks this is different? It's not clear on mobile (android)

README.md
```diff
@@ -41,6 +40,11 @@ This is used in the [winit](#winit) example for the SVG rendering.
 
 A separate integration for playing Lottie animations is available through the [`velato`](https://github.com/linebender/velato) crate.
 
+### Community
+
+- [bevy-vello](https://github.com/vectorgameexperts/bevy-vello): A bevy integration to render lottie or svg files
```
DJMcNab, line 45: I'm not completely opposed to this, although I think it should be a different PR.

We get into a bit of a challenge of which things do we mention here, and what happens if they get out of date, for example - this might be better as an external page, like the Bevy examples page

I think it's a confused message to have both Velato and Velottie here - I'm not sure what we can do about that in this pr discussion though

.gitignore
```diff
@@ -1,6 +1,7 @@
 /target
 Cargo.lock
 /trace.json
+.DS_Store
```
DJMcNab, line 4: On principle, I'm opposed to this - it should be configured in your global gitignore, as it's globally applicable to your system

But practically, I think we should probably add this change. There's no cost to having something stupid in the gitignore.

DJMcNab (commented):
I've not run this yet, but some initial thoughts on the code/changes

This looks useful, thanks!

README.md
m-hugo: look at the line numbers, a newline was removed

README.md
simbleau: I'll remove it

integrations/vello_svg/src/lib.rs
simbleau: Sorry, what? Like this?

```rust
let x: [f64; 6] = [sx, ky, ...].into();
```

integrations/vello_svg/src/lib.rs
simbleau: These changes are detailed here: https://github.com/RazrFalcon/resvg/blob/master/CHANGELOG.md#0340---2023-05-27

Unfortunately the move to TinySkia and f32s in general means a ton of f64 casts.

integrations/vello_svg/src/lib.rs
simbleau: First, usvg switched to f32, reasons here:
https://github.com/RazrFalcon/resvg/blob/master/CHANGELOG.md#0340---2023-05-27

It is the same code as before, and was hard to get right. When usvg changed to [sx, kx, ky,...] instead of [a,b,c,d,e,f], it confused things for me. Fortunately I found this remapping helpful: https://github.com/RazrFalcon/resvg/blob/c6fd7f486aaef81704546f298541513f50f4f868/crates/c-api/lib.rs#L758

Usvg also uses column-major-column-vector matrix notation, therefore it’s ky-kx, not kx-ky.

.gitignore
simbleau: I'll remove it

integrations/vello_svg/src/lib.rs
simbleau: I also think it should be a common convention across the file. I'll try to make it consistent when I get to my next revision.

integrations/vello_svg/src/lib.rs
simbleau: Will be fixed. It was a bad auto-format.

.gitignore
Philipp-M: I would keep it, although I agree that this should probably be ignored globally, I'm not sure how many really have it configured (and then it's annoying, either when it's committed, or when it's always there as unstaged file)

integrations/vello_svg/src/lib.rs
DJMcNab: Yeah, so:
```rust
let arr = [...].map(f64::from);
```

(Where `...` is the components in the right order)

DJMcNab:
@simbleau just want to check on the status of this - I think we're waiting on your changes

simbleau:
> @simbleau just want to check on the status of this - I think we're waiting on your changes

Acknowledged. Will try to prioritize this today or tomorrow.

simbleau:
@DJMcNab Ready to go.

DJMcNab (approved):
Thanks! This works as expected locally and on Android

PR #423, opened 2024-01-24
Replace fello with skrifa for font rendering

dfrg:
This replaces our bespoke fello lib with skrifa from crates.io which brings us one step closer to being able to release.

Causes a slight regression since we no longer support hinting but that will be fixed when support lands upstream (soon!)

### DIFF
--- a/crates/encoding/src/glyph_cache.rs
+++ b/crates/encoding/src/glyph_cache.rs
@@ -52,6 +52,8 @@ impl GlyphCache {
             encoding_cache.encode_fill_style(fill);
             let mut path = encoding_cache.encode_path(true);
             let outline = outlines.get(GlyphId::new(key.glyph_id as u16))?;
+            // FIXME: Re-add hinting when skrifa supports it
+            // Tracking issue <https://github.com/googlefonts/fontations/issues/620>
             let draw_settings = skrifa::outline::DrawSettings::unhinted(size, coords);
             match style {
                 Style::Fill(_) => {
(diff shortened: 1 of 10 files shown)


Philipp-M (approved):
I've not got the deepest knowledge within vello + skrifa, but changes look good to me, and it works on my machine, thanks.

simbleau:
> hinting

What is hinting?

Philipp-M:
I think [wikipedia](https://en.wikipedia.org/wiki/Font_hinting) does a good job explaining that.

crates/encoding/src/glyph_cache.rs
```diff
@@ -49,18 +51,16 @@ impl GlyphCache {
             };
             encoding_cache.encode_fill_style(fill);
             let mut path = encoding_cache.encode_path(true);
+            let outline = outlines.get(GlyphId::new(key.glyph_id as u16))?;
+            let draw_settings = skrifa::outline::DrawSettings::unhinted(size, coords);
```
simbleau, line 55:
```suggestion
            // FIXME: Re-add hinting when skrifa supports it
            let draw_settings = skrifa::outline::DrawSettings::unhinted(size, coords);
```

simbleau (approved):

crates/encoding/src/glyph_cache.rs
dfrg: Sure, a little reminder won't hurt. Also added a link to the tracking issue.

dfrg:
> What is hinting?

The wikipedia article does a good job explaining the concept. Our fello crate does contain an implementation of hinting but it's not production quality. My current task at day job is making it so and landing it in skrifa so this should be done soon.

PR #425, opened 2024-01-25
Force encoding of transform/style following glyph run

### LINKED ISSUE #424: Glyph runs break transform/style optimization
Replacing the body of `SimpleText::add` with the following code appears to affect the transforms of objects drawn after the text. 

```rust
        let brush = brush.unwrap_or(&Brush::Solid(Color::WHITE));
        self.add_run(
            builder,
            font,
            size,
            brush,
            transform,
            None,
            Fill::NonZero,
            text,
        );
```
This only occurs in some cases but seems to be deterministic. I suspect it has something to do with the glyph run resolve code but after a bit of investigation I haven't been able to track it down yet, so leaving this issue as a reminder.

dfrg:
Fixes #424 which describes a case where encoding a glyph run breaks the optimization that omits duplicate transforms/styles from their respective streams.

This uses flags in the encoding to force the next transform/style to be emitted. An alternative implementation would be to unconditionally encode a default transform/style after each glyph run but this seemed like the better approach to me.

Also changes SimpleText::add to use the glyph run API. This means all text in the vello demos now uses that code path rather than encoding glyph outlines directly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ kurbo = "0.10.4"
 [workspace.dependencies]
 bytemuck = { version = "1.12.1", features = ["derive"] }
 skrifa = "0.15.4"
-peniko = { git = "https://github.com/linebender/peniko", rev = "629fc3325b016a8c98b1cd6204cb4ddf1c6b3daa" }
+peniko = { git = "https://github.com/linebender/peniko", rev = "8717635681dedfab3e9f3741fcbc7f3318a82ff0" }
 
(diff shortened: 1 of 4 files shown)


Philipp-M (approved):
I'm still barely beneath the surface of vello (time to dig a little bit deeper), so take this with a grain of salt.

I think I understand the issue, and agree with the solution though.

dfrg:
Thanks Philipp! Your reviews across the whole range of projects have been incredibly helpful in moving these changes forward.

Issue #411, opened 2023-11-14
Update raw-window-handle dependency


undersquire:
[Current version is 0.6](https://crates.io/crates/raw-window-handle/versions), while Vello is currently using 0.5 which causes issues when trying to use winit 0.29+.

xStrom:
For an immediate solution, in order to use `winit` 0.29+ you want to use the `rwh_05` feature of `winit`, which enables `raw-window-handle` 0.5.

Right now `wgpu` has no published release with `raw-window-handle` 0.6, although the support is present in `trunk` thanks to [wgpu#4202](https://github.com/gfx-rs/wgpu/pull/4202). Once `wgpu` gets a new release that contains these changes we will update Vello accordingly.

PR #427, opened 2024-01-30
chore: update wgpu to 0.19

crowlKats:
Also updates winit due to `raw-window-handle` changes

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -53,10 +53,10 @@ struct Args {
     use_cpu: bool,
 }
 
-struct RenderState {
+struct RenderState<'s> {
     // SAFETY: We MUST drop the surface before the `window`, so the fields
     // must be in this order
-    surface: RenderSurface,
+    surface: RenderSurface<'s>,
     window: Window,
 }
 
@@ -67,7 +67,7 @@ fn run(
     render_cx: RenderContext,
     #[cfg(target_arch = "wasm32")] render_state: RenderState,
 ) {
-    use winit::{event::*, event_loop::ControlFlow};
+    use winit::{event::*, event_loop::ControlFlow, keyboard::*};
     let mut renderers: Vec<Option<Renderer>> = vec![];
     #[cfg(not(target_arch = "wasm32"))]
     let mut render_cx = render_cx;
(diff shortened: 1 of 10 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -136,397 +136,412 @@ fn run(
...
-                                        .as_ref()
-                                        .or(profile_stored.as_ref())
+                                Key::Named(NamedKey::ArrowRight) => {
+                                    scene_ix = scene_ix.saturating_add(1)
+                                }
+                                Key::Named(NamedKey::ArrowUp) => complexity += 1,
+                                Key::Named(NamedKey::ArrowDown) => {
+                                    complexity = complexity.saturating_sub(1)
+                                }
+                                Key::Character(key @ "a") | Key::Character(key @ "e") => {
```
DJMcNab, line 167: Previously, rotate anti-clockwise was `q`. It might be worth checking whether your coding font makes the difference between Q and A sufficiently clear

Does rotation still work when caps lock is enabled with this change?

examples/with_winit/src/lib.rs
```diff
@@ -136,397 +136,412 @@ fn run(
...
+                                        transform = Affine::translate(prior_position)
+                                            * Affine::rotate(angle)
+                                            * Affine::translate(-prior_position)
+                                            * transform;
+                                    }
+                                }
+                                Key::Named(NamedKey::Space) => {
+                                    transform = Affine::IDENTITY;
+                                }
+                                Key::Character("s") => {
```
DJMcNab, line 180: Same concern about caps lock for all of these

DJMcNab (commented):
Looks good, thanks for doing this. Some comments from a first pass

I've not run the example yet

Can you please link #411?

examples/with_winit/src/lib.rs
crowlKats: whoops, thats a typo. fixed.

examples/with_winit/src/lib.rs
crowlKats: So regarding caps lock: I didn't consider it even, thinking winit would ignore modifiers, and act like it used to before.
I reworked the code: Winit does expose a way to ignore modifiers on key via a `KeyEventExtModifierSupplement` trait, however oddly enough its not exposed for a web target, so we cannot use it here, so now will just use `to_lowercase`.

examples/with_winit/src/lib.rs
```diff
@@ -136,397 +136,418 @@ fn run(
...
+                                Key::Named(NamedKey::Space) => {
+                                    transform = Affine::IDENTITY;
+                                }
+                                Key::Character(char) => {
+                                    let char = char.to_lowercase();
+                                    match char.as_str() {
+                                        "q" | "e" => {
+                                            if let Some(prior_position) = prior_position {
+                                                let is_clockwise = char == "e";
+                                                let angle = if is_clockwise { -0.05 } else { 0.05 };
```
DJMcNab, line 176: I'm seeing some weird behaviour here, where pressing `e` now rotates anticlockwise

However, this is also present on main, so that's not an issue from this PR.

examples/with_winit/src/lib.rs
```diff
@@ -136,397 +136,418 @@ fn run(
...
+                                                    * Affine::rotate(angle)
+                                                    * Affine::translate(-prior_position)
+                                                    * transform;
+                                            }
+                                        }
+                                        "s" => {
+                                            stats_shown = !stats_shown;
+                                        }
+                                        "d" => {
+                                            complexity_shown = !complexity_shown;
```
DJMcNab, line 187: This no longer works on my machine, but this is also a pre-existing issue from before this PR

DJMcNab (approved):
Sorry it's taken a while to review, I've taken a few weeks off.
Overall, this looks like a good update, thanks.

Unfortunately, this no longer compiles for Android. I needed to make two changes for that to work:
1) Adding the suggested `expect` - I have not yet dug into why this is now fallible in winit. I can do so if you want
2) Changing the run command to `cargo apk run -p with_winit --lib` (that is, the wrong command is documented in the README). I don't know why this change was necessary, but I also need it on main, so it's not a result of this PR.

We will likely add CI for android at some point, but that isn't currently present

ratmice:
I was just about to make a similar PR, and before doing it forgot to look if there were any outstanding.
The only major difference was making the `utils::create_surface` use the safe `create_surface` API, and the ramifications for that to `with_winit::RenderState` having to move the `window` variable out to just being in the `run` loop.

https://github.com/ratmice/vello/tree/update_wgpu

If that kind of change is wanted I could perhaps make a separate PR for it once this one goes in?

>Unfortunately, this no longer compiles for Android. I needed to make two changes for that to work:
>
>    Adding the suggested expect - I have not yet dug into why this is now fallible in winit. I can do so if you want

I had a bit of trouble figuring out which expect this is referring to, perhaps at the end of `run`?

DJMcNab:
The `expect` which is in the compilation failure message? This is at:
```rust
    let event_loop = EventLoopBuilder::with_user_event()
        .with_android_app(app)
        .build();
```
I haven't investigated why this is fallible yet

ratmice:
Ahh, I see.  The fallibility here is in the platform-independent part of winit:
[0.28 EventLoopBuilder::build](https://docs.rs/winit/0.28.7/winit/event_loop/struct.EventLoopBuilder.html#method.build) vs [0.29 EventLoopBuilder::build](https://docs.rs/winit/0.29.10/winit/event_loop/struct.EventLoopBuilder.html#method.build), where `with_android_app` returns a normal `EventLoopBuilder`.

It looks like it only errors in the case that you call `build()` multiple times, which in 0.28 was a panic.
So adding an `expect` there would just be continuing that `panic` tradition.

DJMcNab:
Closing, as this PR was merged as #444 

Thanks for kicking this off!

PR #431, opened 2024-02-12
Renumber resources in CPU path_count shader

raphlinus:
This cleans up a loose end in #421, which renumbered the resources in the path_count GPU shader but not the CPU shader.

It fixes an immediate panic when running with --use-cpu.


DJMcNab (approved):
This fixes the panic for me as well.

Thanks

PR #432, opened 2024-02-13
Remove SceneBuilder/Fragment

### LINKED ISSUE #342: Better composition of `SceneFragment` and `Scene`
Hello there,  I would like to do better composition of fragments and scenes but it's very limiting to do so (Or I missed some public API).

So I would like to propose few things:
1. `SceneFragment` and `Scene` should derive `Clone`.
2. Make their `Encoding` public.
3. Alllow me to modify `SceneFragment` without clearing it.

dfrg:
This removes both `SceneBuilder` and `SceneFragment` while moving all of the "builder" methods directly to `Scene`. Also derives `Clone` for `Scene`.

Fixes #342 and subsumes #243.

Based on #426 which must land first.

### DIFF
--- a/examples/scenes/src/lib.rs
+++ b/examples/scenes/src/lib.rs
@@ -16,7 +16,7 @@ pub use simple_text::SimpleText;
 pub use svg::{default_scene, scene_from_files};
 pub use test_scenes::test_scenes;
 
-use vello::{kurbo::Vec2, peniko::Color, SceneBuilder};
+use vello::{kurbo::Vec2, peniko::Color, Scene};
 
 pub struct SceneParams<'a> {
     pub time: f64,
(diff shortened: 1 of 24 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -376,7 +376,7 @@ fn run(
                 height,
                 antialiasing_method,
             };
-            let mut builder = SceneBuilder::for_scene(&mut scene);
+            scene.reset();
```
armansito, line 379: I like that this is explicit now (instead of implicitly resetting the scene by creating a new fragment.

src/scene.rs
```diff
@@ -20,9 +20,9 @@ use skrifa::instance::NormalizedCoord;
 use vello_encoding::{Encoding, Glyph, GlyphRun, Patch, Transform};
 
 /// Encoded definition of a scene and associated resources.
-#[derive(Default)]
+#[derive(Clone, Default)]
 pub struct Scene {
-    data: Encoding,
+    encoding: Encoding,
```
armansito, line 25: 👍

armansito (approved):
Everything looks good and runs without any issues. This change really improves the scene API and resolves some of the long standing issues around its usability. Thank you for tackling it!

examples/scenes/src/lib.rs
```diff
@@ -43,11 +43,11 @@ pub struct ExampleScene {
 }
 
 pub trait TestScene {
-    fn render(&mut self, sb: &mut SceneBuilder, params: &mut SceneParams);
+    fn render(&mut self, sb: &mut Scene, params: &mut SceneParams);
```
DJMcNab, line 46: `sb` here is short for `SceneBuilder`, which is of course no longer accurate

I'm not certain what we should change it to. Maybe `scene` would be fine?

examples/with_winit/src/lib.rs
dfrg: This was a definite pain point and we have the complaints to prove it :)

examples/scenes/src/lib.rs
dfrg: Oops, good catch. Renamed all of these (and occurrences of `builder`) to `scene`.

PR #435, opened 2024-02-13
Sweep gradients

dfrg:
Implements support for rendering sweep gradients. Just had an itch to get this done, mostly to support COLRv1 fonts.

Based on Skia's `SkSweepGradient` shader. 

Here's a comparison of a sweep gradient with the three extend modes. Skia on top and Vello on bottom.
skia-vello-sweep


examples/scenes/src/test_scenes.rs
```diff
@@ -683,10 +699,12 @@ fn gradient_extend(scene: &mut Scene, params: &mut SceneParams) {
     }
     let extend_modes = [Extend::Pad, Extend::Repeat, Extend::Reflect];
     for (x, extend) in extend_modes.iter().enumerate() {
-        for y in 0..2 {
-            let is_radial = y & 1 != 0;
+        for (y, kind) in [Kind::Linear, Kind::Radial, Kind::Sweep]
+            .into_iter()
+            .enumerate()
```
DJMcNab, line 704: These now go off-screen in the default resolution. I think the `resolution` field of `params` can be set to counteract this (see #453)

src/cpu_shader/draw_leaf.rs
```diff
@@ -107,6 +108,34 @@ fn draw_leaf_main(
...
+                        let inv_mat = [
+                            z[3] * inv_det,
+                            -z[1] * inv_det,
+                            -z[2] * inv_det,
+                            z[0] * inv_det,
+                        ];
+                        let inv_tr = [
+                            -(inv_mat[0] * z[4] + inv_mat[2] * z[5]),
+                            -(inv_mat[1] * z[4] + inv_mat[3] * z[5]),
+                        ];
```
DJMcNab, line 129: Is it worth adding the `inverse` method to `Transform` directly?

DJMcNab (approved):
Should this also be added to the `fine` cpu shader (on a best-effort basis, given we can't really test it)? We probably should have a coherent policy on that (see also #386)

Otherwise, I've checked code quality. Haven't checked the maths, but it works in the test scenes you've created, on both CPU and GPU

examples/scenes/src/test_scenes.rs
dfrg: I wasn't aware this existed! Added to this scene and made sure it scales properly.

dfrg:
> Should this also be added to the fine cpu shader (on a best-effort basis, given we can't really test it)? We probably should have a coherent policy on that (see also https://github.com/linebender/vello/pull/386)

Since we're still missing texture bindings in the CPU pipeline, we can't do much with this and my preference is to defer until that is more fully fleshed out.

I did however go ahead and sneak the CPU-side radial gradient fixes into this PR.

PR #437, opened 2024-02-14
Migrate to our Standard CI

DJMcNab:
See also linebender/peniko#21
I haven't added an MSRV yet, as we're not at that stage

This also lets the headless example use the `--use-cpu` flag. This isn't currently used in CI - I'm planning to make a `tests` crate/PR instead

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,23 @@
+env:
+  # We aim to always test with the latest stable Rust toolchain, however we pin to a specific
+  # version like 1.70. Note that we only specify MAJOR.MINOR and not PATCH so that bugfixes still
+  # come automatically. If the version specified here is no longer the latest stable version,
+  # then please feel free to submit a PR that adjusts it along with the potential clippy fixes.
+  RUST_STABLE_VER: "1.76" # In quotes because otherwise (e.g.) 1.70 would be interpreted as 1.7
+
+
+# Rationale
+#
+# We don't run clippy with --all-targets because then even --lib and --bins are compiled with
+# dev dependencies enabled, which does not match how they would be compiled by users.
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -1,55 +1,117 @@
...
-      # - run: cargo run_wasm -- -p with_winit --bin with_winit_bin --build-only
-      - name: Allow using WebGPU in web_sys
-        run: |
-          echo "RUSTFLAGS=--cfg=web_sys_unstable_apis" >> "$GITHUB_ENV"
-      - run: cargo check -p with_winit --target wasm32-unknown-unknown
+          components: clippy
+
+      - name: cargo clippy (wasm)
+        run: cargo clippy --all-targets --target wasm32-unknown-unknown -- -D warnings
+        env:
```
waywardmonkeys, line 97: This is an improvement!

crates/shaders/src/types.rs
```diff
@@ -20,6 +20,10 @@ pub enum BindType {
 }
 
 impl BindType {
+    // TODO: This is a public method, which means it definitely is not
+    // "dead code". However, rustc seems insitent that it is, and so to not
+    // block forward progress, I shall humour it
+    #[allow(dead_code)]
```
DJMcNab, line 26: See https://xi.zulipchat.com/#narrow/stream/255911-rust/topic/Confusing.20method.20is.20never.20used.20warning

I can't prioritise debugging and resolving this, so think we should land this without that

Philipp-M: Yeah I honestly think it's a bug in Rust, the error warning doesn't make sense to me as the type `BindType` is clearly public...

Philipp-M (approved):
Going towards a more unified CI makes sense. LGTM, thanks :)

PR #438, opened 2024-02-14
feat: svg strokes

simbleau:
Implements SVG strokes.

### DIFF
--- a/examples/scenes/src/download.rs
+++ b/examples/scenes/src/download.rs
@@ -200,7 +200,10 @@ impl SVGDownload {
             bail!("Size limit exceeded");
         }
         if limit_exact && file.stream_position().context("Checking file limit")? != size_limit {
-            bail!("Builtin downloaded file was not as expected");
+            bail!(
+                "Builtin downloaded file was not as expected. Expected {size_limit}, received {}.",
+                file.stream_position().context("Checking file limit")?
+            );
         }
         Ok(())
     }
--- a/examples/scenes/src/download.rs
+++ b/examples/scenes/src/download.rs
@@ -199,11 +199,13 @@ impl SVGDownload {
         if reader.read_exact(&mut [0]).is_ok() {
             bail!("Size limit exceeded");
         }
-        if limit_exact && file.stream_position().context("Checking file limit")? != size_limit {
-            bail!(
-                "Builtin downloaded file was not as expected. Expected {size_limit}, received {}.",
-                file.stream_position().context("Checking file limit")?
-            );
+        if limit_exact {
+            let bytes_downloaded = file.stream_position().context("Checking file limit")?;
+            if bytes_downloaded != size_limit {
+                bail!(
+                    "Builtin downloaded file was not as expected. Expected {size_limit}, received {bytes_downloaded}.",
+                );
+            }
         }
         Ok(())
     }
--- a/examples/scenes/src/download/default_downloads.rs
+++ b/examples/scenes/src/download/default_downloads.rs
@@ -41,7 +41,7 @@ pub(super) fn default_downloads() -> Vec<SVGDownload> {
         },
         SVGDownload {
             builtin:Some(BuiltinSvgProps {
-                info: "resvg",
+                info: "https://github.com/RazrFalcon/resvg-test-suite",
                 license: "MIT",
                 expected_size: 383,
             }),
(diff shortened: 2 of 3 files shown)


integrations/vello_svg/src/lib.rs
```diff
@@ -150,10 +150,28 @@ pub fn render_tree_with<F: FnMut(&mut Scene, &usvg::Node) -> Result<(), E>, E>(
...
-                        // linejoin, etc.
+                        let mut conv_stroke = Stroke::new(stroke.width.get() as f64)
+                            .with_caps(match stroke.linecap {
+                                usvg::LineCap::Butt => vello::kurbo::Cap::Butt,
+                                usvg::LineCap::Round => vello::kurbo::Cap::Round,
+                                usvg::LineCap::Square => vello::kurbo::Cap::Square,
+                            })
+                            .with_join(match stroke.linejoin {
+                                usvg::LineJoin::Miter | usvg::LineJoin::MiterClip => {
+                                    vello::kurbo::Join::Miter
```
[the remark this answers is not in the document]
DJMcNab: I think `Miter` is `MiterClip` - `Miter` in svg appears to be equivalent to `MiterClip` with an infinite limit?

DJMcNab (approved):
Is there a suitably licensed SVG we can add to the downloads to test this feature? I don't think any of existing SVGs have obvious strokes

Otherwise, code looks fine, thanks

DJMcNab:
You're also going to need to rebase past #437

simbleau:
> Is there a suitably licensed SVG we can add to the downloads to test this feature? I don't think any of existing SVGs have obvious strokes
> 
> Otherwise, code looks fine, thanks

Currently can't run because main panics for me.

But http://fabricjs.com/test/svg_import/stroke.html has some SVGs I used locally.

DJMcNab:
@simbleau I believe (as mentioned in [#gpu > main panic](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/main.20panic)) the panic was resolved in #444

If you can add this test svg (as an additional download rather than in the repository), we can merge this soon thereafter

Thanks

examples/scenes/src/download/default_downloads.rs
```diff
@@ -39,5 +39,65 @@ pub(super) fn default_downloads() -> Vec<SVGDownload> {
             url: "https://upload.wikimedia.org/wikipedia/commons/5/58/Coat_of_arms_of_the_Kingdom_of_Yugoslavia.svg".to_string(),
             name: "Coat of Arms of the Kingdom of Yugoslavia".to_string()
         },
+        SVGDownload {
+            builtin:Some(BuiltinSvgProps {
+                info: "resvg",
```
DJMcNab, line 44: I'd like this to be slightly more detailed - maybe a link to https://github.com/RazrFalcon/resvg-test-suite?

examples/scenes/src/download/default_downloads.rs
```diff
@@ -39,5 +39,65 @@ pub(super) fn default_downloads() -> Vec<SVGDownload> {
...
+                license: "MIT",
+                expected_size: 383,
+            }),
+            url: "https://raw.githubusercontent.com/RazrFalcon/resvg-test-suite/master/tests/painting/stroke-dashoffset/default.svg".to_string(),
```
DJMcNab, line 48: I'd consider using a fixed commit here, although knowing about an update is also good.

Maybe we need to just skip the specific item if the size doesn't match, rather than completely bailing out

examples/scenes/src/download.rs
```diff
@@ -200,7 +200,10 @@ impl SVGDownload {
             bail!("Size limit exceeded");
         }
         if limit_exact && file.stream_position().context("Checking file limit")? != size_limit {
-            bail!("Builtin downloaded file was not as expected");
+            bail!(
+                "Builtin downloaded file was not as expected. Expected {size_limit}, received {}.",
+                file.stream_position().context("Checking file limit")?
```
DJMcNab, line 205: We should use the same value here

DJMcNab (approved):

examples/scenes/src/download/default_downloads.rs
simbleau: If the size doesn't match the CLI already gives you the option to continue and ignore the error.

examples/scenes/src/download.rs
simbleau: Sorry, can you clarify? What are you expecting?

examples/scenes/src/download/default_downloads.rs
DJMcNab: I just forgot I implemented that lol. This is a non-issue then

examples/scenes/src/download.rs
simbleau: fixed

PR #439, opened 2024-02-14
CI and local smoke testing

DJMcNab:
To run the tests locally, use:
```rust
cargo test --workspace
```

These are placed in a new (non-publishing) `vello_tests` crate. The lib.rs of this crate has helpers for creating test cases, adapted from the headless example.

There is currently one additional set of tests, the `smoke.rs` tests. This has one test with two instantiations (over the value of `use_cpu`). This test is `simple_square`, which draws a 50px by 50px square, and confirms that 2500 pixels are red (and that the remainder are black).

There is a debug environment variable, which is `VELLO_DEBUG_TESTS`. This controls the debug output to PNG of these tests. These png are output into the `debug_outputs` folder. The environment variable takes a list of tests to debug, separated by commas. E.g. `VELLO_DEBUG_TESTS=simple_square_cpu` would debug the test `simple_square` with `use_cpu` enabled. The `cpu`/`gpu` suffices are optional, e.g. `VELLO_DEBUG_TESTS=simple_square` would output the png for both `simple_square_cpu` and `simple_square_gpu`.
There are also three special values:
- `all`, which debugs every test
- `cpu`, which debugs the CPU version of every test
- `gpu`, which debugs the GPU version of every test

In terms of CI, this uses Mesa (and thus lavapipe?) to run the shaders on the CPU of the runner. This is only possible on Ubuntu. In theory, windows has a fallback CPU implementation of DirectX, but I haven't dug into this yet


waywardmonkeys (approved):
I think this looks good overall.

PR #440, opened 2024-02-14
Bump Peniko to released version.

xorgy:


Cargo.toml
```diff
@@ -53,12 +53,12 @@ wgpu-profiler = { workspace = true, optional = true }
 # Note: we pick up a kurbo dependency transitively through peniko.
 # However, we need at least version 0.10.4 for proper stroking.
 # Next time we bump peniko, this can be removed.
-kurbo = "0.10.4"
+kurbo = "0.11.0"
```
ratmice, line 56: The one thing I notice is this "Next time we bump peniko, this can be removed." comment.

xorgy: Yeah I was looking into that.

xorgy: Works fine without it.

Philipp-M (approved):
LGTM and works for me, thanks.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,9 @@ env:
   # come automatically. If the version specified here is no longer the latest stable version,
   # then please feel free to submit a PR that adjusts it along with the potential clippy fixes.
   RUST_STABLE_VER: "1.76" # In quotes because otherwise (e.g.) 1.70 would be interpreted as 1.7
+  # If updating, ensure that it is also updated in Cargo.toml
+  # Explanation of the version choice
+  MINIMUM_SUPPORTED_RUST_VERSION: "1.70"
```
DJMcNab, line 9: I wouldn't expect this to do anything without the MSRV jobs being added

Cargo.toml
```diff
@@ -18,8 +18,11 @@ members = [
 edition = "2021"
 version = "0.0.1"
 license = "MIT OR Apache-2.0"
-# homepage = "https://vello.dev" - Domain owned by us, but unused at present
-# rust-version = 
+homepage = "https://vello.dev"
```
DJMcNab, line 21: Similarly to Peniko, I'd just remove this for now.

.github/workflows/ci.yml
xorgy: I think that's fair, but it's going to be there afterward anyhow so. :shrug:

Cargo.toml
xorgy: Since it at least goes somewhere, I think it's fine for now. If we publish it and later there's a real homepage there, it'll be better that it had been there.

.github/workflows/ci.yml
Philipp-M: Adding the jobs should probably be its own PR, I think it's ok to add this here though (that we at least know what the current MSRV would be), but well yeah it doesn't do anything.

Cargo.toml
xorgy: I'll remove this for now, since it is controversial.

.github/workflows/ci.yml
xorgy: Removed for now, since controversial.

DJMcNab:
I'd be happy to set an MSRV - it's easy to do, and we can always increase it later

I'd treat it the same way as Peniko, where it's not a promise, it's just an observation based on what testing we do.

It's just that if we are doing it, we should do it properly and check it in CI

PR #441, opened 2024-02-15
Stop ignoring Cargo.lock

DJMcNab:
This follows the guidance in https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

Previous discussion in #256

I've opened this as a draft, because it looks like it requires more CI work to ensure this is updated regularly.
I don't know that we want to commit to developing that


waywardmonkeys (commented):
I don't feel super strongly about this because I can (and have) argued both sides.

I suspect that if this does happen (or even if it doesn't), we should add CI for "minimal versions".

xStrom:
I've been doing some research into it and have a lot more to say about this (and MSRV) and will post a comprehensive explanation in not too distant of a future, hopefully before the end of the month. Until then the short of it is that it's very messy and complicated.

We currently have `Cargo.lock` committed for Xilem, so in that sense not a big issue to add it here for now. Once I post my findings, hopefully we can settle on a unified strategy across repos.

xStrom:
I did get to posting my comprehensive explanation. It is available for reading at [rfcs#5](https://github.com/linebender/rfcs/pull/5).

DJMcNab:
It seems like we'll have a more standardised experience, so I'll close this PR

PR #442, opened 2024-02-16
Add a workflow for testing Android compilation

DJMcNab:


waywardmonkeys (approved):
I don't know anything about `cargo apk` but other than that lack of knowledge, if that part is actually doing what it sounds like, then the rest of this is okay.

I'd not been sure about the addition of caching to the wasm job, but that looks like it is just making that match the other jobs, so okay...

Issue #433, opened 2024-02-13
Create a simpler example


PoignardAzur:
The current examples are too complex. We should create at least one single-file example, with no interaction. Something as dumb as "Draw a single shape (eg a circle) inside a winit window".

The example should fit in a single source file.

PR #443, opened 2024-02-16
Configure command line arguments in Android and WASM

DJMcNab:
Primarily, the two new environment variables are very useful:
1) `VELLO_STATIC_LOG` allows configuring logging for Android (equivalent to `RUST_LOG`, but at compile time)
2) `VELLO_STATIC_ARGS` allows configuring the command line arguments

For example,
```sh
VELLO_STATIC_LOG="with_winit=debug,vello::wgpu_engine=debug" VELLO_STATIC_ARGS="--test-scenes --scene 2 --use-cpu" cargo apk run -p with_winit --lib --release
```
would have debug level logging in `with_winit` and the `wgpu_engine` module, and start at the third test scene, using the CPU shaders

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -727,7 +727,7 @@ fn android_main(app: AndroidApp) {
     let config = android_logger::Config::default();
     // We allow configuring the Android logging with an environment variable at build time
     let config = if let Some(logging_config) = option_env!("VELLO_STATIC_LOG") {
-        let mut filter = env_logger::filter::Builder::new();
+        let mut filter = android_logger::FilterBuilder::new();
         filter.filter_level(log::LevelFilter::Warn);
         filter.parse(logging_config);
         let filter = filter.build();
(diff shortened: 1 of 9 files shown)


waywardmonkeys:
Since this is adding a dependency on `log`, this might be a good time (in a separate issue?) to discuss using `tracing` instead ... There's an open issue about `winit` moving to `tracing` as well.

examples/with_winit/src/lib.rs
```diff
@@ -655,16 +663,36 @@ use winit::platform::android::activity::AndroidApp;
...
+        let filter = filter.build();
+        // This shouldn't be needed in theory, but without this the max
+        // level is set to 0 (i.e. Off)
+        let config = config.with_max_level(filter.filter());
+        config.with_filter(filter)
+    } else {
+        config.with_max_level(log::LevelFilter::Warn)
+    };
+    android_logger::init_once(config);
+    let cache_dir = get_cache_directory(&app)
```
DJMcNab, line 681: This will be used for pipeline caching, but that isn't yet implemented. I could move this to a different PR, but it would be useful for me to have it merged, and I don't think there's any real cost to keeping it here (especially with #433, if with_winit is to be more of a testbed)

xorgy:
I like it.

xorgy (approved):

PR #444, opened 2024-02-19
Upgrade wgpu to 0.19

### LINKED ISSUE #411: Update raw-window-handle dependency

DJMcNab:
Subsumes #427 

Fixes #411

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -658,7 +658,8 @@ fn android_main(app: AndroidApp) {
 
     let event_loop = EventLoopBuilder::with_user_event()
         .with_android_app(app)
-        .build();
+        .build()
+        .expect("Required to continue");
     let args = Args::parse();
     let scenes = args
         .args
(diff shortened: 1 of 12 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -658,7 +658,8 @@ fn android_main(app: AndroidApp) {
 
     let event_loop = EventLoopBuilder::with_user_event()
         .with_android_app(app)
-        .build();
+        .build()
+        .expect("Required to continue");
```
ratmice, line 664: It looks like it only errors in the case that you call build() multiple times, which in 0.28 was a panic.
So adding an expect there would just be continuing that panic tradition.

So this seems fine to me.

DJMcNab: Thanks, that makes sense

ratmice (approved):

ratmice:
Overall this looked fine to me, some of the re-indentation is unfortunate just because it makes the patch a bit hard to read, but
with rustfmt c'est la vie.

PR #446, opened 2024-02-19
Use safe `create_surface` api

ratmice:
Here is an attempt at using the safe surface creation API, I only tested that it runs on wayland and compiles on wasm,
but I've never been able to run on WebGPU here (even without the patch), so I can't test that or other platforms.

Feel it is worth marking as draft because I haven't tested other platforms.


DJMcNab (approved):
Thanks!

ratmice:
I thought to double check that that `'static` lifetime was needed, I was previously needing it for the error

> Borrowed data escapes outside of closure

But somewhere in all the simplifications, that no longer seems to be a problem.  So I've gotten rid of that `'static` bound.

DJMcNab (approved):
Thanks for getting this done so quickly! 

You're going to need to rebase past #448 , but I don't think there should be any conflicts

PR #447, opened 2024-02-19
Update web-sys pin comment.

xStrom:
As we moved past `web-sys` 0.3.65 which the previous comment talked about I wondered if we can remove the pin. No we can't, because `wgpu` depends on an unstable API of `web-sys` that gets changed in patch versions. So I updated our comment to be more current and added a link to a recent [relevant `wgpu` issue](https://github.com/gfx-rs/wgpu/pull/5224).

### DIFF
--- a/examples/with_winit/Cargo.toml
+++ b/examples/with_winit/Cargo.toml
@@ -46,7 +46,8 @@ android_logger = "0.13.0"
 console_error_panic_hook = "0.1.7"
 console_log = "1"
 wasm-bindgen-futures = "0.4.33"
-# Note: pinning the exact dep here because 0.3.65 broke semver. Update this
-# when revving wgpu.
+# Note: pin web-sys to 0.3.67 because wgpu 0.19 depends on that exact version.
+# Update this when revving wgpu. Remove pin when wgpu no longer demands it.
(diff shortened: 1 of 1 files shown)


waywardmonkeys (approved):

examples/with_winit/Cargo.toml
```diff
@@ -46,7 +46,8 @@ android_logger = "0.13.0"
 console_error_panic_hook = "0.1.7"
 console_log = "1"
 wasm-bindgen-futures = "0.4.33"
-# Note: pinning the exact dep here because 0.3.65 broke semver. Update this
-# when revving wgpu.
+# Note: pin web-sys to 0.3.67 because wgpu 0.19 depends on that exact version.
+# Update this when revving wgpu. Remove pin when wgpu no longer demands it.
```
DJMcNab, line 50: This "no longer demands it" is hard to know how to action. I'm not sure that I have a better phrasing, though

DJMcNab (approved):
For clarity, the reason https://github.com/gfx-rs/wgpu/pull/5224#issuecomment-1934772931 doesn't apply to us is because we *do* require webgpu, and don't work with WebGL - meaning that we do unconditionally use the unstable features

PR #448, opened 2024-02-19
Run the smoke tests on macOS arm runners

DJMcNab:
https://developer.apple.com/documentation/paravirtualizedgraphics?language=objc might work here

This would be running actual Metal on actual GPUs, which is exciting

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories - these are included runners, as far as I can see.
I don't know whether we can accidentally use a chargeable runner - I hope not, but I'm not sure how to confirm that

I've also removed an outdated comment, as it's no longer useful

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,12 +39,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        # We use macos-14 as that is an arm runner. These have the virtgpu support we need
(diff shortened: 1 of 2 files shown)


Philipp-M (approved):
> I don't know whether we can accidentally use a chargeable runner

AFAIK all runners are free for public repos that are available without Github Team or Enterprise (which we don't have AFAIK)

LGTM otherwise, thanks :)

waywardmonkeys (approved):

PR #453, opened 2024-02-20
Additional test cases for co-linear stroked curves

armansito:
Ported over some additional test cases from Skia's "tricky cubic strokes" tests that exercise co-linear quadratic curves. These used to render incorrectly until the recent kurbo update, so it's good to have these in Vello.

These new curves are technically conic sections with cusps in them (with weights 1, 0.3, and 1.5). Vello doesn't yet support rational Beziers, so Skia encodes a conic as a path composed of quadratics (with a tolerance of 0.25 which scaled by the applied transform). The new test cases encode the results of this conic->quad conversion.

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use crate::{ExampleScene, SceneConfig, SceneParams, SceneSet};
-use vello::kurbo::{Affine, BezPath, Cap, Ellipse, Join, PathEl, Point, Rect, Stroke};
+use vello::kurbo::{Affine, BezPath, Cap, Ellipse, Join, PathEl, Point, Rect, Shape, Stroke};
 use vello::peniko::*;
 use vello::*;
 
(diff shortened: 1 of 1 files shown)


examples/scenes/src/test_scenes.rs
```diff
@@ -366,6 +429,35 @@ fn tricky_strokes(scene: &mut Scene, _: &mut SceneParams) {
...
+        bigger_flat_conic_as_quads.as_slice(),
+    ];
+    for quads in flat_curves.iter() {
+        let mut path = BezPath::new();
+        path.push(MoveTo((1., 1.).into()));
+        for quad in quads.iter() {
+            path.push(QuadTo(quad[0].into(), quad[1].into()));
+        }
+        let x = (idx % NUM_COLS) as f64 * CELL_SIZE;
+        let y = (idx / NUM_COLS) as f64 * CELL_SIZE;
```
DJMcNab, line 445: These now go off screen at the default window size

DJMcNab (approved):
Presuming that there are no licensing issues with these, and a resolution for the going offscreen question

I haven't really reviewed what the tests are achieving, but they all render the same on CPU and GPU, and the new tests appear.

xStrom:
We have a new Android CI check so you'll need to rebase on `main` to get that to pass.

examples/scenes/src/test_scenes.rs
armansito: I can adjust the CELL_SIZE to adapt to the number of rows (which fixes this on my external monitor).

The layout in general is sensitive to DPI, e.g. at the default window dimensions, everything fits and looks tiny on my MacBook's retina display, but enlarged on my external monitor. I've observed this with all the other test scenes too.

armansito: The test scene now initializes the `resolution` field of `SceneParams`, which fixes the issue for me.

PR #455, opened 2024-02-21
Allow initialising shaders in parallel

DJMcNab:
Headline stats: Reduces shader initialisation time from ~5seconds on Pixel 6 to ~2 seconds (with no compromises on runtime performance)

This is disabled on wasm because there are no threads.
This is disabled on macOS, following the prior art in Bevy (https://github.com/bevyengine/bevy/pull/10812#discussion_r1496138004)

This work queuing system is not intended to be the final form of this, which is why I have made this optional.
This would not be suitable for many consumers, who will likely already have their own thread pool.
We currently just create short lived threads using [`std::thread::scope`](https://doc.rust-lang.org/std/thread/fn.scope.html).

I still think this is useful to have for experimental purposes, and I think it should also be useful in Xilem in its current form. @xorgy might need you to confirm that

This is not async shader initialisation, which I do also plan to add. This is a useful first step towards that, however

### DIFF
--- a/crates/tests/src/lib.rs
+++ b/crates/tests/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{env, fs::File, path::Path, sync::Arc};
+use std::{env, fs::File, num::NonZeroUsize, path::Path, sync::Arc};
 
 use anyhow::{anyhow, bail, Result};
 use vello::{
@@ -62,7 +62,7 @@ pub async fn render(scene: Scene, params: &TestParams) -> Result<Image> {
         RendererOptions {
             surface_format: None,
             use_cpu: params.use_cpu,
-            initialise_in_parallel: false,
+            num_init_threads: NonZeroUsize::new(1),
             antialiasing_support: vello::AaSupport::area_only(),
         },
     )
--- a/crates/tests/src/lib.rs
+++ b/crates/tests/src/lib.rs
@@ -62,6 +62,7 @@ pub async fn render(scene: Scene, params: &TestParams) -> Result<Image> {
         RendererOptions {
             surface_format: None,
             use_cpu: params.use_cpu,
+            initialise_in_parallel: false,
             antialiasing_support: vello::AaSupport::area_only(),
         },
     )
--- a/src/wgpu_engine.rs
+++ b/src/wgpu_engine.rs
@@ -159,16 +159,19 @@ impl WgpuEngine {
         use std::num::NonZeroUsize;
 
         if let Some(mut new_shaders) = self.shaders_to_initialise.take() {
-            let num_threads = num_threads.map(NonZeroUsize::get).unwrap_or_else(|| {
-                // Fallback onto a heuristic. This tries to not to use all threads.
-                // We keep the main thread blocked and not doing much whilst this is running,
-                // so we broadly leave two cores unused at the point of maximum parallelism
-                // (This choice is arbitrary, and could be tuned, although a 'proper' threadpool
-                // should probably be used instead)
-                std::thread::available_parallelism().map_or(2, |it| it.get().max(4) - 2)
-            });
+            let num_threads = num_threads
+                .map(NonZeroUsize::get)
+                .unwrap_or_else(|| {
+                    // Fallback onto a heuristic. This tries to not to use all threads.
+                    // We keep the main thread blocked and not doing much whilst this is running,
+                    // so we broadly leave two cores unused at the point of maximum parallelism
+                    // (This choice is arbitrary, and could be tuned, although a 'proper' threadpool
+                    // should probably be used instead)
+                    std::thread::available_parallelism().map_or(2, |it| it.get().max(4) - 2)
+                })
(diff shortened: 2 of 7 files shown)


crates/tests/src/lib.rs
```diff
@@ -62,6 +62,7 @@ pub async fn render(scene: Scene, params: &TestParams) -> Result<Image> {
         RendererOptions {
             surface_format: None,
             use_cpu: params.use_cpu,
+            initialise_in_parallel: false,
```
ratmice, line 65: I wonder if it would be better to instead of a boolean control this setting via the number of threads,
such as  `num_initialisation_threads: Option<NonZeroUsize>`, with `None` using the heuristics, or
similar to rayon where instead it uses `usize` with `0 == heuristics` such as [rayon's `num_threads`](https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.num_threads)

DJMcNab: Yeah, I think that's a good idea, thanks

xorgy:
Is there any reason somebody would want _not_ to initialize in parallel if it is supported on the platform? Memory?

xorgy:
> I still think this is useful to have for experimental purposes, and I think it should also be useful in Xilem in its current form. @xorgy might need you to confirm that

I'll give it a go and let you know.

DJMcNab:
> Is there any reason somebody would want _not_ to initialize in parallel if it is supported on the platform? Memory?
> 
> (talking about aside from cases where people have their own thread pooling)

There is also a possibility of starving other threads of the resources they need. And yeah, there is some additional (temporary) memory usage required, although I think (but haven't checked exact numbers) that the amount should be minimal

crates/tests/src/lib.rs
xorgy: Could use futures underneath, have a default strategy (just fire up threads), and then have an `enum`/`Fn` interface for hooking in your own dispatch.

xorgy: @ratmice `Option<NonZeroUsize>` is guaranteed to be one word. :+ )

ratmice: Happy to resolve this following [aa0819e](https://github.com/linebender/vello/pull/455/commits/aa0819eb191ac96ff25ce26fa09b940ca2a50bcd), but leaving unresolved in case you want to tackle the futures aspect mentioned above under this too.

waywardmonkeys:
> This is disabled on macOS, following the prior art in

I think you left off something...

DJMcNab:
Thanks, resolved

src/wgpu_engine.rs
```diff
@@ -130,6 +140,77 @@ impl WgpuEngine {
         }
     }
 
+    /// Enable creating any remaining shaders in parallel
+    pub fn use_parallel_initialisation(&mut self) {
+        if self.shaders_to_initialise.is_some() {
```
ratmice, line 146: Scraping the bottom of the barrel a bit, but I suppose that my main question here I guess is since `use_parallel_initialisation` is called immediately after `WgpuEngine::new()`,  and `Vec::new` doesn't allocate until `push`, whether it would be better to just do this initialisation there. 

But came to the realization that since neither `new` or `add_shader` have any hint that things should/should not be conducted in parallel, so to achieve that we would have to add a parameter to `WgpuEngine::new()` unfortunately.
absent that I wonder if it is worth just making this `pub (crate)`, and perhaps also `build_shaders_if_needed`?

it seems fine to disregarding this if there is some reason I don't understand why this should be `pub` or separate from `new`.

DJMcNab: This isn't actually `pub`, in the sense that `wgpu_engine` isn't externally reachable.

There's one more constraint from future work which might be more enlightening. One of the plans I have to improve startup times further is starting up by using the cpu shaders.
That is, the asynchronous shader compilation I mentioned in the PR description.

However, this would only apply to shaders *other* than fine_area. That is, the call order would be:
```rust
add_shader(fine_area);
use_parallel_initialisation();
add_other_shaders();
build_shaders_if_needed_async();
```

ratmice: Ahh, sorry I noticed that `use wgpu`, but for some reason I read the later `use wgpu::{.., WgpuEngine}` as being `pub`, stand corrected.
I guess that I would still consider moving `use_parallel_initialisation` into `WgpuEngine::new`, but also adding a `parallel` flag to `add_shader` like:

```rust
WgpuEngine::new(...);
add_shader(fine_area, false);
add_other_shaders(); // presumable this calls add_shader(..., true);
build_shaders_if_needed_async();
```

But since this isn't pub as I thought, we aren't actually locking ourselves into any one of these design decisions,
which was my main concern. Because of that i'm happy to resolve this.

ratmice (approved):
Given that the bits I was worried about commitment to semver for it turns out are not actually `pub` at the crate level, this seems fine to me.

PR #456, opened 2024-02-22
Standardize copyright headers.

xStrom:
The Linebender standard for copyright headers was decided last year in [kurbo#207](https://github.com/linebender/kurbo/issues/207) as the following:
```
// Copyright <year of file creation> the <Project> Authors
```

 With Vello 0.1 approaching fast it makes sense to get this done now.

This PR converts all file headers to this standard form, including the standard capitalization.

Additionally, three CPU shader files didn't specify Unlicense in the headers which seems like a mistake. I added the Unlicense identifier to match all the other shaders.

### DIFF
--- a/examples/with_winit/src/multi_touch.rs
+++ b/examples/with_winit/src/multi_touch.rs
@@ -1,3 +1,6 @@
+// Copyright 2023 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
 /// Adapted from https://github.com/emilk/egui/blob/212656f3fc6b931b21eaad401e5cec2b0da93baa/crates/egui/src/input_state/touch_state.rs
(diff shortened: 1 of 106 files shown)


waywardmonkeys (approved):
Just curious, what is the difference between files with 2022 and 2023? Creation date?

xStrom:
Yeap, the year marks the creation date.

examples/with_winit/src/multi_touch.rs
```diff
@@ -1,3 +1,6 @@
+// Copyright 2023 the Vello Authors
```
DJMcNab, line 1: I'm not sure this is right. It's mostly the work of the egui authors

xStrom: Good catch! I've created #464 to address this.

PR #461, opened 2024-02-23
Add simpler example

### LINKED ISSUE #433: Create a simpler example

anantnrg:
This PR introduces a new example named `shapes` which draws a stroked rectangle, a filled circle, a filled ellipse and a line to a Winit window. It fits in a single file and is very minimal. The code is commented (can add more if needed) which makes it easier for beginners to get started.

Closes #433 (hopefully).

Any suggestions or criticisms are welcome. Cheers!

### DIFF
--- a/examples/shapes/Cargo.toml
+++ b/examples/shapes/Cargo.toml
@@ -4,10 +4,14 @@ edition.workspace = true
 version.workspace = true
 license.workspace = true
 repository.workspace = true
+publish = false
 
+# The dependencies here are independent from the workspace versions
 [dependencies]
 anyhow = "1.0.80"
 pollster = "0.3.0"
 wgpu = "0.19.1"
 winit = "0.29.10"
-vello = { git = "https://github.com/linebender/vello" }
+vello = { path = "../../" }
+# When using, replace the above line with this:
+#vello = { git = "https://github.com/linebender/vello" }
--- a/examples/shapes/Cargo.toml
+++ b/examples/shapes/Cargo.toml
@@ -5,8 +5,6 @@ version.workspace = true
 license.workspace = true
 repository.workspace = true
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 anyhow = "1.0.80"
 pollster = "0.3.0"
--- a/examples/shapes/Cargo.toml
+++ b/examples/shapes/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "shapes"
+edition.workspace = true
+version.workspace = true
+license.workspace = true
+repository.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.80"
+pollster = "0.3.0"
+wgpu = "0.19.1"
+winit = "0.29.10"
+vello = { git = "https://github.com/linebender/vello" }
--- a/examples/shapes/src/main.rs
+++ b/examples/shapes/src/main.rs
@@ -1,3 +1,6 @@
+// Copyright 2024 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
 use anyhow::Result;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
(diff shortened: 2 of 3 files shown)


examples/shapes/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+    // Create a RenderState wrapped in an Option.
+    let mut render_state = None::<RenderState>;
+    // Cache a window so that it can be reused when the app is resumed after being suspended
+    let mut cached_window = None;
+    // Create a new scene
+    let mut scene = Scene::new();
+
+    // Run the event loop
+    event_loop
+        .run(move |event, event_loop| match event {
```
DJMcNab, line 45: Four of these five comments don't give any information which can't be trivially inferred from the line of code they're attached to.
I think these are likely to add comment fatigue, meaning that users may skip reading the actually useful comments later

examples/shapes/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+                        let circle = Circle::new((420.0, 200.0), 120.0);
+                        let circle_fill_color = Color::rgb(0.9529, 0.5451, 0.6588);
+
+                        let ellipse = Ellipse::new((250.0, 420.0), (100.0, 160.0), -90.0);
+                        let ellipse_fill_color = Color::rgb(0.7961, 0.651, 0.9686);
+
+                        let line = Line::new((260.0, 20.0), (620.0, 100.0));
+                        let line_stroke_color = Color::rgb(0.5373, 0.7059, 0.9804);
+
+                        // Draw the shapes!
```
DJMcNab, line 104: Where does the scene get reset? My reading of this code suggests that the same shapes will be drawn on top each other, repeated for each frame.

This would mean performance would get worse as the program runs, which isn't really the impression we want to give

examples/shapes/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+        .expect("Couldnt run event loop");
+}
+
+fn main() -> Result<()> {
+    // Create the event loop and the rendex context
+    let event_loop = EventLoop::new()?;
+    let render_cx = RenderContext::new().unwrap();
+
+    // Run the event loop
+    run(event_loop, render_cx);
```
DJMcNab, line 191: Since there are no other platforms supported here, I would consider inlining the run function

DJMcNab (commented):
Thanks for putting this together. I have a few nits from an initial scan.

examples/shapes/src/main.rs
DJMcNab: The comment you've left wasn't the one which is informative.
```rust
// Cache a window so that it can be reused when the app is resumed after being suspended
```
is a useful comment - I don't think the others added anything though

examples/shapes/src/main.rs
DJMcNab: You've inlined the arguments to the run function, but I meant was suggesting inlining the function itself

anantnrg: So just remove the `run` function and put all the code in `main`?

examples/shapes/src/main.rs
DasLixou: look one line above

DJMcNab: Thanks, although that was added in response to my comment :laughing:

examples/shapes/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "shapes"
+edition.workspace = true
+version.workspace = true
+license.workspace = true
+repository.workspace = true
+
```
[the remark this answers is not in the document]
anantnrg: This should be fixed right?

examples/shapes/src/main.rs
DasLixou: My question would be why resetting and making new in the first place

examples/shapes/src/main.rs
```diff
@@ -0,0 +1,183 @@
+use anyhow::Result;
```
xStrom, line 1: Please add our copyright header to the new file.

```suggestion
// Copyright 2024 the Vello Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT

use anyhow::Result;
```

anantnrg: Added!

xStrom:
Our CI scripts have changed, so you need to rebase on `main` to get the CI checks to pass.

anantnrg:
Fixed @xStrom!

PoignardAzur (approved):
Solid contribution!

anantnrg:
Thanks :)

Issue #449, opened 2024-02-19
Artifacts on Windows with GPU


DasLixou:
Video from zulip thread: https://xi.zulipchat.com/user_uploads/8784/cbAwPeKKof9rsMvN9E504Gt4/20240219-1541-03.0118905.mp4

Latest main commit

GPU: `AMD Radeon RX 6600`

DasLixou:
Fixed by updating graphic card driver.
Now the console gets spammed full with:
```
[2024-02-23T20:20:15Z ERROR wgpu_hal::auxil::dxgi::exception] ID3D12CommandQueue::ExecuteCommandLists: Using IDXGISwapChain::Present on Command List (0x000002CA5AEE0FD0:'Internal DXGI CommandList'): Resource state (0x1DB321A0: D3D12_RESOURCE_STATE_RENDER_TARGET) of resource (0x000002CA5B831CF0:'Unnamed ID3D12Resource Object') (subresource: 0) is invalid for use as a PRESENT_SOURCE.  Expected State Bits (all): 0x1DB32180: D3D12_RESOURCE_STATE_[COMMON|PRESENT], Actual State: 0x1DB32160: D3D12_RESOURCE_STATE_RENDER_TARGET, Missing State: 0x0: D3D12_RESOURCE_STATE_[COMMON|PRESENT]. [ EXECUTION ERROR #538: INVALID_SUBRESOURCE_STATE]
[2024-02-23T20:20:15Z ERROR wgpu_hal::auxil::dxgi::exception] ID3D12CommandQueue::ExecuteCommandLists: Using ClearRenderTargetView on Command List (0x000002CA2338D6E0:'Unnamed ID3D12GraphicsCommandList Object'): Resource state (0x1DB329C0: D3D12_RESOURCE_STATE_[COMMON|PRESENT]) of resource (0x000002CA233532E0:'Unnamed ID3D12Resource Object') (subresource: 0) is invalid for use as a render target.  Expected State Bits (all): 0x1DB329A0: D3D12_RESOURCE_STATE_RENDER_TARGET, Actual State: 0x1DB32980: D3D12_RESOURCE_STATE_[COMMON|PRESENT], Missing State: 0x4: D3D12_RESOURCE_STATE_RENDER_TARGET. [ EXECUTION ERROR #538: INVALID_SUBRESOURCE_STATE]
[...]

PR #462, opened 2024-02-24
Support `WGPU_BACKEND` and simplify shader util function

DasLixou:
Two mini patches:
* making the `read_draw_tag_from_scene` wgsl function more equal to it's cpu equivalent, just from guessing I think this would safe 2 cycles (because 2 add operations are now either 0 or 1) but that's too much of a microoptimization, I just like to think about that.
* Taking `WGPU_BACKEND` env variable into selection of the backend, which enables me to get rid of the dx12 error spam mentioned in #449 and also bumps my fps from 80-85 on dx12 to 100-120 on vulkan.

### DIFF
--- a/shader/shared/util.wgsl
+++ b/shader/shared/util.wgsl
@@ -13,9 +13,9 @@
 //    * `scene`: array<u32>
 //    * `config`: Config (see config.wgsl)
 fn read_draw_tag_from_scene(ix: u32) -> u32 {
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Thanks, both of these look good.

I've also tested the backend selection locally, and was able to get the gles backend working with no apparent compromises

PR #463, opened 2024-02-25
Add copyright header check to CI.

xStrom:
Since #456 was merged we have standardized copyright headers here. This PR adds a check to the CI to ensure that we keep them nice and consistent.

The script also has two things to quickly support a more complex future.

1. Additional copyright owners in the format of:
```
// Copyright 2024 [Other Text and] the Vello Authors [and Other Text]
```
This has been used for example in Peniko as *the Peniko Authors and the Piet Authors*.

2. Nearby comment with examples on how to add potential new files or directories as exclusions from this check. Currently we don't have anything to exclude, but it's possible we might in the future. The comment should help reduce confusion on how to deal with it.

Worth noting that the ripgrep command itself is cross-platform. However the bash script obviously isn't. However at this point that isn't a big deal, because we don't need to check the headers on every platform. So we just piggyback on the `rustfmt` Ubuntu job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,14 @@ jobs:
       - name: cargo fmt
         run: cargo fmt --all --check
 
+      - name: install ripgrep
+        run: |
+          sudo apt update
+          sudo apt install ripgrep
+
+      - name: check copyright headers
+        run: bash .github/copyright.sh
+
   test-stable:
     runs-on: ${{ matrix.os }}
     strategy:
--- a/.github/copyright.sh
+++ b/.github/copyright.sh
@@ -2,9 +2,9 @@
 
 # Debugging CI quirk
 echo "First run, should be empty:"
-rg "^// Copyright (19|20)[\d]{2} the Vello Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!{shader,src/cpu_shader}"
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,14 @@ jobs:
       - name: cargo fmt
         run: cargo fmt --all --check
 
+      - name: install ripgrep
+        run: |
+          sudo apt update
```
waywardmonkeys, line 40: Does this need a `-y` or is that just something people copy and paste and isn’t actually needed?

waywardmonkeys (approved):
LGTM with one question.

.github/workflows/ci.yml
xStrom: That's a good question. We haven't had `-y` in any of our scripts previously and haven't had a single problem. So maybe GitHub actions have `APT::Get::Assume-Yes` pre-configured? I can look a bit into it later.

xStrom:
Well this worked just fine on my local Ubuntu install, but the GitHub Action runner seems to print out `<stdin>` which causes an error. I'll need to investigate what's going on here, but that will have to wait a bit as I'm out of time right now.

xStrom:
Explicitly setting the search path helped solved that issue. I also tested that the CI properly notices invalid headers with some bogus commits that I later reverted.

This is now ready.

.github/workflows/ci.yml
xStrom: Indeed that was the case. You can see in the [official GitHub Ubuntu image setup script](https://github.com/actions/runner-images/blob/df722a3cf81bbe65c3b964c9b4830c85451542fc/images/ubuntu/scripts/build/configure-apt.sh#L19):
```sh
# Configure apt to always assume Y
echo "APT::Get::Assume-Yes \"true\";" > /etc/apt/apt.conf.d/90assumeyes
```

So `-y` is implicit for GitHub runners and doesn't have to be explicit.

waywardmonkeys (approved):

PR #464, opened 2024-02-25
Add proper attribution to egui authors.

xStrom:
The multi touch example should have a clearer attribution to the original authors. The licenses still match.

### DIFF
--- a/examples/with_winit/src/multi_touch.rs
+++ b/examples/with_winit/src/multi_touch.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 the Vello Authors
+// Copyright 2021 the egui Authors and Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 /// Adapted from https://github.com/emilk/egui/blob/212656f3fc6b931b21eaad401e5cec2b0da93baa/crates/egui/src/input_state/touch_state.rs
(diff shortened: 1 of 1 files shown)


examples/with_winit/src/multi_touch.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 the Vello Authors
+// Copyright 2021 the egui Authors and Vello Authors
```
DJMcNab, line 1: The name used in egui's (MIT) [license](https://github.com/emilk/egui/blob/5cf99c6308e36b5f36861accea6835f190924cce/LICENSE-MIT) is:

> Copyright (c) 2018-2021 Emil Ernerfeldt <emil.ernerfeldt@gmail.com>

Maybe we should ask the author for their preference?

xStrom: The file in question was mostly written by [Ivo Vollrath](https://github.com/quadruple-output) as seen in [egui#306](https://github.com/emilk/egui/pull/306). Beyond that, there have been other contributors.

Given that there are multiple contributors and that the original file doesn't have a header at all, I think it's fine to reference them as *egui Authors*.

DJMcNab (approved):
This is certainly an improvement, thanks.

PR #471, opened 2024-02-26
Refactoring `Recording`

DasLixou:
* Done some documenting over `Command`s and `Recording`.
* `engine.rs` is now called `recording.rs`
* `Id` is now `ResourceId` and uses `NonZeroU64::new_unchecked` as mentioned in the comment
* Some structs/variants with `Buf` got replaced with `Buffer` because `Image` is also not shortened as `Img`.

There are some comments and functions left in `Recording` where I don't know what they do in detail yet so they're left out. But reviewers can insert them of course 🫠

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,7 @@
 
 mod cpu_dispatch;
 mod cpu_shader;
-mod engine;
+mod recording;
 mod render;
 mod scene;
 mod shaders;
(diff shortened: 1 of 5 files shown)


DasLixou:
YAY MY COMMIT IS THERE

PoignardAzur (approved):

PR #472, opened 2024-02-26
Rename `scene.reset()` to `scene.clear()`

### LINKED ISSUE #469: `Scene::reset()` should be renamed to `Scene::clear()`
Should be self-explanatory.

If we visualize Scene as a container for encoded draw commands, `clear()` is the obvious name.

DasLixou:
Closes #469

### DIFF
--- a/crates/encoding/src/encoding.rs
+++ b/crates/encoding/src/encoding.rs
@@ -70,7 +70,7 @@ impl Encoding {
     }
 
     /// Clears the encoding.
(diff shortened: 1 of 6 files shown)


DasLixou:
Closing this and the issue then? (Issue is not from me)

xStrom:
We decided at office hours 2024-02-29 that we will not go ahead with this rename. Thank you for your effort anyway!

PR #473, opened 2024-02-27
More refactoring

DasLixou:
* Renamed `Recording` to `Workflow`
* Made `commands` non-public and made use of `commands()` function
* Move all the wgpu rendering code out of `lib.rs` into seperate file

# Questions
- Why is `wgpu_renderer` even there? Vello is very low level in some cases but then it holds our hands with drawing to the window?

### DIFF
--- a/src/wgpu_renderer.rs
+++ b/src/wgpu_renderer.rs
@@ -97,7 +97,7 @@ impl Renderer {
         texture: &TextureView,
         params: &RenderParams,
     ) -> Result<()> {
-        let (recording, target) = render::render_full(scene, &self.shaders, params);
+        let (workflow, target) = render::render_full(scene, &self.shaders, params);
         let external_resources = [ExternalResource::Image(
             *target.as_image().unwrap(),
             texture,
@@ -105,7 +105,7 @@ impl Renderer {
         self.engine.run_workflow(
             device,
             queue,
-            &recording,
+            &workflow,
             &external_resources,
             "render_to_texture",
             #[cfg(feature = "wgpu-profiler")]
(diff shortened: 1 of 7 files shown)


src/wgpu_renderer.rs
```diff
@@ -0,0 +1,478 @@
+// Copyright 2024 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
xStrom, line 2: Given that this new file is essentially a copy-paste of an old file, let's keep the old file's copyright year of 2022.

```suggestion
// Copyright 2022 the Vello Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT
```

raphlinus:
I'm not a huge fan of "Workflow" as a name; it suggests something higher level. I'm not sure what the best name is; "recording" is borrowed from Skia Graphite. It's actually an abstracted command buffer, with a bit of similarity to a render graph, but neither of those are particularly good stems for snappy names.

DasLixou:
What's with `Commands` then?

armansito:
> * Renamed `Recording` to `Workflow`

I'm not a fan of this rename. The term "recording" is actually not uncommon in this context: with modern GPU APIs, the CPU side "records" a series of commands and submits them to the GPU.

The general idea is that the `Recording` type is meant to encapsulate the pipeline dispatch sequence as well as the resource dependencies in an abstract way, and a GPU API backend-aware engine knows how to "play" the Recording. We don't expose this to clients today (IIRC) but in theory a client could build a `Recording` and "replay" it many times.

> * Why is `wgpu_renderer` even there? Vello is very low level in some cases but then it holds our hands with drawing to the window?

Initially we intended to have different backend runners/engines (such as wgpu, metal, vulkan, etc) and built an abstraction around a GPU command "Recording" that can get executed by a backend engine. As you've noticed, the backend abstraction isn't entirely clean cut since the main vello crate assumes wgpu support.

I posted a similar response to your [zulip thread](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Splitting.20backends.20into.20seperate.20crates.3F), I hope it can provide some more context.

DJMcNab:
I don't think this PR is likely to go anywhere at this point. Thanks for the interest, and you definitely have accurately raised that our API design needs more analysis.

If you wanted to rework this to have only the change with moving the `wgpu` using things into their own file, that would be welcome. But the other changes are less clearly improvements, as discussion has shown. If not, I'd suggest closing this out

PR #475, opened 2024-02-29
Fix negative duration panic on macos

Enter-tainer:
The `winit` example sometimes panics when running on macos. This pr fix this by adding a `.max(0.0)` to clip it to 0 before making a `Duration`

```
can not convert float seconds to Duration: value is negative
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_display
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:196:5
   3: core::time::Duration::from_secs_f64::panic_cold_display
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panic.rs:99:13
   4: core::time::Duration::from_secs_f64
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/time.rs:762:23
   5: with_winit::stats::draw_gpu_profiling::{{closure}}
             at ./examples/with_winit/src/stats.rs:389:17
   6: with_winit::stats::traverse_profiling
             at ./examples/with_winit/src/stats.rs:443:9
   7: with_winit::stats::draw_gpu_profiling
             at ./examples/with_winit/src/stats.rs:348:5
   8: with_winit::run::{{closure}}
             at ./examples/with_winit/src/lib.rs:418:33
```

### DIFF
--- a/examples/with_winit/src/stats.rs
+++ b/examples/with_winit/src/stats.rs
@@ -384,11 +384,21 @@ pub fn draw_gpu_profiling(
             let text_size = (text_height * 0.9) as f32;
             // Text is specified by the baseline, but the y positions all refer to the top of the text
             cur_text_y = text_y + text_height;
-            let label = format!(
-                "{:.2?} - {:.30}",
-                Duration::from_secs_f64(this_time.max(0.0)),
-                profile.label
-            );
+            let label = {
+                if this_time < 0.0 {
+                    format!(
+                        "-{:.2?}(!!) - {:.30}",
+                        Duration::from_secs_f64(this_time.abs()),
+                        profile.label
+                    )
+                } else {
+                    format!(
+                        "{:.2?} - {:.30}",
+                        Duration::from_secs_f64(this_time),
+                        profile.label
+                    )
+                }
+            };
             scene.fill(
                 Fill::NonZero,
                 offset,
--- a/examples/with_winit/src/stats.rs
+++ b/examples/with_winit/src/stats.rs
@@ -385,6 +385,9 @@ pub fn draw_gpu_profiling(
             // Text is specified by the baseline, but the y positions all refer to the top of the text
             cur_text_y = text_y + text_height;
             let label = {
+                // Sometimes, the duration turns out to be negative
+                // We have not yet debugged this, but display the absolute value in that case
+                // see https://github.com/linebender/vello/pull/475 for more
                 if this_time < 0.0 {
                     format!(
                         "-{:.2?}(!!) - {:.30}",
--- a/examples/with_winit/src/stats.rs
+++ b/examples/with_winit/src/stats.rs
@@ -386,7 +386,7 @@ pub fn draw_gpu_profiling(
             cur_text_y = text_y + text_height;
             let label = format!(
                 "{:.2?} - {:.30}",
-                Duration::from_secs_f64(this_time),
+                Duration::from_secs_f64(this_time.max(0.0)),
                 profile.label
             );
             scene.fill(


examples/with_winit/src/stats.rs
```diff
@@ -386,7 +386,7 @@ pub fn draw_gpu_profiling(
             cur_text_y = text_y + text_height;
             let label = format!(
                 "{:.2?} - {:.30}",
-                Duration::from_secs_f64(this_time),
+                Duration::from_secs_f64(this_time.max(0.0)),
```
waywardmonkeys, line 389: I wonder if this is a bug in `wgpu` or something else where a monotonic timer isn't being used ...

DJMcNab:
Thanks for the PR - I just have one question, which is have you validated how negative the number is?

I ask because if it's very small ($\approx -\frac{1}{5000}$, i.e. -0.2ms), then that's probably fine, but if it's much larger than that, I suspect there's some data corruption going on. If that's the case, we should add a note to these text values of the actual amount, instead of just clamping it to zero.

I believe @DasLixou was seeing something similar [on the Zulip](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Help.20me.20helping.20you.20.28yep.2C.20just.20as.20silly.20as.20it.20sounds.29/near/423088899), where the timings from the profiler didn't seem right. Although in that case, it was AMD DirectX12 rather than macOS

Enter-tainer:
It is about -20ms 🤔 I guess it might be a problem?

My device is MacBook M2 Pro 32GB

image

image

Enter-tainer:
```
Negative time: -0.014637250453233719, bits: bf8dfa2280000000
Many same lines...
Negative time: -0.01713495794683695, bits: bf918bd390000000
...
Negative time: -0.01640879176557064, bits: bf90cd7760000000
...
```

DJMcNab:
I think this PR should show the negative value in the visible output (instead of zero). That is, by producing a Duration of the absolute value.
Maybe we should also log?

It's not clear to me why this is happening - @wumpf have you got any ideas?

DasLixou:
OK just read the whole conversation, if this is the same problem as I have, then i think its a bug between wgpu_profiler and the graphics card. Maybe we could do a crappy wgpu render example with wgpu_profiler and see if there's also the problem

Wumpf:
I believe this is somewhere between known & hard to avoid issue and problematic barrier placement in wgpu 🤔
I'm a little bit surprised it happens on dx12 & Vulkan since I've experienced this most commonly on Metal so far: The issue is that right now it's a bit too poorly defined what waits on what in the presence of timer scopes, so pairs of queries may internally be placed with different command buffer/encoder. This is also why the WebGPU spec recently removed timer queries for anything but the ones that are directly placed on compute/draw-pass boundaries.

DJMcNab:
Thanks - at least it's not too unexpected. Is there a wgpu issue we can track?
I'm slightly concerned that WebGPU is removing this - it is useful to have a view into what's taking the most time. 

The main use of the scopes is in `WgpuEngine::run_recording`, and a lot of the runtime/interaction with `wgpu_profiler` is in the `with_winit` crate

Wumpf:
These two are fitting best (given that it's really the issue I'm suspecting):

* https://github.com/gfx-rs/wgpu/issues/2521
* https://github.com/gpuweb/gpuweb/issues/3952

Wumpf:
Since gpu processing of things that aren't protected by barriers can be reordered in execution arbitrarily it is still very relevant. Right now things can happen where one query is part of a block of memory operations and another is part of a dispatch and then the driver/gpu decides to reorder one after the other (that sort of thing is what happens on wgpu's Metal implementation a lot because of the way it creates MTLCommandEncoder on the fly).

DJMcNab (approved):
Thanks. I've also added a comment which explains this

xStrom:
> I don't _think_ those explain the really odd behaviour we're seeing.
> 
> Like, I don't think it's reasonable for a dispatch to finish 20ms _before_ it started.
> 
> Those issues seem to be discussing how two timestamps can happen arbitrarily close to one another, even if the passes themselves are slow. But I don't think that should mean one timestamp happens before the other?

The Vulkan spec only provides [`vkCmdWriteTimestamp`](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkCmdWriteTimestamp) which has happens-after syncing. That means that even a malicious implementation that returns the same value for all calls to `vkCmdWriteTimestamp` is compliant. In less malicious cases you can still end up with a scenario described in the first post of [gpuweb#3952](https://github.com/gpuweb/gpuweb/issues/3952):

```js
writeTimestamp(A)
do_some_gpu_work(...)
writeTimestamp(B)

// Running via Vulkan, gets reordered on the GPU as:

do_some_gpu_work(...)
writeTimestamp(A) // Totally compliant with Vulkan spec
writeTimestamp(B)
```

So the quality of measurements will depend on how much the GPU manufacturer went beyond the spec in retaining execution order. Very unfortunate, especially as the reordering can be dynamic and depend on the measured instructions. So measuring a different execution time may just mean that the GPU optimizer resulted in a different execution order.

---

The issue on macOS is slightly different, because Metal doesn't provide even the already weak options that Vulkan has. Or at least the implementations haven't found those options yet. I didn't find a Metal specific `wgpu` issue but for Dawn there is [this issue](https://bugs.chromium.org/p/dawn/issues/detail?id=1598). Dawn has a custom solution for writing timestamps on Metal and that solution is completely broken in a way where different timestamp writes can be completely reorderd.

```js
writeTimestamp(A)
do_some_gpu_work(...)
writeTimestamp(B)

// Running via Metal, gets even worse:

do_some_gpu_work(...)
writeTimestamp(B)
writeTimestamp(A) // The joys of the current Dawn implementation
```

Even though this is a Dawn issue and I didn't find the matching `wgpu` issue, there is good reason to believe that `wgpu` suffers from the same problem. That's because I looked through the notes of two different WebGPU meetings ([1](https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2023-11-08#problem-measuring-time-deltas-with-writetimestamp-3952), [2](https://github.com/gpuweb/gpuweb/wiki/GPU-Web-2023-11-29#problem-measuring-time-deltas-with-writetimestamp-3952)) and the problem is discussed. There was no mention that it's solved in `wgpu`. Instead we get:

> Couldn't make writeTimestamp work as expected on Metal

> I would like to reintroduce writeTimestamp but not 100% sure how to do it.

> can do an investigation for writeTimestamp offline (for Metal) and come back to writeTimestamp later?

---

In summary, here's where we stand right now:
* `writeTimestamp` was removed from the WebGPU spec ([gpuweb#4370](https://github.com/gpuweb/gpuweb/pull/4370)).
* The already existing implementation on Vulkan may not work as developers expect, because the spec demands only happens-after sync.
* The already existing implementation on Metal is completely unreliable, negative deltas should be fully expected right now.
* Hopefully a better solution is found for Metal, maybe also for Vulkan & DX12. Some discussion of more reliable timers that have a perf hit are being considered. After more clarity is found, timers will be added back to the WebGPU spec.

DJMcNab:
Even taking that into account, which I agree definitely permits *some* negative durations, I still think there's some strangeness. I'm less sure that permits *this* negative duration.

I'm assuming two fairly weak invariants in making that assertion:
1) The timestamp can't be at a time before its command was sent to the GPU
2) The timestamp can't be at a time after its data was recieved from the GPU

The 20ms gap is long enough for a complete cycle to have been done, i.e. if everything is working ccorrectly, we're seeing one of those invariants be broken (also because we're blocking on the timestamp data being returned, because of using `Poll::Wait` in our `maintain` call)
Which means that either:
1) Metal's timestamp API is fundamentally meaningless, in that it just makes up any time it likes
2) We have some data corruption occuring

xStrom:
20ms does seem awfully long for M2 Pro, so indeed there's probably an additional factor.

I checked the [timer Metal code in `wgpu`](https://github.com/gfx-rs/wgpu/blob/badb3c88ea29acb159d333e2f60b1cc305bbd512/wgpu-hal/src/metal/command.rs#L27) and it does seem to have the same strategy as Dawn. It also has a bunch of comments describing various bugs with Metal.

So we know for a fact that both the `wgpu` timer implementation is buggy and that Metal has bugs in relation to timers. Wouldn't be that surprising if there is another bug in there causing the 20ms result.

Wumpf:
excellent writeup @xStrom !
And oh god that metal implementation was awful. There's a higher level issue in wgpu on top around we set up new CommandEncoder: it can very easily happen that the timestamp ends up on the "wrong one" which has nothing to do with the thing you're trying to measure, which is essentially "encouraging" metal to schedule it somewhere else entirely. The only thing that is somewhat reliable is [timestamp writes that are specified as part of a render/compute pass](https://docs.rs/wgpu/latest/wgpu/struct.RenderPassDescriptor.html#structfield.timestamp_writes)

PR #468, opened 2024-02-26
Update documentation

PoignardAzur:
Update README.md
Add ARCHITECTURE.md

Issue #467, opened 2024-02-26
`vello` should import `vello_shaders`


PoignardAzur:
The same preprocessor code is written twice:

- In `crates/shaders/src/compile/preprocess.rs`
- In `src/shaders/preprocess.rs`

[According to Raph](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/The.20preprocess.20code.20seems.20to.20be.20duplicated), this is because the vello_shaders crate was meant for Skia integration, and currently isn't used by Vello.

We should fix that and remove the duplication.

armansito:
We always intended for the main Vello crate to use `vello_shaders` but never got around to it. The crate supports generating native shaders (such as MSL) from the WGSL which allows external engines to target Metal, Vulkan, etc directly. This code path isn't exercised in the Vello repository but we have wanted to have native runners (i.e. alternative engines alongside wgpu) eventually but decided to put that in the backburner due to other priorities.

I think this issue can be repurposed as "Move to shader management via crates/shaders".

PR #478, opened 2024-03-01
Prepare package configurations for publishing.

xStrom:
I went over all the `Cargo.toml` files to get them ready for the upcoming publishing.

* Enabled `publish` for `vello` itself.
* Bumped `vello` version to `0.1.0`.
* Removed the workspace version. Every published crate gets its own dedicated version, which for now only means `vello` & `vello_encoding`.
* Added `rust-version` to document the reality of 1.75 being needed to compile `vello` at this moment.
* Excluded some needless files from being bundled when publishing, which is going to reduce the size of the published package by ~1MB !!
* Promoted some dependencies to workspace dependencies.
* Reordered some properties to get them into consistent order in all files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,13 @@ members = [
 ]
 
 [workspace.package]
+# Vello version, also used by other packages which want to mimic Vello's version.
+# Right now those packages include vello_encoding and vello_shaders.
+#
+# NOTE: When bumping this, remember to also bump the aforementioned other package's
+#       version in the dependencies section at the bottom of this file.
+version = "0.1.0"
+
 edition = "2021"
 rust-version = "1.75"
 license = "Apache-2.0 OR MIT"
@@ -23,7 +30,7 @@ repository = "https://github.com/linebender/vello"
 
 [package]
 name = "vello"
-version = "0.1.0"
+version.workspace = true
 description = "An experimental GPU compute-centric 2D renderer."
 categories = ["rendering", "graphics"]
 keywords = ["2d", "vector-graphics"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ members = [
 # Vello version, also used by other packages which want to mimic Vello's version.
 # Right now those packages include vello_encoding and vello_shaders.
 #
-# NOTE: When bumping this, remember to also bump the aforementioned other package's
+# NOTE: When bumping this, remember to also bump the aforementioned other packages'
 #       version in the dependencies section at the bottom of this file.
 version = "0.1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
 #
 # NOTE: When bumping this, remember to also bump the aforementioned other packages'
 #       version in the dependencies section at the bottom of this file.
+#       Additionally, bump the Vello dependency version in the 'simple' example.
 version = "0.1.0"
 
 edition = "2021"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,5 @@
 [workspace]
 resolver = "2"
-
 members = [
     "crates/encoding",
     "crates/shaders",
@@ -18,22 +17,20 @@ members = [
 
 [workspace.package]
 edition = "2021"
-version = "0.0.1"
-license = "MIT OR Apache-2.0"
+rust-version = "1.75"
+license = "Apache-2.0 OR MIT"
 repository = "https://github.com/linebender/vello"
 
 [package]
 name = "vello"
-description = "An experimental GPU compute-centric 2D renderer"
+version = "0.1.0"
+description = "An experimental GPU compute-centric 2D renderer."
 categories = ["rendering", "graphics"]
 keywords = ["2d", "vector-graphics"]
-
-# This crate is intended for publishing, but not ready yet
-publish = false
-
-version.workspace = true
-license.workspace = true
+exclude = ["/.github/", "/doc/", "/examples/", ".gitignore"]
 edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
 repository.workspace = true
 
 [features]
@@ -42,19 +39,22 @@ hot_reload = []
 buffer_labels = []
 
 [dependencies]
+vello_encoding = { workspace = true }
 bytemuck = { workspace = true }
 skrifa = { workspace = true }
 peniko = { workspace = true }
 wgpu = { workspace = true, optional = true }
-raw-window-handle = "0.6"
-futures-intrusive = "0.5.0"
-vello_encoding = { path = "crates/encoding" }
+raw-window-handle = { workspace = true }
+futures-intrusive = { workspace = true }
 wgpu-profiler = { workspace = true, optional = true }
 
 [workspace.dependencies]
+vello_encoding = { version = "0.1.0", path = "crates/encoding" }
 bytemuck = { version = "1.14.3", features = ["derive"] }
 skrifa = "0.15.5"
 peniko = "0.1.0"
+futures-intrusive = "0.5.0"
+raw-window-handle = "0.6"
 
 # NOTE: Make sure to keep this in sync with the version badge in README.md
 wgpu = { version = "0.19" }
--- a/crates/shaders/Cargo.toml
+++ b/crates/shaders/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "vello_shaders"
 version.workspace = true # We mimic Vello's version
-description = "Vello infrastructure to compile pipelines and shaders."
+description = "Vello infrastructure to preprocess and cross-compile shaders at compile time."
 edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
--- a/crates/shaders/Cargo.toml
+++ b/crates/shaders/Cargo.toml
@@ -1,9 +1,13 @@
 [package]
 name = "vello_shaders"
 version = "0.1.0"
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -18,22 +17,20 @@ members = [
 
 [workspace.package]
 edition = "2021"
-version = "0.0.1"
-license = "MIT OR Apache-2.0"
+rust-version = "1.75"
```
xorgy, line 20: In general we include a comment with the `rust-version` to remind people to also update it in `ci.yml`
Here's what the one out of Peniko looks like: 
```markdown
# If updating, ensure that it is also updated in .github/workflows/ci.yml, and
# update the documented reason
```

Right now, the `ci.yml` asks for 1.76 instead as well, so it might be worth bringing it down to 1.75 there if it is meant to build against 1.75.

xorgy (changes requested):
Looks good overall, but I would like that one added comment and clarification on the MSRV.

Cargo.toml
```diff
@@ -18,22 +17,20 @@ members = [
 
 [workspace.package]
 edition = "2021"
-version = "0.0.1"
```
DJMcNab, line 21: I'd be inclined to keep using this - it's very annoying managing multiple version numbers.

That is, once/if we're publishing `vello_shaders` because we've done #467, we'd want it have the same version number as `vello` itself.
I don't think we're likely to have a breaking release of Vello which isn't also breaking the encoding

crates/tests/Cargo.toml
```diff
@@ -1,13 +1,11 @@
 [package]
 name = "vello_tests"
+description = "Helper code for writing Vello tests."
 edition.workspace = true
-version.workspace = true
```
DJMcNab, line 4: I didn't know we could have a package without a version - that's neat.

crates/tests/Cargo.toml
```diff
@@ -1,13 +1,11 @@
 [package]
 name = "vello_tests"
+description = "Helper code for writing Vello tests."
```
DJMcNab, line 3: I'm not sure this is the right description here

Like yes, the lib crate only includes the helper code, but that's not the raison d'être for this package.

examples/headless/Cargo.toml
```diff
@@ -1,23 +1,19 @@
 [package]
 name = "headless"
-description = "An example showing how to use `vello` to create raster images"
-publish = false
-
-version.workspace = true
-license.workspace = true
+description = "An example showing how to use `vello` to create raster images."
```
DJMcNab, line 3: I *think* this should use Vello (with a capital letter) outside of graves. I'm not even sure that the crate description is markdown rendered.

DJMcNab (approved):
Generally looks reasonable

Cargo.toml
xStrom: There is no MSRV job in the Vello CI and Vello doesn't have a MSRV policy yet. I kept this really simple on purpose so that it's specified for Cargo instead of undefined, because the release is imminent. I tested it locally to work with 1.75. Even if it wouldn't actually work with 1.75 (let's say a dep bumped their MSRV just now) then the experience wouldn't differ from it being undefined, as in both cases the compilation would fail.

A more through pass on the documentation and CI would come with [RFC#5](https://github.com/linebender/rfcs/pull/5), which won't be ready in time for the planned release.

PS. Can't bring CI down to 1.75 because bevy requires 1.76. However that doesn't affect our to-be-published crates.

crates/tests/Cargo.toml
xStrom: Yeah it's a new feature since 1.75. Non-published crates don't benefit from a version so now Cargo supports reducing that noise.

crates/tests/Cargo.toml
xStrom: I just copied this from #468's `ARCHITECTURE.md`. Also we're not publishing this crate, so it doesn't matter that much anyway.

That said, just tell me a better description and I'm all for it.

examples/headless/Cargo.toml
xStrom: Good call!

crates/tests/Cargo.toml
DJMcNab: We should bring this to all the applicable crates. I think `with_winit` still has a version, for example.

crates/tests/Cargo.toml
DJMcNab: I'd probably just have "Integration tests for Vello"

Cargo.toml
xStrom: If we have these smaller packages that are meant to be used by others than Vello, then publishing a breaking change for them when it isn't actually required would be creating a needless hurdle for those other consumers. Even if that isn't a common scenario.

There is also no sense in publishing new versions without any actual changes. So take for example a scenario where `vello` needs a small bug fix and we publish `0.1.1`. Well the encoding/shaders didn't actually receive any changes.

The opposite case is also relevant. Say we make two small patch releases for shaders, going from `0.1.0` to `0.1.2`. However `vello` proper might not need a new release at all.

There is also no way to specify dependencies with the workspace version. That is, under the `[dependencies]` table we still need to manually enter a version requirement for encoding and shaders. This isn't required for unpublished packages, but we're moving into the published state now.

Given all of the above, I consider the extra annoyance of having to bump the encoding/shader package's version field worth the extra flexibility and precision we get.

crates/tests/Cargo.toml
DJMcNab: Oh, I see your commit

Cargo.toml
DJMcNab: > The opposite case is also relevant. Say we make two small patch releases for shaders, going from 0.1.0 to 0.1.2. However vello proper might not need a new release at all.

We probably want to be able to have a release of Vello which definitely has these changes, if we have improved the shaders. I don't want to be recommending that people depend on `vello_shaders = 0.1.2` to guarantee their users actually get the fix the need. That is, `vello_shaders` being updated is sufficient reason for `vello` to be released.

I don't think slightly optimising for the experience of the (very few) consumers who are going to be using `vello_shaders` directly, without using `vello`, at the expense of making things slightly more confusing for those many more who will be using `vello` directly is a good tradeoff. And I think it makes our processes more complicated, to maintain version numbers for each of the independent crates.

And if you're using `vello_encoding`, you want to be using the matching version of `vello` anyway. Having to match those up would make having the dependency much harder.

crates/tests/Cargo.toml
xStrom: I agree and `with_winit` originally had this change too. However turns out that `cargo-apk` still has the old requirement, perhaps because it depends on some older fork of Cargo? I didn't investigate deeply and for now just [added the version back to `with_winit`](https://github.com/linebender/vello/pull/478/commits/88b8ac19afc5dd1f0c861411e7b41ba94fb5847c).

Hopefully `cargo-apk` will get the update, although their warning
> This tool is deprecated in favour of [xbuild](https://github.com/rust-mobile/xbuild).

doesn't exactly give me high confidence.

Maybe we can get them updated, maybe we move to xbuild (?), maybe we need to add version back to even some other examples just for the sake of Android.

Cargo.toml
xStrom: I'm not fully sold, but this topic can be postponed until we're actually doing more of these releases. So for now, in the interest of getting other stuff here merged, I'll just revert the to-be-published packages to a unified version with some comments.

The dependencies will still need manual updates, which I'll also add a comment for.

crates/shaders/Cargo.toml
```diff
@@ -1,10 +1,14 @@
 [package]
 name = "vello_shaders"
-version = "0.1.0"
+version.workspace = true # We mimic Vello's version
+description = "Vello infrastructure to compile pipelines and shaders."
```
DJMcNab, line 4: I don't think vello_shaders handles pipeline compilation. That's handled in whatever integration uses `vello_shaders`.

That is,
```suggestion
description = "Vello infrastructure to preprocess and cross-compile shaders at compile time"
```

Cargo.toml
```diff
@@ -17,23 +16,28 @@ members = [
 ]
 
 [workspace.package]
+# Vello version, also used by other packages which want to mimic Vello's version.
+# Right now those packages include vello_encoding and vello_shaders.
+#
+# NOTE: When bumping this, remember to also bump the aforementioned other packages'
+#       version in the dependencies section at the bottom of this file.
```
DJMcNab, line 23: We should also mention the `simple` example's `Cargo.toml`, which should point to the released version in the comment

DJMcNab (approved):
Thanks for this - these are useful improvements

crates/shaders/Cargo.toml
xStrom: This is another one I got from `ARCHITECTURE.md`. I updated the changes here, and will also submit a comment on the docs PR so that we're more precise there too.

Cargo.toml
xStrom: Good call, I tweaked the simple example and added the comment to the workspace `Cargo.toml` as well.

xorgy (approved):
LGTM

crates/tests/Cargo.toml
xStrom: I looked a bit deeper into this. The issue comes directly from `cargo-apk`'s custom `Cargo.toml` parser. I submitted a patch to them ([cargo-apk#56](https://github.com/rust-mobile/cargo-apk/pull/56)), let's see if and when they accept it.

PR #482, opened 2024-03-01
Bug fix: `with_winit` 0-sized canvas on wasm

simbleau:
Closes #481 

image

PR #483, opened 2024-03-01
feat: GitHub Pages web demo

simbleau:
This will likely need attention from @raphlinus or someone who has administrative privileges on the repo to setup GitHub Pages.

This PR is aimed to package the `with_winit` demo to WASM and serve it with GitHub Pages!
(As soon as #482 is closed - I put a PR for that)

Tasks:
- [x] Wait for WASM to work - #482 
- [x] Enable environments in Admin settings - Needs @raphlinus or @DJMcNab 
- [x] Approve and Merge this PR
- [ ] Update the repo link to https://linebender.github.io/vello

The end result will look like this, except be available on [https://linebender.github.io/vello](https://linebender.github.io/vello)

image

### DIFF
--- a/.github/workflows/web-demo.yml
+++ b/.github/workflows/web-demo.yml
@@ -1,0 +1,66 @@
+env:
+  # We aim to always test with the latest stable Rust toolchain, however we pin to a specific
+  # version like 1.70. Note that we only specify MAJOR.MINOR and not PATCH so that bugfixes still
+  # come automatically. If the version specified here is no longer the latest stable version,
+  # then please feel free to submit a PR that adjusts it along with the potential clippy fixes.
+  RUST_STABLE_VER: "1.76" # In quotes because otherwise (e.g.) 1.70 would be interpreted as 1.7
+
+name: Web Demo Release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  rustfmt:
+    runs-on: ubuntu-latest
+    name: cargo fmt
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: install stable toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_STABLE_VER }}
+          targets: wasm32-unknown-unknown
+
+      - name: install wasm-bindgen
+        uses: jetli/wasm-bindgen-action@v0.2.0
+        with:
+          version: 'latest'
+
+      - name: build (wasm)
+        run: cargo build -p with_winit --bin with_winit_bin --release --target wasm32-unknown-unknown
+        env:
+          RUSTFLAGS: '--cfg=web_sys_unstable_apis'
+
+      - name: package wasm
+        run: |
+          mkdir public
+          wasm-bindgen --target web --out-dir public target/wasm32-unknown-unknown/release/with_winit_bin.wasm --no-typescript
+          cat << EOF > public/index.html
+            <html>
+              <title>Vello Web Demo</title>
+              <meta content=no-cache http-equiv=Cache-control>
+              <meta content=-1 http-equiv=Expires>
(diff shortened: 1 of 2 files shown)


.github/workflows/web-demo.yml
```diff
@@ -0,0 +1,66 @@
+env:
+  # We aim to always test with the latest stable Rust toolchain, however we pin to a specific
+  # version like 1.70. Note that we only specify MAJOR.MINOR and not PATCH so that bugfixes still
+  # come automatically. If the version specified here is no longer the latest stable version,
+  # then please feel free to submit a PR that adjusts it along with the potential clippy fixes.
+  RUST_STABLE_VER: "1.76" # In quotes because otherwise (e.g.) 1.70 would be interpreted as 1.7
```
DJMcNab, line 6: We're not using clippy here, so there's no risk of the workflow breaking between runs. I'd suggest just using `stable`

.github/workflows/web-demo.yml
```diff
@@ -0,0 +1,66 @@
...
+
+name: Web Demo Release
```
DJMcNab, line 8:
```suggestion
name: Web Demo Update
```

.github/workflows/web-demo.yml
```diff
@@ -0,0 +1,66 @@
...
+                    margin: 0;
+                    padding: 0;
+                  }
+                </style>
+              </body>
+            </html>
+          EOF
+
+      - name: deploy
+        uses: peaceiris/actions-gh-pages@v3
```
DJMcNab, line 63: Why is this not using https://github.com/marketplace/actions/deploy-github-pages-site? (i.e. the official one from GitHub). We at least need a comment here explaining that choice

README.md
```diff
@@ -48,6 +48,8 @@ Examples must be selected using the `--package` (or `-p`) Cargo flag.
 
 ### Winit
 
+This example has a demo [available here](https://linebender.github.io/vello) on supporting web browsers.
```
DJMcNab, line 51: I'd put this in the "Web" section, to leave it in context that we don't actively support the web, and that it only works on the browsers which support WebGPU

DJMcNab (commented):
Thanks!

I think this is probably good to have, but there are a few things I'd prefer to be swapped around slightly

.github/workflows/web-demo.yml
simbleau: Added a comment TODO to revisit when stable. While `actions/deploy-pages` is stable, they require you to use [`actions/upload-pages-artifact`](https://github.com/actions/upload-pages-artifact) which is not stable.

simbleau:
@DJMcNab ready again. :)

.github/workflows/web-demo.yml
DJMcNab: I think I'd still much rather us use the official GitHub one - creating a gh-pages branch is messy, and requires additional tokens to be managed.

Like yes, there's risk that things break, but needing to do secrets securely is something I don't want to deal with at the moment

simbleau: Do you mean `secrets.GITHUB_TOKEN`? This is a token provided to all jobs by GitHub, not something needing to be added or managed by us.

I can still work on the official one, though. I'll have it up soon.

DJMcNab: Ah, thanks for the pointer. I'd never dug into it before. My prior experience was with talking to external systems, which of course do need tokens managing.

Even still, I do think it's nicer to not make a gh-pages branch, so thanks for changing that

DJMcNab (approved):
I'm happy to experiment with this - I can't run the example, but I think we added an error page at some point?

simbleau:
> I'm happy to experiment with this - I can't run the example, but I think we added an error page at some point?

The pipeline has to run first. I just kicked it off. Hope it works first try 🤞

PR #484, opened 2024-03-01
doc: Fix rustdoc warning about bare URL.

waywardmonkeys:


DJMcNab (approved):
This is what I was trying to express in https://github.com/linebender/glazier/pull/149#discussion_r1489301778

I don't see any feasible solution other than maybe reporting these through a webhook on Zulip? That is kinda terrible UX though
Because these are nightly docs, we don't want it to break/block randomly for contributors

waywardmonkeys:
> This is what I was trying to express in [linebender/glazier#149 (comment)](https://github.com/linebender/glazier/pull/149#discussion_r1489301778)
> 
> I don't see any feasible solution other than maybe reporting these through a webhook on Zulip? That is kinda terrible UX though Because these are nightly docs, we don't want it to break/block randomly for contributors

I understand and share your frustration! I don't have an answer at the moment or I would have included it in this PR. :(

DJMcNab:
Yeah, no worries. I'm happy to get this landed though. I'll let you press the button

xStrom (approved):

xStrom:
I'll add it to the queue just to make sure it won't miss the release train.

waywardmonkeys:
@xStrom Thanks. I had a bunch of meetings and stuff and got lost.

PR #486, opened 2024-03-01
Toggle Profiling

DJMcNab:
Allow it to be toggled by the <kbd>G</kbd> key, and controlled by a command line flag

Also properly maintain the vsync state - even across suspend/resumes between screens - and enable it to be toggled with a command line flag


armansito (approved):
Looks good!

PR #489, opened 2024-03-02
Support image, very basic clip-path for vello_svg and various improvements

Enter-tainer:
- Bump usvg to v0.40
- Support basic clip-path(`clip-path` with exactly 1 `path`)
- Support embeded image
- Take path/image visibility into consideration
- Support stroke/fill paint order

Due to [upstream bug](https://github.com/RazrFalcon/resvg/issues/722), `abs_transform` is not accurate for `use` with positions in v0.40. So this pr calculate `transform` manually.

TODO:

- [x] try recover original error processing logic
- [x] properly handle error in image related code path
- [x] correct image viewbox & clipping

### DIFF
--- a/integrations/vello_svg/src/geom.rs
+++ b/integrations/vello_svg/src/geom.rs
@@ -1,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// copied from https://github.com/RazrFalcon/resvg/blob/4d27b8c3be3ecfff3256c9be1b8362eb22533659/crates/resvg/src/geom.rs
+
+/// Converts `viewBox` to `Transform` with an optional clip rectangle.
+///
+/// Unlike `view_box_to_transform`, returns an optional clip rectangle
+/// that should be applied before rendering the image.
+pub fn view_box_to_transform_with_clip(
+    view_box: &usvg::ViewBox,
+    img_size: usvg::tiny_skia_path::IntSize,
+) -> (usvg::Transform, Option<usvg::NonZeroRect>) {
+    let r = view_box.rect;
+
+    let new_size = fit_view_box(img_size.to_size(), view_box);
+
+    let (tx, ty, clip) = if view_box.aspect.slice {
+        let (dx, dy) = usvg::utils::aligned_pos(
+            view_box.aspect.align,
+            0.0,
+            0.0,
+            new_size.width() - r.width(),
+            new_size.height() - r.height(),
+        );
+
+        (r.x() - dx, r.y() - dy, Some(r))
+    } else {
+        let (dx, dy) = usvg::utils::aligned_pos(
+            view_box.aspect.align,
+            r.x(),
+            r.y(),
+            r.width() - new_size.width(),
+            r.height() - new_size.height(),
+        );
+
+        (dx, dy, None)
+    };
+
+    let sx = new_size.width() / img_size.width() as f32;
+    let sy = new_size.height() / img_size.height() as f32;
+    let ts = usvg::Transform::from_row(sx, 0.0, 0.0, sy, tx, ty);
+
+    (ts, clip)
+}
+
+/// Fits size into a viewbox.
+pub fn fit_view_box(size: usvg::Size, vb: &usvg::ViewBox) -> usvg::Size {
+    let s = vb.rect.size();
+
+    if vb.aspect.align == usvg::Align::None {
+        s
+    } else if vb.aspect.slice {
+        size.expand_to(s)
+    } else {
+        size.scale_to(s)
+    }
+}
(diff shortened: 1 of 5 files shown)


integrations/vello_svg/src/geom.rs
```diff
@@ -0,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// copied from https://github.com/RazrFalcon/resvg/blob/4d27b8c3be3ecfff3256c9be1b8362eb22533659/crates/resvg/src/geom.rs
```
Enter-tainer, line 5: I don't really know the copyright/license problem here. Is it ok to use this code? And ci looks not happy with this.

integrations/vello_svg/src/geom.rs
```diff
@@ -0,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
xStrom, line 3: The SPDX header type is explicitly mentioned as allowed in the [MPL FAQ under Q27](https://www.mozilla.org/en-US/MPL/2.0/FAQ/#machine-readable-header) and also [under Q4](https://www.mozilla.org/en-US/MPL/2.0/FAQ/#apply). Also [MPL section 3.4](https://www.mozilla.org/en-US/MPL/2.0/#notices) states that *You may not remove or alter the **substance** of any license notices*.

Thus the MPL header can be replaced with a header that matches our style while keeping the same substance, as the SPDX format is explicitly allowed and conveys the same meaning.





```suggestion
// Copyright 2018 Yevhenii Reizner
// SPDX-License-Identifier: MPL-2.0
```

integrations/vello_svg/src/geom.rs
xStrom: Yes it is ok. The MPL 2.0 license is compatible enough with Apache 2.0 OR MIT, and we already import `usvg` which is licensed under MPL 2.0 so this has no effect on the license mixture of `vello_svg`.

We can replace the copyright notice header of this file with a different format, which I showed in a sibling comment.

You will also need to edit `.github/copyright.sh` and look for the first long regex line that starts with `output=`. Replace that line with the following:

```sh
output=$(rg "^// Copyright (19|20)[\d]{2} (.+ and )?the Vello Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2\.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!{shader,src/cpu_shader}" -g "!integrations/vello_svg/src/geom.rs" .)
```

Enter-tainer:
I wonder if we have test infra for rendering? I guess it would be helpful to difftest with mature svg renderers like resvg

simbleau:
> I wonder if we have test infra for rendering? I guess it would be helpful to difftest with mature svg renderers like resvg

We do not. We are not focused on correctness, unlike resvg. This integration is more of an example than an actual SVG renderer.

DJMcNab:
We do have [`crates/tests`](https://github.com/linebender/vello/tree/main/crates/tests), which is some basic test infrastructure.
It would be very cool to test against different renderers, although I do not want us to *commit* any test SVGs, because of repository size concerns

xStrom:
Just for clarification, is the `geom.rs` file introduced only because of the `usvg` bug? Which is to ask, if `usvg` gets fixed do we no longer need this file?

Enter-tainer:
> Just for clarification, is the `geom.rs` file introduced only because of the `usvg` bug? Which is to ask, if `usvg` gets fixed do we no longer need this file?

i'm reworking related code and i hope i will remove it once this pr becomes ready

Enter-tainer:
> Just for clarification, is the geom.rs file introduced only because of the usvg bug? Which is to ask, if usvg gets fixed do we no longer need this file?

It's related to usvg's bug. It's a util file from resvg, and the it is not in public API so I have to copy it here.

It is used to calculated the size and transform of the image tags in a svg tree. When I send the last comment I want to reimplement it but eventually found that simply reusing resvg's util is much easier.

DJMcNab (changes requested):
The code looks reasonable, but I'm just noting here that this PR is blocked on deciding what to do about the MPL code. See [Zulip](https://xi.zulipchat.com/#narrow/stream/419691-linebender/topic/Vello.20SVG/near/424893347)

I expect that we should discuss this in tomorrow's office hours

simbleau:
Update on decision: This is blocked because I will be bringing in this code into a new project that will be made public soon. (linebender/vello_svg)

Going forward, it was decided that vello will have a minimal story around SVG (just enough to run examples like Ghostscript_Tiger, Paris-30k, etc.)

I will close this PR once the new repo is seeded.

Enter-tainer:
closing this pr since https://github.com/linebender/vello_svg is ready

PR #491, opened 2024-03-03
Expand 'NDC'

waywardmonkeys:


xStrom (approved):
Seems useful, I certainly hadn't heard of NDC before.

PR #492, opened 2024-03-03
ci: Add typos check for common spelling errors.

waywardmonkeys:
This will fail until #490 and #491 land.

I'm not thrilled about the discoverability here if a CI check fails due to this. We don't really have a place for docs about this yet (even a `CONTRIBUTING.md`) ...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,3 +290,12 @@ jobs:
       # We test documentation using nightly to match docs.rs. This prevents potential breakages
       - name: cargo doc
         run: cargo doc --workspace --locked --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples
+
+  # If this fails, consider changing your text or adding something to `.typos.toml`
+  typos:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: check typos
+        uses: crate-ci/typos@v1.21.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -291,7 +291,7 @@ jobs:
       - name: cargo doc
         run: cargo doc --workspace --locked --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples
 
-  # If this fails, consider changing your text or adding something to `.typos.toml`
+  # If this fails, consider changing your text or adding something to .typos.toml
   typos:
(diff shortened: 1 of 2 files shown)


xStrom:
How would we deal with false positives in the future, is there a way to whitelist words/files?

It falsely identifies NDC as being an incorrect AND. We're going to unpack NDC for other reasons, but it wasn't really a typo and I can see similar scenarios in the future where we don't want to unpack the acronym.

waywardmonkeys:
Yes, it is configurable: https://crates.io/crates/typos-cli

And that's where having some contribution docs / guidelines might be useful ...

xStrom:
Looks like dealing with false positives isn't too difficult.

As for docs, well I think it's completely fine to just create a new `CONTRIBUTING.md` with a basic section about this typo checker and a link to its crate, repo, or docs. This section should probably make it clear that if there is a false positive we encourage the configuration change to suppress it.

If we end up wanting a different `CONTRIBUTING.md` we can do that work later in a different PR.

Also worth adding a short comment next to the CI step mentioning that there is more info in `CONTRIBUTING.md`. Actually, might even be worth making it an extra step, something like this:
```yml
- name: help with typos
  run: echo "If any of the detected typos are false positives, see CONTRIBUTING.md for guidance."
  if: failure()
```

.github/workflows/ci.yml
```diff
@@ -290,3 +290,12 @@ jobs:
       # We test documentation using nightly to match docs.rs. This prevents potential breakages
       - name: cargo doc
         run: cargo doc --workspace --locked --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples
+
+  # If this fails, consider changing your text or adding something to `.typos.toml`
```
xStrom, line 294: These comments aren't markdown, so I'd avoid the ticks.

```suggestion
  # If this fails, consider changing your text or adding something to .typos.toml
```

xStrom (approved):
Thanks, let's try it out. 🚀

.typos.toml
```diff
@@ -0,0 +1,8 @@
+# Corrections take the form of a key/value pair. The key is the incorrect word
```
xStrom, line 1: Probably worth linking to the reference here.

```suggestion
# See configuration reference https://github.com/crate-ci/typos/blob/master/docs/reference.md

# Corrections take the form of a key/value pair. The key is the incorrect word
```

PR #496, opened 2024-03-04
Checkpoint of Euler spiral based stroke expansion

raphlinus:
This branch contains an implementation of stroke expansion based on Euler spirals. At this point, the CPU shader implementation (in `cpu-shader/flatten.rs`) should be a reasonably solid implementation, with care taken for numerical robustness in particular, but more validation is needed.

In particular, some of the "tricky stroke" cases have a semi-circle taken out near the endpoints. This is the expected rendering from slightly perturbing the co-linear cubic Bézier, then drawing just the parallel curve and cap. However, it is not strongly correct rendering (in the sense of the Nehab 2020 "Converting stroked primitives to filled primitives" paper sense).

The next major step is translation of that logic to WGSL so strokes can be fully evaluated on GPU.

Progress toward #303


src/cpu_shader/euler.rs
```diff
@@ -0,0 +1,277 @@
...
+//! Utility functions for Euler Spiral based stroke expansion.
+
+// Use the same constants as the f64 version.
+#![allow(clippy::excessive_precision)]
+
+use super::util::Vec2;
+use std::f32::consts::FRAC_PI_4;
+
+#[derive(Debug)]
+pub struct CubicParams {
```
armansito, line 13: I suggest documenting this parameterization for a cubic and briefly explain how it relates to the Euler spiral, in a few words.

src/cpu_shader/euler.rs
```diff
@@ -0,0 +1,277 @@
...
+    pub p1: Vec2,
+    pub params: EulerParams,
+}
+
+impl CubicParams {
+    /// Compute parameters from endpoints and derivatives.
+    pub fn from_points_derivs(p0: Vec2, p1: Vec2, q0: Vec2, q1: Vec2, dt: f32) -> Self {
+        let chord = p1 - p0;
+        // Robustness note: we must protect this function from being called when the
+        // chord length is (near-)zero.
```
armansito, line 41: Maybe the CPU shaders should assert that the chord length is less than some epsilon (I'm not sure if Rust panics on a floating point division by zero).

src/cpu_shader/euler.rs
```diff
@@ -0,0 +1,277 @@
...
+    // by chord.
+    pub fn est_euler_err(&self) -> f32 {
+        // Potential optimization: work with unit vector rather than angle
+        let cth0 = self.th0.cos();
+        let cth1 = self.th1.cos();
+        if cth0 * cth1 < 0.0 {
+            // Rationale: this happens when fitting a cusp or near-cusp with
+            // a near 180 degree u-turn. The actual ES is bounded in that case.
+            // Further subdivision won't reduce the angles if actually a cusp.
+            return 2.0;
```
armansito, line 72: As we discussed briefly, I would elaborate on the choice of `2.0` here: what is the expected downstream effect and why is that acceptable.

src/cpu_shader/euler.rs
```diff
@@ -0,0 +1,277 @@
...
+        }
+        let e0 = (2. / 3.) / (1.0 + cth0);
+        let e1 = (2. / 3.) / (1.0 + cth1);
```
armansito, line 75: Is there are scenario in which `cth0` and `cth1` are both negative (so they pass the check above) and at least one of them has a value in the vicinity of -1?

src/cpu_shader/euler.rs
```diff
@@ -0,0 +1,277 @@
...
+            self.p0.x + chord.x * x - chord.y * y,
+            self.p0.y + chord.x * y + chord.y * x,
+        )
+    }
+}
+
+/// Integrate Euler spiral.
+///
+/// TODO: investigate needed accuracy. We might be able to get away
+/// with 8th order.
```
armansito, line 197: Eventually we should have a wiki page (as well as the definitive paper) that explain the formulations here. Maybe add a disclaimer for now that a lot of the magic numbers here were derived by hand and needs to be documented?

src/cpu_shader/flatten.rs
```diff
@@ -202,17 +149,68 @@ fn output_two_lines_with_transform(
...
+    center: Vec2,
+    angle: f32,
+    transform: &Transform,
+    line_ix: &mut usize,
+    lines: &mut [LineSoup],
+    bbox: &mut IntBbox,
+) {
+    let mut p0 = transform.apply(begin);
+    let mut r = begin - center;
+    let tol: f32 = 0.5;
```
armansito, line 165: Is `0.5` the tolerance value we want to go with?

src/cpu_shader/flatten.rs
```diff
@@ -202,17 +149,68 @@ fn output_two_lines_with_transform(
...
+    const MAX_LINES: u32 = 1000;
+    let n_lines = if theta <= ROBUST_EPSILON {
+        MAX_LINES
+    } else {
+        MAX_LINES.min((std::f32::consts::TAU / theta).ceil() as u32)
+    };
+
+    let th = angle / (n_lines as f32);
+    let c = th.cos();
+    let s = th.sin();
```
armansito, line 178: nit: This could be rewritten as `let (s, c) = th.sin_cos();`

src/cpu_shader/flatten.rs
```diff
@@ -202,17 +149,68 @@ fn output_two_lines_with_transform(
...
     path_ix: u32,
     local_to_device: &Transform,
     offset: f32,
+    start_p: Vec2,
+    end_p: Vec2,
     line_ix: &mut usize,
     lines: &mut [LineSoup],
     bbox: &mut IntBbox,
 ) {
+    log!("@@@ flatten_euler: {:#?}", cubic);
```
armansito, line 212: nit: comment out or remove debug log?

src/cpu_shader/flatten.rs
```diff
@@ -202,17 +149,68 @@ fn output_two_lines_with_transform(
...
+    let tol: f32 = 0.5;
+    let radius = tol.max((p0 - transform.apply(center)).length());
+    let x = 1. - tol / radius;
+    let theta = (2. * x * x - 1.).clamp(-1., 1.).acos();
+    const MAX_LINES: u32 = 1000;
+    let n_lines = if theta <= ROBUST_EPSILON {
+        MAX_LINES
+    } else {
+        MAX_LINES.min((std::f32::consts::TAU / theta).ceil() as u32)
+    };
```
armansito, line 174: This formulation is different from what's in the WGSL (and the removed lines below):

```
let MIN_THETA = 0.0001;
let tol = 0.1;
let radius = max(tol, length(p0 - transform_apply(transform, center)));
let theta = max(MIN_THETA, 2. * acos(1. - tol / radius));

// Always output at least one line so that we always draw the chord.
let n_lines = max(1u, u32(ceil(angle / theta)));
```

I believe the WGSL version is correct (apart from the tolerance and MIN_THETA values, which need tuning). I would keep the CPU and WGSL versions consistent.

shader/flatten.wgsl
```diff
@@ -188,96 +357,118 @@ fn flatten_cubic(cubic: CubicPoints, path_ix: u32, local_to_device: Transform, o
...
-        let qp2 = eval_cubic(p0, p1, p2, p3, t);
-        var qp1 = eval_cubic(p0, p1, p2, p3, t - 0.5 * step);
-        qp1 = 2.0 * qp1 - 0.5 * (qp0 + qp2);
-
-        // TODO: Estimate an accurate subdivision count for strokes
-        let params = estimate_subdiv(qp0, qp1, qp2, scaled_sqrt_tol);
-        keep_params[i] = params;
-        val += params.val;
-        qp0 = qp2;
+    // Drop zero length lines.
```
armansito, line 360: Maybe mention here why the exact matches are valid (e.g. any invariants about prior detection of zero-length lines).

shader/flatten.wgsl
```diff
@@ -188,96 +357,118 @@ fn flatten_cubic(cubic: CubicPoints, path_ix: u32, local_to_device: Transform, o
...
+        last_q = eval_cubic_and_deriv(p0, p1, p2, p3, DERIV_EPS).deriv;
+    }
+    var last_t = 0.0;
+    var lp0 = t_start;
+    loop {
+        let t0 = f32(t0_u) * dt;
+        if t0 == 1.0 {
+            break;
+        }
+        loop {
```
armansito, line 382: Should there be a short explainer here about the derivative threshold, the iterative search to handle cusps, and the 3 cases below?

src/cpu_shader/flatten.rs
```diff
@@ -1,94 +1,51 @@
...
-fn approx_parabola_inv_integral(x: f32) -> f32 {
-    x * (1.0 - B + (B * B + 0.5 * x * x)).sqrt()
-}
-
-#[derive(Clone, Copy, Default)]
-struct SubdivResult {
-    val: f32,
-    a0: f32,
-    a2: f32,
+// TODO: remove this
```
armansito, line 17: Remove uses of `log!`? Or should we leave them in as a no-op?

P.S.: IWBN to have a debug macro that's optionally enabled for this purpose in general.

src/scene.rs
```diff
@@ -116,7 +116,7 @@ impl Scene {
         const SHAPE_TOLERANCE: f64 = 0.01;
         const STROKE_TOLERANCE: f64 = SHAPE_TOLERANCE;
 
-        const GPU_STROKES: bool = false; // Set this to `true` to enable GPU-side stroking
+        const GPU_STROKES: bool = true; // Set this to `true` to enable GPU-side stroking
```
armansito, line 119: 🎉

armansito (approved):
Great progress and exciting milestone! I left a number of minor suggestions, largely around documentation. This will get better as we work through the writeup soon but I think some things could be explained briefly in the code now.

src/cpu_shader/euler.rs
raphlinus: It does not. I'll add an assert and some comments, that should help us debug the case.

And fun! When I added the assertion it got triggered. I'll add the robustness logic for this now.

src/cpu_shader/euler.rs
raphlinus: I wrote a paragraph explaining the tradeoff. It's also something that can potentially be explained in the paper.

src/cpu_shader/flatten.rs
raphlinus: That's a very good question and we should take a closer look. One other thing I've noticed in testing is the second longpathdash now exceeds memory limits, and we should really understand why.

src/cpu_shader/flatten.rs
raphlinus: I'm ok leaving them in for the time being, I found it useful while developing to be able to turn it on. I agree that a more sophisticated log story would be nice.

shader/flatten.wgsl
raphlinus: That feels more like paper than code, but I'll add a sentence each.

src/cpu_shader/flatten.rs
raphlinus: I hadn't intended to change this. I'm pretty sure that what happened is that #415 didn't make it into the euler-spiral-strokes branch. I've restored it, and good catch.

raphlinus:
Ok, I think I've addressed the review comments, but I may have missed one.

Remaining issues:

* investigate why longpathdash is taking more memory
* revisit tolerance tuning parameters

These two items may not be completely separate :)

shader/flatten.wgsl
armansito: I agree. I think the paper will be the in-depth reference for all of this BUT I would like the code to eventually contain explainers and relevant citations as much as possible. I think the comments can be improved but I'm OK with the current state of things.

src/cpu_shader/euler.rs
armansito: Consider adding assertions here that `cth0` and `cth1` won't cause divisions by zero.

src/cpu_shader/euler.rs
armansito: Maybe add a short sentence about the effect of `2.0` vs another number like `1.0`, e.g. "Returning an error value of 2.0 terminates the subdivisions because so and so".

src/cpu_shader/euler.rs
raphlinus: Ok, so when I add those assertions they do trigger on tricky_strokes, specifically 'M0 0.5 C 0 0 20 0 10 0`. I'm going to max with 1e-9 here (and in wgsl) to protect against divide by zero, as the correct behavior is to subdivide.

PR #497, opened 2024-03-04
fix(wasm): GpuProfilerSettings not initialized on startup

### LINKED ISSUE #494: Web Demo hangs after a minute
After about a minute or two, the web demo (http://linebender.org/vello/) hangs.

Zulip ref: https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Vello.3A.20wasm.20web.20demo

Seemingly because the GpuProfiler?

Stack trace:
Stack

simbleau:
Closes #494


DJMcNab (approved):
Thanks

I think it would be worth tracking down the actual issue as well, but I don't think it's that useful to track it in our issue tracker, as, whilst we support the web, we don't necessarily support the profiling

PR #499, opened 2024-03-05
Fix the formatting of the `with_winit` example

DJMcNab:
Rustfmt gave up previously because of the `store_profiling`'s nested depth


Philipp-M (approved):
LGTM

PR #500, opened 2024-03-05
Disable the async pipeline by default

### LINKED ISSUE #494: Web Demo hangs after a minute

DJMcNab:
The async pipeline is not wired up to do robust readback, so it has few advantages (see #366)

This PR also enables the non-async pipeline's interaction with the GPU profiling (which should make it work on wasm)

@simbleau this might also properly fix #494, incidentally.

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -55,6 +55,16 @@ struct Args {
     /// Use `0` for an automatic choice
     #[arg(long, default_value_t=default_threads())]
     num_init_threads: usize,
+    /// Use the asynchronous pipeline (if available) for rendering
+    ///
+    /// The asynchronous pipeline is one approach for robust memory - see
+    /// <https://github.com/linebender/vello/issues/366>
+    ///
+    /// However, it also has potential latency issues, especially for
+    /// accessibility technology, as it (currently) blocks the main thread for
+    /// extended periods
+    #[arg(long)]
+    async_pipeline: bool,
 }
 
 fn default_threads() -> usize {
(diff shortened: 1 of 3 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -451,8 +461,9 @@ fn run(
                             .surface
                             .get_current_texture()
                             .expect("failed to get surface texture");
-                        #[cfg(not(target_arch = "wasm32"))]
-                        {
+                        // Note: we don't run the async/"robust" pipeline, as
+                        // it requires more async wiring for the readback.
```
simbleau, line 465: Could we get more detail here? What is meant by "async wiring" - Is this a limitation of !Send futures, single-threaded wasm, etc. - What *exactly* is the issue on WASM that we aren't doing async pipelines there?

simbleau (approved):

examples/with_winit/src/lib.rs
DJMcNab: The thread is [#gpu > async on wasm](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/async.20on.20wasm). Broadly, the issue is: if we spawn a future, that needs to keep hold of the `scene`. But that's incompatible with lifetimes of keeping the scene around (to reuse allocations, etc.)

In theory, we should do this much more carefully, but we haven't yet. That might also be useful in contexts outside of browsers too, allowing multi-threaded scene rendering

PR #503, opened 2024-03-06
Integrate Android Tracing into with_winit

DJMcNab:
This is using our new [Tracing Android Trace](https://github.com/linebender/android_trace/tree/main/tracing_android_trace)

This in turn connects with https://developer.android.com/topic/performance/tracing/custom-events-native

You should use Android GPU Inspector (https://gpuinspector.dev/) to read these events (with the `atrace_apps: "*"` set)

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,6 +161,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_trace"
+version = "0.1.0"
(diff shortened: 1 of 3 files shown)


xorgy (approved):
with_winit is getting chunky, but I don't see anything wrong with this, except that naming unused values with `let` is not to my taste.

DJMcNab:
Without doing an explicit drop, I don't know of a better way to do this.

That is, the value is used, but only in its `Drop` implementation. The [tracing docs](https://docs.rs/tracing/latest/tracing/) recommend the same pattern

PR #504, opened 2024-03-06
Improve the CI by checking each package and feature separately.

xStrom:
As was [discussed in Zulip](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Continuous.20Integration.20Upgrades/near/425034305) we've had hidden breakages lately. One major source of these issues has been Cargo's workspace dependency feature unification. So if some package in a workspace enables a feature, all other packages may incorrectly depend on that feature without specifying the requirement.

Our CI already split `--lib` & `--bins` from other targets for that reason, however the Vello repo has its examples as separate packages, and also just has more packages in general. Thus a more robust CI is needed.

Because we have a lot of different repos in Linebender, I strive to make changes in a widely applicable way. That means avoiding specifying package names if I can help it. Luckily `cargo-hack` has a much better `--workspace` implementation that doesn't do feature unification and instead runs the commands on each workspace member separately.

With the introduction of `cargo-hack` we now also have access to its feature matrix options. I went with `--each-feature` and `--optional-deps` to run checks with every feature separately. (*In the future we can also consider adding a less frequent job with `--feature-powerset --depth 10` or something like that to really test the combinations.*)

I also converted the wasm32 job to be more like the main job and that already caught a bunch of issues.

---
* Hid the CPU shaders behind the `wgpu` feature, as right now that is just the reality.
* Fixed the `full` feature of `vello_encoding`.
* Fixed the `wgpu-profiler` feature of `with_winit` when targeting wasm32.
* Added missing `js` feature to `getrandom` dependencies.
* Added dummy `lib.rs` files to our binary-only examples. Unfortunately right now this is the easiest solution to the problem that Cargo will error out with `--lib` if there is no library. This requirement might be removed by `cargo-hack` in the future as they do have [an aspirational issue](https://github.com/taiki-e/cargo-hack/issues/189) describing this problem.


examples/headless/Cargo.toml
```diff
@@ -17,3 +17,6 @@ pollster = { workspace = true }
 env_logger = "0.11.2"
 png = "0.17.13"
 futures-intrusive = { workspace = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2.12", features = ["js"] }
```
DJMcNab, line 22: Does headless not already depend on `scenes`, seen as you're adding this here

.github/workflows/ci.yml
```diff
@@ -85,33 +101,21 @@ jobs:
...
-
-      - name: cargo clippy (all features) (auxiliary)
-        run: cargo clippy --workspace --tests --benches --examples --all-features -- -D warnings
+      - name: cargo clippy
+        run: cargo hack clippy --workspace --each-feature --optional-deps -- -D warnings
+  
+      - name: cargo clippy (auxiliary)
+        run: cargo hack clippy --workspace --each-feature --optional-deps --tests --benches --examples -- -D warnings
 
       # At the time of writing, we don't have any tests. Nevertheless, it's better to still run this
```
DJMcNab, line 110: This is no longer true :)

Although I suspect our tests could easily get flaky if we add more, given that `wgpu` intentionally serialises their tests. We probably should make something related to their system

DJMcNab (approved):
Thanks - it's good to get these sorted

examples/headless/Cargo.toml
xStrom: Good catch! I added this because `headless` errored out before `scenes`, but yes now that `scenes` has it that's all we need.

PR #505, opened 2024-03-07
Refactor CI jobs to allow for more concurrency.

xStrom:
With #504 merged we now do Clippy checks for each package and feature separately. That is great for catching issues but it comes with the downside of increasing the wall time our CI takes to run.

The increased wall time comes due to several factors:
* There is just more work to do, as different features mean we need to compile packages multiple times.
* More compilation artifacts means it also takes more time to compress/decompress by the cache step.
* The `cargo test` pass that comes after `cargo clippy` will do feature unification and so is guaranteed to force another compilation pass on a bunch of crates.

This PR here addresses the third point, i.e. that `cargo test` is more decoupled from `cargo clippy` now. That also means that there are fewer reasons to keep it in the same job. Thus this PR splits the `cargo clippy + test` jobs into separate `cargo clippy` and `cargo test` jobs.

Splitting the jobs will have a significant impact on wall time. For wall time what ultimately matters is the slowest job, which is the old `cargo clippy + test` on Windows. I estimate that this PR will cut its wall time by ~40% from 12m30s to 7m30s. (***Edit:** The slowest Windows test on this PR took 7m39s. There will be run-to-run variance of course, but it's the ballpark.*)

The pre- and post-step overhead of the old job is about 17% - most of which is cache related and will be split between jobs now too. Not a clear split as there is some overlap in artifacts, but e.g. for Windows instead of a single 900MB cache we get a 350MB one and a 700MB one. So splitting the cache jobs won't result in a big CPU time increase and helps cut down the wall time even more.

---

I also switched all our macOS jobs to be explictly `macos-14`, because it's just so much faster. The following table shows the time comparison with cold cache runs. N=1 but the difference is huge and matches by previous casual observations, so I'm not concerned with run-to-run variance.

| Job | Time | Diff |
|---|---|---|
| cargo clippy (macos-latest) | 10m 51s | |
| cargo clippy (macos-14) | 2m 53s | -73% |
| &nbsp; | &nbsp; | &nbsp; |
| cargo doc (macos-latest) | 2m 25s | |
| cargo doc (macos-14) | 1m 39s | -32% |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,10 +214,9 @@ jobs:
 
   doc:
     name: cargo doc
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [windows-latest, macos-14, ubuntu-latest]
+    # NOTE: We don't have any platform specific docs in this workspace, so we only run on Ubuntu.
+    #       If we get per-platform docs (win/macos/linux/wasm32/..) then doc jobs should match that.
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,14 @@ env:
 #
 # Using cargo-hack also allows us to more easily test the feature matrix of our packages.
 # We use --each-feature & --optional-deps which will run a separate check for every feature.
+#
+# We use macos-14 explictly instead of macos-latest because:
+#   * macos-latest currently points to macos-12
+#   * macos-14 provides us with the GPU support we want for testing
+#   * macos-14 comes with the M1 CPU which compiles our code much faster than the older runners
+# This explicit dependency can be switched back to macos-latest once it points to macos-14,
+# which is expected to happen sometime in Q2 FY24 (April – June 2024).
+# https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/
 
 name: CI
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -54,11 +62,71 @@ jobs:
       - name: check copyright headers
         run: bash .github/copyright.sh
 
+  clippy-stable:
+    name: cargo clippy
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [windows-latest, macos-14, ubuntu-latest]
```
simbleau, line 70:
```suggestion
        # TODO: Switch to macos-latest after macos-14 is latest.
        os: [windows-latest, macos-14, ubuntu-latest]
```

xStrom: There are three places that it needs to be changed right now, and I expect that count to grow soon with [rfcs#5](https://github.com/linebender/rfcs/pull/5). That's why I added the note about this to the top of the file, as opposed to one of those numerous places.

.github/workflows/ci.yml
```diff
@@ -167,12 +212,12 @@ jobs:
...
           RUSTFLAGS: '-D warnings'
 
-  docs:
+  doc:
     name: cargo doc
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        os: [windows-latest, macos-14, ubuntu-latest]
```
DJMcNab, line 220: Is it necessary to document for all platforms here? We don't have any platform specific code at this level

DJMcNab (approved):

.github/workflows/ci.yml
xStrom: Yeah, the Vello repo doesn't have an active need for it. It doesn't increase wall time in most scenarios, but if there are a bunch of PRs doing CI at the same time, these extra jobs can actually start to matter. Thus I reduced the doc job to only Ubuntu here and added a note about potential future expansion.

PR #512, opened 2024-03-09
Switch from `dialoguer` to `inquire`

### LINKED ISSUE #511: Switch from `dialoguer` to `inquire`
I can't compile vello anymore on 1.76 stable for very strange reasons related to `lazy_static`, see https://github.com/console-rs/dialoguer/issues/306

`dialoguer` is less maintained, less stars, etc. than `inquire`, and dialoguer never made the change to `once_cell` and I think in recent versions of rust (1.75+?) a dep with lifetimes in lazy_static are no longer producing macro-expansion that is valid.

The issue is simple... replace dialoguer with inquire. Should take 5 minutes, but I wanted to explain the purpose here.

We only use `dialoguer` on native to assist with downloads, which we could easily replace with the "Confirm" equivalent on inquire: https://github.com/mikaelmello/inquire?tab=readme-ov-file#Confirm

simbleau:
Closes #511

### DIFF
--- a/examples/scenes/Cargo.toml
+++ b/examples/scenes/Cargo.toml
@@ -18,7 +18,7 @@ instant = { workspace = true }
 # Used for the `download` command
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 byte-unit = "4.0.19"
-dialoguer = "0.10.4"
+inquire = "0.7"
 ureq = "2.9.6"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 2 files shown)


xStrom:
`inquire` does look like a nice package. One slight downside being that all its nice features also make it heavier. However this is for an example and so it doesn't really matter.

examples/scenes/Cargo.toml
```diff
@@ -18,7 +18,7 @@ instant = { workspace = true }
 # Used for the `download` command
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 byte-unit = "4.0.19"
-dialoguer = "0.10.4"
+inquire = "0.7"
```
xStrom, line 21: I'd like to also specify the patch version so that `cargo upgrade` can edit it when needed.

```suggestion
inquire = "0.7.0"
```

xStrom (approved):

PR #513, opened 2024-03-09
Bring in Linebender wide `rustfmt` settings

### LINKED ISSUE #510: rustfmt settings (linebender-wide?)
In setting up the seed for vello_svg, a comment came up that there's a fragmented story around rustfmt.toml settings across linebender.

"Kurbo, Glazier, Druid, and Piet have the following. Vello and Xilem right now don't have the file at all. I think going with the Kurbo one is fine." - @xStrom 

```toml
max_width = 100
use_field_init_shorthand = true
newline_style = "Unix"
```

And I'd love to add the following, which keeps imports very tidy: 
```toml
imports_granularity = "Crate"
```

Is this something that needs more discussion or can I chug along with a PR to add all of these settings?

simbleau:
Closes #510 

Applies the standard rustfmt settings - But also with a one-time assist from `imports_granularity = "Module"` to get things more consistent there until `imports_granularity` is stable.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,3 @@
+max_width = 100
+use_field_init_shorthand = true
+newline_style = "Unix"
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,3 +1,4 @@
 max_width = 100
 use_field_init_shorthand = true
 newline_style = "Unix"
+# TODO: imports_granularity = "Module" - Wait for this to be stable.
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,4 +1,3 @@
-max_width = 100
 use_field_init_shorthand = true
 newline_style = "Unix"
 # TODO: imports_granularity = "Module" - Wait for this to be stable.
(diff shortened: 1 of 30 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,4 @@
+max_width = 100
```
DJMcNab, line 1: I don't think we should repeat this default value (also mentioned in linebender/velato#13)

simbleau: Do you have an alternative suggestion? This is the value used in other linebender repos, e.g. glazier/peniko (if I recall correctly which repos those were)

DJMcNab: I would suggest not including it. I.e.
```suggestion
```

xStrom: It is included in the other repos indeed. However given that it is the default value, it doesn't really matter. I don't have strong feelings either way, so I guess let's just remove it.

DJMcNab: To clarify, I would like to remove it because it suggests we're choosing something other than the default

The line is doing nothing but increasing the potential confusion. I don't think it can have any advantage

DJMcNab (approved):

PR #515, opened 2024-03-09
Upgrade naga dependency to 0.19

sepcnt:


simbleau (approved):
I'll approve, seems straight forward. However I think someone who works with the shaders more should approve too, e.g. @armansito . Was there a reason we were on naga 0.13? That seems... very old. I also assumed naga was included by wgpu already (if not as a feature?), so i'm equally confused.

raphlinus:
The shaders crate is primarily for Skia integration and is not directly consumed by Vello, though it should be (see #467). It definitely makes sense to bump this, as my new shader code makes use of constants and so that's a source of warnings on #496. It's possible that this will break some stuff on the Skia side, so it does make sense for @armansito to take another look.

Also, our policy is for the author to merge, so I'll turn on write access for @sepcnt.

Thanks!

waywardmonkeys:
> I'll approve, seems straight forward. However I think someone who works with the shaders more should approve too, e.g. @armansito . Was there a reason we were on naga 0.13? That seems... very old. I also assumed naga was included by wgpu already (if not as a feature?), so i'm equally confused.

They skipped from release 0.14 to 0.19 when they folded it in with the wgpu release train and versioning.

armansito:
Thank you for the change, the naga dependency was just stale and needed to be updated.

PR #516, opened 2024-03-09
Better error types

### LINKED ISSUE #285: Error handling
At the moment, our error types are just `Box<dyn Error>`:
https://github.com/linebender/vello/blob/61ae3cde03a4cc88d6b312fd6e18384d1fd5dfa3/src/engine.rs#L29

I'm actually having trouble finding places where this is constructed, as e.g.:
https://github.com/linebender/vello/blob/61ae3cde03a4cc88d6b312fd6e18384d1fd5dfa3/src/engine.rs#L160-L166
Also returns an `Error`, despite being infallible (any error which occurs is handled by wgpu's [error handler](https://docs.rs/wgpu/latest/wgpu/struct.Device.html#method.push_error_scope) instead)

The only case I can see is in:
https://github.com/linebender/vello/blob/61ae3cde03a4cc88d6b312fd6e18384d1fd5dfa3/src/lib.rs#L156

I'm sure there are probably other reasons it's so widely used, though

I think the standard is to use [`derive(Error) from thiserror`](https://docs.rs/thiserror/latest/thiserror/) for making error enums. 

In #284, @dfrg correctly observed pointed out that finishing clips is a another point where errors can occur. This is an interesting case, because there are a number of philosophies around handling it. Skia ignores similar cases, and returning an error would be rather unergonomic for the majority of cases where this wouldn't be hit.

I see some possible solutions:
- Return an error, and let callers deal with it
- Return an error, but add a helper which just drops the error
- Return an error, but add a helper which panics on error
- Poison the `SceneBuilder`, meaning that unless the a special method (e.g. `take_errors()->Option<...>`) is called, the `SceneBuilder` panics when dropped

DasLixou:
Using custom `thiserror` errors instead of `Box<dyn Error>`, closes #285

This removed the use of `Result` as a return type for many functions, including `RenderContext::new()`. This "could" fail when no backend feature is activated for wgpu, but because we don't vendor them, this should never happen. (also there is no failable version of this function)

I also had to make a `RendererError` which I totally hate, but the problem here is that without the `wgpu-profiler` feature this function never fails, so should I remove the `Result` and just unwrap for the `GpuProfiler`?

There are still many unwraps, which may(?) still be handled correctly with errors, but for most of them I'm not sure. They are in many functions which had `Result` return types but still, `unwrap` was called, so was it intention?


src/lib.rs
```diff
@@ -186,6 +182,32 @@ impl FromIterator<AaConfig> for AaSupport {
...
+#[derive(Error, Debug)]
+pub enum AsyncRenderError {
+    #[error("error while run recording")]
+    WgpuRecordingError(#[from] WgpuRecordingError),
+    #[error(transparent)]
+    BufferAsyncError(#[from] wgpu::BufferAsyncError),
+    #[error("channel was closed")]
+    ChannelWasClosed,
+}
+
```
simbleau, line 210: I don't like this as it is... I'd be a fan of some top-level error type like `VelloError`, this feels a little error-heavy.

In addition, when I read this, I think "in what context would a channel be closed? I'm trying to render." - So the errors aren't helpful. Every error should be an error that a benefits a developer to identify a mistake or change their control flow.

src/shaders.rs
```diff
@@ -82,7 +82,7 @@ pub fn full_shaders(
     device: &Device,
     engine: &mut WgpuEngine,
     options: &RendererOptions,
-) -> Result<FullShaders, Error> {
+) -> FullShaders {
```
simbleau, line 85: This would be a breaking change and one I'm unsure of. Where did the error go?

src/util.rs
```diff
@@ -24,17 +23,31 @@ pub struct DeviceHandle {
...
+
+impl Default for RenderContext {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl RenderContext {
-    pub fn new() -> Result<Self> {
+    pub fn new() -> Self {
```
simbleau, line 41: Once again, breaking change. Where did the error go?

src/wgpu_engine.rs
```diff
@@ -131,6 +132,14 @@ enum TransientBuf<'a> {
...
 }
 
+#[derive(Error, Debug)]
+pub enum WgpuRecordingError {
+    #[error("buffer for indirect dispatch not in map")]
+    IndirectDispatchBufferNotInMap,
+    #[error("buffer for download not in map")]
+    DownloadBufferNotInMap,
+}
+
```
simbleau, line 142: Similar comments to the first, this doesn't help a developer make adjustments they need. The error doesn't feel like it's necessary.

simbleau (changes requested):
I think this needs to be redone with a high-level `VelloError`, and only have variants which apply to helping a developer. Simply saying "channel was closed" in an error message is only confusing, for example, and doesn't address how this came about or how to fix it, or an event a developer cares about.

src/shaders.rs
DasLixou: There was never one. This was also mentioned in the issue, where they are multiple `Result` returning functions which just straight-up never fail

src/util.rs
DasLixou: Here again, there was no way to return an `Err`, just panic or `Ok` at the end

src/wgpu_engine.rs
DasLixou: Yeah I should dig deeper into the errors and try to get what they mean and also document that right and maybe also ask in zulip for explanations. Currently it's just `.ok_or(Err("".into()))?` rewritten as enum variant

src/lib.rs
simbleau: A root-level error is common for targeted crates. I could be persuaded if there's a reason we shouldn't use a root-level error type. But in my opinion, Vello is doing 1 job, so I don't see why we need multiple error types everywhere.

The error types I would expect to see, which matter to a developer:
```
enum VelloError {
    // The target does not support WebGPU, e.g. your web browser (wasm32)
    WebGpuNotSupported,
    // Features that WebGpu needs are not available (e.g. Compute shaders)
    GpuFeaturesNotAvailable(list),
    // Failed to get or create a surface for the GPU device
    SurfaceCreation(#[from] wgpu::Error)
    // Shader errors, wgpu pipeline errors, etc.
    Render(#[from] source)
}
```

src/shaders.rs
simbleau: I'm wondering why it didn't return `Infallible` before, but that's good to know.

src/lib.rs
```diff
@@ -447,11 +480,7 @@ impl Renderer {
             let buf_slice = bump_buf.slice(..);
             let (sender, receiver) = futures_intrusive::channel::shared::oneshot_channel();
             buf_slice.map_async(wgpu::MapMode::Read, move |v| sender.send(v).unwrap());
-            if let Some(recv_result) = receiver.receive().await {
-                recv_result?;
-            } else {
-                return Err("channel was closed".into());
-            }
+            receiver.receive().await.expect("channel was closed")?;
```
DasLixou, line 483: So I was unsure if this could even happen and what it would mean for the end user if the channel was closed, so I replaced this with a panic

src/util.rs
```diff
@@ -44,20 +50,20 @@ impl RenderContext {
...
+            .ok_or(VelloError::CouldntFindSuitableDevice)?;
 
         let device_handle = &self.devices[dev_id];
         let capabilities = surface.get_capabilities(&device_handle.adapter);
         let format = capabilities
             .formats
             .into_iter()
             .find(|it| matches!(it, TextureFormat::Rgba8Unorm | TextureFormat::Bgra8Unorm))
-            .expect("surface should support Rgba8Unorm or Bgra8Unorm");
+            .ok_or(VelloError::UnsupportedSurfaceFormat)?;
```
DasLixou, line 66: Replaced panic with error here

src/wgpu_engine.rs
```diff
@@ -539,10 +538,9 @@ impl WgpuEngine {
                                 .with_parent(Some(&query));
                             cpass.set_pipeline(&wgpu_shader.pipeline);
                             cpass.set_bind_group(0, &bind_group, &[]);
-                            let buf = self
-                                .bind_map
-                                .get_gpu_buf(proxy.id)
-                                .ok_or("buffer for indirect dispatch not in map")?;
+                            let buf = self.bind_map.get_gpu_buf(proxy.id).ok_or(
+                                VelloError::UnavailableBufferUsed(proxy.name, "indirect dispatch"),
```
DasLixou, line 542: I made `"indirect dispatch"` and `"download"` strings here. Should I make an enum for those two so it is easier to match on the error?

DasLixou:
I would like to also make the CPU variant non-panic but there's no output of the function saying that it isn't available, just that it's in wrong format.
https://github.com/linebender/vello/blob/main/src/wgpu_engine.rs#L512

src/lib.rs
```diff
@@ -504,7 +512,7 @@ impl Renderer {
         scene: &Scene,
         surface: &SurfaceTexture,
         params: &RenderParams,
-    ) -> Result<Option<BumpAllocators>, AsyncRenderError> {
+    ) -> VResult<Option<BumpAllocators>> {
```
Philipp-M, line 515: I think you can just use `Result` instead of `VResult`. This is more common IME.

DasLixou: So should I make them all explicit?

src/util.rs
Philipp-M: I don't think we currently have to be that careful with breaking changes, when they make sense. vello is still (pre-)alpha. (Especially if they create less pain, changing to from `Self` -> `Result<Self>` would be more pain to refactor downstream). 

That said, I haven't yet looked deep enough whether this can not fail under normal circumstances though.

src/lib.rs
Philipp-M: No I meant you could just do:

```rust
pub(crate) type Result<T> = Result<T, VelloError>;
```

DasLixou: I wouldn't do that. For a `Box<dyn Error>` this is ok, but for specific Errors, it should at least have the `V` before.

Philipp-M: Well I guess you could do that everytime you're only using `Result<T, VelloError>` in a module (and no other type of `Result`, it's only syntax sugar to reduce boilerplate anyway, and that is relatively common practice IME. It should of course not be exported from the lib (in case I certainly would agree).

DasLixou: Nevermind, won't happen. In the same file i also use `Result<(), wgpu::Error>` so it gives name clash.

DJMcNab: I believe the fix for that is:
```rust
type Result<T, E=VelloError> = Result<T, E>;
```

src/shaders.rs
waywardmonkeys, line 85: Is there value in having this part of the change be a separate PR that can land sooner than the rest? (Like I ended up doing with #547 for another part as I didn' know it was covered here.)

waywardmonkeys (dismissed):

src/util.rs
```diff
@@ -24,17 +24,23 @@ pub struct DeviceHandle {
     pub queue: Queue,
 }
 
+impl Default for RenderContext {
```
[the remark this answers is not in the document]
DasLixou: Oh I see

src/shaders.rs
DasLixou: probably.. It has some unwraps in it so I was hoping to get discussion instead of removing the `Result` but.. silence

src/lib.rs
```diff
@@ -184,6 +179,45 @@ impl FromIterator<AaConfig> for AaSupport {
...
+    /// Make sure that you have a surface which provides one of those texture formats.
+    #[cfg(feature = "wgpu")]
+    #[error("Couldn't find `Rgba8Unorm` or `Bgra8Unorm` texture formats for surface")]
+    UnsupportedSurfaceFormat,
+
+    /// Error when using a buffer inside a recording while it's not available.
+    /// Check if you have created it and not freed before its last usage.
+    #[cfg(feature = "wgpu")]
+    #[error("Buffer '{0}' is not available but used for {1}")]
+    UnavailableBufferUsed(&'static str, &'static str),
```
simbleau, line 204: Should this only exist if the profiler feature is active, as the docs suggest?

src/lib.rs
```diff
@@ -184,6 +179,45 @@ impl FromIterator<AaConfig> for AaSupport {
...
+
+    /// Error when creating [`GpuProfiler`].
+    /// See [`wgpu_profiler::CreationError`] for more information.
+    #[cfg(feature = "wgpu-profiler")]
+    #[error("Couldn't create wgpu profiler")]
+    ProfilerCreationError(#[from] wgpu_profiler::CreationError),
+}
+
+#[allow(dead_code)]
+pub(crate) type Result<T, E = VelloError> = std::result::Result<T, E>;
```
simbleau, line 219: Why is this dead code? Can we just remove this?

src/lib.rs
simbleau: Why are you unsure if it can happen? There should be a soundness "guarantee" with any panic. I'd rather not upgrade the error to a panic unless we are sure.

src/util.rs
```diff
@@ -50,15 +50,15 @@ impl RenderContext {
...
+            .ok_or(VelloError::CouldntFindSuitableDevice)?;
 
         let device_handle = &self.devices[dev_id];
         let capabilities = surface.get_capabilities(&device_handle.adapter);
         let format = capabilities
             .formats
             .into_iter()
             .find(|it| matches!(it, TextureFormat::Rgba8Unorm | TextureFormat::Bgra8Unorm))
-            .expect("surface should support Rgba8Unorm or Bgra8Unorm");
+            .ok_or(VelloError::UnsupportedSurfaceFormat)?;
```
simbleau, line 61: Love this.

src/wgpu_engine.rs
```diff
@@ -553,7 +552,7 @@ impl WgpuEngine {
                     let src_buf = self
                         .bind_map
                         .get_gpu_buf(proxy.id)
-                        .ok_or("buffer not in map")?;
+                        .ok_or(VelloError::UnavailableBufferUsed(proxy.name, "download"))?;
```
simbleau, line 555: This error doesn't feel meaningful... is there a way for the developer to react if this is triggered? I don't think so... this error would only happen if we (as Vello maintainers) aren't handling the buffers correctly. In such case, I'd accept a panic here instead with a safety comment.

simbleau (changes requested):
Another preliminary pass. It looks better.

src/lib.rs
DasLixou: Where is that written in the docs?

src/lib.rs
DasLixou: yeah, that can be removed.

src/lib.rs
simbleau: "/// Error when using a buffer inside a recording"

Aren't the recordings related to the `wgpu-profiler` feature? Genuinely not sure, need some guidance.

waywardmonkeys:
@DasLixou How far away from ready is this? Any chance it could be done for Vello 0.2?

DasLixou:
I would have to look over this, will probably do a rebase, but IIRC nothing really blocked this, only some not-100%-sure decisions

src/lib.rs
```diff
@@ -184,6 +179,45 @@ impl FromIterator<AaConfig> for AaSupport {
     }
 }
 
+/// Errors that can occur in vello.
+#[derive(Error, Debug)]
+pub enum VelloError {
+    /// Error when vello couldn't find a wgpu device with suitable features.
+    #[cfg(feature = "wgpu")]
+    #[error("Couldn't find suitable device")]
+    CouldntFindSuitableDevice,
```
DJMcNab, line 188: I think I'd prefer this to be called `NoCompatibleDevice`?

DJMcNab (commented):
Sorry for not reviewing this earlier - some initial comments, although I've not done a full review

DasLixou:
image
Sometimes merge conflicts are just hilarious 😂 (at least they're easy)

DasLixou:
That should be it

DasLixou:
Also just checked that everything runs fine

simbleau:
I have many comments that are unresolved still

src/lib.rs
DJMcNab: I agree with @DasLixou that this error case is impossible, and so panicking is fine.

Anyway, it's in the legacy `_async` code, which we're no longer recommending, so I think this is fine

src/lib.rs
simbleau: Please don't mark my comments as resolved unless you've responded to them.

src/wgpu_engine.rs
DJMcNab: This is a question about where our abstraction boundaries lie. Concretely, there are effectively two parts:
1) The recording abstraction
2) Connecting that abstraction to the Vello shaders

As seen in #416, the recording can theoretically be used for additional functionality by end-users.
Our current API here is very much research quality, and so. I think it's likely that we'll want to have slightly more fine-grained error types for this in future (i.e. a different error at the recording/higher-level Vello levels), but this is fine for this PR.

DJMcNab (approved):
Once this is rebased/updated with main we can land this. We aren't committing to any stability guarantees, so I think it's perfectly reasonable to land this as an improvement.

This doesn't really change the API we expect our users to use, as seen in `with_winit` not needing to be updated.

src/wgpu_engine.rs
simbleau: I think it's fine, personally.

src/lib.rs
simbleau: If it was unclear, I'm asking whether or not this could be `#[cfg(all(feature = "wgpu", feature = "wgpu-profiler"))]`

src/lib.rs
```diff
@@ -183,6 +178,46 @@ impl FromIterator<AaConfig> for AaSupport {
...
+
+    /// Failed to create [`GpuProfiler`].
+    /// See [`wgpu_profiler::CreationError`] for more information.
+    #[cfg(feature = "wgpu-profiler")]
+    #[error("Couldn't create wgpu profiler")]
+    ProfilerCreationError(#[from] wgpu_profiler::CreationError),
+}
+
+#[allow(dead_code)] // this can be unused when wgpu feature is not used
+pub(crate) type Result<T, E = VelloError> = std::result::Result<T, E>;
```
simbleau, line 219: This is just a side note but I've always disliked the specialized `Result` types that crates do. It just seems like a bad pattern across the rust ecosystem.

src/lib.rs
```diff
@@ -183,6 +178,46 @@ impl FromIterator<AaConfig> for AaSupport {
     }
 }
 
+/// Errors that can occur in vello.
+#[derive(Error, Debug)]
+pub enum VelloError {
```
simbleau, line 183: Should this be `#[non_exhaustive]`?

src/lib.rs
DJMcNab: I think this is fine here. We're not re-exporting it, which is correct.

simbleau:
> This doesn't really change the API we expect our users to use, as seen in `with_winit` not needing to be updated.

That's because the example panics or discards most errors. Currently, that's not giving me the confidence I want yet.

I think approving this is a bit hasty as-is. It's definitely an improvement, but the error types will have a big effect on the crate going forward. 

I'd prefer if we were specific with the feature flags on error types, whether `VelloError` should be `non_exhaustive`, and (personal opinion) get rid of the specialized `Result` type we've made, and generally address the comments that are withstanding so there is less uncertainty.

simbleau:
Agree with Chad's comment above, but glad to see all the improvements in such a short turnaround. Nice work @DasLixou !

Approving

simbleau (approved):

src/lib.rs
```diff
@@ -183,6 +178,47 @@ impl FromIterator<AaConfig> for AaSupport {
     }
 }
 
+/// Errors that can occur in vello.
+#[derive(Error, Debug)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
xStrom: I renamed it to `Error`.

PR #517, opened 2024-03-09
Update `skrifa` to `0.16.0`

DasLixou:


xStrom:
Unfortunately there is no changelog so I can't be sure if we need to change anything on our side when upgrading. Perhaps @dfrg can give this a quick look as he should be more informed.

DasLixou:
I looked at the commits and couldn't find anything so I just yoloed into CI to see if something breaks. But yeah asking is definitely the better way to go

dfrg:
It’s probably not worth tracking skrifa very closely for the time being. We often bump versions for internal reasons to propagate changes in underlying dependencies to other projects. This will likely settle down in the next few months.

There shouldn’t be any breaking changes that affect anything in vello so this is safe to merge.

Philipp-M (approved):

Issue #291, opened 2023-03-07
Validation Error on Default Scene Render


power-unsealed:
The following example tries to render an empty scene.

```rust
use vello::{util::RenderContext, Renderer, RendererOptions, Scene, RenderParams, peniko::Color};
use wgpu::{TextureDescriptor, Extent3d, TextureDimension, TextureFormat, TextureUsages, TextureViewDescriptor};
use pollster::FutureExt as _;

fn main() {
    let mut context = RenderContext::new().unwrap();
    let device_id = context.device(None).block_on().unwrap();
    let device_handle = &mut context.devices[device_id];
    let device = &device_handle.device;
    let queue = &device_handle.queue;
    let mut renderer = Renderer::new(
        device,
        &RendererOptions { surface_format: None }
    ).unwrap();

    let scene = Scene::default();
    let (width, height) = (1024, 1024);
    let texture = device.create_texture(&TextureDescriptor {
        label: Some("target"),
        size: Extent3d { width, height, depth_or_array_layers: 1 },
        mip_level_count: 1,
        sample_count: 1,
        dimension: TextureDimension::D2,
        format: TextureFormat::Rgba8Unorm,
        usage: TextureUsages::STORAGE_BINDING | TextureUsages::COPY_SRC,
        view_formats: &[],
    });
    let texture_view = texture.create_view(&TextureViewDescriptor::default());
    let params = RenderParams { base_color: Color::BLACK, width, height };

    renderer.render_to_texture(device, queue, &scene, &texture_view, &params).unwrap();
}
```

The code crashes at runtime with the following output:
```
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a ComputePass
      note: encoder = `<CommandBuffer-(0, 1, Vulkan)>`
    In a dispatch command, indirect:false
      note: compute pipeline = `pathtag_reduce`
    Buffer is bound with size 16 where the shader expects 20 in group[0] compact index 2
```

The same happens, when rendering an empty scene to a `winit` window.

DJMcNab:
Thanks for the report.
This is because of https://www.w3.org/TR/webgpu/#minimum-buffer-binding-size

I think the correct fix *might* be to skip the parts of the render which fail if there are no paths, which may well be everything before `fine`. In theory, this case should be equivalent with clearing the given texture to the specified background colour? Can we make that "optimisation" to avoid needing to think about this.

For now, in user code, a workaround would be to always render something, e.g. a triangle of the background colour.

[...]

PR #518, opened 2024-03-10
Resurrect pico_svg and use it to replace vello_svg in the examples.

xorgy:


README.md
```diff
@@ -120,14 +120,6 @@ More formal benchmarks are on their way.
 
 ## Integrations
 
-### SVG
```
DJMcNab, line 123: We should probably mention [`vello_svg`](https://github.com/linebender/vello_svg/) here. Maybe that should be a follow-up PR?

examples/scenes/src/svg.rs
```diff
@@ -94,16 +96,43 @@ pub fn svg_function_of<R: AsRef<str>>(
...
+                            );
+                        }
+                    }
+                }
+                eprintln!("Encoded svg {name} in {:?}", start.elapsed());
+                (new_scene, size.to_vec2())
+            }
+            std::result::Result::Err(e) => {
+                eprintln!("{:?}", e);
+                (Scene::new(), Vec2::ZERO)
```
DJMcNab, line 133: If you press <kbd>S</kbd> to close the stats view, this runs into #291 

That's probably a good thing, but we should have an empty test scene as well

examples/scenes/src/svg.rs
```diff
@@ -94,16 +96,43 @@ pub fn svg_function_of<R: AsRef<str>>(
...
+                                &stroke.path,
+                            );
+                        }
+                    }
+                }
+                eprintln!("Encoded svg {name} in {:?}", start.elapsed());
+                (new_scene, size.to_vec2())
+            }
+            std::result::Result::Err(e) => {
+                eprintln!("{:?}", e);
```
DJMcNab, line 132: This could be much more descriptive

examples/scenes/src/svg.rs
```diff
@@ -94,16 +96,43 @@ pub fn svg_function_of<R: AsRef<str>>(
...
-            .unwrap_or_else(|e| panic!("failed to parse svg file {name}: {e}"));
-        eprintln!("Parsed svg {name} in {:?}", start.elapsed());
-        let start = Instant::now();
-        let mut new_scene = Scene::new();
-        vello_svg::render_tree(&mut new_scene, &svg);
-        let resolution = Vec2::new(svg.size.width() as f64, svg.size.height() as f64);
-        eprintln!("Encoded svg {name} in {:?}", start.elapsed());
-        (new_scene, resolution)
+        match PicoSvg::load(contents, 1.0) {
+            std::result::Result::Ok(PicoSvg { items, size }) => {
```
DJMcNab, line 102: Why use the fully qualified path here?

examples/scenes/src/svg.rs
DJMcNab, line 132: This is printing the string `Wrong` for me. That isn't helpful.

I can't even find the string "Wrong" in the source code, so it's not even partly trackable.

examples/scenes/src/pico_svg.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub path: BezPath,
+}
+
+struct Parser<'a> {
+    scale: f64,
+    items: &'a mut Vec<Item>,
+}
+
+impl PicoSvg {
+    pub fn load(xml_string: &str, scale: f64) -> Result<PicoSvg, Box<dyn std::error::Error>> {
```
DJMcNab, line 41: Why not use `anyhow::Result`? `scenes` already depends on that, and it lets us add context to the errors

examples/scenes/src/pico_svg.rs
```diff
@@ -0,0 +1,271 @@
...
+            match vs.as_slice() {
+                &[x, y, width, height] => (Point { x, y }, Size { width, height }),
+                _ => (
+                    Point::ZERO,
+                    Size {
+                        width: 300.0,
+                        height: 150.0,
+                    },
+                ),
+            }
```
DJMcNab, line 60: This is getting the wrong viewbox for the Forma paris-30k I have locally. It's not clear whether it's using the fallback. It might be good to have a printout for the two fallback conditions?

examples/scenes/src/svg.rs
xorgy: Conflicting `Ok` and `Result` in the namespace.

README.md
xorgy: Now that it's published I'll add it back in and point to the repo. :+1:

examples/scenes/src/svg.rs
DJMcNab: Hmm, that's annoying.

Maybe it should it use the same fix as in #516 , where the helper  type is defined as:
```rust
type Result<T, E = {default}> = Result<T, E>;
```
?

examples/scenes/src/svg.rs
xorgy: Now returns an obnoxious fuchsia unit square instead of an empty scene.

examples/scenes/src/svg.rs
```diff
@@ -94,16 +96,51 @@ pub fn svg_function_of<R: AsRef<str>>(
...
+                                &stroke.path,
+                            );
+                        }
+                    }
+                }
+                eprintln!("Encoded svg {name} in {:?}", start.elapsed());
+                (new_scene, size.to_vec2())
+            }
+            std::result::Result::Err(e) => {
+                eprintln!("Failed to load svg, with Err({e:?})");
```
DJMcNab, line 132:
```suggestion
                eprintln!("Failed to load svg, with Err({e})");
```
Errors should implement Display, which is the better value to show for them

examples/scenes/src/svg.rs
xorgy: I'm too dumb for this, sorry.

examples/scenes/src/pico_svg.rs
xorgy: I'll give that a shot.

examples/scenes/src/pico_svg.rs
xorgy: This works for me on the paris-30k from the MPVG publication package, but I think it's technically incorrect when width, height, and viewBox are all defined at the same time.

examples/scenes/src/svg.rs
DJMcNab: Hmm, I'm surprised that this fails. I'll need to dig into it more myself.

I don't really understand how it could fail

examples/scenes/src/pico_svg.rs
xorgy: It's a bit of a mess mixing the two in pico_svg, since roxml is using std:: as well.

examples/scenes/src/svg.rs
xorgy: Comes from a number parsing error somewhere, but doesn't seem to be related to the transform parsing.

examples/scenes/src/svg.rs
```diff
@@ -94,16 +96,51 @@ pub fn svg_function_of<R: AsRef<str>>(
...
+                                &stroke.path,
+                            );
+                        }
+                    }
+                }
+                eprintln!("Encoded svg {name} in {:?}", start.elapsed());
+                (new_scene, size.to_vec2())
+            }
+            Err(e) => {
+                eprintln!("Failed to load svg, with Err({e})");
```
DJMcNab, line 132:
```suggestion
                eprintln!("Failed to load svg: {e}");
```

examples/scenes/src/pico_svg.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+fn parse_transform(transform: &str) -> Affine {
+    let mut nt = Affine::IDENTITY;
+    for ts in transform.split(')').map(str::trim) {
+        nt *= if let Some(s) = ts.strip_prefix("matrix(") {
+            let vals = s
+                .split(|c| matches!(c, ',' | ' '))
+                .map(str::parse)
+                .collect::<Result<Vec<f64>, ParseFloatError>>()
```
DJMcNab, line 191:
```suggestion
                .collect::<Result<Vec<f64>, _>>()
```

DJMcNab (commented):
I've only skimmed the code, but it looks reasonable.

I don't think we can land this in a form which doesn't support all of the default downloads.
1) Coat of arms of Poland gives "Unable to parse a number"
2) More concerning, the SVG stroke test examples don't output any elements at all

The solution for 1 could be removing Coat of arms of Poland. But I don't think that's viable for issue 2.

xorgy:
@DJMcNab processing the examples with usvg CLI makes them all render, most of them correctly (except the ones that rely on stroke-linejoin and other things that are not decoded in pico_svg).

DJMcNab (approved):
> @DJMcNab processing the examples with usvg CLI makes them all render, most of them correctly (except the ones that rely on stroke-linejoin and other things that are not decoded in pico_svg). With many of them, the issue is the use of CSS instead of attributes where it's not necessary.

I hope you can understand my thinking that's not a great end state. If the stroke features are out of scope, then I'd be happy for us to remove those examples.
I really don't think we should be downloading SVGs which don't work in the default state.
An alternative would be to host the converted SVGs ourselves (but not in this repo)

I ran into https://github.com/anishathalye/assets, which seems like a reasonable solution (of course, being careful about licenses). I think we should do something similar for the readme image, for example

So, approving conditional on removing the "broken" assets from the default_downloads, but with suggestions of alternative solutions which we could use.

dfrg:
My understanding was that we’d punt fancy SVG functionality to vello_svg and this repo would focus purely on tests that are directly relevant to vello dev, specifically tiger and Paris (and maybe Poland?). I don’t see why we need the download bits at all.

PR #520, opened 2024-03-14
Experimental code to increase robustness of cubic params

raphlinus:
This only calls atan2 when the derivatives are sufficiently large.

WIP, not to commit, as GPU and CPU are out of sync.

### DIFF
--- a/shader/flatten.wgsl
+++ b/shader/flatten.wgsl
@@ -76,16 +76,33 @@ struct EulerSeg {
     params: EulerParams,
 }
 
+// Threshold below which a derivative is considered too small.
+const DERIV_THRESH: f32 = 1e-6;
+const DERIV_THRESH_SQUARED: f32 = DERIV_THRESH * DERIV_THRESH;
+// Amount to nudge t when derivative is near-zero.
+
 /// Compute cubic parameters from endpoints and derivatives.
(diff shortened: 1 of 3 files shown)


shader/flatten.wgsl
```diff
@@ -200,7 +217,7 @@ fn es_seg_from_params(p0: vec2f, p1: vec2f, params: EulerParams) -> EulerSeg {
 
 fn es_seg_eval_with_offset(es: EulerSeg, t: f32, offset: f32) -> vec2f {
     let chord = es.p1 - es.p0;
-    let scaled = offset / length(chord);
+    let scaled = offset / max(length(chord), TANGENT_THRESH);
```
armansito, line 220: Not sure if this should be capped at `sqrt(TANGENT_THRESH)` instead of `TANGENT_THRES`.

armansito:
I pushed commits with a fix for divides by zero chord length, a minor clean up of constants, and a port of the fixes to the CPU euler code.

Seeing as the AMD issues likely have a different root cause, I think we can move this PR out of draft and land it while investigating the other issues.

armansito:
This is superseded by #521, which incorporates the fixes from here plus some restructuring that makes the code cleaner.

raphlinus:
Indeed, closing in favor of #521.

PR #524, opened 2024-03-18
Pipeline caching!

DJMcNab:
Supercedes #459

This is the version of pipeline caching using gfx-rs/wgpu#5319

This leads to *massive* improvements in startup time, especially on Android.

### DIFF
--- a/examples/with_winit/src/minimal_pipeline_cache.rs
+++ b/examples/with_winit/src/minimal_pipeline_cache.rs
@@ -1,0 +1,111 @@
+//! Code related to managing wgpu pipeline caches, especially on Android
+
+use std::{
+    io::ErrorKind,
+    path::{Path, PathBuf},
+};
+
+use vello::wgpu::{self, AdapterInfo, Device, Features, PipelineCache};
+use winit::event_loop::EventLoop;
+
+#[cfg(target_os = "android")]
+fn get_cache_directory_android<T>(event_loop: &EventLoop<T>) -> anyhow::Result<PathBuf> {
+    use std::path::PathBuf;
+
+    use anyhow::Context;
+    use winit::platform::android::EventLoopExtAndroid;
+
+    let app = event_loop.android_app();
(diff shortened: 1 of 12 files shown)


PoignardAzur:
Not a blocker for this PR, but `RendererOptions` should probably be non_exhaustive to avoid breaking changes in cases like this?

examples/with_winit/src/minimal_pipeline_cache.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/// Load a pipeline cache for the given device, or create one if not present.
+///
+/// # Safety
+///
+/// The directory should only have been written into by [`write_pipeline_cache`].
+///
+/// We recognise that this is an impossible standard, but that's because Vulkan
+/// makes UB unavoidable around invalid pipeline caches, so just do your best.
```
PoignardAzur, line 71: That's unfortunate. Can we at least embed a checksum in the cache file? If possible, one with a program-local salt so it's extra hard to get the wrong cache by accident?

DJMcNab: I struggle to see the point. wgpu already does some sanity checks here, and it's still possible to do UB (if *something* decides to write to the wrong place with a forged cache with our hash in place).

I don't believe adding it here for this example to be a good use of my time.

If you think this would be a good use of your time, there is space reserved for a hash in the header wgpu writes. The place to contribute that hashing is: https://github.com/gfx-rs/wgpu/blob/trunk/wgpu-core/src/pipeline_cache.rs. The thing that you'd need to use as salt is probably instead (most of) the data in `PipelineCacheHeader`

DJMcNab:
I'll make it non-exhaustive here. There's little advantage to splitting it out.

xorgy (approved):
Looks good. This is self contained, and errors therein are failsafe (apart from if JVM stuff explodes, but that will never be safe).

PR #525, opened 2024-03-18
[shaders] Use constants in switch statements

armansito:
Naga didn't support constant evaluation until recently, which prevented constant declarations from being referenced in switch case statements. Now that naga supports consteval (starting in 0.19) and the feature is available in vello, this PR updates the shaders to reference constants directly instead of numerical literals.


raphlinus (approved):
This is a nice improvement, only having access to literal numbers in those positions felt very ugly. Thanks!

PR #526, opened 2024-03-19
Allow large numbers of draw objects

### LINKED ISSUE #334: Flicking problem
I have test a svg file, which is not that big, not as big as the CIA map case.

When I loaded the file, and zoom in, I find that the screen is flicking, some small parts not rendering correctly.

Is this because of float precision problem?

Quit sure there is no clipping in this file

test

raphlinus:
Previously there was a limit of workgroup size squared for the number of draw objects, which is 64k in practice. This PR makes each workgroup iterate multiple blocks if that limit is exceeded, borrowing a technique from FidelityFX sort.

WIP, this causes hangs on mac. Uploading to test on other hardware.

Also contains some changes for testing that may not want to be committed as is.

Fixes #334

### DIFF
--- a/shader/draw_leaf.wgsl
+++ b/shader/draw_leaf.wgsl
@@ -51,11 +51,9 @@ var<workgroup> sh_scratch: array<DrawMonoid, WG_SIZE>;
 
 @compute @workgroup_size(256)
 fn main(
-    @builtin(global_invocation_id) global_id: vec3<u32>,
     @builtin(local_invocation_id) local_id: vec3<u32>,
     @builtin(workgroup_id) wg_id: vec3<u32>,
 ) {
-    let ix = global_id.x;
     // Reduce prefix of workgroups up to this one
     var agg = draw_monoid_identity();
     if local_id.x < wg_id.x {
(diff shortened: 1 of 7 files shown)


shader/draw_leaf.wgsl
```diff
@@ -74,184 +72,198 @@ fn main(
...
     workgroupBarrier();
-    var m = sh_scratch[0];
-    workgroupBarrier();
-    let tag_word = read_draw_tag_from_scene(ix);
-    agg = map_draw_tag(tag_word);
-    sh_scratch[local_id.x] = agg;
-    for (var i = 0u; i < firstTrailingBit(WG_SIZE); i += 1u) {
+    var prefix = sh_scratch[0];
+
+    let num_blocks_total = (config.n_drawobj + (WG_SIZE - 1u)) / WG_SIZE;
```
armansito, line 77: nit: You could remove the parentheses around `WG_SIZE - 1u`.

shader/draw_leaf.wgsl
```diff
@@ -74,184 +72,198 @@ fn main(
...
+    let n_blocks_base = num_blocks_total / WG_SIZE;
+    let remainder = num_blocks_total % WG_SIZE;
+    let first_block = n_blocks_base * wg_id.x + min(wg_id.x, remainder);
+    let n_blocks = n_blocks_base + u32(wg_id.x < remainder);
```
armansito, line 81: If I understand lines 79-81 correctly, they distribute the remainder across workgroups to load balance? I would add a comment here that clarifies the block/workgroup/thread assignment.

I would also consider moving these calculations to a helper somewhere in shared/ so that they can be reused in `shader/draw_reduce.wgsl` (though I don't feel strongly about this).

shader/draw_reduce.wgsl
```diff
@@ -13,20 +13,29 @@ var<storage> scene: array<u32>;
...
 @compute @workgroup_size(256)
 fn main(
-    @builtin(global_invocation_id) global_id: vec3<u32>,
     @builtin(local_invocation_id) local_id: vec3<u32>,
+    @builtin(workgroup_id) wg_id: vec3<u32>,
 ) {
-    let ix = global_id.x;
-    let tag_word = read_draw_tag_from_scene(ix);
-    var agg = map_draw_tag(tag_word);
+    let num_blocks_total = (config.n_drawobj + (WG_SIZE - 1u)) / WG_SIZE;
```
armansito, line 27: Same comment about explaining the block->thread assignment.

armansito (approved):
Looks good!

shader/draw_leaf.wgsl
raphlinus: Basically the reason for these parentheses was that I didn't trust constant evaluation to understand the associativity of addition. But I think it's better to trust the compiler here, and have the cleanest expression of programmer intent.

PR #527, opened 2024-03-21
Revert workaround for miscompilation

raphlinus:
We have an ugly workaround for miscompilation of a workgroup uniform load (see #199). The miscompilation no longer repros, and it's not clear we still need it. Possibly something changed in naga, or possibly Apple drivers have improved.

It would be possible to go a little further and trim the allocation of the Paths array so it's not rounded up to workgroup size, but the practical benefit of that is marginal.

Sending this as a PR to see if there still may be problems.


DJMcNab (dismissed):
Approving based on positive response in Zulip thread.

armansito (approved):

PoignardAzur:
Is this still relevant?

PR #531, opened 2024-03-23
Improve MSRV docs.

xStrom:
This PR does minor wording adjustments to the MSRV notice and also moves the Cargo command to its own line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,13 +236,17 @@ This version of Vello has been verified to compile with **Rust 1.75** and later.
 Future versions of Vello might increase the Rust version requirement.
 It will not be treated as a breaking change and as such can even happen with small patch releases.
 
-> [!TIP]
-> As time has passed, some of Vello's dependencies could have released versions with a higher Rust requirement.
-> If you encounter a compilation issue due to a dependency and don't want to upgrade your Rust toolchain, then you could downgrade the dependency.
-> ```sh
-> # Use the problematic dependency's name and version
-> cargo update -p package_name --precise 0.1.1
-> ```
+<details>
+<summary>Click here if compiling with MSRV fails.</summary>
+
+As time has passed, some of Vello's dependencies could have released versions with a higher Rust requirement.
+If you encounter a compilation issue due to a dependency and don't want to upgrade your Rust toolchain, then you could downgrade the dependency.
+
+```sh
+# Use the problematic dependency's name and version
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -231,14 +231,19 @@ VELLO_STATIC_LOG="vello=trace" VELLO_STATIC_ARGS="--test-scenes" cargo apk run -
...
-This version of Vello has been verified to compile with Rust 1.75 and later.
-
-Note that since then, one of Vello's dependencies might have released a new version with a higher requirement.
-Thus if you encounter a compilation issue due to a dependency, you should downgrade it via `cargo update -p package_name --precise 0.1.1` (with a real name and version).
+This version of Vello has been verified to compile with **Rust 1.75** and later.
 
 Future versions of Vello might increase the Rust version requirement.
 It will not be treated as a breaking change and as such can even happen with small patch releases.
 
+> [!TIP]
```
DJMcNab, line 239: See [#linebender > GitHub (markdown) alerts](https://xi.zulipchat.com/#narrow/stream/419691-linebender/topic/GitHub.20.28markdown.29.20alerts) on using these alerts. It has some unfortunate consequences

README.md
```diff
@@ -231,14 +231,19 @@ VELLO_STATIC_LOG="vello=trace" VELLO_STATIC_ARGS="--test-scenes" cargo apk run -
...
+> As time has passed, some of Vello's dependencies could have released versions with a higher Rust requirement.
+> If you encounter a compilation issue due to a dependency and don't want to upgrade your Rust toolchain, then you could downgrade the dependency.
+> ```sh
+> # Use the problematic dependency's name and version
+> cargo update -p package_name --precise 0.1.1
+> ```
```
DJMcNab, line 245: I'm not completely convinced that we actually need this whole section to give this much detail. It seems like too much detail for those who don't need this, but too little detail for anyone who actually does (e.g. why do we have dependencies which can do these kinds of publish?)

Can we make "downgrade the dependency" a link to cargo update's doc?

DJMcNab (approved):
This is an improvement in the formatting, so approving. But I'm not convinced the content is actually that useful to have front-and-center

README.md
xStrom: I don't have strong feelings about this syntax, so I'll remove it from this PR so that discussion on the syntax can remain separated.

DJMcNab (approved):

PR #532, opened 2024-03-23
Remove stale clippy installations.

xStrom:
The `test-stable` and `test-stable-wasm` CI jobs don't actually use Clippy so we don't need to spend time installing it.


waywardmonkeys (approved):
IiRC, clippy is part of the base profile, so this won’t save any time. But it is more correct.

xStrom:
Clippy is part of the default profile, however the action we use opts-in to the [minimal profile](https://rust-lang.github.io/rustup/concepts/profiles.html) which does not come with Clippy.

PR #535, opened 2024-03-27
Make draw_leaf pass uniformity analysis

raphlinus:
It was in fact uniform, but required a little arithmetic to infer that. Refactor the math slightly so it's clear even to a simplistic analysis.


armansito (approved):
I verified the fix against Tint. Thanks!

PR #537, opened 2024-03-28
Add robustness to GPU shaders

raphlinus:
Make each stage quit early if a previous stage has failed.

The CPU shaders are minimally changed to be layout compatible. For the most part, they'll panic on a bounds check if sizes are exceeded. That's arguably useful for debugging, but a case can be made they should have the same behavior as the GPU shaders.

Work towards #366

### DIFF
--- a/shader/binning.wgsl
+++ b/shader/binning.wgsl
@@ -53,7 +53,7 @@ var<workgroup> sh_bitmaps: array<array<atomic<u32>, N_TILE>, N_SLICE>;
 // store count values packed two u16's to a u32
 var<workgroup> sh_count: array<array<u32, N_TILE>, N_SUBSLICE>;
 var<workgroup> sh_chunk_offset: array<u32, N_TILE>;
-var<workgroup> sh_atomic_failed: u32;
+var<workgroup> sh_previous_failed: u32;
 
 @compute @workgroup_size(256)
 fn main(
@@ -66,10 +66,10 @@ fn main(
     }
     if local_id.x == 0u {
         let failed = bump.lines > config.lines_size;
-        sh_atomic_failed = u32(failed);
+        sh_previous_failed = u32(failed);
     }
     // also functions as barrier to protect zeroing of bitmaps
-    let failed = workgroupUniformLoad(&sh_atomic_failed);
+    let failed = workgroupUniformLoad(&sh_previous_failed);
     if failed != 0u {
         if global_id.x == 0u {
             bump.failed |= STAGE_FLATTEN;
--- a/shader/binning.wgsl
+++ b/shader/binning.wgsl
@@ -53,6 +53,7 @@ var<workgroup> sh_bitmaps: array<array<atomic<u32>, N_TILE>, N_SLICE>;
 // store count values packed two u16's to a u32
 var<workgroup> sh_count: array<array<u32, N_TILE>, N_SUBSLICE>;
 var<workgroup> sh_chunk_offset: array<u32, N_TILE>;
+var<workgroup> sh_atomic_failed: u32;
 
 @compute @workgroup_size(256)
 fn main(
@@ -63,7 +64,18 @@ fn main(
     for (var i = 0u; i < N_SLICE; i += 1u) {
         atomicStore(&sh_bitmaps[i][local_id.x], 0u);
     }
-    workgroupBarrier();
+    if local_id.x == 0u {
+        let failed = bump.lines > config.lines_size;
+        sh_atomic_failed = u32(failed);
+    }
+    // also functions as barrier to protect zeroing of bitmaps
+    let failed = workgroupUniformLoad(&sh_atomic_failed);
+    if failed != 0u {
+        if global_id.x == 0u {
+            bump.failed |= STAGE_FLATTEN;
+        }
+        return;
+    }
 
     // Read inputs and determine coverage of bins
     let element_ix = global_id.x;
--- a/shader/coarse.wgsl
+++ b/shader/coarse.wgsl
@@ -73,6 +73,9 @@ fn alloc_cmd(size: u32) {
         let ptcl_dyn_start = config.width_in_tiles * config.height_in_tiles * PTCL_INITIAL_ALLOC;
         var new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
         if new_cmd + PTCL_INCREMENT > config.ptcl_size {
+            // This sets us up for technical UB, as lots of threads will be writing
+            // to the same locations. But I think it's fine, and predicating the
+            // writes would probably slow things down.
             new_cmd = 0u;
             atomicOr(&bump.failed, STAGE_COARSE);
         }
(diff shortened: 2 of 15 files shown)


raphlinus:
Possibly this can be merged as-is, but perhaps more work should be done so that clients can handle the failures reasonably. For now, rendering appears to freeze (zoom into longpathdash, then zoom back out; rendering will resume).

shader/binning.wgsl
```diff
@@ -63,7 +64,18 @@ fn main(
...
     }
-    workgroupBarrier();
+    if local_id.x == 0u {
+        let failed = bump.lines > config.lines_size;
+        sh_atomic_failed = u32(failed);
+    }
+    // also functions as barrier to protect zeroing of bitmaps
+    let failed = workgroupUniformLoad(&sh_atomic_failed);
+    if failed != 0u {
+        if global_id.x == 0u {
```
DJMcNab, line 74: It probably isn't worth having a comment that this is to avoid data races here

shader/coarse.wgsl
```diff
@@ -73,6 +73,9 @@ fn alloc_cmd(size: u32) {
         let ptcl_dyn_start = config.width_in_tiles * config.height_in_tiles * PTCL_INITIAL_ALLOC;
         var new_cmd = ptcl_dyn_start + atomicAdd(&bump.ptcl, PTCL_INCREMENT);
         if new_cmd + PTCL_INCREMENT > config.ptcl_size {
+            // This sets us up for technical UB, as lots of threads will be writing
+            // to the same locations. But I think it's fine, and predicating the
+            // writes would probably slow things down.
```
DJMcNab, line 78: Would it be reasonable to have `new_cmd=cmd_offset` here? I think that would avoid the UB - instead we'd just overwrite in the same location in each loop

shader/coarse.wgsl
```diff
@@ -152,11 +155,16 @@ fn main(
...
+            || atomicLoad(&bump.seg_counts) > config.seg_counts_size;
         // Reuse sh_part_count to hold failed flag, shmem is tight
-        sh_part_count[0] = atomicLoad(&bump.failed);
+        sh_part_count[0] = u32(failed);
     }
     let failed = workgroupUniformLoad(&sh_part_count[0]);
-    if (failed & (STAGE_BINNING | STAGE_TILE_ALLOC | STAGE_PATH_COARSE)) != 0u {
+    if failed != 0u {
+        if wg_id.x == 0u && local_id.x == 0u {
+            atomicOr(&bump.failed, STAGE_COARSE);
```
DJMcNab, line 166: What does this represent? I would think this should be `STAGE_{BEFORE_COARSE}`, and then only if the check on `seg_counts` failed

shader/path_count_setup.wgsl
```diff
@@ -16,8 +16,12 @@ let WG_SIZE = 256u;
 
 @compute @workgroup_size(1)
 fn main() {
-    let lines = atomicLoad(&bump.lines);
-    indirect.count_x = (lines + (WG_SIZE - 1u)) / WG_SIZE;
+    if atomicLoad(&bump.failed) != 0u {
+        indirect.count_x = 0u;
```
DJMcNab, line 20: I'm impressed that this works. Reading the specs suggest it's fine. :shipit:

DJMcNab (approved):
A few nits, but generally this looks like a decent improvement

I think the CPU shaders probably do need some rethinking, but I'm not sure the best way to do so

shader/coarse.wgsl
raphlinus: Intriguing idea! However, that won't quite avoid UB, as cmd_offset will edge into the allocation following this one. Setting it to `cmd_limit - (PTCL_INCREMENT - PTCL_HEADROOM)` *almost* works, but only if it's not in its initial segment. I can't think of a good solution in that case, as we still ideally want the limit where it is so it accurately allocates as if there were enough memory, for the purposes of reporting the size back. My gut feeling is that if we were very concerned about the technical UB, we should in fact predicate the writes.

shader/binning.wgsl
```diff
@@ -63,7 +64,18 @@ fn main(
     for (var i = 0u; i < N_SLICE; i += 1u) {
         atomicStore(&sh_bitmaps[i][local_id.x], 0u);
```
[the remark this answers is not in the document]
raphlinus: Ah, I didn't realize that was a strong guarantee. In WebGPU world, it's probably worth skipping this explicit zeroing, but in native world it might be worth compiling with zeroing by infrastructure disabled, in which case we would need this.

shader/coarse.wgsl
raphlinus: I'll make the change, now that I'm thinking these flags will be interpreted by the client, not just to early-out downstream.

shader/binning.wgsl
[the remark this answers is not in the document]
DJMcNab: Good point - for us it's not impactful, but e.g. before https://github.com/linebender/vello/pull/363 this would have mattered for the MSL conversion

shader/coarse.wgsl
DJMcNab: Ah, because the allocations are variably sized? I can't say I'm that happy about adding UB, but I do agree that it's unlikely to cause a problem in practise.

I wonder how bad the cost of writing to the same location is in terms of memory bandwidth/cache coherency?
I agree that this is probably fine as-is, though

DJMcNab (approved):
In its current state, the fact that there's no feedback at all when we miss a frame isn't ideal. I guess for debugging we could have a feature which slightly increases the effective alpha (/applies a red tinged overlay?) when rendering fails

Although that would have to increase each frame (which might actually be ideal, to give a view on the latency we get in dealing with under-allocation in #366.

shader/path_count_setup.wgsl
armansito: Yes, this works and it's the only way I'm aware of that allows you to "abort" this type of indirect dispatch (there are more sophisticated ways with bindless, see for example: https://developer.apple.com/documentation/metal/indirect_command_encoding/encoding_indirect_command_buffers_on_the_gpu?language=objc).

Interestingly, I couldn't find any _explicit_ wording in the WebGPU, Metal, Vulkan, or D3D12 docs that this is the expected behavior but "0" falls within the accepted range for all of them. See also this past discussion: https://github.com/gpuweb/gpuweb/issues/1045

armansito (approved):
Approved with one question.

shader/coarse.wgsl
```diff
@@ -152,11 +155,19 @@ fn main(
     // We need to check only prior stages, as if this stage has failed in another workgroup, 
     // we still want to know this workgroup's memory requirement.   
     if local_id.x == 0u {
+        var failed = atomicLoad(&bump.failed) & (STAGE_BINNING | STAGE_TILE_ALLOC | STAGE_FLATTEN);
+        if atomicLoad(&bump.seg_counts) > config.seg_counts_size {
+            failed |= STAGE_PATH_COUNT;
```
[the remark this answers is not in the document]
raphlinus: Basically because path_count doesn't bind config. I'm also a bit wary of divergence but there's probably no meaningful impact on performance. I think it basically comes down to a style issue whether you tick the flag there or later.

If you can really rely on buffer robustness, then maybe at some point you can drop the write predication and just look at the read after the fact. One thing at the back of my head is the possibility of wrapping u32, but I think I'll choose not to worry about that too much right now.

PR #538, opened 2024-03-31
Impl `From<Encoding>` for `Scene`.

### LINKED ISSUE #530: Appending an `Encoding` to a `Scene`
I don't see a way to append an `Encoding` to a `Scene` at the moment.

I could see doing this as:

* New method on `Scene`: `Scene::append_encoding()`
* New method on `Scene`: `Scene::get_encoding_mut()` and then work with that underlying encoding
* Allow wrapping an existing `Encoding` as a `Scene` and then use the existing `Scene::append()`

waywardmonkeys:
This allows creating a `Scene` with a pre-existing `Encoding`.

Fixes #530.

### DIFF
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -221,7 +221,8 @@ impl Scene {
 
 impl From<Encoding> for Scene {
     fn from(encoding: Encoding) -> Self {
-        // TODO: Properly update the estimator for the encoding.
+        // It's fine to create a default estimator here, and that field will be
+        // removed at some point - see https://github.com/linebender/vello/issues/541
         Self {
             encoding,
             #[cfg(feature = "bump_estimate")]
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -219,6 +219,17 @@ impl Scene {
     }
 }
 
+impl From<Encoding> for Scene {
+    fn from(encoding: Encoding) -> Self {
+        // TODO: Properly update the estimator for the encoding.
+        Self {
+            encoding,
+            #[cfg(feature = "bump_estimate")]
+            estimator: vello_encoding::BumpEstimator::default(),
+        }
+    }
+}
+
 /// Builder for encoding a glyph run.
 pub struct DrawGlyphs<'a> {
     encoding: &'a mut Encoding,


src/scene.rs
```diff
@@ -219,6 +223,17 @@ impl Scene {
     }
 }
 
+impl From<Encoding> for Scene {
+    fn from(encoding: Encoding) -> Self {
+        // TODO: Properly update the estimator for the encoding.
```
DJMcNab, line 228: And this is why I'd make it a `from_encoding` method, to document that the bump estimator will undercount items added in this method

waywardmonkeys: In other conversations, it has already been said that the implementation strategy of the estimator will be changing and this issue would go away.

src/scene.rs
```diff
@@ -14,8 +14,12 @@ use vello_encoding::{Encoding, Glyph, GlyphRun, Patch, Transform};
 
 /// The main datatype for rendering graphics.
 ///
-/// A Scene stores a sequence of drawing commands, their context, and the
+/// A `Scene` stores a sequence of drawing commands, their context, and the
 /// associated resources, which can later be rendered.
+///
+/// A `Scene` can be created in a default (empty) state via `Scene::new()`
+/// or via `Default`. It can also be created from a pre-existing [`Encoding`]
+/// using `From`.
```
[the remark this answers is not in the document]
waywardmonkeys: I think it aids in discoverability. It helps make it clear that this isn't a type you'd be getting from somewhere else.

src/scene.rs
PoignardAzur: Note that if you're adding a TODO to the code, it's usually good practice to join a link to a ticket (usually a github issue) tracking down whatever needs to be done to remove that TODO.

waywardmonkeys: It matches the existing TODO comments elsewhere in the file for the same feature. (Also, out of roughly 80 TODOs in this repo, approximately 2 have an issue number in them.)

src/scene.rs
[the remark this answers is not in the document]
DJMcNab: Can you point to pre-existing examples of this style of paragraph in idiomatic libraries?

Just to check myself, I've looked in [`Vec`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html), [`HashMap`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html),  [`bevy_ecs::World`](https://docs.rs/bevy/latest/bevy/ecs/world/struct.World.html), and [`BezPath`](https://docs.rs/kurbo/latest/kurbo/struct.BezPath.html).

The closest I have found is "A HashMap with a known list of items can be initialized from an array", which *is* calling out a `From` impl, but makes no mention of the `new` associated function or `Default` impls.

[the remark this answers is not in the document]
waywardmonkeys: This far out-exceeds my budget of caring. I will remove it when I have the time.

src/scene.rs
PoignardAzur: We can do a comments pass at some point, but I think at least new comments should follow that practice.

DJMcNab: We could point to https://github.com/linebender/vello/issues/530#issuecomment-2016585267

DJMcNab (approved):
Thanks - this looks like it should give more flexibility to consumers

The concern about bump estimation and the future state of that are valid, but we're in flux at the moment, and @armansito's comment in the linked issue is clear enough that the bump estimator field will change substantially

I don't think adding a link to that comment is too important, although it would probably be an improvement

src/scene.rs
DJMcNab: Or indeed https://github.com/linebender/vello/issues/541

PR #539, opened 2024-04-01
[binning] Do not use numerical operators on atomics

armansito:
This fixes a comparison expression and a compound-assignment expression that used an atomic as an operand. These should be treated as illegal based on my understanding of the WGSL specification (and are rejected by Tint) but Naga accepts them and injects appropriate atomic operations during code generation.

See https://github.com/gfx-rs/wgpu/issues/5474 for more details.


raphlinus (approved):
Good catch, and thanks also for raising the overly lenient behavior of naga.

PR #540, opened 2024-04-02
[fine] Coverage-mask specialization and some shader feature cleanup

armansito:
This change introduces a variant of the fine rasterizer stage that is geared towards coverage mask rendering. Specifically:

1. The "r8" permutations can be used to define the texel format of the target texture as `r8unorm`, which is supported via a non-standard extension in Dawn.
2. The "full" macro has been repurposed such that it specifically applies to the gradient and image support in the fine stage. The vello_shaders crate now supports a "full" feature flag. When this is omitted,  path rendering with solid brushes and clipping are supported across all pipelines.

This PR also includes some minor cleanups:
- Remove the `force_rw_storage` feature, which is no longer used.
- Remove the case block in fine that handles the `CMD_STROKE` PTCL command, which is now obsolete.

----
###  Background on the "full" variant
Previously the "full" macro was introduced to distinguish the complete imaging model of piet-gpu from the more limited "reduced" variant, which was intended to produce coverage masks for an atlas based path renderer. This reduced variant was limited to only paths without any blends, clips, or other non-path draw objects. This specialization wasn't fully carried forward with the move from piet-gpu to Vello, since some of the excluded features (in particular anti-aliased clipping) turned out to be a desirable feature for atlasing.

While there is still room for pipeline specializations and optimizations geared towards coverage-mask atlas based path renderers, the existing meaning of "full" left some important features out (such as the PTCL and stroke style / path_ix monoids used in stroke expansion).

### DIFF
--- a/crates/shaders/Cargo.toml
+++ b/crates/shaders/Cargo.toml
@@ -10,9 +10,13 @@ repository.workspace = true
 publish = false # Remove this when the package is ready for publishing
 
 [features]
-default = ["compile", "wgsl", "msl"]
+default = ["compile", "wgsl", "msl", "full"]
 compile = ["naga", "thiserror"]
 
+# Enables the complete imaging model. When this feature is disabled, the fine rasterization
+# stage only supports drawing paths with a solid brush and clipping, and the shaders can
+# not be run with an encoding that contains gradient fills and images.
+full = []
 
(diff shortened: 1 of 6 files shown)


raphlinus (approved):
This looks very useful for Skia, and will have minimal impact on Vello main. Thanks!

Issue #541, opened 2024-04-03
Improving bump estimation accuracy and performance


armansito:
The bump buffer size estimation utility is currently integrated into the scene construction interface of the scene API. In this model, each `Scene` object owns a `BumpEstimator` that maintains intermediate tallies of encoded data for a single instance of `Encoding`. When a `Scene` gets appended to another `Scene`, the bump tallies of the argument scene fragment also get appended to the tallies of the appendee, after a heuristic-driven scale is applied to it based on the provided transform.

This design has some significant drawbacks:
- The size of the render target is not known during scene construction. The `segment` and `tile` buffer counts are currently likely to be massively overestimated when objects are scaled up to have significant portions lie outside the viewport. The estimator can take this into account to discard culled objects. While the current estimate for the "line soup" buffer is independent of the render target, there are good reasons to apply viewport clipping and culling during the curve flattening stage.
- The heuristic based scaling is less accurate compared to relying on precisely transformed coordinates.
- Glyphs and other shapes/resources that get resolved late are currently ignored as estimating them at encoding time is tricky.

The most straightforward solution to this is to run the estimation during scene resource resolution (see `vello_encoding::Resolver::resolve()`). Resolution happens every time a scene gets rendered at which point all fragment data, precise absolute transforms, late bound resources, and render target parameters (most importantly the viewport dimensions) are available.

This approach isn't without it's own drawbacks. The `BumpEstimator` has a modest but non-zero cost on CPU-time performance. Measurements on the current integration show a 1.75x - 2.3x increase to encoding time when the `bump_estimate` feature is enabled. This impact can be significant for complex scenes.

This performance impact is likely hard to avoid (at least without optimizations) for dynamic scenes, however there can be a significant advantage to avoiding this computation on every frame for a static scene (such as a SVG scene where the user only interacts with the transform) or reused scene fragments. Given the trade-offs, I have the following thoughts / proposals:

1. Move the estimation to resolve time so that the overestimation drawbacks above can be avoided. This can allow the BumpEstimate results to be more seamlessly integrated with the `Layout` and `RenderConfig` structures that are used to set up a vello render.

[...]

PR #542, opened 2024-04-03
New test scene: `hugepath`

armansito:
This test draws lines with very large floating point y-coordinates (spanning `12345678901234567890.0` and `-9.8765432109876543210e+19` respectively). This causes an overflow in bump estimation and results in rendering issues in Vello.

### Bump Estimation
This test causes a panic in debug builds with the bump_estimate feature enabled:
```
thread 'main' panicked at /Users/armansito/work/vello/crates/encoding/src/estimate.rs:157:26:
attempt to multiply with overflow
```
This happens because the line generates an exorbitant amount of tile crossings which causes the segment count to overflow. This can be avoided by taking the viewport size and culling into account during estimation since a majority of the shape lies outside the viewport. See #541.

### Rendering Issues
1. The frame rate (on my M1 Pro) doesn't go higher than 6 fps for this scene.
2. The expected rendering result is similar to the following (barring colors):
<p align="center">
Screenshot 2024-04-02 at 10 36 57 PM
</p>
With vello, the top 3 strokes (winding from positive to negative) are missing under the default transform:
<p align="center">
Screenshot 2024-04-02 at 10 41 09 PM
</p>
Under rotation, (particularly >90&deg;) causes artifacts that look like this:
<p align="center">
Screenshot 2024-04-02 at 10 45 46 PM
</p>


DJMcNab:
I think I see the opposite problem (?) when pressing the <kbd>C</kbd> key in `with_winit`, where the strokes for the frame times are scaled to `1/u64::MAX`, which seems to lead to terrible performance (~200ms frame times)

DJMcNab (approved):
Thanks, this looks like it brings in a real case where we have issues

I don't know what the path towards fixing this would be

DJMcNab:
The only blocker is the clippy failure now

waywardmonkeys:
@armansito Is this something that we want to land? Trying to clear up some PRs before doing a Vello 0.2 release...

PoignardAzur:
Is it worth rebasing and merging this?

PR #544, opened 2024-04-03
Restore glyph hinting support

dfrg:
Updates skrifa to 0.19.0 and restores hinting support. Our previous glyph scaler came with its own instance cache but skrifa does not, so this also adds a very simple LRU cache to manage instances.

There will likely be adjustments to this when we do "real" glyph caching and start retaining our caches across frames but this should be sufficient for now and will hopefully improve text quality in xilem on low-resolution displays.

### DIFF
--- a/crates/encoding/src/glyph_cache.rs
+++ b/crates/encoding/src/glyph_cache.rs
@@ -7,8 +7,8 @@ use super::{Encoding, StreamOffsets};
 
 use peniko::kurbo::{BezPath, Shape};
 use peniko::{Fill, Style};
-use skrifa::instance::NormalizedCoord;
-use skrifa::outline::OutlinePen;
+use skrifa::instance::{NormalizedCoord, Size};
+use skrifa::outline::{HintingInstance, HintingMode, LcdLayout, OutlineGlyphFormat, OutlinePen};
 use skrifa::{GlyphId, OutlineGlyphCollection};
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Default, Debug)]
@@ -24,12 +24,14 @@ pub struct GlyphKey {
 pub struct GlyphCache {
     pub encoding: Encoding,
     glyphs: HashMap<GlyphKey, CachedRange>,
+    hinting: HintCache,
 }
 
 impl GlyphCache {
     pub fn clear(&mut self) {
         self.encoding.reset();
         self.glyphs.clear();
+        // No need to clear the hinting cache
     }
 
     pub fn get_or_insert(
@@ -43,6 +45,7 @@ impl GlyphCache {
         let size = skrifa::instance::Size::new(font_size);
         let is_var = !coords.is_empty();
         let encoding_cache = &mut self.encoding;
+        let hinting_cache = &mut self.hinting;
         let mut encode_glyph = || {
             let start = encoding_cache.stream_offsets();
             let fill = match style {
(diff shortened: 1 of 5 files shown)


xorgy (approved):
I don't see anything wrong with this.

crates/encoding/src/glyph_cache.rs
```diff
@@ -55,9 +58,18 @@ impl GlyphCache {
             encoding_cache.encode_fill_style(fill);
             let mut path = encoding_cache.encode_path(true);
             let outline = outlines.get(GlyphId::new(key.glyph_id as u16))?;
-            // FIXME: Re-add hinting when skrifa supports it
```
DJMcNab, line 58: :fire:

crates/encoding/src/glyph_cache.rs
```diff
@@ -145,3 +157,111 @@ impl OutlinePen for BezPathPen {
...
+        }
+    }
+
+    fn instance(&self) -> Option<HintingInstance> {
+        HintingInstance::new(self.outlines, self.size, self.coords, HINTING_MODE).ok()
+    }
+}
+
+const HINTING_MODE: HintingMode = HintingMode::Smooth {
+    lcd_subpixel: Some(LcdLayout::Horizontal),
```
DJMcNab, line 196: I'm guessing this will need to be wired through at some point?

examples/scenes/src/simple_text.rs
```diff
@@ -92,6 +92,7 @@ impl SimpleText {
             .glyph_transform(glyph_transform)
             .normalized_coords(var_loc.coords())
             .brush(brush)
+            .hint(false)
```
DJMcNab, line 95: I would expect us to want to test this?
Is it because SimpleText is also used for the headless example?

I see in Zulip that you've mentioned that  text should be on a pixel baseline, which we don't enforce in our current examples.

I still would imagine that hinting could be productively used in e.g. the stats displays?

crates/encoding/src/glyph_cache.rs
```diff
@@ -24,12 +24,14 @@ pub struct GlyphKey {
...
     pub encoding: Encoding,
     glyphs: HashMap<GlyphKey, CachedRange>,
+    hinting: HintCache,
 }
 
 impl GlyphCache {
     pub fn clear(&mut self) {
         self.encoding.reset();
         self.glyphs.clear();
+        // No need to clear the hinting cache
```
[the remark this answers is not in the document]
dfrg: Exactly. In reality, none of this is retained right now so it’s mostly irrelevant. This module will need some substantial rework when we do rasterized glyph caching.

crates/encoding/src/glyph_cache.rs
dfrg: That’s an open question. The skrifa hinter supports a lot of modes because we have a requirement of matching FreeType. There’s an argument to be made that vello should always just use a modern hinting mode but that’s open to discussion. Ultimately, most fonts these days have hints generated by ttfautohint which ignores these things. Only the MS ClearType fonts really take advantage of them and we can potentially choose a sane default for those.

examples/scenes/src/simple_text.rs
dfrg: This is disabled for the test scenes because we have some animated text which doesn’t work well with hinting (it snaps to pixels as it animates causing annoying jitter) and everything currently uses the same code path. We can definitely change this but I wanted to get the functionality in for xilem where hinting is more important.

PR #546, opened 2024-04-06
Fix typo in panic/expect message in `simple` example

DasLixou:
Unify with other `.expect` messages


xStrom (approved):
This was probably on purpose as there are some people that like this way of writing those messages. That said, I'm not part of that group, and more importantly consistency is good and the other messages are indeed written like your PR here. Thus I think we can merge this. Thanks!

PR #547, opened 2024-04-11
`RenderContext::new()` needn't return `Result`.

waywardmonkeys:
When the support was added for multiple devices, this function no longer could return an error, so a `Result` is not needed and just makes things more confusing in sample code.


DJMcNab (approved):
Sure, this seems fine. This does conflict with #516, but that's a bigger refactor

PR #555, opened 2024-04-21
Better glyph caching

dfrg:
This enables caching for variable glyphs, adds support for pruning the cache and retains the Resolver in the Renderer struct so that glyphs can be cached across frames.

Still so much to do here but this should remove our performance cliffs and better position us for future changes.

### DIFF
--- a/crates/encoding/src/glyph_cache.rs
+++ b/crates/encoding/src/glyph_cache.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use std::collections::HashMap;
+use std::rc::Rc;
 
 use super::{Encoding, StreamOffsets};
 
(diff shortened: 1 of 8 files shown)


crates/encoding/src/glyph_cache.rs
```diff
@@ -5,157 +5,245 @@ use std::collections::HashMap;
...
+    ) -> Option<GlyphCacheSession<'a>> {
+        let font_id = font.data.id();
+        let font_index = font.index;
+        let font = skrifa::FontRef::from_index(font.data.as_ref(), font.index).ok()?;
+        let map = if !coords.is_empty() {
+            // This is still ugly in rust. Choices are:
+            // 1. multiple lookups in the hashmap (implemented here)
+            // 2. always allocate and copy the key
+            // 3. use unsafe
+            // Pick 1 bad option :(
```
xorgy, line 50: :heart:

xorgy (approved):
Aside from the lint error, I don't see anything here more than the most minor nitpicks.
It also works and doesn't cause crashing for me in example applications or Xilem, including when I exercise variable fonts.

nicoburns:
Does this cache all glyphs or only variable glyphs?

dfrg:
> Does this cache all glyphs or only variable glyphs?

It caches all glyphs now

PR #556, opened 2024-04-22
Test that `Renderer` + `Scene` are `Send`.

waywardmonkeys:
Usages within Bevy and other frameworks are simpler when these types are `Send`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,6 +202,13 @@ pub struct Renderer {
     #[cfg(feature = "wgpu-profiler")]
     pub profile_result: Option<Vec<wgpu_profiler::GpuTimerQueryResult>>,
 }
+// This is not `Send` (or `Sync`) on WebAssembly as the
+// underlying wgpu types are not. This can be enabled with the
+// `fragile-send-sync-non-atomic-wasm` feature in wgpu.
+// See https://github.com/gfx-rs/wgpu/discussions/4127 for
+// further discussion of this topic.
+#[cfg(all(feature = "wgpu", not(target_arch = "wasm32")))]
+static_assertions::assert_impl_all!(Renderer: Send);
 
 /// Parameters used in a single render that are configurable by the client.
 pub struct RenderParams {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3821,6 +3821,7 @@ dependencies = [
  "peniko",
  "raw-window-handle",
  "skrifa",
+ "static_assertions",
  "vello_encoding",
  "wgpu",
  "wgpu-profiler",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ peniko = { workspace = true }
 wgpu = { workspace = true, optional = true }
 log = { workspace = true }
 raw-window-handle = { workspace = true }
+static_assertions = { workspace = true }
 futures-intrusive = { workspace = true }
 wgpu-profiler = { workspace = true, optional = true }
 
@@ -76,6 +77,7 @@ peniko = "0.1.0"
 futures-intrusive = "0.5.0"
 raw-window-handle = "0.6.0"
 smallvec = "1.13.2"
+static_assertions = "1.1.0"
 
 # NOTE: Make sure to keep this in sync with the version badge in README.md
 wgpu = { version = "0.19.3" }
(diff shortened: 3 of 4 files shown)


waywardmonkeys:
I wanted this in my own integration, but was reminded when I saw the change from `Rc` to `Arc` during the development of #555.  This will help to catch that issue earlier in the development cycle.

src/lib.rs
```diff
@@ -687,3 +687,19 @@ impl BlitPipeline {
...
+
+    #[test]
+    fn check_scene_is_send() {
+        assert_is_send::<super::Scene>();
+    }
+
+    #[test]
+    #[cfg(all(feature = "wgpu", not(target_arch = "wasm32")))]
+    fn check_renderer_is_send() {
+        assert_is_send::<super::Renderer>();
```
DJMcNab, line 703: I'm not certain whether we definitely want this, but we should at least make changing this be intentional, so this check is good.

src/lib.rs
```diff
@@ -687,3 +687,19 @@ impl BlitPipeline {
...
+mod tests {
+    fn assert_is_send<T: Send>() {}
+
+    #[test]
+    fn check_scene_is_send() {
+        assert_is_send::<super::Scene>();
+    }
+
+    #[test]
+    #[cfg(all(feature = "wgpu", not(target_arch = "wasm32")))]
```
DJMcNab, line 701: Can we have a comment explaining why this fails on wasm?

src/lib.rs
```diff
@@ -687,3 +687,19 @@ impl BlitPipeline {
...
+#[cfg(test)]
+mod tests {
+    fn assert_is_send<T: Send>() {}
+
+    #[test]
+    fn check_scene_is_send() {
+        assert_is_send::<super::Scene>();
+    }
+
+    #[test]
```
DJMcNab, line 700: The `#[test]` attribute is a bit weird here. I think a `const _: () = {}` block is more idiomatic for this kind of check.

DJMcNab (approved):
Thanks!

I would change this to use `const _: ()` instead, but that's an easy change

src/lib.rs
waywardmonkeys: Sure, but I don't know a clear place to point to that adequately describes the status.

There is https://github.com/gfx-rs/wgpu/discussions/4127 ... but some of the discussion is happening in the WebGPU groups (about what guarantees should be provided by the browsers).

Fundamentally, right now, `wgpu` types aren't `Send`, but can be forced to be with an extra feature flag:

```
## Implement `Send` and `Sync` on Wasm, but only if atomics are not enabled.
##
## WebGL/WebGPU objects can not be shared between threads.
## However, it can be useful to artificially mark them as `Send` and `Sync`
## anyways to make it easier to write cross-platform code.
## This is technically *very* unsafe in a multithreaded environment,
## but on a wasm binary compiled without atomics we know we are definitely
## not in a multithreaded environment.
fragile-send-sync-non-atomic-wasm = [
    "hal/fragile-send-sync-non-atomic-wasm",
    "wgc/fragile-send-sync-non-atomic-wasm",
    "wgt/fragile-send-sync-non-atomic-wasm",
]
```

src/lib.rs
waywardmonkeys: I've seen it done this way in Bevy and some other crates in the past.

I did it this way partly so that there is an entry in the test output indicating which checks it performed, so that I would know in this case that it actually checked the `Renderer` and not a dumb mistake in the `cfg` check.

src/lib.rs
waywardmonkeys: It is required for how the `with_bevy` sample is written.

src/lib.rs
DJMcNab: Right, this is probably fine then

src/lib.rs
DJMcNab: I still think I'd prefer it - now that we know that the check is working as expected.

src/lib.rs
waywardmonkeys: I've added a comment.

src/lib.rs
waywardmonkeys: If there's a precise change you want to see here with material benefits, please provide it. This is close to how various other crates do it (bevy, static assertions crate, etc).

DJMcNab: I thought my suggestion is exactly what the static-assertions crate implements?

waywardmonkeys: Ends up it is a regular dependency.

DJMcNab (approved):
Thanks!

PR #559, opened 2024-04-27
Initial update to winit 0.30.0

waywardmonkeys:
This doesn't yet implement the new application handler trait.


waywardmonkeys:
Marking this as ready for review in this form ... I really think that fixing it nicely will take rewriting a good chunk of `with_winit` which could happen incrementally.

xorgy (approved):

PR #561, opened 2024-04-29
Remove Bevy example.

waywardmonkeys:
This will be out of date with respect to WGPU versions shortly and won't compile. Additionally, it adds significantly to CI time and is only an example, while another crate provides a usable integration of Bevy and released versions of Vello.


simbleau:
Personally, I'm good with this because it doesn't use bevy in the way I'd expect: 2d.

It renders a texture to put on a cube, instead of splatting a fullscreen quad to render a vello scene. It was an odd choice to begin with, but I suppose did the job.

However if we do intend to support bevy as a first class platform (on the same priority as SVG/Lottie, at least) - then we should move bevy_vello into linebender.

DJMcNab (approved):

waywardmonkeys:
As decided during office hours, landing this.

@simbleau I think you should bring up the `bevy_vello` proposal!

PR #563, opened 2024-05-01
Use vello_shaders crate to load and preprocess WGSL

### LINKED ISSUE #467: `vello` should import `vello_shaders`

armansito:
This replaces the shader preprocessing code in the vello crate with invocations to the vello_shaders crate, which supports both reading from a directory at runtime (for hot_reload) and bundlng pre-compiled shader code into the binary (which is suitable for wasm).

This also moves the CPU shader code over to the vello_shaders crate behind the "cpu" feature.

Resolves #467


Cargo.toml
```diff
@@ -55,6 +55,7 @@ workspace = true
 
 [dependencies]
 vello_encoding = { workspace = true }
+vello_shaders = { workspace = true }
```
DJMcNab, line 58: Do we still need the `compile` feature to be enabled even if we're not hot reloading?

I guess it doesn't really matter, because we share the Naga version with `wgpu`

waywardmonkeys:
Since the copyright check is failing, you'll need to modify the 2 checks to account for the new location of the CPU shaders.

That's in `.github/copyright.sh`

DJMcNab (approved):
Awesome. I looked into this previously, but thought the hot reload support would have been dropped - I missed that `from_dir` function.

It's unfortunate that we can't use the pipeline struct of `vello_shaders` (because it doesn't allow parallel initialisation), but that could be resolved in the future, I guess?

src/shaders.rs
```diff
@@ -123,52 +92,48 @@ pub fn full_shaders(
...
                 $name,
                 $bindings,
-                &$defines,
-                CpuShaderType::Present(cpu_shader::$name)
+                CpuShaderType::Present(vello_shaders::cpu::$name)
             )
         };
         ($name:ident, $bindings:expr) => {
-            add_shader!($name, $bindings, &full_config)
+            add_shader!($name, $bindings)
```
DasLixou, line 106: This is the exact as above and can be removed

DasLixou: Pretty interesting, does rust not cover equal macro branches? O.o

armansito: Good catch!

Cargo.toml
armansito: That's a good point. Even though the features match, I think there is merit to excluding the feature when it's not needed. Though I generally tend to rely on linkers being smart about excluding unused code from binaries (which is desirable in the wasm build case) I went ahead and made "compile" an optional feature, which feels right to me.

armansito:
> It's unfortunate that we can't use the pipeline struct of `vello_shaders` (because it doesn't allow parallel initialisation), but that could be resolved in the future, I guess?

This is something I've been considering. I'm not entirely convinced yet that the shaders crate should abstract the pipeline initialization (which is what the currently unused PipelineHost trait declarations were intended for). If we do that, then it might make sense to have the shaders crate handle multi-threaded pipeline compilation.

I should write down my thoughts on this in an issue. I'm particularly thinking about non-wgpu engines and I'm not sure yet how to best organize that code.

PR #564, opened 2024-05-03
`add_shader` on the engine doesn't return a failure.

waywardmonkeys:
Simplify this by recognizing that `Engine::add_shader` won't return an `Error` so it doesn't need to return a `Result`.

### DIFF
--- a/src/shaders.rs
+++ b/src/shaders.rs
@@ -121,7 +121,7 @@ pub fn full_shaders(
                 } else {
                     $cpu
                 },
-            )?
+            )
         }};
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This is related to #516, but not handled there, as it removes things that claim they can return an error, but don't actually.

DJMcNab (approved):
I suspect the idea might have been for `shaders` to be "generic" over the engine being used, and some engines could reasonably have errors here.

But saying that, we don't have that set up, so this is fine.

PR #566, opened 2024-05-05
vello_tests: Remove `decode_image()`, dep on `image`

waywardmonkeys:
This wasn't used here and shouldn't be needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2702,7 +2702,6 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "futures-intrusive",
- "image",
  "png",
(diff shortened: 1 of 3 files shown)


DJMcNab:
Hmm, this is probably fine, but we are going to want to do snapshot tests at some point

DJMcNab (approved):

waywardmonkeys:
It does use the `png` crate for writing out images ... and we'd probably use the image-diffing systems for snapshot testing as discussed in xilem.

And if we did pull `image` back in here, hopefully we'd `default-features = false` and only build what was needed.

PR #571, opened 2024-05-07
deps: Update image to 0.25.1

waywardmonkeys:
changelog: https://github.com/image-rs/image/blob/main/CHANGES.md

Since we only use this for jpegs, the main change is that it switches to `zune-jpeg` which is faster than the old decoder.


xStrom (approved):
Thanks!

PR #573, opened 2024-05-09
[encoding] Fix bug in join estimate

armansito:
The buffer size estimator wasn't accounting for the inner-join line segments when counting joins. This resulted in an underestimate which is particularly a problem when the scene content consists only of line-tos (curves mask the problem since they are always over-estimated).

### DIFF
--- a/crates/encoding/src/estimate.rs
+++ b/crates/encoding/src/estimate.rs
@@ -225,6 +225,10 @@ impl BumpEstimator {
                 self.segments += count * arc_lines * count_segments_for_line_length(line_len);
             }
         }
+
+        // Count inner join lines
+        self.lines.linetos += count;
+        self.segments += count_segments_for_line_length(scaled_width) * count;
     }
 }
 


DJMcNab (approved):
I don't fully understand this at this point, but we don't use the estimator ourselves.

So happy to land this!

raphlinus (approved):
Yup. Of course, this will have to be revised when we get fully correct inner joins in place. That makes estimation harder, as the math for whether the join is included within the stroke is pretty deep geometry. On the other hand, doing the estimation for arcs will be much nicer, as the number of segments is bounded with arcs... so much deep work still to be done.

armansito:
> I don't fully understand this at this point, but we don't use the estimator ourselves.
> 
> So happy to land this!

The estimator is effectively guessing a count of the outline segments that get written out to the line soup buffer. When outputting strokes, we output the outer join according to the style but render the inner join as a single line segment. So with a stroked path, every join needs an extra count of 1 to account for that.

I documented some of the general algorithm in Section 8 of the [stroke expansion paper](https://arxiv.org/abs/2405.00127). Figure 11 may be helpful as it illustrates the join segments.

PR #575, opened 2024-05-13
Disable zeroing workgroup memory for compute shaders.

waywardmonkeys:
This is a performance improvement for shader compilation.

See https://github.com/gfx-rs/wgpu/pull/5508

### DIFF
--- a/src/wgpu_engine.rs
+++ b/src/wgpu_engine.rs
@@ -646,7 +646,7 @@ impl WgpuEngine {
             entry_point: "main",
             compilation_options: PipelineCompilationOptions {
                 zero_initialize_workgroup_memory: false,
-                ..PipelineCompilationOptions::default()
+                ..Default::default()
             },
         });
         WgpuShader {
--- a/src/wgpu_engine.rs
+++ b/src/wgpu_engine.rs
@@ -644,7 +644,10 @@ impl WgpuEngine {
             layout: Some(&compute_pipeline_layout),
             module: &shader_module,
             entry_point: "main",
-            compilation_options: PipelineCompilationOptions::default(),
+            compilation_options: PipelineCompilationOptions {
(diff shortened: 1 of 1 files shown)


src/wgpu_engine.rs
```diff
@@ -644,7 +644,10 @@ impl WgpuEngine {
             layout: Some(&compute_pipeline_layout),
             module: &shader_module,
             entry_point: "main",
-            compilation_options: PipelineCompilationOptions::default(),
+            compilation_options: PipelineCompilationOptions {
+                zero_initialize_workgroup_memory: false,
+                ..PipelineCompilationOptions::default()
```
DasLixou, line 649: Do we have a policy of whether using `..Default::default()` or `..TheWholeLongStructName::default()` ?

raphlinus: I personally go for the former.

Also, as a point of curiosity, less noisy default syntax was one of the things discussed at the unconference for how the Rust language could better support UI and application programming.

DasLixou: Only thing I know is that there was a [rust internals discussion](https://internals.rust-lang.org/t/add-default-keyword/20810/13) about a `default` keyword like in C#

DasLixou, line 649:
```suggestion
                ..Default::default()
```

DasLixou: oh wait I can also put it here :0 github actually surprises me sometimes
```suggestion
                ..Default::default()
```

DJMcNab (approved):
I have the `Default::default` version running on my phone with no issues and the expected performance improvement

So once we apply that suggestion, I'm happy for this to land.

waywardmonkeys:
Used the proposed change and set merge when ready as I am in bed to sleep.

PR #579, opened 2024-05-14
Reference `vello` via workspace dep.

waywardmonkeys:
This makes `vello` the same as the other crates like `vello_encoding` and `vello_shaders`.

This change isn't done for `examples/simple/Cargo.toml` as changing that `path` to a `workspace` results in a warning from cargo.


xStrom (approved):
Makes sense, thanks!

waywardmonkeys:
@xStrom I'd done some patches (locally) that moved the crates around for the new standard and this part of that annoyed me, so I figured it was better to do this as a first step.

PR #582, opened 2024-05-16
tests: Support check-cfg for skip_gpu_tests.

waywardmonkeys:


xStrom (approved):
[Neat.](https://blog.rust-lang.org/2024/05/06/check-cfg.html) Thanks!

PR #584, opened 2024-05-16
[shaders] More graceful error handling

armansito:
The transition of the vello crate to use the vello_shaders crate regressed developer ergonomics around compilation failures, in which compilation failures always caused a panic and the displayed debug-formatted error message wasn't pretty printed. The panic was particularly bad with hot-reload since it effectively undid its usefulness by crashing the whole process on shader misspellings.

This change alleviates those regressions by adding better display formatting to WGSL parse errors returned from the shaders crate and propagating errors up to clients instead of always panicking. The Debug formatting is still a little wonky so call-sites are currently responsible for formatting errors with Display to get a more readable compiler message. This PR only updates `with_winit` with pretty-printing.

### DIFF
--- a/vello_shaders/src/compile/mod.rs
+++ b/vello_shaders/src/compile/mod.rs
@@ -232,7 +232,7 @@ impl ShaderInfo {
                         }
                     } else {
                         let source = preprocess::preprocess(&contents, &defines, &imports);
-                        match Self::new(&shader_name, source, "main") {
+                        match Self::new(shader_name, source, "main") {
                             Ok(shader_info) => {
                                 info.insert(shader_name.to_string(), shader_info);
                             }
--- a/vello_shaders/src/compile/mod.rs
+++ b/vello_shaders/src/compile/mod.rs
@@ -19,15 +19,25 @@ use crate::types::{BindType, BindingInfo, WorkgroupBufferInfo};
 
 #[derive(Error, Debug)]
 pub enum Error {
-    #[error("failed to parse shader: {0}")]
-    Parse(#[from] wgsl::ParseError),
+    #[error("failed to parse shader ({name}): {msg}")]
+    Parse {
+        name: String,
+        msg: String,
+        error: wgsl::ParseError,
+    },
 
-    #[error("failed to validate shader: {0}")]
-    Validate(#[from] WithSpan<ValidationError>),
+    #[error("failed to validate shader ({name}): {msg}")]
+    Validate {
+        name: String,
+        msg: String,
+        error: WithSpan<ValidationError>,
+    },
 
     #[error("missing entry point function")]
     EntryPointNotFound,
 }
+
+pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug)]
 pub struct ShaderInfo {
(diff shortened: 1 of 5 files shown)


crates/shaders/build.rs
```diff
@@ -28,7 +28,13 @@ fn main() {
         .and_then(|p| Path::new(&p).parent().map(|p| p.to_owned()))
         .unwrap_or(PathBuf::from("../../"));
     let shader_dir = Path::new(&workspace_dir).join("shader");
-    let mut shaders = compile::ShaderInfo::from_dir(shader_dir);
+    let mut shaders = match compile::ShaderInfo::from_dir(shader_dir) {
+        Ok(s) => s,
+        Err(err) => {
```
DJMcNab, line 33: In theory, we should coalesce issues from multiple shaders.

In practise, you'll only be working on one shader at a time, so it should be fine. I guess if you modify an include in an incompatible way, this could be an issue

crates/shaders/src/compile/mod.rs
```diff
@@ -18,8 +18,12 @@ use crate::types::{BindType, BindingInfo, WorkgroupBufferInfo};
 
 #[derive(Error, Debug)]
 pub enum Error {
-    #[error("failed to parse shader: {0}")]
-    Parse(#[from] wgsl::ParseError),
+    #[error("failed to parse shader ({name}): {msg}")]
```
DJMcNab, line 21: It would be really nice if we could point to file and line. I understand why we can't (because we'd effectively need to handle source maps), but it would be good for DX.

OTOH, as it's only "internal" DX, that's not that important

DJMcNab (commented):
I have not run this myself - I'll try and so on Monday if this hasn't been otherwise approved.

This looks like it should be a good improvement, although the requirement to print that way is very unfortunate. I feel like we should just be avoiding unwrap here, if the data will be duplicated and repeated.

An example output from a failure would be good to have for reviewing

crates/shaders/src/compile/mod.rs
```diff
@@ -18,8 +18,12 @@ use crate::types::{BindType, BindingInfo, WorkgroupBufferInfo};
...
+    Parse {
+        name: String,
+        msg: String,
+        error: wgsl::ParseError,
+    },
 
     #[error("failed to validate shader: {0}")]
     Validate(#[from] WithSpan<ValidationError>),
```
DJMcNab, line 29: The error message for validation errors is unfortunately terrible, though. Again, not really a job for this PR

DJMcNab (approved):
Sorry for the delay.

This looks good and is important. I would delay this until after vello#590 then rebase onto that.

crates/shaders/src/compile/mod.rs
armansito: Fortunately `emit_to_string` prints out line numbers (so `msg` contains them) and `{name}` here is will match the name of the pipeline being compiled, e.g.:

```
Failed to create renderer: Failed to compile shader: failed to parse shader (pathtag_reduce2): error: expected identifier, found '32'
   ┌─ wgsl:87:14
   │
87 │     path_ix: 32,
   │              ^^ expected identifier
```

That said the line numbers don't actually line up with the real files because following the preprocess step and won't contain any info about the included files. So yeah, we'd need some type of source mapping (which is beyond the scope of this PR).

crates/shaders/src/compile/mod.rs
armansito: Done. This will now result in an error like this on validation:

```shell
Failed to create renderer: Failed to compile shader: failed to validate shader (flatten): error: Function [30] 'es_seg_flatten_evolute' is invalid
    ┌─ wgsl:673:1
    │
673 │ ╭ fn es_seg_flatten_evolute(
674 │ │     es: EulerSeg,
675 │ │     path_ix: u32,
676 │ │     transform: Transform,
677 │ │     chord_len: f32,
678 │ │     tol: f32,
679 │ │ ) -> vec2f {
    │ ╰────────────^ naga::Function [30]
    │
    = The `return` value None does not match the function return value
```

vello_shaders/src/compile/mod.rs
```diff
@@ -17,18 +18,62 @@ pub mod msl;
...
+            msg: source.emit_msg(wgsl),
+            source,
+        }
+    }
+}
+
+impl InnerError {
+    fn emit_msg(&self, wgsl: &str) -> String {
+        match self {
+            Self::Parse(e) => e.emit_to_string(wgsl),
```
DJMcNab, line 70: Can we use `emit_to_string_with_path` here? The line numbers will still be wrong, but it means that the string won't just be `"wgsl"`

DJMcNab (approved):
Thanks for getting these improvements in; this looks like it should be an improvement in DX and restore the hot reloading properly.

The source mapping issue is a pain, but it's not important enough to fix as we have no external users.

vello_shaders/src/compile/mod.rs
armansito: Done.

PR #586, opened 2024-05-23
Remove download command, keep download directory.

waywardmonkeys:
This still allows loading from the download directory, but removes the command that would download to it, allowing people to populate it themselves.

The files downloaded were parsed with the `vello_svg` parser, but that was removed from here and moved to its own home, so files loaded would need some preprocessing (to be defined elsewhere).


DJMcNab (approved):
Is it worth changing the name of the `downloads` folder (e.g. to `svgs`?)

I don't think downloads is quite the right name anymore, although I'm struggling to give a compellingly better suggestion.

waywardmonkeys:
Going to merge this for now and open an issue about 1) renaming it and 2) documenting what needs to be done to have an SVG load from here (the preprocessing to let pico-svg understand it).

PR #587, opened 2024-05-26
clippy: Fix new `doc_lazy_continuation` lint

waywardmonkeys:


DJMcNab (approved):
Is there a reason to not also add this to the lints table? If it's enabled by default, it should have been caught by ci on https://github.com/linebender/vello/pull/580

waywardmonkeys:
It is default in `nightly` and was added in 1.80, so it isn't available yet in `stable`.

Issue #548, opened 2024-04-12
Full system hang on Apple M1 8GB


jrmoulton:
I'm experimenting with swapping in Vello as the renderer for [floem](https://github.com/lapce/floem) and I'm running into an issue where, when using Vello, 
- my entire system will hang
- everything but the mouse is unresponsive
- within about a minute the screen and mouse will also become unresponsive (screen dark with only dim backlight showing)
- the system requires a reboot. 

I get this issue in the `with_winit` example in the Vello examples and also in the `editor` example in floem. 

In the with_winit example it will happen when I zoom in too far into the Ghostscript tiger. I have noticed that there is a limit to the amount that I can zoom in and this causes some stuttering but this is separate from when it becomes unresponsive. 

In the `editor` example the issue is caused when I delete several characters from the starting text in the editor. 

Prior to hanging macOS activity monitor doesn't indicate high memory pressure. 

The issue isn't consistent (and reproducing takes a long time) but it does happen regularly (within 30 seconds of doing the above listed actions). 

Apple M1 MacBook Air
8GB RAM
Sonoma 14.2.1 (23C71)

bram209:
I had the same issue when aggressively zooming in and out. No stuttering, but suddenly the whole system hangs. I am on a M2 Pro 32GB

raphlinus:
We'll look into it. A good way of isolating this is to turn on use_cpu, then edit the code to set `force_gpu_from` to various values. That said, when doing aggressive zooming, it's most likely that there will be a panic from buffer overflow - one of the things to do soon is to chance the CPU shaders to match the behavior of the GPU shaders (report a failure and continue).

armansito:
I ran into this with some scene content today. I'll investigate.

DJMcNab:
Hi @jrmoulton. We think that #551 or #553 might have fixed this issue. Does this still happen for you on main?

raphlinus:
I'm able to repro this fairly straightforwardly on M1. There are two separate issues, both of which need to be fixed.

The first is that zooming in causes unboundedly large memory usage, specifically in flatten. This is because it's flattening all outlines, using the transform to determine the number of subdivisions. What needs to happen is culling the flattened line segments to the viewport. That's also related to vello#542.

[...]

PR #589, opened 2024-05-29
Workaround to avoid full hang

raphlinus:
As the scale factor becomes unreasonably large, the number of subdivisions of an Euler spiral into lines grows, making the flatten stage take too long.

This patch just bounds the number of subdivisions, so it will eventually make progress. It will be slow, but not hang.

A better solution would be to aggressively cull so it only generates geometry inside the viewport, but that is considerably more complicated.

Workaround for #548


waywardmonkeys:
Given the amount of pain involved in this, I think that maybe having a comment in one of the implementations would be reasonable to explain that this particular `clamp` is important and a bit about why...

xStrom (approved):
Not hanging the system is good!

PR #590, opened 2024-05-30
Migrate the repository to a virtual workspace.

xStrom:
The CI, VSCode config, `ARCHITECTURE.md` and some `README.md` links have been updated to reflect the new directory structure.

The WGSL shaders now live in the `vello_shaders` subdirectory. Vello hot reloading depends on our repo structure, which is fine as decided at last office hours due to this being a development feature.

The examples continue to live in a root `examples` directory, as that is how it remained for [Parley](https://github.com/linebender/parley) as well. I did this for consistency and don't really feel strongly either way if people think we should move them under `vello/examples` instead. If we had examples without `Cargo.toml` that would make more sense, but we don't, so ..

The repo `README.md` and the `vello` sub-directory `README.md` are exact clones and both need updating. That is intentionally planned as follow-up work to reduce the review friction here.


vello/src/shaders.rs
```diff
@@ -62,7 +62,7 @@ pub fn full_shaders(
     #[cfg(feature = "hot_reload")]
     let mut shaders = vello_shaders::compile::ShaderInfo::from_dir(concat!(
```
DJMcNab, line 63: Can `vello_shaders` instead export the directory the shaders are located in?

xStrom: You mean the `shader` directory name, or some more absolute path?

xStrom: Alright I added a new function that returns an absolute path to the shader directory. It depends on using Cargo and the path is computed at compile time.

If someone has a build cache and then moves the directories around, they will need to `cargo clean` or trigger a rebuild in another way.

vello_shaders/src/compile/mod.rs
```diff
@@ -198,3 +204,11 @@ fn postprocess(wgsl: &str) -> String {
     }
     output
 }
+
+/// Returns the absolute path to the directory containing the WGSL shaders.
+///
+/// The path is determined at compile time.
+pub fn shader_dir() -> &'static PathBuf {
+    static SHADER_DIR: OnceLock<PathBuf> = OnceLock::new();
+    SHADER_DIR.get_or_init(|| Path::new(env!("CARGO_MANIFEST_DIR")).join("shader"))
```
DJMcNab, line 213: If we used `concat!`, I don't think the runtime string manipulation would be needed.

I guess this is about using the proper separator?

vello_shaders/src/compile/mod.rs
```diff
@@ -198,3 +204,11 @@ fn postprocess(wgsl: &str) -> String {
     }
     output
 }
+
+/// Returns the absolute path to the directory containing the WGSL shaders.
+///
+/// The path is determined at compile time.
```
DJMcNab, line 210: And may be invalid at runtime, especially if the program is deployed on a different machine

As a different point:
This is also a reproducible builds challenge, which I guess we should document?
Idk, maybe even making it `doc(hidden)`

vello_shaders/src/compile/mod.rs
xStrom: Just using `concat!` is buggy because we don't know if `CARGO_MANIFEST_DIR` ends with a separator or not. Instead of doing some sort of `.ends_with` dance and potentially missing some other edge cases, better to use the `path` module.

vello_shaders/src/compile/mod.rs
xStrom: Sure, I added a note about the impact on reproducible builds.

vello/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+```shell
+cargo run -p with_winit
+```
+
+![image](https://github.com/linebender/vello/assets/8573618/cc2b742e-2135-4b70-8051-c49aeddb5d19)
+
+It is used as the rendering backend for [Xilem], a Rust GUI toolkit.
+
+> [!WARNING]
```
DJMcNab, line 31: We should change the formatting of this now, to use something which renders well on crates.io

That might just be using the ⚠️ emoji

DJMcNab (commented):
Will give a final check on Monday if this hasn't landed by then

vello/README.md
xStrom: Updating readme files is intentionally out of scope and planned as follow-up work to reduce the review friction here.

vello/LICENSE-APACHE
```diff
@@ -0,0 +1,202 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
DJMcNab, line 179: I don't think we need this appendix here - see e.g. https://github.com/rust-lang/rust/pull/67734

README.md
DJMcNab, line 1: I'd add a big comment at the top that this is "temporarily" duplicated in the `vello` folder, and updates should be preferred there.

vello/README.md
```diff
@@ -0,0 +1,296 @@
...
+Licensed under either of
+
+- Apache License, Version 2.0
+   ([LICENSE-APACHE](LICENSE-APACHE) or <http://www.apache.org/licenses/LICENSE-2.0>)
+- MIT license
+   ([LICENSE-MIT](LICENSE-MIT) or <http://opensource.org/licenses/MIT>)
+
+at your option.
+
+In addition, all files in the [`vello_shaders/shader`](https://github.com/linebender/vello/tree/main/vello_shaders/shader) and [`vello_shaders/src/cpu`](https://github.com/linebender/vello/tree/main/vello_shaders/src/cpu)
```
DJMcNab, line 277: These additional caveats are not important true for `vello` itself. I imagine this has the same reasoning as above though?

vello_shaders/src/compile/mod.rs
```diff
@@ -129,6 +130,11 @@ impl ShaderInfo {
         })
     }
 
+    /// Same as [`ShaderInfo::from_dir`] but uses the default shader directory provided by [`shader_dir`].
+    pub fn from_default() -> HashMap<String, Self> {
```
DJMcNab, line 134: Maybe `from_builtin`/`from_included` is a better name?

We should also point to the `SHADERS` constant as preferred by most users

vello_shaders/src/compile/mod.rs
```diff
@@ -198,3 +204,12 @@ fn postprocess(wgsl: &str) -> String {
     }
     output
 }
+
+/// Returns the absolute path to the directory containing the WGSL shaders.
+///
+/// The path is determined at compile time.
```
DJMcNab, line 210:
```suggestion
/// The path is determined at compile time and is only valid on the compiling machine.
```

DJMcNab (approved):
Thanks for getting this done. A few small nits, I don't think any are necessarily blocking

PR #591, opened 2024-05-31
Re-export wgpu

### LINKED ISSUE #585: Can we re-export `wgpu` with the feature?
I didn't realize until now that the recent `wgpu` feature given to Vello doesn't re-export the version, e.g. `vello::wgpu::<what I need>`

Could we re-export this with the feature flag on?

Particularly, some of the examples in velato/vello_svg use `wgpu` types, and I'd prefer not to keep separately maintained versions in sync.
image

tannal:
These changes first re-export wgpu under the vello module.
Then, all examples are updated to use vello::wgpu.

Fixes #585

### DIFF
--- a/examples/headless/src/main.rs
+++ b/examples/headless/src/main.rs
@@ -10,11 +10,11 @@ use clap::Parser;
 use scenes::{ImageCache, SceneParams, SceneSet, SimpleText};
 use vello::kurbo::{Affine, Vec2};
 use vello::util::RenderContext;
-use vello::{block_on_wgpu, RendererOptions, Scene};
-use wgpu::{
-    BufferDescriptor, BufferUsages, CommandEncoderDescriptor, Extent3d, ImageCopyBuffer,
+use vello::wgpu::{
+    self, BufferDescriptor, BufferUsages, CommandEncoderDescriptor, Extent3d, ImageCopyBuffer,
     TextureDescriptor, TextureFormat, TextureUsages,
 };
+use vello::{block_on_wgpu, RendererOptions, Scene};
 
 fn main() -> Result<()> {
     #[cfg(not(target_arch = "wasm32"))]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -951,7 +951,6 @@ dependencies = [
  "pollster",
  "scenes",
  "vello",
- "wgpu",
 ]
 
 [[package]]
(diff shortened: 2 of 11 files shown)


DJMcNab (commented):
Thanks for your interest. The actual changes look like a fine implementation of this.
I'm somewhat conflicted on whether we want this. There are two classes of projects which we expect to use Vello:
1) Ones which use Vello for all of their rendering
2) Ones which embed Vello within an existing wgpu application

Of these, option 1 probably should want this re-export, but option 2 would I think actively not want to use this re-export. And having this re-export means that it becomes harder to maintain consistency in which import path is used.
I think we probably will end up with this, but the lack of tooling support means that it does have the potential for negative consequences.

xStrom (commented):
Please rebase on `main` to resolve the merge conflict.

examples/headless/src/main.rs
```diff
@@ -10,11 +10,11 @@ use clap::Parser;
 use scenes::{ImageCache, SceneParams, SceneSet, SimpleText};
 use vello::kurbo::{Affine, Vec2};
 use vello::util::RenderContext;
-use vello::{block_on_wgpu, RendererOptions, Scene};
-use wgpu::{
+use vello::wgpu::{
     BufferDescriptor, BufferUsages, CommandEncoderDescriptor, Extent3d, ImageCopyBuffer,
```
xStrom, line 14: Here we can add the `self` import.

```suggestion
use vello::wgpu::{
    self, BufferDescriptor, BufferUsages, CommandEncoderDescriptor, Extent3d, ImageCopyBuffer,
```

xStrom (commented):
The re-export is fine.

I don't like the increased verbosity of the example code though. Let's keep the code unchanged and instead add the required `use vello::wgpu;` statements to the files that need them.

xStrom:
I see you removed all changes to examples so the added re-export isn't actually used anywhere. Is this intentional or would you be willing to still remove the `wgpu` dependencies where possible?

Specifically, the `headless`, `simple`, `with_winit`, and `vello_tests` packages have `wgpu` in their `Cargo.toml` files that could be removed. Also the line *"as well as examples/simple/Cargo.toml"* comment line removed from above the `wgpu` specification in the workspace `Cargo.toml`.

tannal:
It's not intentional, I will remove those wgpu dependencies. 
I think I just need to import `vello::wgpu` in those files so that the import path is not changed.

xStrom (approved):
Excellent, thank you for getting this done!

PR #593, opened 2024-05-31
deps: Update and sync direct dep versions

waywardmonkeys:
We now state that we depend on the same version of direct dependencies as is used in the `Cargo.lock`.

Also, update to the latest versions of our direct dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -720,9 +720,9 @@ dependencies = [
 
 [[package]]
 name = "font-types"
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
I'm guessing this is with 0.2 in mind?

waywardmonkeys:
> I'm guessing this is with 0.2 in mind?

Yep… get most things in place…

PR #594, opened 2024-05-31
Update `CHANGELOG.md`

waywardmonkeys:


DJMcNab (approved):
I have not validated that this is every change we want to include, but these look like a good set of additions.

Thanks!

PR #595, opened 2024-05-31
examples: Use `web-time` instead of `instant`

waywardmonkeys:
Recently, `instant` has been marked as unmaintained by the maintainer. A suggested replacement is `web-time`, which is used by `winit` and is already in our dependency tree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1047,18 +1047,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
This conflicts with both #590 and #593, but I'm happy to rebase as needed.

DJMcNab (approved):
Thanks

PR #599, opened 2024-06-06
vello_shaders: Rename use of `gen`

waywardmonkeys:
This becomes a reserved word in the 2024 edition.


waywardmonkeys:
https://doc.rust-lang.org/nightly/edition-guide/rust-2024/gen-keyword.html

DJMcNab (approved):
That module isn't public, right?

waywardmonkeys:
Correct. The module is private and the`SHADERS` from it is re-exported.

PR #602, opened 2024-06-06
Update `README.md` files.

xStrom:
I moved around the various extra license info to correctly reflect the new project structure.

In the `vello` readme I replaced the GitHub specific markdown syntax with emojis for better crates.io visuals. I also edited some text to comply with one-sentence-per-line. The directory references got updated to reflect reality.

The repo level readme got a comment noting that it needs restructuring. I also synced some changes I made in `vello/README.md` to the text that is still present in the repo readme.

The `vello_shaders` and `vello_encoding` packages get community & license sections.

### DIFF
--- a/vello/README.md
+++ b/vello/README.md
@@ -108,7 +108,7 @@ renderer
 surface_texture.present();
 ```
 
-See the [`../examples`](https://github.com/linebender/vello/tree/main/examples) directory to see how that code integrates with frameworks like winit and bevy.
+See the repository's [`examples`](https://github.com/linebender/vello/tree/main/examples) directory for code that integrates with frameworks like winit and bevy.
 
 ## Performance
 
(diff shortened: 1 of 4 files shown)


vello/README.md
```diff
@@ -107,7 +108,7 @@ renderer
 surface_texture.present();
 ```
 
-See the [`examples/`](https://github.com/linebender/vello/tree/main/examples) folder to see how that code integrates with frameworks like winit and bevy.
+See the [`../examples`](https://github.com/linebender/vello/tree/main/examples) directory to see how that code integrates with frameworks like winit and bevy.
```
DJMcNab, line 111: I think this probably likely to be more/just as confusing - I'd suggest maybe:
```suggestion
The [`examples`](https://github.com/linebender/vello/tree/main/examples) folder in the repository shows that code integrated with frameworks like winit and bevy.
```

The specific issue is that this README is intended to be read on crates.io and docs.rs, with the GitHub use cases only peripherally

vello/README.md
```diff
@@ -232,13 +236,13 @@ cargo update -p package_name --precise 0.1.1
...
+All public content can be read without logging in.
 
-Contributions are welcome by pull request. The [Rust code of conduct] applies.
+Contributions are welcome by pull request.
+The [Rust code of conduct] applies.
 
-Unless you explicitly state otherwise, any contribution intentionally submitted
-for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
-licensed as noted in the "License" section, without any additional terms or conditions.
+Unless you explicitly state otherwise, any contribution intentionally submitted for inclusion in the work by you, as defined in the Apache 2.0 license, shall be licensed as noted in the [License](#license) section, without any additional terms or conditions.
```
DJMcNab, line 245: For android_trace, I excluded this sentence from the crate level README, as it is only really meaningful at a repository level.

DJMcNab (approved):
Some good and necessary improvements here.

I've got a few thoughts on small details which I'd prefer to see changed, but the proposed versions are also entirely reasonable.

vello/README.md
xStrom: I simplified it and removed the confusing dots.

vello/README.md
xStrom: Crates can have different licenses, so there is some potential value there. I don't care about this specific sentence much, because it doesn't change anything legally, and the same sentence is already present in the Apache license.

I guess I'm fine with only having it in the repo level readme.

vello/README.md
```diff
@@ -28,7 +28,8 @@ cargo run -p with_winit
 
 It is used as the rendering backend for [Xilem], a Rust GUI toolkit.
 
-> [!WARNING]
+> ⚠️ WARNING
+>
 > Vello can currently be considered in an alpha state. In particular, we're still working on the following:
```
DJMcNab, line 33: Is this release good enough to call beta? I think probably not, but it's plausible, so I thought I'd check

vello/README.md
```diff
@@ -183,7 +187,8 @@ The [`with_winit`](#winit) example supports running on Android, using [cargo apk
 cargo apk run -p with_winit
 ```
 
-> [!TIP]
+> 💡 TIP
+>
```
DJMcNab, line 191: It might be worth removing the web/android/with_winit detailed running instructions from this crate-level readme

I think that should be a post-release follow-up, probably.

vello/README.md
xStrom: Probably not, pending some more robustness work.

PR #604, opened 2024-06-06
Update `CHANGELOG.md` in preparation for v0.2.0.

xStrom:
I quickly went over the commits and added some things that are of some relevance to consumers of Vello.

I also switched to the more compact markdown format that we used in [Druid](https://raw.githubusercontent.com/linebender/druid/master/CHANGELOG.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 # Changelog
 
-The latest published Vello release is [0.2.0](#020---2024-06-07) which was released on 2024-06-07.
-You can find its changes [documented below](#020---2024-06-07).
+The latest published Vello release is [0.2.0](#020---2024-06-08) which was released on 2024-06-08.
+You can find its changes [documented below](#020---2024-06-08).
 
 ## [Unreleased]
 
@@ -19,7 +19,7 @@ You can find its changes [documented below](#020---2024-06-07).
 
 ### Fixed
 
-## [0.2.0] - 2024-06-07
+## [0.2.0] - 2024-06-08
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


waywardmonkeys (approved):

CHANGELOG.md
```diff
@@ -1,37 +1,99 @@
...
+- Euler spiral based stroke expansion. ([#496] by [@raphlinus])
+- Sweep gradients. ([#435] by [@dfrg])
+- Bump allocation estimation. ([#436], [#454], [#522] by [@armansito])
+- Impl `From<Encoding>` for `Scene`. ([#538] by [@waywardmonkeys])
+- Glyph hinting support. ([#544] by [@dfrg])
+- Better glyph caching. ([#555] by [@dfrg])
+- `vello_shaders` crate to load and preprocess WGSL. ([#563] by [@armansito])
+- Coverage-mask specialization. ([#540] by [@armansito])
+- Support for the `#enable` post-process directive. ([#550] by [@armansito])
+- `wgpu-profiler` feature. ([#502] by [@DJMcNab])
```
DJMcNab, line 35: As far I can remember, this is localised to `with_winit`; not really something which has been "added" in a changelog sense

Is it worth making an `examples` header or similar?

xStrom: The feature also exists in `vello`, but apparently didn't work properly before, because you modified `vello` code in #502. So could move it under the fixed section, but because it didn't really work before, this is really the first release that has it.

As for an examples header, I'm leaning towards nope. In Druid we documented a bunch of stuff, including examples, in the changelog. However I think that leads to too much noise and so now I'm thinking that the changelog should focus more strictly on changes that directly impact an existing consumer.

DJMcNab: I believe it must have worked in 0.1.0 (see e.g. Xilem), but then regressed then unregressed

That is, I still don't think that this pr changed anything in the change from 0.1.0 to 0.2.0

xStrom: Alright, I'll remove it then.

PR #605, opened 2024-06-06
Bump Vello version to 0.2.0.

xStrom:
Also enable publishing for `vello_shaders`.


DJMcNab (approved):
Will be good to get this released

PR #609, opened 2024-06-10
[pico_svg] Encode group transform in scene

raphlinus:
Build the scene recursively, applying the transform at group nodes, rather than flattening to a single list of items. This causes the transform to be applied correctly to stroke styles, which among other things fixes the rendering of the waves example from the Nehab timings data set.

We're deliberately keeping things minimal, but if we were to add other properties to groups (clips, opacity, etc), then this change would support that.

### DIFF
--- a/examples/scenes/src/svg.rs
+++ b/examples/scenes/src/svg.rs
@@ -71,6 +71,38 @@ fn example_scene_of(file: PathBuf) -> ExampleScene {
     }
 }
 
+fn render_svg_rec(items: &[crate::pico_svg::Item]) -> Scene {
+    let mut scene = Scene::new();
+    for item in items {
+        use crate::pico_svg::Item;
+        match item {
+            Item::Fill(fill) => {
+                scene.fill(
+                    Fill::NonZero,
+                    Affine::IDENTITY,
+                    fill.color,
+                    None,
+                    &fill.path,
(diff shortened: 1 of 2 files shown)


examples/scenes/src/svg.rs
```diff
@@ -71,6 +71,38 @@ fn example_scene_of(file: PathBuf) -> ExampleScene {
     }
 }
 
+fn render_svg_rec(items: &[crate::pico_svg::Item]) -> Scene {
```
DJMcNab, line 74: I don't really like this function name; I'm not familiar with this `_rec` naming scheme being used.

Unfortunately, I don't have a better suggestion at the moment, so it's probably fine to keep?

examples/scenes/src/svg.rs
```diff
@@ -71,6 +71,38 @@ fn example_scene_of(file: PathBuf) -> ExampleScene {
...
+    let mut scene = Scene::new();
```
DJMcNab, line 75: It's probably premature optimisation to re-use a single child scene for the top-level of each child, to cut down on the number of allocations slightly.

DJMcNab (approved):
I've not tested this on the scene which sees an improvement, but this is only in example code, so it should be fine

PR #610, opened 2024-06-11
Add some snapshot test infrastructure

### LINKED ISSUE #608: Snapshot testing
We want to ensure that Vello reliably renders as expected, even in the face of potentially significant refactors (e.g. #574), and with unusual scenes (e.g. #542). Additionally, as our pipeline changes (e.g. #607), we need to ensure that the CPU shaders keep up-to-date (especially if and/or when we get #485). This requires automated testing.

Some groundwork was laid in #439, which got us running on an emulated Mesa GPU on Linux, and physical M1 GPUs on macOS. However, the tests added there are extremely simple, doing property based testing on a small handful of scenes. We also want to do snapshot testing. This would involve snapshots of "hand-written" scenes, such as to emulate a [masonry](https://github.com/linebender/xilem) scene or to include tricky cases, and a number of SVG scenes. The canonical results for these testst would be stored in this repository [^1]. We discovered in https://github.com/linebender/xilem/pull/233 that our rendering isn't completely stable cross-platform (presumably due to fast math), so we need to compare the images using something like [nv-flip](https://crates.io/crates/nv-flip), as done by wgpu.

There are some additional kinds of tests which can be performed. These are longer-term plans, and should have their own issues if and/or when we decide to do them. These include:
- [ ] Comparative tests against other renderers (skia, tiny-skia?, webrender?, cocoa?)
- [ ]  It would also be possible to use diffing against previous CI artefacts. This is what bevy is exploring in https://github.com/bevyengine/bevy/pull/13248. This would allow using significantly more tests, but we would need to set up infrastructure for comparing and approving these images (as Bevy's infrastructure is closed and proprietary). 
- [x] Comparing the CPU and GPU pipelines, removing the need to store an on-disk snapshot

[^1]: We might be better off doing something like [piet-snapshots](https://github.com/linebender/piet-snapshots) or `git lfs`. We get 1GiB of bandwidth a month of lfs, which might be a problem for CI. piet-snapshots is less likely to run into this, because we get free bandwidth. See e.g. https://estebangarcia.io/how-we-saved-on-github-lfs-bandwidth/

DJMcNab:
Fixes #608 

This does not add a wide gamut of tests - I imagine that the `test_scenes` will provide some good fodder here.
There are some quality-of-life features included:

- You can use `VELLO_TEST_CREATE=all` to create new tests. It is recommended to set this on your local machine (and it won't be set on CI)
- You can use `VELLO_TEST_UPDATE=all` to update with new contents if the test fails. (Note that if there's a minor change which passes the test, this won't update)

These variables can also filter to specific tests if needed.
The saved snapshots are always created using the GPU renderer, but it is set up to be easy to use the CPU renderer for tests as well.

Using `cargo nextest` for running tests is recommended; indeed, it seems to be now required, as the tests segfault without it


vello_tests/Cargo.toml
```diff
@@ -16,3 +16,7 @@ anyhow = { workspace = true }
 pollster = { workspace = true }
 png = "0.17.13"
 futures-intrusive = { workspace = true }
+nv-flip = "0.1.2"
+image = { workspace = true, features = ["png"] }
+
+scenes = { workspace = true }
```
PoignardAzur, line 22: If you're adding a dependency to the `scenes` example, you might at well make it into a full `vello_test_scenes` root crate. Having a crate depend on an example is somewhat confusing.

PoignardAzur (commented):
Okay, so I might need to look deeper into this code, but my takeaway after a first pass through is that it seems like the test code has evolved into something of a labyrinth, and documentation would be welcome.

It feels like some of the confusing architecture I saw existed prior to this PR, though, so I'm not sure how much of a fuss to make about it now.

vello_tests/Cargo.toml
DJMcNab: I'm hesitant, mostly because of the assets directory. Specifically, the issue is:
1) `vello_test_scenes` makes use of the `examples/assets`.
2) I think it's clearer to have `assets` under `examples`, because that's the main route through which users use them
3) I don't think a top-level `vello_test_scenes` should be using assets in the `examples/assets` folder.

vello_tests/tests/simplest.rs
```diff
@@ -0,0 +1,79 @@
...
+    let params = TestParams {
+        use_cpu,
+        ..TestParams::new("filled_square", 20, 20)
+    };
+    match snapshot_test_sync(scene, &params)
+        .and_then(|mut snapshot| snapshot.assert_mean_less_than(0.01))
+    {
+        Ok(()) => (),
+        Err(e) => panic!("{:#}", e),
+    }
```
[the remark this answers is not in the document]
DJMcNab: Hmm, yeah. Looking into it, `anyhow::Error` does the right thing with unwrap. That's good to know.

vello_tests/src/lib.rs
```diff
@@ -129,35 +135,44 @@ pub async fn render(scene: Scene, params: &TestParams) -> Result<Image> {
...
     writer.write_image_data(image.data.data())?;
     writer.finish()?;
-    println!("Wrote result ({width}x{height}) to {out_path:?}");
-
     Ok(())
 }
 
-pub fn should_debug_png(name: &str, use_cpu: bool) -> bool {
-    if let Ok(val) = env::var("VELLO_DEBUG_TEST") {
+pub fn env_var_relates_to(env_var: &'static str, name: &str, use_cpu: bool) -> bool {
```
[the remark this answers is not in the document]
DJMcNab: I've made it private, and added a small comment

vello_tests/Cargo.toml
PoignardAzur: What's wrong with having assets in vello_test_scenes?

DJMcNab: - The existing downloads folder is under `examples/assets` and is `.gitignore`d, which means that moving that folder will require manual work from anyone who has been using the downloads folder
- In my opinion, `examples/assets` is a much more natural place to find assets used by the examples than in a test crate - most users who care about examples.
- I would expect users looking for example uses of Vello would be much more likely to look under examples than in a test support crates. Although our entire setup isn't really great for people wanting to find examples.
- I also don't really see that it's worth the noise of having an extra top-level crate for it.

I don't really have a *strong* position on this, but I think leaving it as-is is the least bad option. Happy to change it if you want to impose a different way though :)

DJMcNab:
@PoignardAzur I think this is blocked on your review, except the one open question (on which my position is that I don't think it needs to change, but will move if there's strong pushback)

PoignardAzur (approved):

PoignardAzur:
Sorry, I thought I had done the re-review.

PR #612, opened 2024-06-13
Remove the `Pipelines` API

DJMcNab:
This has a few issues:
- Pipelines cannot be optional, and so all pipelines will be compiled even if unused. This includes all 6 variants of `fine`
- It cannot support multi-threaded initialisation
- It has, as far as I can tell, never been used (based on a search of all of GitHub, including [google/skia](https://github.com/google/skia/))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ You can find its changes [documented below](#020---2024-06-08).
 
 ### Fixed
 
+## Removed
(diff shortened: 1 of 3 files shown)


dfrg (approved):
I still think the concept of generating pipeline instantiations for the shader set is useful but this particular implementation was designed for a different time so clearing it out makes sense.

PR #613, opened 2024-06-13
examples: Use `clamp`

waywardmonkeys:
This satisfies a clippy lint that is newly enabled in 1.79.


DJMcNab (approved):

xStrom:
Aha, the return of `clippy::manual_clamp`. Looks like it's much more conservative this time around and works on const values only.

Last time it was enabled (in 1.66) it suggested code changes that were not equivalent and even caused panics, e.g. the following code [in Druid](https://github.com/linebender/druid/pull/2317):
```rust
#[allow(clippy::manual_clamp)] // Usable space could be below the minimum bar size.
let length = length.max(bar_min_size).min(usable_space);
// e.g. length.max(6.0).min(3.0) would be .clamp(6.0, 3.0) which panics
```

PR #615, opened 2024-06-18
Render COLR Glyphs (Emoji)

### LINKED ISSUE #536: Trouble drawing emojis
I'm trying to render emojis using `winit`. I've gotten normal text to render just fine following along from [here](https://gist.github.com/dwuggh/cfc290bcfdb7192582e350b432d097d5). I'm using an emoji font (I know it has emojis since the `charmap.map(c)` is returning a nonzero `gid`, and is returning a zero `gid` for normal ascii). The following code I think provides enough context to understand what I'm doing:

```rust
scene
    .draw_glyphs(&self.font)
    .font_size(font_size)
    .transform(Affine::translate((110.0, 700.0)))
    .glyph_transform(None)
    .draw(
        peniko::Fill::NonZero,
        text.chars().map(|c| {
            let x = pen_x;
            let gid = charmap.map(c).unwrap_or_default();
            pen_x += metrics.advance_width(gid).unwrap_or_default();
            vello::glyph::Glyph {
                id: gid.to_u16() as u32,
                x,
                y: pen_y,
            }
        }),
    );
```

The `text` has the emojis we're rendering from. In my case I'm using "abc🪓123" which turns into GID_0, GID_0, GID_0, GID_3750, GID_3192, GID_3198, GID_3196, respectively. The `charmap` holds the charmap from a font_ref from the font. I'm using the NotoColorEmoji-Regular.ttf font. I'm on macOS 14.2.1 on M1 Pro. When it renders I see the box for missing glyphs for abc but just black for the rest.

I'm starting to think that I'm doing emojis the wrong way because the object returned by `draw_glyphs` has a brush object that defaults to pure black, but I don't want my emojis to be all black. I've looked around and I can't find anything about how to render an emojis. I'm happy to provide a gist for people in the future trying to do a similar thing.

In short, is this something that is possible? And what files/docs should I go find to get the information I need to understand how to render emojis?

Thank you for such a wonderful library.

EDIT: explaining the parts of the code snippet

DJMcNab:
This is following the breadcrumbs laid out in [#gpu > Emoji Rendering](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Emoji.20rendering)

This supports rendering COLR emoji; we assume that these are reasonably well behaved.
image

For examples, we use a very small subset of [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji), which is approximately 5.5kB.

Fixes #536 

We will need a follow up issue for `sbix`/`CBDT`. I presume we're ignoring svg-in-opentype.


dfrg:
Yes, this is looking good so far. Note that you’ll need to grab the `CPAL` table to resolve palette indices to actual colors. We don’t do this by default because we need to allow custom overrides by index for the web. 

Thanks for taking this on!

examples/scenes/src/simple_text.rs
```diff
@@ -14,10 +14,13 @@ use vello::Scene;
 // On Windows, can set this to "c:\\Windows\\Fonts\\seguiemj.ttf" to get color emoji
 const ROBOTO_FONT: &[u8] = include_bytes!("../../assets/roboto/Roboto-Regular.ttf");
 const INCONSOLATA_FONT: &[u8] = include_bytes!("../../assets/inconsolata/Inconsolata.ttf");
+const NOTO_COLOR_EMOJI: &[u8] =
```
DJMcNab, line 17: To test this, you need to import the font yourself.

vello/src/scene.rs
```diff
@@ -334,23 +345,284 @@ impl<'a> DrawGlyphs<'a> {
...
     /// The `style` parameter accepts either `Fill` or `&Stroke` types.
     pub fn draw(mut self, style: impl Into<StyleRef<'a>>, glyphs: impl Iterator<Item = Glyph>) {
-        let resources = &mut self.encoding.resources;
-        self.run.style = style.into().to_owned();
-        resources.glyphs.extend(glyphs);
-        self.run.glyphs.end = resources.glyphs.len();
-        if self.run.glyphs.is_empty() {
+        let font_index = self.run.font.index;
+        let font = skrifa::FontRef::from_index(self.run.font.data.as_ref(), font_index).unwrap();
+        if font.colr().is_ok() {
```
DJMcNab, line 350: What's the actual right way to detect this?

vello/src/scene.rs
```diff
@@ -334,23 +345,284 @@ impl<'a> DrawGlyphs<'a> {
...
+        let transform = conv_skrifa_transform(transform);
+        let prior_transform = self.last_transform();
+        self.transform_stack.push(prior_transform * transform);
+    }
+
+    fn pop_transform(&mut self) {
+        self.transform_stack.pop();
+    }
+
+    fn push_clip_glyph(&mut self, _: GlyphId) {
```
DJMcNab, line 439: This really does not match the Vello imaging model - I'm inclined to leave glyphs which require this as unsupported for the moment

DJMcNab (commented):
Some helping comments

vello/src/scene.rs
dfrg: This is just a `push_layer` with the glyph outline as the path and the `Mix::Clip` blending mode.

dfrg: (and the current transform)

dfrg: How so? Clips and layers share the blend stack in vello. We shouldn’t need to track anything but transforms.

vello/src/scene.rs
dfrg: This is fine but I’d also check for cpal here since you need both.

vello/src/scene.rs
DJMcNab: There's no way to implement "fill the entire current clip area" from the scene API, as far as I know
Nor is there "add a layer atop the entire current clip area"

Which means we need to store the innermost clip area at all times, which as far as I can requires a new stack of the outlines/bezpaths. Maybe I'm missing some bit of API surface? 

And if we get a:
```
push_clip
push_layer
pop_clip
pop_layer
```
interface, we need even more messiness.
Plus we have to ignore a `push_layer` before anything else, because that's meaningless, but then we have to ignore the corresponding `pop_layer`. It seems to be a massive headache which I would rather not deal with if we can get away with it

dfrg: There are definitely ways to produce degenerate scenes but the flattened representation is generated from a tree so clips/layers will always be properly ordered and popped assuming there’s not a parsing error.

This work looks great to me so far and I’d be happy to pick it up if you’d like to merge and move on.

vello/src/scene.rs
```diff
@@ -333,24 +344,340 @@ impl<'a> DrawGlyphs<'a> {
...
+
+    fn push_clip_glyph(&mut self, glyph_id: GlyphId) {
+        let Some(outline) = self.outlines.get(glyph_id) else {
+            eprintln!("Didn't get expected outline");
+            return;
+        };
+
+        let mut path = BezPathOutline(BezPath::new());
+        let draw_settings =
+            DrawSettings::unhinted(skrifa::instance::Size::unscaled(), [].as_slice());
```
dfrg, line 460: This should use the same coords/LocationRef as passed to the paint method. Fonts can have both variable paints and variable outlines and they share the same variation space.

vello/src/scene.rs
```diff
@@ -333,24 +344,340 @@ impl<'a> DrawGlyphs<'a> {
...
+        brush: skrifa::color::Brush<'_>,
+    ) {
+        let Some(outline) = self.outlines.get(glyph_id) else {
+            eprintln!("Didn't get expected outline");
+            return;
+        };
+
+        let mut path = BezPathOutline(BezPath::new());
+        let draw_settings =
+            DrawSettings::unhinted(skrifa::instance::Size::unscaled(), [].as_slice());
```
dfrg, line 543: Same wrt coords.

vello/src/scene.rs
```diff
@@ -333,24 +344,340 @@ impl<'a> DrawGlyphs<'a> {
...
+            Gradient::new_sweep(conv_point(c0), start_angle, end_angle)
+                .with_extend(conv_extend(extend))
+                .with_stops(ColorStopsConverter(color_stops, cpal)),
+        ),
+    }
+}
+
+fn color_index(cpal: &'_ Cpal<'_>, palette_index: u16) -> Color {
+    let actual_colors = cpal.color_records_array().unwrap().unwrap();
+    let color = actual_colors[usize::from(palette_index)];
```
dfrg, line 653: The spec defines palette index 0xFFFF as “foreground color”. I guess we can use the provided brush for this, maybe taking the “dominant” color if it’s a gradient? Either way, we should probably avoid the potential panic here.

vello/src/scene.rs
```diff
@@ -333,24 +344,340 @@ impl<'a> DrawGlyphs<'a> {
...
+fn conv_point(point: skrifa::raw::types::Point<f32>) -> Point {
+    Point::new(point.x.into(), point.y.into())
+}
+
+fn conv_extend(extend: skrifa::color::Extend) -> Extend {
+    match extend {
+        skrifa::color::Extend::Pad => Extend::Pad,
+        skrifa::color::Extend::Repeat => Extend::Reflect,
+        skrifa::color::Extend::Reflect => Extend::Repeat,
+        // TODO: Error reporting on unknown variant?
```
dfrg, line 666: This seems like a reasonable default. Lots of broken fonts in the wild and there’s not much we can do about it now :(

vello/src/scene.rs
```diff
@@ -333,24 +344,340 @@ impl<'a> DrawGlyphs<'a> {
...
+            skrifa::color::CompositeMode::DestOver => Compose::DestOver,
+            skrifa::color::CompositeMode::SrcIn => Compose::SrcIn,
+            skrifa::color::CompositeMode::DestIn => Compose::DestIn,
+            skrifa::color::CompositeMode::SrcOut => Compose::SrcOut,
+            skrifa::color::CompositeMode::DestOut => Compose::DestOut,
+            skrifa::color::CompositeMode::SrcAtop => Compose::SrcAtop,
+            skrifa::color::CompositeMode::DestAtop => Compose::DestAtop,
+            skrifa::color::CompositeMode::Xor => Compose::Xor,
+            skrifa::color::CompositeMode::Plus => Compose::Plus,
+            // TODO:
```
dfrg, line 519: Yes, this conversion is annoying since the COLR spec conflates color mixing and composition.

vello/src/scene.rs
```diff
@@ -333,24 +344,446 @@ impl<'a> DrawGlyphs<'a> {
...
+            self.run.glyphs.start = self.run.glyphs.end;
+            self.run.stream_offsets = self.scene.encoding.stream_offsets();
+            outline_count += self.draw_outline_glyphs(clone_style_ref(&style), outline_glyphs);
+
+            let Some((color, glyph)) = final_glyph.take() else {
+                // All of the remaining glyphs were outline glyphs
+                break;
+            };
+
+            if glyph.id == 2 {
```
dfrg, line 434: Lingering debug stuff

dfrg (approved):
Thanks Daniel! This covers a lot of ground and is a great foundation for full emoji support.

PR #617, opened 2024-06-21
Fix dead code warnings and `warn(unreachable_pub)`

DJMcNab:
This also unblocks the upgrade to Rust 1.79.

As far as I can tell, this doesn't make anything newly private, it just properly tracks the private state of all items.

The new 1.79 dead code warnings were caused by the build script for `vello_shaders`, which uses custom path imports to import the same modules as `vello_shaders` itself uses.


waywardmonkeys (approved):
I'm kind of mixed on the whole `warn(unreachable_pub)` thing ... it feels like it adds a lot of visual noise.

But if that's fine, then I'm fine.

dfrg:
I’ll just say that I tend to view `pub(crate)` as a mechanism for providing internal access to “private” portions of public APIs. That is, I see it as a tool for enabling exceptions. The `unreachable_pub` warning essentially forces all internal APIs to be marked `pub(crate)` and IMO dilutes the value and increases noise. It seems to me that having private modules with public members is a perfectly valid way to express “internal API” in Rust. 

As a compromise, I do think it’s reasonable to tag _types_ with `pub(crate)` or `pub(super)` to prevent accidental exposure if that’s a concern but extending it to all methods and fields feels a bit too noisy. 

I’m curious as to how others feel about this.

waywardmonkeys:
I think that is a good expression of how I feel.

DJMcNab:
I think that having `unreachable_pub` enabled for the published crates is useful, because it clearly delineates what is exported against what isn't. In my mind, it's important for `pub` to actually represent something which is pub.

But I'll undo it for `scenes` and `with_winit`, which should reduce the noise of this change

PR #618, opened 2024-06-24
Hack around unsupported glyphs

DJMcNab:
Currently, if we get a glyph which the current font "doesn't support" (such as if that font only supports bitmap images, which we don't), then the entire transforms stream gets misaligned. That leads to some funny rendering, such as 
Screenshot of the mason example, with misaligned text, some upside down characters and incorrect looking layout.

This PR works around this, by instead of skipping these glyphs, inserting an empty encoding into their location.


xorgy (approved):
LGTM, hope we don't have much use for this in the future! :+ )

PR #619, opened 2024-06-26
Document the MSRV in CHANGELOG.md

DJMcNab:
I think this is probably a useful property of a release to document.


waywardmonkeys (approved):
Yes! I often have to go and look at the version listing on crates.io to see this for other crates.

PR #621, opened 2024-06-28
[vello_shaders] Bring back the Bazel build hack

armansito:
Skia, an external consumer of this crate, uses [Bazel rules](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/bazel/external/vello/BUILD.bazel) to compile Rust code. Due to limitations in Bazel's rust support, the CARGO_MANIFEST_DIR environment variable doesn't get set properly for nested crates since Bazel seems to assign the workspace root instead (as the [BUILD](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/bazel/external/vello/BUILD.bazel) file needs to be specified relative to the external repository root).

Until we figure out a better way to manage the bazel integration, this PR reintroduces a recently removed hack to work around this by allowing the crate manifest path to be assigned explicitly, using a new environment variable called `BAZEL_CRATE_MANIFEST_PATH`. This is intended to be the absolute path to `vello_shaders/Cargo.toml` within the bazel sandbox file system.

### DIFF
--- a/vello_shaders/src/compile/mod.rs
+++ b/vello_shaders/src/compile/mod.rs
@@ -274,5 +274,24 @@ fn postprocess(wgsl: &str) -> String {
 // NOTE: Embedding build environment info into the code makes reproducible builds trickier.
 pub fn shader_dir() -> &'static PathBuf {
     static SHADER_DIR: OnceLock<PathBuf> = OnceLock::new();
-    SHADER_DIR.get_or_init(|| Path::new(env!("CARGO_MANIFEST_DIR")).join("shader"))
+    SHADER_DIR.get_or_init(|| manifest_dir().join("shader"))
+}
+
+// In a regular cargo build the manifest directory is simply given by CARGO_MANIFEST_DIR.
+//
+// Skia, an external consumer of this crate, uses Bazel rules to compile Rust code. Due to
+// limitations in Bazel's rust support, Skia maintains its own build definitions
+// (https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/bazel/external/vello/BUILD.bazel).
+//
+// Because of the current setup, Bazel sets CARGO_MANIFEST_DIR to the workspace root instead of the
+// actual crate being built. This could be improved but until then, we work around this by allowing
+// the absolute path to the vello_shader crate's manifest to be specified using the
+// `BAZEL_CRATE_MANIFEST_PATH` build script environment variable. This should never be set when
+// using cargo.
+fn manifest_dir() -> PathBuf {
+    use std::env;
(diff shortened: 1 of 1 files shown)


armansito:
I think a better solution going forward might be to host this bazel file directly in the vello repo, which may fix the issue with `CARGO_MANIFEST_DIR`.

The reason Skia requires Bazel rules to build the crates (as opposed to listing them as a third-party cargo dependencies, which Bazel **does** support) is that I couldn't find a cleaner way to conditionally toggle crate features (particularly `msl` and `wgsl`) from the client-side Bazel rules.

vello_shaders/src/compile/mod.rs
```diff
@@ -274,5 +274,24 @@ fn postprocess(wgsl: &str) -> String {
...
+// (https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/bazel/external/vello/BUILD.bazel).
+//
+// Because of the current setup, Bazel sets CARGO_MANIFEST_DIR to the workspace root instead of the
+// actual crate being built. This could be improved but until then, we work around this by allowing
+// the absolute path to the vello_shader crate's manifest to be specified using the
+// `BAZEL_CRATE_MANIFEST_PATH` build script environment variable. This should never be set when
+// using cargo.
+fn manifest_dir() -> PathBuf {
+    use std::env;
+    env::var_os("BAZEL_CRATE_MANIFEST_PATH")
```
DJMcNab, line 293:
```suggestion
    env::var_os("UNSTABLE_BAZEL_VELLO_SHADERS_CRATE_MANIFEST_PATH")
```
I don't think this is *strictly* necessary, but I want to emphasise that this is both a targeted hack, and isn't something we're treating as a stable behaviour (i.e. that it could change in a patch release) - because as I understand it, Skia will depend on Vello on a specific git revision anyway.

DJMcNab (approved):
Some of the documentation in that BUILD file suggests that this hack was needed because the `shader` directory is outside `vello_shaders`. That is no longer true since #590. 

Does that change the equation here?

Approving anyway, because the actual change is fine.

vello_shaders/src/compile/mod.rs
armansito: Done.

armansito:
> Some of the documentation in that BUILD file suggests that this hack was needed because the `shader` directory is outside `vello_shaders`. That is no longer true since #590.
> 
> Does that change the equation here?

That was part of the original reason for this hack. I was hoping that #590 would address it but I ran into some limitations. I'll update that comment in Skia when I roll this PR in.

> 
> Approving anyway, because the actual change is fine.

Thanks!

PR #625, opened 2024-07-08
clippy: Fix `manual_pattern_char_comparison` lint

waywardmonkeys:
Using this pattern of code allows for more efficient code and is more concise.

This is a lint that slated to be added in 1.80.


DJMcNab (approved):
Oh, that's a cool feature!

Clearly I should investigate the implementations of `Pattern`.

PR #626, opened 2024-07-09
Migrate examples to winit's new "app" API

yutannihilation:
A followup of https://github.com/linebender/vello/pull/559

I see https://github.com/linebender/vello/pull/559#issuecomment-2088285475 indicates there's some plan of more significant rewrite, so please feel free to close this. This is mainly my practice of migration to winit's new API.

cf. [changelog](https://docs.rs/winit/latest/winit/changelog/index.html#deprecated)

### DIFF
--- a/examples/simple/src/main.rs
+++ b/examples/simple/src/main.rs
@@ -28,7 +28,7 @@ enum RenderState<'s> {
     Suspended(Option<Arc<Window>>),
 }
 
-struct RenderStateWithContext<'s> {
+struct SimpleVelloApp<'s> {
     // The vello RenderContext which is a global context that lasts for the
     // lifetime of the application
     context: RenderContext,
@@ -45,7 +45,7 @@ struct RenderStateWithContext<'s> {
     scene: Scene,
 }
 
-impl<'s> ApplicationHandler for RenderStateWithContext<'s> {
+impl<'s> ApplicationHandler for SimpleVelloApp<'s> {
     fn resumed(&mut self, event_loop: &ActiveEventLoop) {
         let RenderState::Suspended(cached_window) = &mut self.state else {
             return;
@@ -167,7 +167,7 @@ impl<'s> ApplicationHandler for RenderStateWithContext<'s> {
 
 fn main() -> Result<()> {
     // Setup a bunch of state:
-    let mut state = RenderStateWithContext {
+    let mut app = SimpleVelloApp {
         context: RenderContext::new(),
         renderers: vec![],
         state: RenderState::Suspended(None),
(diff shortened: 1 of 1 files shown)


examples/simple/src/main.rs
```diff
@@ -27,149 +28,156 @@ enum RenderState<'s> {
     Suspended(Option<Arc<Window>>),
 }
 
-fn main() -> Result<()> {
-    // Setup a bunch of state:
-
-    // The vello RenderContext which is a global context that lasts for the lifetime of the application
-    let mut render_cx = RenderContext::new();
+struct RenderStateWithContext<'s> {
```
DJMcNab, line 31: I think we should rename this to something which describes what it is for, rather than what it contains. E.g. `SimpleVelloApp`?

DJMcNab (approved):
This still works for me. Thanks!

Are you planning on also applying this to the `with_winit` example, or should we just land this as-is?

examples/simple/src/main.rs
yutannihilation: Thanks for the suggestion! `SimpleVelloApp` sounds better.

yutannihilation:
> Are you planning on also applying this to the `with_winit` example, or should we just land this as-is?

Thanks, I think I can address it in this pull request.

yutannihilation:
Sorry, I tried `with_winit`, but it crashes on my local after my rewrite and it will probably take some time to figure out what's happening due to my lack of knowledge. So, I'd like to not include it in this pull request. 

```
     Running `target\debug\with_winit_bin.exe`
[2024-07-15T15:04:06.672Z ERROR wgpu_hal::vulkan::instance] GENERAL [Loader Message (0x0)]
        loader_scanned_icd_add: Driver C:\Windows\System32\DriverStore\FileRepository\u0379858.inf_amd64_1719b9caae095668\B377432\.\amdvlk64.dll says it supports interface version 6 but still exports core entrypoints (Policy #LDP_DRIVER_6)
[2024-07-15T15:04:06.673Z ERROR wgpu_hal::vulkan::instance]     objects: (type: INSTANCE, hndl: 0x1f291a0ccd0, name: ?)
Initialising in parallel using 14 threads
Parsed svg Ghostscript_Tiger in 8.9593ms
Encoded svg Ghostscript_Tiger in 799.6µs
thread 'main' panicked at examples\with_winit\src\lib.rs:511:22:
failed to get surface texture: Timeout
error: process didn't exit successfully: `target\debug\with_winit_bin.exe` (exit code: 101)
```

Here's the migrated code of `with_winit` (WIP).

https://github.com/linebender/vello/compare/main...yutannihilation:winit-0.30-new-api-2

DJMcNab:
Hmm. I'll land this then. Would you mind making a draft follow-up PR with those changes?

yutannihilation:
Sure, I will!

PR #627, opened 2024-07-14
Correct example code in README.md

kmoon2437:
Corrected some incorrect parts of the example code in README.md file(i.e. using `vello::peniko::Color`, `vello::kurbo::Circle` as `vello::Color`, `vello::Circle`) and added `use`s.


DJMcNab (approved):
Thanks for bringing this to our attention! The incorrect import paths is especially egregious; it's good to fix that.

I've tweaked things slightly, just to reduce the amount of boilerplate being shown.

PR #628, opened 2024-07-15
Migrate "with_winit" examples to winit's new "app" API

yutannihilation:
Second round of #626

Currently, this crashes with the following on my local, so I think I'm doing something wrong.

```
     Running `target\debug\with_winit_bin.exe`
[2024-07-15T15:33:30.317Z ERROR wgpu_hal::vulkan::instance] GENERAL [Loader Message (0x0)]
        loader_scanned_icd_add: Driver C:\Windows\System32\DriverStore\FileRepository\u0379858.inf_amd64_1719b9caae095668\B377432\.\amdvlk64.dll says it supports interface version 6 but still exports core entrypoints (Policy #LDP_DRIVER_6)
[2024-07-15T15:33:30.318Z ERROR wgpu_hal::vulkan::instance]     objects: (type: INSTANCE, hndl: 0x1ca928fbea0, name: ?)
Initialising in parallel using 14 threads
Parsed svg Ghostscript_Tiger in 8.3298ms
Encoded svg Ghostscript_Tiger in 893.9µs
thread 'main' panicked at examples\with_winit\src\lib.rs:511:22:
failed to get surface texture: Timeout
error: process didn't exit successfully: `target\debug\with_winit_bin.exe` (exit code: 101)

```

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -670,6 +670,7 @@ fn run(
         context: render_cx,
         renderers,
         state: render_state,
+        #[cfg(not(target_arch = "wasm32"))]
         cached_window: None,
         use_cpu: args.use_cpu,
         num_init_threads: args.num_init_threads,
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -110,7 +110,9 @@ struct VelloApp<'s> {
     #[cfg(not(target_arch = "wasm32"))]
     cached_window: Option<Arc<Window>>,
 
+    #[cfg(not(target_arch = "wasm32"))]
     use_cpu: bool,
+    #[cfg(not(target_arch = "wasm32"))]
     num_init_threads: usize,
 
     scenes: Vec<ExampleScene>,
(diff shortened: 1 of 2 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -88,30 +92,540 @@ struct RenderState<'s> {
...
+                    height,
+                    antialiasing_method,
+                };
+                self.scene.reset();
+                if let Some(resolution) = scene_params.resolution {
+                    // Automatically scale the rendering to fill as much of the window as possible
+                    // TODO: Apply svg view_box, somehow
+                    let factor = Vec2::new(width as f64, height as f64);
+                    let scale_factor = (factor.x / resolution.x).min(factor.y / resolution.y);
+                    self.transform *= Affine::scale(scale_factor);
```
DJMcNab, line 467: Why are you modifying `self.transform` here?

Changing this to match the prior logic resolves the crashing issue for me.

DJMcNab (commented):
Other than the logic you've changed, this works fine for me.

examples/with_winit/src/lib.rs
yutannihilation: Thanks for catching this! Fixing this mistake works for me, too.

yutannihilation:
Thanks for reviewing!

PR #629, opened 2024-07-16
Implement workaround to prevent panics with empty scenes

PoignardAzur:
I know someone is going to answer: "This isn't the optimal solution, we should fix this problem at the root and add checks to the GPU pipelines to avoid the panics".

Yes, well, people have been saying we should do that for months, and the panic is still here. If someone wants to implement the proper solution, they can do so after this is merged.

People are still having crashes because of this. We need a fix *last month*, pretty or not.


DJMcNab:
Superseded by #630

PR #631, opened 2024-07-16
Update to wgpu 0.20.1

waywardmonkeys:
This also updates `wgpu-core` and `wgpu-hal` to the corresponding versions (which are 0.21.x).

This fixes some undefined behavior when building with Rust 1.79 and using the Direct3D 12 backend: https://github.com/gfx-rs/wgpu/issues/5812

It also fixes an issue on OpenGL where non-sRGB was still using sRGB: https://github.com/gfx-rs/wgpu/pull/5642

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2808,9 +2808,9 @@ dependencies = [
 
 [[package]]
 name = "wgpu"
-version = "0.20.0"
+version = "0.20.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
@simbleau I think you had an issue with color on OpenGL before. Would the sRGB-related fix here have helped you?

waywardmonkeys:
@DJMcNab The other important part is that the wgpu 0.20.0 release was yanked.

PR #634, opened 2024-07-16
Allow access to adapter

cfagot:
Currently there is no way to access the adapter outside of Vello, which prevents users of Vello from doing things like checking surface capabilities. This pull request exposes the adapter.


DJMcNab (approved):
Thanks! I'm planning on exploring a much more well-thought through API, but this makes the util more useful.

Fwiw, the intention of `util` is that if you need something which util doesn't provide, you should be able to use the lower-level primitives yourself. But as I say, very little API design work has gone into Vello, so that might not be possible.

PR #635, opened 2024-07-19
Update to wgpu 22.1.0

waywardmonkeys:


waywardmonkeys:
This is blocked on https://github.com/Wumpf/wgpu-profiler/pull/75

That in turn is very likely blocked on getting a 22.0.1 release of `wgpu`.

Opening this here now to track that it is in progress and what's blocking it as well as call out a couple of discussion points.

1) In `vello_shaders`, we have a feature `msl` which could be changed to only happen on Apple targets by using `msl-out-if-target-apple` instead with Naga 22.  I don't know what the upstream users of this feature want.
2) We could start to enable the work from #524 to improve startup time when using Vulkan.
3) ???

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This release has an [MSRV][] of 1.75.
 
 ### Changed
 
+- Updated `wgpu` to 22.1.0. ([#635] by [@waywardmonkeys])
```
DJMcNab, line 22: We don't have a consistent policy here. Is it worth marking this as a breaking change?

vello/src/wgpu_engine.rs
```diff
@@ -678,6 +678,7 @@ impl WgpuEngine {
                 zero_initialize_workgroup_memory: false,
                 ..Default::default()
             },
+            cache: None,
```
DJMcNab, line 681:
```suggestion
            // TODO: Support providing a cache here
            cache: None,
```

DJMcNab (approved):

CHANGELOG.md
waywardmonkeys: I think it should be for now ... if we don't, it would cause some problems for users of `bevy_vello`.

DJMcNab: I wasn't suggesting we don't treat it as a breaking change. I mean to ask whether we should call it out as a breaking change in the changelog

waywardmonkeys: Oh, sorry. Doing too many things at once. How should we do that? I'm okay with doing so here.

waywardmonkeys: I did this like your own entry now. Will merge soon.

PR #638, opened 2024-07-22
Update typos, fix typo, improve config.

waywardmonkeys:
Instead of matching anywhere in a word, we can match the exact identifier. This will help reduce not finding typos.


DJMcNab (approved):
Thanks!

PR #641, opened 2024-07-25
Support bitmap fonts

DJMcNab:
Resolve bitmap Emoji fonts at scene construction time (rather than the ideal resolve time for glyph caching) for expedience.

Current status: Layout details aren't entirely correct


vello/src/scene/bitmap.rs
```diff
@@ -0,0 +1,374 @@
...
+// This code (apart from imports) is not modified from upstream, and
+// we don't want to touch this code to minimise licensing complications.
+#![allow(warnings)]
+// Taken from https://github.com/googlefonts/fontations/blob/cbdf8b485e955e3acee40df1344e33908805ed31/skrifa/src/bitmap.rs
+// Used under the Apache 2.0 OR MIT License.
+
+// The MIT license used is in LICENSE-MIT-SKRIFA in the crate root or
+// at https://github.com/googlefonts/fontations/blob/cbdf8b485e955e3acee40df1344e33908805ed31/LICENSE-MIT
+// The Apache 2.0 License is the same as LICENSE-APACHE in the crate root, or
+// at https://github.com/googlefonts/fontations/blob/cbdf8b485e955e3acee40df1344e33908805ed31/LICENSE-APACHE
```
xStrom, line 13:
```suggestion
// Copyright 2024 the Vello Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT

#![allow(warnings)]
// Based on https://github.com/googlefonts/fontations/blob/cbdf8b485e955e3acee40df1344e33908805ed31/skrifa/src/bitmap.rs
```

.github/copyright.sh
```diff
@@ -7,7 +7,7 @@
 #   -g "!src/special_directory"
 
 # Check all the standard Rust source files
-output=$(rg "^// Copyright (19|20)[\d]{2} (.+ and )?the Vello Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2\.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!vello_shaders/{shader,src/cpu}" .)
+output=$(rg "^// Copyright (19|20)[\d]{2} (.+ and )?the Vello Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2\.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!vello_shaders/{shader,src/cpu}" -g "!vello/src/scene/bitmap.rs" .)
```
xStrom, line 10:
```suggestion
output=$(rg "^// Copyright (19|20)[\d]{2} (.+ and )?the Vello Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2\.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!vello_shaders/{shader,src/cpu}" .)
```

xStrom (commented):
I don't think there is any license issue here. Our licenses match exactly, and we already have Google LLC code in the repo as-is.

I think our generic `LICENSE-MIT` is sufficient. We can also add *Colin Rothfels* to our `AUTHORS` file, even though that doesn't have an impact on copyright.

Thus I think we should just keep it simple, use our standard copyright header, not modify the copyright check, not add `LICENSE-MIT-SKRIFA`, and not worry about making changes to the `bitmap.rs` file.

vello/src/scene.rs
```diff
@@ -511,7 +511,8 @@ impl<'a> DrawGlyphs<'a> {
 
                             if !mask.is_packed {
                                 // TODO: Error once?
-                                log::warn!("Unpacked mask data not handled");
+                                log::warn!("Unpacked mask data in font not yet supported");
+                                // TODO: How do we get the font name here?
```
waywardmonkeys, line 515: If we were using `tracing`, we'd be able to have some context data around that made this more clear, perhaps? (Clearly not a thing for this PR to solve.)

DJMcNab: I suspect that getting that information "just-in-case" all the time would be too expensive. This is a potentially hot-spot in the API.

waywardmonkeys: Ideally, it would be somewhere further up the call stack, although for fonts, maybe not quite far enough.

xorgy (approved):

PR #643, opened 2024-07-25
[Experiment] Store snapshots in `git lfs`

DJMcNab:
This a living experiment to determine how much LFS bandwidth we would use if we stored our snapshot tests in Git LFS.

The intention is once this is merged, we will then add some additional snapshot tests, which use this new folder, but which won't fail (at least not on CI?) if LFS downloading failed. That way, we have a graceful degradation if we do run out of bandwidth, but improving resilience if LFS is working.


.github/workflows/ci.yml
```diff
@@ -216,6 +216,9 @@ jobs:
         run: cargo nextest run --workspace --locked --all-features
         env:
           VELLO_CI_GPU_SUPPORT: ${{ matrix.gpu }}
+          # If LFS downloading failed, we still want CI to work.
+          # Any values still in the cache will still work.
+          VELLO_SKIP_LFS_SNAPSHOTS: all
```
Philipp-M, line 221: Has this occurred already (as I haven't used such yet, but I don't think it hurts anyway)?

vello_tests/src/lib.rs
```diff
@@ -171,6 +176,20 @@ pub fn write_png_to_file(
...
+    if max_size_in_bytes.is_some_and(|max_size_in_bytes| size > max_size_in_bytes) {
+        std::fs::rename(out_path, &oversized_path)?;
+        bail!(
+            "File was oversized, expected {} bytes, got {size} bytes. New file written to {to}",
+            max_size_in_bytes.unwrap(),
+            to = oversized_path.display()
+        );
+    } else {
+        // Intentionally do not handle errors here
+        drop(std::fs::remove_file(oversized_path));
```
Philipp-M, line 191: Just for my understanding, why is this removed? It shouldn't even be there should it? Or is this because when a previous image resulted `oversized` and is now not anymore, that this will be cleaned up?

vello_tests/src/snapshot.rs
```diff
@@ -95,33 +106,78 @@ impl Snapshot<'_> {
...
+    /// This test will ensure that files produced are no more than 4KiB.
+    Smoke,
+    /// Run the test in a git LFS managed directory.
+    ///
+    /// This test will ensure that files produced are no larger than 128KiB.
+    Lfs,
+}
+
+impl SnapshotDirectory {
+    fn max_size_in_bytes(self) -> u64 {
```
Philipp-M, line 124: Not that it does matter, I think this could even be `const`

Philipp-M (approved):
Since there wasn't backlash in office hours, let's try that experiment, thanks for implementing, LGTM

.github/workflows/ci.yml
DJMcNab: I can comment this out, and say that if you need to uncomment it, to let me know and to just do so.

vello_tests/src/lib.rs
DJMcNab: Yeah, if the previous run was oversized, we no longer need this file.

It's pretty poor UX to have old snapshot files "hanging around"

vello_tests/src/snapshot.rs
DJMcNab: It's always used with dynamic arguments, but I see your point. As it happens, I originally wrote it as `const`, then removed that again. If you add this as a suggestion, I'll apply it.

.github/workflows/ci.yml
Philipp-M: I mean I'm mostly indifferent, but I think it's probably not bad to know (i.e. in failures) if this is really an issue (IME it's not).

vello_tests/src/snapshot.rs
Philipp-M: As I say it doesn't really matter, it's not that it helps much, I think you could keep it as that, just something I noticed

PR #645, opened 2024-07-26
Revert "[Experiment] Store snapshots in `git lfs`"

DJMcNab:
Reverts linebender/vello#643

This is a temporary reversal whilst we work out our merge queue story.

The current status is that GitHub actions doesn't actually reuse caches properly - see https://github.com/actions/cache/issues/1243, so we need to manage a workaround ourselves.


waywardmonkeys:
Could you just disable running these in CI instead?

DJMcNab:
That is vello#646.

Issue #644, opened 2024-07-26
Corruption issue with clipping


nicoburns:
The output rendered by Vello is sometimes corrupted when using clips such as:

```rust
scene.push_layer(Mix::Clip, 1.0, transform, &clip);
scene.pop_layer()
```

This occurs:
- With or without anything being drawn in the clipped layer
- With only small numbers of clips (see below for an example with only 3)
- With no clip nesting (although the example below might nest - not sure)
- With various different blend modes (`Blendmode::default()` and `Mix::Normal` both also exhibit this behaviour)
- With various transform/clip shapes (these do appear to sometimes affect the nature/position of the corruption)

## Runnable example

The `vello-clip-bug` branch of Blitz (https://github.com/DioxusLabs/blitz/tree/vello-clip-bug) reproduces this with `cargo run --release --example google` (also reproduces in debug mode). I think this is the best repro because it happens with only 3 clips (assuming my counting is correct). Vello-related code is almost all in `packages/blitz/src` (and there isn't much else in that dir). The vello setup code is in `renderer.rs`. The vello scene creation code is in `renderer/render.rs`.

There is a suspicion that this could be caused by overlapping clips: https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Vello.3A.20.60push_layer.60.20causing.20black.20rectangle

## Screenshots

The issue causes corruption like the following:

<img width="912" alt="Screenshot 2024-07-25 at 20 59 55" src="https://github.com/user-attachments/assets/2ce0ad32-5a8f-4d76-806d-d476a3e0fd65">

It is supposed to look like:

<img width="912" alt="Screenshot 2024-07-26 at 13 06 06" src="https://github.com/user-attachments/assets/9965b612-4842-47b4-85af-a408eeb721c7">

raphlinus:
I've done some investigation on this. I believe the main culprit is the zero-size clip that's being pushed - if I tweak the code so the width is at least 1e-3 (instead of 0), then the rendering is as intended.

To be clear, that is a bug in Vello and should be fixed. But it also suggests a workaround.

Other findings: use_cpu makes no difference, and it's not an issue with the blend stack size - that said, I think we should prioritize that last soon, as it's likely we'll run into it as applications use deeply nested clips.

nicoburns:
[...]

PR #649, opened 2024-07-30
Clip bug reproduction(s) (DO NOT MERGE)

nicoburns:
This PR is to track reproduction attempts for https://github.com/linebender/vello/issues/644

Current reproduction attempt is in the `simple` example of this branch.


DJMcNab:
Following through the commits, I'm not sure a regression test for this ever landed.

It would be good to get one if so.

PR #651, opened 2024-07-31
Encode valid empty paths for empty clips

### LINKED ISSUE #644: Corruption issue with clipping

raphlinus:
There was logic to encode a valid empty path when the clip path was empty, but this logic was ineffective because of empty line segment culling. This patch directly encodes an empty path (single zero-length line).

Also includes a doc improvement for `push_layer`.

Fixes #644

### DIFF
--- a/vello/src/scene.rs
+++ b/vello/src/scene.rs
@@ -77,6 +77,9 @@ impl Scene {
         clip: &impl Shape,
     ) {
         let blend = blend.into();
+        if blend.mix == Mix::Clip && alpha != 1.0 {
+            log::warn!("Clip mix mode used with semitransparent alpha");
+        }
         let t = Transform::from_kurbo(&transform);
         self.encoding.encode_transform(t);
         self.encoding.encode_fill_style(Fill::NonZero);
@@ -84,6 +87,12 @@ impl Scene {
             // If the layer shape is invalid, encode a valid empty path. This suppresses
             // all drawing until the layer is popped.
             self.encoding.encode_empty_shape();
+            #[cfg(feature = "bump_estimate")]
+            {
+                use peniko::kurbo::PathEl;
+                let path = [PathEl::MoveTo(Point::ZERO), PathEl::LineTo(Point::ZERO)];
+                self.estimator.count_path(path.into_iter(), &t, None);
+            }
         } else {
             #[cfg(feature = "bump_estimate")]
             self.estimator.count_path(clip.path_elements(0.1), &t, None);
--- a/vello/src/scene.rs
+++ b/vello/src/scene.rs
@@ -66,6 +66,9 @@ impl Scene {
     /// until the layer is popped.
     ///
     /// **However, the transforms are *not* saved or modified by the layer stack.**
+    ///
+    /// Clip layers (`blend` = [`Mix::Clip`]) should have an alpha value of 1.0.
+    /// For an opacity group with non-unity alpha, specify [`Mix::Normal`].
     pub fn push_layer(
         &mut self,
         blend: impl Into<BlendMode>,
@@ -80,8 +83,7 @@ impl Scene {
         if !self.encoding.encode_shape(clip, true) {
             // If the layer shape is invalid, encode a valid empty path. This suppresses
             // all drawing until the layer is popped.
-            self.encoding
-                .encode_shape(&Rect::new(0.0, 0.0, 0.0, 0.0), true);
+            self.encoding.encode_empty_shape();
         } else {
             #[cfg(feature = "bump_estimate")]
             self.estimator.count_path(clip.path_elements(0.1), &t, None);
--- a/vello_encoding/src/encoding.rs
+++ b/vello_encoding/src/encoding.rs
@@ -240,6 +240,11 @@ impl Encoding {
         encoder.finish(true) != 0
     }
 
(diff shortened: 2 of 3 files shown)


nicoburns (approved):
I don't have enough context to know whether there is an appropriate fix. But there is nothing obviously wrong with this PR.

vello/src/scene.rs
```diff
@@ -66,6 +66,9 @@ impl Scene {
     /// until the layer is popped.
     ///
     /// **However, the transforms are *not* saved or modified by the layer stack.**
+    ///
+    /// Clip layers (`blend` = [`Mix::Clip`]) should have an alpha value of 1.0.
+    /// For an opacity group with non-unity alpha, specify [`Mix::Normal`].
```
waywardmonkeys, line 71: I'd like to reword this to not use the term "non-unity" .. I know what it means, but I don't think all of our readers will.

vello/src/scene.rs
```diff
@@ -80,8 +83,7 @@ impl Scene {
         if !self.encoding.encode_shape(clip, true) {
             // If the layer shape is invalid, encode a valid empty path. This suppresses
             // all drawing until the layer is popped.
-            self.encoding
-                .encode_shape(&Rect::new(0.0, 0.0, 0.0, 0.0), true);
+            self.encoding.encode_empty_shape();
         } else {
```
waywardmonkeys, line 87: Should this have a bit for the `bump_estimate` feature to count the 1 segment (empty) path or is that not needed since it is empty?

vello_encoding/src/encoding.rs
```diff
@@ -240,6 +240,12 @@ impl Encoding {
         encoder.finish(true) != 0
     }
 
+    pub fn encode_empty_shape(&mut self) {
```
waywardmonkeys, line 243: This needs a doc comment.

waywardmonkeys (commented):
I feel like the bit about this needing to bypass the path collapsing needs to be captured in a comment somewhere.

I think that we need some infrastructure where we can add code-based tests for things where we validate the encoding. But perhaps not as part of this.  (If we had a way to validate that the encoding is okay, then that would open up to doing some fuzz testing to find ways to generate invalid encodings.)

vello/src/scene.rs
DJMcNab, line 71: Can we add a warning for this being violated?

DJMcNab (commented):
I think it would be worth adding a new screenshot test for this (potentially with a few different variations?). That would also let you test out the new testing workflow.

vello/src/scene.rs
dfrg: I believe clip is just an optimization so what about removing it from the public API and having normal+over+opaque encode as a clip?

DJMcNab:
I see that none of the review comments have been addressed. Backed out the merge in case a push has been missed.

raphlinus:
Argh, I was on the bus and missed that there were open review comments. Will address.

vello/src/scene.rs
raphlinus: I'm open to adding a warning but don't know what the technology for that is.

And, as discussed in the [Zulip thread](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Vello.3A.20.60push_layer.60.20causing.20black.20rectangle/near/460112832), it's not just an optimization, it has different semantics when doing nontrivial blending. I'll improve the docs a bit. Also, as discussed in that thread, it's likely that support for non-isolated blend groups should become a feature request, but that's out of scope for this PR.

vello_encoding/src/encoding.rs
DJMcNab: This is still the case

raphlinus: Duh, I had forgot to Cmd-S my buffer. Sorry about that; it feels like today is a day I shouldn't be operating heavy machinery.

PR #652, opened 2024-08-01
Upload screenshots as artifacts when tests fail

DJMcNab:
See b285dd28333fd26414234f39cdecd4e08ca6171c for an example of how this works!


waywardmonkeys (approved):
GO GO GO, this is SO cool.

Only thing would be that perhaps the screenshots could be bigger, but then there's more cost to that.

Philipp-M (approved):
Yeah, agree, this is pretty neat.

DJMcNab:
I've removed this from the milestone, as it's entirely a CI-related change, and doesn't impact users.

PR #654, opened 2024-08-04
feat: derives for `AaSupport`

simbleau:
Doesn't play well with bevy_vello since I can't interoperate these types or clone, etc.

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -134,15 +134,25 @@ use wgpu::{Device, Queue, SurfaceTexture, TextureFormat, TextureView};
 use wgpu_profiler::{GpuProfiler, GpuProfilerSettings};
 
 /// Represents the antialiasing method to use during a render pass.
-#[derive(Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum AaConfig {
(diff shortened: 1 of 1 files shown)


vello/src/lib.rs
```diff
@@ -134,14 +134,25 @@ use wgpu::{Device, Queue, SurfaceTexture, TextureFormat, TextureView};
...
+impl From<AaConfig> for AaSupport {
+    fn from(value: AaConfig) -> Self {
+        Self {
+            area: value == AaConfig::Area,
+            msaa8: value == AaConfig::Msaa8,
+            msaa16: value == AaConfig::Msaa16,
+        }
+    }
+}
+
```
DJMcNab, line 153: We should either:
```suggestion
```
or implement [`BitOr`](https://doc.rust-lang.org/std/ops/trait.BitOr.html) for `AaSupport` if we're doing this.

DJMcNab (commented):
Thanks! We really should enable some of the clippy lints in this repository as well.

See the one comment for the blocking concern.

vello/src/lib.rs
simbleau: Yeah, removing makes sense now that I know how the support/config work.

waywardmonkeys (approved):

PR #657, opened 2024-08-06
Allow deeper blend stacks by spilling to a buffer

DJMcNab:
This brings in support for blend spilling (which was supported in the old piet-gpu).

I don't have a good heuristic for how big to make the buffer. That is something which will need to be addressed in #606 (or its successors). I just guessed that 256 spills would be fine. I think this is probably too small - I suspect we'll get feedback from @TrueDoctor about this.

I have confirmed that the robustness works as expected with the GPU shaders.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This release has an [MSRV][] of 1.75.
 
 ### Added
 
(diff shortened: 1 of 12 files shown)


raphlinus (approved):
Thanks, this looks good. We'll do some followup work with the robust dynamic memory.

PR #659, opened 2024-08-06
Fix error in clip reduction

### LINKED ISSUE #648: Clips cause compute shaders to take 10 seconds and results in corruption
Rendering the following scene freezes my desktop environment for about 10 seconds before showing a corrupted result:

```
let mut scene = Scene::new();

for _i in 0..43 {
    {
        let mut clip: BezPath = Default::default();
        clip.move_to((0., 0.));
        clip.line_to((1., 1.));
        scene.push_layer(BlendMode::default(), 1., Affine::IDENTITY, &clip);
    }
    {
        let mut clip: BezPath = Default::default();
        clip.move_to((0., 0.));
        clip.line_to((200., 200.));
        scene.push_layer(BlendMode::default(), 1., Affine::IDENTITY, &clip);
    }
    {
        let mut clip: BezPath = Default::default();
        clip.move_to((0., 0.));
        clip.line_to((300., 300.));
        scene.push_layer(BlendMode::default(), 1., Affine::IDENTITY, &clip);
    }

    scene.pop_layer();
    scene.pop_layer();
    scene.pop_layer();
}

scene.fill(
    vello::peniko::Fill::NonZero,
    Affine::IDENTITY,
    Color::rgb(1., 0., 0.),
    None,
    &Rect::new(40.0, 40.0, 180.0, 180.0),
);
```

Whether the output is corrupted or not seems to be based on chance, which might point to some kind of memory corruption. The final fill is not necessary to cause the freeze, only to sometimes observe corruption. Using two clips instead of 3 doesn't reproduce the issue, even if we increase the number of iterations to arrive at the same overall number of clips. It doesn't reproduce with 42 iterations instead of 43. According to amdgpu_top, while my desktop environment is frozen, the GPU is actually busy running the compute shaders. It takes those 10s every time a frame is drawn. The problem does not reproduce with the CPU shaders.

My environment:
- `Linux chulak 6.8.12 #1-NixOS SMP PREEMPT_DYNAMIC Thu May 30 07:49:53 UTC 2024 x86_64 GNU/Linux`
- `AMD Radeon RX 6700 XT (radeonsi, navi22, LLVM 17.0.6, DRM 3.57, 6.8.12)`
- Vello at 3ee3bea02164c5a816fe6c16ef4e3a810edb7620

raphlinus:
One bit of logic from the GLSL (piet-gpu) version of clip_reduce didn't get copied over.

Fixes #648

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -68,6 +68,7 @@ export_scenes!(
     brush_transform(brush_transform: animated),
     blend_grid(blend_grid),
     deep_blend(deep_blend),
+    many_clips(many_clips),
     conflation_artifacts(conflation_artifacts),
     labyrinth(labyrinth),
     robust_paths(robust_paths),
@@ -82,7 +83,11 @@ export_scenes!(
 /// Implementations for the test scenes.
 /// In a module because the exported [`ExampleScene`](crate::ExampleScene) creation functions use the same names.
 mod impls {
+    use std::f64::consts::PI;
+
     use crate::SceneParams;
+    use rand::Rng;
+    use rand::{rngs::StdRng, SeedableRng};
     use vello::kurbo::{
         Affine, BezPath, Cap, Circle, Ellipse, Join, PathEl, Point, Rect, Shape, Stroke, Vec2,
     };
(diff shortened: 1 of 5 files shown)


DJMcNab (commented):
I think this also needs a test to be added.

I'm presuming that would just involve creating 256 clip areas?

examples/scenes/src/test_scenes.rs
```diff
@@ -1094,6 +1099,32 @@ mod impls {
...
+        base_tri.move_to((-50.0, 0.0));
+        base_tri.line_to((25.0, -43.3));
+        base_tri.line_to((25.0, 43.3));
+        for y in 0..10 {
+            for x in 0..10 {
+                let translate =
+                    Affine::translate((100. * (x as f64 + 0.5), 100. * (y as f64 + 0.5)));
+                const CLIPS_PER_FILL: usize = 3;
+                for _ in 0..CLIPS_PER_FILL {
+                    let rot = Affine::rotate(rng.gen_range(0.0..PI));
```
DJMcNab, line 1115: I believe that the root path is rotationally symmetric every $$\frac{2\pi}{3}$$ radians, but this is doing a rotation every $$\pi$$ radians.

I think that means there's a bias towards the range which is equivalent to $$\[0, \frac{\pi}{3})$$.

Practically, I don't think this matters here, but I'm wondering if this is an intentional choice.

(And of course, my maths understanding could be very wrong :P)

examples/scenes/src/test_scenes.rs
```diff
@@ -1094,6 +1099,32 @@ mod impls {
...
+                    scene.push_layer(Mix::Clip, 1.0, translate * rot, &base_tri);
+                }
+                let rot = Affine::rotate(rng.gen_range(0.0..PI));
+                let color = Color::rgb(rng.gen(), rng.gen(), rng.gen());
```
DJMcNab, line 1119: This produces remarkably pleasing colours considering that it's completely random - it's probably a good technique to use in other examples.

DJMcNab (approved):
It's so nice seeing new scenes come with a test, because seeing them like this is much easier than running them.

examples/scenes/src/test_scenes.rs
raphlinus: Heh, you're right, there is a bias. Not thinking it through carefully, I assumed the symmetry would make it still uniform, and I wanted to make the code concise. I think I'll leave it, in the sprit of kintsugi.

DJMcNab: We could use $$\tau$$ ([`TAU`](https://doc.rust-lang.org/stable/std/f64/consts/constant.TAU.html)) here to still be fairly concise.

But practically, this is great!

PR #661, opened 2024-08-07
Free the `blend_spill` buffer

DJMcNab:
This was missed in #657

I have been running into consistent issues with freeing things incorrectly; I think we need to develop better tools for this.

This will be a consideration for the planned API changes.


Philipp-M (approved):
LGTM

Issue #640, opened 2024-07-24
Drop shadows


StarArawn:
Support for drop shadows would be excellent! I did this in my own custom UI renderer by using SDF's for quads, but I imagine for any shape it might be a little bit more tricky.. 

You can see my quad sdf box shadow code here if it helps at all:
https://github.com/StarArawn/kayak_ui/blob/main/src/render/unified/shaders/sample_quad.wgsl#L86

raphlinus:
For arbitrary shapes (and blurs of layers other than images) the best bet is to do image filters, which are on the longer term roadmap but require quite a bit of architectural and infrastructure work.

Doing blurred rounded rectangles shouldn't be too hard. The code linked above is simple and efficient, but not especially accurate (you'll have curvature discontinuities) if you look at isolines. A better approach is the [blurred rounded rectangle](https://raphlinus.github.io/graphics/2020/04/21/blurred-rounded-rects.html) technique from my blog a few years back.

Tactically, I think the best way to implement this is as another type of gradient. The PR to add conical gradients (#435) is probably a good guideline for what's involved. In brief, getting the gradient parameters encoded, doing processing (roughly analogous to a vertex shader) in draw_leaf, and of course computing the alpha values (analogous to a fragment shader) in fine.

This is a great issue for someone to get up to speed on the Vello codebase, help would be appreciated.

msiglreith:
I would take a stab at implementing as it seems pretty straightforward and I'm also in need for it 👀

PR #665, opened 2024-08-15
Add support for blurred rounded rectangle.

msiglreith:
Implements the blurred rounded rectangle approximation outlined in https://raphlinus.github.io/graphics/2020/04/21/blurred-rounded-rects.html. Closes https://github.com/linebender/vello/issues/640

The convolution is mostly taken verbatim from https://git.sr.ht/~raph/blurrr translated to wgsl. Added input bounds for the `erf` function due to artifacts at `std_dev < 0.1`.

Possible optimizations would include moving parts of the precomputation to the CPU side, right now each tile recalculates the same values.

Scene output:
grafik


vello/src/scene.rs
```diff
@@ -106,6 +106,37 @@ impl Scene {
         self.encoding.encode_end_clip();
     }
 
+    /// Draw a rounded rectangle blurred with a gaussian filter.
+    pub fn draw_blurred_rounded_rect(
+        &mut self,
+        transform: Affine,
+        brush: Color,
+        size: impl Into<Size>,
```
DJMcNab, line 114: I'd *probably* expect to see this be a `Rect`, to create flexibility in where the origin is (even if internally, we just make it use a transform to get to the current format).

At the moment, it's quite unclear to me where the origin actually is, except for by looking at the example

vello/src/scene.rs
```diff
@@ -106,6 +106,37 @@ impl Scene {
...
+        radius: f64,
```
DJMcNab, line 115: Not because I expect you to do this, but because it's important that it's raised, but is it at all plausible for the corner radii to be different? I.e. make this function accept a [`RoundedRect`](https://docs.rs/kurbo/latest/kurbo/struct.RoundedRect.html) - which despite the docs, can have different radii on each corner since https://github.com/linebender/kurbo/pull/166

I don't have a good understanding of the underlying maths to make this work

vello_encoding/src/encoding.rs
```diff
@@ -433,6 +433,26 @@ impl Encoding {
             }));
     }
 
+    // Encodes a blurred rounded rectangle.
```
DJMcNab, line 436:
```suggestion
    // Encodes a blurred rounded rectangle brush.
```

examples/scenes/src/test_scenes.rs
```diff
@@ -1686,4 +1688,34 @@ mod impls {
...
     }
+
+    pub(super) fn blurred_rounded_rect(scene: &mut Scene, params: &mut SceneParams) {
+        scene.fill(
+            Fill::NonZero,
+            Affine::IDENTITY,
+            Color::rgba8(255, 255, 255, 255),
+            None,
+            &Rect::new(0.0, 0.0, 1000.0, 1000.0),
+        );
```
DJMcNab, line 1699:
```suggestion
    params.base_color = Some(Color::WHITE);
```

examples/scenes/src/test_scenes.rs
```diff
@@ -1686,4 +1688,34 @@ mod impls {
             splash_screen(scene, params);
         }
     }
+
+    pub(super) fn blurred_rounded_rect(scene: &mut Scene, params: &mut SceneParams) {
```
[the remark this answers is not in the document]
DJMcNab: I'd consider also validating a few edge cases. Namely:
1) A skewed affine transform
2) A stretch affine transformation
3) A circle (i.e. a square with radius half the side length)
4) An "overflowing" radius. E.g. a `(500., 240.)` circle with a radius of `200`.

If you're doing this, you should add a resolution to the test (`params.resolution = ...`)

I have tested these, and they work as expected (with the mental model of a clamped radius)

vello_shaders/shader/coarse.wgsl
DJMcNab, line 1: I think all of the shader logic (except for that in `fine`) needs to be duplicated in the CPU shaders. At the moment, they panick on the test scene.

DJMcNab (commented):
This is looking really good, and works well!

The main blocking concern is copying the logic into the CPU shaders

vello/src/scene.rs
msiglreith: I thought about this as well but seems tricky at least.
If it would be dependent on the (continuous) distance alone it would be fine in my opinion but the parameters (e.g squircle shape) derived from the corner radius would introduce dicontinuities at the quadrant boundaries, probably visible at larger filter lengths. We could derive shared parameters for the whole shape with some tradeoff in quality but without larger adjustments to the approach.

DJMcNab: Yeah - that's fine. I've not thought about it deeply, and I don't think that it is a necessary feature.

Thought it would be important to at least ensure that it is discussed, as much for when Raph comes through to gut-check this.

msiglreith:
Thanks for the review! I added the snapshot tests along with support in the cpu pipeline (besides fine rasterization)

DJMcNab:
One last thought: I'd like to see this used in stats displays for background. I've put my implementations on Zulip, because I'm running into issues.

Otherwise, we're just waiting on Raph for licensing confirmation on the code used from him, I think

msiglreith:
thanks for testing, should be fixed. A fill style was missing resulting in the rect path being interpreted as stroke.

vello/src/scene.rs
nicoburns: > I don't think that it is a necessary feature.

FWIW, this definitely is a necessary feature for CSS box-shadow support. Not sure how common they are in the wild, but the spec certainly allows authors to specify such shadows.

jrmoulton:
I've been testing this in my vello test branch for floem. 

Works great!

It would be nice to have some docs for the std dev arg. I can see that it affects blur but that's as much as I can see. I imagine most people will be approaching this API from a web-dev background and looking for blur and won't have an understanding of the terminology in terms of gaussian blur.

vello/src/scene.rs
jkelleyrtp: https://github.com/DioxusLabs/blitz/blob/main/packages/blitz/src/renderer/multicolor_rounded_rect.rs

we implemented the maths here, you can just copy it if you need to

raphlinus: Just rounded rect can be done without too much difficulty in a distance field context by, basically, setting the radius by quadrant. But in the blurred case you will definitely see discontinuities depending on the exact parameters. I think it is possible to fix, but will require some doing.

Of course another possibility is to render the rounded rect as a vector shape (as in the code Jon linked) and blur it, but (a) that's going to be slower than doing it in a shader, even with heavier math, and (b) we don't have the infrastructure in place. It would be nice to have though, as you do need that for text shadows.

I don't recommend using clips for this, it adds complexity and possible performance issues, as well as potential conflation artifacts.

vello/src/scene.rs
```diff
@@ -106,6 +106,36 @@ impl Scene {
...
+        &mut self,
+        transform: Affine,
+        rect: Rect,
+        brush: Color,
+        radius: f64,
+        std_dev: f64,
+    ) {
+        // The impulse response of a gaussian filter is infinite.
+        // For performance reason we cut off the filter at some extent where the response is close to zero.
+        let kernel_size = 4.0 * std_dev;
```
raphlinus, line 120: I believe 2.5 is adequate here, and didn't see any artifacts when I tweaked it.

raphlinus (approved):
This looks great, thanks! I'm not sure exactly what needs to be done to fix the snapshot issue. Also, there's been a bit more discussion about multiple radii, that would be a good followup issue (or, better, PR).

vello/src/scene.rs
msiglreith: I quickly tested how the artifacts look like and there are indeed visible discontinuities at larger kernel sizes (see figure below). It seems to be primarily caused by the `scale` factor. Using a fixed or interpolated value would be an option here. But I would postpone it for now as followup.
grafik

DJMcNab:
@msiglreith a reminder that the policy is author-merged, and I believe you have permission to press the button.

PR #671, opened 2024-08-24
Add SDL2 & C-powered window handling compatibility.

TheNachoBIT:
This pull request adds `create_surface_unsafe()`, that allows you to add unsafe surface targets.

This allows Vello to be used in SDL2, and it should also work in other libraries like GLFW, but that's not tested yet.

Here's a screenshot of Vello running in SDL2, i'll push an example of it soon :D.

image

### DIFF
--- a/examples/simple_sdl2/src/main.rs
+++ b/examples/simple_sdl2/src/main.rs
@@ -33,7 +33,9 @@ pub fn main() {
 
     let surface_future = unsafe {
         context.create_render_surface(
-            context.instance.create_surface_unsafe(wgpu::SurfaceTargetUnsafe::from_window(&window).unwrap()),
+            context
+                .instance
+                .create_surface_unsafe(wgpu::SurfaceTargetUnsafe::from_window(&window).unwrap()),
             width,
             height,
             wgpu::PresentMode::AutoVsync,
--- a/examples/simple_sdl2/src/main.rs
+++ b/examples/simple_sdl2/src/main.rs
@@ -40,7 +40,8 @@ pub fn main() {
         context.create_render_surface(
             context
                 .instance
-                .create_surface_unsafe(wgpu::SurfaceTargetUnsafe::from_window(&window).unwrap()),
+                .create_surface_unsafe(wgpu::SurfaceTargetUnsafe::from_window(&window).unwrap())
+                .unwrap(),
             width,
             height,
             wgpu::PresentMode::AutoVsync,
--- a/examples/simple_sdl2/src/main.rs
+++ b/examples/simple_sdl2/src/main.rs
@@ -1,0 +1,146 @@
+extern crate sdl2;
+
+use sdl2::event::Event;
+use sdl2::keyboard::Keycode;
+use std::time::Duration;
+
+use std::num::NonZeroUsize;
+
+use vello::util::{RenderContext, RenderSurface};
+use vello::{AaConfig, DebugLayers, Renderer, RendererOptions, Scene};
+use vello::kurbo::{Affine, Circle, Ellipse, Line, RoundedRect, Stroke};
+use vello::peniko::Color;
+
+use vello::wgpu;
+
+use pollster;
+
+pub fn main() {
+    let sdl_context = sdl2::init().unwrap();
+    let video_subsystem = sdl_context.video().unwrap();
+
+    let width: u32 = 800;
+    let height: u32 = 600;
+
+    let window = video_subsystem.window("Vello SDL2 Demo", width, height)
+        .position_centered()
+        .metal_view()
+        .build()
+        .unwrap();
+
+    let mut context = RenderContext::new();
+
+    let surface_future = unsafe { 
+        context.create_surface_unsafe(
+            wgpu::SurfaceTargetUnsafe::from_window(&window).unwrap(),
+            width,
+            height,
+            wgpu::PresentMode::AutoVsync,
+        )
+    };
+
+    let surface = pollster::block_on(surface_future).expect("Error creating surface.");
+    
+    let mut renderers: Vec<Option<Renderer>> = vec![];
+    
+    renderers.resize_with(context.devices.len(), || None);
+    renderers[surface.dev_id].insert(create_vello_renderer(&context, &surface));
+
+    let mut scene = Scene::new();
+
+    let mut event_pump = sdl_context.event_pump().unwrap();
+
+    'running: loop {
+
+        scene.reset();
+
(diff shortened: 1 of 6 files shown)


TheNachoBIT:
So, i added the SDL2 example, but it seems that the tests can't find the library because you have to download SDL2 first.

The unsafe function works, and SDL2 works aswell if you have it installed, but i don't know how can i make the checks successful, so i'm gonna need a little bit of help with that ;_;

TheNachoBIT:
LET'S GOOOOOOOOOO THE BUNDLED FEATURE WORKED!!!

Now "wasm32" is the only target left and then all of the checks will be done :')

vello/src/util.rs
```diff
@@ -80,6 +80,52 @@ impl RenderContext {
         Ok(surface)
     }
 
+    /// Creates a new unsafe surface for the specified window and dimensions.
+    ///
+    /// # Safety
+    /// This function works exactly like `create_surface()`.
+    /// Treat it like `create_surface()` and you should not have any issues :D.
+    pub async unsafe fn create_surface_unsafe<'w>(
```
DJMcNab, line 88: Let's just expose a safe function which accepts a Surface, and document that `RenderSurface` drops the associated surface when it is dropped.

I.e.
```rust
pub async fn create_render_surface(&mut self, surface: wgpu::Surface<'w>, ...)->Result<RenderSurface<'w>>;
```
That way, we don't need to think about unsafety in the core Vello library.

The existing `create_surface` can be refactored in terms of this.

DJMcNab (commented):
Supporting more use cases is always good, but I don't think this is the right API for this behaviour.

The safety comment on `create_surface_unsafe` is not good, and so I have suggested a better alternative which forces users who are working unsafely to do their own safety work.

Fwiw, it is also intended to be possible to use Vello without using the `util` module, so sdl2 should already be supported.

vello/src/util.rs
TheNachoBIT: [Something like this?](https://github.com/bella-project/vello/blob/309d86c703e8a96c5918cdeeaaa57e10fd49a049/vello/src/util.rs#L59)

[Usage in simple_sdl2.](https://github.com/bella-project/vello/blob/309d86c703e8a96c5918cdeeaaa57e10fd49a049/examples/simple_sdl2/src/main.rs#L34)

vello/src/util.rs
```diff
@@ -46,7 +46,25 @@ impl RenderContext {
         height: u32,
         present_mode: wgpu::PresentMode,
     ) -> Result<RenderSurface<'w>> {
-        let surface = self.instance.create_surface(window.into())?;
+        self.create_render_surface(
+            self.instance.create_surface(window.into()),
```
DJMcNab, line 50:
```suggestion
            self.instance.create_surface(window.into())?,
```

vello/src/util.rs
```diff
@@ -46,7 +46,25 @@ impl RenderContext {
...
+            height,
+            present_mode,
+        )
+        .await
+    }
+
+    /// Creates a new render surface for the specified window and dimensions.
+    pub async fn create_render_surface<'w>(
+        &mut self,
+        surface_result: Result<Surface<'w>, CreateSurfaceError>,
```
DJMcNab, line 61:
```suggestion
        surface: Surface<'w>,
```

vello/src/util.rs
```diff
@@ -46,7 +46,25 @@ impl RenderContext {
...
+        width: u32,
+        height: u32,
+        present_mode: wgpu::PresentMode,
+    ) -> Result<RenderSurface<'w>> {
+        let surface = surface_result?;
+
```
DJMcNab, line 67:
```suggestion
```

examples/simple_sdl2/src/main.rs
```diff
@@ -0,0 +1,147 @@
+// Copyright 2024 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
DJMcNab, line 3:
```suggestion

//! Vello can also be used with non-Winit crates which provide a `RawWindowHandle`.
//! This example uses it with [`sdl2`].
//!
//! Vello however is primarily designed for Xilem, which uses Winit, and so support for non-Winit crates
//! is on a tier-2 basis, i.e. it is checked in CI, but is not rarely manually validated.
```

DJMcNab (approved):
Thanks, this version is much easier to approve, as it keeps unsafe out of the core of Vello.

My one remaining concern is whether we already required a C compiler to develop in this repository. I think it's likely that we did, but it's not something I was previously aware of

I have run the new example, but haven't reviewed it extremely carefully

TheNachoBIT:
@DJMcNab Ayyy! Thank you so much for approving! :D

> My one remaining concern is whether we already required a C compiler to develop in this repository. I think it's likely that we did, but it's

There's no need to have a C compiler unless you're using a C-powered window handler, so stuff that isn't running under SDL2 will compile correctly if you don't have one :D

The only thing missing in the example is the wasm32, because idk if Rust's SDL2 supports `wasm32-unknown-unknown` but it does support `wasm32-unknown-emscripten`.

DJMcNab:
So to clarify what I mean when I say that - it's that if you don't have a C compiler on your system, and you clone the `https://github.com/linebender/vello/` repository, then if you run `cargo clippy` in this repository, that will fail to compile `sdl2`. That's not a blocker, by any means, but it might be a new requirement which we have created which wasn't there before.

DJMcNab:
Why not just add the new example to the `NO_WASM_PKGS` in `.github/ci.yml`?

TheNachoBIT:
> So to clarify what I mean when I say that - it's that if you don't have a C compiler on your system, and you clone the https://github.com/linebender/vello/ repository, then if you run cargo clippy in this repository, that will fail to compile sdl2.

Ooooooh i see, yeah that adds a new requirement to run cargo clippy successfully. Glad it's not a blocker tho c:

> Why not just add the new example to the NO_WASM_PKGS in .github/ci.yml?

Can you do that? Nice! i'll do it in just a sec :D

DJMcNab:
To clarify, the suggestions need to be applied or rebuffed before this is merged.

TheNachoBIT:
Of course! I'll apply those changes! Gimmie a sec rq :D

TheNachoBIT:
I'm glad that i'm not the only one that constantly forgets to run clippy lmao

Let me fix it rq c:

TheNachoBIT:
There ya go, i think *that* should've be it.

PR #673, opened 2024-08-28
Properly write-protect segments

DJMcNab:
Extracted from #606. Before this, we would still render even if the segments buffer was being overflowed.

I believe that this was missed in #537

### DIFF
--- a/vello/src/render.rs
+++ b/vello/src/render.rs
@@ -429,7 +429,7 @@ impl Render {
         recording.dispatch(
             shaders.path_tiling_setup,
             wg_counts.path_tiling_setup,
-            [bump_buf, indirect_count_buf.into(), ptcl_buf],
+            [config_buf, bump_buf, indirect_count_buf.into(), ptcl_buf],
         );
         recording.dispatch_indirect(
             shaders.path_tiling,
--- a/vello/src/shaders.rs
+++ b/vello/src/shaders.rs
@@ -194,7 +194,7 @@ pub(crate) fn full_shaders(
             Buffer,
         ]
     );
-    let path_tiling_setup = add_shader!(path_tiling_setup, [Buffer, Buffer, Buffer]);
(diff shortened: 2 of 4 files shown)


DJMcNab:
@94bryanr, does this resolve the issue you were seeing?

94bryanr:
I am still seeing the visual artifact of nothing rendered below a certain line when the resolution is stretched. This is with the new configurations DebugLayers set to none and MemoryHints set to Performance.

DJMcNab:
We probably need code which reproduces this issue to be able to help

dominikh:
Just editing the `simple` example to fill a shape of a large size, say 8000x8000, will demonstrate the problem when rendering to a window roughly larger than 4096x4096 (it really depends on the area). At some point, new tiles stop being rendered.

DJMcNab:
Just done some experiments, and have been able to reproduce, and isolated this to us only supporting 256 bins. I've opened #680 to track this.

However, the issue @94bryanr seems to be a separate issue again, interestingly.
We need a minimal reproduction for that issue as well. I have explained why it looks like a different issue in #680.

raphlinus:
There is a reason we don't have a check in path_tiling, though it's possible I missed something. Essentially, the number of tiles written in path_tiling should always be less than or equal to the count determined in the path_count stage. That, in turn, is checked against the buffer size in [coarse](https://github.com/linebender/vello/milestone/2), and then downstream of that, path_tiling_setup [zeroes out](https://github.com/linebender/vello/milestone/2) the dispatch of path_tiling, so effectively it doesn't run.

If my analysis is correct, then we probably should close this without merging, but I am very open to an argument that something might still go wrong that this check would mitigate.

DJMcNab:
This was discussed in [#gpu > Segments and seg_counts buffer sizes](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Segments.20and.20seg_counts.20buffer.20sizes). We decided that follow-up work is to make the segments and seg_counts buffer use the same capacity (even though they have different bumps).

I'll do that in a different PR.

PR #674, opened 2024-08-28
Experiments on frame pacing in Android

DJMcNab:
This is PR has two things:
1) The proof of concept integration into `with_winit` of gfx-rs/wgpu#6149
2) The start of a frame pacing library using this integration. This is targeted only at Android to reduce scope/risk, but similar ideas will ultimately need to be implemented for other backends. The current state of this library is as a set of ideas, with no actual implementation work.

See [#gpu > Latency Investigation](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Latency.20Investigation)


examples/scenes/src/images.rs
```diff
@@ -43,7 +43,7 @@ impl ImageCache {
 }
 
 fn decode_image(data: &[u8]) -> anyhow::Result<Image> {
-    let image = image::io::Reader::new(std::io::Cursor::new(data))
+    let image = image::ImageReader::new(std::io::Cursor::new(data))
```
waywardmonkeys, line 46: I was going to do this one as a separate PR ... if you don't mind. (Been looking at what'll happen as we get ready to release a new Vello...)

examples/with_winit/src/lib.rs
```diff
@@ -701,6 +770,8 @@ fn run(
     let debug = vello::DebugLayers::none();
 
     let mut app = VelloApp {
+        present_id: 0,
+        google_display_timing_ext_devices: vec![None; renderers.len()],
```
waywardmonkeys, line 774: Should keep the same order as the declaration and put them in the opposite order under `renderers` ?

examples/scenes/src/images.rs
DJMcNab: Sure! This was just to unblock me as I needed to use `cargo update`

waywardmonkeys: Done in #682.

PR #675, opened 2024-08-28
Fix panic caused by improperly aligned read

timtom-dev:
Under certain circumstances, Vello could attempt an improperly aligned read, which would cause bytemuck to panic.

This uses `bytemuck::pod_read_unaligned` instead of `bytemuck::from_bytes` to achieve the same thing without the possibility of a panic.


DJMcNab (approved):
Thanks!

To clarify for future readers, the reason we haven't seen this before is that it only happens with an allocator which doesn't inadvertently over-align, which the default allocators on most platforms do.

PR #677, opened 2024-08-30
wgpu: Prefix WGPU labels with "vello."

waywardmonkeys:
This makes them easier to identify as being from Vello when embedding into an application.

### DIFF
--- a/vello/src/render.rs
+++ b/vello/src/render.rs
@@ -178,22 +178,25 @@ impl Render {
             // is zero.
             packed.resize(size_of::<u32>(), u8::MAX);
         }
-        let scene_buf = ResourceProxy::Buffer(recording.upload("scene", packed));
+        let scene_buf = ResourceProxy::Buffer(recording.upload("vello.scene", packed));
         let config_buf = ResourceProxy::Buffer(
-            recording.upload_uniform("config", bytemuck::bytes_of(&cpu_config.gpu)),
+            recording.upload_uniform("vello.config", bytemuck::bytes_of(&cpu_config.gpu)),
         );
         let info_bin_data_buf = ResourceProxy::new_buf(
             buffer_sizes.bin_data.size_in_bytes() as u64,
-            "info_bin_data_buf",
+            "vello.info_bin_data_buf",
         );
         let tile_buf =
-            ResourceProxy::new_buf(buffer_sizes.tiles.size_in_bytes().into(), "tile_buf");
-        let segments_buf =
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
For example, in Apple's tools:

image

But when you have your own application and a bunch of pipelines, then it is harder to identify who something belongs to without it being clear in the name.

DJMcNab (approved):
Would it be worth automatically stripping these prefices in `with_winit` for display?

Otherwise, seems reasonable. Should we also apply this for buffers as well?

waywardmonkeys:
Maybe and maybe. I confess to feeling lazy about the first. The second, less so ... but I hadn't run into it yet in the tools. :) Will think about it for a couple of hours!

DJMcNab:
Laziness about the first is very reasonable, fwiw.

waywardmonkeys:
I changed this to use `vello.` as a prefix instead of `vello_`.

I updated it to handle various buffers as well.

For debug stuff, I prefixed with `vello.debug.`

In some cases, I changed a space character to an underscore.

vello/src/render.rs
```diff
@@ -178,9 +178,9 @@ impl Render {
...
             packed.resize(size_of::<u32>(), u8::MAX);
         }
-        let scene_buf = ResourceProxy::Buffer(recording.upload("scene", packed));
+        let scene_buf = ResourceProxy::Buffer(recording.upload("vello.scene", packed));
         let config_buf = ResourceProxy::Buffer(
-            recording.upload_uniform("config", bytemuck::bytes_of(&cpu_config.gpu)),
+            recording.upload_uniform("vello.config", bytemuck::bytes_of(&cpu_config.gpu)),
         );
         let info_bin_data_buf = ResourceProxy::new_buf(
             buffer_sizes.bin_data.size_in_bytes() as u64,
```
DJMcNab, line 186: There are lots of other calls to `new_buf`; why have you changed only the ones you have?

waywardmonkeys: because I wasn't looking at new_buf but `upload` / `upload_uniforms` ... I'll hit the others now, I guess. Thanks!

waywardmonkeys: Fixed and force pushed.

waywardmonkeys:
One more thing it seems ... looking at it now under the Metal tools.

DJMcNab (approved):
This will accrue merge conflicts, and seems fine to land

Issue #680, opened 2024-09-02
Support more than 256 bins


DJMcNab:
Currently, if we try and perform a render with more than 256 bins (e.g. a render larger than 4096 by 4096 pixels), we stop rendering below the 256th bin. 
I believe that this limit only occurs in `binning`, as all the algorithms assume there are fewer bins than the workgroup size of that element.

This was debugged following the trail from https://github.com/linebender/vello/issues/548#issuecomment-2307690939.
I don't really understand how this issue is occurring in that case, as even full screen on a retina 5k Mac screen, there should still be only 240 bins being rendered. Additionally, the video demonstrating this has in-congruent behaviour with this theory, as the "Try the tutorial" button is still appearing. This suggests that user is experiencing a different issue.

DeadPoetSpoon:
If try to render something to a huge picture for printing.

casey:
I think I'm able to reproduce this issue. I'm rendering on a 5120x2880 display, but due to the way my renderer works, textures have to be square, so I'm rendering to a 5120x5120 texture, which is enough to hit this issue.

Additionally, and I believe this is the issue closed as not planned in #548, if I try to render to a 5809x5809 texture, the renderer becomes unresponsive and causes some kind of system-wide GPU hiccup.  I'm on an M3 Sequoia mac. (5808 * 5808 % 256 == 0, which is slightly suspicious.)

cococry:
I think fixing this by adding a Y Dimension to the Work-group counts in the binning and coarse dispatch will solve this issue. This can then be used as a page for bin groups (sized 256).

94bryanr:
Hello, so I ran into #548  again (related to bin overflow) and can explain a little more about what is happening. I was able to at least stop the issue from bringing the entire OS down as #548 and a few other issues report.

First - the problem IS that the bin count is exceeding the 256 limit on MacOS when rendering on 4k screens. This happens because the OS provides a display option for external monitors to scale all of the content rendered to them. This is set in the Display settings as a "Larger Text .... More Space" slider. When the user chooses one of these scaling levels, instead of rendering at a lower resolution then scaling up to 4k, the OS supersamples at ~2x the chosen resolution. So when someone is on a 4k display scaled with one of the "larger text" options (which is pretty common I'd guess - pretty sure this happens by default) the surface vello sees is actually almost 2x higher than 4k resolution. This is what causes the number of bins to exceed 256 and causes the visual artifact where bins lower on the screen appear black.

[...]

PR #681, opened 2024-09-02
Add a failing test and warning for #680

DJMcNab:
See #680 for details.

This test will currently pass, but only because the behaviour is incorrect.

This adds a warning for when this bug would apply, for users

### DIFF
--- a/vello/src/render.rs
+++ b/vello/src/render.rs
@@ -168,7 +168,7 @@ impl Render {
         }
         let cpu_config =
             RenderConfig::new(&layout, params.width, params.height, &params.base_color);
-        // The coarse workgroup counts is the number of bins active.
+        // HACK: The coarse workgroup counts is the number of active bins.
         if (cpu_config.workgroup_counts.coarse.0
             * cpu_config.workgroup_counts.coarse.1
             * cpu_config.workgroup_counts.coarse.2)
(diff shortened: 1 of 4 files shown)


vello/src/render.rs
```diff
@@ -168,6 +168,19 @@ impl Render {
         }
         let cpu_config =
             RenderConfig::new(&layout, params.width, params.height, &params.base_color);
+        // The coarse workgroup counts is the number of bins active.
```
DJMcNab, line 171:
```suggestion
        // HACK: The coarse workgroup counts is the number of active bins.
```

vello_tests/src/snapshot.rs
```diff
@@ -147,14 +147,24 @@ pub fn smoke_snapshot_test_sync(scene: Scene, params: &TestParams) -> Result<Sna
     pollster::block_on(snapshot_test(scene, params, SnapshotDirectory::Smoke))
 }
 
-/// Run a snapshot test of the given scene.
```
waywardmonkeys, line 150: Why remove this?

waywardmonkeys (approved):

vello_tests/src/snapshot.rs
DJMcNab: I honestly have no idea :upside_down_face:

My original plan in this PR was to downscale the image for the test, but I realised just directly testing the behaviour would be better.
(Fwiw, once #680 is fixed, then downscaled tests is *probably* the right approach there)

PR #685, opened 2024-09-10
Attempt add single pass scan.

b0nes164:
Adds single pass scan with new GPU and CPU shader. Removes previous tree based scans.


b0nes164:
Although tests are passing, out of an abundance of caution, I want to add seperate tests specifically for the correctness of the scan. These should act as a litmus test for general compatibility with single-pass monoid techniques, which would be useful as more shaders are converted to single pass.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,425 @@
...
+    member_ix: u32,
+    aggregate: u32,
+    spin_count: ptr<function, u32>,
+    prev: ptr<function, u32>,
+    reduction_complete: ptr<function, bool>,
+    inclusive_complete: ptr<function, bool>
+){
+    let payload: u32 = atomicLoad(&reduced[lookback_ix][member_ix]);
+    let flag_value: u32 = payload & FLAG_MASK;
+    if(flag_value == FLAG_REDUCTION){
```
waywardmonkeys, line 56: This is a change in style. The shaders you're replacing don't seem to use `(...)` around the `if` condition.

There's also a bunch of other style changes with missing spaces before `{` and `(` as well as in comments like above with `//For`.

vello/src/render.rs
```diff
@@ -194,61 +194,35 @@ impl Render {
...
             "vello.tagmonoid_buf",
         );
-        let pathtag_scan = if use_large_path_scan {
-            shaders.pathtag_scan_large
-        } else {
-            shaders.pathtag_scan
-        };
+        let reduced_buf = BufferProxy::new(
+            buffer_sizes.path_reduced.size_in_bytes().into(),
+            "reduced_buf",
```
waywardmonkeys, line 203: This and the one below should be prefixed with "vello." so that it is clear that they're from Vello when using the labels in various GPU tools.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
b0nes164: What is the style guidance for the spacing after the rightmost semicolon `for (var spin_count: u32 = 0u; spin_count < MAX_SPIN_COUNT; )` ?

raphlinus (commented):
Ok, I think I have my head around this now. It feels like this is going to be a workable pattern. I think shared memory usage can be reduced, and there are some stylistic things that can be improved.

My overriding concern is the array structure for the monoid - we somehow have the worst of both worlds where the fields aren't named and there's also repetition. I think it can be improved by leaning in more heavily to arrays.

I didn't try this, but am assuming it works. Also, the main focus of my review is on the scan wgsl shader. I did skim the rest but will rely on other reviewers for any fine details I may have missed.

I'm not approving at this point, but think it can be landed soon.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,425 @@
...
+            agg[3] += other[3];
+            agg[4] += other[4];
+        }
+        workgroupBarrier();
+        sh_scratch[local_id.x] = agg;
+    }
+
+    //Broadcast the results and flag into device memory
+    if local_id.x == WG_SIZE - 1u {
+        if part_ix != 0u {
```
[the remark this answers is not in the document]
b0nes164: I came down on the verbose side of this to avoid an `if` for each flag, but I will change this.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,425 @@
...
+            var red1 = false;
+            var red2 = false;
+            var red3 = false;
+            var red4 = false;
+            
+            //Lookback, with a single thread
+            //Last thread in the workgroup has the complete aggregate
+            if local_id.x == WG_SIZE - 1u {
+                for (var spin_count = 0u; spin_count < MAX_SPIN_COUNT; ) {
+                    //TRANS_IX
```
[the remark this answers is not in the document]
b0nes164: I also thought about looping it, but again decided to be verbose. In general I think some sort of performance benchmarking is necessary, as I am afraid the issues I mentioned previously with naga could cause performance regressions.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,425 @@
...
+                        }
+                    } else {
+                        spin_count++;
+                    }
+                }
+
+                //If we didn't complete the lookback within the allotted spins,
+                //prepare for the fallback by broadcasting the lookback tile id
+                //and states of the tagmonoid struct members
+                if sh_lock == LOCKED {
```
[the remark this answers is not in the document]
b0nes164: As the same thread is read/writing from shared memory, this should not be a RAW.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,425 @@
...
+let FLAG_MASK: u32 = 3;
+
+//For the decoupled fallback
+let MAX_SPIN_COUNT: u32 = 4;
+let LOCKED: u32 = 1;
+let UNLOCKED: u32 = 0;
+
+var<workgroup> sh_broadcast: u32;
+var<workgroup> sh_lock: u32;
+var<workgroup> sh_scratch: array<array<u32, 5>, WG_SIZE>;
```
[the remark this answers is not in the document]
b0nes164: I think it's possible to do this without any additional barriers, but it might be a little ugly.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -24,69 +24,49 @@ let LG_WG_SIZE = 8u;
...
+
+//TODO: There has to be a better way to initialize an array?
+fn clear_pathtag()->array<u32, PATH_MEMBERS>{
+    var a: array<u32, PATH_MEMBERS>;
+    a[0] = 0u;
+    a[1] = 0u;
+    a[2] = 0u;
+    a[3] = 0u;
+    a[4] = 0u;
+    return a;
```
DJMcNab, line 59:
```
return array(0u, 0u, 0u, 0u, 0u);
```
should also work. I don't think there's a nice syntax for a zero-initialised array.

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,271 @@
...
+                //Is there an alternative to this besides a giant switch statement or
+                //5 individual reductions?
+                let f_word = scene[config.pathtag_base + local_id.x + fallback_ix * WG_SIZE];
+                var f_agg: pathtag_wrapper;
+                f_agg.p = reduce_tag_arr(f_word);
+                sh_scratch[local_id.x] = f_agg.p;
+                for (var i = 0u; i < LG_WG_SIZE; i += 1u) {
+                    workgroupBarrier();
+                    let index = local_id.x - (1u << i);
+                    if index >= 0u {
```
DJMcNab, line 201: `>=0u` is suspect. E.g.:
```rust
fn main(){
    let index: u32 = 10;
    if index >= 0u32 {
    }
}
```
fails `clippy::absurd_extreme_comparisons`

DJMcNab (commented):
Some points from browsing

vello_shaders/shader/pathtag_scan_csdldf.wgsl
```diff
@@ -0,0 +1,259 @@
...
+                                inc_complete.s[i] = true;
+                            }
+                        }
+                    }
+
+                    //At this point, the reductions are guaranteed to be complete,
+                    //so try unlocking, else, keep looking back
+                    var all_complete = inc_complete.s[0];
+                    for (var i = 1u; i < PATH_MEMBERS; i += 1u) {
+                        all_complete = all_complete && inc_complete.s[i];
```
raphlinus, line 228: I think you can write this `&=`, but it's a very minor stylistic point.

raphlinus (approved):
Thanks for addressing my comments. I also verified that it works (on Apple M1) with the test scenes. I think it's ready to go in, though I didn't go over it with a fine-toothed comb.

DJMcNab:
To reduce the risk from this, I'd propose that we land this after we release [version 0.3.0](https://github.com/linebender/vello/milestone/2). I'm hopeful that we can do this after office hours this week, but we'll see.

b0nes164:
Following up on my conversation with Raph, FXC is still blocking on this:

In the looped version FXC is giving the following errors:

```
vello.pathtag_scan_csdldf(345,45-67): warning X3550: array reference cannot be used as an l-value; not natively addressable, forcing loop to unroll
vello.pathtag_scan_csdldf(310,21-31): error X3511: unable to unroll loop, loop does not appear to terminate in a timely manner (101 iterations) or unrolled loop is too large, use the [unroll(n)] attribute to force an exact higher number
vello.pathtag_scan_csdldf(299,9-19): error X3511: forced to unroll loop, but unrolling failed.
```

This is because the loop inside the lookback/fallback block cannot be unrolled.

Going back to the pre-looping implementation, FXC fails to compile due to uniformity analysis:
```
Internal error: FXC D3DCompile error (0x80004005): pathtag_scan_csdldf(484,29-61): error X4026: thread sync operation must be in non-varying flow control, due to a potential race condition this sync is illegal, consider adding a sync after reading any values controlling shader execution at this point
pathtag_scan_csdldf(548,26-43): error X4026: this variable dependent on potentially varying data: local_id
pathtag_scan_csdldf(664,30-76): error X4026: this variable dependent on potentially varying data: _e992
pathtag_scan_csdldf(451,30-36): error X4026: this memory access dependent on potentially varying data
pathtag_scan_csdldf(452,22-36): error X4026: this variable dependent on potentially varying data: _e565
pathtag_scan_csdldf(473,29-40): error X4026: this variable dependent on potentially varying data: loop_init_1
pathtag_scan_csdldf(479,30-47): error X4026: this variable dependent on potentially varying data: _e609
pathtag_scan_csdldf(472,21-31): error X4026: this loop dependent on potentially varying data
pathtag_scan_csdldf(473,29-40): error X4026: this variable dependent on potentially varying data: loop_init_1
```
It considers the barrier on [line 107](https://github.com/linebender/vello/pull/685/files#diff-e5907d2585aa514e1b15cf91a83a6b622e7ed6ab82237ad315facffb3b62dc5dR107) illegal because the control flow is dependent on shared memory.

raphlinus:
Ah, very interesting. Looking at the code, I'm surprised this runs at all, the WGSL uniformity analysis should reject it (if it doesn't, that's a bug in naga). I haven't tried it, but I'd expect Chrome/tint to also reject it.

Fortunately, it should be fairly easily fixable. Instead of using `sh_lock` directly in the loop predicate, do `let lock = workgroupUniformLoad(&sh_lock)` and predicate off that (this intrinsic also contains a barrier, so the separate `workgroupBarrier` can be eliminated).

I don't have my head around the loop unrolling yet. This is just `PATH_MEMBERS`, no? In which case you'd think the compiler would be able to unroll it a known number of times. One thing I noticed: we should now be able to use `const` instead of `let` (when the shaders were originally written, naga didn't support `const`), but I'd be surprised if that made a difference at the FXC level.

b0nes164:
After changing `sh_lock` to use `workgroupUniformLoad`, the shader does indeed pass uniformity analysis. I also had to change the read on [line 170](https://github.com/linebender/vello/pull/685/files#diff-e5907d2585aa514e1b15cf91a83a6b622e7ed6ab82237ad315facffb3b62dc5dR170). 

As for the loop unrolling, I think you're dead on with regard to the naga transpilation. Even after changing to `const`, subbing literals into the loops, and changing the `spin_count` loop to a fixed number, it's still complaining about loop unrolling, and in places that should be fine, like the initial local scan. Using the naga cli, it does indeed compile to `while` loops. 

With regard to metal, I had the same exact performance issues. I think I was getting like 60% of the speed from wgsl->naga->msl as opposed to glsl->glslangvalidator->SPIR-V->spirv_cross->msl, and that's *after* I unrolled the device level loads/stores by hand.

DJMcNab:
What's the status of this PR? It looks to me that we're at least waiting for the change described in https://github.com/linebender/vello/pull/685#issuecomment-2372509910 to land?

b0nes164:
Naga emitting `while` loops instead of `for` loops is still blocking. This is breaking shader compilation on FXC, and causing general suboptimal performance on metal.

 I reimplemented my prefix sum library in Dawn/Tint to gather data on how significant the slowdown is, and as the data comes in I'll use it to open up an issue requesting the change in naga. That might take a while though.

A more immediate option is to revert the `for` loop changes, as that would fix FXC's dynamic indexing problem.

DJMcNab:
> Naga emitting while loops instead of for loops is still blocking. This is breaking shader compilation on FXC, and causing general suboptimal performance on metal.

Is this being tracked on the wgpu side?

b0nes164:
> Is this being tracked on the wgpu side?

No, I have the dawn/tint results in hand, but I'm waiting for fresh benchmarks on wgpu/naga 23.0. Once I have those, I'll open up an issue.

b0nes164:
See [wgpu#6521](https://github.com/gfx-rs/wgpu/issues/6521)

PR #686, opened 2024-09-10
Optimizations for Winit Examples.

TheNachoBIT:
This optimizes examples by using `ControlFlow::Wait` instead of `ControlFlow::Poll`. This fix was found by Micah Johnston, and this PR implements it in both "simple" and "with_winit" examples. 

- "simple" now runs at 0.1-0.2% CPU usage.
- "with_winit" now runs at 1.0-1.8%.

[You can check more info about it here](https://xi.zulipchat.com/#narrow/stream/147921-general/topic/I've.20made.20a.20Vello.20game.20engine.20to.20prove.20that.20i.20was.20wrong.2E/near/468637207)


examples/with_winit/src/lib.rs
```diff
@@ -168,7 +168,9 @@ struct VelloApp<'s> {
 
 impl<'s> ApplicationHandler<UserEvent> for VelloApp<'s> {
     #[cfg(target_arch = "wasm32")]
-    fn resumed(&mut self, _event_loop: &winit::event_loop::ActiveEventLoop) {}
+    fn resumed(&mut self, event_loop: &winit::event_loop::ActiveEventLoop) {
+        event_loop.set_control_flow(ControlFlow::Wait);
```
madsmtm, line 172: The default is `Wait`, so you could also just not set it

TheNachoBIT: Oh neat! I'll change it once i'm available

TheNachoBIT: Thank u for the info! :D

Cargo.lock
```diff
@@ -3410,3 +3410,11 @@ checksum = "ec866b44a2a1fd6133d363f073ca1b179f438f99e7e5bfb1e33f7181facfe448"
 dependencies = [
  "zune-core",
 ]
+
```
DJMcNab, line 3413: Why do you keep trying to make this change?
We've already had to fix this once in #679

examples/simple/src/main.rs
```diff
@@ -75,14 +75,13 @@ impl<'s> ApplicationHandler for SimpleVelloApp<'s> {
         // Save the Window and Surface to a state variable
         self.state = RenderState::Active(ActiveRenderState { window, surface });
 
-        event_loop.set_control_flow(ControlFlow::Poll);
+        event_loop.set_control_flow(ControlFlow::Wait);
```
DJMcNab, line 78: https://github.com/linebender/vello/pull/686/files#r1752659047 also applies here

examples/simple/src/main.rs
```diff
@@ -109,7 +108,6 @@ impl<'s> ApplicationHandler for SimpleVelloApp<'s> {
             WindowEvent::Resized(size) => {
                 self.context
                     .resize_surface(&mut render_state.surface, size.width, size.height);
-                render_state.window.request_redraw();
```
DJMcNab, line 112: :+1: Requesting redraw when resized is implicit, right?

vello_encoding/Cargo.toml
```diff
@@ -27,4 +27,4 @@ bytemuck = { workspace = true }
 skrifa = { workspace = true, optional = true }
 peniko = { workspace = true }
 guillotiere = { version = "0.6.2", optional = true }
-smallvec = { workspace = true }
+smallvec = { workspace = true }
```
DJMcNab, line 30: Why have you removed this newline?

DJMcNab (approved):
This works for me.

I am slightly concerned that we'll lose the non-vsync behaviour on Wayland if we add a call to [`pre_present_notify`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.pre_present_notify), but that doesn't seem to be an issue for now.

The comments here do need addressing.

examples/simple/src/main.rs
madsmtm: Should be, if not, it's a bug

vello_encoding/Cargo.toml
TheNachoBIT: If i'm being honest, i don't know, i don't even remember lmao, there's no problem if you want that newline back c:

Cargo.lock
TheNachoBIT: I don't know really, i don't touch Cargo.lock so it's probably changed when i run clippy.

When i was working on the SDL2 branch, i had this issue of Cargo.lock being constantly "outdated" when i ran the tests, and the fix was to run clippy, which it'll change the file.

I'm sorry if that caused issues, i'll see if i can fix that on my end.

vello_encoding/Cargo.toml
TheNachoBIT: There, i *think* that it got back to how it was before this PR

examples/with_winit/src/lib.rs
TheNachoBIT: There, i think all of the suggestions are now applied :D

Cargo.lock
DJMcNab: Oh, that's a bit annoying. Thanks for sorting it.

TheNachoBIT: No problem! :D

examples/simple/src/main.rs
TheNachoBIT: Done! :D

DJMcNab (approved):
I'm surprised this actively makes a difference, as we should be spending most of the waiting time in [`get_current_texture`](https://docs.rs/wgpu/latest/wgpu/struct.Surface.html#method.get_current_texture).
But we should never overestimate the capacity of things to be wrong!

Thanks!

TheNachoBIT:
No problem! My pleasure to help! :D

This was mostly a Winit issue.

The problem with `Poll` is that if there's no events, it constantly tries to catch nothing.
With `Wait`, it does nothing if there's no events, leaving room for more important operations to be running, that's why there's a big difference.

DJMcNab:
Oh, so `RedrawRequest` was being deferred somehow? That's janky

PR #688, opened 2024-09-13
Make `sdl2` example optional

DJMcNab:
sdl2 doesn't build (trivially) on Android. This breaks rust-analyzer in this workspace.

To work around this, I propose that we make it not build by default.

We emulate an optional package using cargo features.


DJMcNab:
Never mind - I fixed this with:
```json
  "rust-analyzer.cargo.extraArgs": [
    "--exclude",
    "simple_sdl2"
  ]
```

PR #691, opened 2024-09-19
Remove "experimental" verbiage.

waywardmonkeys:
Vello is in use by others and we're actively encouraging such use.

While Vello has some current shortcomings and issues, that's different from being "experimental".


DJMcNab (approved):
We still have the warning about being an alpha lower down, so this is a reasonable change to make.

DJMcNab:
I've also applied this to the repository description

Issue #650, opened 2024-07-31
Infinitesimal stroked bezier fills whole tile


dominikh:
The scene
```
let mut scene = Scene::new();
let mut p: BezPath = BezPath::default();
p.move_to((399.997, 400.0002));
p.quad_to((399.998, 400.0002), (400., 400.));
scene.stroke(
    &Stroke::default(),
    Affine::IDENTITY,
    Color::rgb(1., 0., 0.),
    None,
    &p,
);
```

renders as a solid square
![image](https://github.com/user-attachments/assets/6bc1f127-6444-45c8-9f19-92a85c1ed5f9)

This seems very sensitive to the coordinates, and many translations of the coordinates don't reproduce the issue. The same issue reproduces with a cubic bezier, with the point `(400., 400.)` repeated.

raphlinus:
It's very likely this is another manifestation of #644, which I believe is a watertightness issue in the expanded stroke outlines. It's in my queue to look at.

raphlinus:
I'm not able to repro directly (or #616 either) from the given example, but I do see an artifact with the following:

```rust
        let mut p: BezPath = BezPath::default();
        p.move_to((399.997, 400.0002));
        p.quad_to((399.998, 400.0002), (400., 400.));
        p.line_to((390., 390.));
        let a = Affine::translate((31.90625, 23.98875));
        scene.stroke(
            &Stroke::new(2.0).with_caps(Cap::Butt),
            a,
            Color::WHITE,
            None,
            &p,
        );
```

It definitely appears to be a watertightness bug, and I suspect it's the same underlying issue. I should be able to track it down now that I've got a repro.

Curious why the variation in reproducibility, but we'll most likely get to that when we have a proposed fix.

DJMcNab:
I suspect the difference could be a fast-math issue.

That's my general heuristic to explain if something is different on Mac :P

raphlinus:
I have a tentative analysis and proposed fix for this particular case. Basically the tangent computed for the start cap is not bit-identical to the tangent for the parallel curves (even though mathematically it's the same). Change line 847 of flatten.rs from:

```rust
                    let tangent = cubic_start_tangent(pts.p0, pts.p1, pts.p2, pts.p3);
```

to

```rust
                    let tangent = pts.p3 - pts.p0;
```

I'm not sure this fixes everything (still investigating), but did want to send an update.

raphlinus:
[...]

PR #695, opened 2024-09-20
Make tangents at subpath end watertight

raphlinus:
At subpath end, the last path segment is encoded with the end point chosen so that the vector from the current point to the end point is the tangent, ie the same vector as cubic_start_tangent applied to the first segment in the subpath. In those cases, compute the tangent by subtracting those two points, rather than cubic_start_tangent applied to the line. These are mathematically identical, but may give different results because of roundoff.

There are two cases: an open subpath, in which case the case is applied to the tangent at draw_start_cap, or an open subpath, where the logic is in the tangent calculation in read_neighboring_segment.

Both GPU and CPU shaders are updated. It hasn't been carefully validated

### DIFF
--- a/vello_shaders/shader/flatten.wgsl
+++ b/vello_shaders/shader/flatten.wgsl
@@ -795,7 +795,10 @@ fn read_neighboring_segment(ix: u32) -> NeighboringSegment {
     let is_closed = (tag.tag_byte & PATH_TAG_SEG_TYPE) == PATH_TAG_LINETO;
     let is_stroke_cap_marker = (tag.tag_byte & PATH_TAG_SUBPATH_END) != 0u;
     let do_join = !is_stroke_cap_marker || is_closed;
-    let tangent = cubic_start_tangent(pts.p0, pts.p1, pts.p2, pts.p3);
+    var tangent = pts.p3 - pts.p0;
+    if !is_stroke_cap_marker {
+        tangent = cubic_start_tangent(pts.p0, pts.p1, pts.p2, pts.p3);
+    }
     return NeighboringSegment(do_join, tangent);
 }
 
@@ -844,7 +847,7 @@ fn main(
             if is_stroke_cap_marker {
                 if is_open {
                     // Draw start cap
-                    let tangent = cubic_start_tangent(pts.p0, pts.p1, pts.p2, pts.p3);
+                    let tangent = pts.p3 - pts.p0;
                     let offset_tangent = offset * normalize(tangent);
                     let n = offset_tangent.yx * vec2f(-1., 1.);
                     draw_cap(path_ix, (style_flags & STYLE_FLAGS_START_CAP_MASK) >> 2u,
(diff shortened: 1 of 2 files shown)


dominikh:
> There are two cases: an open subpath, in which case the case is applied to the tangent at draw_start_cap, or an open subpath, where the logic is in the tangent calculation in read_neighboring_segment.

I think this is erroneously saying "open subpath" twice.

dominikh:
I can confirm that with the new `flatten.wgsl`, #650 is fixed for me. I also didn't notice any incorrect rendering in any of my (manual) testing.

raphlinus:
> I think this is erroneously saying "open subpath" twice (in the PR description as well as the commit message).

Thanks, the second one should be "closed," and I've fixed that in the comment above. Glad to hear this works!

DJMcNab (approved):
This works for me on the regression test. I'd like to see an actual test added. I.e.:
```patch
diff --git a/vello_tests/src/lib.rs b/vello_tests/src/lib.rs
index e97910593..8ee990ceb 100644
--- a/vello_tests/src/lib.rs
+++ b/vello_tests/src/lib.rs
@@ -16,6 +16,7 @@ use vello::wgpu::{
     self, BufferDescriptor, BufferUsages, CommandEncoderDescriptor, Extent3d, ImageCopyBuffer,
     TextureDescriptor, TextureFormat, TextureUsages,
 };
+use vello::AaConfig;
 use vello::{block_on_wgpu, util::RenderContext, RendererOptions, Scene};
 
 mod compare;
@@ -32,6 +33,7 @@ pub struct TestParams {
     pub base_colour: Option<Color>,
     pub use_cpu: bool,
     pub name: String,
+    pub anti_aliasing: AaConfig,
 }
 
 impl TestParams {
@@ -42,6 +44,7 @@ impl TestParams {
             base_colour: None,
             use_cpu: false,
             name: name.into(),
+            anti_aliasing: AaConfig::Area,
         }
     }
 }
@@ -87,7 +90,7 @@ pub async fn get_scene_image(params: &TestParams, scene: &Scene) -> Result<Image
             surface_format: None,
             use_cpu: params.use_cpu,
             num_init_threads: NonZeroUsize::new(1),
-            antialiasing_support: vello::AaSupport::area_only(),
+            antialiasing_support: std::iter::once(params.anti_aliasing).collect(),
         },
     )
     .or_else(|_| bail!("Got non-Send/Sync error from creating renderer"))?;
@@ -97,7 +100,7 @@ pub async fn get_scene_image(params: &TestParams, scene: &Scene) -> Result<Image
         base_color: params.base_colour.unwrap_or(Color::BLACK),
         width,
         height,
-        antialiasing_method: vello::AaConfig::Area,
+        antialiasing_method: params.anti_aliasing,
         debug: vello::DebugLayers::none(),
     };
     let size = Extent3d {
```

raphlinus:
I agree we could use better testing, but I'm on the road right now so not in a great position to write and validate tests.

PR #696, opened 2024-09-21
Fix `clippy::too_long_first_doc_paragraph` lint

waywardmonkeys:
This is a newer lint in nightlies.


DJMcNab (approved):
This doc comment is hard for me to parse, but splitting this out doesn't make that worse. Thanks!

PR #697, opened 2024-09-23
Update Skrifa to 0.22.0

DJMcNab:
The main consequence of this is that there is a new auto-hinter.

I've also removed the `glyph` module from Vello, as the actual code in there is never used, as far as I can see.


vello_encoding/src/glyph_cache.rs
```diff
@@ -245,13 +245,19 @@ pub(crate) struct HintKey<'a> {
...
 }
 
-const HINTING_MODE: HintingMode = HintingMode::Smooth {
-    lcd_subpixel: Some(LcdLayout::Horizontal),
-    preserve_linear_metrics: true,
+const HINTING_OPTIONS: HintingOptions = HintingOptions {
+    engine: skrifa::outline::Engine::AutoFallback,
+    target: skrifa::outline::Target::Smooth {
+        mode: skrifa::outline::SmoothMode::Lcd,
+        // TODO: Set based on whether the anti-aliasing mode is Area?
```
dfrg, line 256: I believe we should always set this to false as you’ve done here. See https://github.com/googlefonts/fontations/issues/1080

vello_encoding/src/glyph_cache.rs
```diff
@@ -245,13 +245,19 @@ pub(crate) struct HintKey<'a> {
...
+        symmetric_rendering: false,
+        // TODO: Does Parley handle layout in a hinting-aware way?
```
dfrg, line 258: Parley expects linear metrics. Hinting aware layout requires scaling glyph outlines which is incredibly expensive.

dfrg (approved):
At some point, we might want to have a separate cache for autohinting instances the same way we do for glyf and cff but this looks good for now, thanks!

PR #699, opened 2024-09-23
Update deps, stable Rust, typos-cli.

waywardmonkeys:


waywardmonkeys:
@DJMcNab I can't check that the updates that are Android specific are okay or not...

PR #700, opened 2024-09-24
Add `draw_blurred_rounded_rect_in` (intended for box shadows)

DJMcNab:
This avoids needing a clip layer for a rounded rectangle which doesn't need to draw the inner parts (e.g. a transparent background with a box shadow).

The previous test image was oversized, so this also fixes that (and makes any future such image fail the tests as expected). I imagine that the author of that test manually copied the value into the place, not realising the size of the image.

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -87,6 +87,7 @@ mod impls {
     use std::f64::consts::PI;
 
     use crate::SceneParams;
+    use kurbo::RoundedRect;
     use rand::Rng;
     use rand::{rngs::StdRng, SeedableRng};
     use vello::kurbo::{
@@ -1719,15 +1720,6 @@ mod impls {
             params.time.sin() * 50.0 + 50.0,
         );
 
-        // Stretch affine transformation.
-        scene.draw_blurred_rounded_rect(
-            Affine::translate((600.0, 600.0)) * Affine::scale_non_uniform(2.2, 0.9),
-            rect,
-            Color::BLACK,
-            radius,
-            params.time.sin() * 50.0 + 50.0,
-        );
-
         // Circle.
         scene.draw_blurred_rounded_rect(
             Affine::IDENTITY,
@@ -1744,6 +1736,29 @@ mod impls {
             Color::BLACK,
             150.0,
             params.time.sin() * 50.0 + 50.0,
+        );
+
+        // An enulated box shadow, to demonstrate the use of `draw_blurred_rounded_rect_in`.
+        let std_dev = params.time.sin() * 50.0 + 50.0;
+        let kernel_size = 2.5 * std_dev;
+
+        // TODO: Add utils to Kurbo for ad-hoc composed shapes
+        let shape = BezPath::from_iter(
+            rect.inflate(kernel_size, kernel_size)
+                .path_elements(0.1)
+                .chain(
+                    RoundedRect::from_rect(rect, radius)
+                        .to_path(0.1)
+                        .reverse_subpaths(),
+                ),
+        );
+        scene.draw_blurred_rounded_rect_in(
+            &shape,
+            Affine::translate((600.0, 600.0)) * Affine::scale_non_uniform(2.2, 0.9),
+            rect,
+            Color::BLACK,
+            radius,
+            std_dev,
         );
     }
 }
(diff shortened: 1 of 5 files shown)


examples/scenes/src/test_scenes.rs
```diff
@@ -1745,5 +1737,28 @@ mod impls {
             150.0,
             params.time.sin() * 50.0 + 50.0,
         );
+
+        // An enulated box shadow, to demonstrate the use of `draw_blurred_rounded_rect_in`.
```
raphlinus, line 1741: I think this should be "emulated"

vello/src/scene.rs
```diff
@@ -125,13 +125,37 @@ impl Scene {
...
+        self.draw_blurred_rounded_rect_in(&shape, transform, rect, brush, radius, std_dev);
+    }
+
+    /// Draw a rounded rectangle blurred with a gaussian filter in `shape`.
+    ///
+    /// For performance reasons, `shape` should not extend more than approximately 2.5 times
+    /// `std_dev` away from the edges of `rect` (as any such points will not be perceptably painted to,
+    /// but calculations will still be performed for them).
+    ///
+    /// This method is a more advanced version of [`Self::draw_blurred_rounded_rect`].
```
raphlinus, line 138: This comment is ok, but I think I might phrase it a little more directly:

``This method effectively draws the blurred rounded rectangle clipped to the given shape. If just the blurred rounded rectangle is desired without clipping, use the simpler [`Self::draw_blurred_rounded_rect`].``

raphlinus (approved):
Very cool! Just a couple comment suggestions.

nicoburns:
I was intending to do a proper review, but my comments would be:

- looks good!
- I assume there is some efficiency advantage to this over a short-lived layer?
- Because it looks like the hard (well, more just awkward) bit (creating the clip shape) is still left to user code.
- It would definitely nice to get some kind of helper for this "rounded rect donut" shape

Finally a question:
- Would this work for inset shadows? I guess it would and I could just use a regular rect as the clip?

DJMcNab:
- thanks
- Yes, I believe so - it reduces the number of segments which need to be processed, and avoids some copies of data in fine.
- Yeah - I didn't want to look into the exact semantics of box shadow's shapes
- I did originally have this utility, but it made some really ugly looking boundaries. I don't think that such a niche shape is a meaningful addition to the imaging model

- Yeah, just using the rounded rectangle as the drawing shape should work, if inset shadows are what I think they are.

nicoburns:
> Yes, I believe so - it reduces the number of segments which need to be processed, and avoids some copies of data in fine.

Awesome.

> Yeah, just using the rounded rectangle as the drawing shape should work, if inset shadows are what I think they are.

Excellent, I'll try this soon :)

> I did originally have this utility, but it made some really ugly looking boundaries. I don't think that such a niche shape is a meaningful addition to the imaging model
> Yeah - I didn't want to look into the exact semantics of box shadow's shapes

Perhaps instead of some generic/composable utility in kurbo there ought to some super-simple `box_shadow` method on `Scene` that just generates `shape` from `rect` as in your example code above and then defers to `draw_blurred_rounded_rect_in`? Which we could add once we've nailed down exactly what the clip path should be for a given rect/shadow I guess.

DJMcNab:
Again, I don't really see the advantage of that - that was part of my original implementation, but I removed it for simplicity.

As you say, that wouldn't work for inset shadows; I don't really think it is very useful.

nicoburns:
My motivation for such a feature is that:

```rust
let kernel_size = 2.5 * std_dev;
let shape = BezPath::from_iter(
    rect.inflate(kernel_size, kernel_size)
        .path_elements(0.1)
        .chain(
            RoundedRect::from_rect(rect, radius)
                .to_path(0.1)
                .reverse_subpaths(),
        ),
);
```

is not particularly simple or obvious to construct to those without knowledge of vector graphics who just want to a draw a box shadow with sensible defaults. It also requires quite a strong knowledge of kurbo to get it that simple (consider the [code we currently have in Blitz for this](https://github.com/DioxusLabs/blitz/blob/main/packages/blitz/src/renderer/multicolor_rounded_rect.rs#L218)).

As the shape can be derived from the rect and shadow parameters, it would be convenient if there was a method that encapsulated that knowledge. Then the API for consumers of Vello would be the same simpler API as that of `draw_blurred_rounded_rect`. For inset shadows we'd need an additional method.

But eh, it's not the most important thing in the world. As-is, this seems like a nice improvement on what we had before.

PR #701, opened 2024-09-26
Add fn encoding_mut() to Scene.

timtom-dev:
This is a useful escape hatch to implement things not supported by the standard API.

### DIFF
--- a/vello/src/scene.rs
+++ b/vello/src/scene.rs
@@ -65,6 +65,11 @@ impl Scene {
         &self.encoding
     }
 
+    /// Returns a mutable reference to the underlying raw encoding.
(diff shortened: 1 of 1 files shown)


vello/src/scene.rs
```diff
@@ -65,6 +65,11 @@ impl Scene {
         &self.encoding
     }
 
+    /// Returns a mutable reference to the underlying raw encoding.
```
DJMcNab, line 68:
```suggestion
    /// Returns a mutable reference to the underlying raw encoding.
    ///
    /// This can be used to more easily create invalid scenes, and so should be used with care.
```

DJMcNab (approved):
I think this API is worth having, but we should discourage its use where possible.

PR #703, opened 2024-09-26
Experimental integration with Achoreographer

DJMcNab:
This is an alternative to #674.

It has significantly better latency performance, with a few peculiar behaviours I can't quite reason about. In particular, it seems to like falling back to 45 fps, but not consistently. More experimentation is needed, I think.

Overall, for the amount of code, the latency reduction is excellent. With this:
trace from Android GPU inspector, showing three well-paced frames
Total latency: ~33ms-43ms

vs without this:
trace from Android GPU inspector, showing three poorly-paced frames. The image is much wider than the previous image
Total latency. Highly variable: ~43ms-53ms

In both of these cases, 16ms is due to compositing, and ~10ms is due to rendering. That is, this integration still leaves some latency on the table, because the callback arrives at vsync, but our rendering doesn't take all of the time the next compositing time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1329,13 +1329,12 @@ dependencies = [
 [[package]]
 name = "ndk"
 version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3f42e7bbe13d351b6bead8286a43aac9534b82bd3cc43e47037f012ebfd62d4"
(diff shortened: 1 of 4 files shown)


DJMcNab:
I've been doing some more thinking about this - this actually needs to be a *component* of/complement to #674.

The Android side docs are kind of terrible, but AChoreographer is extremely useful for us to have.

PR #705, opened 2024-09-26
Match CPU/GPU logic for start tangents

### LINKED ISSUE #704: Watertightness regression in longpathdash
As of #695 we have a regression in longpathdash, generating artifacts consistent with loss of watertightness. The artifacts are present for both butt and round caps.

longpathdash_butt

longpathdash_round

The most likely explanation is a discrepancy between the way tangents are computed on the CPU when encoding the cap and on the GPU. My guess is that the above patch made these match for cubic segments but not for lines.

Thanks to @DJMcNab for noticing; there's a [Zulip thread](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Longpathdash.20-.20artifacts) with a bit of discussion.

raphlinus:
When encoding a start tangent for an end cap or closed segment, use logic designed to match the GPU computed tangent for the first segment.

Note: this changes the GPU calculation to write out the lerp calculation explicitly rather than use the mix intrinsic, so we can rely on the rounding behavior. In the presence of fastmath, the rounding behavior is not guaranteed, but it is verified to work on M1.

Fixes #704, unless we get bitten by fastmath.

### DIFF
--- a/vello_encoding/src/path.rs
+++ b/vello_encoding/src/path.rs
@@ -769,7 +769,7 @@ impl<'a> PathEncoder<'a> {
         Some(pt)
     }
 
-    // Similar to `start_tangent_for_curve` but for a line.
+    /// Similar to [`Self::start_tangent_for_curve`] but for a line.
     fn start_tangent_for_line(&self, p1: (f32, f32)) -> Option<(f32, f32)> {
         let p0 = (self.first_point[0], self.first_point[1]);
         let pt = if (p1.0 - p0.0).abs() > EPSILON || (p1.1 - p0.1).abs() > EPSILON {
--- a/vello_encoding/src/path.rs
+++ b/vello_encoding/src/path.rs
@@ -524,7 +524,7 @@ impl<'a> PathEncoder<'a> {
         }
         if self.state == PathState::MoveTo {
             // Ensure that we don't end up with a zero-length start tangent.
-            let Some((x, y)) = self.start_tangent_for_curve((x, y), None, None) else {
+            let Some((x, y)) = self.start_tangent_for_line((x, y)) else {
                 return;
             };
             self.first_start_tangent_end = [x, y];
@@ -552,7 +552,7 @@ impl<'a> PathEncoder<'a> {
         }
         if self.state == PathState::MoveTo {
             // Ensure that we don't end up with a zero-length start tangent.
-            let Some((x, y)) = self.start_tangent_for_curve((x1, y1), Some((x2, y2)), None) else {
+            let Some((x, y)) = self.start_tangent_for_quad((x1, y1), (x2, y2)) else {
                 return;
             };
             self.first_start_tangent_end = [x, y];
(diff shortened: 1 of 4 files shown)


raphlinus:
The current state of this is "works for me," and I'm feeling a bit uncertain whether it's fully portable. If we *do* get hit by fastmath issues, which is entirely plausible given the weak guarantees of the language, then I'm inclined to take this in a somewhat different direction: revert the encoding changes so that the final segment writes the unmodified point values, but change the GPU-side tangent calculation so that it's based on the encoded values rather than the cubic points after degree raising.

So I'm uploading this hoping to get more feedback. If it generally works, I think we can go forward.

DJMcNab:
As I mentioned [on Zulip](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Longpathdash.20-.20artifacts), I still see some watertightness failures detected by #416 with this PR. Are those out-of-scope for this change?

I would also like to see a snapshot test for the longpathdash test scene added (ideally validated as showing the failure)

raphlinus:
I'll dig into this. I do think fixing watertightness issues is in scope, but also would consider unblocking 0.3 if we have no known regressions.

raphlinus:
As of the most recent commit, the new test is validated by failing on main.

vello_encoding/src/path.rs
```diff
@@ -773,6 +768,39 @@ impl<'a> PathEncoder<'a> {
         };
         Some(pt)
     }
+
+    // Similar to `start_tangent_for_curve` but for a line.
```
DJMcNab, line 772:
```suggestion
    /// Similar to [`Self::start_tangent_for_curve`] but for a line.
```
and similarly elsewhere

DJMcNab (approved):
I can't meaningfully review the maths here, but I can confirm it does resolve the artifacts

:shipit:

PoignardAzur:
Okay, so I'm a bit late to review this, but one nitpick:

This PR says it matches CPU and GPU logic. But the main change in shader code is:

```wgsl
    // Degree-raise
    if seg_type == PATH_TAG_LINETO {
        p3 = p1;
        p2 = p3 + (1.0 / 3.0) * (p0 - p3);
        p1 = p0 + (1.0 / 3.0) * (p3 - p0);
    } else if seg_type == PATH_TAG_QUADTO {
        p3 = p2;
        p2 = p1 + (1.0 / 3.0) * (p2 - p1);
        p1 = p1 + (1.0 / 3.0) * (p0 - p1);
    }
```

Whereas the equivalent code in the CPU shader in `vello_shaders/src/cpu/flatten.rs` is still:

```rust
    // Degree-raise
    if seg_type == PATH_TAG_LINETO {
        p3 = p1;
        p2 = p3.mix(p0, 1.0 / 3.0);
        p1 = p0.mix(p3, 1.0 / 3.0);
    } else if seg_type == PATH_TAG_QUADTO {
        p3 = p2;
        p2 = p1.mix(p2, 1.0 / 3.0);
        p1 = p1.mix(p0, 1.0 / 3.0);
    }
```

That is, same as the old GPU code.

Not sure what's going on here. Probably the CPU shader should be edited to match the change in the GPU shader?

DJMcNab:
Those look the same to me? Rust doesn't have fast math.

PoignardAzur:
Well the GPU side went from using `mix` to explicitly using operators. I assumed there was a reason for that. Maybe I'm reading too much into it.

DJMcNab:
> Note: this changes the GPU calculation to write out the lerp calculation explicitly rather than use the mix intrinsic, so we can rely on the rounding behavior. In the presence of fastmath, the rounding behavior is not guaranteed, but it is verified to work on M1.

From the pr description

PoignardAzur:
I should really read PR descriptions more often.  I don't know when I got the habit of not doing that.

PR #712, opened 2024-10-03
Various docs improvements

DJMcNab:
Adds intra-doc links as appropriate.
Explains anti-aliasing methods.


vello/src/lib.rs
```diff
@@ -132,22 +132,52 @@ use wgpu::{Device, Queue, SurfaceTexture, TextureFormat, TextureView};
...
 /// Represents the antialiasing method to use during a render pass.
+///
+/// Can be configured for a render operation by setting [`RenderParams::antialiasing_method`].
+/// Each value of this can only be used if the corresponding field on [`AaSupport`] was used.
+///
+/// This can be converted into an `AaSupport` using [`Iterator::collect`],
+/// as `AaSupport` implements `FromIterator`.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum AaConfig {
+    /// Area antialiasing, where the proportion of an edge pixel covered by a path determines its contribution.
```
xStrom, line 143:
```suggestion
    /// Area anti-aliasing, where the proportion of an edge pixel covered by a path determines its contribution.
```
There's a mix of *antialiasing* and *anti-aliasing* in this PR and in old comments. Not exactly a blocker, but I would personally do a search/replace and have all documentation call it *anti-aliasing*.

raphlinus: Area anti-aliasing, where the alpha value for a pixel is computed from integrating the winding number over its square area. This technique produces very accurate values when the shape has winding number of 0 or 1 everywhere, but can result in conflation artifacts otherwise.

PR #716, opened 2024-10-04
Run `cargo update && cargo upgrade`

DJMcNab:
The important updates here are the upgrade to Skrifa 0.22.3, and png 0.17.14.

Exactly what that Skrifa update does isn't very clear, since I've not been able to find fontation's changelog (cc @dfrg). This does mean we are testing with what our users will actually use in the imminent 0.3.0 release.


dfrg (approved):

dfrg:
The skrifa changes include full match for FT hinted metrics (which isn’t relevant for us) along with some substantial improvements to TrueType scaling performance. There might be more changes as we prep for expanded launch in Chrome over the next week.

PR #726, opened 2024-10-29
Install wasm-bindgen at a fixed version for deployment

DJMcNab:
This should stop the red cross on GitHub

### DIFF
--- a/.github/workflows/web-demo.yml
+++ b/.github/workflows/web-demo.yml
@@ -24,9 +24,10 @@ jobs:
           targets: wasm32-unknown-unknown
 
       - name: install wasm-bindgen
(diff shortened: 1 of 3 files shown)


xStrom (approved):
I guess it makes sense to pin these so they always match, given that `wasm-bindgen` doesn't seem to follow semver.

DJMcNab:
It's kind of semver, kind of not. `wasm-bindgen` the CLI tool is only supported in lock-step with the same version of `wasm-bindgen` the package. But like, updating them both as an atomic unit follows semver.

It's not clear to me why `wasm-bindgen-up` doesn't exist, though. To just handle that automatically.

PR #734, opened 2024-11-05
Match CI doc testing with docs.rs.

xStrom:
* Remove `-Zunstable-options -Zrustdoc-scrape-examples` from CI. It wasn't even enabled for docs.rs and we don't have any examples by Cargo's definition, so not worth enabling scraping at this point.
* Treat doc warnings as errors. Historically we've not done so due to various `rustdoc` bugs giving false positives but I got it to pass without failure right now, so perhaps better times have arrived.
* Pass `--all-features` at docs.rs to match our CI and reduce the maintenance burden of manually syncing the features list.
* Target only `x86_64-unknown-linux-gnu` on docs.rs as we don't have any platform specific docs.
* Enable the `doc_auto_cfg` feature for docs.rs which will show a little tip next to feature gated functionality informing of the crate feature flag.

### DIFF
--- a/vello/src/wgpu_engine.rs
+++ b/vello/src/wgpu_engine.rs
@@ -38,7 +38,7 @@ pub(crate) struct WgpuEngine {
     #[cfg(not(target_arch = "wasm32"))]
     shaders_to_initialise: Option<Vec<UninitialisedShader>>,
     pub(crate) use_cpu: bool,
-    /// Overrides from a specific `Image`'s [`id`](peniko::Image::id) to a wgpu `Texture`.
+    /// Overrides from a specific `Image`'s [`id`](peniko::Blob::id) to a wgpu `Texture`.
     ///
     /// The `Texture` should have the same size as the `Image`.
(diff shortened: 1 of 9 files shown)


vello/src/wgpu_engine.rs
```diff
@@ -38,7 +38,7 @@ pub(crate) struct WgpuEngine {
     #[cfg(not(target_arch = "wasm32"))]
     shaders_to_initialise: Option<Vec<UninitialisedShader>>,
     pub(crate) use_cpu: bool,
-    /// Overrides from a specific `Image`'s [`id`](peniko::Image::id) to a wgpu `Texture`.
+    /// Overrides from a specific `Image`'s [`id`](peniko::Blob::id) to a wgpu `Texture`.
```
waywardmonkeys, line 41: Feel like this might be better phrased to indicate that the `id` comes from the image's blob ...

waywardmonkeys (approved):

waywardmonkeys (approved):

PR #735, opened 2024-11-06
Update to wgpu 23.0.0

waywardmonkeys:
This also updates the MSRV to 1.76 to match the updated MSRV of wgpu 23.

### DIFF
--- a/vello/src/wgpu_engine.rs
+++ b/vello/src/wgpu_engine.rs
@@ -331,7 +331,7 @@ impl WgpuEngine {
             layout: Some(&pipeline_layout),
             vertex: wgpu::VertexState {
                 module,
-                entry_point: vertex_main,
+                entry_point: Some(vertex_main),
                 buffers: vertex_buffer
                     .as_ref()
                     .map(core::slice::from_ref)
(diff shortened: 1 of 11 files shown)


vello/src/wgpu_engine.rs
```diff
@@ -827,7 +827,7 @@ impl WgpuEngine {
             label: Some(label),
             layout: Some(&compute_pipeline_layout),
             module: &shader_module,
-            entry_point: "main",
+            entry_point: Some("main"),
```
DJMcNab, line 830: I'd like to try:
```suggestion
            entry_point: None,
```
here

DJMcNab (approved):
Barring wgpu-profiler, this looks correct.

Nice that our MSRV checks are working as designed

vello/src/wgpu_engine.rs
waywardmonkeys: Done.

waywardmonkeys:
@DJMcNab This is now failing on macOS only in the `many_bins_gpu` test that you added as part of #680 / #681. Confirmed that it happens locally on this branch and not on current `main`.

Thoughts or pointers?

DJMcNab:
That is extraordinarily bizarre; it seems to have fixed #680, but only on macOS? That doesn't make sense.

I don't really know how to begin debugging this, as I don't have a Mac.
One thing to try is reproducing #680 in an actual window - i.e make a window larger than your screen using the "drag down, expand/resize up" trick.

That test is effectively a reverse regression test, i.e. that if we fix the mentioned issue, we should also close #680.
I'd be reasonably happy to just ignore the test on macOS and move on...

DJMcNab (approved):

PR #737, opened 2024-11-11
Add CI support for doc tests.

xStrom:
As [`cargo-nextest` does not support running doc tests](https://github.com/nextest-rs/nextest/issues/16), we need a `cargo test --doc` step as well.


DJMcNab (approved):
See also https://github.com/linebender/xilem/issues/500

PR #738, opened 2024-11-18
Fix a problem with dependencies which prevents from running examples on MacOS

binadamu-isiyoonekana:
This PR updates some dependencies so that to be able to run examples  (at least on MacOS).

For instance, when trying to execute the `simple` example,  no shape is actually rendered (when executing `cargo run --package simple`) without this update (again, on a MacOS, x86_64)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -216,27 +216,12 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bit-set"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f0481a0e032742109b1133a095184ee93d88f3dc9e0d28a5d033dc77a073f44f"
-dependencies = [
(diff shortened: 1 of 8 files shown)


xStrom:
The update to wgpu 23 is already ongoing in #735. Can you try running examples using that branch? Would be good to know if your issues go beyond the wgpu update.

binadamu-isiyoonekana:
> The update to wgpu 23 is already ongoing in #735. Can you try running examples using that branch? Would be good to know if your issues go beyond the wgpu update.

Ooops... I did not see #735 PR. Sorry. It in fact fixes the same issue. So let me close this PR.

waywardmonkeys:
I'm confused as to why `simple` isn't rendering something and `with_winit` is ... that seems to be a bug that should be looked into separately from the update of `wgpu`?

binadamu-isiyoonekana:
> Right, but the question is "Why is it failing in just the one example?" and going back, it seems to be failing for a while for me too. Is this due to an OS update or ... ?

Gotcha! Now I understand your question. I'll also dig more on my side.

Issue #306, opened 2023-04-12
Cosmic Text example


nicoburns:
It would be great to have an example of using [cosmic-text](https://github.com/pop-os/cosmic-text) with Vello. This would be super-helpful for [Blitz](https://github.com/DioxusLabs/blitz) to get high quality text layout / rendering running on top Vello Long term hopfully [parley](https://github.com/dfrg/parley) will also be an option, but it seems that at it isn't ready yet in that it is lacking both documentation and a crates.io release so for the time being cosmic-text is looking like the best option.

@dfrg I believe you had this working. Would you be willing to publish it as an example?

dfrg:
I’d love to get this in as an example. There is a minor complication around selecting a unique id for the fonts coming from cosmic-text. This can be resolved in a few different ways— I’ll see what I can do. 

The major blocker is having simple infrastructure for spinning up new examples. Right now, we just dump everything into the with_winit crate and I’m not sure if that is the right call here.

axelkar:
Is it possible to do layout with skrifa or rendering with COSMIC Text?

axelkar:
Nevermind that - Do I take individual glyphs from COSMIC Text and just shove them down `vello::DrawGlyphs`?

dfrg:
Yes, the draw glyphs API is the intended entry point for this and my prior experiment had this working with cosmic-text though I’m sure the code is massively out of date at this point.

PR #739, opened 2024-11-19
Add cosmic text example

AustinMReppert:
Add an example for cosmic text.

Related:
[306](https://github.com/linebender/vello/issues/306)


DJMcNab:
Thanks, it's really good to see this.

Would you consider making this a test scene (i.e. in the `scenes` crate?). Possibly behind a feature gate.
The global context of fonts can be part of the shared `SceneParams` .

I think that would be useful because this new example has a lot of interleaving of logic between cosmic text and Vello and winit.
Additionally, it could allow us to use cosmic text in our tests.

If not, I think we can still land this nearly as-is, although it being another usage of winit is likely to cause us pain in future when we need to perform updates.

AustinMReppert:
Sure, I will do that.

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,425 @@
...
+        // Copy fonts from cosmic_text, so vello can use them
+        let font_faces: Vec<(ID, u32)> = font_system
+            .db()
+            .faces()
+            .map(|face| (face.id, face.index))
+            .collect();
+
+        for (font_id, index) in font_faces {
+            if let Some(font) = font_system.get_font(font_id) {
+                let resource = Arc::new(font.data().to_vec());
```
DJMcNab, line 47: There are two allocations here, one to put the data in the vec, and one to allocate the `Arc`. The first could be avoided by creating an adapter which implements `AsRef<[u8]>` by containing an `Arc<Font>` and using the data parameter. The second would use https://docs.rs/bytemuck/latest/bytemuck/allocation/trait.TransparentWrapperAlloc.html#method.wrap_arc, and is definitely out-of-scope.

The best resolution to this would probably be to have a comment: "These allocations are avoidable, but used for simplicity.", maybe with a link to this comment?

examples/scenes/src/lib.rs
```diff
@@ -34,6 +40,8 @@ pub struct SceneParams<'a> {
     pub resolution: Option<Vec2>,
     pub base_color: Option<vello::peniko::Color>,
     pub complexity: usize,
+    #[cfg(feature = "cosmic_text")]
+    pub cosmic_text_scene_state: &'a CosmicTextSceneState,
```
DJMcNab, line 44: This is underdocumented, so I can understand this being unclear, but test scenes can have state. 

The main result of this is that that we can make this feature gate be "localised", by making the `cosmic_text_scene::text_scene` function return an `impl TestScene`, which will be retained.
For an example, see the `MMark` scene.

This does lead to some things which are unidiomatic, such as not sharing the fonts between the scenes, but I think for this peripheral functionality, keeping it as localised as possible is better than improving the efficiency.

examples/scenes/src/test_scenes.rs
```diff
@@ -53,6 +61,7 @@ macro_rules! scene {
 }
 
 export_scenes!(
+    #[cfg(feature = "cosmic_text")] cosmic_text_scene(cosmic_text),
```
DJMcNab, line 64: This can use:
```suggestion
    #[cfg(feature = "cosmic_text")] cosmic_text_scene(crate::cosmic_text_scene::cosmic_text(), "cosmic text", false),
```

examples/scenes/src/test_scenes.rs
```diff
@@ -53,6 +61,7 @@ macro_rules! scene {
...
     splash_with_tiger(impls::splash_with_tiger(), "splash_with_tiger", false),
```
DJMcNab, line 65: The splash screen should still be the first screen, to show the controls to anyone.

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,425 @@
+// Copyright 2024 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
DJMcNab, line 3: A tiny bit of documentation would be good:
```suggestion

//! An example integrating Cosmic text with Vello.
//! This renders a read-only snapshot of a simplified text editor using system fonts.

```

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,425 @@
...
+            });
+        }
+
+        buffer_glyphs.buffer_lines.push(buffer_line);
+    }
+
+    buffer_glyphs
+}
+
+// Copied directly from cosmic_text.
```
DJMcNab, line 359: Can we have a link to the source?

DJMcNab (commented):
This is looking good and runs as expected. Thanks for pulling this together

The exact integration with scenes could be cleaner, given that there's only one scene; if my guidance isn't clear enough, let me know and I can apply it myself.

AustinMReppert:
Thanks for the feedback. I will work on this tomorrow.

examples/scenes/src/cosmic_text_scene.rs
waywardmonkeys: Not in time for this patch, but improving this is also a design goal for splitting these resource handles out of Peniko. (https://github.com/linebender/peniko/pull/68)

examples/scenes/src/test_scenes.rs
AustinMReppert: Done

examples/scenes/src/lib.rs
AustinMReppert: Done.

examples/scenes/src/cosmic_text_scene.rs
AustinMReppert: I have used an adapter and added a link to this comment.

examples/scenes/src/cosmic_text_scene.rs
AustinMReppert: Done

examples/scenes/src/cosmic_text_scene.rs
AustinMReppert: Done

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,452 @@
...
+                    .font_size(buffer_glyphs.font_size)
+                    .brush(glyph_color)
+                    .transform(text_transform)
+                    .draw(Fill::NonZero, glyphs.into_iter());
+            }
+        }
+    }
+}
+
+pub struct CosmicTextSceneState {
```
DJMcNab, line 70: "State" is rarely a meaningful suffix; I think this would be clearer if it were removed.

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,452 @@
...
+            if let Some(font) = font_system.get_font(font_id) {
+                // For now use an adapter, to avoid cloning the entire font data.
+                // For alternatives, see https://github.com/linebender/vello/pull/739#discussion_r1858293718
+                let font_blob = Blob::new(Arc::new(CosmicFontBlobAdapter::new(font)));
+                let vello_font = Font::new(font_blob, index);
+                vello_fonts.insert(font_id, vello_font);
+            }
+        }
+
+        let text = "おはよう (ja) (ohayō) 🌅✨ (morning), こんにちは (ja) (konnichi wa) ☀️😊 (daytime), こんばんは (ja) (konban wa) 🌙🌟 (evening)";
```
DJMcNab, line 121: For future reviewers:

Source: https://github.com/pop-os/cosmic-text/blob/1f4065c1c3399efad58841082212f7c039b58480/sample/hello.txt#L113

These translations do also appear to be correct.

examples/headless/Cargo.toml
```diff
@@ -20,3 +20,6 @@ pollster = { workspace = true }
 env_logger = "0.11.5"
 png = "0.17.14"
 futures-intrusive = { workspace = true }
+
+[features]
+cosmic_text = ["scenes/cosmic_text"]
```
DJMcNab, line 25: Our style is to have a final newline at the end of all files.
(This also applies to the other files where you've chosen to omit this.)

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,452 @@
...
+    Attrs, Buffer, Cursor, Edit, Editor, FontSystem, LayoutRun, Metrics, Selection, Shaping,
+};
+
+use unicode_segmentation::UnicodeSegmentation;
+
+use std::cmp;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+pub struct CosmicData {}
```
DJMcNab, line 24: What is this for?

DJMcNab (commented):
What needs to happen to get this out of draft? It's looking quite good

examples/scenes/src/cosmic_text_scene.rs
AustinMReppert: Removed

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,450 @@
...
+        }
+
+        buffer_glyphs.buffer_lines.push(buffer_line);
+    }
+
+    buffer_glyphs
+}
+
+// Copied directly from https://github.com/pop-os/cosmic-text/blob/58c2ccd1fb3daf0abc792f9dd52b5766b7125ccd/src/edit/editor.rs#L66
+fn cursor_position(cursor: &Cursor, run: &LayoutRun) -> Option<(i32, i32)> {
```
DJMcNab, line 385: I think this is exposed via https://docs.rs/cosmic-text/latest/cosmic_text/struct.Editor.html#method.cursor_position; I'd much prefer we used that if at all feasible.

DJMcNab (commented):
Would you be able to also document that this exists at the top level of the repository, in https://github.com/linebender/vello?tab=readme-ov-file#integrations

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,450 @@
...
+        }
+
+        for glyph in layout_run.glyphs {
+            let mut glyph_color = match glyph.color_opt {
+                Some(color) => Color::from_rgba8(color.r(), color.g(), color.b(), color.a()),
+                None => text_color,
+            };
+
+            if let Some(editor_info) = &editor_info {
+                if text_color.components != editor_info.selected_text_color.components {
```
waywardmonkeys, line 336: This isn't the right way to compare, but I understand that it is the easier for now.

We're addressing this in https://github.com/linebender/color/pull/76 but it'll take some time, so this can get fixed after something else lands and gets propagated here.

examples/scenes/src/cosmic_text_scene.rs
```diff
@@ -0,0 +1,450 @@
...
+                        {
+                            glyph_color = editor_info.selected_text_color;
+                        }
+                    }
+                }
+            }
+
+            if let Some((last_font, last_glyph_color)) = last_font {
+                if last_font != glyph.font_id
+                    || last_glyph_color.components != glyph_color.components
```
waywardmonkeys, line 351: As above.

examples/scenes/src/cosmic_text_scene.rs
nicoburns: A wrapper/adapter is fine, but we should probably just submit a PR to cosmic-text to make `Font` implement `AsRef<[u8]>`. Alternatively, the correct type can be obtained using `unsafe { font_system.db().make_shared_face_data(font_id).unwrap() }`.

examples/scenes/src/cosmic_text_scene.rs
waywardmonkeys: If you rebase on top of the current `main` as of a couple of minutes ago, the color now impl's `PartialEq`, so you can just `if text_color != ...` instead of looking at the components.

waywardmonkeys:
Is there anything left here other than what I just commented on? Or is this basically ready to go? It would be nice to land it soon!

AustinMReppert:
I will rebase this hopefully in the next day or two

examples/scenes/src/cosmic_text_scene.rs
AustinMReppert: That function only returns the top left cursor. I think the drawing assumes that there may be multi-line cursors. Although, the editor currently only has one cursor.

README.md
```diff
@@ -146,6 +146,15 @@ A separate Linebender integration for playing Lottie animations is available thr
 
 A separate Linebender integration for rendering raw scenes or Lottie and SVG files in [Bevy] through [`bevy_vello`](https://github.com/linebender/bevy_vello).
 
+### Cosmic Text
+
+An example scene demonstrating the integration of COSMIC text for font loading and text layout through [COSMIC Text Scene](https://github.com/linebender/vello/blob/main/examples/scenes/src/cosmic_text_scene.rs).
```
DJMcNab, line 151: This should probably use a relative path (so that e.g. people viewing this in their editor get the right path). It also lets us double check the link before merging.
See e.g. the links to the license files at the bottom of this file.

I do see that there are other cases in this file where that isn't done - it's fine not to fix that.

examples/headless/Cargo.toml
DJMcNab: This appears to have re-regressed in the same place.

examples/scenes/src/cosmic_text_scene.rs
DJMcNab: The reason I want that is for licensing. In the form you've currently included it, we would need to include the full System76 MIT license somewhere (or use this section of code only under the Apache 2.0 license, probably moving it into its own file).

vello_tests/Cargo.toml
```diff
@@ -22,3 +22,6 @@ nv-flip = "0.1.2"
 image = { workspace = true, features = ["png"] }
 
 scenes = { workspace = true }
+
+[features]
+cosmic_text = ["scenes/cosmic_text"]
```
DJMcNab, line 27: And here

DJMcNab (commented):
This is still blocked on the licensing question for the newly imported code - I'd still prefer us to just avoid importing that code entirely, but I might be missing something.

DJMcNab:
@xStrom are you willing to review the name licensing things I've done here? I think I've covered it properly

AustinMReppert:
Closing this as not planned.

DJMcNab:
Thanks for your contribution, and sorry that we weren't able to get this reviewed and landed in a timely manner.

> Closing this as not planned.

To be completely clear to future readers, this is "not planned" by the PR author, not "not planned" by the project. We still want an example like this.

PR #740, opened 2024-11-22
Correct the capitalisation of the `png ` tag

DJMcNab:
See https://github.com/googlefonts/fontations/issues/1167#issuecomment-2387032814

Should make linebender/xilem#420 work.

### DIFF
--- a/vello/src/scene/bitmap.rs
+++ b/vello/src/scene/bitmap.rs
@@ -162,7 +162,7 @@ impl<'a> BitmapStrike<'a> {
         match &self.0 {
             StrikeKind::Sbix(sbix, metrics) => {
                 let glyph = sbix.glyph_data(glyph_id).ok()??;
-                if glyph.graphic_type() != Tag::new(b"PNG ") {
+                if glyph.graphic_type() != Tag::new(b"png ") {
                     return None;
                 }
                 let glyf_bb = metrics.bounds(glyph_id).unwrap_or_default();


Philipp-M (approved):

vello/src/scene/bitmap.rs
```diff
@@ -162,7 +162,7 @@ impl<'a> BitmapStrike<'a> {
         match &self.0 {
             StrikeKind::Sbix(sbix, metrics) => {
                 let glyph = sbix.glyph_data(glyph_id).ok()??;
-                if glyph.graphic_type() != Tag::new(b"PNG ") {
+                if glyph.graphic_type() != Tag::new(b"png ") {
```
waywardmonkeys, line 165: It would be nice to note some provenance for this tag. Perhaps https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6sbix.html

DJMcNab: The canonical form of this code doesn't live here; I've intentionally touched it as little as possible. Instead, it can be found at https://github.com/googlefonts/fontations/blob/skrifa-bitmaps/skrifa/src/bitmap.rs.

The place to record this feedback is probably in the first linked issue.

waywardmonkeys:
(I think we need a lot more provenance links in general!)

waywardmonkeys:
This needs a changelog entry.

PR #741, opened 2024-11-22
Update the CHANGELOG

DJMcNab:
If I would evaluate the CHANGELOG before the next release, I would reach the decision that #740 doesn't need a changelog entry However, @waywardmonkeys believes that it does, so the only viable way for this to end up in the changelog is to do it now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ This release has an [MSRV][] of 1.76.
 ### Fixed
 
 - Offset in image rendering, and sampling outside correct atlas area ([#722][] by [@dfrg])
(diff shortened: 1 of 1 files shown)


waywardmonkeys (approved):

waywardmonkeys:
Was going to say it might be nice to reword the commit message, but ah well.

DJMcNab:
Removing a pr from the merge queue is perfectly fine.

I don't *think* this is one which needs to be remediated, but I'm willing to be overruled, and this is a scenario where rewriting history would be feasible.

waywardmonkeys:
I'll keep that in mind about removing it in the future.

No point in bothering to do a force push or anything, it isn't a big deal.

PR #742, opened 2024-11-24
Update to new `peniko` with `color`

waywardmonkeys:


waywardmonkeys:
This is a work in progress showing a version of Vello that is using the new `color` crate and the corresponding updated `peniko`.  Both are git dependencies here (`color` being a transitive one).

There's plenty of stuff to do after the initial compilation, and a few things are still not working here like the code that needs `PartialCmp` for `Color`.

Once this is all working, we can look at getting rid of `ColorF64`  and more!

waywardmonkeys:
When reviewing this, keep in mind the list of things that is already in #749.

CHANGELOG.md
```diff
@@ -18,13 +18,19 @@ This release has an [MSRV][] of 1.82.
...
 ### Fixed
 
 - Offset in image rendering, and sampling outside correct atlas area ([#722][] by [@dfrg])
 - Inference conflict when using Kurbo's `schemars` feature ([#733][] by [@ratmice][])
 - Detection of PNG format bitmap fonts, primarily for Apple systems ([#740][] by [@LaurenzV])
 
+### Color Changes
+
+As part of an initiative to improve color handling across the ecosystem (and especially within Linebender crates), Vello is now using the new [`color`] crate.
+This is the first step towards providing richer color functionality, better handling of color interpolation, and more.
```
DJMcNab, line 32: Maybe this should be in `Highlights`?

examples/headless/src/main.rs
```diff
@@ -145,7 +146,7 @@ async fn render(mut scenes: SceneSet, index: usize, args: &Args) -> Result<()> {
             .args
             .base_color
             .or(scene_params.base_color)
-            .unwrap_or(vello::peniko::Color::BLACK),
+            .unwrap_or(palette::css::BLACK),
```
DJMcNab, line 149: I wonder if it would actually be more idiomatic to import the full `css` palette, i.e. this would be `css::BLACK`.

The `palette` identifier here reads a bit out of place.

vello_tests/snapshots/many_clips.png
DJMcNab, line 1: Nice!

vello_tests/tests/property.rs
```diff
@@ -65,8 +65,8 @@ fn empty_scene(use_cpu: bool) {
     assert_eq!(image.format, Format::Rgba8);
     for pixel in image.data.data().chunks_exact(4) {
         let &[r, g, b, a] = pixel else { unreachable!() };
-        let image_color = Color::rgba8(r, g, b, a);
-        if image_color != color {
+        let image_color = Color::from_rgba8(r, g, b, a);
+        if image_color.premultiply().difference(color.premultiply()) > 1e-4 {
```
DJMcNab, line 69: Why this premultiply step?

DJMcNab (approved):
I've not actually run this, but that's what tests are for :upside_down_face:

CHANGELOG.md
waywardmonkeys: Moved (part of next push).

vello_tests/tests/property.rs
waywardmonkeys: There's no `AlphaColor::difference` because of perceptual difference vs alpha.

waywardmonkeys: (I've previously pasted this code to Raph with no objections.)

examples/headless/src/main.rs
waywardmonkeys: I've gone back and forth with this a couple of times in my changes. This indicates that it is coming from a palette of colors and not that it has anything to do directly with CSS.

waywardmonkeys: If we aren't careful, I'll end up with `use peniko::color::palette::css as css_palette;` !

DJMcNab: I read `css::BLACK` as `CSS black`, i.e. black"from CSS", so as defined in CSS. It seems like Bevy mostly always imports them fully, so would use `BLACK` here.

(The swapped capitalisation in that first sentence is quite funny)

DJMcNab: This doesn't need to be resolved now, and so can land as-is.

PR #744, opened 2024-11-25
Add Linebender lint set v2.

xStrom:
Actually satisfying a good chunk of these lints is deferred and they are in various `allow` blocks.

I deleted the `clone_style_ref` function because `StyleRef` implements `Copy`.

The `rand` method `gen` has to be a raw `r#gen` until they publish a new version. They have that method renamed in [`master`](https://github.com/rust-random/rand).

Other lint based changes should be straightforward to understand.

### DIFF
--- a/vello/src/debug/renderer.rs
+++ b/vello/src/debug/renderer.rs
@@ -1,6 +1,10 @@
 // Copyright 2023 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// size_of is not part of the prelude until Rust 1.80 and our MSRV is below that
+#[allow(unused_imports)]
+use core::mem::size_of;
+
 use super::DebugLayers;
 use crate::{
     debug::validate::{validate_line_soup, LineEndpoint},
--- a/vello/src/debug/renderer.rs
+++ b/vello/src/debug/renderer.rs
@@ -72,7 +72,7 @@ impl DebugRenderer {
             },
             // This mirrors the layout of the PathBbox structure.
             Some(wgpu::VertexBufferLayout {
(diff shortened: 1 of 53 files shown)


waywardmonkeys (approved):

vello/src/debug/renderer.rs
```diff
@@ -1,6 +1,10 @@
 // Copyright 2023 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// size_of is not part of the prelude until Rust 1.80 and our MSRV is below that
+#[allow(unused_imports)]
+use core::mem::size_of;
```
Philipp-M, line 6: I wonder whether we should just update MSRV to avoid accidentally having these as (unnecessary) artifacts later, as this isn't used here anyways?

waywardmonkeys: The MSRV on `peniko`, `vello`, and everything above it will soon be 1.82. Everything being `masonry` and the Xilem family.

xStrom: Yeah the 1.82 update is coming real soon and I made sure to have all of these as exact copies so it will be easy to remove.

xStrom: > as this isn't used here anyways?

It is used if you check the full PR's diff.

xStrom: The `allow` is necessary because of the `unused_qualifications` Rust compiler lint, which triggers with 1.82 because `size_of` is part of the prelude since 1.80 and thus this explicit import is not needed in 1.82.

The import is still necessary because `size_of` is not yet part of the prelude in 1.76.

PR #745, opened 2024-11-27
Update to MSRV 1.82 in preparation for color

DJMcNab:


vello/src/recording.rs
```diff
@@ -168,7 +168,10 @@ impl Recording {
     /// Dispatch a compute shader where the size is determined dynamically.
     /// The `buf` argument contains the dispatch size, 3 `u32` values beginning
     /// at the given byte `offset`.
-    #[allow(unused)]
+    #[cfg_attr(
+        not(feature = "debug_layers"),
+        expect(unused, reason = "Currently only used by the debug layers")
+    )]
```
waywardmonkeys, line 174: This is `pub`, so why is it needing to be marked as `unused`?

vello/src/debug/renderer.rs
```diff
@@ -211,8 +213,10 @@ impl DebugRenderer {
         }
     }
 
-    #[allow(clippy::too_many_arguments)]
-    // #[expect(clippy::too_many_arguments, reason="This function is internal, so the argument count doesn't cause issues for consumers.")]
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "This function is internal, so the argument count doesn't cause issues for consumers."
+    )]
```
waywardmonkeys, line 219: Not sure what value `clippy::too_many_arguments` brings to the table.

waywardmonkeys: Yeah, but the number of times I've gotten that lint and thought "Oh, I'll restructure the code" vs the number of times that we suppress it ...

xStrom: It's true that an actually unreasonable number of arguments would be caught by human review.

A compromise could be to increase our [too-many-arguments-threshold](https://doc.rust-lang.org/clippy/lint_configuration.html#too-many-arguments-threshold) from 7 to something higher.

DJMcNab: I think it's something good to reason about for externally facing APIs, but for this kind of internal stuff ignoring it is fine. I don't think there's a way to configure the two separately.

What solution do I need to do here to get this PR landable?

xStrom: I think this can remain as-is for this PR, and let's come up with a Linebender-wide solution in [Zulip](https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Standard.20Lint.20set).

vello/src/recording.rs
DJMcNab: Hmm, actually it wasn't unused. Sorry I didn't analyse this fully.

vello_encoding/src/lib.rs
```diff
@@ -32,6 +32,8 @@
 #![allow(
     unnameable_types,
     clippy::shadow_unrelated,
+    // Only applies on WASM?
+    variant_size_differences,
```
waywardmonkeys, line 36: It is more likely that it is 32 bit targets, not just wasm, like I ran into in Peniko. But some 32 bit targets can't do AtomicU64, and so we don't support building on them all.

vello/src/util.rs
```diff
@@ -137,7 +140,13 @@ impl RenderContext {
...
         let features = adapter.features();
         let limits = Limits::default();
-        #[allow(unused_mut)]
+        #[cfg_attr(
+            not(feature = "wgpu-profiler"),
+            expect(
+                unused_mut,
+                reason = "Mutation is only expected with the wgpu-profiler feature"
+            )
+        )]
```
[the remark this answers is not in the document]
DJMcNab: The reason for this pattern is for future compatibility. I've done a shadowing based approach which I think works well.

xStrom (approved):
Thanks, good work! 🚀

PR #747, opened 2024-11-28
Remove skrifa from vello's public API

nicoburns:

- To align vello's representation of "normalized coordinates" with that of swash/parley (avoiding a boilerplatey conversion when integrating parley with vello)
- To make skrifa version bumps in vello non-breaking changes.

### DIFF
--- a/examples/scenes/src/simple_text.rs
+++ b/examples/scenes/src/simple_text.rs
@@ -4,10 +4,9 @@
 use std::sync::Arc;
 
 use skrifa::prelude::NormalizedCoord;
-use skrifa::{raw::FontRef, MetadataProvider};
+use skrifa::{raw::{FontRef, FileRef}, MetadataProvider};
 use vello::kurbo::Affine;
-use vello::peniko::{color::palette, Blob, Brush, BrushRef, Font, StyleRef};
-use vello::skrifa::{raw::FontRef, MetadataProvider};
+use vello::peniko::{color::palette, Blob, Brush, BrushRef, Font, StyleRef, Fill};
 use vello::{Glyph, Scene};
 
 // This is very much a hack to get things working.
@@ -198,7 +197,6 @@ impl SimpleText {
         transform: Affine,
         text: &str,
     ) {
-        use vello::peniko::Fill;
         let brush = brush.unwrap_or(&Brush::Solid(palette::css::WHITE));
         self.add_run(
             scene,
@@ -214,7 +212,6 @@ impl SimpleText {
 }
 
 fn to_font_ref(font: &Font) -> Option<FontRef<'_>> {
-    use skrifa::raw::FileRef;
     let file_ref = FileRef::new(font.data.as_ref()).ok()?;
     match file_ref {
         FileRef::Font(font) => Some(font),
--- a/examples/scenes/src/simple_text.rs
+++ b/examples/scenes/src/simple_text.rs
@@ -4,9 +4,12 @@
 use std::sync::Arc;
 
 use skrifa::prelude::NormalizedCoord;
-use skrifa::{raw::{FontRef, FileRef}, MetadataProvider};
+use skrifa::{
+    raw::{FileRef, FontRef},
+    MetadataProvider,
+};
 use vello::kurbo::Affine;
-use vello::peniko::{color::palette, Blob, Brush, BrushRef, Font, StyleRef, Fill};
+use vello::peniko::{color::palette, Blob, Brush, BrushRef, Fill, Font, StyleRef};
 use vello::{Glyph, Scene};
 
 // This is very much a hack to get things working.
--- a/examples/scenes/src/simple_text.rs
+++ b/examples/scenes/src/simple_text.rs
@@ -3,7 +3,6 @@
 
 use std::sync::Arc;
 
-use skrifa::prelude::NormalizedCoord;
 use skrifa::{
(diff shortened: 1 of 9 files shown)


vello/src/scene.rs
```diff
@@ -400,7 +400,7 @@ impl<'a> DrawGlyphs<'a> {
     }
 
     /// Sets the normalized design space coordinates for a variable font instance.
-    pub fn normalized_coords(mut self, coords: &[NormalizedCoord]) -> Self {
+    pub fn normalized_coords(mut self, coords: impl Iterator<Item = i16>) -> Self {
```
nicoburns, line 403: We could require `ExactSizeIterator` here if we want to.

dfrg:
Note this type is still visible from `vello_encoding`: https://github.com/linebender/vello/blob/e6ead09604d9b64229577f039883dfb4df598997/vello_encoding/src/encoding.rs#L548

This is a bit more tricky because skrifa actually wants a `&[NormalizedCoord]` type. We can either keep an internal buffer in the resolver and do a copy or just use bytemuck to cast the slice (`font_types::F2Dot14` impls the necessary traits)

vello/src/scene.rs
dfrg: I don’t think it would hurt to have a `pub type NormalizedCoord = i16` in vello_encoding

examples/scenes/src/simple_text.rs
```diff
@@ -161,7 +162,13 @@ impl SimpleText {
             .font_size(size)
             .transform(transform)
             .glyph_transform(glyph_transform)
-            .normalized_coords(var_loc.coords())
+            .normalized_coords(
+                var_loc
+                    .coords()
+                    .iter()
+                    .copied()
+                    .map(NormalizedCoord::to_bits),
```
DJMcNab, line 170: I think:
```suggestion
                    .map(|coord| coord.to_bits()),
```
should work and is slightly nicer

examples/scenes/src/simple_text.rs
```diff
@@ -210,7 +217,7 @@ impl SimpleText {
 }
 
 fn to_font_ref(font: &Font) -> Option<FontRef<'_>> {
-    use vello::skrifa::raw::FileRef;
+    use skrifa::raw::FileRef;
```
DJMcNab, line 220: Is there much advantage to this import being here?

vello/src/scene.rs
DJMcNab: What does `ExactSizeIterator` win us?

DJMcNab: I think we definitely should document this slightly better, even if just to say "this will be provided by your text layout implementation". The current format is quite unactionable.

DJMcNab (commented):
Looks good. This also needs a CHANGELOG entry

waywardmonkeys:
@nicoburns Can we wrap this one up and land it? I'd like to make sure this is in our upcoming release.

examples/scenes/src/simple_text.rs
DJMcNab: Looking again, I don't think this is necessary.

examples/scenes/src/simple_text.rs
nicoburns: I don't think so. That was where it already was, but I've moved it to the top.

vello/src/scene.rs
nicoburns: > What does `ExactSizeIterator` win us?

Potentially slightly cheaper allocation. But I suspect `size_hint` will work even without `ExactSizeIterator`, and that even if it doesn't the cost will be small.

nicoburns:
How much do we still care about this being public in `vello_encoding`? Seems to me that it's probably less important to do semver compatible releases there? It's easy enough to change everything to i16, but it involves changing a lot of the internal uses of `NormalizedCoord` into `i16` which seems like somewhat of a code quality regression from a readability / type safety point of view.

examples/scenes/src/simple_text.rs
DJMcNab: Ah, apologies. I probably didn't originally notice that this was only a change and not an addition.

DJMcNab:
I don't hate making `vello_encoding` intentionally not follow semver, but we'd need to do two things:
1) Document that being the case
2) Use an `=` dependency specification on it from Vello.

nicoburns:
> I don't hate making `vello_encoding` intentionally not follow semver, but we'd need to do two things:
> 
> 1. Document that being the case
> 2. Use an `=` dependency specification on it from Vello.

My suggestion would be for `vello_encoding` to continue to follow semver, but to aggressively bumping it's version number beyond that of the main `vello` crate's (as required). I feel like "you're using the low-level API, so you have to deal with breaking changes more often" is kinda reasonable, especially if the "breaking change" is just a dep bump. Also, I believe most users of `vello_encoding` are unlikely to have large crates.io dep trees. So conflicting versions is likely to be less of an issue.

DJMcNab:
Looking again, I'm not sure that would be practical; Vello itself exposes vello_encoding as public API. I don't know the right answer here

DJMcNab:
I'd probably suggest making our own `NormalizedCoord` type, and then as Chad says, using `bytemuck` as/when needed.

waywardmonkeys (approved):

PR #748, opened 2024-11-28
Spell `color` consistently as `color`

waywardmonkeys:


waywardmonkeys:
Noticed this due to working on #742. Separated it out rather than folding it into that.

Considered putting a thing in `.typos.toml` to catch future instances, but opted not to for now.

vello_tests/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub use snapshot::{
 pub struct TestParams {
     pub width: u32,
     pub height: u32,
-    pub base_colour: Option<Color>,
+    pub base_color: Option<Color>,
```
xorgy, line 63: Perhaps we could go the other way and introduce more diverse spellings of the word color, instead of stamping them out.
```suggestion
    /// La couleur d'arrière-plan, le cas échéant.
    pub base_colour: Option<Color>,
```

xorgy (approved):

PR #751, opened 2024-11-28
Update to Rust 1.83

DJMcNab:
Major changes are some new clippy lints and updates to existing lints.

We don't really use const, which are the other main changes from this release.


waywardmonkeys (changes requested):
I'm currently a "no" on being forced to write code like this with `needless_lifetimes`. It brings no value that I see.

waywardmonkeys (approved):
I'm going to give in on these needless `needless_lifetime` changes since they're a default in `clippy` and I can see an argument for them.

But I will be pushing for a reduction in our lint configuration separate from this PR.

PR #752, opened 2024-11-28
Run `cargo upgrade && cargo update`

DJMcNab:


waywardmonkeys (approved):

xorgy:
@DJMcNab 
This has caused a very noticeable regression on my system† (both under X11 and Wayland). It misses groups of between 3 and 5 frames in a row, roughly once every 0.6 seconds. I bisected to this commit.

† RADV Mesa 24.3 on Radeon VII, Linux

PR #753, opened 2024-11-29
pico_svg: Simplify color parsing.

waywardmonkeys:
The color parsing in `color` is more complete than the older code in `peniko`, so this can be simplified now.

Also, switch from `peniko::Color` to `AlphaColor<Srgb>` to continue down the path of using `peniko::Color` less.

Finally, add a couple of tests to show the color parsing working with a couple of the supported syntaxes.

### DIFF
--- a/examples/scenes/src/pico_svg.rs
+++ b/examples/scenes/src/pico_svg.rs
@@ -8,8 +8,7 @@ use std::str::FromStr;
 use roxmltree::{Document, Node};
 use vello::{
     kurbo::{Affine, BezPath, Point, Size, Vec2},
-    peniko::color::{self, palette},
-    peniko::Color,
+    peniko::color::{self, palette, AlphaColor, DynamicColor, Srgb},
 };
 
 pub struct PicoSvg {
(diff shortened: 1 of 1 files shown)


examples/scenes/src/pico_svg.rs
```diff
@@ -288,3 +282,38 @@ fn modify_opacity(color: Color, attr_name: &str, node: Node<'_, '_>) -> Color {
...
+    fn assert_close_color(c1: AlphaColor<Srgb>, c2: AlphaColor<Srgb>) {
+        const EPSILON: f32 = 1e-4;
+        assert_eq!(c1.cs, c2.cs);
+        for i in 0..4 {
+            assert!((c1.components[i] - c2.components[i]).abs() < EPSILON);
+        }
+    }
+
+    #[test]
+    fn parse_colors() {
```
DJMcNab, line 300: Hmm. Is it best practise to have tests which are only testing your dependencies? More tests are always good, I suppose.

The thing which gives me pause is that if this stops working, there isn't any remediation possible in this repository. But OTOH, this test failing would still be a good thing, as it's testing an assumption we're making, and would catch that before "shipping" it.

DJMcNab (approved):

examples/scenes/src/pico_svg.rs
waywardmonkeys: Partly it is a signal that this works since Raph asked and I wasn't fully sure.

PR #754, opened 2024-11-29
Remove `full` feature.

waywardmonkeys:
This was needed for an integration and according to Raph is no longer needed.


waywardmonkeys:
I'm curious about the `r8` define in the shaders as well and if that's still useful.

DJMcNab:
r8 can be useful for Glyph caching, as the glyph outlines can be used as a mask (saying that, subpixel AA is a potential factor here)

waywardmonkeys:
Moved question about `r8` to issue #755.

DJMcNab (approved):

dfrg:
This was meant for an earlier Skia integration which is no longer on the roadmap (replaced by potential future work with sparse strips), so removing this makes sense to me.

PR #756, opened 2024-11-29
Remove `r8` permutations of shaders

### LINKED ISSUE #755: Fate / status of `r8` ifdef in shaders
In `fine.wgsl`:

```
#ifdef r8
// The R8 variant is only available via an internal extension in Dawn native
// (see https://dawn.googlesource.com/dawn/+/refs/heads/main/docs/tint/extensions/chromium_internal_graphite.md).
#enable chromium_internal_graphite;
#endif
```

```
@group(0) @binding(5)
#ifdef r8
var output: texture_storage_2d<r8unorm, write>;
#else
var output: texture_storage_2d<rgba8unorm, write>;
#endif
```

Is this used by anything still?

It seems like `r8unorm` is a valid part of the WebGPU specification now, so maybe this is a hold over from earlier days.

Is this something that can be used from the renderer code or was that part of something else as well?

waywardmonkeys:
These aren't used and it was discussed in #755 to remove them.

Fixes #755.


waywardmonkeys:
By not having the r8 permutations of the shaders, the `shaders.rs` is 160k smaller!

DJMcNab (approved):
Formalising approval from #755

PR #757, opened 2024-12-01
Remove `buffer_labels` feature, no longer optional

waywardmonkeys:
This extra memory usage from this is tiny in comparison to even a simple use of Vello, so simplify things by removing the feature that made it optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This is the first step towards providing richer color functionality, better hand
 - Breaking: Updated to new `peniko` and `color` is now used for all colors ([#742][] by [@waywardmonkeys])
 - Breaking: The `full` feature is no longer present as the full pipeline is now always built ([#754][] by [@waywardmonkeys])
 - The `r8` permutation of the shaders is no longer available ([#756][] by [@waywardmonkeys])
+- Breaking: The `buffer_labels` feature is no longer present as the labels are always configured ([#757][] by [@waywardmonkeys])
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


raphlinus (approved):
I'm ok with this change, as simpler is better, but I should explain why it's conditional. We don't care about the memory for the label, that's trivial. The problem is that we have a buffer pool to encourage reuse, and reuse is conditional on the `BufferProperties` being `==`. If the name is in the buffer properties, then buffers are locked to that name. I'm not sure this is a problem in practice.

DJMcNab (approved):
I concur with the earlier approval.

One option for regaining the same advantage is to be conditional on whether the wgpu `DISCARD_HAL_LABELS` feature is enabled. That's out-of-scope for this PR, and is probably not needed.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This is the first step towards providing richer color functionality, better hand
 - Breaking: Updated to new `peniko` and `color` is now used for all colors ([#742][] by [@waywardmonkeys])
 - Breaking: The `full` feature is no longer present as the full pipeline is now always built ([#754][] by [@waywardmonkeys])
 - The `r8` permutation of the shaders is no longer available ([#756][] by [@waywardmonkeys])
+- Breaking: The `buffer_labels` feature is no longer present as the labels are always configured ([#757][] by [@waywardmonkeys])
```
DJMcNab, line 29: This is missing the link definition. We can either make a PR now, or put it in another one if we remember?

waywardmonkeys: I'll PR it now.

PR #758, opened 2024-12-01
Update color byte packing

waywardmonkeys:
Colors are now packed in little-endian u32s rather than big-endian which lets us get rid of some swizzling within the shaders.

We also get to take advantage of the new `to_u8_array` functions to simplify the image creation for some tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -435,7 +435,7 @@ dependencies = [
 [[package]]
 name = "color"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


raphlinus (approved):
Yes, this is better, thanks. Probably something will break somewhere, but at least we'll be able to see

waywardmonkeys:
If anything breaks, then it'll just be another step towards having consistent color byte encoding!

PR #759, opened 2024-12-01
Remove `ColorF64`

waywardmonkeys:
This is a step towards using `color` for the gradients, first moving to use the `lerp` function in `AlphaColor` and getting rid of `ColorF64` since our colors are now `f32`-based.


raphlinus (approved):
Looks good, thanks.

I'm a bit curious about the snapshot changes. They're 1 lsb changes, due to roundoff? I guess it's unrealistic to expect them to be bit identical.

waywardmonkeys:
It was a very small change, just above the permitted deviation.

PR #760, opened 2024-12-01
Hoist base and blur color unpack4x8unorm from loop

waywardmonkeys:
The calls to unpack the base color for clearing a tile and the blur color for blurred box code don't need to happen within the loop and we can hoist them out to ensure that they're treated more optimally.


raphlinus (approved):
Yes, better not to expect a sufficiently smart compiler, especially for shaders. I am curious which shader compilers do and which do not perform this optimization, if any.

waywardmonkeys:
I don’t even know how to find that out!

waywardmonkeys:
I know now.

waywardmonkeys:
Described in https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/Manually.20hoisting.20an.20unpack.20in.20a.20loop

PR #761, opened 2024-12-02
shaders: Remove workaround for `const` / `let` globals

waywardmonkeys:
In the past, naga had issues with the correct form for this code, so a workaround was put into place to let the shaders-as-written use `let`, but then turn it into `const`.

This hasn't been strictly followed since then with a couple of things using `const` anyway.


waywardmonkeys:
Take note that a few things were commented out and using hand-written constants but now are using the const to evaluate the value.

DJMcNab (approved):

PR #768, opened 2024-12-13
Migrate to notify-debouncer-full

### LINKED ISSUE #767: Regression: Repeating chunks of 3 to 5 missed frames about twice a second in with_winit following #752
Something in #752 (bisected) is causing the with_winit example to miss groups of 3 to 5 frames at a regular interval now.

I am running it on Mesa 24.3 RADV, on a Radeon VII. This seems to be a present issue maybe so I suspect others may also be affected. The issue is present on both X11 and Wayland, though on Wayland it has some additional effect on present.

DJMcNab:
It's unclear why notify-debouncer-mini ever worked for us

Fixes #767

### DIFF
--- a/examples/with_winit/src/hot_reload.rs
+++ b/examples/with_winit/src/hot_reload.rs
@@ -4,19 +4,31 @@
 use std::time::Duration;
 
 use anyhow::Result;
-use notify_debouncer_mini::notify::RecursiveMode;
-use notify_debouncer_mini::{new_debouncer, DebounceEventResult};
+use notify_debouncer_full::notify::RecursiveMode;
+use notify_debouncer_full::{new_debouncer, DebounceEventResult};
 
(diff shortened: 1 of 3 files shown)


xorgy (approved):

xorgy:
Resolves the issue at least. :+1:

examples/with_winit/src/hot_reload.rs
```diff
@@ -4,19 +4,31 @@
...
 
 pub fn hot_reload(mut f: impl FnMut() -> Option<()> + Send + 'static) -> Result<impl Sized> {
     let mut debouncer = new_debouncer(
         Duration::from_millis(500),
+        None,
         move |res: DebounceEventResult| match res {
-            Ok(_) => f().unwrap(),
+            Ok(events) => {
+                for event in events {
+                    // Don't hot reload if the file was only read (i.e. by us...)
```
xorgy, line 17: :joy:

PR #769, opened 2024-12-17
Update Peniko

waywardmonkeys:


waywardmonkeys:
This builds on https://github.com/linebender/peniko/pull/85

vello_encoding/src/ramp_cache.rs
```diff
@@ -35,13 +36,14 @@ impl RampCache {
...
-        if let Some(entry) = self.map.get_mut(stops) {
+        if let Some(entry) = self.map.get_mut(&CacheKey(ColorStops(stops.into()))) {
             entry.1 = self.epoch;
             entry.0
         } else if self.map.len() < RETAINED_COUNT {
             let id = (self.data.len() / N_SAMPLES) as u32;
             self.data.extend(make_ramp(stops));
-            self.map.insert(stops.into(), (id, self.epoch));
+            self.map
+                .insert(CacheKey(ColorStops(stops.into())), (id, self.epoch));
```
waywardmonkeys, line 46: I'm not thrilled about this change to `ColorStops(stops.into())` rather than perhaps going and doing some better impls for `From` or something in Peniko.

waywardmonkeys:
@DJMcNab This is to help make sure that all is well before we go and publish the new version of Peniko so this doesn't have to land by itself yet ...

vello_encoding/src/ramp_cache.rs
waywardmonkeys: https://github.com/linebender/peniko/pull/86 allows for `CacheKey(ColorStops::from(stops))` / `CacheKey(stops.into())`.

waywardmonkeys: But really this copying of the color stops just to serve as key seems suboptimal since what we could probably do this with the `Range<usize>` that we have available in the resolver code. But that's not for this PR.

waywardmonkeys: We can't currently do that ... I just tried, but we can't rely on data in the `Resources::color_stops` being uniqued, so we can't use the range of it as a key.

vello_encoding/src/ramp_cache.rs
```diff
@@ -35,13 +36,14 @@ impl RampCache {
     }
 
     pub(crate) fn add(&mut self, stops: &[ColorStop]) -> u32 {
-        if let Some(entry) = self.map.get_mut(stops) {
+        if let Some(entry) = self.map.get_mut(&CacheKey(ColorStops(stops.into()))) {
```
DJMcNab, line 39: This is not ideal, as it allocates "unnecessarily". It really complicates the traits in the Color crate to make this work nicely, I think.

DJMcNab (approved):
I think this is fine. IWBN to avoid the need to allocate for `get` to allocate. But given that we probably should be using the entry API here anyway (which would allocate), it's probably fine.

PR #774, opened 2024-12-19
wgpu: Initialize InstanceDescriptor from env.

waywardmonkeys:
This makes the instance descriptor fully configurable from the environment using the underlying wgpu mechanisms.


DJMcNab (approved):
This whole util module is known to be a bit of a disaster, but this is an improvement.

PR #783, opened 2025-01-07
Make `ImageFormat::from_wgpu`'s signature non-total

### LINKED ISSUE #781: Make `ImageFormat::from_wgpu` fallible
From the conversation in #780.

`ImageFormat::from_wgpu` is currently infallible, but only supports a subset of `wgpu` formats, and panics with `unimplemented!()` on unsupported formats.

It might be worthwhile to change this to return `Option<Self>`, to make it clear that not all `wgpu` formats are supported.

DJMcNab:
Fixes #781

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -432,7 +432,8 @@ impl Renderer {
         engine.build_shaders_if_needed(device, options.num_init_threads);
         let blit = options
             .surface_format
-            .map(|surface_format| BlitPipeline::new(device, surface_format, &mut engine));
+            .map(|surface_format| BlitPipeline::new(device, surface_format, &mut engine))
+            .transpose()?;
         #[cfg(feature = "debug_layers")]
         let debug = options
             .surface_format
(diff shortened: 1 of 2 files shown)


vello/src/lib.rs
```diff
@@ -782,11 +790,17 @@ impl Renderer {
             .as_ref()
             .expect("renderer should have configured surface_format to use on a surface");
         let mut recording = Recording::default();
-        let target_proxy = ImageProxy::new(width, height, ImageFormat::from_wgpu(target.format));
+        let target_proxy = ImageProxy::new(
+            width,
+            height,
+            ImageFormat::from_wgpu(target.format)
+                .expect("`TargetTexture` always has a supported texture format"),
+        );
```
waywardmonkeys, line 798: Just checking, but why error here and return an Error below? No value in having a Result from here as well?

And same for above in the same code structure ...

waywardmonkeys (approved):

vello/src/lib.rs
DJMcNab: If you follow the data flow, it's impossible for this to fail. Returning an error would suggest it's possible for the user to address a failure here, which it wouldn't be.

The whole design of the types here is quite complicated, so I just did the minimum thing which worked.

waywardmonkeys: SGTM

casey:
Nice!

PR #766, opened 2024-12-12
Image x/y extend, alpha, nearest neighbor sampling

dfrg:
This adds support for image extend modes, alpha and nearest neighbor sampling to fine and plumbs support for all through the pipeline.

Extend modes with bilinear filtering (bottom right is x repeat and y reflect):
image_extend_bilinear

Extend modes with nearest neighbor filtering:
image_extend_nearest

Image alpha:
image_alpha

PR #784, opened 2025-01-08
Render graph for image filter effects

DJMcNab:
image

This is a design for how image filters could be added. This is currently built entirely[^1] on top of the existing API, and so can theoretically land with minimal risk of breaking existing consumers (maybe after 0.4).

The core concept is that a `Gallery`  (~one per thread) stores the operations which will create a `Painting` (semantically similar to a texture). All galleries can then be combined together in a rendering operation, which will return the texture of a specific Painting, which can be used as you like.
This allows scenes to have a dependency on other paintings, which will be rendered before the scene is rendered. The above image was generated with a "main" scene, which is then blurred. The top part of the image is an inclusion of the rendering, and the bottom part is an inclusion of the blurred version.

Things needing to happen in this PR:
- Docs
- Tests
- Design review

Follow-up work needed (after this PR):
- More filters (grayscale, hue shift, etc.)
- Proper blur filters (y-axis-box-blur, gaussian, etc.)
- Custom operations?
- Greater cleverness around texture atlases
- Batching pre-fine rendering stages

[^1]: We peek into the internals of `override_image` slightly.


nicoburns:
One filter which I don't see listed above is opacity. I'm thinking (but I don't know what I'm talking about) that this might make a for a good first "real world" use case given as I'm assuming that correct opacity blending is easier to implement than a high-quality blur filter, and I suspect it's a more commonly used than something like greyscale.

DJMcNab:
Maybe? It seems more likely that you'd want to use the `alpha` property on the resulting `Image` (once #766 lands) in most cases, and so such an opacity filter would be more an attractive nuisance.

Issue #662, opened 2024-08-09
Zero-width stroke acts as fill


dominikh:
This
```
let stroke = Stroke::new(0.0);
let rect = RoundedRect::new(10.0, 10.0, 240.0, 240.0, 20.0);
let rect_stroke_color = Color::rgb(0.9804, 0.702, 0.5294);
self.scene.stroke(&stroke, Affine::IDENTITY, rect_stroke_color, None, &rect);
```
renders as
![image](https://github.com/user-attachments/assets/89d1fb27-a75e-4c5e-8f2c-6a14ec33fbb3)

DJMcNab:
Thanks for the report! I suspect what's happening here is that the expanded lines overlap, and that confuses something.

But I think @raphlinus will be more likely to give an easy answer.

For the moment, I'd recommend capping your strokes to be at least 0.01 units, presuming that the issue occurs only exactly at 0.

dfrg:
I think the easy fix is to just detect this in `Scene::stroke` and encode nothing. We’d need to do the same for glyph runs.

raphlinus:
Yeah, internally we use a special value of 0.0 of the stroke width to indicate fill. I agree with @dfrg that the easiest thing is to encode nothing. Alternatively, we could switch to using a value of -1.0 (which I believe we did in the past), and have the predicate in the shaders that that the value is negative.

TheNachoBIT:
I've made tests and i found out that every `Style` [(source code just in case)](https://github.com/linebender/vello/blob/main/vello_encoding/src/path.rs) that had a `line_width` of 0.0 gets automatically rendered as a fill no matter what flags in `flags_and_miter_limit` has.

Maybe this is a dumb/not a good idea, but perhaps there could've be a flag for `Style` that explicitly sets if its a fill or a stroke?

PR #785, opened 2025-01-08
Don't encode empty strokes

DJMcNab:
Fixes https://github.com/linebender/vello/issues/662

I was unable to isolate why this bug originally occurs (making the stroke expansion conditional on the width being non-zero appeared to work, but artificially increasing the offset didn't change anything), as I failed to find the condition mentioned in https://github.com/linebender/vello/issues/662#issuecomment-2289723432.

However, I was able to make the fix suggested by Chad in https://github.com/linebender/vello/issues/662#issuecomment-2283730287.

This fix wouldn't apply to manually created (or GPU encoded scenes), but that's pretty much fine, as we don't really support those.

### DIFF
--- a/vello_tests/tests/regression.rs
+++ b/vello_tests/tests/regression.rs
@@ -6,7 +6,7 @@
 use scenes::ImageCache;
 use vello::{
     AaConfig, Scene,
-    kurbo::{Affine, RoundedRect, Stroke},
+    kurbo::{Affine, Rect, RoundedRect, Stroke},
     peniko::{Extend, ImageQuality, color::palette},
 };
 use vello_tests::{TestParams, smoke_snapshot_test_sync, snapshot_test_sync};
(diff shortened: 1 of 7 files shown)


vello_tests/tests/regression.rs
```diff
@@ -24,3 +25,46 @@ fn rounded_rectangle_watertight() {
...
+    );
+    let params = TestParams::new(
+        "text_stroke_width_zero",
+        (font_size * 6.) as _,
+        (font_size * 1.25).ceil() as _,
+    );
+    snapshot_test_sync(scene, &params)
+        .unwrap()
+        .assert_mean_less_than(0.001);
+}
```
DJMcNab, line 91: This needs a test for glyphs

raphlinus (approved):
Feel free to add a test for glyph encoding.

vello_tests/tests/regression.rs
DJMcNab: Oops, there is a test for glyphs!

Issue #749, opened 2024-11-28
Post 742 (color!) followups


waywardmonkeys:
Once #742 lands, there some stuff that we should fix up and improve and I don't want to keep expanding the scope of #742:

* [ ] Remove `ColorF64`: A first pass wouldn't use the good gradient interpolation yet, just use `DynamicColor` instead.
  This is complicated by the fact that gradients and non-gradient color data are being written in different byte orders.
* [ ] Actually do gradients.
* [ ] Use a consistent byte order for color data being sent to the GPU?
* [ ] `vello_encoding::RenderConfig::new` doesn't need to take a reference to the base color, nor does it have to use `peniko::Color`.
  Instead, it can perhaps use `impl Into<DrawColor>` and simplify calling code as that allows a variety of color types to be used.
  This would probably not change a lot though since we don't want to use `DrawColor` in `RenderParams` from the main `vello` crate.
* [ ] Switch more example code from using `Color::from_*` or `Color::new` to using colors from `color::palette::css` where the actual color doesn't really matter.
  Notably, some of the examples should be able to entirely stop using `peniko::Color`.
* [ ] pico_svg's `parse_color` has some fallback for `rgb(...)` syntax that is probably not needed anymore.
* [ ] Consider changing some `to_u32` methods in `color` to be `to_be_u32` and `to_le_u32` or some other naming and making it more clear up through the stack.
* [ ] `DrawColor` has an comment about the endianness, so when fixing things above, make sure that this is corrected.
* [ ] Why does the blurred rounded rect code only support a color as a brush and not `impl Into<BrushRef<'b>>`?

PR #787, opened 2025-01-14
Examples: Replace Color::from_* and Color::new with palette::css where possible

BloodStainedCrow:
As mentioned in #749, this replaces all uses of `Color::from_*` and `Color::new` with `palette::css` constants when possible.

This also allowed me to remove the import of `peniko::Color`, in all but one example project.
I do not know how to use the `examples/scenes` lib, so it is *technically* untested, but I tried to match the colors as best as possible, which should be good enough.


BloodStainedCrow:
Clippy failed because the import is unly used behind the `#[cfg(feature = "wgpu-profiler")]` flag, which seems to not be set. Should I put the import behind a #[cfg(feature = "wgpu-profiler")]`?

DJMcNab:
Thanks for this! To use the `scenes` library, you should run one of the `with_winit` or `headless` examples, which both use `scenes` to provide their content.

> I stumbled onto this by trying to copy the example code into my own project, but discovered it does not compile with the vello version on crates.io.

Hmm, I'm unsure how this change would have an impact there - the palettes used would also not be available on the release on crates.io.

`main` is used for development of future versions, and to see examples for a specific release, you should use that release's tag (https://github.com/linebender/vello/releases/tag/v0.3.0 for v0.3.0). Maybe it would help if we had a comment about this in our README?

> Clippy failed because the import is only used behind the #[cfg(feature = "wgpu-profiler")] flag, which seems to not be set. Should I put the import behind a #[cfg(feature = "wgpu-profiler")]?

I think that's right. It's the solution we used to import `Line`.

DJMcNab:
How did you find these matching colours? Were they all identical to the CSS counterpart? (if the colours have changed, the snapshot tests will need to be updated)

BloodStainedCrow:
Oh, I did not realize the exact values of the example colours mattered for tests, so I just chose the ones from the css list, which matches most closely...

BloodStainedCrow:
> > I stumbled onto this by trying to copy the example code into my own project, but discovered it does not compile with the vello version on crates.io.
> 
> Hmm, I'm unsure how this change would have an impact there - the palettes used would also not be available on the release on crates.io.
This change won´t help. I was just making sure, that the issue is known!

BloodStainedCrow:
How are the snapshot tests generated? Since they will need to be updated.

DJMcNab:
You should use `cargo nextest run` to run the tests, and then follow the guidance to set the environment variable for blessing the tests which should be printed for you.

DJMcNab:
You also need to have git lfs enabled.

BloodStainedCrow:
Ah, that is what i feared. Some test pass on my machine with the modified colors (due to the leniency in the tests), but since this introduces baseline error into each test, they now are much more sensitive, so are failing on i.e. macos.

BloodStainedCrow:
So I should probably regenerate all snapshot test, to ensure this is not a problem in the future (nobody likes sporadic failures)

BloodStainedCrow:
Is it okay if I add a flag `VELLO_TEST_FORCE_UPDATE` to update all snapshot tests, even if they pass?

vello_tests/snapshots/stroke_styles.png
DJMcNab, line 1: I don't know if it's just me, but this colour scheme looks worse to me. The previous scheme all had a very similar saturation, whereas now the green looks more saturated than the orange and the blue (plus the new white, which also stands out)

examples/scenes/src/test_scenes.rs
```diff
@@ -169,10 +169,10 @@ mod impls {
...
         move |scene, params| {
             let colors = [
-                Color::from_rgba8(140, 181, 236, 255),
-                Color::from_rgba8(246, 236, 202, 255),
-                Color::from_rgba8(201, 147, 206, 255),
-                Color::from_rgba8(150, 195, 160, 255),
+                palette::css::SKY_BLUE,
+                palette::css::LIGHT_SALMON,
+                palette::css::LAVENDER,
+                palette::css::PALE_GREEN,
```
DJMcNab, line 175: How did you come to these colours? Given that this was originally the same set of colours as at line 349, it's surprising that three of the four colour names used are different.

examples/scenes/src/mmark.rs
```diff
@@ -118,13 +119,13 @@ impl TestScene for MMark {
...
-    Color::from_rgba8(0x80, 0x80, 0x80, 0xff),
-    Color::from_rgba8(0xc0, 0xc0, 0xc0, 0xff),
-    Color::from_rgba8(0xe0, 0x10, 0x40, 0xff),
+    palette::css::BLACK,
+    palette::css::GRAY,
+    palette::css::LIGHT_GRAY,
+    palette::css::BLACK,
+    palette::css::GRAY,
+    palette::css::LIGHT_GRAY,
+    palette::css::DEEP_PINK,
```
DJMcNab, line 128: These need to match the [motionmark benchmark](https://github.com/WebKit/MotionMark/blob/baf37c1fcb690abbe048e249e248e7847bddd34f/MotionMark/tests/core/resources/canvas-tests.js#L75-L76)

DJMcNab (commented):
The problem with updating *all* the tests is that a lot of them won't have actually changed. I'd suggest either lowering the threshold of similarity temporarily, or just deleting and regenerating the "source" snapshots for the tests you want to keep.

Regenerating the full suite is an operation we effectively never want to use.

DJMcNab:
I'm going to close this; it's a useful change, and someone else can pick it up in future. But it isn't a critical fix, and is only for example/internal testing code, and so I'm not going to drive it to completion.

Thanks for starting this, and sorry we weren't able to land it!

PR #792, opened 2025-01-17
Correct vertical offset for Apple Color Emoji

dfrg:
Experimentation indicates that CoreText special cases the Apple Color Emoji font and adds a 100 font unit vertical offset when rendering. This PR adds a hack to mimic that behavior for correct emoji placement on Mac.

Also fixes a tiny general sbix bug where we were adding rather than subtracting the height of the image.

parley's vello_editor with this fix applied:
Screenshot 2025-01-17 at 10 39 45 AM

### DIFF
--- a/vello/src/scene/bitmap.rs
+++ b/vello/src/scene/bitmap.rs
@@ -13,7 +13,8 @@ use skrifa::{
         types::{GlyphId, Tag},
         FontData, TableProvider,
     },
-    MetadataProvider,
+    string::StringId,
+    FontRef, MetadataProvider,
 };
 
 /// Set of strikes, each containing embedded bitmaps of a single size.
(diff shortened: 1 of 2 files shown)


vello/src/scene/bitmap.rs
```diff
@@ -138,11 +140,32 @@ impl<'a> BitmapStrikes<'a> {
 #[derive(Clone)]
 enum StrikesKind<'a> {
     None,
-    Sbix(sbix::Sbix<'a>, GlyphMetrics<'a>),
+    Sbix(sbix::Sbix<'a>, SbixKind, GlyphMetrics<'a>),
     Cbdt(CbdtTables<'a>),
     Ebdt(EbdtTables<'a>),
 }
 
+#[derive(Copy, Clone, PartialEq)]
```
waywardmonkeys, line 148: Could this get a doc comment to indicate that it is for this particular workaround?

dfrg: Sure, added a note. Let me know if that is sufficient.

waywardmonkeys: Yep!

waywardmonkeys (approved):
I'm tired at the end of a very long day, so maybe someone else might take a look, but LGTM.

DJMcNab (approved):
In theory we should test this, but that's not practically possible. Code looks good, and has been confirmed to work. I also agree with the location of the fix

Thanks

PR #796, opened 2025-01-19
Fix `DrawColor` endianness docs

tomcur:
This was missed when Color switched to packing as little-endian.


tomcur:
I've added the change to `vello_encoding::DrawColor` to the docs as breaking, under the assumption that's tracked.

waywardmonkeys (approved):

PR #797, opened 2025-01-20
Use `Color::from_rgb8`, update to Peniko 0.3.1

waywardmonkeys:
This is a short hand that was added in Color 0.2.3. This change only impacts tests, examples, and docs.

The update to Peniko 0.3.1 brings a requirement for color 0.2.3.


Cargo.lock
```diff
@@ -429,9 +429,9 @@ dependencies = [
 
 [[package]]
 name = "color"
-version = "0.2.2"
+version = "0.2.3"
```
DJMcNab, line 432: This is tricky... Our readme docs require you to have 0.2.2, but if Color is already in your Cargo.lock, updating to Vello 0.4 wouldn't automatically bring in Color 0.2.3. I think on balance, just requiring 0.2.3 in Cargo.toml is the best course of action here.

But the reason we might not want this is because the released Vello crate works with at least 0.2.2.

DJMcNab (approved):

Cargo.lock
tomcur: I was doubting the same thing. FWIW, it probably wouldn't break anything, as it's just examples and one block of `ignored` docstring code.

waywardmonkeys: Not sure if we even build with 0.2.2 and a Peniko would only pull something older…

DJMcNab: To be specific, the thing it would break is people who use the example code to their own project. We have had a few cases of people trying that with different versions, so having example code which can actually be broken just using only the Vello crate is not ideal.

waywardmonkeys: Ends up that there's enough improvements since color 0.2.0 that we should just do a peniko 0.3.1 that uses color 0.2.3 and make sure that we're all in sync .. this will make it easier when we do Parley and Masonry releases / updates as well.

See https://github.com/linebender/peniko/pull/95

tomcur (approved):

DJMcNab (approved):

PR #798, opened 2025-01-21
Test `DrawColor` endianness and premultiplication

tomcur:
`DrawColor` docs got out of sync (see https://github.com/linebender/vello/pull/796), and could get out of sync again: the code doing the packing and the code reading it do not live in this file.

Test whether it is premultiplied for good measure.


waywardmonkeys (approved):
Maybe add a note to update the docs if this test changes and is meant to change?

PR #799, opened 2025-01-21
Use an array texture for the image atlas

DJMcNab:
This has a few tradeoffs, and should probably be treated as a proof of concept only:
1) Massively increases the number of images we can support (untested...)
2) Simplifies atlas allocation complexity, as we don't need to restart to increase the allocation
3) Doesn't allow any images larger than 2048x2048 to be included (currently).
    This is not a fundamental limitation
4) Shouldn't impact compatibility, as 256 layers are supported even in `downlevel_webgl2_defaults`
5) Reduces the memory efficiency of our atlases (as we just abandon partially-filled layers)
6) The Recording level code for this is very janky. We've previously discussed wanting to significantly clean up the recording as not-fit-for-purpose.

cc @nicoburns, have you run into issues with there being too many images?

This is a potential precursor to bindless, as:
1) Bindless can potentially contain atlas allocated (although they would then need to be non-array) textures

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -523,12 +523,14 @@ impl Renderer {
         let target_proxy = ImageProxy::new(
             width,
             height,
+            1,
             ImageFormat::from_wgpu(target.format)
                 .expect("`TargetTexture` always has a supported texture format"),
         );
         let surface_proxy = ImageProxy::new(
             width,
             height,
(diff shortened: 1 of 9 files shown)


DJMcNab:
I think that this is pedagogically interesting, but makes too many tradeoffs to be useful at the moment.

Issue #549, opened 2024-04-12
Rendering to a surface and transparency


waywardmonkeys:
It would be nice if the `render_to_surface` function would respect a transparent `base_color` and then show whatever is already rendered to the surface.

The main issue here is that `BlendState::ALPHA_BLENDING` would need to be set in the `ColorTargetState` in the `BlitPipeline`, but that's set for the `Renderer` and the background color is set in the `RenderParams` which is passed at each invocation of `render_to_surface`.

This also demonstrates that `Renderer` should perhaps be two different types, with a `SurfaceRendener` wrapping `Renderer` and storing the `BlitPipeline` and `TargetTexture` members, since they aren't used when rendering to a texture.

There are other issues with the current code, like if you're rendering to multiple surfaces and they're different sizes, then you'd be constantly creating and re-creating the `TargetTexture`.

Alternatively, maybe the `render_to_surface` should just be somewhere else and not part of the core library ...

waywardmonkeys:
This would also impact using xilem in the current form if you have other content behind the UI (like a 3D application or game).

cfagot:
I've been playing with integrating xilem rendering with the rest of my rendering pipeline. What I'm finding is that, as you suggest, render_to_surface isn't that useful outside of vello only rendering.

Instead I'm using render_to_texture and after making a couple things in vello public (BlitPipeline and TextureTarget) and adding a couple tweaks (blend mode added to the blit pipeline being key) it integrates pretty cleanly with a standard wgpu pipeline, either as a new render pass or embedded within another render pass.

If you are willing to skip anti-aliasing, you don't even need the blit pipeline public and can just blend onto the screen as-is instead.

PR #801, opened 2025-01-24
Add transparency support to render_to_surface.

TheNachoBIT:
This tiny PR adds the feature suggested in #549.

It adds transparency support for `render_to_surface`, with the addition of being able to render scenes on top of already rendered stuff with this function by giving you the option to choose if you want to clear the screen before drawing via an extra argument called "clear". If the check is true, it clears the screen before drawing the scene; if it is false, it doesn't.

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -502,6 +502,7 @@ impl Renderer {
         scene: &Scene,
         surface: &SurfaceTexture,
         params: &RenderParams,
+        clear: bool,
     ) -> Result<()> {
         let width = params.width;
         let height = params.height;
@@ -539,7 +540,10 @@ impl Renderer {
             vertex_buffer: None,
             resources: vec![ResourceProxy::Image(target_proxy)],
             target: surface_proxy,
-            clear_color: Some([0., 0., 0., 0.]),
+            clear_color: match clear {
+                true => Some([0., 0., 0., 0.]),
+                false => None,
+            },
         });
 
         let surface_view = surface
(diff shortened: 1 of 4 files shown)


vello/src/lib.rs
```diff
@@ -502,6 +502,7 @@ impl Renderer {
         scene: &Scene,
         surface: &SurfaceTexture,
         params: &RenderParams,
+        clear: bool,
```
DJMcNab, line 505: A bare boolean parameter is very rarely understandable. I think the best solution here would be to move `clear` into being a field of `RendererOptions`. (We'd need to also then make it a field of `Renderer`).

Clearly that's not ideal in all cases. I hope that we can remove `render_to_surface` entirely *soon*. But that's been planned for probably a year, so...

DJMcNab (commented):
I'd like a slight change in the API, but these changes look reasonable

vello/src/lib.rs
TheNachoBIT: I personally think that `render_to_surface` shouldn't be removed, mostly because it is a function that i depend a lot on to display to the screen directly, without having to set it up manually.

And moving it to the `Renderer` kills a bit of flexibility, since by using this bool inside of the function, it allows you to split scenes in separate draw calls if you need to, and still being able to use the same renderer.

TheNachoBIT: Still, if you guys remove it at some point, it's totally fine! But, i think this function works pretty well, and this fix should allow the surface to have transparency for now c:

DJMcNab: `render_to_surface` is a really terrible API, for a number of reasons. It's definitely going.
With the new wgpu release, you will be able to use [`TextureBlitter`](https://docs.rs/wgpu/latest/wgpu/util/struct.TextureBlitter.html), which will make `render_to_surface` almost trivial.

I don't know if we'll end up writing our own utility for it.

PR #802, opened 2025-01-28
Remove `Arc` from `override_image`

DJMcNab:
A gain from wgpu 0.24


tomcur (approved):
Neat!

DJMcNab:
This API is currently untested, but #784 makes use of it, and we don't need it here any more.

PR #803, opened 2025-01-29
Remove `render_to_surface`

### LINKED ISSUE #549: Rendering to a surface and transparency

DJMcNab:
Replace with blitting/texture utilities in the `utils` module.
Fixes #549

Incompatible with #801. I have a follow-up which shows you would do transparency, but my current thinking is that you should drop off the "utils" happy path if you need that. I could be convinced otherwise.

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -441,8 +441,8 @@ impl Renderer {
     ///    that texture to the surface.
     ///    This pattern is supported by the [`util`] module.
     /// 2) Call `render_to_texture` directly on the [`SurfaceTexture`][wgpu::SurfaceTexture]'s texture, if
-    ///    it has the right usages. This should generally be avoided, as it's a potential performance pitfall
-    ///    on GPUs where the render pipeline method of writing to surfaces is assumed and optimised for.
+    ///    it has the right usages. This should generally be avoided, as some GPUs assume that you will not
+    ///    be rendering to the surface using a compute pipeline, and optimise accordingly.
     pub fn render_to_texture(
         &mut self,
         device: &Device,
@@ -515,10 +515,6 @@ impl Renderer {
     /// Renders a scene to the target texture using an async pipeline.
     ///
     /// Almost all consumers should prefer [`Self::render_to_texture`].
-    ///
-    /// The texture is assumed to be of the specified dimensions and have been created with
-    /// the [`wgpu::TextureFormat::Rgba8Unorm`] format and the [`wgpu::TextureUsages::STORAGE_BINDING`]
-    /// flag set.
     ///
     /// The return value is the value of the `BumpAllocators` in this rendering, which is currently used
     /// for debug output.
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -31,12 +31,9 @@
 //! let (width, height) = ...;
 //! let device: wgpu::Device = ...;
 //! let queue: wgpu::Queue = ...;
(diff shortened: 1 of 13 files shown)


jaredoconnell (commented):
I am not familiar enough with how all of this works to give an approval, but I didn't see any red flags, and it looks well documented.

vello/src/lib.rs
```diff
@@ -597,398 +493,64 @@ impl Renderer {
...
+//             return Err(Error::DownloadError("linesoup"));
+//         };
+
+//         let lines = lines_buf.slice(..bump.lines as u64 * size_of::<LineSoup>() as u64);
+//         let (sender, receiver) = futures_intrusive::channel::shared::oneshot_channel();
+//         lines.map_async(wgpu::MapMode::Read, move |v| sender.send(v).unwrap());
+//         receiver.receive().await.expect("channel was closed")?;
+//         Ok(Self { lines })
+//     }
+// }
```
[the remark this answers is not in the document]
DJMcNab: Essentially, the hope is to restore the debug pipeline, and this is a part of it.

This section probably is peripheral enough not to be worth keeping

vello/src/lib.rs
```diff
@@ -452,6 +434,15 @@ impl Renderer {
...
     /// flag set.
+    ///
+    /// If you want to render Vello content to a surface (such as in a UI toolkit), you have two options:
+    /// 1) Render to an intermediate texture, which is the same size as the surface.
+    ///    You would then use [`TextureBlitter`][wgpu::util::TextureBlitter] to blit the rendered result from
+    ///    that texture to the surface.
+    ///    This pattern is supported by the [`util`] module.
+    /// 2) Call `render_to_texture` directly on the [`SurfaceTexture`][wgpu::SurfaceTexture]'s texture, if
+    ///    it has the right usages. This should generally be avoided, as it's a potential performance pitfall
+    ///    on GPUs where the render pipeline method of writing to surfaces is assumed and optimised for.
```
waywardmonkeys, line 445: Something about this phrasing is confusing to me.

waywardmonkeys (approved):
I think I'm good with this once the clippy and doc stuff is fixed.

This is a step forward and we'll want to iterate more before the next release!

Thanks for pulling out the async changes as that made this easier to understand.

vello/src/lib.rs
DJMcNab: I've tried to rephrase this.

DJMcNab:
The async pipeline with `debug_layers` is not working in this PR. However, this is not really a regression from main, as it also doesn't work there.

waywardmonkeys (approved):

PR #804, opened 2025-01-29
Transparency example (PR only)

DJMcNab:
Demonstration of how you'd get transparency with #803.

I'm not planning to merge this, because I don't think that this should be part of `utils`. However, I'm happy for someone to want to make the follow-up of making this configurable (cc @TheNachoBIT), if they still want to use `utils` and have a transparent surface render. OTOH, #801 is for a different use case, which I would argue should just use a custom `TextureBlitter`.

### DIFF
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -76,6 +76,7 @@ export_scenes!(
     labyrinth(labyrinth),
     robust_paths(robust_paths),
     base_color_test(base_color_test: animated),
+    translucent_base(translucent_base),
     clip_test(clip_test: animated),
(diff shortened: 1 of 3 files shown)


nicoburns:
My use case for this would be rendering a Vello scene such that it could then be composited on top of some other wgpu generated texture such that the other content "shows through" where the Vello scene is entirely transparent.

More concretely the use case I am imagining is something like rendering UI in Blitz on top of gpu-rendered content generated by something like bevy or custom wgpu rendering code. This could be useful for game UI, and it could also be useful for applications like Figma / Graphite / CAD that want standard widgets for their control UI but custom drawing for the "content".

Would the approach in this PR be suitable for that?

DJMcNab:
Sure, I believe so. You'd want the texture you're using for compositing to have the `RENDER_ATTACHMENT` usage.

PR #820, opened 2025-02-20
Endian

tomcur:
(On top of https://github.com/linebender/vello/pull/817 to prevent conflicts. See https://github.com/linebender/vello/commit/6c4d4cc55f173b344f12306f4ef6a4d5d9ecac0a for the changes here.)

WGSL expects little-endian data, but Vello currently doesn't convert to little-endian, except for colors: `DrawColor` specifies its endianness explicitly, with conversion handled by the Color crate. On big-endian hosts, the other encodings would break.

Whether big-endian hosts are a valid target for Vello, I do not know. This is opened in part as a place to have that discussion. If this is desirable, other places will need attention. I have not taken a look yet.

### DIFF
--- a/vello_encoding/src/encoding.rs
+++ b/vello_encoding/src/encoding.rs
@@ -22,9 +22,8 @@ pub struct Encoding {
     /// The path tag stream.
     pub path_tags: Vec<PathTag>,
     /// The path data stream.
-    ///
     /// Stores all coordinates on paths.
-    /// Stored as u32 as all comparisons are performed bitwise.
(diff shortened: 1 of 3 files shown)


tomcur:
Consider all the structs with `Bytemuck::Pod` that would also need treatment. Doing this right would be somewhat involved.

DJMcNab:
I think that it would be good to get this right, but it's hard because of the CPU shaders; the CPU shaders need to have the opposite transformation.

My preference (as we discussed yesterday) would be to just fail to compile for a big endian system (and adding a non-feature `cfg` to disable that check).

PR #823, opened 2025-02-25
Add font embolden

jrmoulton:
This is basically just a direct port of the swash implementation of font embolden. 

It does seem to have about twice the "strength" of the swash implementation though and I haven't been able to figure out why.


jrmoulton:
Need to investigate:

- [ ] Strength double
- [ ] Interaction with fonts that support bold
- [ ] Doing embolden in stroke expansion on the GPU

jrmoulton:
This is now working. 

It has path expansion/stem darkening working on the GPU and properly works with Bever, Miter, and Round joins. These are all configurable. 

The changes are implemented in both the cpu and gpu flatten shader.

raphlinus:
A quick update, this was discussed in [Renderer office hours](https://docs.google.com/document/d/1r28-uhSl3w9-MhPIfgt7uQXDxPBBVzkw25tlS3b6VMM/edit?usp=sharing), also see [transcript](https://docs.google.com/document/d/1srVD8Y-0ifdjpmDDN5o4ElURee85UIFx76wwdPhkCGQ/edit?usp=sharing) of the discussion.

This would be great to have in some form, but the big question is review bandwidth as it's a fairly major change and has a risk of breaking things. We haven't forgotten about it, but need to decide what to do.

jrmoulton:
I have updated this PR to allow the embolden to take a vec2 for separate emboldening of x and y. 

I will add some comments to the PR to highlight parts that need particular review and open questions.

vello_encoding/src/path.rs
```diff
@@ -833,6 +902,174 @@ impl skrifa::outline::OutlinePen for PathEncoder<'_> {
...
+    }
+}
+
+/// Encoder for path segments with winding information.
+pub struct WindingPathEncoder<'a> {
+    /// The underlying path encoder
+    encoder: PathEncoder<'a>,
+    path_windings: &'a mut Vec<f32>,
+    /// Collects points from all subpaths for a single winding calculation
+    all_glyph_points: Vec<Point>,
```
raphlinus, line 916: It's not obvious to me why we're storing the points rather than accumulating the area on the fly.

vello_encoding/src/path.rs
```diff
@@ -833,6 +902,174 @@ impl skrifa::outline::OutlinePen for PathEncoder<'_> {
...
+            }
+
+            self.winding_stored = true;
+        }
+
+        self.encoder.finish(insert_path_marker)
+    }
+}
+
+fn compute_winding(points: &[Point]) -> u8 {
```
raphlinus, line 1042: Unless I'm missing something, this has the possibility of mis-computing area, as I believe it includes a component from the last point of a subpath to the `moveto` opening the next subpath, and this should not be included. I think this logic would be fairly straightforward to accumulate on the fly.

vello_shaders/shader/flatten.wgsl
```diff
@@ -417,7 +424,7 @@ fn flatten_euler(
             var n_frac: f32;
             var robust = ESPC_ROBUST_NORMAL;
             if abs(k1) < K1_THRESH {
-                let k = es.params.k0;
+                let k = k0 + 0.5 * k1;
```
raphlinus, line 427: Reason for this change? They should be mathematically identical, and the old version is a touch less work.

jrmoulton:
> I'm not yet convinced the logic for separate x and y emboldening is correct; if I'm reading the code correctly, what you have is basically a variable width that's linearly interpolated from the "projected" boldening at the edges. This will incorrectly compute the cusp in extreme cases, which of course we don't expect for font rendering. The more principled way to do it would be to nonuniform scale the path, apply the stroke offset, then inverse nonuniform scale. That's very similar to the way stroke expansion works in the presence of an eccentric affine transformation (in this case, there's no pre-transform, just the application of a post-transform). However, I suspect that the difference between what you have and the "fully correct" version is subtle and is probably not worth additional complexity/rework, so I'm just mentioning it.

Ok I'll fix the items from the feedback and take a look at this and report back. 

> Overall, very cool

❤️

PR #826, opened 2025-02-27
Introduce Vello Hybrid crates structure for CPU/GPU Sparse Rendering

grebmeg:
This PR introduces a new modular crate structure for Vello Sparse Strips implementation, based on the sparse rendering approach outlined in [Potato: a hybrid CPU/GPU 2D renderer design](https://docs.google.com/document/d/1gEqf7ehTzd89Djf_VpkL0B_Fb15e0w5fuv_UzyacAPU/edit?tab=t.0#heading=h.27frrspcevv4) by @raphlinus.

Changes
- Added `vello_api` – Defines public API types shared across implementations.
- Added `vello_common` – Provides shared data structures and utilities for rendering.
- Added `vello_cpu` – Implements a CPU-based renderer optimized for multithreading and SIMD.
- Added `vello_hybrid` – A hybrid CPU/GPU renderer, balancing workload between CPU and GPU.


vello_hybrid/vello_api/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+This crate defines the public API types for the Vello rendering. It provides common interfaces and data structures used across different implementations, including CPU, GPU, and hybrid rendering backends.
+
+## Features
+- Shared API types for Vello's rendering pipeline.
+- Interfaces for render contexts and rendering options.
+- Designed for compatibility across CPU and GPU implementations.
+
+## Usage
+This crate is intended to be used by other Vello components and external consumers needing a stable API.
```
LaurenzV, line 22: Is it intended to be used by external consumers? I would say apart from `vello/hybrid` and `vello-cpu` it doesn't really make sense to be used by others, no?

vello_hybrid/README.md
```diff
@@ -0,0 +1,40 @@
+# Vello Hybrid Crates
+
```
DJMcNab, line 2: It would be nice to have an introduction sentence here, explaining what this folder actually is. We currently explain reasonably well what it contains, but not e.g. why it exists

Something like:

```suggestion

We're developing a new implementation of Vello, which is:
- Compatible with a wider set of devices
- Removes some performance cliffs
- Works in more memory conditions (i.e. when there is less memory available)

We are using this folder to develop this implementation, which is currently not suitable for production use.
Our plan is to move the packages in this folder to the top level of the repository once they are ready to be used.

```

vello_hybrid/README.md
```diff
@@ -0,0 +1,40 @@
...
+Licensed under either of
+
+- Apache License, Version 2.0 ([LICENSE-APACHE](LICENSE-APACHE) or <http://www.apache.org/licenses/LICENSE-2.0>)
+- MIT license ([LICENSE-MIT](LICENSE-MIT) or <http://opensource.org/licenses/MIT>)
+
+at your option.
+
+[Rust code of conduct]: https://www.rust-lang.org/policies/code-of-conduct
+[Vello]: https://github.com/linebender/vello
+[the changelog]: https://github.com/linebender/vello/tree/main/CHANGELOG.md
```
DJMcNab, line 40: This file needs a newline at the end.

vello_hybrid/vello_api/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "vello_api"
+version.workspace = true
+description = "Defines the public API types for Vello, providing a stable interface for CPU and Hybrid rendering implementations."
+categories = ["rendering", "graphics"]
+keywords = ["2d", "vector-graphics"]
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+
```
DJMcNab, line 11:
```suggestion

# Until we make our initial release, avoid accidental publishing
publish = false
```

vello_hybrid/vello_api/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Minimum supported Rust Version (MSRV)
+
+TODO: Fill in the MSRV when known.
```
DJMcNab, line 26: Is this because of the Edition 2024 question? Otherwise, I'd expect to set this to the repository MSRV by default.

DJMcNab (commented):
Thanks! It'll be good to get the ball rolling on this.

Just a few details I'd like to see addressed (plus CI)

vello_hybrid/vello_api/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright 2020 the Vello Authors
```
tomcur, line 1: If we expect not to copy too much non-trivial code from current-vello, it _might_ be possible to go with the "no attribution" version. E.g., `color` has: https://github.com/linebender/color/blob/ba0bbaea84baf9e578429aacd3c39aae6ca7e92a/LICENSE-MIT (@DJMcNab).

vello_hybrid/vello_cpu/README.md
```diff
@@ -0,0 +1,47 @@
...
+[![Linebender Zulip chat.](https://img.shields.io/badge/Linebender-%23gpu-blue?logo=Zulip)](https://xi.zulipchat.com/#narrow/stream/197075-gpu)
+[![GitHub Actions CI status.](https://img.shields.io/github/actions/workflow/status/linebender/vello/ci.yml?logo=github&label=CI)](https://github.com/linebender/vello/actions)
+
+</div>
+
+This crate implements the CPU-based renderer for Vello. It is optimized for running on lower-end GPUs or CPU-only environments, leveraging parallelism and SIMD acceleration where possible.
+
+## Features
+- Fully CPU-based path rendering pipeline
+- SIMD and multithreading optimizations
```
tomcur, line 18: Is this crate also where e.g. SIMD strip generation would live, which a hybrid renderer might want to use?

LaurenzV: I think for the common parts this should probably also live in `vello_common`.

vello_hybrid/vello_api/LICENSE-MIT
DJMcNab: I think the attribution is fine to have. Something which my previous claims had missed, was that to merge multiple MIT licenses, it's *probably* valid to just have one block with multiple attributions, and the single block of license text.
That is, if we're taking code from a project with a different attribution here (e.g. from swash for #823), we can add their attributions to the license.

For reference, Rust has moved back to the form with attribution.

tomcur: > to merge multiple MIT licenses, it's probably valid to just have one block with multiple attributions, and the single block of license text

That's certainly how I'd read "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software." If that's indeed the case, it's much less painful to merge.

tomcur: Yes, I think we're all agreed that would be OK. There has been some confusion and churn around this, see e.g. the Rust project moving to the "no-attribution" MIT license: https://github.com/rust-lang/rust/commit/2a8807e889a43c6b89eb6f2736907afa87ae592f.

(And last December reintroducing the attribution string, but without a year: https://github.com/rust-lang/rust/commit/f9c16997dc016a3ef1456f56df2ab564a1c48cb2.)

vello_hybrid/vello_cpu/README.md
tomcur: > the common parts will live in vello_common, and vello_cpu and vello_hybrid should remain independent of each other

That would also be my preference.

grebmeg:
> Thanks! It'll be good to get the ball rolling on this.
> 
> Just a few details I'd like to see addressed (plus CI)

Hey @DJMcNab, thanks for reviewing! I’ve addressed your comments and fixed the CI errors. Let me know if there’s anything else that needs tweaking!

DJMcNab:
I don't want to waste too much time bike shedding this. I'd probably lean towards renaming the directory to something like `sparse_strips`, to avoid "Vello overload" in the filesystem paths.

But ultimately I also think not doing that would be fine. If merging this now unblocks more work today, we (@tomcur, myself or anyone else) can merge it as-is.

tomcur (approved):

PR #827, opened 2025-03-03
Add the code for `vello_api`

LaurenzV:
It`s possible I later on realize there is something else missing, but I think for now this should cover most of what is needed for that crate, based on my current version. One thing that will need to be added later on is the render context trait.


sparse_strips/vello_api/src/flatten.rs
```diff
@@ -0,0 +1,60 @@
...
+//! Types for flattening curves.
+
+/// A point.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Point {
+    /// The x coordinate of the point.
+    pub x: f32,
+    /// The y coordinate of the point.
+    pub y: f32,
+}
```
tomcur, line 13: If my reasoning about `Tile` and `Strip` living in `sparse_strips/vello_common` is sound, `Point` and `FlatLine` should probably also live there.

tomcur (commented):
These types make sense as a first merge to get some code in, but I'd expect most of these types to go into `sparse_strips/vello_common`. I've added two notes inline, based on the assumption that `sparse_strips/vello_api` should only contain types intended to be used by consumers of the renderer.

sparse_strips/vello_api/src/tile.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct Tile {
+    /// The index of the tile in the x direction.
+    pub x: i32,
+    /// The index of the tile in the y direction.
+    pub y: u16,
+    /// The start point of the line in that tile.
+    pub p0: Point,
+    /// The end point of the line in that tile.
+    pub p1: Point,
+}
```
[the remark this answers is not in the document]
LaurenzV: Yeah I was kind of confused about what should go into `vello_api` and what shouldn't... Should only non-internal things go there? Which basically would only include a re-export of `peniko` and maybe the renderer trait, at which point I'm wondering what the point of `vello_api` is in the first place, since that's not a lot...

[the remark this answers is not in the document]
tomcur: Yes, I would argue that's the right direction for now. If it makes sense, we can always delete `vello_api` later.

sparse_strips/vello_api/src/execute.rs
```diff
@@ -0,0 +1,54 @@
...
+    Auto,
+    /// Force the usage of neon SIMD instructions. This will lead to panics in case
+    /// the CPU doesn't support the target feature `neon`.
+    #[cfg(all(target_arch = "aarch64", feature = "simd"))]
+    Neon,
+    /// Force the usage of AVX2 SIMD instructions. This will lead to panics in case
+    /// the CPU doesn't support the target features `avx2` and `fma`.
+    #[cfg(all(target_arch = "x86_64", feature = "simd"))]
+    Avx2,
+}
```
tomcur, line 26: One alternative would be to lift this to the type-level to allow for free branching due to monomorphization, but perhaps that should be an implementation detail and not be exposed on the API. On the other hand, if we ever get something like [struct target features](https://github.com/rust-lang/rfcs/pull/3525), the consumer could guide autovectorization through monomorphization.

That said, I'd be happy to land this as-is and revisit when we have a clearer picture.

sparse_strips/vello_api/src/lib.rs
```diff
@@ -4,3 +4,9 @@
 //! This crate defines the public API types, providing a stable interface for CPU and hybrid
 //! CPU/GPU rendering implementations. It provides common interfaces and data structures used
 //! across different implementations
+
+#![forbid(unsafe_code)]
+
+pub use peniko::*;
```
tomcur, line 10:
```suggestion
pub use peniko;
pub use peniko::color;
pub use peniko::kurbo;
```

tomcur (changes requested):
One bikeshed regarding the peniko import.

tomcur (approved):
LGTM, the exact module structure may be something to be revisited when we have a clearer picture, but perhaps concrete renderer implementations just re-export the types they need.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -0,0 +1,26 @@
...
+//! Types for paints.
+
+use peniko::color::{AlphaColor, Srgb};
+
+// TODO: This will probably turn into a generic type where
+// vello-hybrid and vello-cpu provide their own instantiations for
+// a `Pattern` type.
+/// A paint used for filling or stroking paths.
+#[derive(Debug, Clone)]
+pub enum Paint {
```
grebmeg, line 13: Can this be used only internally, or is it also available for external consumers?

My initial thought about `vello_api` was that it would act as an entry point for external usage — kind of like how wgpu is structured. As I understand it, wgpu-api itself is implemented under the wgpu crate, but the real core is in wgpu-core, which then uses wgpu-hal. wgpu-hal abstracts over different graphics APIs with backends for Vulkan, Metal, Direct3D, and OpenGL.

I was imagining something similar here — where consumers of `vello_api` could use a single API and choose the rendering path dynamically, depending on the device. For example, they could switch between a CPU renderer, a hybrid CPU/GPU mode, or maybe even enable compute if the device supports it.

With that in mind, I think it’d be great to add an example showcasing how to use it later. Also, adding some common types and structures to `vello_common` for now seems like a safe move — we can always upstream them into `vello_api` if they prove useful for consumers.

That said, at this stage, `vello_api` doesn’t seem critical yet. We can probably build and refine it after merging the two implementations we have in progress.

Curious to hear your thoughts — maybe we can chat about this during Office Hours?

sparse_strips/vello_api/src/execute.rs
```diff
@@ -0,0 +1,54 @@
...
+pub enum ExecutionMode {
+    /// Only use scalar execution. This is recommended if you want to have
+    /// consistent results across different platforms and want to avoid unsafe code,
+    /// and is the only option if you disabled the `simd` feature. Performance will be
+    /// worse, though.
+    Scalar,
+    /// Select the best execution mode according to what is available on the host system.
+    /// This is the recommended option for highest performance.
+    #[cfg(feature = "simd")]
+    Auto,
```
grebmeg, line 17: Do I understand correctly that WebAssembly has its own [wasm32-simd128](https://doc.rust-lang.org/beta/core/arch/wasm32/index.html) instruction set (currently in beta), and that this instruction set is not currently supported in the current implementation?

sparse_strips/vello_api/src/paint.rs
taj-p: > For example, they could switch between a CPU renderer, a hybrid CPU/GPU mode, or maybe even enable compute if the device supports it.

Would this be a runtime switch or build time feature? Runtime determination could see increased bundle sizes

grebmeg: For the web target, I believe it would be better to introduce a build-time feature, such as cpu, hybrid, etc., to prevent unnecessary increases in bundle size. This is something we’ll need to work out.

sparse_strips/vello_api/src/execute.rs
LaurenzV: Yes I currently did not write any SIMD code for that, but it's definitely planned. But it probably won't be a focus of my thesis (unless I implement everything else and still have time).

Cargo.toml
```diff
@@ -86,6 +86,7 @@ clippy.wildcard_dependencies = "warn"
 
 [workspace.dependencies]
 vello = { version = "0.4.0", path = "vello" }
+vello_api = { path = "sparse_strips/vello_api" }
```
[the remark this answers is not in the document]
LaurenzV: Hmm, I wouldn't see it this way, the only reason I added it here is so that everywhere else we use it we can just do `vello_api = {workspace = true}` to avoid duplicating the path to the crate. I think everyone who works with this repo is aware that this is in-progress and experimental, so I wouldn't worry about it too much. Especially since we aren't gonna publish it.

sparse_strips/vello_api/src/paint.rs
LaurenzV: > I was imagining something similar here — where consumers of vello_api could use a single API and choose the rendering path dynamically, depending on the device. For example, they could switch between a CPU renderer, a hybrid CPU/GPU mode, or maybe even enable compute if the device supports it.

I think that will already be possible with the `RenderContext` trait, no? Since both renderers will implement that trait you can dynamically choose whether you want to use the CPU-only version or the hybrid version.

> That said, at this stage, vello_api doesn’t seem critical yet. We can probably build and refine it after merging the two implementations we have in progress.

Yeah as I mentioned, to me it's still not 100% clear what the purpose of `vello_api` will be (I've never used wgpu, so unfortunately I can't really comment on that either^^), but I think we can figure it out as we go.

> Would this be a runtime switch or build time feature? Runtime determination could see increased bundle sizes

So, for  `vello_cpu` I would definitely like to have support for runtime detection, so that you can have a portable binary with the best possible performance regardless of which CPU you run on. With that said, to address your concerns:

1) I would be open to adding additional feature flags so that you can disable specific instruction sets you don't want to use at build time, if you think that's better. Then you can have the best of both worlds.
2) My understanding was that for `vello_hybrid` you are mostly targeting WASM? In which case this wouldn't be relevant for you anyway, because the other SIMD kernels will only get compiled for x86/aarch64. Or am I missing something?
3) Since in `vello_common` all of the SIMD stuff is behind feature flags, it is possible to build `vello_cpu` with runtime detection in mind, while vello_hybrid can combine the code in a way so that everything happens at build time, so I don't think there should be any big issues. :)

Cargo.toml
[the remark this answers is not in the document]
tomcur: I don't think adding a crate to the workspace signals readiness, but an argument could be made for ordering the sparse strip crates after the main Vello crates. A note could then be added stating they're the experimental sparse strip crates.

sparse_strips/vello_api/src/paint.rs
tomcur: > So, for vello_cpu I would definitely like to have support for runtime detection, so that you can have a portable binary with the best possible performance regardless of which CPU you run on.

By lifting the CPU feature sets to the type-level the compiler is able to remove unused specializations after monomorphization, also for explicit branches like checking for [`TypeId::of`](https://doc.rust-lang.org/stable/std/any/struct.TypeId.html#method.of). That could help reduce the burden of feature flag sprinkling a bit, by not requiring every individual specialization to be feature flagged.

sparse_strips/vello_api/src/execute.rs
DJMcNab: Can you clarify which part of this is in beta? [`caniuse`](https://caniuse.com/wasm-simd) indicates that it's baseline, and the instructions have a stabilisation Rustc version of 1.54. E.g. [`v128`](https://doc.rust-lang.org/stable/core/arch/wasm32/struct.v128.html).

sparse_strips/vello_api/src/paint.rs
DJMcNab: My baseline assumption is that Vello API would be the equivalent of Piet, and is something we've decided to defer entirely:

[Quoting Raph](https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/.5BProposal.5D.20Vello.20repo.20restructuring.20for.20Potato.20integration/near/502148419):
> The initial API should not be trait based. Like Bintje, I have code in a private branch that stores the transform stack so that is part of the intent of the design. (the code I have is for a tiny-skia backend; I don't think I've published this anywhere. Taking a quick look at the code from Bintje, the logic is similar enough there's not a lot of value in mine). And yes, let's have a separate issue for the medium to longer term API considerations.

(I haven't looked at what's actually been merged here)

sparse_strips/vello_api/src/execute.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Different execution modes for kernels.
+
+#[derive(Copy, Clone, Debug)]
+/// The execution mode used for the rendering process.
+pub enum ExecutionMode {
```
DJMcNab, line 8: This definitely needs to be `non_exhaustive`

sparse_strips/vello_api/src/execute.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// Only use scalar execution. This is recommended if you want to have
+    /// consistent results across different platforms and want to avoid unsafe code,
+    /// and is the only option if you disabled the `simd` feature. Performance will be
+    /// worse, though.
+    Scalar,
+    /// Select the best execution mode according to what is available on the host system.
+    /// This is the recommended option for highest performance.
+    #[cfg(feature = "simd")]
```
DJMcNab, line 16: I don't think this should be feature flagged; we should just always have this be the default

sparse_strips/vello_api/src/execute.rs
LaurenzV: I don’t think it’s beta, at least tiny-skia uses WASM intrinsics and works fine on stable, AFAIK?

sparse_strips/vello_api/src/execute.rs
```diff
@@ -0,0 +1,54 @@
...
+        Self::Auto
+    }
+}
+
+#[cfg(not(feature = "simd"))]
+impl Default for ExecutionMode {
+    fn default() -> Self {
+        Self::Scalar
+    }
+}
```
DJMcNab, line 40: I'm 90% sure that these could use the `#[default]` attribute with `cfg_attr`

sparse_strips/vello_api/src/paint.rs
DJMcNab: I'm not actually sure what this merged content meaningfully represents as a grouping. Almost all of this should probably be in Vello Common?

Certainly the SIMD stuff seems a bit out-of-place in this crate.

DJMcNab: In my mind, the crate which does the dynamic dispatch would be Vello, probably.
Vello would probably only use the CPU mode if:
1) You ask it to directly; OR
2) The adapter available is detected to be a CPU fallback and you don't ask to use it anyway; OR
3) You don't provide a GPU (which is equivalent to 1, really)

LaurenzV: > In my mind, the crate which does the dynamic dispatch would be Vello, probably.

How would that work if you use `vello_cpu` directly, though? I was under the assumption that those crates can be used independently, no? Or do you think in this case it's the responsibility of the user to do the dispatching themselves?

sparse_strips/vello_api/src/execute.rs
grebmeg: Oh, yeah, sorry! I shouldn’t have written “beta”. What I meant is that there’s still a large group of users on Safari < 16.4, which don’t support WASM SIMD.

sparse_strips/vello_api/src/paint.rs
grebmeg: I agree with @DJMcNab on deferring the use of `vello_api`.

> My baseline assumption is that Vello API would be the equivalent of Piet, and is something we've decided to defer entirely:

It looks like there are different understandings of that crate. How about we discuss it later — either in Zulip or during Office Hours? For now, what do you think about using `vello_common` instead of `vello_api`?

PR #828, opened 2025-03-03
Add the code for `vello_common`

LaurenzV:
Note that this is not ready for merging yet as I'm still tracking down a regression I introduced. I didn't notice this because in my branch all affected tests use the rectangle fast path, which I removed here for now. So that might take a while.

But other than that, this part of the code should be pretty much done (and reviewable).


LaurenzV:
I'm not sure what to do about all those truncation clippy lints... Can we ignore them for now?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub fn delta(&self) -> i32 {
+        (self.p1.y == 0.0) as i32 - (self.p0.y == 0.0) as i32
+    }
+}
+
+/// Handles the tiling of paths.
+#[derive(Clone, Debug)]
+pub struct Tiles {
+    tile_buf: Vec<Tile>,
+    tile_index_buf: Vec<TileIndex>,
```
tomcur, line 73: I can imagine this indirection potentially causing a slowdown. That said, I have measured a similar indirection somewhere else and that didn't cause any real slowdown. It most likely does depend on cache availability, and as the number of tiles will generally be greater than the number of lines, it _might_ be an issue for very large paths.

tomcur (commented):
I've not looked in detail yet; I won't have time for that until tomorrow, but if others don't see big blockers I'd be happy landing this and picking up improvements from there.

I have some thoughts about tile generation, and specifically culling geometry to the left of the viewport while accounting for winding calculations. That's probably better left for a follow-up.

sparse_strips/vello_common/src/tile.rs
LaurenzV: I measured it with 2 SVGs and only saw modest but not unsignificant improvements in time (see https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/CPU.20sparse.20strip.20rendering.20to.20pixels/near/499992836). It also would make things easier in case we want to add more attributes to tiles in the future, since we don't need to worry about additional overhead for sorting. But I'm definitely open to revisiting this at a later point.

LaurenzV:
Okay, I fixed the test failures from my branch. So the main thing missing now is figuring out the clippy thing and perhaps additional reviews.

LaurenzV:
I've decided to just disable the clippy lint for now.

LaurenzV:
I'll further try to make the API a bit nicer.

sparse_strips/vello_common/src/lib.rs
```diff
@@ -4,3 +4,18 @@
 //! This crate contains core data structures and utilities shared across crates. It includes
 //! foundational types for path geometry, tiling, and other common operations used in both CPU and
 //! hybrid CPU/GPU rendering.
+
+#![cfg_attr(not(feature = "simd"), forbid(unsafe_code))]
+#![allow(
+    clippy::cast_possible_truncation,
+    reason = "We temporarily ignore those because the casts\
+only break in edge cases, and some of them are also only related to conversions between f32 and f64."
```
tomcur, line 12: Does Clippy warn when going from `f32` to `f64`? Warning when going from `f64` to `f32` makes sense.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -0,0 +1,203 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Rendering strips.
+
+use crate::footprint::Footprint;
+use crate::peniko::Fill;
```
tomcur, line 7: I'd prefer bringing in Peniko as a crate dependency to keep the import paths slightly simpler (and allow grouping external and internal imports). As it's versioned at the workspace-level already, there shouldn't be any issues. The re-exports in `vello_api` are a convenience for consumers and can be useful if they have multiple versions of, say, Kurbo.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    /// Check whether two tiles are at the same location.
+    pub fn same_loc(&self, other: &Self) -> bool {
+        self.x == other.x && self.same_row(other)
+    }
+
+    /// Check whether two tiles are on the same strip.
+    pub fn same_strip(&self, other: &Self) -> bool {
+        self.same_row(other) && (other.x - self.x).abs() <= 1
+    }
```
tomcur, line 59: This name could be confusing, as it's only checking if two tiles are adjacent on the same row. To know whether they're on the same strip requires knowing which tiles are in between these two.

It could be rewritten (and dropping the `abs` for a very slight performance improvement) as

```suggestion
    /// Check whether `self` is the previous tile on the same row as `other`, i.e., it is adjacent to the left of `other`.
    pub fn prev_loc(&self, other: &Self) -> bool {
        self.same_row(other) && self.x + 1 == other.x
    }
```

tomcur (commented):
This looks good in general, I added some thoughts inline.

I expect some simplifications in tile and strip generations to be possible, but some of those simplifications depend on other changes/decisions. This PR definitely should not be blocked on those.

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Flatten a stroked bezier path into line segments.
+pub fn stroke(path: &BezPath, style: &Stroke, affine: Affine, line_buf: &mut Vec<Line>) {
+    line_buf.clear();
+
+    // TODO: Temporary hack to ensure that strokes are scaled properly by the transform.
+    let tolerance = TOL / affine.as_coeffs()[0].abs().max(affine.as_coeffs()[3].abs());
+
+    let lines: LoweredPath<kurbo::Line> =
+        flatten::stroke::stroke_undashed(path.iter(), style, tolerance);
```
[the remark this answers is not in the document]
LaurenzV: Sure, I can change that!

sparse_strips/vello_common/src/tile.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    pub(crate) fn index(&self) -> usize {
+        self.index as usize
+    }
+}
+
+const fn scale_up(z: f32) -> f32 {
+    z * 4.0
+}
```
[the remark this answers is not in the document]
LaurenzV: Oops I already had changed that, looks like it got lost somehow. 😅

sparse_strips/vello_common/src/lib.rs
LaurenzV: I think it was f64 to f32. Yes, it does make sense, but for now I would prefer not having dozens of `try_from` in the code.

sparse_strips/vello_common/src/tile.rs
LaurenzV: I'm not sure I follow here, because if two tiles are adjacent to each other then they automatically will be in the same strip, I think? At least in the current code.

LaurenzV: Ah, you mean two tiles could technically be on the same strip if they are not adjacent, but in this case the method would return false? That makes sense I suppose.

sparse_strips/vello_common/src/lib.rs
tomcur: I agree this is fine, I was just a bit confused by the wording, as it sounds like it warned when going from f64 to f32. I would suggest using `expect` instead of `allow`, so this attribute auto-cleans.

As a separate note, you can also `#[expect(...)]` individual items, so you do not necessarily need `try_from` if you don't allow truncation globally. See [this example](https://github.com/linebender/color/blob/ba0bbaea84baf9e578429aacd3c39aae6ca7e92a/color/src/colorspace.rs#L482-L498) in the Color crate.

sparse_strips/vello_common/src/tile.rs
tomcur: Indeed, I'm bikeshedding the name of the method, as it confused me a little bit.

tomcur: Indeed, I'm bikeshedding the name of the method, as it confused me a little bit. The functionality is fine, but a name indicating it's just checking for tile adjacency would be good.

sparse_strips/vello_common/src/wide.rs
```diff
@@ -0,0 +1,245 @@
...
+            if strip.x >= self.width as i32 {
+                // Don't render strips that are outside the viewport.
+                continue;
+            }
+
+            if strip.y >= self.height as u16 {
+                // Since strips are sorted by location, any subsequent strips will also be
+                // outside the viewport, so we can abort entirely.
+                break;
+            }
```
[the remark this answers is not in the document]
LaurenzV: How so? 🤔  It is much more likely that we hit the first statement (i.e. after the end of each row), while the second statement will only hit once in the very end, and then we abort anyway.

sparse_strips/vello_common/src/lib.rs
LaurenzV: I'll change the crate-level attribute to `expect` now, but let's defer the individual items part to a later point.^^

tomcur (commented):
Thanks for getting this up quickly! I'm leaving some time for other reviews.

The algorithms can probably be polished somewhat (and I've not examined them in complete detail); I'll likely have time for a few follow-ups. As we start iterating, ideally we'd have snapshot tests and some benchmarks to consistently measure the performance impact of changes. Ghostscript Tiger is already in the repo, that could be used to start with. Being able to measure individual stages of the pipeline would be useful.

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -107,20 +107,8 @@ pub fn fill(path: &BezPath, affine: Affine, line_buf: &mut Vec<Line>) {
...
-        flatten::stroke::stroke_undashed(path.iter(), style, tolerance);
-    for line in &lines.path {
-        let scaled_p0 = affine * line.p0;
-        let scaled_p1 = affine * line.p1;
-        let p0 = Point::new(scaled_p0.x as f32, scaled_p0.y as f32);
-        let p1 = Point::new(scaled_p1.x as f32, scaled_p1.y as f32);
-        line_buf.push(Line::new(p0, p1));
-    }
+    let expanded = kurbo::stroke(path.iter(), style, &StrokeOpts::default(), TOL);
+    fill(&expanded, affine, line_buf);
```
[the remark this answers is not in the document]
LaurenzV: Oh, is the kurbo stroke expansion also cubic to line? I thought it’s cubic-to-cubic.

LaurenzV:
> As we start iterating, ideally we'd have snapshot tests and some benchmarks to consistently measure the performance impact of changes.

Yes, I have both, snapshot tests and also benchmarks in my own repo, but I'd like to get this PR and the vello_cpu PR merged before we add them, that's why it would be good to get those merged soon. But I guess let's give it some time in case anyone else wants to take a look. :)

sparse_strips/vello_common/src/strip.rs
```diff
@@ -0,0 +1,203 @@
...
+
+                // Note: We are iterating in column-major order because the inner loop always
+                // has a constant number of iterations, which makes it more SIMD-friendly. Worth
+                // running some tests whether a different order allows for better performance.
+                for x in x0..x1 {
+                    // Relative x offset of the start point from the
+                    // current column.
+                    let rel_x = p0.x - x as f32;
+
+                    for y in 0..4 {
```
raphlinus, line 85: Might not be a bad idea to change some of these hardcoded constants to `STRIP_HEIGHT`, though perhaps that should be a separate PR when we actually test that it works when varied.

raphlinus (approved):
I'm in favor of this going in and iterating. A couple changes suggested in the fill rule logic.

Coarse rasterization will become more sophisticated. Either there's a separate file for the coarse rasterization logic, separate from the wide tile representation (the latter of which is consumed downstream), or I think `wide.rs` should be renamed `coarse.rs`.

In any case, good to see the progress!

tomcur (approved):
`vello_common` needs to be added to the workspace Cargo.toml's `[workspace.dependencies]`. (It probably makes sense to group the sparse strip crates separately from the main Vello crates for now.)

sparse_strips/vello_common/src/wide.rs
```diff
@@ -0,0 +1,246 @@
...
+
+//! Generating and processing wide tiles.
+
+use crate::color::{AlphaColor, Srgb};
+use crate::strip::{Strip, STRIP_HEIGHT};
+use vello_api::paint::Paint;
+use vello_api::peniko::Fill;
+
+/// The width of a wide tile.
+pub const WIDE_TILE_WIDTH: usize = 256;
```
tomcur, line 12: This should probably be `u32`, matching `Tile`. (Practically `u16` is probably big enough to measure the viewport.)

sparse_strips/vello_common/src/strip.rs
LaurenzV: Just an oversight!

sparse_strips/vello_common/src/strip.rs
```diff
@@ -0,0 +1,203 @@
...
+                };
+            }
+
+            match fill_rule {
+                Fill::NonZero => {
+                    fill!(|area: f32| (area.abs().min(1.0) * 255.0 + 0.5) as u32)
+                }
+                Fill::EvenOdd => {
+                    fill!(|area: f32| {
+                        let area_abs = area.abs();
```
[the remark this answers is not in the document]
LaurenzV: The formula looks indeed better! But if I understand correctly removing the macro would imply doing the branching inside of the for loop, which I would like to avoid for performance reasons.

sparse_strips/vello_common/src/wide.rs
LaurenzV: Will defer this for now since it requires quite a few changes in other parts of the code.

PR #830, opened 2025-03-04
Add the code for `vello_cpu`

LaurenzV:
Doesn't include the tests yet which will come after this PR. But I verified that the tests pass using this setup.


DJMcNab:
(the tests are excluded at my request, because we need to make sure that the LFS settings are all done properly)

sparse_strips/vello_cpu/src/fine.rs
```diff
@@ -1,12 +1,12 @@
 // Copyright 2025 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+//! Fine rasterization.
+//!
```
DJMcNab, line 5:
```suggestion
```
or:
```suggestion
//!
//! Fine rasterization converts from tiles containing lines into strips with alpha values for each pixel.
```

(I don't know if that description is true, which is why we should have that description :) but just to give a flavour of what I'd like to see)

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub(crate) paint: Paint,
+    pub(crate) stroke: Stroke,
+    pub(crate) transform: Affine,
+    pub(crate) fill_rule: Fill,
+    pub(crate) blend_mode: BlendMode,
+}
+
+impl RenderContext {
+    /// Create a new render context.
+    pub fn new(width: usize, height: usize) -> Self {
```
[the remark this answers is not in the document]
LaurenzV: Indeed, but it's the same problem as in the last PR that this would require quite a lot of changes now and is probably most suitable as a separate PR, is it okay to add it as a TODO for now?

LaurenzV:
> The RenderContext should probably change to having a transform stack (and in the future a clip stack as well), allowing somewhat more efficient pushing and popping of layers, but that's not a blocker for this PR. 

Yeah, that makes sense!

> As a general note, my preference would be to move away from encoding things in the pipeline as u8 and u32, unless there's a strong reason for doing that.

Yes, we definitely have to address that once we want to support f32-based rendering, though as mentioned before I would leave that for once the basic structure is set up.

sparse_strips/vello_cpu/src/render.rs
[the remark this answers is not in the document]
tomcur: You can also consider taking `u16` as parameters and casting them to `usize` inside the method body (with a TODO note).

sparse_strips/vello_cpu/src/util.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use peniko::color::{PremulColor, Srgb};
+
+pub(crate) trait ColorExt {
+    /// Using the already-existing `to_rgba8` is slow on x86 because it involves rounding, so
```
raphlinus, line 7: I wonder if this should be addressed in the color crate. It's a very complex issue and possibly depends on the `target-cpu` setting or should be implemented using multiversioning. I believe the `round_ties_even()` function has better autovectorization support than the C round semantics. For (non-HDR) colors, just adding 0.5 and doing a truncate (implicit in the `as` here) is the right thing, but for more general rounding cases there are a lot of things that will go wrong: truncation is not floor for negative numbers, and for larger numbers, it's not idempotent for integers. But here it's guaranteed to be non-HDR, enforced by the output type.

I think the right thing to do here is file a bug or PR against color, then plan on removing this code when that's released and we bump the dep.

sparse_strips/vello_cpu/src/fine.rs
```diff
@@ -0,0 +1,176 @@
...
+//! Fine rasterization runs the commands in each wide tile to determine the final RGBA value
+//! of each pixel and pack it into the pixmap.
+
+use crate::util::ColorExt;
+use vello_common::coarse::{Cmd, WIDE_TILE_WIDTH};
+use vello_common::paint::Paint;
+use vello_common::strip::STRIP_HEIGHT;
+
+pub(crate) const COLOR_COMPONENTS: usize = 4;
+pub(crate) const TOTAL_STRIP_HEIGHT: usize = STRIP_HEIGHT * COLOR_COMPONENTS;
```
raphlinus, line 13: I don't love this name, it's not a height. I'm not sure what a less clunky name would be, maybe `STRIP_HEIGHT_COMPONENTS`?

sparse_strips/vello_cpu/src/fine.rs
```diff
@@ -0,0 +1,176 @@
...
+
+pub(crate) mod fill {
+    // See https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators for the
+    // formulas.
+
+    use crate::fine::{COLOR_COMPONENTS, TOTAL_STRIP_HEIGHT};
+    use crate::util::scalar::div_255;
+
+    pub(crate) fn src_over(target: &mut [u8], cs: &[u8; COLOR_COMPONENTS]) {
+        let _as = cs[3] as u16;
```
raphlinus, line 142: Why the underscore in the name?

sparse_strips/vello_cpu/src/lib.rs
```diff
@@ -4,3 +4,15 @@
 //! This crate implements a CPU-based renderer, optimized for SIMD and multithreaded execution.
 //! It is optimized for CPU-bound workloads and serves as a standalone renderer for systems
 //! without GPU acceleration.
+
+#![expect(
+    clippy::cast_possible_truncation,
+    reason = "We cast u16s to u8 in various places where we\
```
raphlinus, line 10: I don't think you need the backslash, the reason should be (and is) short enough to fit on one line.

raphlinus (approved):
Generally looks good, and think in generally we should land things early and then iterate.

I agree with Tom that more precise types for things (colors, chunks of alpha values). Part of the reason why the prototypes use primitive types is bad habits from GPU land. But Tom's suggestion is valid: on the Rust side we should use good types, then depend on bytemuck to bit-reinterpret the types to u32 etc.

sparse_strips/vello_cpu/src/fine.rs
LaurenzV: Because `as` is a reserved keyword, though it probably is not needed for the `ab` below, I will change that.

sparse_strips/vello_cpu/src/util.rs
LaurenzV: Created an issue: https://github.com/linebender/color/issues/142

PR #831, opened 2025-03-05
`vello_hybrid` implementation

grebmeg:
Based on #818: This PR adds a new hybrid sparse strip CPU/GPU renderer (`vello_hybrid`). The CPU handles path processing and initial geometry setup, while the GPU accelerates compositing and blending operations. The architecture consists of `RenderContext` for rendering operations, `Renderer` for GPU resource management, and `RenderData` container for processed geometry. The implementation supports both windowed and headless rendering, includes various examples, and moves some common functionality to shared modules.

```shell
# Render svg to file
cargo run -p vello_hybrid --example render_to_file examples/assets/Ghostscript_Tiger.svg target/Ghostscript_Tiger_VH.png
```

Ghostscript_Tiger_VH

```shell
# Render svg to window
cargo run -p vello_hybrid --example simple
cargo run -p vello_hybrid --example svg examples/assets/Ghostscript_Tiger.svg
```

image


sparse_strips/vello_hybrid/src/simd.rs
DJMcNab, line 1: I'd probably tear out the SIMD code entirely from this version, to make the future refactorings clearer

sparse_strips/vello_hybrid/src/strip.rs
```diff
@@ -0,0 +1,209 @@
...
+                            let xx0 = startx + (y0 - starty) * slope;
+                            let xx1 = startx + (y1 - starty) * slope;
+                            let xmin0 = xx0.min(xx1);
+                            let xmax = xx0.max(xx1);
+                            let xmin = xmin0.min(1.0) - 1e-6;
+                            let b = xmax.min(1.0);
+                            let c = b.max(0.0);
+                            let d = xmin.max(0.0);
+                            let a = (b + 0.5 * (d * d - c * c) - xmin) / (xmax - xmin);
+                            areas[x as usize][y] += a * dy;
```
DJMcNab, line 158: I get a panic on this line:
`index out of bounds: the len is 4 but the index is 4`, just running:
```sh
> cargo run --example gpu
```

grebmeg:
cc @taj-p

sparse_strips/vello_hybrid/src/gpu.rs
```diff
@@ -0,0 +1,648 @@
...
+        let current_height = self.alphas_texture.height();
+
+        if alpha_texture_width > current_width || alpha_texture_height > current_height {
+            // Ensure dimensions don't exceed WebGL2 limits
+            assert!(
+                alpha_texture_height <= max_texture_dimension_2d,
+                "Alpha texture height exceeds WebGL2 limit"
+            );
+
+            self.alphas_texture = self.device.create_texture(&wgpu::TextureDescriptor {
```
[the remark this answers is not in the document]
tomcur: I suspect textures are more performant, but I also haven't measured. I wouldn't say the uniform-buffer approach requires many queue submissions, instead, it does one draw call with a dynamic offset per uniform-sized slice of the big alpha mask buffer: https://github.com/tomcur/bintje/blob/b80f1fe6a53581b988cde1048c63c70b341de0aa/bintje_wgpu/src/lib.rs#L374-L383.

The queue is only submitted when the alpha mask buffer would be overrun (though perhaps double-buffering the alpha mask buffer could reduce the overhead of submission). If I remember correctly, with an alpha mask buffer of 512 KiB a 1600x1600 tiger is two queue submissions.

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -11,7 +11,29 @@ repository.workspace = true
...
+[features]
+# Feature for development-only performance measurement
+perf_measurement = []
 
 [lints]
 workspace = true
+
+[dependencies]
+vello_common = { workspace = true }
+vello_cpu = { workspace = true }
```
LaurenzV, line 23: As mentioned in zulip, this dependency probably shouldn't be there?

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -11,7 +11,29 @@ repository.workspace = true
...
+peniko = { workspace = true }
+kurbo = { workspace = true }
+flatten = { git = "https://github.com/linebender/gpu-stroke-expansion-paper", rev = "827ccf6" }
```
LaurenzV, line 26: Same here, I don't think vello_hybrid needs that dependency?

sparse_strips/vello_cpu/src/lib.rs
```diff
@@ -10,11 +10,9 @@
     reason = "We cast u16s to u8 in various places where we know for sure that it's < 256"
 )]
 
-mod pixmap;
 mod render;
 
 mod fine;
 mod util;
 
-pub use pixmap::Pixmap;
```
LaurenzV, line 19: I don't think this should disappear, `vello_cpu` should at least re-export the type, no?

LaurenzV (commented):
I didn't look very closely at the `vello_hybrid` stuff because that's outside of my area of expertise, I'll leave that to someone else, but the other stuff looks good, just a few comments.

sparse_strips/vello_common/src/pico_svg.rs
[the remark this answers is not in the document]
DJMcNab: Yeah, having svg code in `vello_common` isn't really right, but I agree that this is fine for now.

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -11,7 +11,29 @@ repository.workspace = true
...
+
+# The following dependencies are for GPU. Maybe they should be optional, or the crate
+# structure refactored so there's a CPU-only renderer that doesn't depend on these.
```
DJMcNab, line 29: This is no longer true

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -11,7 +11,29 @@ repository.workspace = true
...
+bytemuck = { workspace = true, features = ["derive"] }
+wgpu = { version = "24.0.1" }
+winit = "0.29"
+pollster = "0.3"
+async-executor = "1.8"
```
DJMcNab, line 34: I think that all three of these should probably be `dev-dependencies`. At the very least, `winit` should not be dependency of this crate.

sparse_strips/vello_hybrid/examples/svg.rs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                match window_event {
+                    WindowEvent::RedrawRequested => {
+                        renderer.render_to_surface(&render_data);
+                        window.request_redraw();
+                    }
+                    WindowEvent::CloseRequested => {
+                        target.exit();
+                    }
+                    _ => (),
```
DJMcNab, line 73: I think it would be worth the complexity to get scroll wheel and/or drag to work here, if at all feasible. With the recgonition that it means that you have to re-render the SVG's items to the render context

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -0,0 +1,81 @@
...
+// 2. A solid color region for fully opaque areas
+//
+// The alpha values are stored in a texture and sampled during fragment shading.
+// This approach optimizes memory usage by only storing alpha data where needed.
+
+struct Config {
+    width: u32,
+    height: u32,
+    strip_height: u32,
+    alpha_texture_width: u32,
```
DJMcNab, line 17: I think you might be able to use [`textureDimensions`](https://www.w3.org/TR/WGSL/#texturedimensions) to get the size of this texture.

sparse_strips/vello_hybrid/src/gpu.rs
```diff
@@ -0,0 +1,649 @@
...
+
+/// GPU renderer for the hybrid rendering system
+#[derive(Debug)]
+pub struct Renderer {
+    /// The GPU device
+    pub device: Device,
+    /// Command queue for the GPU
+    pub queue: Queue,
+    /// Surface for presenting rendered content, if applicable
+    pub surface: Option<Surface<'static>>,
```
DJMcNab, line 62: This feels like the wrong abstraction layer. I strongly suspect that the renderer should just be rendering to a user-provided `TextureView` of dimensions provided by the user

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
DJMcNab, line 1: I've tried to follow this, but couldn't really. I know however that the lack of comments was inherited from #818, so I'm not expecting you to add more explanation

sparse_strips/vello_hybrid/src/gpu.rs
```diff
@@ -0,0 +1,649 @@
...
+impl Renderer {
+    /// Creates a new renderer
+    ///
+    /// The target parameter determines if we render to a window or headless
+    pub async fn new(target: RenderTarget) -> Self {
+        let instance = wgpu::Instance::new(&wgpu::InstanceDescriptor {
+            backends: wgpu::Backends::all(),
+            flags: wgpu::InstanceFlags::empty(),
+            backend_options: wgpu::BackendOptions::default(),
+        });
```
DJMcNab, line 147: We should really be working on a user-provided instance and device

sparse_strips/vello_hybrid/src/gpu.rs
```diff
@@ -0,0 +1,649 @@
...
+        let texture_width = self.alphas_texture.width();
+        let texture_height = self.alphas_texture.height();
+        let mut alpha_data = vec![0_u32; (texture_width * texture_height) as usize];
+
+        // Fill the buffer with alpha data
+        for (idx, alpha) in render_data.alphas.iter().enumerate() {
+            if idx < alpha_data.len() {
+                alpha_data[idx] = *alpha;
+            }
+        }
```
DJMcNab, line 459: Without digging into specifics, this looks an awful lot like [`slice::copy_from_slice`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.copy_from_slice) (indeed, I'm slightly surprised that clippy hasn't said anything)

sparse_strips/vello_hybrid/src/perf_measurement.rs
```diff
@@ -0,0 +1,127 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Performance measurement functionality for GPU operations.
+
+use std::sync::Mutex;
+use wgpu::{Buffer, BufferUsages, CommandEncoder, Device, QuerySet, QueryType, Queue};
+
+/// Handles performance measurement using GPU timestamp queries
```
DJMcNab, line 9: :eyes:

Was wgpu-profiler evaluated for this use?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /// Render the current content to a pixmap.
+    pub async fn render_to_pixmap(&self, pixmap: &mut Pixmap) {
+        let render_data = self.prepare_render_data();
+        let mut renderer = Renderer::new(RenderTarget::Headless {
+            width: pixmap.width as u32,
+            height: pixmap.height as u32,
+        })
+        .await;
```
DJMcNab, line 145: This is also a very strange API; making a new `Renderer` entirely for a single render operation is likely to be extraordinarily expensive

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Fill a path with the current paint and fill rule.
+    pub fn fill_path(&mut self, path: &BezPath) {
+        flatten::fill(path, self.transform, &mut self.line_buf);
+        self.render_path(self.fill_rule, self.paint.clone());
+    }
+
+    /// Stroke a path with the current paint and stroke settings.
+    pub fn stroke_path(&mut self, path: &BezPath) {
+        flatten::stroke(path, &self.stroke, self.transform, &mut self.line_buf);
+        self.render_path(Fill::NonZero, self.paint.clone());
```
DJMcNab, line 90: Hmm, so we finalise the render for each path as we go? Interesting. My gut reaction is that it is simpler, but has costs around how much batching we are doing, as well as blocking the UI thread on things which should be on the render thread (slightly moot at the moment as we do that all on a single thread)

DJMcNab (commented):
Some of the API uses around the GPU/surface lifetimes live in very surprising places here. But it's exciting to get this moving forward

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -0,0 +1,81 @@
...
+
+struct VertexOutput {
+    @location(0) tex_coord: vec2<f32>,
+    @location(1) @interpolate(flat) dense_end: u32,
+    @location(2) @interpolate(flat) color: u32,
+    @builtin(position) position: vec4<f32>,
+};
+
+@group(0) @binding(1)
+var<uniform> config: Config;
```
taj-p, line 36: To be explored separately, I wonder whether it's worth separating `config` out from `alphas_texture` to reduce state change in the pipeline.

sparse_strips/vello_hybrid/src/gpu.rs
grebmeg: Oops, yep! `slice::copy_from_slice` can be used here, but also needed to align the source buffer with the texture size. Thank you!

Fixed in 2357a16

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -0,0 +1,81 @@
...
+    out.dense_end = instance.col + dense_width;
+    let pix_x = f32(x0) + f32(width) * x;
+    let pix_y = f32(y0) + y * f32(config.strip_height);
+    let gl_x = (pix_x + 0.5) * 2.0 / f32(config.width) - 1.0;
+    let gl_y = 1.0 - (pix_y + 0.5) * 2.0 / f32(config.height);
+    out.position = vec4<f32>(gl_x, gl_y, 0.0, 1.0);
+    out.tex_coord = vec2<f32>(f32(instance.col) + x * f32(width), y * f32(config.strip_height));
+    out.color = instance.rgba;
+    return out;
+}
```
taj-p, line 59:
```suggestion
@vertex
fn vs_main(
    @builtin(vertex_index) in_vertex_index: u32,
    instance: StripInstance,
) -> VertexOutput {
    var out: VertexOutput;
    // Map vertex_index (0-3) to quad corners:
    // 0 → (0,0), 1 → (1,0), 2 → (0,1), 3 → (1,1)
    let x = f32(in_vertex_index & 1u);
    let y = f32(in_vertex_index >> 1u);
    // Unpack the x and y coordinates of the strip instance
    let x0 = instance.xy & 0xffffu;
    let y0 = instance.xy >> 16u;
    // Unpack the width and dense width of the strip instance
    let width = instance.widths & 0xffffu;
    let dense_width = instance.widths >> 16u;
    // Calculate the ending x-position of the dense region 
    // Used in the fragment shader to determine whether to sample alpha
    out.dense_end = instance.col + dense_width;
    // Calculate the pixel coordinates of the vertex
    let pix_x = f32(x0) + f32(width) * x;
    let pix_y = f32(y0) + y * f32(config.strip_height);
    // Convert pixel coordinates to normalized device coordinates
    let ndc_x = (pix_x + 0.5) * 2.0 / f32(config.width) - 1.0;
    let ndc_y = 1.0 - (pix_y + 0.5) * 2.0 / f32(config.height);

    out.position = vec4<f32>(ndc_x, ndc_y, 0.0, 1.0);
    out.tex_coord = vec2<f32>(f32(instance.col) + x * f32(width), y * f32(config.strip_height));
    out.color = instance.rgba;
    return out;
}
```

Added minor comments

sparse_strips/vello_hybrid/src/perf_measurement.rs
grebmeg: Nope, but it definitely should be! I’ve removed that custom profiler and will evaluate/add `wgpu-profiler` in a separate PR. Thanks for pointing it out!

Removed in 83148ee

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -0,0 +1,81 @@
...
+    let x = u32(floor(in.tex_coord.x));
+    var alpha = 1.0;
+    // TODO: This is a branch, but we can make it branchless by using a select
+    // would it be faster to do a texture lookup for every pixel?
+    if x < in.dense_end {
+        let y = u32(floor(in.tex_coord.y));
+        // Read alpha value from texture
+        // Calculate texture coordinates based on x for the u32 value
+        let tex_x = x % config.alpha_texture_width;
+        let tex_y = x / config.alpha_texture_width;
```
taj-p: If we made texture_width a power of 2 (which max texture width usually is right?), we could use bit shifting instead here I think:


```wgsl
        let width_bits = firstTrailingBit(config.alpha_texture_width);
        let tex_x = x & (config.alpha_texture_width - 1u);  // x % width
        let tex_y = x >> width_bits;                        // x / width
```

taj-p (commented):
I notice you're pushing now, so submitting review early

sparse_strips/vello_hybrid/src/gpu.rs
```diff
@@ -0,0 +1,600 @@
...
+            label: Some("Strips Buffer"),
+            size: std::mem::size_of::<GpuStrip>() as u64 * initial_strips_capacity,
+            usage: wgpu::BufferUsages::VERTEX | wgpu::BufferUsages::COPY_DST,
+            mapped_at_creation: false,
+        });
+
+        // Create initial texture for alpha values
+        // It will be recreated if needed in prepare
+        let initial_alpha_texture_width = 64;
+        let initial_alpha_texture_height = 64;
```
taj-p, line 309: Should the initial alpha texture dimensions correspond with the number of initial sparse strips? I.e. 64 strips is (64 * 4=) 256 alpha values

sparse_strips/vello_cpu/src/lib.rs
grebmeg: For the same reason, I believe yes.

Fixed in d1776cbdad6ac2efd500926d600024f263981581

sparse_strips/vello_hybrid/examples/simple.rs
DJMcNab, line 1: This example is running into https://github.com/gfx-rs/wgpu/pull/6997. I've put a note on the mentioned backport into the wgpu maintainer meeting notes, and it doesn't actually break anything (it only means we don't get a clean shutdown)

sparse_strips/vello_hybrid/examples/svg.rs
DJMcNab, line 1: This panics on paris-30k:
```
cargo run -p vello_hybrid --example svg -j 6 --release -- ./examples/assets/downloads/paris-30k.svg
   Compiling vello_common v0.4.0 (/home/djmcnab/Documents/repositories/linebender/vello/sparse_strips/vello_common)
   Compiling vello_hybrid v0.4.0 (/home/djmcnab/Documents/repositories/linebender/vello/sparse_strips/vello_hybrid)
    Finished `release` profile [optimized] target(s) in 1.99s
     Running `target/release/examples/svg ./examples/assets/downloads/paris-30k.svg`

thread 'main' panicked at sparse_strips/vello_common/src/coarse.rs:85:24:
index out of bounds: the len is 1128 but the index is 1128
```

sparse_strips/vello_hybrid/src/gpu.rs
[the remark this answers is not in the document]
tomcur: Yes, I believe those numbers are correct, with a 512 KiB alpha mask buffer sliced to 16 KiB uniform buffers, you get 32 draw calls per submission. I don't remember what the size the second submission was, but I believe the big alpha mask buffer could be increased in size by a few orders of magnitude.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,268 @@
...
+                let bg = tile.bg.to_rgba8().to_u32();
+                if bg != 0 {
+                    let strip = GpuStrip {
+                        x: tile_x as u16,
+                        y: tile_y as u16,
+                        width: WIDE_TILE_WIDTH as u16,
+                        dense_width: 0,
+                        col: 0,
+                        rgba: bg,
+                    };
```
tomcur, line 222: It doesn't really matter at this stage, but in the rare case where we're running on a big-endian host, these should be `u16::from_ne_bytes((tile_x as u16).to_le_bytes())`.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,268 @@
...
+                                match cmd_strip.paint {
+                                    Paint::Solid(color) => color,
+                                    _ => peniko::color::AlphaColor::TRANSPARENT,
+                                };
+                            let strip = GpuStrip {
+                                x: (tile_x as u32 + cmd_strip.x) as u16,
+                                y: tile_y as u16,
+                                width: cmd_strip.width as u16,
+                                dense_width: cmd_strip.width as u16,
+                                col: cmd_strip.alpha_ix as u32,
```
tomcur, line 254: With the current commands, `width` and `dense_width` are always either the same, or `dense_width` is 0. If that stays this way, the absence of an alpha mask could be encoded as, say, `col: u32::MAX`. On the other hand, your current encoding would allow merging alpha fills and sparse fills into one command, meanings paths would be drawn with at most two vertex instances per row per wide tile.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,268 @@
...
+                                    Paint::Solid(color) => color,
+                                    _ => peniko::color::AlphaColor::TRANSPARENT,
+                                };
+                            let strip = GpuStrip {
+                                x: (tile_x as u32 + fill.x) as u16,
+                                y: tile_y as u16,
+                                width: fill.width as u16,
+                                dense_width: 0,
+                                col: 0,
+                                rgba: color.to_rgba8().to_u32(),
```
tomcur, line 239: I believe this should be the following.

```suggestion
                                rgba: color.premultiply().to_rgba8().to_u32(),
```

The same for the `AlphaFill` command below and the `bg` above.

tomcur:
> Ghostscript_Tiger_VH

Looking at the pixel coverage of whiskers (and pixel values of those), I believe the image is missing unpremultiplication.

sparse_strips/vello_hybrid/src/render.rs
grebmeg: > On the other hand, your current encoding would allow merging alpha fills and sparse fills into one command, meanings paths would be drawn with at most two vertex instances per row per wide tile

Could you clarify this part? Are you referring to a more advanced batching logic?

grebmeg (commented):
> Looking at the pixel coverage of the whiskers (and pixel RGBA values of those), I believe the image is missing unpremultiplication.

Fixed in cecfb41

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -11,7 +11,29 @@ repository.workspace = true
 # Prevent accidental publishing until the initial release
 publish = false
 
-[dependencies]
+[features]
+# Feature for development-only performance measurement
+perf_measurement = []
```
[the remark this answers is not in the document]
grebmeg: Fixed in 9bb8b97

sparse_strips/vello_hybrid/src/gpu.rs
```diff
@@ -0,0 +1,649 @@
...
+
+        let current_width = self.alphas_texture.width();
+        let current_height = self.alphas_texture.height();
+
+        if alpha_texture_width > current_width || alpha_texture_height > current_height {
+            // Ensure dimensions don't exceed WebGL2 limits
+            assert!(
+                alpha_texture_height <= max_texture_dimension_2d,
+                "Alpha texture height exceeds WebGL2 limit"
+            );
```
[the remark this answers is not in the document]
grebmeg: So, this part has changed a bit — now texture width is accessed in the shader. We dynamically resize the texture if its height is insufficient. The only remaining step is to check `max_texture_dimension_2d` overhead, but from what I see, it’s just a call over internally stored data in `wgpu`. Plus, the limit is only retrieved when resizing is necessary in `prepare` method, so it might not be a major issue. That said, I agree it’s worth verifying — I’ll look into it later.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
grebmeg: Yep, it make sense! Thanks!

Fixed in 3d4ed71

sparse_strips/vello_hybrid/examples/simple.rs
grebmeg: Thank you! 🙏  I’m wondering which platform you encountered the issue on?

sparse_strips/vello_hybrid/examples/svg.rs
grebmeg: Nice example for benchmarking the renderer! I couldn’t reproduce the same issue, but the output looks odd. I think `render_svg` method is quite simplistic, lacking even transform support. So, I believe we should first transition to using `usvg`. Would you mind if I explore this in a separate PR?

image

sparse_strips/vello_hybrid/src/gpu.rs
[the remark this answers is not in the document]
DJMcNab: It's not worth worrying about, given that it's not on the hot path. It does clone a reasonably large structure out, but it's still an extremely easy case for the compiler to optimise.

sparse_strips/vello_hybrid/examples/simple.rs
DJMcNab: This issue occurs on Wayland.

sparse_strips/vello_hybrid/examples/svg.rs
DJMcNab: Yeah, that's reasonable.
The panic seems to now be resolved, luckily. I'm not sure what's changed to make that so.

sparse_strips/vello_hybrid/src/render.rs
DJMcNab: My current preference would be to have a:
```rust
#[cfg(all(not(vello_big_endian_unchecked), target_endian = "big"))]
compile_fail!("Vello currently does not support big endian targets. Enable the `vello_big_endian_unchecked` cfg flag to try and run it anyway.");
```

This is however better as a follow-up.

sparse_strips/vello_hybrid/src/gpu.rs
DJMcNab: I'm not keen to defer this, but for expedience we could do? I'm not entirely sure how this API was reached in the first place.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -0,0 +1,131 @@
...
+fn unpack_alphas_from_channel(rgba: vec4<u32>, channel_index: u32) -> u32 {
+    switch channel_index {
+        case 0u: { return rgba.x; }
+        case 1u: { return rgba.y; }
+        case 2u: { return rgba.z; }
+        case 3u: { return rgba.w; }
+        // Fallback, should never happen
+        default: { return rgba.x; }
+    }
+}
```
DJMcNab, line 131: This should have a final newline. You can normally configure your editor to add these automatically.

Unfortunately, I'm not aware of a good formatter for WGSL files, so this can't really be checked on CI.

sparse_strips/vello_hybrid/src/gpu.rs
grebmeg: I haven’t commented on it because I’ve been refactoring this part according to your comment, which I agree with. I find the current Vello approach in this part suitable for this case, so I’ve implemented a similar solution with some modifications.

Fixed in a2c2ac6

sparse_strips/vello_hybrid/src/gpu.rs
grebmeg: Fixed in a2c2ac6

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
grebmeg: Oh, my bad! I missed that.

> You can normally configure your editor to add these automatically.

Yeah, I recently switched to vscode for some projects and haven’t configured everything yet.

Fixed in 93038ab

DJMcNab (commented):
I've mentioned a blocking concern in [#gpu > Vello Hybrid](https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/Vello.20Hybrid/with/505086068). Once we get that resolved, I'm happy for this to land from my perspective.

The GPU resource management story is something that makes me quite sad; there are quite a few changes I want to make here. However, those are better as follow-ups.

sparse_strips/vello_hybrid/src/render.rs
tomcur: > Could you clarify this part? Are you referring to a more advanced batching logic?

It could be done in batching, but the logic fits better in command generation (after strip rendering). The idea is that if you're encoding `width` and `dense_width` for draw commands, you can fold `CmdFill` and `CmdAlphaFill` into a single command. That would reduce the number of vertex instances required to at most two per path per wide tile, reducing uploads to the GPU. You could go one further and encode a `dense_offset` as well, requiring at most one vertex instance per wide tile for drawing paths.

I've added a note to the renderer office hours agenda.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
taj-p: I think WebGPU can compartmentalise state change if we move less frequently changing resources into lower indexed groups. During a resize operation (i.e. user increases Canvas size each frame), we would need to rebound both the texture and config. If we moved the config to its own group, WebGPU wouldn't need to rebuild its state for the texture too (assuming the config is in group 1 and texture in group 0).

That said, I think it's not worth implementing right now. I think it makes sense as a follow up awhile from now!

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /// Stroke a path with the current paint and stroke settings.
+    pub fn stroke_path(&mut self, path: &BezPath) {
+        flatten::stroke(path, &self.stroke, self.transform, &mut self.line_buf);
+        self.render_path(Fill::NonZero, self.paint.clone());
+    }
+
+    /// Fill a rectangle with the current paint and fill rule.
+    pub fn fill_rect(&mut self, rect: &Rect) {
```
taj-p, line 92: Should we re-expect kurbo so that consumers don't need to pull it in themselves?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        Self {
+            render_bind_group_layout,
+            render_pipeline,
+            resources: None,
+        }
+    }
+
+    /// Prepare the GPU buffers for rendering
+    pub fn prepare(
```
taj-p, line 200:
```suggestion
   fn prepare(
```

Currently this is only called in `render_to_texture`. Maybe we should keep it private until we think it should be exposed?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -0,0 +1,365 @@
...
+    pub width: u32,
+    /// Height of the rendering target
+    pub height: u32,
+    /// Height of a strip in the rendering
+    pub strip_height: u32,
+}
+
+/// Options for the renderer
+#[derive(Debug)]
+pub struct RendererOptions {}
```
taj-p, line 40: WDYT of providing a default implementation for RendererOptions?

sparse_strips/vello_hybrid/src/scene.rs
taj-p: Maybe also `pub use peniko` too?

DJMcNab: The state we've ended up with ([#linebender > Re-exports policy](https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Re-exports.20policy)) is that this should just have a `pub use peniko` for now.

DJMcNab (approved):
There's still a lot of careful thought that needs to go into this, but I think the easiest way for me to make that happen is to just write the code out myself as a follow-up.

The core of this is very solid, and it works well for me.

tomcur (approved):
I'm happy to land this. I haven't looked very closely at the API, but I agree with @DJMcNab that the core looks solid.

Given the time zone differences and the work that's best based on top of this PR, Daniel and I discussed merging this now to unblock that work during our day (instead of Linebender's usual policy of waiting for @grebmeg for the final merge).

PR #832, opened 2025-03-05
Add integration tests for `vello_cpu`

LaurenzV:
Based on https://github.com/linebender/vello/pull/830, which should be merged first.

I tried to follow the steps for Git LFS, but I'm not sure whether it actually worked.


sparse_strips/vello_cpu/.gitattributes
```diff
@@ -0,0 +1 @@
+refs/** filter=lfs diff=lfs merge=lfs -text
```
DJMcNab, line 1: This isn't going to work on CI; you need to also add it to the files setup.

I've opened #833 to hopefully make this clearer

sparse_strips/vello_cpu/tests/util.rs
```diff
@@ -0,0 +1,167 @@
...
+use vello_common::kurbo::{Rect, Shape};
+use vello_cpu::pixmap::Pixmap;
+use vello_cpu::render::RenderContext;
+
+const REPLACE: bool = option_env!("REPLACE").is_some();
+
+static REFS_PATH: LazyLock<PathBuf> =
+    LazyLock::new(|| PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("refs"));
+static DIFFS_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
+    let path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("diffs");
```
DJMcNab, line 28: This... doesn't look right where we're using [cargo-nextest](https://nexte.st/), as each test would overwrite the entire diffs folder.

sparse_strips/vello_cpu/tests/util.rs
```diff
@@ -0,0 +1,167 @@
...
+use std::cmp::max;
+use std::io::Cursor;
+use std::path::PathBuf;
+use std::sync::LazyLock;
+use vello_common::color::palette;
+use vello_common::kurbo::{Rect, Shape};
+use vello_cpu::pixmap::Pixmap;
+use vello_cpu::render::RenderContext;
+
+const REPLACE: bool = option_env!("REPLACE").is_some();
```
DJMcNab, line 23: Why is this a compile time environment variable?

LaurenzV: The main reason is that it allows you to do `REPLACE="" cargo test` in case you want to replace all failing tests which is much more convenient than having a true/false flag that you need to set/unset each time you want to do that. Although it could just be a runtime check to avoid having to recompile, if that's what you mean... I think originally I used the flag in multiple places in the code, that's why I did that.

sparse_strips/vello_cpu/tests/util.rs
LaurenzV: Hmm okay... It's a bit annoying having to manually clear the diff folder, but I guess I can remove it for now.

sparse_strips/vello_cpu/.gitattributes
LaurenzV: Sorry, but this is still confusing me a bit... :( How exactly can I tell whether it works? I mean CI seems to pass, but I still can't tell whether it's actually fetching things from LFS or whether the images are still checked in as normal files?

sparse_strips/vello_cpu/tests/util.rs
DJMcNab: Yeah. I think it should be a runtime check, so you don't need to recompile.

sparse_strips/vello_cpu/.gitattributes
DJMcNab: The changes you've made seem to work. If you commit a change to a test without the corresponding snapshot changing, does that work for you?

DJMcNab:
I think you'll probably need https://github.com/linebender/xilem/pull/881/commits/d4dca3ec9becdb0c225aaf6ba2aeef2c86658a1d

LaurenzV:
I mean, the other temporary option is to just store all screenshots in the same folder for now. Though I'm not sure how great it is to mix sparse_strip stuff with other vello stuff.

xStrom:
Having a single `vello_tests/lfs` directory that contains all the LFS stuff does indeed seem like the simplest solution so we can just move on.

```
# Having all the LFS stuff with a single root path:
vello_tests/lfs
vello_tests/lfs/snapshots
vello_tests/lfs/refs

# Would mean we can just use:
vello_tests/lfs/**/*.png
```

LaurenzV:
Okay, I will try to do that tomorrow!

sparse_strips/vello_cpu/.gitignore
```diff
@@ -0,0 +1 @@
+diffs
```
DJMcNab, line 1: We should have a final newline here

tomcur, line 1: Needs a trailing newline.

sparse_strips/vello_cpu/tests/issues.rs
```diff
@@ -0,0 +1,285 @@
...
+use vello_common::kurbo::{BezPath, Stroke};
+use vello_common::peniko::Fill;
+
+mod util;
+
+// All issue numbers starting from here refer to the issue numbers in the old repo
+// at https://github.com/LaurenzV/cpu-sparse-experiments/issues.
+
+#[test]
+fn issue_2_incorrect_filling_1() {
```
DJMcNab, line 17: I don't think test names referring to issue numbers from a different repository are helpful.
I'd suggest either just `incorrect_filling_1` or maybe `rotated_square`?

It's definitely good to have regression tests, but a lot of these are testing baseline functionality, and it's just a historical coincidence that they had corresponding issues.

LaurenzV:
We figured out the CI thing, so after final approval this can be merged!

sparse_strips/vello_cpu/tests/issues.rs
```diff
@@ -0,0 +1,285 @@
...
+use vello_common::color::palette::css::{DARK_BLUE, LIME};
+use vello_common::kurbo::{BezPath, Stroke};
+use vello_common::peniko::Fill;
+
+mod util;
+
+// All issue numbers starting from here refer to the issue numbers in the old repo
+// at https://github.com/LaurenzV/cpu-sparse-experiments/issues.
+
+#[test]
```
DJMcNab, line 16: We should have a link to the relevant issue, i.e.
```suggestion
#[test]
// https://github.com/LaurenzV/cpu-sparse-experiments/issues/2
```

.gitattributes
```diff
@@ -1,3 +1,4 @@
 # LFS settings
 # If changing, also change in .github/workflows/ci.yml
 vello_tests/snapshots/*.png filter=lfs diff=lfs merge=lfs -text
+sparse_strips/vello_cpu/refs/*.png filter=lfs diff=lfs merge=lfs -text
```
tomcur, line 4: For least astonishment, should we place these in `sparse_strips/vello_cpu/snapshots`, like `vello_tests/snapshosts` and `parley/parley/tests/snapshots`?

sparse_strips/vello_cpu/tests/util.rs
DJMcNab, line 1: This whole module is messy. I'm not claiming it's incorrect to have, as we haven't worked out the right patterns here. A similar pattern is used in Xilem Core.

.gitattributes
LaurenzV: Works for me.

sparse_strips/vello_cpu/tests/util.rs
```diff
@@ -0,0 +1,164 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![allow(missing_docs, reason = "we don't need docs for testing")]
+#![allow(
+    unused,
+    reason = "for some reason cargo reports the functions/variables here are unused
+when running `cargo test`, even though they are used."
+)]
```
DJMcNab, line 9: This is because some of the code is unused in certain test cases. That is, this module is compiled at least twice, once for `issues.rs` and once for `basic.rs`.

DJMcNab (approved):
Excellent to get some tests in!

sparse_strips/vello_cpu/tests/util.rs
```diff
@@ -0,0 +1,164 @@
...
+
+fn is_pix_diff(pixel1: &Rgba<u8>, pixel2: &Rgba<u8>) -> bool {
+    if pixel1.0[3] == 0 && pixel2.0[3] == 0 {
+        return false;
+    }
+
+    pixel1.0[0] != pixel2.0[0]
+        || pixel1.0[1] != pixel2.0[1]
+        || pixel1.0[2] != pixel2.0[2]
+        || pixel1.0[3] != pixel2.0[3]
```
tomcur, line 163: This can be written more concisely as the following.

```suggestion
    if pixel1.0[3] == 0 && pixel2.0[3] == 0 {
        false
    } else {
        pixel1 != pixel2
    }
```

tomcur: The entire if-statement could be folded into a single expression, but the explicit branches as you've written now do make the intention clearer.

sparse_strips/vello_cpu/tests/issues.rs
LaurenzV: Will just do this for all then!

tomcur (approved):

sparse_strips/vello_cpu/tests/util.rs
DJMcNab: I'd even consider:
```suggestion
    if pixel1.0[3] == 0 && pixel2.0[3] == 0 {
        return false
    }

    pixel1 != pixel2
```
to maybe be better style, to make it really obvious that it's a "both transparent" short-circuit

LaurenzV:
I think the Git LFS thing got messed up again now... https://github.com/linebender/vello/blob/tests/sparse_strips/vello_cpu/snapshots/filled_circle_with_opacity.png It's not showing it anymore as being stored there, even though I think I changed all of the paths... Maybe it's because I added them in the same commit.

LaurenzV:
Okay, seems to work now again.

PR #833, opened 2025-03-05
Refactor LFS file handling in CI slightly

DJMcNab:
This ensures:
1) The hashing differences on windows and linux don't impact us - https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Parley.20CI.20failing.20.28due.20to.20LFS.20failure.3F.29/near/500383610
2) Only explicitly asked for LFS files are downloaded (avoiding accidental successes if these get out of sync)
3) There's a single source of truth for which files are impacted

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ env:
   # List of packages that can not target Wasm.
   # `vello_tests` uses `nv-flip`, which doesn't support Wasm.
   NO_WASM_PKGS: "--exclude vello_tests --exclude simple_sdl2"
+  # The files stored in LFS the tests need to access
+  LFS_FILES: "'vello_tests/snapshots/*.png'"
 
 
 # Rationale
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ env:
   # List of packages that can not target Wasm.
   # `vello_tests` uses `nv-flip`, which doesn't support Wasm.
   NO_WASM_PKGS: "--exclude vello_tests --exclude simple_sdl2"
+  # The files stored in LFS the tests need to access
+  LFS_FILES: "vello_tests/snapshots/*.png"
```
DJMcNab, line 19: I'm not sure how well this will handle adding additional files.
I think the approach is to add a comma within the quotes, which I think might work in both places. But it's a bit unclear

xStrom: We can cross that bridge when we actually need to. It's just a question of style - there are multiple possible solutions. 👍

xStrom (approved):

.github/workflows/ci.yml
DJMcNab: @xStrom we need to cross this bridge in #832
I'm a bit lost

DJMcNab: The issue is isomorphic to https://github.com/orgs/community/discussions/25761#discussioncomment-6508758, I believe

xStrom: Aha, that soon. I can look at it later today and/or tomorrow if you guys haven't solved it before then.

PR #834, opened 2025-03-05
Add the `vello_toy` crate

LaurenzV:
This PR proposes adding a new crate called `vello_toy`, containing a number of small binary crates that build on top of the other ones. Currently, it only contains a single tool `debug` that allows to visualize the intermediate stages of sparse strip rendering, but I think there are a couple of other binary crates that could be added in the future (another idea of mine being for example building a very simplistic SVG renderer on top of usvg).

The current tool I added serves two purposes:
- It allows newcomers to experiment with and visualize the various stages, which makes it easier to understand how everything works.
- It allows for debugging issues in specific parts of the pipelines more easily.

I know there could be more documentation on what exactly each of the parameters mean and how to use them, but since it's mainly for development purposes I think this should be enough. To demonstrate this with an example:

```
cargo run --bin debug -- --path "M 5 5 L 40 23 L 7 44 Z"  --stages ls,ta,ti
```
grafik

```
cargo run --bin debug -- --path "M 5 5 L 40 23 L 7 44 Z"  --stages ls,ta,ti,sa,s
```
grafik

```
cargo run --bin debug -- --path "M 5 5 L 40 23 L 7 44 Z"  --stages ls,ta,ti,sa,s,wt
```
grafik


LaurenzV:
If we don't want to create a separate crate I'm open for that, but I'm not sure if something like this is worth putting in `examples`, maybe it is? Another disadvantage of that would be that `vello_common` would require a dev dependency on clap and svg.

grebmeg:
That sounds like a great idea! I really liked your debug example in the `cpu-sparse-experiments` repo and actually started building a web debugger on top of it (attached a video). I wanted to share it earlier but didn’t get the chance. Initially, I thought it might be a good fit as one of the examples. What do you think about the web debugger? Would it be useful for you as well?


https://github.com/user-attachments/assets/9dbf4b80-deb9-4284-82ec-ad1f5dda56d5

LaurenzV:
Ohh yeah, this looks pretty neat! :) Maybe that would be a better way of implementing this, since it's interactive.

DJMcNab:
> This should be `**/debug.svg`, right? Otherwise it will only get excluded at the root of the repo.

Plausibly? I was intentionally only having it be at the root; just because having "global" `gitignore` things has the potential to be very confusing at later points.
In most cases, if you're running the code, it will be from the workspace root.

LaurenzV:
I have no idea what's going on with formatting... I ran it locally on my computer (I have the newest version), and that's the formatting I get, but for some reason CI doesn't like it. The weird thing is that the exact same import order worked fine in CI yesterday! But it doesn't seem like the version changed or anything...

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
+debug.svg
```
DJMcNab, line 2: Fwiw, this is still equivalent to `**/debug.svg`, as it doesn't have the leading `/`.
From the [`.gitignore` docs](https://git-scm.com/docs/gitignore#_pattern_format):

>For example, "`**/foo`" matches file or directory "`foo`" anywhere, the same as pattern "`foo`".

sparse_strips/vello_toy/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "vello_toy"
+version.workspace = true
```
DJMcNab, line 3: I think this is a package which we'll never release, and so it should not have a version.
See also that this package doesn't have `LICENSE-MIT`/`LICENSE-APACHE` files.

```suggestion
```

sparse_strips/vello_toy/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "A collection of different binary crates for interacting with the sparse rendering stack."
+categories = ["rendering", "graphics"]
+keywords = ["2d", "vector-graphics"]
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+# Prevent accidental publishing until the initial release
```
DJMcNab, line 11:
```suggestion
```
As in https://github.com/linebender/vello/blob/a61783396b6d00f187235b2b1fccb0d91144d92a/vello_tests/Cargo.toml#L7-L9

sparse_strips/vello_toy/src/debug.rs
```diff
@@ -0,0 +1,442 @@
...
+    type Err = String;
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        match input.to_lowercase().as_str() {
+            "ls" => Ok(Self::LineSegments),
+            "ta" => Ok(Self::TileAreas),
+            "ti" => Ok(Self::TileIntersections),
+            "sa" => Ok(Self::StripAreas),
+            "s" => Ok(Self::Strips),
+            "wt" => Ok(Self::WideTiles),
```
DJMcNab, line 383:
```suggestion
            "ls" | "line_segments" => Ok(Self::LineSegments),
            "ta" | "tile_areas" => Ok(Self::TileAreas),
            "ti" | "tile_intersections" => Ok(Self::TileIntersections),
            "sa" | "strip_areas" => Ok(Self::StripAreas),
            "s" | "strips" => Ok(Self::Strips),
            "wt" | "wide_tiles" => Ok(Self::WideTiles),
```

sparse_strips/vello_toy/src/debug.rs
```diff
@@ -0,0 +1,442 @@
...
+            _ => Err(format!("invalid stage: {}", input)),
```
DJMcNab, line 384:
```suggestion
            _ => Err(format!("invalid stage: {}. Expected one of `line_segments`, `tile_areas`, `tile_intersections`, `strip_areas`, `strips`, or `wide_tiles`, or their acronym", input)),
```

sparse_strips/vello_toy/src/debug.rs
```diff
@@ -0,0 +1,442 @@
...
+
+fn parse_path(val: &str) -> Result<BezPath, String> {
+    BezPath::from_svg(val).map_err(|_| "failed to parse the SVG path".to_string())
+}
+
+fn parse_fill_rule(val: &str) -> Result<Fill, String> {
+    match val {
+        "nonzero" => Ok(Fill::NonZero),
+        "evenodd" => Ok(Fill::EvenOdd),
+        _ => Err(format!("unsupported fill rule: {}", val)),
```
DJMcNab, line 440:
```suggestion
        _ => Err(format!("unsupported fill rule: {}. Expected one of `nonzero` or `evenodd`", val)),
```

DJMcNab (approved):
This is great! A few notes on improving UX cheaply.

DJMcNab:
And yes, absolutely if the core of this can be turned into an interactive web tool, that would be very welcome.

I guess it's not really something we can render ourselves, because debugging a broken renderer using itself doesn't seem like a good time :)

PR #835, opened 2025-03-05
Simplify strip generation and analytical coverage calculation

tomcur:
The footprint calculation is removed from strip generation. Analytical pixel coverage and the strip boundaries are now calculated in a single forward scan through the tiles. This simplification potentially makes it easier to base SIMD versions on the math here. (As an aside, AVX512 could do a 4x4 tile in `f32` in one go.)

This is a performance improvement: the code in main generates strips for a 200x200 Tiger in 0.475 ms on my hardware (1000 iterations). The code in this PR generates the strips in 0.359 ms. 

This also appears to improve robustness: for example, the first tiger is rendered by main, the right tiger is rendered by the code in this PR.

test-orig 
test

The calculations here are ready for lines being removed from tiles, which is a likely future optimization. With lines currently being packed into tiles, some calculations here do a bit more work than necessary.

There are two follow-ups to this I'm planning. One is culling tiles to the left of the viewport, which will allow storing tile x-coordinates unsigned. The other is to elide horizontal geometry at the tile-generation stage, as horizontal geometry doesn't impact winding. That would include generalizing sparse fills to sparse alpha masks, where the alpha values are invariant over the columns.

As drive-by changes I've improved some comments, and changed some constants to be associated constants on `Tile` and `WideTile` for easier importing. As we're still a bit messy with `u16`, `i32`, `usize`, etc., some more clean-up will be required in the future changes.

Some snapshots have some slight rounding differences of colors on some pixels. The color values of `filled_vertical_hairline_rect.png` match those calculated by hand, whereas the old snapshot on main does not.


sparse_strips/vello_common/src/coarse.rs
```diff
@@ -9,7 +9,7 @@ use vello_api::paint::Paint;
 use vello_api::peniko::Fill;
 
 /// The width of a wide tile.
-pub const WIDE_TILE_WIDTH: usize = 256;
+pub const WIDE_TILE_WIDTH: usize = WideTile::WIDTH as usize;
```
tomcur, line 12: These changes are unfortunate, but I don't think doing the clean up in this PR is the right timing. When left-of-viewport tile culling lands and the x-coordinates go unsigned, is probably a better time.

LaurenzV:
Interesting, I'm pretty sure that at least at some point the tiger rendered without robustness issues on my branch... Perhaps I introduced a bug on the way, but if it's solved now, then even better! :)

LaurenzV:
I backported this to my branch and ran the banchmarks again (on M1). Those were the results I got before:

```
render_strips/gs_tiger - Scalar
                        time:   [707.92 µs 709.66 µs 711.48 µs]
render_strips/gs_tiger - Neon
                        time:   [577.51 µs 578.98 µs 580.57 µs]
render_strips/coat_of_arms - Scalar
                        time:   [3.2907 ms 3.3023 ms 3.3129 ms]
render_strips/coat_of_arms - Neon
                        time:   [2.6149 ms 2.6317 ms 2.6523 ms]
```

And those are the results I get now:

```
render_strips/gs_tiger - Scalar
                        time:   [569.88 µs 573.73 µs 579.45 µs]
                        change: [-20.007% -19.613% -19.069%] (p = 0.00 < 0.05)

render_strips/coat_of_arms - Scalar
                        time:   [2.7394 ms 2.7943 ms 2.8646 ms]
                        change: [-16.775% -15.853% -14.766%] (p = 0.00 < 0.05)
```

So it definitely seems like a win in scalar mode, and it's now about roughly as fast as the manual NEON implementation, so good job! :) I wonder if there is even more to be gained from manually vectorizing this, or whether the compiler can already perfectly do that.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -41,100 +42,51 @@ pub fn render(
...
-            if same_strip {
-                fp.extend(3);
-            }
+    /// A special tile to keep the logic below simple.
+    const GATE_CLOSER: Tile = Tile {
+        x: i32::MAX,
+        y: u16::MAX,
+        p0: Point::ZERO,
+        p1: Point::ZERO,
+    };
```
LaurenzV, line 69: Just a personal opinion, so feel free to ignore, but I think I would prefer `SENTINEL_TILE` or `DUMMY_TILE`.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -41,100 +42,51 @@ pub fn render(
...
-            }
+    // The strip we're building.
+    let mut strip = Strip {
+        x: prev_tile.x * Tile::WIDTH as i32,
+        y: prev_tile.y * Tile::HEIGHT,
+        col: alpha_buf.len() as u32,
+        winding: 0,
+    };
 
+    for tile in tiles.iter().copied().chain([GATE_CLOSER]) {
```
LaurenzV, line 79: Do we really need to copy each tile here? Using references doesn't work? I'd assume this would be slightly cheaper?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -155,36 +106,163 @@ pub fn render(
...
+        // Push out the strip if we're moving to a next strip.
+        if !prev_tile.same_loc(&tile) && !prev_tile.prev_loc(&tile) {
+            if !prev_tile.same_row(&tile) {
+                winding_delta = 0;
+            }
+
+            debug_assert_eq!(
+                (prev_tile.x + 1) * Tile::WIDTH as i32 - strip.x,
+                alpha_buf.len() as i32 - strip.col as i32
+            );
```
LaurenzV, line 125: Could you maybe add a brief comment explaining what exactly this checks?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -155,36 +106,163 @@ pub fn render(
...
-        fp.merge(&cur_tile.footprint());
+        for y_idx in 0..Tile::HEIGHT {
+            let px_top_y = y_idx as f32;
+            let px_bottom_y = 1. + y_idx as f32;
+
+            let ymin = f32::max(line_top_y, px_top_y);
+            let ymax = f32::min(line_bottom_y, px_bottom_y);
+
+            let mut acc = 0.;
+            for x_idx in 0..Tile::WIDTH {
```
LaurenzV, line 232: This was mentioned in the meeting briefly, but could you explain again whether in theory, it'd be possible to reverse the order of for loops and re-introduce the footprint, so we don't unnecessarily include columns that are full of 0? What is the main rationale for changing this, just performance?

LaurenzV (commented):
Some comments, but I will leave the proper review of the logic to Raph.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -36,6 +36,12 @@ pub struct Tile {
 }
 
 impl Tile {
+    /// The width of a tile in pixels.
+    pub const WIDTH: u16 = 4;
+
+    /// The height of a tile in pixels.
+    pub const HEIGHT: u16 = 4;
+
```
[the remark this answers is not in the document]
tomcur: Yes, I believe there was some talk about that possibility.

sparse_strips/vello_common/src/strip.rs
tomcur: The `copied()` doesn't change generated code, it's a construct of the Rust type system; but I think getting owned `Tile`s here is more ergonomic within the loop body.

sparse_strips/vello_common/src/strip.rs
tomcur: That should be possible. The current order makes sense for writing Rust in a natural way, as you first calculate the `y`-values per row and then iterate the columns in the row using those values. You could also calculate all `y`-values upfront and collect them into a `Tile::HEIGHT`-sized one-dimensional array, and then iterate the columns as the outer loop. I've not measured whether that has a performance impact.

Note that the SIMD versions are likely going to calculate multiple columns at the same time (in the AVX512 case you could do an entire 4x4 tile at once). It's possible for certain configurations of tile size that just checking and omitting zero-columns after the fact may be faster than calculating a footprint upfront. I suspect that may be the case for 4x4 tiles.

One other consideration is that it's not just zeros we're interested in, but in principle all trailing columns that are identical to the column before them can be dropped.

tomcur:
> I wonder if there is even more to be gained from manually vectorizing this, or whether the compiler can already perfectly do that.

I suspect it can be improved by manual vectorization. For an optimized standard x86 build the compiler doesn't vectorize this, but it does completely unroll the nested loops and does some smart register allocation. For a tile size of e.g. 16x16 it stops doing that, which appears to have a serious performance impact.

sparse_strips/vello_common/src/strip.rs
LaurenzV: Fair enough, a while back I did an analysis of the length of the footprints for the ghostscript tiger and coat of arms, and in 80/85% of the cases it was always 4 wide, so it's possible that just always calculating everything does lead to better performance, though it probably depends on the exact workload. 

But I think this might be a question reconsidering for 8x8 tiles, on the one hand because there probably will be more tiles that don't use the full width, but on the other hand also because of the increased memory usage, and even with SIMD you will have a lot of columns that need processing.

tomcur: That's good to know. Glyphs would generate many narrow tiles, though, and those should definitely be optimized for as well. Footprints or something similar could be added back in, perhaps on the level of individual tiles (if lines are going to be removed from the tile packing, you need to do the bounds calculations anyway, and this already does most of the calculations necessary).

LaurenzV (approved):
As mentioned we should definitely also ask for approval from @raphlinus, but overall this seems like a step in the right direction!

sparse_strips/vello_common/src/strip.rs
raphlinus: This captures exactly my thoughts. I'm in favor of simplification, and rather easily buy the argument that footprints aren't worth the complexity for 4x4, but I definitely see us moving to larger tiles (especially for CPU/GPU hybrid), and there I think *some* kind of footprint mechanic would be worthwhile.

raphlinus (approved):
Ok, this looks good, and thanks. I haven't gone through the math with a fine-grained comb, but do understand the general approach and am convinced by the test data.

It's clear that the speed gain is mostly because the new code lends itself well to autovectorization, but I also see opportunities to squeeze more. I also didn't look at x86 code yet. It wouldn't be surprising to me if autovectorization (of the core area accumulation loop) could only exploit 4 lanes because of the apparent data dependency of `acc`, which if so I'm confident we can improve by using standard prefix sum techniques.

I have very complicated feelings about the footprints, but those are mostly in line with the existing comment thread. I'll start a new thread on Zulip with half-baked thoughts on medium-grain sparseness.

PR #836, opened 2025-03-05
Fix clippy::large_enum_variant warning in simple example.

xorgy:


xorgy:
Encountered this with updated clippy.
```rust
warning: large size difference between variants
  --> examples/simple/src/main.rs:29:1
   |
29 | / enum RenderState<'s> {
30 | |     Active(ActiveRenderState<'s>),
   | |     ----------------------------- the largest variant contains at least 304 bytes
31 | |     // Cache a window so that it can be reused when the app is resumed after being suspended
32 | |     Suspended(Option<Arc<Window>>),
   | |     ------------------------------ the second-largest variant contains at least 8 bytes
33 | | }
   | |_^ the entire enum is at least 304 bytes
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
   = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields to reduce the total size of the enum
   |
30 -     Active(ActiveRenderState<'s>),
30 +     Active(Box<ActiveRenderState<'s>>),
   |
```

examples/simple/src/main.rs
```diff
@@ -20,15 +20,14 @@ use winit::window::Window;
 use vello::wgpu;
 /// Simple struct to hold the state of the renderer
```
DJMcNab, line 21: I wouldn't normally describe an enum as a struct.

examples/simple/src/main.rs
```diff
@@ -93,12 +95,12 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
         event: WindowEvent,
     ) {
         // Ignore the event (return from the function) if
-        //   - we have no render_state
+        //   - `RenderState` is `Suspended`
```
DJMcNab, line 98: Maybe just "the app is suspended"

examples/simple/src/main.rs
```diff
@@ -121,12 +123,8 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
                 // Re-add the objects to draw to the scene.
                 add_shapes_to_scene(&mut self.scene);
 
-                // Get the RenderSurface (surface + config)
-                let surface = &render_state.surface;
-
                 // Get the window size
-                let width = surface.config.width;
-                let height = surface.config.height;
+                let wgpu::SurfaceConfiguration { width, height, .. } = surface.config;
```
DJMcNab, line 127: I'd probably restore the explicit access to width and height here. I don't think this form is clearer, as `SurfaceConfiguration` is a weird type.

examples/simple/src/main.rs
nicoburns: I believe it is the window which is suspended, not necessarily the whole app?

xorgy: Suspended RenderState is not necessarily suspended app. It is also the default state during initialization.

examples/simple/src/main.rs
xorgy: I wouldn't either, this is just a residual comment lol.

examples/simple/src/main.rs
xorgy: Personally I don't see the point of this comment... it just restates the obvious meaning of the code that directly follows.

xorgy: I've written a more concise comment that addresses the purpose of the code rather than restating it.

nicoburns (approved):

waywardmonkeys:
Since @DJMcNab had comments, I'm going to leave it for him to confirm that he's happy with the resolution.

DJMcNab (approved):
LGTM

examples/simple/src/main.rs
DJMcNab: This `simple` example is designed to be over-commented. It's not my preference, but my preference would be to not have this example at all :)

I agree that the new shorter comment is fine.

PR #837, opened 2025-03-06
Use correct seperation in each place needed when loading LFS

DJMcNab:
https://github.com/actions/toolkit/blob/d70fb49aaa0517f8c92e4e2e9406c8a5414f3fec/packages/glob/src/glob.ts#L11 indicated that newlines were supported, and it worked

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,7 +215,7 @@ jobs:
           enableCrossOsArchive: true
   
       - name: Checkout LFS files
-        run: git lfs checkout '${{ join(fromJson(env.LFS_FILES), ',') }}'
+        run: git lfs checkout '${{ join(fromJson(env.LFS_FILES), fromJson(''' ''')) }}'
         continue-on-error: true
 
       - name: install stable toolchain
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ env:
   # List of packages that can not target Wasm.
   # `vello_tests` uses `nv-flip`, which doesn't support Wasm.
   NO_WASM_PKGS: "--exclude vello_tests --exclude simple_sdl2"
-  # The files stored in LFS the tests need to access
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -168,7 +168,7 @@ jobs:
 
     - id: calc-hash
       name: Calculate LFS hash
-      run: echo "result=${{ hashFiles(env.LFS_FILES) }}" | tee -a "$GITHUB_OUTPUT"
+      run: echo "result=${{ hashFiles(join(fromJson(env.LFS_FILES), fromJson('"\n"'))) }}" | tee -a "$GITHUB_OUTPUT"
```
LaurenzV, line 171: Just doing `fromJson('\n')` doesn't work?

.github/workflows/ci.yml
```diff
@@ -215,7 +215,7 @@ jobs:
           enableCrossOsArchive: true
   
       - name: Checkout LFS files
-        run: git lfs checkout '${{ env.LFS_FILES }}'
+        run: git lfs checkout '${{ join(fromJson(env.LFS_FILES), ''' ''') }}'
```
LaurenzV, line 218: What exactly does ''' ''' do?

LaurenzV (approved):
If it works, it works! Just two things I'm wondering about.

.github/workflows/ci.yml
DJMcNab: I don't think so. This monstrosity is what was required to get a `\n` character into GitHub's expression context.

`fromJson('\n')` would be parsing the string `\n` as a json object, which would error at the `\`, as the only valid first characters in a JSON file are `[`, `{`, [whitespace], and `"`.
And just `\n` is also the literal string `\n`, which isn't the right join.

.github/workflows/ci.yml
DJMcNab: It's the string `' '`, but quoted in the way GitHub requires, with a `'` quoted as `''` inside a string.

PR #840, opened 2025-03-08
Add a very basic SVG render tool to `vello_toy`

LaurenzV:
Example:
`cargo run --bin svg --release  -- --path "C:\Users\lstam\Documents\Programming\vello\examples\assets\Ghostscript_Tiger.svg" --scale 5`

svg

Obviously far from feature complete, but my goal for now is to only implement very basic stuff.


sparse_strips/vello_toy/src/svg.rs
```diff
@@ -0,0 +1,254 @@
...
+#[derive(Parser, Debug)]
+struct Args {
+    /// How much to scale the output image.
+    #[arg(long, default_value_t = 1.0)]
+    pub scale: f32,
+    /// The path to the SVG file.
+    #[arg(long)]
+    pub path: String,
+    /// The target runtime for calculating how many iterations to run, in milliseconds.
+    #[arg(long, default_value_t = 2000)]
```
DJMcNab, line 99: I'd prefer for this to default to running a single iteration, i.e.:
```suggestion
    /// For simple benchmarking, our current recommended value is 2000 (i.e. 2 seconds).
    #[arg(long, default_value_t = 0)]
```

I do agree that the benchmarking is a useful tool, but I don't think it should be the default.
(I can also see arguments for a `benchmark` flag, but that's probably too much complication for now)

sparse_strips/vello_toy/src/svg.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        if runtime.as_millis() > args.runtime as u128 {
+            break;
+        }
+    }
+
+    let avg_runtime = (runtime.as_millis() as f32) / (num_iters as f32);
+
+    eprintln!(
+        "ran {} iterations, with an average runtime of {}ms to render {}.",
```
DJMcNab, line 63: Given that we run until the current item finishes, I think the priority should be on the `average_runtime`, rather than on the number of iterations.

sparse_strips/vello_toy/src/svg.rs
```diff
@@ -0,0 +1,254 @@
...
+        num_iters, avg_runtime, file_name
```
DJMcNab, line 64: Would it also make sense to display the standard deviation? I think that's harder to calculate; we'd need to save all the results?
So probably not worth doing, then

DJMcNab (approved):
LGTM. A couple of details

DJMcNab:
Is that referring to #831 (adding Vello Hybrid, which has now landed)?

LaurenzV:
Yep!

LaurenzV:
Would still be nice to have and I'll probably look into it again at some point, but since this branch is very stale I'll close it for now.

PR #841, opened 2025-03-08
Fix COLRv1 for Segoe UI Emoji

dfrg:
The yx and xy transform components were swapped in the conversion. Also the repeat/reflect extend modes.

See zulip: https://xi.zulipchat.com/#narrow/channel/205635-text/topic/Purple.20emoji.20on.20Windows/with/503831553

PR #844, opened 2025-03-09
Add v0.4.1 changelog entry

nicoburns:
See [#gpu > Vello 0.4.1 release](https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/Vello.200.2E4.2E1.20release)

This release is to backport #841

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 # Changelog
 
-The latest published Vello release is [0.4.1](#041---2025-03-09) which was released on 2025-03-09.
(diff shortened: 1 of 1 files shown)


waywardmonkeys (approved):
LGTM, but should probably only land once 0.4.1 does in case something changes over in that area.

waywardmonkeys:
And thanks @DJMcNab for fixing it up!

DJMcNab:
My plan is the same as for 0.2.1, which is to land the changelog PR on main, cherry-pick that and the other relevant PRs into a "release" branch, get that approved and then land the PR, then make the release. So in that plan, I merge this one now.

I've double checked everything merged since 0.4.0, and nothing else should go in here as well.

PR #845, opened 2025-03-10
Remove line packing from tiles, slightly improve tile culling

tomcur:
This reduces the tile memory footprint and drops the requirement for tile x-coordinates to be signed. We may be able to drop the `TileIndex` indirection soon. 

Coarse winding calculation is performed in tile generation, like it more-or-less was before. In principle it requires just a single bit in the tile packing. Calculating the coarse winding could be performed in the strip generation as well, but I suspect it to be slightly more performant this way. I haven't accurately measured that.

This is one step towards what [Raph mentioned](https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/CPU.20sparse.20strip.20rendering.20to.20pixels/near/500019583). Quoting:

> pack x, y, line id, and winding number delta into a single u32 and sort that. This is a very tight packing with u32 and may overflow in some cases (very large number of lines per path), but I think should be very efficient. Then the determination of p0 and p1 within the tile happens after sorting. A major motivation for this change is that it can be done on GPU in hybrid modes, but the approach to parallelism also works with SIMD, I think.

Tiles above, to the right and below the viewport are now culled. Tiles to the left of the viewport are clamped but not culled yet. Currently strip generation has some code special-casing those tiles.

On my machine, paris-30k measures with #835 as the "before" and after the changes here as the following. (With 20 iterations, appears to be accurate to about ~0.2ms, but we should think about better benchmarking.)

```
Tile generation
    old: 24.394854ms
    new: 17.620834ms
Tile sorting
    old: 8.922622ms
    new: 8.146455ms
Strip generation
    old: 33.187645ms
    new: 31.825052ms
```

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -5,7 +5,7 @@
 
 use peniko::Fill;
 
-use crate::flatten::Point;
+use crate::flatten::Line;
 use crate::tile::{Tile, Tiles};
 
 // Note that this will probably disappear and be turned into a const generic in the future.
@@ -39,6 +39,7 @@ pub fn render(
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u32>,
     fill_rule: Fill,
+    lines: &[Line],
 ) {
     strip_buf.clear();
 
@@ -64,8 +65,8 @@ pub fn render(
     const SENTINEL: Tile = Tile {
         x: i32::MAX,
         y: u16::MAX,
-        p0: Point::ZERO,
-        p1: Point::ZERO,
+        line_idx: 0,
+        winding: false,
     };
 
     // The strip we're building.
@@ -77,6 +78,14 @@ pub fn render(
     };
 
     for tile in tiles.iter().copied().chain([SENTINEL]) {
+        let line = lines[tile.line_idx as usize];
+        let tile_left_x = tile.x as f32 * Tile::WIDTH as f32;
+        let tile_top_y = tile.y as f32 * Tile::HEIGHT as f32;
+        let p0_x = line.p0.x - tile_left_x;
+        let p0_y = line.p0.y - tile_top_y;
+        let p1_x = line.p1.x - tile_left_x;
+        let p1_y = line.p1.y - tile_top_y;
+
         // Push out the winding as an alpha mask when we move to the next location (i.e., a tile
         // without the same location).
         if !prev_tile.same_loc(&tile) {
@@ -127,14 +136,30 @@ pub fn render(
             );
             strip_buf.push(strip);
 
-            // Once we've reached the `SENTINEL` tile, emit a final strip.
-            if tile.y == u16::MAX && tile.x == i32::MAX {
-                strip_buf.push(Strip {
-                    x: i32::MAX,
-                    y: u16::MAX,
-                    col: alpha_buf.len() as u32,
-                    winding: 0,
-                });
+            let is_sentinel = tile.y == u16::MAX && tile.x == i32::MAX;
+            if !prev_tile.same_row(&tile) {
+                // Emit a final strip in the row if there is non-zero winding for the sparse fill,
+                // or unconditionally if we've reached the sentinel tile to end the path (the `col`
+                // field is used for width calculations).
+                if winding_delta != 0 || is_sentinel {
+                    strip_buf.push(Strip {
+                        x: i32::MAX,
+                        y: prev_tile.y * Tile::HEIGHT,
+                        col: alpha_buf.len() as u32,
+                        winding: winding_delta,
+                    });
+                }
+
+                winding_delta = 0;
+                accumulated_winding.fill(0.);
+
+                #[expect(clippy::needless_range_loop, reason = "dimension clarity")]
+                for x in 0..Tile::WIDTH as usize {
+                    location_winding[x].fill(0.);
+                }
+            }
+
+            if is_sentinel {
                 break;
             }
 
@@ -149,14 +174,6 @@ pub fn render(
             accumulated_winding.fill(winding_delta as f32);
         }
         prev_tile = tile;
-
-        // TODO: lines are currently still packed into tiles. This will probably change, in which
-        // case we will have to translate the lines to have the tile's top-left corner as origin.
-        // let line = lines[tile.line_idx as usize];
-        let p0_x = tile.p0.x; // - tile_left_x;
-        let p0_y = tile.p0.y; // - tile_top_y;
-        let p1_x = tile.p1.x; // - tile_left_x;
-        let p1_y = tile.p1.y; // - tile_top_y;
 
         // TODO: horizontal geometry has no impact on winding. This branch will be removed when
         // horizontal geometry is culled at the tile-generation stage.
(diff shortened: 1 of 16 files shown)


sparse_strips/vello_common/src/strip.rs
```diff
@@ -38,103 +39,63 @@ pub fn render(
...
+    };
+
+    for tile in tiles.iter().copied().chain([SENTINEL]) {
+        let line = lines[tile.line_idx as usize];
+        let tile_left_x = tile.x as f32 * Tile::WIDTH as f32;
+        let tile_top_y = tile.y as f32 * Tile::HEIGHT as f32;
+        let p0_x = line.p0.x - tile_left_x;
+        let p0_y = line.p0.y - tile_top_y;
+        let p1_x = line.p1.x - tile_left_x;
+        let p1_y = line.p1.y - tile_top_y;
```
tomcur, line 87: With these calculations placed at the front of the loop, strip generation measures at ~1-2% faster for paris-30k on my machine.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -3,48 +3,42 @@
...
-    pub p0: Point,
-    /// The end point of the line in that tile.
-    pub p1: Point,
+    /// The index of the line this tile belongs to into the line buffer.
+    pub line_idx: u32,
+    /// Whether the line crosses the top edge of the tile.
+    ///
+    /// Lines making this crossing increment or decrement the coarse tile winding, depending on the
+    /// line direction.
+    pub winding: bool,
```
tomcur, line 25: This can become a single bit in a tight packing.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -3,48 +3,42 @@
...
 /// A tile represents an aligned area on the pixmap, used to subdivide the viewport into sub-areas
 /// (currently 4x4) and analyze line intersections inside each such area.
 ///
 /// Keep in mind that it is possible to have multiple tiles with the same index,
 /// namely if we have multiple lines crossing the same 4x4 area!
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub struct Tile {
     /// The index of the tile in the x direction.
     pub x: i32,
```
tomcur, line 16: This should become `u16`, probably as part of this PR, which I'd prefer doing as a separate commit (many non-interesting changes will follow from that). With this PR currently being based on top of another PR, I'll hold off on that for a bit to keep things clear.

LaurenzV: The end goal is that a `Tile` can be represented by just u32, right? This would mean that we need to first check whether any of the components can overflow, and if so use a "long path" that doesn't contain this tight packing, right? In practice this would hopefully never be necessary unless there are a lot of lines in a single path or x/y are very big, but for correctness it would still be necessary to have.

sparse_strips/vello_common/src/tile.rs
LaurenzV: Before, we had a delta which can be -1, 0, 1. Am I correct in the assumption that `winding == true` corresponds to either -1/1, the exact delta being determined after the sorting?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -128,269 +117,130 @@ impl Tiles {
...
-            // strip generation code, but it works for now.
-            if p.x.fract() == 0.0 {
-                Point {
-                    x: p.x + SCALED_X_NUDGE_FACTOR,
-                    y: p.y,
-                }
-            } else {
-                p
-            }
-        };
```
LaurenzV, line 158: All of this is not necessary anymore due to your new strip generation algorithm?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -128,269 +117,130 @@ impl Tiles {
...
+    /// Populate the tiles' container with a buffer of lines.
+    ///
+    /// Tiles exceeding the top, right or bottom of the viewport (given by `width` and `height` in
+    /// pixels) are culled.
+    //
+    // TODO: Tiles are clamped to the left edge of the viewport, but lines fully to the left of the
+    // viewport are not culled yet. These lines impact winding, and would need forwarding of
+    // winding to the strip generation stage.
```
LaurenzV, line 141: Above you mentioned x can be stored as u16, so how do you distinguish between lines at 0 and lines strictly to the left?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -128,269 +117,130 @@ impl Tiles {
...
+            } else {
+                (p1_x, p0_x)
+            };
+            let (line_top_y, line_top_x, line_bottom_y, line_bottom_x) = if p0_y < p1_y {
+                (p0_y, p0_x, p1_y, p1_x)
+            } else {
+                (p1_y, p1_x, p0_y, p0_x)
+            };
+
+            // For ease of logic, special-case purely vertical tiles.
```
LaurenzV, line 182: Are purely horizontal lines not worth special-casing?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -128,269 +117,130 @@ impl Tiles {
...
-                for i in 0..tile_count_x - 1 {
-                    let yclip = yclip0 + i as f32 * sign * slope;
-                    let yfrac = scale_up(yclip).max(NUDGE_FACTOR);
-                    let packed = Point::new(xclip, yfrac);
-
-                    push_tile(x, y, last_packed, packed);
-
-                    last_packed = Point::new(packed.x + flip, packed.y);
+            } else {
+                let x_slope = (p1_x - p0_x) / (p1_y - p0_y);
```
LaurenzV, line 197: Can this not be NaN for horizontal lines?

LaurenzV (commented):
I think the proper review will wait until the previous PR has been merged for easiniess, but just a couple of (perhaps dumb) questions from my side. Otherwise, the overall direction seems sensible to me!

tomcur:
Spot-checking some pixels, the changed tests all appear to just be the rounding of pixel coverage falling differently. I haven't checked too closely, but it's possible the area coverage calculation are by coincidence slightly more symmetrical now, check e.g. the left and right edges of `eo_filling_missing_anti_aliasing.png` before and after these changes.)

sparse_strips/vello_common/src/tile.rs
tomcur: No, but it can be `inf`. The `min` and `max` below take care of that. Admittedly, I haven't verified that _too_ closely since horizontal lines don't impact geometry and can be elided, which could be done shortly after this lands in a follow-up PR.

sparse_strips/vello_common/src/tile.rs
tomcur: They are worth special-casing. Horizontal geometry is parallel to the winding scan direction, and so doesn't impact winding. We should evaluate removing them entirely. That does most likely entail adding a sparse alpha mask draw command to the wide tiles.

sparse_strips/vello_common/src/tile.rs
tomcur: This PR adds logic to strip generation to check for left-of-viewport lines. With some plumbing of winding/coverage from earlier pipeline stages into strip generation, those checks are best moved back to tile generation in a follow-up PR. I believe Raph would like to cull some geometry even earlier than tile generation in future changes, which might be able to use some of that same plumbing.

sparse_strips/vello_common/src/tile.rs
tomcur: Yes, `false` corresponds to `0`, and `true` corresponds to `-1` if the line is oriented down and `1` if the line is oriented up. In this PR the sign is resolved during strip generation.

sparse_strips/vello_common/src/tile.rs
tomcur: Indeed. I've attempted to encode as many of the edge-cases into IEEE 754 float semantics as I could (mostly around `inf` and `nan`), to have as few branches and instructions as possible. It could perhaps be improved more, but I haven't found that yet.

sparse_strips/vello_toy/src/debug.rs
```diff
@@ -196,15 +202,25 @@ fn draw_tile_areas(document: &mut Document, tiles: &Tiles) {
...
+        // TODO: how to handle line intersections now lines are not explicitly segmented by tile
+        // generation anymore?
+        let p0 = Point {
+            x: line.p0.x - x as f32,
+            y: line.p0.y - y as f32,
+        };
+        let p1 = Point {
+            x: line.p1.x - x as f32,
+            y: line.p1.y - y as f32,
+        };
```
tomcur, line 223: This is a bit hacky at the moment to get tests to pass. Lines are no longer segmented per tile explicitly, so this is drawing the same line multiple tiles. @LaurenzV do you have some ideas about this? Segmenting here manually might be a bit misleading, as with the code in this PR that doesn't actually happen internally.

sparse_strips/vello_common/src/tile.rs
LaurenzV: Nice!

sparse_strips/vello_toy/src/debug.rs
LaurenzV: I'm fine with removing the tile intersection points if the underlying algorithm is different now.

sparse_strips/vello_common/src/tile.rs
raphlinus: Yes, you'll overflow the u32 packing if the line count exceeds the space available, which I think will be rare, but there would need to be a fallback as it's not impossible.

I guess a major empirical question is whether sorting of u32's is actually much faster than u64's. If not, then it's probably better to avoid the complexity of having two code paths. This question is potentially going to depend on a lot of things, like the architecture it's running on and whether we try to do alternative sorting strategies like radix sort.

raphlinus (approved):
This looks good, and the measured time changes justify the change. It's also consistent with changes I proposed for the sorting.

There's definitely more work to be done to get to the end state, but this seems like a good step.

sparse_strips/vello_toy/src/debug.rs
tomcur: I've removed the intersections, but am opening a follow-up soon after this.

tomcur:
To reduce some effort, tile x-coord data type clean up and packing the tile more tightly are best done together, I think.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -204,23 +221,53 @@ pub fn render(
...
+                let px_top_y = y_idx as f32;
+                let px_bottom_y = 1. + y_idx as f32;
+
+                let ymin = f32::max(ymin, px_top_y);
+                let ymax = f32::min(ymax, px_bottom_y);
+
+                let h = (ymax - ymin).max(0.);
+                accumulated_winding[y_idx as usize] += sign * h;
+
+                for x_idx in 0..Tile::HEIGHT {
```
dominikh, line 262: @tomcur should this have said Tile::WIDTH?

tomcur: Yes, thank you! Opened #848.

PR #847, opened 2025-03-11
Clean up coordinate data types

tomcur:
(For ease of review, this PR has been split. The second part now lives here: https://github.com/linebender/vello/pull/852).

This cleans up the data types around integer `Tile`, `Strip`, `WideTile` and pixel coordinates: 1-dimensional indices are now consistently `u16`, and indices into 2-dimensional buffers are (mostly) `usize`. Making tile x-coordinates unsigned is possible since the left-of-viewport winding calculation changes in https://github.com/linebender/vello/pull/845.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -3,35 +3,34 @@
 
 //! Generating and processing wide tiles.
 
-use crate::color::{AlphaColor, Srgb};
-use crate::strip::{STRIP_HEIGHT, Strip};
-use vello_api::paint::Paint;
-use vello_api::peniko::Fill;
-
-/// The width of a wide tile.
-pub const WIDE_TILE_WIDTH: usize = WideTile::WIDTH as usize;
+use crate::{
+    color::{AlphaColor, Srgb},
+    strip::Strip,
+    tile::Tile,
+};
+use vello_api::{paint::Paint, peniko::Fill};
 
 /// A container for wide tiles.
 #[derive(Debug)]
 pub struct Wide {
     /// The width of the container.
-    pub width: usize,
+    pub width: u16,
     /// The height of the container.
-    pub height: usize,
+    pub height: u16,
     /// The wide tiles in the container.
     pub tiles: Vec<WideTile>,
 }
 
 impl Wide {
     /// Create a new container for wide tiles.
-    pub fn new(width: usize, height: usize) -> Self {
-        let width_tiles = width.div_ceil(WIDE_TILE_WIDTH);
-        let height_tiles = height.div_ceil(STRIP_HEIGHT);
-        let mut tiles = Vec::with_capacity(width_tiles * height_tiles);
+    pub fn new(width: u16, height: u16) -> Self {
+        let width_tiles = width.div_ceil(WideTile::WIDTH);
+        let height_tiles = height.div_ceil(Tile::HEIGHT);
+        let mut tiles = Vec::with_capacity(usize::from(width_tiles * height_tiles));
 
         for w in 0..width_tiles {
             for h in 0..height_tiles {
-                tiles.push(WideTile::new(w * WIDE_TILE_WIDTH, h * STRIP_HEIGHT));
+                tiles.push(WideTile::new(w * WideTile::WIDTH, h * Tile::HEIGHT));
             }
         }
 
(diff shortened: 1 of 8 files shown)


sparse_strips/vello_common/src/tile.rs
```diff
@@ -5,24 +5,33 @@
...
+pub const MAX_LINES_PER_PATH: u32 = 1 << 31;
+
 /// A tile represents an aligned area on the pixmap, used to subdivide the viewport into sub-areas
 /// (currently 4x4) and analyze line intersections inside each such area.
 ///
 /// Keep in mind that it is possible to have multiple tiles with the same index,
 /// namely if we have multiple lines crossing the same 4x4 area!
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(C)]
```
tomcur, line 19: Putting `#[repr(C)]` on this turns out to be a very slight net win (100 iterations of paris-30k).

````
Tile generation
    without: 16.65825ms
    with:    16.540007ms
Tile sorting
    without: 7.7405343ms
    with:    7.038436ms
Strip generation
    without: 31.125238m
    with:    31.41686ms

File: sparse_strips/vello_cpu/src/fine.rs [hunk 2]
```diff
@@ -108,23 +118,26 @@ impl<'a> Fine<'a> {
 }
 
 fn pack(out_buf: &mut [u8], scratch: &ScratchBuf, width: usize, height: usize, x: usize, y: usize) {
-    let base_ix = (y * STRIP_HEIGHT * width + x * WIDE_TILE_WIDTH) * COLOR_COMPONENTS;
+    let base_ix = (y * usize::from(Tile::HEIGHT) * width + x * usize::from(WideTile::WIDTH))
+        * COLOR_COMPONENTS;
 
     // Make sure we don't process rows outside the range of the pixmap.
-    let max_height = (height - y * STRIP_HEIGHT).min(STRIP_HEIGHT);
+    let max_height = (height - y * usize::from(Tile::HEIGHT)).min(usize::from(Tile::HEIGHT));
```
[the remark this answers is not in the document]
LaurenzV: Not sure though, maybe it's not worth it.

sparse_strips/vello_common/src/tile.rs
tomcur: Manually packing a `u64` in the order of `(y, x, line_idx, winding)` and sorting that directly leads to the following.

```
Tile generation  16.987038ms
Tile sorting      7.2218933ms
Strip generation 30.619955ms
```

I think strip generation is improved because cache locality for lines is improved. The tile generation is slower, likely because of the required bit shifting of `line_idx` one to the left. Tile sorting is a bit slower as it now sorts on the entire tile instead of just the coordinates.

We're now in micro-optimization land. Manually packing into a `u64` in the order of `(y, x, winding, line_idx)` leads to the following.

```
Tile generation  16.529137ms
Tile sorting      7.315445ms
Strip generation 29.878174ms
```

This is better for tile generation, and the reason strip generation is improved even more might be because the coarse winding branch inside strip generation is better predicted. Tile sorting is a bit slower than the previous option, likely because line indices are already somewhat sorted from tile generation, but winding is sorted on first, leading to more swapping of tiles.

LaurenzV: Oh, no, sorry. 😅 I just meant inside of the code, you should write a single sentence what the reason for `repr[C]` is (i.e.: `We use repr(c) because benchmarking showed it leads to slightly better performance). That's it. :)

tomcur: I understood, and you are right, but I experimented with some other packings in the meantime. What I wrote above are the results.

tomcur:
Updated the main PR text for the new commit moving to `u64` packing.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -98,71 +97,61 @@ impl Wide {
...
                 // Don't render strips that are outside the viewport.
                 continue;
             }
 
-            if strip.y >= self.height as u16 {
-                // Since strips are sorted by location, any subsequent strips will also be
-                // outside the viewport, so we can abort entirely.
-                break;
-            }
-
```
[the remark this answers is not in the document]
tomcur: Yes. The winding scan direction is horizontal, left-to-right, so tiles above or below (or to the right of) the viewport can be culled without additional mechanisms required.

sparse_strips/vello_cpu/image.png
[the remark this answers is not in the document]
tomcur: Nope!

sparse_strips/vello_cpu/src/fine.rs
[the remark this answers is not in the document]
LaurenzV: Yep I mainly meant for code readability, but it's fine I think.

LaurenzV (approved):
I haven't tried to understand the new logic in great detail, but since tests are passing I'm not very concerned.

Other than that, apart from the image that shouldn't be there, looks good to me!

sparse_strips/vello_common/src/tile.rs
```diff
@@ -17,22 +17,22 @@ pub const MAX_LINES_PER_PATH: u32 = 1 << 31;
...
     ///
-    /// The index is the unsigned number in the 31 least significant bits of this value.
+    /// The layout is as follows, with the bit indices from least to most significant.
     ///
-    /// The last bit is 1 if and only if the lines crosses the tile's top edge. Lines making this
-    /// crossing increment or decrement the coarse tile winding, depending on the line direction.
-    pub packed_winding_line_idx: u32,
-}
+    /// [ 0, 31): The 31-bit index of the line this tile belongs to into the line buffer, plus
+    /// whether the line crosses the top edge of the tile, packed together;
```
dominikh, line 26: > plus whether the line crosses the top edge of the tile, packed together;

That's not quite right, is it? [0, 31) is just the pure index, no other info packed into it.

tomcur: Indeed, that was a vestigial comment from this PR's second commit.

PR #849, opened 2025-03-12
Lift tile width restriction

tomcur:
With the recent changes, the scalar code in `vello_common` and `vello_cpu` is now generic over tile width, but not yet over tile height due to the alpha mask encoding.


tomcur:
(Other tile sizes are not tested, and can't be tested yet, as `Tile` is not `const` generic over its size).

LaurenzV (approved):

Issue #850, opened 2025-03-12
sparse_strips/vello_cpu: broken path clipping


dominikh:
On b05dea1ca469cd736df5b122f4769683ae672ae1, drawing 

```
let mut ctx = get_ctx(256, 256, false);
let circle = Circle::new((128.0, 128.0), 128.0+1.0);
ctx.set_paint(LIME.into());
ctx.fill_path(&circle.to_path(0.1));
```

looks like 

![Image](https://github.com/user-attachments/assets/053c73ae-8b32-4b99-8b89-00112deaca5e)

It has looked like that since 4a75eced04cdd8cb188542d64bb6f18cc35b9821.

/cc @tomcur

PR #851, opened 2025-03-13
Fix sparse fills extending beyond the viewport

tomcur:
Fixes https://github.com/linebender/vello/issues/850.

A vestigial branch left in https://github.com/linebender/vello/pull/845 reset the `winding_delta` too early during strip rendering.


DJMcNab (approved):
Approving based entirely on tests passing.

PR #852, opened 2025-03-13
Pack `Tile` into 64 bits

tomcur:
(Split off from https://github.com/linebender/vello/pull/847.)

The fields of `Tile` are now packed such that its value in memory can readily be interpreted as a `u64` consisting, from most to least significant, of the numeric values of 16-bit `y`, 16-bit `x`, 1-bit `winding`, and 31-bit `line_index`. This means its value in memory can be used directly for sorting the tiles.

Sorting on a `u32` of only `y` and `x` is slightly faster for sorting (by about ~0.3 ms on paris-30k), but sorting the entire `u64` speeds up strip rendering (by about ~1.5ms on paris-30k). Presumably sorted line indices help cache locality during strip rendering, and ordering the winding bit before the line index also improves strip rendering performance, maybe because of branch predictability during strip rendering (see [this comment](https://github.com/linebender/vello/pull/847#discussion_r1990081159)).

On my machine, averaging over 500 iterations paris-30k on main and in this PR measure as:

```
Tile generation
    before: 17.329454ms
    after:  16.633348ms
Tile sorting
    before: 8.120621ms
    after:  7.3823915ms
Strip generation
    before: 31.896929ms
    after:  29.898085ms
```

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -15,23 +15,45 @@ pub const MAX_LINES_PER_PATH: u32 = 1 << 31;
 ///
 /// Keep in mind that it is possible to have multiple tiles with the same index,
 /// namely if we have multiple lines crossing the same 4x4 area!
-#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+///
+/// # Note
+///
+/// This struct is `#[repr(C)]`, but the byte order of its fields is dependent on the endianness of
+/// the compilation target.
+#[derive(Debug, Clone, Copy)]
 #[repr(C)]
-pub struct Tile(
-    /// The packed tile data.
+pub struct Tile {
+    // The field ordering is important.
+    //
+    // The given ordering (variant over little and big endian compilation targets), ensures that
+    // `Tile::to_bits` doesn't do any actual work, as the ordering of the fields is such that the
+    // numeric value of a `Tile` in memory is identical as returned by that method. This allows
+    // for, e.g., comparison and sorting.
+    #[cfg(target_endian = "big")]
+    /// The index of the tile in the y direction.
+    pub y: u16,
+
+    #[cfg(target_endian = "big")]
+    /// The index of the tile in the x direction.
+    pub x: u16,
+
+    /// The index of the line this tile belongs to into the line buffer, plus whether the line
+    /// crosses the top edge of the tile, packed together.
     ///
-    /// The layout is as follows, with the bit indices from least to most significant.
+    /// The index is the unsigned number in the 31 least significant bits of this value.
     ///
-    /// - `[ 0, 31)`: The 31-bit index of the line this tile belongs to into the line buffer;
-    /// - `[31, 32)`: The 1-bit coarse winding of the tile. This is `1` if and only if the lines
-    ///   crosses the tile's top edge. Lines making this crossing increment or decrement the coarse
-    ///   tile winding, depending on the line direction;
-    /// - `[32, 48)`: The 16-bit x-coordinate; and
-    /// - `[48, 64)`: The 16-bit y-coordinate.
-    ///
-    /// Note the byte layout in memory depends on the endianness of the compilation target.
-    pub u64,
-);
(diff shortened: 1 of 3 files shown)


LaurenzV:
> The order of components from most to least significant is 16-bit y and x, 1-bit winding, and 31-bit line_index. Sorting on a u32 of only y and x is slightly faster for sorting (by about ~0.3 ms on paris-30k), but sorting the entire u64 speeds up strip rendering (by about ~1.5ms on paris-30k). 

Have you also tried u32 but with everything packed, i.e. the original suggestion by Raph? Would at least be interesting to try to see whether there are any performance gains, or whether just sticking to u64 is the best.

tomcur:
> Have you also tried u32 but with everything packed, i.e. the original suggestion by Raph? Would at least be interesting to try to see whether there are any performance gains, or whether just sticking to u64 is the best.

Not yet, but I agree it would be interesting to test.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -5,24 +5,54 @@
...
+/// the compilation target.
 #[derive(Debug, Clone, Copy)]
+#[repr(C)]
 pub struct Tile {
+    // The field ordering is important.
+    //
+    // The given ordering (variant over little and big endian compilation targets), ensures that
+    // `Tile::to_bits` doesn't do any actual work, as the ordering of the fields is such that the
+    // numeric value of a `Tile` in memory is identical as returned by that method. This allows
+    // for, e.g., comparison and sorting.
```
LaurenzV, line 31: Have you measured whether that actually makes any performance difference?

LaurenzV (approved):

tomcur:
See https://godbolt.org/z/44ffv3v3G for codegen examples depending on field order.

sparse_strips/vello_common/src/tile.rs
tomcur: Also see https://godbolt.org/z/44ffv3v3G, which shows why the difference is so big.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -33,36 +63,91 @@ impl Tile {
...
+        (self.packed_winding_line_idx & MAX_LINES_PER_PATH) != 0
+    }
+
+    /// Return the `u64` representation of this tile.
+    ///
+    /// This is the u64 interpretation of `(y, x, packed_winding_line_idx)` where `y` is the
+    /// most-significant part of the number and `packed_winding_line_idx` the least significant.
+    #[inline(always)]
+    const fn to_bits(self) -> u64 {
+        ((self.y as u64) << 48) | ((self.x as u64) << 32) | self.packed_winding_line_idx as u64
```
DJMcNab, line 120: Could this use `bytemuck`? (we should probably have a test that this methods matches bytemuck, anyway)

waywardmonkeys: I'm not sure what that would gain here?

LaurenzV: I would also prefer avoiding adding additional dependencies for now, unless strictly necessary.

tomcur: It should indeed match `bytemuck`, and I'm not against bytemuck if it makes our lives easier. For CPU-rendering, the dep tree from `vello_cpu` upwards may never need the dependency.

DJMcNab: What we're asking the compiler to do is the conversion from `Tile` (which is 64 bits in memory) to the equivalent `u64` (which is the same 64 bits in memory). We're expressing that as a mathematical operation, and expecting the compiler to recognise that it is a no-op. But it seems to me that just doing the transmute directly would be a better indication of what we want.

But given that we don't already have bytemuck in tree here, I think this is *fine*

tomcur: Perhaps [safe transmutation](https://github.com/rust-lang/rfcs/blob/master/text/2835-project-safe-transmute.md) is getting somewhere. This is an interesting blog post about it: https://jack.wrenn.fyi/blog/safety-goggles-for-alchemists/.

If we do find another need for `bytemuck` in tree, I'd be happy for this to be bytemucked as well.

PR #853, opened 2025-03-13
vello_hybrid: Account for SVG group transforms

tomcur:
Not very important, as this code is likely to go away, but it makes testing a bit easier.

Paris-30k before:

before

Paris-30k after:

after

Ignore some of the rendering artifacts, those apparently only happen on some systems are unrelated to these changes.


DJMcNab (approved):
LGTM

PR #854, opened 2025-03-13
Vello Hybrid: Render into a user-provided RenderPass

DJMcNab:
Some of the follow-up from #831.

This strategy might not work as well if/when we start doing clipping and blending. Things get messy there; ultimately I *suspect* it's still right to do the "final" render into a user-provided pass, but we want to do intermediate renders ourselves? 
Nevertheless, for the current state this is the correct API; we definitely should not be doing our own queue submission, anyway.

I've run out of time for today, so have only run the `svg` example (which worked as expected). I expect the others to work, but I'd recommend checking that

### DIFF
--- a/sparse_strips/vello_hybrid/examples/simple.rs
+++ b/sparse_strips/vello_hybrid/examples/simple.rs
@@ -168,19 +168,16 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
                         occlusion_query_set: None,
                         timestamp_writes: None,
                     });
-                    self.renderers[surface.dev_id]
-                        .as_mut()
-                        .unwrap()
-                        .render_to_texture(
-                            &self.scene,
-                            &mut pass,
-                            &RenderParams {
-                                base_color: Some(palette::css::BLACK),
-                                width,
-                                height,
-                                strip_height: 4,
-                            },
-                        );
+                    self.renderers[surface.dev_id].as_mut().unwrap().render(
+                        &self.scene,
+                        &mut pass,
+                        &RenderParams {
+                            base_color: Some(palette::css::BLACK),
+                            width,
+                            height,
+                            strip_height: 4,
+                        },
+                    );
                 }
 
                 device_handle.queue.submit([encoder.finish()]);
--- a/sparse_strips/vello_hybrid/examples/simple.rs
+++ b/sparse_strips/vello_hybrid/examples/simple.rs
@@ -20,6 +20,7 @@ use vello_hybrid::{
     RenderParams, Renderer, Scene,
     util::{RenderContext, RenderSurface},
 };
+use wgpu::RenderPassDescriptor;
 use winit::{
     application::ApplicationHandler,
     event::WindowEvent,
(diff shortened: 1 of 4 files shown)


tomcur (commented):
```shell
cargo run --release --package vello_hybrid --example render_to_file ./examples/assets/Ghostscript_Tiger.svg ./target/tiger.png
```

Works for me, and leaves the issue mentioned on Zulip ([#gpu > Vello Hybrid @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/Vello.20Hybrid/near/505252172)) unchanged.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -325,48 +325,19 @@ impl Renderer {
     /// Render to a texture
     pub fn render_to_texture(
         &mut self,
-        device: &Device,
-        queue: &Queue,
         scene: &Scene,
-        texture_view: &TextureView,
-        render_params: &RenderParams,
+        render_pass: &mut RenderPass<'_>,
+        _render_params: &RenderParams,
```
grebmeg, line 330: If `render_params` isn’t being used, should we consider removing it for now?

sparse_strips/vello_hybrid/examples/simple.rs
```diff
@@ -141,41 +142,51 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
...
-                    &device_handle.device,
-                    &mut encoder,
-                    &surface.target_view,
-                    &surface_texture
-                        .texture
-                        .create_view(&wgpu::TextureViewDescriptor::default()),
-                );
+                {
+                    let mut pass = encoder.begin_render_pass(&RenderPassDescriptor {
+                        label: Some("Redner to Texture Pass"),
```
grebmeg, line 162:
```suggestion
                        label: Some("Render to Texture Pass"),
```

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -325,48 +325,19 @@ impl Renderer {
     /// Render to a texture
     pub fn render_to_texture(
```
grebmeg, line 326: Should we rename this method since it’s now decoupled from textures and tied to the render pass? Maybe just `render`?

grebmeg (approved):
Nice! The calling code now has full control over the command encoder, pass descriptor, and queue submission. I left a few minor comments — nothing major.

sparse_strips/vello_hybrid/src/render.rs
DJMcNab: I think so; looking at `fine`, the clear colour for the blend layers is transparent black:
https://github.com/linebender/vello/blob/fafba255bdf9584dae0e4b7b621e2a0cf2df7281/vello_shaders/shader/fine.wgsl#L977

 That means that if we render into a user-provided `RenderPass`, we never need to know the clear colour, so we can also remove that field.

DJMcNab:
The one new change is here is removing the `strip_height` `RenderParam`. It would always be set to 4, and isn't meaningful to set to a different value.

PR #855, opened 2025-03-14
Tidy up handling of truncation and lints in Vello Hybrid

DJMcNab:
The main new API is `DimensionConstraints::convert_dimension(value: f64) -> u16`, which is a panicking conversion from `f64` to `u16`.

This also changes `DimensionConstraints::calculate_dimensions` to always return a value within the range, even if it means that the aspect ratio is clamped.

### DIFF
--- a/sparse_strips/vello_hybrid/src/util.rs
+++ b/sparse_strips/vello_hybrid/src/util.rs
@@ -308,6 +308,11 @@ impl DimensionConstraints {
     }
 
     /// Calculate dimensions while preserving aspect ratio within constraints
+    ///
+    /// Some viewboxes could never fit inside this constraint. For example, if the constraint for both axes
+    /// is 100.0..=2000.0, if `original_width` is `2.` and `original_height` is `1000.`, there is clearly
+    /// no way for that to fit within the constraints.
+    /// In these cases, this method clamps to within the ranges (respecting the constraints but losing the aspect ratio).
     pub fn calculate_dimensions(&self, original_width: f64, original_height: f64) -> (f64, f64) {
         // Ensure we have non-zero input dimensions
         let original_width = original_width.max(1.0);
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_hybrid/src/util.rs
```diff
@@ -319,29 +318,50 @@ impl DimensionConstraints {
...
+        (
+            width.clamp(min_width, max_width),
+            height.clamp(min_height, max_height),
+        )
+    }
+
+    /// Converts a floating point dimension to a u16.
+    /// This can be used for values returned from [`calculate_dimensions`](Self::calculate_dimensions).
+    /// Note that:
+    /// 1) If the
```
tomcur: Fragment

sparse_strips/vello_hybrid/src/util.rs
```diff
@@ -319,29 +318,50 @@ impl DimensionConstraints {
...
-            )
+            ((original_width * ratio), (original_height * ratio))
         } else {
             (original_width, original_height)
-        }
+        };
+        // Some viewboxes could never fit inside the given ranges (for example, for a constraint on both axes
+        // of 100.0..=2000.0), an `original_width` of `2.` with an `original_height` of `1000.`, there is clearly
+        // no way to fit that within the constraints.
+        // For these cases, we just clamp. Note that in theory.
```
tomcur, line 341: Fragment

tomcur (approved):
This looks reasonable.

One alternative is to take `u16`s in `DimensionConstraints::new` directly or otherwise only allowing a `DimensionConstraints` to be constructed with a range that can unconditionally be cast to `u16` without truncation. That would reduce the number of runtime checks required when using the constructed `DimensionConstraints`.

sparse_strips/vello_hybrid/src/util.rs
tomcur: It might be worth having this example in the method docstring.

PR #857, opened 2025-03-14
Vello Hybrid: Tidy up unused things from util

DJMcNab:
Note that I want to rethink a good amount of this util module anyway. It's easier to do that if there isn't a lot of cruft in the way.

These are all things which were needed in the old Vello, but aren't needed in Vello Hybrid


tomcur (approved):
I haven't tried running the examples.

PR #858, opened 2025-03-16
sparse_strips: Depend on kurbo/peniko via vello_api

waywardmonkeys:
The `vello_api` dependency already provides a re-export of `kurbo` and `peniko`, so use them via that or the re-export of that in `vello_common`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2698,8 +2698,6 @@ dependencies = [
 name = "vello_common"
 version = "0.4.0"
 dependencies = [
- "kurbo",
- "peniko",
  "roxmltree",
  "vello_api",
 ]
@@ -2709,9 +2707,7 @@ name = "vello_cpu"
 version = "0.4.0"
 dependencies = [
  "image",
- "kurbo",
  "oxipng",
- "peniko",
  "vello_common",
 ]
 
@@ -2731,8 +2727,6 @@ name = "vello_hybrid"
 version = "0.4.0"
 dependencies = [
  "bytemuck",
- "kurbo",
- "peniko",
  "png",
  "pollster",
  "roxmltree",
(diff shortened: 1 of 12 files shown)


LaurenzV:
Iirc I did that because @tomcur preferred depending on them that way to simplify the imports.

waywardmonkeys:
This change is laying the groundwork to reduce some configuration duplication when I submit a follow up PR that does some `no_std` work.

nicoburns:
I don't really work on Vello much. But I'm going to put my support behind direct imports of `peniko`/`kurbo`. It's very awkward to import types through a re-export (esp. as there are often several re-exports to choose from). And it also doesn't play very nicely with Rust Analyzer auto imports.

waywardmonkeys:
> I don't really work on Vello much. But I'm going to put my support behind direct imports of `peniko`/`kurbo`. It's awkward to import types through a re-export, both when writing the code and when reading it (esp. as there are often several re-exports to choose from). And it also doesn't play very nicely with Rust Analyzer auto imports.

Why do we have to have this same discussion every single time? If we're going to use direct imports, then remove the re-exports. There's no issue with Rust Analyzer auto-imports that I know about (but I don't use RA) when there is one primary path to the stuff, which is the case when it is only coming in via the `vello_api` or `vello_common` re-export.

nicoburns:
> Why do we have to have this same discussion every single time?

I guess because we never agreed on a solution. But I will concede and butt out here to avoid further bikeshedding.

tomcur (approved):
> Iirc I did that because @tomcur preferred depending on them that way to simplify the imports.

> This change is laying the groundwork to reduce some configuration duplication when I submit a follow up PR that does some no_std work.

I indeed suggested going with direct dependencies, mostly following current-`vello` (e.g. see `peniko` in [`vello_encoding/Cargo.toml`](https://github.com/linebender/vello/blob/5f4126245e94a828095521d399c16df57fcea206/vello_encoding/Cargo.toml) and [`vello/Cargo.toml`](https://github.com/linebender/vello/blob/main/vello/Cargo.toml)).

I'm easily convinced going via re-exports makes sense, especially with feature-dependent dependency features.

PR #860, opened 2025-03-17
Vello Hybrid: Move `RenderContext` into `examples/common`

DJMcNab:
Exposing `RenderContext` publicly from Vello is one of the greatest mistakes, as it's a significant layering break. However, it does simplify the multi-GPU code we have for our examples significantly, so removing it entirely is quite an involved process. For now, we can move it to `examples/common` in Vello Hybrid.

(I tend to think that our "simple" example is an anti-pattern; it's not our job to teach people how to set up wgpu, and getting that entire integration exactly right is *really hard*; to call an example "simple" is the same as calling it "wrong". There absolutely is a role for "idiomatic wgpu + winit integration", but trying to fit it in a single file whilst also teaching our API is not really conducive to that)


waywardmonkeys (approved):
Someone else should also review this, but I'm very in favor of this change.

DJMcNab:
> You mentioned that it’s not our responsibility to teach people how to set up wgpu, but I think it’s valuable to have a somewhat comprehensive integration in our examples — both to reflect our understanding of how it should be used and to allow others to spot potential issues and provide feedback.

Basically, my issue is with the name "simple"; the reality of such an example is that (say) 80% of the code is just "generic winit+wgpu boilerplate". It would be valuable for that to be taught somewhere, but our examples should be focused on our speciality, which is our own API/integration with Vello. Ideally, we'd contribute such a comprehensive integration to [Learn WGPU](https://sotrh.github.io/learn-wgpu/). But getting it right is hard:
1) It needs to properly handle web (truly async handling of resource creation, etc.)
2) It should handle multi-GPU well; ideally also handling multi-window
3) It should support running on Android (although I'm happy to leave that on hold until we see how https://github.com/mwcampbell/android-view shakes up)

Incidentally, I have for a while wanted to create a "Vidi" crate - see [#glazier > Vidi](https://xi.zulipchat.com/#narrow/channel/351333-glazier/topic/Vidi). The exact shape would probably not be as proposed in that thread, but it would try and provide what you need for an idiomatic integration between Vello and Winit. That being said, we're pretty far from that.

> Regarding the “simple” example — are you suggesting we remove it? Or perhaps rename the “svg” example to something more fitting?

My gut reaction is to remove the current "simple" example, maybe rename `render_to_file` to "simple" and fill it with comments. I'm not sure what we should do with the `svg` example; I don't want it to end up as monstrous as `with_winit`.

PR #862, opened 2025-03-18
`vello_api`: Make `no_std` and add CI

waywardmonkeys:
This makes the workspace dependencies for `kurbo` and `peniko` not enable default features, and therefore have to enable them in the crates that use them.

Add CI based on other Linebender crates, except that there is now a variable to specifically control which targets are checked for `no_std` compatibility.

This also means that `no_std` crates are checked for MSRV compliance which was not currently happening for the sparse strips crates.

Other crates will get `no_std` support in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,11 @@ env:
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
   RUST_MIN_VER_PKGS: "-p vello -p vello_encoding -p vello_shaders -p vello_api -p vello_common -p vello_cpu -p vello_hybrid"
+  # List of packages that will be checked for `no_std` builds.
+  # This should be limited to packages that are intended for publishing.
+  RUST_NO_STD_PKGS: "-p vello_api"
+  # List of features that depend on the standard library and will be excluded from no_std checks.
+  FEATURES_DEPENDING_ON_STD: "std,default"
   # List of packages that can not target Wasm.
   # `vello_tests` uses `nv-flip`, which doesn't support Wasm.
   NO_WASM_PKGS: "--exclude vello_tests --exclude simple_sdl2"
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -12,6 +12,11 @@ env:
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
   RUST_MIN_VER_PKGS: "-p vello -p vello_encoding -p vello_shaders"
```
DJMcNab, line 14: I think these not having MSRV checked was an oversight (not that it matters when MSRV is MaxSRV, of course)

That is, we should add the sparse strips packages here.

sparse_strips/vello_api/Cargo.toml
```diff
@@ -15,7 +15,10 @@ publish = false
 peniko = { workspace = true }
 
 [features]
+default = ["std"]
```
DJMcNab, line 18: Probably worth adding some basic docs here (i.e. a comment above the libm/std features)

I'd probably also lean towards grouping `libm` and `std`.

DJMcNab (approved):
LGTM. We may as well fix the MSRV thing you noticed whilst we're here

tomcur (approved):
I'm in favor of climbing this ladder. Having `alloc`-only `vello_common` and perhaps `vello_cpu` would be nice. The main thing lost without `std` is threading. Hmm... and `core` doesn't have CPU feature detection yet.

.github/workflows/ci.yml
waywardmonkeys: Done in PR #863.

waywardmonkeys:
> I'm in favor of climbing this ladder. Having `alloc`-only `vello_common` and perhaps `vello_cpu` would be nice. The main thing lost without `std` is threading. Hmm... and `core` doesn't have CPU feature detection yet.

`alloc`-only `vello_common` will need `pico_svg` to go away. I have a branch locally that does most of the work already.

xorgy (approved):

PR #864, opened 2025-03-18
Panic instead of silently failing in `vello_hybrid`

taj-p:
Small changes in `vello_hybrid` to ensure that panics are called when invariants are not obeyed rather than silently failing.

Implementation like the below means that incorrect use of the API results in no feedback to the user. We should panic in such cases because the user has not followed our architectural invariant of `prepare` being called before `render`.

```rs
fn render (...) {
        let Some(resources) = &self.resources else {
            return;
        };
       ....
```

---

I think we could explore returning some struct from `prepare` that can be passed to `render` to ensure that prepare must be called before render, but I leave that as a separate exercise.

---

Tip: Review with whitespace hidden:

image

### DIFF
--- a/sparse_strips/vello_hybrid/src/render.rs
+++ b/sparse_strips/vello_hybrid/src/render.rs
@@ -363,7 +363,10 @@ impl Renderer {
     ) {
         // TODO: Consider API that forces the user to call `prepare` before `render`.
         // For example, `prepare` could return some struct that is consumed by `render`.
-        let resources = &self.resources.as_ref().expect("`prepare` should be called before `render`");
+        let resources = &self
+            .resources
+            .as_ref()
+            .expect("`prepare` should be called before `render`");
         let render_data = scene.prepare_render_data();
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -309,47 +308,46 @@ impl Renderer {
...
-            );
-        }
+        // Prepare alpha data for the texture with 4 alpha values per texel
+        let texture_width = resources.alphas_texture.width();
+        let texture_height = resources.alphas_texture.height();
+        assert!(
+            render_data.alphas.len() <= (texture_width * texture_height * 4) as usize,
+            "Alpha texture dimensions are too small to fit the alpha data"
+        );
+        let mut alpha_data = render_data.alphas.clone();
```
grebmeg, line 328: Ah, yes, of course! Thanks for the correction!

grebmeg (approved):

PR #866, opened 2025-03-19
Store alphas as `u8` instead of explicitly packing

tomcur:
This stops packing alpha values explicitly into `u32`s, which was a leftover from using `u32` GPU alpha storage buffers. This simplifies the code a bit and is a performance improvement. On my machine, 500 iterations of Paris-30k measure as the following.

```
Strip rendering
    before: 29.770845ms
    after:  28.914503ms
CPU fine
    before: 13.367204ms
    after:  12.358518ms
```

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -15,8 +15,8 @@ pub struct Strip {
     pub x: u16,
     /// The y coordinate of the strip, in user coordinates.
     pub y: u16,
-    /// The index into the alpha buffer
-    pub col: u32,
+    /// The index into the alpha buffer.
+    pub alpha_idx: u32,
     /// The winding number at the start of the strip.
     pub winding: i32,
 }
@@ -33,7 +33,7 @@ impl Strip {
 pub fn render(
     tiles: &Tiles,
     strip_buf: &mut Vec<Strip>,
-    alpha_buf: &mut Vec<u32>,
+    alpha_buf: &mut Vec<u8>,
     fill_rule: Fill,
     lines: &[Line],
 ) {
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_common/src/strip.rs
```diff
@@ -83,23 +83,17 @@ pub fn render(
...
                         }
-
-                        alpha_buf.push(alphas);
                     }
                 };
             }
             match fill_rule {
                 Fill::NonZero => {
-                    fill!(|area: f32| area.abs().min(1.0))
+                    fill!(|area: f32| area.abs())
```
tomcur, line 96: The winding is now cast directly to `u8`, which saturates, so this `min` is unnecessary. Dropping this reduces strip rendering time by about `0.5ms` by itself.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -118,21 +112,21 @@ pub fn render(
         if !prev_tile.same_loc(&tile) && !prev_tile.prev_loc(&tile) {
             debug_assert_eq!(
                 (prev_tile.x + 1) * Tile::WIDTH - strip.x,
-                (alpha_buf.len() - strip.col as usize) as u16,
+                ((alpha_buf.len() - strip.alpha_idx as usize) / usize::from(Tile::HEIGHT)) as u16,
```
LaurenzV, line 115: Not sure if that makes sense, but could we maybe turn alpha_idx into a very thin wrapper around u32 that provides a col() method that automatically calculates the division? So that we don't have to repeat it everywhere?

LaurenzV (approved):

sparse_strips/vello_common/src/strip.rs
tomcur: Once `Tile` is `const`-generic over size, that wrapper would also have to be generic, which is not ideal. It could be a method on `Strip`, as that would have to be generic anyway, but I'm not completely convinced that it would help legibility much. Perhaps `alpha_idx` should just become `col` again, doing the multiplication by `Tile::HEIGHT` elsewhere to get the actual index.

I understand your concern... I'm not sure what would be easiest for consumers of `vello_common`.

PR #870, opened 2025-03-20
`vello_hybrid`: Fix pixel coord -> NDC transform

tomcur:
I believe the coordinate systems are aligned at the pixel edge, not the pixel center. This fixes rendering artifacts on my system (AMD GPU), where with the previous shader code the locations sampled during fragment shading probably sometimes got bumped to the next pixels.

This should be confirmed to work on systems where it worked previously as well.

### DIFF
--- a/sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
+++ b/sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
@@ -69,8 +69,8 @@ fn vs_main(
     let pix_y = f32(y0) + y * f32(config.strip_height);
     // Convert pixel coordinates to normalized device coordinates (NDC)
     // NDC ranges from -1 to 1, with (0,0) at the center of the viewport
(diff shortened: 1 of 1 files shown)


tomcur:
```shell
cargo run --release --package vello_hybrid --example render_to_file ./examples/assets/Ghostscript_Tiger.svg ./target/tiger.png
```

# Before
tiger_before

# After
tiger

DJMcNab (approved):
I can't approve from first principles, but it doesn't seem to be a regression on my machine.

The relevant reading is probably: https://gpuweb.github.io/gpuweb/#abstract-opdef-rasterize

tomcur:
> I can't approve from first principles, but it doesn't seem to be a regression on my machine.
> 
> The relevant reading is probably: https://gpuweb.github.io/gpuweb/#abstract-opdef-rasterize

The equation for framebuffer coordinates given NDC position and viewport offset and size agree with the new equations in this PR, given zero viewport offset.

grebmeg (approved):
Awesome! This also fixes the issue on the Tiger’s whisker in the screenshot below — thank you!

Before | After
|---|---|
image|image

PR #872, opened 2025-03-20
[Vello Hybrid]: WebGL2 Compatibility

taj-p:
# Updates to Support WebGL2

In order to support WebGL2 downlevel compatibility, we need to:

- Use 16 byte aligned uniforms
- Allow passing surface formats to the renderer (`Bgra8Unorm` is not supported)
- Use downlevel compatible WGSL (polyfill `unpack4x8unorm` and not use `firstTrailingBit`)

---

Also, adds in CI and the below test crate:

## WebGL Demo

Uses Vello Hybrid with a WebGL2 backend in the browser - this is to ensure Vello Hybrid remains WebGL2 compliant.

## Development

Run with `cargo run_wasm -p webgl`.

## Testing

In order to test this crate, you need to have [`wasm-pack`] installed. Install it using
the steps found in https://rustwasm.github.io/wasm-pack/installer/.

Thereafter, for interactive test sessions, run:

```
wasm-pack test --chrome
# Navigate to printed URL
```

[`wasm-pack`]: https://rustwasm.github.io/wasm-pack/

## Example of success

image

## Example of error

image


sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -101,11 +104,9 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
         let texel_index = alphas_index / 4u;
         // Which channel (R,G,B,A) in the texel
         let channel_index = alphas_index % 4u;
-        // Calculate texture coordinates using bitwise operations
-        // This is more efficient than using modulo and division when width is a power of 2
-        let alphas_tex_width_bits = firstTrailingBit(alphas_tex_width);
```
taj-p, line 106: WebGL2 doesn't support `firstTrailingBit` -> Need to precompute

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -129,3 +130,16 @@ fn unpack_alphas_from_channel(rgba: vec4<u32>, channel_index: u32) -> u32 {
         default: { return rgba.x; }
     }
 }
+
+// Polyfills `unpack4x8unorm`.
+//
+// Downlevel targets do not support native WGSL `unpack4x8unorm`.
+fn unpack4x8unorm(rgba_packed: u32) -> vec4<f32> {
```
taj-p, line 137: WebGL2 doesn't support `unpack4x8unorm`. Needs a polyfill.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -115,7 +116,7 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
         // Extract the alpha value for the current y-position from the packed u32 data
         alpha = f32((alphas_u32 >> (y * 8u)) & 0xffu) * (1.0 / 255.0);
     }
-    // Apply the alpha value to the unpacked RGBA color
+    // Apply the alpha value to the manually unpacked RGBA color
```
taj-p, line 119:
```suggestion
    // Apply the alpha value to the unpacked RGBA color
```

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -113,9 +119,7 @@ impl Renderer {
     /// Creates a new renderer
     ///
     /// The target parameter determines if we render to a window or headless
-    pub fn new(device: &Device, _options: &RendererOptions) -> Self {
-        let format = wgpu::TextureFormat::Bgra8Unorm;
```
taj-p, line 117: WebGL2 doesn't support `Bgra8Unorm` texture formats. I think it's better to allow consumers to decide target format

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -17,6 +17,9 @@ struct Config {
     height: u32,
     // Height of a strip in the rendering
     strip_height: u32,
+    // Number of trailing zeros in alphas_tex_width (log2 of width).
+    // Pre-calculated on CPU since WebGL2 doesn't support firstTrailingBit.
+    alphas_tex_width_bits: u32,
```
taj-p, line 22: Beyond the issue of WebGL2 not supporting `firstTrailingBit`. Uniforms in WebGL2 need to be 16 byte aligned.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -17,6 +17,9 @@ struct Config {
     height: u32,
     // Height of a strip in the rendering
     strip_height: u32,
+    // Number of trailing zeros in alphas_tex_width (log2 of width).
+    // Pre-calculated on CPU since WebGL2 doesn't support firstTrailingBit.
```
taj-p, line 21:
```suggestion
    // Pre-calculated on CPU since WebGL2 doesn't support `firstTrailingBit`.
```

waywardmonkeys (commented):
Is there a way to test this (especially in CI) since this is typically the sort of thing that is rather fragile and easy to break?

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -129,3 +130,16 @@ fn unpack_alphas_from_channel(rgba: vec4<u32>, channel_index: u32) -> u32 {
...
+// Downlevel targets do not support native WGSL `unpack4x8unorm`.
+fn unpack4x8unorm(rgba_packed: u32) -> vec4<f32> {
+    // Extract each byte and convert to float in range [0,1]
+    return vec4<f32>(
+        f32((rgba_packed >> 0u) & 0xFFu) / 255.0,  // r
+        f32((rgba_packed >> 8u) & 0xFFu) / 255.0,  // g
+        f32((rgba_packed >> 16u) & 0xFFu) / 255.0, // b
+        f32((rgba_packed >> 24u) & 0xFFu) / 255.0  // a
+    );
+}
```
grebmeg, line 145: [nit]: trailing newline missing

grebmeg (approved):

taj-p:
> Is there a way to test this (especially in CI) since this is typically the sort of thing that is rather fragile and easy to break?

Good idea @waywardmonkeys - I'll time box adding a WebGL example to build and run in CI. I'll reserve time on Monday for this.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
DJMcNab: What's the failure mode here? Is it an error from wgpu, or does the webgl implementation give an error? If the webgl implementation gives an error, I think the wgpu team would like to know this.

If wgpu is erroring, ideally, we would tell wgpu to run in a fully webgl2 compatible mode, cross-platform. But that's not to be dealt with in this PR.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
DJMcNab: (You should be able to configure your editor to add these automatically)

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
DJMcNab: Does this have a measurable performance cost cross-platform? I don't know if this is used on a hot path

DJMcNab (commented):
This seems reasonable. I agree with the concern that it seems impractical to not regress this with:
1) No extant CI tests
2) No extant runnable on web version outside of CI

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
taj-p: > If wgpu is erroring, ideally, we would tell wgpu to run in a fully webgl2 compatible mode, cross-platform. But that's not to be dealt with in this PR.

WGPU is erroring - Yes, totally agree!

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
taj-p: My IDE didn't like `.wgsl` 😅

taj-p:
> This seems reasonable. I agree with the concern that it seems impractical to not regress this with:
> 
> 1. No extant CI tests
> 2. No extant runnable on web version outside of CI

Yup - agree and good call! Added in https://github.com/linebender/vello/pull/872/commits/289bf9065f81a05f23fb8405ef3724b800de14ce.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
taj-p: > Does this have a measurable performance cost cross-platform?

Probably measurable, but likely small. We don't have benchmarks yet for Vello Hybrid, so it's hard to say.

This is run once per pixel for each wide tile command. So, it could be considered a hot path. 

We could consider preprocessing the shaders to allow directives like `#ifdef webgl /* add polyfill */ #endif`, but I think that's out of scope at this stage and better to be looked at in the future (?).

waywardmonkeys: Might be worth filing a tracking issue for this on our end. Also might be worth bringing this up with upstream wgpu.

taj-p: Good idea!

I created an issue (https://github.com/linebender/vello/issues/876). 

---

As per WGPU's README, WebGL support is "best effort" and I imagine an API like `unpack4x8unorm` which is available in OpenGL 4 but not 3 (which WebGL2 uses) could fall into the bucket of "best effort" but not full support. 

I suspect WGPU would need to support a `BROWSER_WEBGL` backend like its [`BROWSER_WEBGPU` backend](https://github.com/gfx-rs/wgpu/blob/ccaf9a84f8118f4a991e0046dbaecef4ea1c6950/wgpu-types/src/lib.rs#L249), which seems onerous.

In any case, I'll investigate existing WGPU WebGL issues and WGPU code separately later this week.

.github/workflows/ci.yml
```diff
@@ -295,6 +295,14 @@ jobs:
       - name: cargo test compile
```
DJMcNab, line 295: There's a comment a line above which we should delete.

.github/workflows/ci.yml
```diff
@@ -295,6 +295,14 @@ jobs:
...
         run: cargo test --workspace ${{ env.NO_WASM_PKGS }} --locked --target wasm32-unknown-unknown --all-features --no-run
 
+      - name: install wasm-pack
+        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```
DJMcNab, line 299: We can use the install-action we already use for this (see https://github.com/taiki-e/install-action/blob/main/TOOLS.md)

sparse_strips/vello_hybrid/examples/webgl/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+}
+
+/// Creates a `HTMLCanvasElement` of the given dimensions and renders the given `Scene` into it.
+#[cfg(target_arch = "wasm32")]
+pub async fn render_scene(scene: vello_hybrid::Scene, width: u16, height: u16) {
+    console_error_panic_hook::set_once();
+    console_log::init_with_level(log::Level::Debug).unwrap();
+    let canvas = web_sys::Window::document(&web_sys::window().unwrap())
```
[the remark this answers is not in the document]
DJMcNab: It would be nice to handle resizing to the full screen size; winit does handle that, but you can also use ResizeObserver manually (?)

sparse_strips/vello_hybrid/examples/webgl/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        });
+        renderer.render(&scene, &mut pass, &params);
+    }
+
+    queue.submit([encoder.finish()]);
+    surface_texture.present();
+}
+
+/// Draws a blue triangle into the given `Scene`.
+pub fn draw_triangle(ctx: &mut vello_hybrid::Scene) {
```
DJMcNab, line 155: It might be nice to embed the GhostScript tiger SVG into the WASM binary, and display that. I don't think it would require significantly more code?

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
DJMcNab: I *think* that's not a wgpu error. That is, it's wgpu just forwarding the failure generated by your browser. It looks very similar to https://github.com/gfx-rs/wgpu/issues/4525

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
DJMcNab: See https://github.com/gfx-rs/wgpu/blob/ccaf9a84f8118f4a991e0046dbaecef4ea1c6950/wgpu-types/src/lib.rs#L147-L148
That is, it's expected to support WebGL

DJMcNab (approved):
I think this is landable

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
taj-p: Ah yes - you're right!

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
taj-p: I guess! The readme conflicts with that expectation, but I think you've found the culprit here: https://github.com/gfx-rs/wgpu/issues/4525 🙏

.github/workflows/ci.yml
taj-p: Deleted in https://github.com/linebender/vello/pull/872/commits/dea6c9a761ac3bec2fbed2525eab8c2f69d89b81

sparse_strips/vello_hybrid/examples/webgl/src/lib.rs
taj-p: Added ghosttiger in https://github.com/linebender/vello/pull/872/commits/03c2ef38ae4b82a3df9177b70b471beae861a0a2

image

sparse_strips/vello_hybrid/examples/webgl/src/lib.rs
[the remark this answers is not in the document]
taj-p: I updated the code to initially set the canvas to full screen size. I timeboxed making it dynamically resize with `window.onresize`, but couldn't complete in the time. I'll make a note to come back to this another time.

PR #873, opened 2025-03-21
Remove tile height restriction from `vello_common`, add to `vello_hybrid`

tomcur:
With the changes to alpha mask packing in https://github.com/linebender/vello/pull/866 the scalar code in `vello_common` and `vello_cpu` now works for arbitrary tile sizes.

Most tile size configurations I've checked pass the tests, some tile heights (e.g., 7) cause rounding differences in the rasterized images.

`vello_hybrid`'s shader code assumes a tile height of 4, so a check has been introduced in that crate.


LaurenzV (approved):
Cool, at some point in the future we should probably at unit tests for tiling at different sizes, but only once the code has more or less stabilized.

grebmeg (approved):
> vello_hybrid's shader code assumes a tile height of 4, so a check has been introduced in that crate.

I haven’t looked into how to remove the restriction in `vello_hybrid`, but could it be fixed by simply using `strip_height` [here](https://github.com/grebmeg/vello/blob/bd74626471ace62d7365f310e8c22dee788df50f/sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl#L103) instead of `4u`?

tomcur:
> I haven’t looked into how to remove the restriction in `vello_hybrid`, but could it be fixed by simply using `strip_height` [here](https://github.com/grebmeg/vello/blob/bd74626471ace62d7365f310e8c22dee788df50f/sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl#L103) instead of `4u`?

Unfortunately not, currently alpha values for columns fit exactly into individual texel color channels (each 32-bit color channel coding for 4 alpha values), and the `y`-component is used only for indexing within a selected color channel ([here](https://github.com/grebmeg/vello/blob/bd74626471ace62d7365f310e8c22dee788df50f/sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl#L118)). If the strip height is any other value, alpha values for columns will no longer be aligned exactly to channels, and the `x` and `y`-components both need to be taken into account.

PR #874, opened 2025-03-21
Add support for indexed paints

LaurenzV:
See [#vello > Sparse Strips: Generic Paint Commands](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Sparse.20Strips.3A.20Generic.20Paint.20Commands/with/507715909)

### DIFF
--- a/sparse_strips/vello_api/src/paint.rs
+++ b/sparse_strips/vello_api/src/paint.rs
@@ -5,8 +5,6 @@
 
 use peniko::color::{AlphaColor, PremulRgba8, Srgb};
 
-pub(crate) const MAX_INDEX: usize = u32::MAX as usize;
-
 /// A paint that needs to be resolved via its index.
 // In the future, we might add additional flags, that's why we have
 // this thin wrapper around u32, so we can change the underlying
@@ -17,7 +15,10 @@ pub struct IndexedPaint(u32);
 impl IndexedPaint {
     /// Create a new indexed paint from an index.
     pub fn new(index: usize) -> Self {
-        assert!(index <= MAX_INDEX, "exceeded the maximum number of paints");
+        assert!(
+            index <= u32::MAX as usize,
+            "exceeded the maximum number of paints"
+        );
 
         #[allow(
             clippy::cast_possible_truncation,
--- a/sparse_strips/vello_api/src/paint.rs
+++ b/sparse_strips/vello_api/src/paint.rs
@@ -3,7 +3,7 @@
 
 //! Types for paints.
 
-use peniko::color::{AlphaColor, Srgb, PremulRgba8};
+use peniko::color::{AlphaColor, PremulRgba8, Srgb};
 
 pub(crate) const MAX_INDEX: usize = u32::MAX as usize;
 
@@ -19,10 +19,13 @@ impl IndexedPaint {
     pub fn new(index: usize) -> Self {
         assert!(index <= MAX_INDEX, "exceeded the maximum number of paints");
 
-        #[allow(clippy::cast_possible_truncation, reason = "Guaranteed to be in-bounds")]
+        #[allow(
+            clippy::cast_possible_truncation,
+            reason = "Guaranteed to be in-bounds"
+        )]
         Self(index as u32)
     }
-    
+
     /// Return the index of the paint.
     pub fn index(&self) -> usize {
         self.0 as usize
@@ -35,14 +38,14 @@ pub enum Paint {
     /// A premultiplied RGBA8 color.
     Solid(PremulRgba8),
     /// An paint that needs to be resolved via an index.
-    Indexed(IndexedPaint)
+    Indexed(IndexedPaint),
 }
 
 impl From<AlphaColor<Srgb>> for Paint {
     fn from(value: AlphaColor<Srgb>) -> Self {
         // TODO: This might be slow on x86, see https://github.com/linebender/color/issues/142.
         // Since we only do that conversion once per path it might not be critical, but should
-        // still be measured. This also applies to all other usages of `to_rgba8` in the current 
+        // still be measured. This also applies to all other usages of `to_rgba8` in the current
         // code.
         Self::Solid(value.premultiply().to_rgba8())
     }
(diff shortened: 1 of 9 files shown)


sparse_strips/vello_api/src/paint.rs
```diff
@@ -3,24 +3,51 @@
...
+        assert!(
+            index <= u32::MAX as usize,
+            "exceeded the maximum number of paints"
+        );
+
+        #[allow(
+            clippy::cast_possible_truncation,
+            reason = "Guaranteed to be in-bounds"
+        )]
+        Self(index as u32)
```
DJMcNab, line 27: Nit: We should be able to use https://doc.rust-lang.org/stable/std/primitive.u32.html#impl-TryFrom%3Cusize%3E-for-u32 here

sparse_strips/vello_api/src/paint.rs
```diff
@@ -3,24 +3,51 @@
...
+    }
+
+    /// Return the index of the paint.
+    pub fn index(&self) -> usize {
+        self.0 as usize
```
DJMcNab, line 32: This is robust, because the value definitely already fit in a usize due to the constructor (but this might cause `cast_possible_truncation` on 16 bit platforms)

This could also use TryFrom, which should optimise to the same, except be sanity checked on 16 bit platforms.

DJMcNab (commented):
This looks reasonable to me. Not approving as I've not read through carefully

sparse_strips/vello_api/src/paint.rs
LaurenzV: So you mean `try_from` and then `unwrap`?

DJMcNab: Yeah

sparse_strips/vello_api/src/paint.rs
```diff
@@ -3,24 +3,42 @@
...
-    /// A solid color.
-    Solid(AlphaColor<Srgb>),
-    /// A gradient.
-    Gradient(()),
-    /// A pattern.
-    Pattern(()),
+    /// A premultiplied RGBA8 color.
+    Solid(PremulRgba8),
+    /// A paint that needs to be resolved via an index.
+    Indexed(IndexedPaint),
```
grebmeg, line 33: [This point](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Sparse.20Strips.3A.20Generic.20Paint.20Commands/near/507637923) made by Bruce seems entirely reasonable:

> Once wider color gamuts and stuff come into play, you won’t always want `PremulRgba8`

However, I understand your intent to avoid performing a conversion for each command, and I suspect that `IndexedPaint` might be able to handle this issue on its own.

Since we’re only using `PremulRgba8` in the renderer atm, I think it makes sense to keep the code simple for now and revisit this once we actually need `AlphaColor<CS>` — especially with achieved performance in mind. I think, avoiding premature optimization at this stage helps keep things more maintainable and focused.

That said, I’m okay with landing this PR as is — but that’s just my perspective. I’d appreciate if someone else could also take a look, just to make sure we’re aligned before merging.

LaurenzV: Perhaps we can just change it so that all paints are indexed in the future, but we would need to measure whether it has a performance impact.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -216,23 +217,23 @@ impl Scene {
                 for cmd in &wide_tile.cmds {
                     match cmd {
                         vello_common::coarse::Cmd::Fill(fill) => {
-                            let color: AlphaColor<Srgb> = match fill.paint {
+                            let color: PremulRgba8 = match fill.paint {
                                 Paint::Solid(color) => color,
-                                _ => AlphaColor::TRANSPARENT,
+                                _ => AlphaColor::<Srgb>::TRANSPARENT.premultiply().to_rgba8(),
```
waywardmonkeys, line 222: Since this enum only has 2 variants now, I think maybe we should be explicit here (and below) to help make it easier to find the uses of the indexed paint for updating them in the future by using `Paint::Indexed(_)` rather than `_`.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -216,23 +217,23 @@ impl Scene {
...
-                                rgba: color.premultiply().to_rgba8().to_u32(),
+                                rgba: color.to_u32(),
                             });
                         }
                         vello_common::coarse::Cmd::AlphaFill(cmd_strip) => {
-                            let color: AlphaColor<Srgb> = match cmd_strip.paint {
+                            let color: PremulRgba8 = match cmd_strip.paint {
                                 Paint::Solid(color) => color,
-                                _ => AlphaColor::TRANSPARENT,
+                                _ => AlphaColor::<Srgb>::TRANSPARENT.premultiply().to_rgba8(),
```
waywardmonkeys, line 236: Same as a comment above.

sparse_strips/vello_hybrid/src/scene.rs
waywardmonkeys: Or even having them be `Paint::Indexed(_) => todo!(),`

sparse_strips/vello_api/src/paint.rs
tomcur: There is a related point about the color space used for compositing. We currently composite using a linear blend in non-linear sRGB, but that may be "less correct" than compositing in a linear color space (see e.g. this post related to color blending and glyphs https://skia.org/docs/dev/design/raster_tragedy/, but this is also an interesting post, though about 3D: https://www.colour-science.org/anders-langlands/).

We may want to let users choose the destination/blend color space.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Seems like clippy doesn't like having todo, so I will change it back to unimplemented.

DJMcNab: The reason for that clippy lint is to enable this workflow:
1) `todo!` is for things which you expect to fix in this PR
2) `unimplemented!` is for things you expect to come later

And clippy enforcing the first of those is quite nice.

LaurenzV: Okay, then I will change it back.

DJMcNab (approved):
I suspect we'll need to think about this more carefully in the future, but for the moment getting this landed can allow that collaboration on main.

PR #877, opened 2025-03-25
Follow-up fix for encoding data as `u32`

### LINKED ISSUE #875: vello_encoding panics while loading CIA World Map in debug build with_winit
When I run with_winit, it panics due to an arithmetic overflow in `vello_encoding/src/path.rs` while encoding the scene.

waywardmonkeys:
The original work was done in PR #817 and as noticed in issue #875, there was still a place where there was a length manipulation that was in terms of `u8` rather than `u32`.

Fixes #875.

### DIFF
--- a/vello_encoding/src/path.rs
+++ b/vello_encoding/src/path.rs
@@ -683,7 +683,7 @@ impl<'a> PathEncoder<'a> {
             self.close();
         }
         if self.state == PathState::MoveTo {
-            let new_len = self.data.len() - 8;
+            let new_len = self.data.len() - 2;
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
MAYBE.

tomcur (approved):
These things are so easy to miss...

xorgy (approved):

waywardmonkeys:
Going to skip the test for now ... it wouldn't really catch much going forward unless I did extensive testing of all sorts of details of the encoding. I'd rather do that on future encoding stuff in the new implementation of things.

PR #878, opened 2025-03-26
[`vello_cpu`]: Basic clipping

grebmeg:
This PR introduces various changes to `vello_common` and `vello_cpu`, focusing on modifying coarse and fine rasterizations to support clipping, and updating `RenderContext` to handle it. It also includes updates to test cases to reflect these changes, along with new snapshot tests for clipping functionality. Based on https://github.com/linebender/vello/pull/818/commits/184dd5ba2bc5d535c78e131c536f0faa0157b626 (#818).

### Overview:
- Added nested clipping functionality with new stacks (`state_stack`, `clip_stack`) and structs (`SceneState`, `Clip`)
- Introduced clip commands (`PushClip`, `PopClip`, `ClipFill`, `ClipStrip`) and tracking of zero/non-zero clips in `WideTile`
- Updated `Fine` to handle multiple clip layers via stacked scratch buffers (`Vec<ScratchBuf>`) and new methods (`clip_fill`, `clip_strip`)
- Adjusted render pipeline (`RenderContext`) to properly pop all clips (`pop_clips`) before fine rasterization
- Created new snapshot tests (e.g. `clip_triangle_with_star`, `clip_rectangle_with_star_…`, `clip_with_*`) to verify nested, transformed clipping.

Output from some clipping tests:
| Test | Snapshot |
|:----:|:--------:|
clip_triangle_with_star|clipped_triangle_with_star|
clip_rectangle_and_circle|clip_rectangle_and_circle|
clip_rectangle_with_star_evenodd|clip_rectangle_with_star_evenodd|
clip_rectangle_with_star_nonzero||


sparse_strips/vello_cpu/snapshots/clipped_triangle_with_star.png
LaurenzV, line 1: Feel free to ignore, but for my OCD it would be great if the triangle could be centered in the picture. 👀 And in general, I've tried to use 100x100 images for such basic functionality test cases, t would be great if we can keep this up, but if there is a good reason to use more here also feel free to ignore!

Other suggestions for additional test cases from my side would be on the one hand simple rectangle clipping, on the other hand the `star_path` is also a good candiate for that, because it allows you to test nonzero and evenodd clipping (though my understanding is that only the first is implemented so far? In which case you can only add the first one). Other than that, it would also be great to see test cases for the interactions with transforms currently in place, but also feel free to ignore for now if you want the main focus to be on getting this merged soon.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -94,6 +94,14 @@ impl RenderContext {
         self.stroke_path(&rect.to_path(DEFAULT_TOLERANCE));
     }
 
+    /// Clip a path.
+    pub fn clip(&mut self, path: &BezPath) {
```
LaurenzV, line 98: I think this should be  `push_clip`, and there also should be a corresponding `pop_clip` which the user can call (or rather _has to call_ at some point) to pop the last clip path? I don't thinnk it's worth it making `render_to_pixmap` take a mutable reference to `self` for popping the clip paths. In my eyes, it's in the responsibility of the user to do so, and we can add a panic in `render_pixmap` in case not all clip paths have been popped.

LaurenzV (commented):
Haven't reviewed the logic in detail (and I know you aren't done yet, just two things I wanted to point out). Overall looks very nice though, thanks for your work on this!

sparse_strips/vello_cpu/src/render.rs
LaurenzV: In this case, does the suggestion I made make sense?

grebmeg: > I think this should be push_clip, and there also should be a corresponding pop_clip which the user can call (or rather has to call at some point) to pop the last clip path?

I thought most renderers (like Canvas 2D, Skia, Cairo, etc.) typically call it just clip? Internally in sparse_strips, it’s implemented using a push/pop stack model, but explicit push_clip / pop_clip naming is used in coarse rasterization.

> I don't thinnk it's worth it making render_to_pixmap take a mutable reference to self for popping the clip paths. In my eyes, it's in the responsibility of the user to do so, and we can add a panic in render_pixmap in case not all clip paths have been popped.

I’ve updated the code to make `finish` public and added a panic in `render_to_pixmap` if not all clip paths have been popped.

grebmeg:
Hey everyone, I think this is ready for an initial review. Feel free to take a look and share any feedback or ideas!

grebmeg:
cc @taj-p

LaurenzV:
Ah, sorry, I might not have been clear enough regarding the star test case. Instead of something like this

```rs
#[test]
fn clip_rectangle_with_star_nonzero() {
    let mut ctx = get_ctx(100, 100, true);

    let clip_rect = Rect::new(5.0, 30.0, 50.0, 70.0);
    let star_path = crossed_line_star();

    let stroke = Stroke::new(1.0);
    ctx.set_paint(DARK_BLUE.into());
    ctx.set_stroke(stroke);
    ctx.stroke_rect(&clip_rect);

    ctx.clip(&clip_rect.to_path(0.1));

    ctx.set_paint(REBECCA_PURPLE.into());
    ctx.set_fill_rule(Fill::NonZero);
    ctx.fill_path(&star_path);

    check_ref(&mut ctx, "clip_rectangle_with_star_nonzero");
}
```

I meant something like this
```
#[test]
fn clip_rectangle_with_star_nonzero() {
    let mut ctx = get_ctx(100, 100, true);

    let rect = Rect::new(0.0, 0.0, 100.0, 100.0);
    let star_path = crossed_line_star();

    let stroke = Stroke::new(1.0);
    ctx.set_paint(DARK_BLUE.into());
    ctx.set_stroke(stroke);
    ctx.stroke_rect(&rect);

    
    ctx.clip(&star_path, FillRule::NonZero);

    ctx.set_paint(REBECCA_PURPLE.into());
    ctx.fill_path(&rect.to_path(0.1));

    check_ref(&mut ctx, "clip_rectangle_with_star_nonzero");
}
``` 

Regarding the `EvenOdd`, I meant that in theory you should be able to do something like this as well:
`ctx.clip(&star_path, FillRule::EvenOdd);`. 

In which case the result should look like this (with purple instead, though):
image


However, as I mentioned it seems like currently, clipping is only support with `NonZero`? In which case the test `clip_rectangle_with_star_evenodd` could be left out. But I think for future-prooing, it would still nice if `clip` takes a fill rule argument, with a note that currently only `NonZero` is supported. What do you think?

sparse_strips/vello_cpu/src/render.rs
LaurenzV: Let's see what @raphlinus says. But I think it should be possible that if you for example push two clip paths, in the API it should be possible that you can selectively just pop one of them and then keep drawing with the first clip path in place instead of having to always push _all_ clip paths at once that are set.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -106,61 +140,386 @@ impl Wide {
...
+            let wtile_x0 = (x0 / WideTile::WIDTH).max(bbox[0]);
             // It's possible that a strip extends into a new wide tile, but we don't actually
             // have as many wide tiles (e.g. because the pixmap width is only 512, but
             // strip ends at 513), so take the minimum between the rounded values and `width_tiles`.
-            let tile_x1 = x1.div_ceil(WideTile::WIDTH).min(width_tiles);
+            let wtile_x1 = x1.div_ceil(WideTile::WIDTH).min(bbox[2]);
+
+            // Adjust column starting position if needed to respect clip boundaries
             let mut x = x0;
+            if (bbox[0] * WideTile::WIDTH) > x {
```
LaurenzV, line 176: Maybe store this value in a veriable like `clip_x`?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -106,61 +140,386 @@ impl Wide {
...
             }
 
+            // Determine if the region between this strip and the next should be filled
+            // based on the fill rule (NonZero or EvenOdd)
             let active_fill = match fill_rule {
                 Fill::NonZero => next_strip.winding != 0,
                 Fill::EvenOdd => next_strip.winding % 2 != 0,
             };
 
+            // If region should be filled and both strips are on the same row,
```
LaurenzV, line 203: Thanks for adding some comments for these!

sparse_strips/vello_cpu/src/render.rs
grebmeg: Ah, gotcha! Yeah, it definitely should be. What do you think about having `save`/`restore` methods for that? I added them in 5f512ba1 commit, along with a test.

sparse_strips/vello_common/src/coarse.rs
grebmeg: Yeah, good idea!

Fixed in d1c0f751

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -106,63 +142,409 @@ impl Wide {
...
+    /// intersection with the current clip region, and updates the clip stack.
+    ///
+    /// # Algorithm overview:
+    /// 1. Calculate bounding box of the clip path
+    /// 2. Intersect with current clip bounding box
+    /// 3. For each tile in the intersected bounding box:
+    ///    - If covered by zero winding: `push_zero_clip`
+    ///    - If fully covered by non-zero winding: do nothing (clip is a no-op)
+    ///    - If partially covered: `push_clip`
+    pub fn push_clip(&mut self, strips: Vec<Strip>, fill_rule: Fill) {
```
grebmeg, line 240: > However, as I mentioned it seems like currently, clipping is only support with NonZero? In which case the test clip_rectangle_with_star_evenodd could be left out. But I think for future-prooing, it would still nice if clip takes a fill rule argument, with a note that currently only NonZero is supported. What do you think?

Hey @LaurenzV, sorry for the misunderstanding, and thanks for clarifying! I believe fill rules for clipping are now supported with this 2ed675e4 commit.

nonzero|evenodd
---|---
clip_rectangle_with_star_nonzero|clip_rectangle_with_star_evenodd

LaurenzV: Lovely!

raphlinus:
Internally we use push and pop for clips, but the more standard API is save/restore, where restore pops all clips since the last save. It makes sense to primarily expose that interface. That said, for blends it's common to have explicit [pushlayer](https://learn.microsoft.com/en-us/windows/win32/api/d2d1/nf-d2d1-id2d1rendertarget-pushlayer(constd2d1_layer_parameters__id2d1layer)) exposed in the interface. I don't have strong feelings on the exact form of the API, open to discussion and what's easiest to implement.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -106,63 +142,409 @@ impl Wide {
...
+                    wtile_x += 1;
+                }
+
+                // Pop zero clips for tiles that had zero winding or didn't match fill rule
+                // TODO: The winding check is probably not needed; if there was a fill,
+                // the logic below should have advanced tile_x.
+                let is_inside = match fill_rule {
+                    Fill::NonZero => strip.winding != 0,
+                    Fill::EvenOdd => strip.winding % 2 != 0,
+                };
```
LaurenzV, line 446: So can this be removed then?

sparse_strips/vello_cpu/tests/util.rs
```diff
@@ -37,14 +37,14 @@ pub(crate) fn get_ctx(width: u16, height: u16, transparent: bool) -> RenderConte
     ctx
 }
 
-pub(crate) fn render_pixmap(ctx: &RenderContext) -> Pixmap {
+pub(crate) fn render_pixmap(ctx: &mut RenderContext) -> Pixmap {
```
LaurenzV, line 40: Hmm, I think I would still prefer this not taking a &mut, and we instead call `finish` in each of the clip path tests, so that we don't have to pass mut for every test cases, even the ones that don't need it.

LaurenzV (approved):
As mentioned I haven't reviewed the actual logic in detail, but given the discussion in the office hours I think this is safe to land (the change to make bbox a struct would indeed be nice to have, though).

sparse_strips/vello_common/src/coarse.rs
grebmeg: I think so — `wtile_x` should be advanced in the code below, but I feel like I need to double-check it more thoroughly. If you don’t mind, I’ll leave it as-is for now and come back to it a bit later.

sparse_strips/vello_cpu/tests/util.rs
grebmeg: Yep, makes sense to me!

Fixed in 4d1b9e2b

grebmeg:
> Internally we use push and pop for clips, but the more standard API is save/restore, where restore pops all clips since the last save. It makes sense to primarily expose that interface. That said, for blends it's common to have explicit [pushlayer](https://learn.microsoft.com/en-us/windows/win32/api/d2d1/nf-d2d1-id2d1rendertarget-pushlayer(constd2d1_layer_parameters__id2d1layer)) exposed in the interface. I don't have strong feelings on the exact form of the API, open to discussion and what's easiest to implement.

Hey @raphlinus, thanks for the comment! I don’t have strong preferences either, so if you think it makes more sense to use a different API style, I’m happy to update the code. Open to whatever direction makes the most sense.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -137,23 +156,36 @@ impl RenderContext {
...
         }
     }
 
+    /// Finish the coarse rasterization prior to fine rendering.
+    ///
+    /// This method is called when the render context is finished with rendering.
+    /// It pops all the clips from the wide tiles.
+    pub fn finish(&mut self) {
+        self.wide.pop_clips();
+    }
```
DJMcNab, line 187: I'm slightly concerned by this method - it seems hard to reason about when you need to call it, and very easy to miss.

"This method is called" is very passive. Who should call it?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -73,6 +73,17 @@ impl RenderContext {
...
+    /// Save the current scene state.
+    pub fn save(&mut self) {
+        self.wide.state_stack.push(SceneState { n_clip: 0 });
+    }
+
+    /// Restore the previous scene state.
+    pub fn restore(&mut self) {
+        self.wide.pop_clips();
+        self.wide.state_stack.pop();
+    }
```
DJMcNab, line 85: As far as I can tell, the panic for calling this too many times is an `unwrap` in `finish` or the next `restore`. Is that what we want?

DJMcNab (commented):
A few discussion points.

sparse_strips/vello_cpu/src/render.rs
grebmeg: Would a better name help clarify this, perhaps something like `pop_all_clips`, along with documentation explaining when and why it should be used? 

Or maybe we could align with the `vello_hybrid` API by introducing `prepare_render_data` method, which could serve as the final step before fine rasterization in both renderers? For the second, I think `render_to_pixmap` method can be seen as an analogue to `Renderer::render` in `vello_hybrid`, where in `prepare` we call `prepare_render_data`. We could apply the same approach in `vello_cpu` — the logic would be simple, but unifying the structure across both renderers would be a nice improvement.

sparse_strips/vello_cpu/src/render.rs
grebmeg: Good callout! I can definitely rewrite this in a safer way that avoids panicking. That said 🤔 I think I’d still prefer to panic explicitly here, since unbalanced save/restore calls would indicate a programming error. Continuing execution might just lead to more confusing behavior down the line. Of course, it’s possible there are other use cases I’m not seeing yet — happy to hear your thoughts!

DJMcNab: Sure. Panicking is reasonable, but the panic we get does not at all make it clear what you've done wrong.

That is, the issue isn't (necessarily) the panic, it's that we don't have a way to lead the user back to the correct way.
(An alternative API would return a token from `save` which you have to pass into `restore` in the right order, although that's probably too convoluted)

sparse_strips/vello_cpu/src/render.rs
DJMcNab: I'm not sure. Documenting when it should be used would be a good first step.

LaurenzV: I think the best thing in the long term would be a `save`/`restore` like API where as the user you just need to make sure that you calls to save and restore are balanced, and the rest is taken care of behind the scenes.

PR #879, opened 2025-03-26
Add example using Android View for Vello Hybrid

DJMcNab:
DO NOT MERGE!

This is a proof of concept only at the moment.

cc for awareness @mwcampbell 

[Android View](https://github.com/mwcampbell/android-view) is the upcoming new way to make Android apps with Rust.


DJMcNab:
Closing: This is still useful, but it is not at all intended for merge (it's less hacky than I expected, but still pretty hacky).

Of potential interest is the device management, which is quite different to current Vello, because we are assuming there's only one surface at a time. This is because Android View doesn't currently have a way to have process-wide state (which I think is *probably* a mistake, although I recognise how the model makes that hard; because you can have multiple view types linked together in the same app).

I really hate how much boilerplate work doing this properly requires, which no-one bothers to do.

PR #881, opened 2025-03-27
Add unit tests for tiling

LaurenzV:
It would be nice to also add some tests for how it behaves at tile boundaries, but since at least from my experiments we do not seem to produce ideal results for those cases (yet?), I've left them out for now.

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -531,8 +531,6 @@ mod tests {
         );
     }
 
-    // At tile boundary
-
     #[test]
     // See https://github.com/LaurenzV/cpu-sparse-experiments/issues/46.
     fn infinite_loop() {
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -322,7 +322,216 @@ impl Tiles {
 #[cfg(test)]
 mod tests {
     use crate::flatten::{Line, Point};
-    use crate::tile::Tiles;
+    use crate::tile::{Tile, Tiles};
+
+    #[test]
+    fn cull_line_at_top() {
+        let line = Line {
+            p0: Point { x: 3.0, y: -5.0 },
+            p1: Point { x: 9.0, y: -1.0 },
+        };
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&[line], 100, 100);
+
+        assert!(tiles.is_empty());
+    }
+
+    #[test]
+    fn cull_line_at_right() {
+        let line = Line {
+            p0: Point { x: 101.0, y: 0.0 },
+            p1: Point { x: 103.0, y: 20.0 },
+        };
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&[line], 100, 100);
+
+        assert!(tiles.is_empty());
+    }
+
+    #[test]
+    fn cull_line_at_bottom() {
+        let line = Line {
+            p0: Point { x: 30.0, y: 101.0 },
+            p1: Point { x: 35.0, y: 105.0 },
+        };
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&[line], 100, 100);
+
+        assert!(tiles.is_empty());
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_common/src/tile.rs
```diff
@@ -322,7 +322,214 @@ impl Tiles {
...
+        };
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&[line], 100, 100);
+
+        assert_eq!(tiles.tile_buf, vec![Tile::new(0, 0, 0, true)]);
+    }
+
+    #[test]
+    fn vertical_straight_line() {
```
tomcur, line 380:
```suggestion
    fn horizontal_straight_line() {
```

sparse_strips/vello_common/src/tile.rs
```diff
@@ -322,7 +322,214 @@ impl Tiles {
...
+            vec![
+                Tile::new(0, 0, 0, false),
+                Tile::new(1, 0, 0, false),
+                Tile::new(2, 0, 0, false),
+            ]
+        );
+    }
+
+    #[test]
+    fn horizontal_straight_line() {
```
tomcur, line 401:
```suggestion
fn vertical_straight_line() {
```

sparse_strips/vello_common/src/tile.rs
```diff
@@ -322,7 +322,214 @@ impl Tiles {
...
+    fn partially_cull_line_exceeding_viewport() {
+        let line = Line {
+            p0: Point { x: -2.0, y: -3.0 },
+            p1: Point { x: 2.0, y: 1.0 },
+        };
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&[line], 100, 100);
+
+        assert_eq!(tiles.tile_buf, vec![Tile::new(0, 0, 0, true)]);
```
tomcur, line 376: It may be slightly faster to compare to a slice, sparing the allocation:

```suggestion
        assert_eq!(&tiles.tile_buf, &[Tile::new(0, 0, 0, true)]);
```

tomcur (commented):
> It would be nice to also add some tests for how it behaves at tile boundaries, but since at least from my experiments we do not seem to produce ideal results for those cases (yet?), I've left them out for now.

What are the issues you're seeing? Currently, for x-coordinates at tile boundaries, lines should be assigned to the tile they are on the left boundary of (i.e., the tile to the line's right). The same is done per-pixel in the coverage calculations.

For lines only touching the top or bottom edge of a tile, iirc they don't belong to that tile.

This behavior isn't really _necessary_, but it is important to be consistent.

LaurenzV:
> What are the issues you're seeing?

Sorry, issues was probably not the right phrasing. I didn't mean that there is a bug or anything, but simply that I wasn't sure whether it is the expected behavior (as you mentined for example, if a line is at a tile-aligned x coordinate, we also create a tile for the right side of the line, even though strictly speaking it probably wouldn't be necessary). I'm not sure if it's worth testing for this though since it's kind of an implementation detail, so I think the current tests are a good start.

tomcur (approved):
Very useful to have these tests to ensure we don't suddenly start emitting unnecessary tiles. I have not manually traced the lines specified here to check whether the specified tiles are correct, but I suspect you have.

A few more places can use arrays instead of `vec`, I pointed at one inline.

> (as you mentined for example, if a line is at a tile-aligned x coordinate, we also create a tile for the right side of the line, even though strictly speaking it probably wouldn't be necessary)

(In that case, only the right tile should be created. The left tile should only be created if the line approaches that coordinate from the left.)

sparse_strips/vello_common/src/tile.rs
```diff
@@ -322,7 +322,214 @@ impl Tiles {
...
+            p1: Point { x: 8.5, y: 1.0 },
+        };
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&[line], 100, 100);
+        tiles.sort_tiles();
+
+        assert_eq!(
+            tiles.tile_buf,
+            vec![
```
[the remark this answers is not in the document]
LaurenzV: Ah, sorry about that, shoul've seen those myself. 😅

PR #882, opened 2025-03-28
vello_hybrid: reuse allocation for staging alpha data

tomcur:
This drops the need to always allocate for alphas in `Renderer::prepare`.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render.rs
+++ b/sparse_strips/vello_hybrid/src/render.rs
@@ -59,6 +59,9 @@ pub struct Renderer {
     pub render_pipeline: RenderPipeline,
     /// GPU resources for rendering (created during prepare)
     resources: Option<GpuResources>,
+
+    /// Scratch buffer for staging alpha texture data.
+    alpha_data: Vec<u8>,
 }
 
 /// Contains the data needed for rendering
@@ -196,6 +199,7 @@ impl Renderer {
             render_bind_group_layout,
             render_pipeline,
             resources: None,
+            alpha_data: Vec::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
It's a bit of a shame that this copy exists at all. It doesn't seem extremely necessary (it looks like it's only to pad out the data to be a rectangular region in the texture?

tomcur:
That's right. It might be possible to have `RenderData::alphas` already be padded. However, scheduling of GPU commands will have to change in the future (e.g., currently if alphas were to overflow the texture size, things will break), so it's probably better to wait until that's settled somewhat before optimizing this padding too much.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -333,8 +342,9 @@ impl Renderer {
             render_data.alphas.len() <= (texture_width * texture_height * 16) as usize,
             "Alpha texture dimensions are too small to fit the alpha data"
         );
-        let mut alpha_data = render_data.alphas.clone();
-        alpha_data.resize((texture_width * texture_height * 16) as usize, 0);
+        // After this copy to `self.alpha_data`, there may be stale trailing alpha values. These
+        // are not sampled, so can be left as-is.
+        self.alpha_data[0..render_data.alphas.len()].copy_from_slice(&render_data.alphas);
```
grebmeg, line 347: Just an idea — would it be more performant to use `std::mem::swap` with the cloned alphas from `scene.prepare_render_data` first, and then resize afterward? Something like:

```rust
std::mem::swap(&mut self.alpha_data, &mut render_data.alphas);
if self.alpha_data.len() < (texture_width * texture_height * 16) as usize {
    self.alpha_data
        .resize((texture_width * texture_height * 16) as usize, 0);
}
```

DJMcNab: I think in that scheme, we don't even need the "intermediate" buffer at all. That is, it's not doing anything.

I think the right change here would have been to just resize the existing `render_data.alphas`; when I reviewed this I didn't realise the fact that we owned those (ideally, they'd also get reused...)

tomcur: I was thinking it's likely those buffers will get reused, once draw scheduling becomes more complex.

PR #883, opened 2025-03-31
[Sparse Strip]: Text API (outlines only)

taj-p:
## Intent

Adds an API to `vello_cpu` and `vello_hybrid` for drawing text: 

```rs
// A Vec of glyph positions: Vec<Glyph { id: u32,  x: f32, y: f32 }>
let glyphs = // ...

scene.glyph_run(&font)
    .normalized_coords(...)
    .font_size(size)
    .hint(true)
    .fill_glyphs(glyphs.iter());
```

## Background

This API does not perform text shaping and layout. Those functions should be performed by other libraries like Parley or Cosmic Text. The API proposed here merely takes glyph positions and rendering attributes and renders them to screen.

## Comparison with Vello

The API should feel familiar to Vello's implementation with one change: We defer to the scene's style and transform.

- The glyph run doesn't support a global transform as [in Vello](https://github.com/linebender/vello/blob/bf95b8501a9ab92aa5474095cc05dd885588fe1e/vello/src/scene.rs#L380-L383). 
- Instead of [specifying the brush](https://github.com/linebender/vello/blob/bf95b8501a9ab92aa5474095cc05dd885588fe1e/vello/src/scene.rs#L435-L447), we use the stroke / fill settings applied to the scene and choose between the two via `[stroke|fill]_glyphs(...)`.

Happy to change this, but I don't see why the consumer can't use the existing transform/paint/stroke APIs on `Scene`. In other words, the scene's transform/paint/stroke applying to every primitive except text seems unexpected. 

If text deserves special attention, I'm keen to understand that choice and this PR is certainly flexible to that (and any other) change. Please let me know 🙏 .

## Design

To support both CPU and Hybrid variants of the renderer, we expose a `GlyphRunBuilder` from `vello_common` that accepts a `GlyphRenderer` trait. The builder encapsulates the logic to "prepare" some glyph for rendering before passing it to the `GlyphRenderer` to draw to screen. I've implemented `GlyphRenderer` for both our CPU and Hybrid variants. The CPU variant implementation is copied below:

https://github.com/linebender/vello/blob/87d18011da75a1d04e31859c7b7a278c2bbd099d/sparse_strips/vello_cpu/src/render.rs#L185-L209

## Next Steps

This PR only supports outlined glyphs. We need to support Emoji through bitmap and colr variants. I'd prefer to do that in separate PRs and keep this PR focused on the API and overall strategy.

Unfortunately, I suspect these other two variants cannot be started until we support:

1. Clipping for colr glyphs (via [`ColorPainter`](https://github.com/googlefonts/fontations/blob/618fcf9a45aec9244b69e8c923e6698be6c155c4/skrifa/src/color/mod.rs#L203))
2. Images for bitmap glyphs

Perhaps since clipping is underway, I could look at image support? But, again, I worry that, like fast text rendering, image rendering would likely need some form of caching 🤔 .

## Notes

I haven't implemented caching of glyphs nor hinting instances. As per discussions, we wanted to keep the text API free from caching for now.


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -174,3 +181,29 @@ impl RenderContext {
...
+    fn fill_glyphs(&mut self, glyphs: impl Iterator<Item = PreparedGlyph>) {
+        for glyph in glyphs {
+            match glyph {
+                PreparedGlyph::Outline(glyph) => {
+                    let transform = self.transform * glyph.local_transform;
+                    flatten::fill(&glyph.path, transform, &mut self.line_buf);
+                    self.render_path(self.fill_rule, self.paint.clone());
+                }
+            }
+        }
```
taj-p, line 195: I think that if we know the font/run doesn't contain any non-outline glyph, we could provide a fast path that avoids this per glyph match. But, I'd like to explore that once we've added in those glyph types.

sparse_strips/vello_common/Cargo.toml
```diff
@@ -15,6 +15,7 @@ publish = false
 vello_api = { workspace = true, default-features = true }
 # for pico_svg
 roxmltree = "0.20.0"
+skrifa = { workspace = true }
```
LaurenzV, line 19: I think this should definitely be behind a feature flag, as not everyone might nee text rendering, and it's a pretty big dependency. But I think for now just a TODO would be fine.

sparse_strips/vello_api/src/glyph.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// The font-specific identifier for this glyph.
+    ///
+    /// This ID is specific to the font being used and corresponds to the
+    /// glyph index within that font. It is *not* a Unicode code point.
+    pub id: u32,
+    /// X-offset in run, relative to transform.
+    pub x: f32,
+    /// Y-offset in run, relative to transform.
+    pub y: f32,
+}
```
LaurenzV, line 18: I think most shaping libraries will give you outputs where you have x_advance/y_advance and x_offset/y_offset instead of absolute positions, so in the future it might be nice if we provide the user with a way or directly using those, instead of having to calculate the absolute positions manually. That said, I think this is a good starting point and we should keep things simple for now.

sparse_strips/vello_api/src/glyph.rs
```diff
@@ -0,0 +1,18 @@
...
+//! Types for glyphs.
+
+/// Positioned glyph.
+#[derive(Copy, Clone, Default, Debug)]
+pub struct Glyph {
+    /// The font-specific identifier for this glyph.
+    ///
+    /// This ID is specific to the font being used and corresponds to the
+    /// glyph index within that font. It is *not* a Unicode code point.
+    pub id: u32,
```
LaurenzV, line 13: I think there are two choices here, either we decide to completely hide skrifa from the public API  ( in which case we can just use u32 here), in which case we should build our own wrapper for `NormalizedCoord` further below, or we decide to just reuse it, in which case we might as well use the GlyphID type by skrifa.

Not sure what the best approach is.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    /// Set the font size in pixels per em.
+    pub fn font_size(mut self, size: f32) -> Self {
+        self.run.font_size = size;
+        self
+    }
+
+    /// Set the per-glyph transform. Can be used to apply skew to simulate italic text.
+    pub fn glyph_transform(mut self, transform: Affine) -> Self {
+        self.run.glyph_transform = Some(transform);
```
LaurenzV, line 73: I'm just wondering, how does hinting interact with glyph transforms? My understanding is that hinting doesn't make a lot of sens if you for example have a rotated glyph?

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,190 @@
...
+        self.renderer
+            .stroke_glyphs(Self::prepare_glyphs(&self.run, glyphs));
+    }
+
+    fn prepare_glyphs(
+        run: &GlyphRun,
+        glyphs: impl Iterator<Item = &'a Glyph>,
+    ) -> impl Iterator<Item = PreparedGlyph> {
+        let font = skrifa::FontRef::from_index(run.font.data.as_ref(), run.font.index).unwrap();
+        let outlines = font.outline_glyphs();
```
LaurenzV, line 106: I'm also not sure what the best way of caching the `FontRef` is. What I've done in other crates is to use a self-referential struct where we store the FontRef inside of the struct that holds the underlying data, but unfortunately this is not possible without `unsafe` or for example using the [yoke](https://docs.rs/yoke/latest/yoke/) crate. But we can also ignore this for now.

sparse_strips/vello_common/src/glyph.rs
LaurenzV: For example, if someone sets the font size to 12 but adds a scaling transform of 2, I guess this should acutally mean that the font size used for hinting should be 24? But maybe those are details we don't need to worry about.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: I don't think that this match will be super expensive in comparison to the other stuff that follows.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -174,3 +181,29 @@ impl RenderContext {
...
 }
+
+impl GlyphRenderer for RenderContext {
+    fn fill_glyphs(&mut self, glyphs: impl Iterator<Item = PreparedGlyph>) {
+        for glyph in glyphs {
+            match glyph {
+                PreparedGlyph::Outline(glyph) => {
+                    let transform = self.transform * glyph.local_transform;
+                    flatten::fill(&glyph.path, transform, &mut self.line_buf);
+                    self.render_path(self.fill_rule, self.paint.clone());
```
LaurenzV, line 192: I think that glyphs always should be rendered with `NonZero`?

sparse_strips/vello_cpu/tests/basic.rs
```diff
@@ -522,6 +526,98 @@ fn filled_vertical_hairline_rect_2() {
...
+    ctx.set_paint(REBECCA_PURPLE.with_alpha(0.5).into());
+    ctx.glyph_run(&font)
+        .font_size(font_size)
+        .glyph_transform(Affine::skew(-20_f64.to_radians().tan(), 0.0))
+        .fill_glyphs(glyphs.iter());
+
+    check_ref(&ctx, "skewed_glyphs");
+}
+
+fn layout_glyphs(text: &str, font_size: f32) -> (Font, Vec<Glyph>) {
```
LaurenzV, line 575: It's also okay for now, but in the future we should probably add a dev dependency on rustybuzz for shaping so that we can render complex text, too.

LaurenzV (commented):
There are definitely some things we will need to change in the long run, but I think it's a good start! :)

sparse_strips/vello_cpu/tests/basic.rs
tomcur: A comment here (and in vello_hybrid) that this is not full shaping, but just used to get some simple text on screen, may be useful. Just in case someone copies the code into their own projects.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Consumes the builder and strokes the glyphs with the current configuration.
+    pub fn stroke_glyphs(self, glyphs: impl Iterator<Item = &'a Glyph>) {
+        self.renderer
+            .stroke_glyphs(Self::prepare_glyphs(&self.run, glyphs));
+    }
+
+    fn prepare_glyphs(
+        run: &GlyphRun,
+        glyphs: impl Iterator<Item = &'a Glyph>,
+    ) -> impl Iterator<Item = PreparedGlyph> {
```
tomcur, line 104: This allocates a `BezPath` for each `PreparedGlyph`. As we don't look at multiple glyphs at the same time, we could instead reuse a single underlying `BezPath` allocation. Perhaps by using something like a [lending iterator](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#what-are-gats), or shuffling code around to prevent needing an iterator.

sparse_strips/vello_api/src/glyph.rs
DJMcNab: I think that this is probably for a different level in the stack. That is, your layout engine accepts this information from the shaper, and computes the absolute positions.

sparse_strips/vello_common/Cargo.toml
DJMcNab: I can't think of very many use cases which don't need text at all.
The key question in my mind becomes do you want scenes to be composable. Currently, that isn't supported in sparse strips (AIUI). Doing text externally is hard because of the hinting question; you can't compose scenes with text, because hinting needs to know about the absolute size, and whether the rotation is off-axis.
But OTOH, I guess this composition is messy for text already, because for example you "need" to know the optical size.

Some of this opens the door to text rendering not even being part of these crates at all. Tiny Skia makes that choice. That is, remove `draw_glyphs` from being a method on the context and instead into its own module entirely, say.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /// Set whether font hinting is enabled.
+    pub fn hint(mut self, hint: bool) -> Self {
+        self.run.hint = hint;
+        self
+    }
+
+    /// Set normalized variation coordinates for variable fonts.
+    pub fn normalized_coords(mut self, coords: Vec<NormalizedCoord>) -> Self {
```
DJMcNab, line 84: This type should be as in #747

DJMcNab: Interesting about this taking in a `Vec`. I don't fully hate it - it's honest about the allocations. But it would also probably be better to re-use the same allocation instead (or get a lifetime from a passed in slice)

sparse_strips/vello_common/src/glyph.rs
DJMcNab: Isn't making a `FontRef` really cheap?

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,190 @@
...
+            .stroke_glyphs(Self::prepare_glyphs(&self.run, glyphs));
+    }
+
+    fn prepare_glyphs(
+        run: &GlyphRun,
+        glyphs: impl Iterator<Item = &'a Glyph>,
+    ) -> impl Iterator<Item = PreparedGlyph> {
+        let font = skrifa::FontRef::from_index(run.font.data.as_ref(), run.font.index).unwrap();
+        let outlines = font.outline_glyphs();
+        let size = Size::new(run.font_size);
```
DJMcNab, line 107: Either the current transform shouldn't apply, or we should scale the size used for hinting based on the current transform. I think that the latter is more sensible.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,190 @@
...
+        let normalized_coords = run.normalized_coords.as_slice();
+        let hinting_instance = if run.hint {
```
DJMcNab, line 109: Should this still hint if there is e.g. an active rotation?

sparse_strips/vello_cpu/tests/basic.rs
DJMcNab: For tests, this is fine. We need to caveat it heavily, though

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,249 @@
...
+            let advance = glyph_metrics.advance_width(gid).unwrap_or_default();
+            let x = pen_x;
+            pen_x += advance;
+            Some(Glyph {
+                id: gid.to_u32(),
+                x,
+                y: pen_y,
+            })
+        })
+        .collect::<Vec<_>>();
```
DJMcNab, line 220: I'd lean towards just using Parley here. Like yes, it's a lot more expensive, but we expect the majority of users of this to use it.
We have evidence of people copying very similar code uncritically, because text layout is not a familiar question to most people.

It would also be nice to use system fonts, but using Roboto would also be reasonable.

Yes, it makes the code longer, but it better emphasises the separation of concerns.

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    ctx.set_paint(palette::css::WHITE.into());
+    let transform = Affine::scale(2.0).then_translate((0., f64::from(size) * 2.0).into());
+    ctx.set_transform(transform);
+
+    // Fill the text
+    ctx.glyph_run(&font)
+        .normalized_coords(vec![])
```
DJMcNab, line 228: It could be fun to really demonstrate variable font support here. The one tricky thing is that we don't have one in-repo. https://github.com/linebender/xilem/tree/main/xilem/resources/fonts/roboto_flex would be possible here (100KB)

sparse_strips/vello_hybrid/examples/text.rs
DJMcNab, line 1: I'm also slightly concerned by the proliferation of examples here. What happens when we need to e.g. update winit; that seems like it would become very painful.

DJMcNab (commented):
Thanks! It's really great to get this working. It's impressive how little code this takes.

Nothing fundamental here, but no shortage of nits!

It will be good to get benchmarks in for this (not now)

sparse_strips/vello_common/src/glyph.rs
dfrg: Vello examines the transform and if it contains a uniform scale with no rotation or skew, then it directly applies the scale to the requested font size. Otherwise it disables hinting for the run.

sparse_strips/vello_common/src/glyph.rs
dfrg: It’s fairly cheap. Certainly much cheaper than calling `outline_glyphs()`. I’d like to make a cacheable accelerator for this but it hasn’t been a priority so far.

sparse_strips/vello_api/src/glyph.rs
taj-p: > I think this is a good starting point and we should keep things simple for now.

👍 

> I think that this is probably for a different level in the stack.

Agree

Going by both your opinions, it sounds like we can safely resolve this comment without update.

sparse_strips/vello_common/src/glyph.rs
taj-p: Great idea! I've stopped exposing Skrifa API and used a normalised coords slice in https://github.com/linebender/vello/pull/883/commits/db226ef37898591d6704af2a5c3c38c1652f26c3

sparse_strips/vello_api/src/glyph.rs
taj-p: Thanks for the context Daniel. We've removed Skrifa from the public API in https://github.com/linebender/vello/pull/883/commits/db226ef37898591d6704af2a5c3c38c1652f26c3. 

Laurenz, this means that we've chosen this route which you suggested: "in which case we should build our own wrapper for NormalizedCoord further below".

sparse_strips/vello_cpu/src/render.rs
taj-p: Nice catch - hardcoded to `NonZero` in https://github.com/linebender/vello/pull/883/commits/897c7a0e8eb6efb25875dd2e567465e2bbc606b0

sparse_strips/vello_cpu/tests/basic.rs
taj-p: Added heavy disclaimer to this function in https://github.com/linebender/vello/pull/883/commits/6ccfb54068b27d0ba2a7241e57b477a319a3aad1

sparse_strips/vello_common/Cargo.toml
```diff
@@ -15,6 +15,8 @@ publish = false
 vello_api = { workspace = true, default-features = true }
 # for pico_svg
 roxmltree = "0.20.0"
+bytemuck = { workspace = true }
```
DJMcNab, line 18: So far, we've worked really hard to avoid this dependency edge (on bytemuck in vello_common or vello_cpu) (I'm not entirely sure why, but I believe it to be well-motivated)

There isn't really a good way to avoid this for this case; dropping down to writing our own unsafe is definitely worth avoiding though.

I'm on mobile, so unfortunately can't easily link to the prior discussion here. I know it was on one of the previous sparse strips PRs.

Concretely, we should be sure this isn't getting default features. We *only* use cast_slice from this dependency. (This might already be the case, although I doubt it - I've not checked the virtual manifest)

LaurenzV: I believe skrifa already depends on bytemuck so we can't avoid it anyway for text rendering. But yes, would be nice to only pick the features actually needed.

sparse_strips/vello_common/src/glyph.rs
taj-p: > then it directly applies the scale to the requested font size.

I implemented this in https://github.com/taj-p/vello/pull/4 to get a feel for how it might work.

I'm unsure whether that behaviour is desirable because of the below two reasons. Please let me know if I've made some mistake in my thinking - I'm keen for other's opinion.

1. It creates a coupling on `font_size` and `transform` between typesetter and renderer, which leads to confusing behaviour and API consumption.

Consider the below code, which produces the below image:

```rs
    let size = 10;
    let glyph_positions = typesetter.run(..., size);

    renderer.glyph_run()
        .font_size(size)
        .hint(true)
        .fill_glyphs(glyph_positions);
```

image

If, for whatever reason, we now need to apply a transform to our renderer (perhaps the user has zoomed in by a factor of 2), then the path of least resistance would be to simply add a transform:

```rs
    let size = 10;
    let glyph_positions = typesetter.run(..., size);

    renderer.set_transform(Affine::scale(2.0));

    renderer.glyph_run()
        .font_size(size)
        .hint(true)
        .fill_glyphs(glyph_positions);
```

But, this produces illegible text. `glyph_positions` was based on a font size of 10, but the renderer, opaque to the consumer, is actually choosing a font size of 20 (and not obeying the passed in `size`). Thus, the size used to generate `glyph_positions` is half what the renderer used. This produces text like:

image

We could scale the glyph positions by the scale, but, IIUC, glyph positions aren't perfectly linear to font size (especially aspects like line height IIUC). This means that text may not be incorrectly laid out if we simply scaled glyph positions by 2 in this case.


To fix this situation, we would need to produce this confusing code:

```rs
    let size = 20;
    let scale = 2;
    let glyph_positions = typesetter.run(..., size);

    renderer.set_transform(Affine::scale(scale));

    renderer.glyph_run()
        .font_size(size / scale) // need to downscale size because its upscaled via transform
        .hint(true)
        .fill_glyphs(glyph_positions);
```

2. I'm not sure if it's always correct.

Let's imagine you're building a piece of software to design posters. If you zoom into the poster and text is, essentially, re-shaped due to the coupling of the renderer's transform to the font size, the user won't know how pixels will ultimately display when they print the poster because it looks different at different scales.

---

I'm curious to understand how Parley and application code generally handle this coupling. 

I would have thought it simplest to leave the decision to the consumer by keeping the renderer "dumb" (i.e. simply scale the outlines). They can choose to scale outlines with `transform` or via increasing font size to their typesetter; this allows for both types of consumer to achieve what they want with a straightforward API.

I think preventing hinting for rotated/skewed text is straightforward (https://github.com/linebender/vello/pull/883/commits/a57bed9e440eebfd3d756b183ffc025eb7e4560e). But, making the choice to scale font size for consumers doesn't seem ideal. In any case, if we don't prefer this default behaviour, I suspect configuration would be the next step? 

That said, I'm keen for everyone's thoughts!

sparse_strips/vello_common/src/glyph.rs
taj-p: Please see https://github.com/linebender/vello/pull/883#discussion_r2021986713 - does that make sense to you?

sparse_strips/vello_common/src/glyph.rs
taj-p: Disabled for rotation/skew in https://github.com/linebender/vello/pull/883/commits/a57bed9e440eebfd3d756b183ffc025eb7e4560e

sparse_strips/vello_hybrid/examples/text.rs
taj-p: I'll migrate this example to use Parley. I might not get time to it today.

sparse_strips/vello_common/src/glyph.rs
DJMcNab: Current Vello makes the assumption that:
> We could scale the glyph positions by the scale, but, IIUC, glyph positions aren't perfectly linear to font size (especially aspects like line height IIUC).

Is not true, and so scales the glyph positions. If that is true, then things are, as you say, a lot more complicated.
We certainly haven't seen any issues yet from making this assumption.
My understanding is that the `opsz` variable font axis is intended for this use case, to allow fonts to customise themselves based on the visual size. If these things could depend on the font size, that would probably be unnecessary.
Running through the Skrifa code, it certainly looks like this is only used as a dumb scale to multiply glyph outlines. It does impact the `hdmx` table results, but we access those anyway. (cc @dfrg I'm not sure why Skrifa accesses the hdmx table when hinting is disabled; it seems designed for only hinted use cases)

My gut reaction here is that when making the unhinted glyph outlines, we should use `Size::unscaled`, and fully perform the scaling ourselves.
And that by requesting hinting you have requested the glyphs to be hinted to their actual rendered size.

Then it's up to the user what `opsz` they use, but getting the right glyphs is up to them using the right settings for their layout engine.

(But again, I'm nowhere near a domain expert here, and am not planning on becoming one)

dfrg: Good catch! In vello (classic?), we capture the additional scale and apply it to the translation component of the glyph transform later:

https://github.com/linebender/vello/blob/61f82e4e66a9111d6b2ee17f173167c3601c2e55/vello_encoding/src/resolve.rs#L334

We also enable the “preserve linear metrics” flag in the hinter to ensure that horizontal metrics are not affected by hinting and the linear scale factor is valid.

None of the shapers take hinting into account by default although HarfBuzz can if you provide custom font funcs that call back into the scaler for hinted advance widths.

DJMcNab: This comment is unabashedly off-topic; the main actionable thing from this thread is that the only thing which needs to change is scaling the glyph translations, because we copied in the hinting settings. Using unscaled outlines would be mostly a caching advantage (to make boundaries clearer), which might not really be needed, and can be a follow-up.

It would be theoretically "better" to properly take into account the extra width requested by the hinter, right? However:
1) It's absurdly expensive (if there's no `hdmx` table; it shouldn't be in theory, but it requires passing "caches" around properly, and so you end up double-hinting).
2) `opsz` is now just a more suitable technology for this, with fewer pitfalls, but worse font support.

I guess the question I'm dancing around is where is hinting for the printer evaluated (and who configures those hinting settings). That is, can the layout engine not being aware of this butcher documents which are printed. Is this something which has happened to anyone before.

sparse_strips/vello_common/src/glyph.rs
taj-p: Prevented per glyph path allocation in https://github.com/linebender/vello/pull/883/commits/c458e8c694a69de11c5ac8343f5ad5805223df90

sparse_strips/vello_common/Cargo.toml
taj-p: Nice catch! Yes, this would have pulled in "derive" from the virtual manifest. I've set bytemuck to use no features in https://github.com/linebender/vello/pull/883/commits/b25c5076cdef25942dd2fb64a19e65988011eea3.

sparse_strips/vello_common/src/glyph.rs
taj-p: Thanks @dfrg and @DJMcNab for the careful consideration 🙏 . 

> we capture the additional scale and apply it to the translation component of the glyph transform later:

Ah yes, that makes sense.

> Digging into this further, this is a complete disaster, and there's no plausible sound way to deal with this!

🙈 😅 

> the main actionable thing from this thread is that the only thing which needs to change is scaling the glyph translations, because we copied in the hinting settings.

Please correct my misunderstanding, but I believe, like Vello, we apply scaling to the glyphs later by applying the scene's existing transform after the glyph's local transform (which contains glyph position).

https://github.com/linebender/vello/blob/18bf5cd85d6cea6c780f5f2d1366024faf635c6b/sparse_strips/vello_hybrid/src/scene.rs#L274-L275

I've added a snapshot test for scaled glyphs in https://github.com/linebender/vello/pull/883/commits/f0af18e474a504d3afee7ccc95be68dabbd41006.

taj-p: In either case, I've created https://github.com/taj-p/vello/pull/5 as a fast follow up on this PR which applies font size scaling with the suggested glyph position scaling. Perhaps we can continue this discussion there outside of this PR?

sparse_strips/vello_hybrid/examples/text.rs
taj-p: Updated example to use Parley in https://github.com/linebender/vello/pull/883/commits/7fdd1afe4ed17f2ba6b358f329c23f05333bb8cc

sparse_strips/vello_hybrid/examples/text.rs
taj-p: I ran out of time this morning to finish this. I've put it on my list of things to do.

sparse_strips/vello_hybrid/examples/text.rs
taj-p: Agree! Updating winit can be a pain! After https://github.com/taj-p/vello/pull/5, I'll refactor Vello Hybrid examples to instead use 1 winit example that cycles through a set number of scenes (while being able to pass flags to load a particular scene on start) (similar to how the root `examples/scenes` crate works) - WDYT @grebmeg ?

sparse_strips/vello_common/src/glyph.rs
DJMcNab: Right. I think I see.

It would take me a bit of time to carefully reason about this. But if it works, then I don't really think I need to do that!

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,235 @@
...
+enum Style {
+    Fill,
+    Stroke,
+}
+
+/// A sequence of glyphs with shared rendering properties.
+#[derive(Clone, Debug)]
+struct GlyphRun<'a> {
+    /// Font for all glyphs in the run.
+    pub font: Font,
```
DJMcNab, line 164: It's not immediately obvious to me why the fields of this private struct are public. But OTOH, if clippy isn't complaining about it, that's probably my lack of understanding here.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -0,0 +1,235 @@
...
+            skrifa::FontRef::from_index(self.run.font.data.as_ref(), self.run.font.index).unwrap();
+        let outlines = font.outline_glyphs();
+        let size = Size::new(self.run.font_size);
+        let hinting_instance = if self.run.hint {
+            // Only transformations including uniform scaling and translation can be hinted.
+            // Rotated, skewed, and other transformations cannot be hinted.
+            let [a, b, c, d, _, _] = self.run.transform.as_coeffs();
+            if a == d && b == 0.0 && c == 0.0 {
+                // TODO: Cache hinting instance.
+                HintingInstance::new(&outlines, size, self.run.normalized_coords, HINTING_OPTIONS)
```
DJMcNab, line 111: So just to really lay this out, currently we scale glyphs post-hinting. Which as an MVP is *fine*, but it is definitely wrong behaviour (because you're not really hinting unless you have a scale of exactly 1.0!).

We should have a comment here, probably.

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,260 @@
...
+                device_handle.device.poll(wgpu::Maintain::Poll);
+            }
+            _ => {}
+        }
+    }
+}
+
+fn draw_text(
+    ctx: &mut Scene,
+    text: String,
```
DJMcNab, line 206: Should this be a &str?

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,260 @@
...
+    font_cx: &mut FontContext,
+    layout_cx: &mut LayoutContext<ColorBrush>,
+) {
+    let mut builder = layout_cx.ranged_builder(font_cx, &text, 1.0);
+    builder.push_default(FontFamily::parse("Roboto").unwrap());
+    builder.push_default(StyleProperty::LineHeight(1.3));
+    builder.push_default(StyleProperty::FontSize(32.0));
+
+    let mut layout: Layout<ColorBrush> = builder.build(&text);
```
DJMcNab, line 215: In theory, we should be caching the layout. I don't want to block this PR on that, though

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+    });
+
+    let run = glyph_run.run();
+    let font = run.font();
+    let font_size = run.font_size();
+    let normalized_coords = run
+        .normalized_coords()
+        .iter()
+        .map(|coord| NormalizedCoord::from_bits(*coord))
```
DJMcNab, line 250: This is where we'd use `bytemuck::cast_slice`, I think.

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,260 @@
...
+    let mut app = App {
+        context: RenderContext::new(),
+        font_cx: FontContext::new(),
+        layout_cx: LayoutContext::new(),
+        renderers: vec![],
+        state: RenderState::Suspended(None),
+        scene: Scene::new(900, 600),
+    };
+
+    app.font_cx.collection.register_fonts(ROBOTO_FONT.to_vec());
```
DJMcNab, line 51: One of "always using roboto" and collecting system fonts is unnecessary.

In terms of teaching how we expect people to use this, I'd probably lean towards only using system fonts (acknowledging the difficulty that has with web support - but this example doesn't yet run on the web)

sparse_strips/vello_hybrid/examples/text.rs
```diff
@@ -0,0 +1,260 @@
...
+use vello_hybrid::{RenderParams, Renderer, Scene};
+use wgpu::RenderPassDescriptor;
+use winit::{
+    application::ApplicationHandler,
+    event::WindowEvent,
+    event_loop::{ActiveEventLoop, EventLoop},
+    window::{Window, WindowId},
+};
+
+const ROBOTO_FONT: &[u8] = include_bytes!("../../../examples/assets/roboto/Roboto-Regular.ttf");
```
DJMcNab, line 28: I agree, not dealing with variable fonts now is right

DJMcNab (approved):
A few nits, but otherwise I think this is ready.

raphlinus:
Thanks for all the comments in review. One point I'll make (based on the initial comments) is that emoji is not *necessarily* bitmap based, although that will be the case on Apple platforms. The COLRv1 format is outline based with a fairly sophisticated imaging model. It should be possible to port Vello's implementation (which lives mostly in vello/src/scene.rs).

sparse_strips/vello_common/src/glyph.rs
taj-p: Nice catch! Made private in https://github.com/linebender/vello/pull/883/commits/6e661a49209423ebbbeb76ea79c1f97204380907

sparse_strips/vello_common/src/glyph.rs
taj-p: Good callout. Added comment in https://github.com/linebender/vello/pull/883/commits/8284434061acb5f6f89c26c9daa014c88d3169be. I also only applied hinting when scale is 1.0. Immediately after this, I'll prepare https://github.com/taj-p/vello/pull/5 for review.

sparse_strips/vello_hybrid/examples/text.rs
taj-p: Made &str in https://github.com/linebender/vello/pull/883/commits/3d6735d75224003bdeb1d333121d90d7b11cd02d

sparse_strips/vello_hybrid/examples/text.rs
taj-p: https://github.com/linebender/vello/pull/883/commits/72ac1457a9598ef95144eea09747b3fc73b3e9db

sparse_strips/vello_hybrid/examples/text.rs
taj-p: Use system fonts in https://github.com/linebender/vello/pull/883/commits/9ab73412723bd9336585ba84700c101528aeefad and https://github.com/linebender/vello/pull/883/commits/d4d5658cb7151e7f4a7e580484ca2f3401a52c0c

taj-p: Ah, it fails CI when we try to build the `vello_hybrid` crate to Wasm. Instead of opting `vello_hybrid` out of Wasm build, I've reverted these changes. I will implement them as part of this [idea](https://github.com/linebender/vello/pull/883#discussion_r2023744690) next.

sparse_strips/vello_hybrid/examples/text.rs
grebmeg: 💯 yes — thank you!

PR #886, opened 2025-04-02
Remove the SDL2 example

DJMcNab:
If someone cares to work out why it doesn't compile and then brings it back, then they can do. 

But I don't care to, and we need CI to be working.


DJMcNab:
FYI @TheNachoBIT

waywardmonkeys:
Easier to just replace SDL2 with SDL3! See PR #887.

DJMcNab:
The interim plan is to land this as-is (i.e. only removing it from the workspace), then land #887 at our leisure.

PR #887, opened 2025-04-02
Replace `simple_sdl2` with `simple_sdl3`

waywardmonkeys:
SDL2 has an issue in CI with newer versions of cmake and instead of figuring out how to fix that, we can just use SDL3 now since it has been released.


taj-p:
@waywardmonkeys - does this need a review, is there still work to be done, or can it be closed?

PR #891, opened 2025-04-03
[vello_cpu] Refactor the test harness

LaurenzV:
See [#vello > Test reorganization](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Test.20reorganization/with/509872511)


sparse_strips/vello_cpu/tests/mod.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![allow(missing_docs, reason = "we don't need docs for testing")]
+#![allow(clippy::cast_possible_truncation, reason = "not critical for testing")]
+
```
DJMcNab, line 6: We should probably have a tiny bit of docs here. E.g. mentioning that `util` is for shared things.

(There are arguments for splitting `util` into e.g. `harness` and something else, say `shared`, `content`, or indeed `util`)

DJMcNab (approved):
This is even cleaner than I thought it would be.

PR #892, opened 2025-04-03
Refactor Snapshot and Generated Image Handling

spirali:
# Refactor Snapshot and Generated Image Handling

This pull request changes how snapshots and currently generated images are handled in tests.

1. It moves `smoke_snapshots` to `snapshots/smoke` and original `snapshots` to `snapshots/lfs`.

2. Instead of generating images into the same directory as snapshots, it generates images into a separate directory `current`.

3. In comparison tests, it generates CPU/GPU tests in separate directories.

## Example (Snapshot Tests)

Original layout:

* `smoke_snapshots/xxx.png` (reference)
* `smoke_snapshots/xxx.cpu.png` (CPU generated)
* `smoke_snapshots/xxx.gpu.png` (GPU generated)

New layout:

* `snapshots/smoke/xxx.png` (reference)
* `current/cpu/smoke/xxx.png` (CPU generated)
* `current/gpu/smoke/xxx.png` (GPU generated)

## Example (Comparison Tests)

Original layout:

* `comparison/xxx.cpu.png`
* `comparison/xxx.gpu.png`

New layout:

* `comparison/cpu/xxx.png`
* `comparison/gpu/xxx.png`

## Why?

The important property of this change is that files to be compared have the same filenames in the same nested directory structure. This allows for simple directory-to-directory comparisons for all tests at once. For example, `current/cpu` vs `snapshots`, or `comparison/cpu` vs `comparison/gpu`.

This is a preparatory step for full Kompari integration. However, this layout allows immediate use of `kompari-cli` to generate diff reports.

Note: This PR lefts `debug_outputs` as it is.


DJMcNab:
I'll review on Monday. Just a passing thought that the main we thing we need to be sure of is that a `oversized.png` never ends up in the `snapshots` directory.

DJMcNab (approved):
This looks good, except for the one thing about oversized images. However, two things are in our favour there:
1) The test will never succeed if it generates an oversized file.
2) The snapshot sizes are still checked on CI

I think the quickest course of action to resolve is to just re-add `*.oversized.png` to the `snapshots/gitignore`.

DJMcNab (approved):

PR #885, opened 2025-04-01
Add basic support for radial, linear and conic gradients

LaurenzV:
Not done yet and not ready for review (radial gradients as well as code clean up & documentation still missing). For now, I will also only be targeting correctness and completely ignore good performance.

PR #893, opened 2025-04-03
Add support for linear, radial and sweep gradients

LaurenzV:
Superserdes #885 to reduce the clutter of commits I produced. 😅 

A first version of linear, radial and sweep gradients. Has support for all important attributes on gradients as well as arbitrary affine transforms.

As mentioned before, performance has not been my major focus for this, but instead correctness. There probably are some easy (and some harder) approaches for improving performance, if you have thoughts on this feel free to share them, but for now I would like to land this more or less in the current shape and defer optimizations to later.


sparse_strips/vello_cpu/src/fine/gradient.rs
```diff
@@ -0,0 +1,136 @@
...
+        let pad = self.gradient.pad;
+        let extend = |val| extend(val, pad, self.gradient.clamp_range);
+        let mut pos = self.cur_pos;
+
+        for pixel in col.chunks_exact_mut(COLOR_COMPONENTS) {
+            let dist = extend(self.kind.cur_pos(&pos));
+            self.advance(dist);
+            let range = self.cur_range;
+
+            for (col_idx, comp) in pixel.iter_mut().enumerate() {
```
dominikh, line 86: Here, `col_idx` is the index of the color channel, right? I'd probably pick a different name, to avoid confusion with columns.

sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -93,53 +124,141 @@ impl<'a> Fine<'a> {
...
-        match paint {
-            Paint::Solid(c) => {
-                let color = c.to_u8_array();
-
-                let target =
-                    &mut scratch[x * TILE_HEIGHT_COMPONENTS..][..TILE_HEIGHT_COMPONENTS * width];
+    pub fn fill(
+        &mut self,
+        x: usize,
+        tile_x: u16,
```
dominikh, line 130: `fill`, `strip` (and thus `run_cmd`), and `pack` all need to know the tile coordinates now. Might be worth storing in `fine` instead?

sparse_strips/vello_cpu/tests/gradient.rs
```diff
@@ -0,0 +1,1229 @@
...
+        opacity: 0.5,
+    };
+
+    ctx.set_paint(gradient);
+    ctx.fill_rect(&rect);
+
+    check_ref(&ctx, "gradient_with_global_alpha");
+}
+
+mod linear {
```
dominikh, line 57: Let's add a test case for a purely vertical linear gradient, e.g. from (0, 0) to (0, 90)

sparse_strips/vello_cpu/snapshots/gradient_radial_center_offset_top_right.png
dominikh, line 1: That's a rough looking circle.

sparse_strips/vello_cpu/src/fine/gradient.rs
```diff
@@ -0,0 +1,136 @@
...
+            kind,
+        }
+    }
+
+    fn advance(&mut self, target_pos: f32) {
+        while target_pos > self.cur_range.x1 || target_pos < self.cur_range.x0 {
+            if self.range_idx == 0 {
+                self.range_idx = self.gradient.ranges.len() - 1;
+            } else {
+                self.range_idx -= 1;
```
grebmeg, line 45: I’m curious about the reasoning behind searching backward through the gradient ranges — was it for performance, correctness, or some edge case handling?

sparse_strips/vello_cpu/snapshots/gradient_radial_center_offset_top_right.png
grebmeg, line 1: Just to double-check — is the heavy pixelation at the pattern borders expected in the test cases with offsets and `Extend::Repeat` for radial gradients?

sparse_strips/vello_cpu/tests/gradient.rs
```diff
@@ -0,0 +1,1206 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use vello_common::kurbo::{Affine, Point, Rect};
+use vello_common::peniko::GradientKind;
+use vello_cpu::paint::Gradient;
+use crate::util::{check_ref, get_ctx, stops_blue_green_red_yellow, stops_green_blue};
+
+pub(crate) fn tan_45() -> f64 {
+    45.0_f64.to_radians().tan()
+}
```
dominikh, line 11: Isn't this just `1`?

grebmeg (commented):
Nice work! 🔥 This is a very clever set of optimizations with a unified approach for handling various gradient types. As you mentioned, performance hasn’t been your main focus at this stage, so I haven’t looked at that aspect too closely — though I believe there are areas that could be optimized further down the line.

For now, the implementation looks solid, and the snapshot coverage is great, so I’d be happy to approve this PR. That said, it might be good to bring it up during the next office hours — just to make sure no one else is actively reviewing or planning to.

One question I do have (or perhaps a concern) is around how we plan to evolve both renderers. Specifically, it seems like parts of this PR (for example, `paint.rs`) could potentially live in `vello_common` so that gradient handling can be shared with `vello_hybrid`? Would be great to clarify the direction there.

sparse_strips/vello_cpu/snapshots/gradient_radial_center_offset_top_right.png
LaurenzV: I think that's simply just because there is no anti-aliasing. Here is a very similar version made with Blend2D:
blend2d_radial_gradient_offset

sparse_strips/vello_cpu/tests/gradient.rs
LaurenzV: Yeah. :p Though I do feel the intent is clearer when writing it this way.

sparse_strips/vello_cpu/src/paint.rs
```diff
@@ -0,0 +1,844 @@
...
+const DEGENERATE_THRESHOLD: f32 = 1.0e-6;
+
+/// A kind of paint used for drawing shapes.
+#[derive(Debug, Clone)]
+pub enum PaintType {
+    /// A solid color.
+    Solid(AlphaColor<Srgb>),
+    /// A gradient.
+    Gradient(Gradient),
+}
```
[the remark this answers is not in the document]
LaurenzV: I didn't use the `Brush` type because it's missing some things. For gradients, it doesn't have a transform attribute and it also has something called "hue direction"  which we don't support. For patterns, it's even more of a problem because it doesn't store a pixmap. So for now, it's better to use our own types, I think.

Any suggestions for a rename?

sparse_strips/vello_cpu/src/paint.rs
```diff
@@ -0,0 +1,844 @@
...
+            clamp_range,
+        };
+
+        let idx = paints.len();
+        paints.push(encoded.into());
+
+        Paint::Indexed(IndexedPaint::new(idx))
+    }
+
+    fn validate(&self) -> Option<Paint> {
```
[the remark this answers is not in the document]
LaurenzV: Fair enough, I changed it so that it always returns the first stop's color, whenever possible. That was a bit arbitrary.

sparse_strips/vello_cpu/src/paint.rs
```diff
@@ -0,0 +1,844 @@
...
+                // This corresponds to (y_2 - y_1) in the formula, but because of the above reasons
+                // shortens to y_2.
+                let y2_minus_y1 = norm.1;
+                // This corresponds to (x_2 - x_1) in the formula, but because of the above reasons
+                // shortens to x_2.
+                let x2_minus_x1 = norm.0;
+                // Note that we can completely disregard the x_2 * y_1 - y_2 * x_1 factor, since
+                // y_1 and x_1 are both 0.
+
+                let end_val = (dx * dx + dy * dy).sqrt();
```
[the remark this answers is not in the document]
LaurenzV: Yes, it will be equal to the distance between p0 and p1, which is also the same length as the norm vector.

sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -93,53 +109,131 @@ impl<'a> Fine<'a> {
...
+                                blend_buf,
+                                color_buf.chunks_exact(4).map(|e| [e[0], e[1], e[2], e[3]]),
+                            );
+                        } else {
+                            // Similarly to solid colors we can just override the previous values
+                            // if all colors in the gradient are fully opaque.
+                            $iter.run(blend_buf);
+                        }
+                    };
+                }
```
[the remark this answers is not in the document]
LaurenzV: I think here I did it because you can't have generics in a closure (necessary for the iterators), and if I create a function I have to pass the blend_buf and color_buf explicitly. But I think I can change it if you want.

sparse_strips/vello_cpu/snapshots/gradient_radial_center_offset_top_right.png
LaurenzV: Yes I would say that's expected, because there is not anti-aliasing between gradient stops, and since the resolution is so low it becomes very noticeable.

sparse_strips/vello_cpu/src/fine/mod.rs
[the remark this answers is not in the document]
DJMcNab: I think passing `blend_buf` and `color_buf` explicitly would be better. Using macros here hides that those variables are being modified at the call site. The extra indentation forced by this being a macro also isn't great.
There are use cases where macros make a big difference, but if a macro is acting like a function, it probably should be a function.

(This is a comment made entirely locally that is, I've not looked into the details of the types: can this be a method on `GradientFiller`? It seems to already have `g`, which means it should have `g.has_opacities`)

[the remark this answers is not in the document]
LaurenzV: Something I just realized is that I will need a similar method for images, in which case I have to duplicate the logic, while with a macro I can just reuse it (as long as the methods are named the same).

sparse_strips/vello_cpu/src/paint.rs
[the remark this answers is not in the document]
raphlinus: Ideally we do use the `Brush` type for this. Ultimately we do want to support hue direction, and I think transform might be something to pass in separately. Having duplicated types for this in the ecosystem is not great.

raphlinus (commented):
I had a look. Generally it's good, and I think the gradient range approach is a good one, to somewhat decouple the input semantics from the rendering details.

As I mentioned, I think it would be better to use peniko types if possible. There are some significant difference, especially peniko supports different color spaces (this is why hue direction is important, it affects the results in cylindrical color spaces). It's possible that can be followup work, but I also think it might be less painful to get at least the types correct early (ie, just converting to Srgb as opposed to calling the gradient iterator in the `color` crate, which incidentally was meant for exactly this use case).

Like Daniel, I'm not a huge fan of macros where they can be avoided. I didn't analyze it carefully, but I have a feeling they could be replaced by generic code, possibly at the expense of making the types more complicated.

I'm ok with this landing and making further improvements as followons, especially to keep momentum going with images.

LaurenzV:
> As I mentioned, I think it would be better to use peniko types if possible. 

I definitely agree with that in the long-term, but I think it would be easier to use the custom types now. I'm not really sure how you would pass the transform extra in a nice way (in tiny-skia and Blend2D the transform is also part of the gradient itself), and especially since the `Brush` type won't work for images at all at the moment, I'm not sure there is much to gain now. I do reuse other parts of peniko (for example we reuse the `GradientKind`), so the API divergence isn't that big if we change it in the future. :)

But I noticed that there also is the `ColorStops` type which I'm not using, would it be okay with you if I try to integrate that but keep the transform thing as is for now?

> Like Daniel, I'm not a huge fan of macros where they can be avoided. I didn't analyze it carefully, but I have a feeling they could be replaced by generic code, possibly at the expense of making the types more complicated.

I will take another look at that!

raphlinus:
> But I noticed that there also is the `ColorStops` type which I'm not using, would it be okay with you if I try to integrate that but keep the transform thing as is for now?

Yes, that seems like an improvement.

sparse_strips/vello_cpu/src/paint.rs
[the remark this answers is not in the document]
LaurenzV: Alternatively, could we make a "flash release"  of peniko to add a `transform` attribute? This way we could at least use the type for gradients.

sparse_strips/vello_cpu/tests/gradient.rs
tomcur: It may be better to return `1.`.  This expression evaluates to one ULP below 1 (`0.9999999999999999`). It may also be helpful to mark this function `const`.

LaurenzV:
Okay, so:
1) I switched to using `ColorStops` from peniko.
2) I moved the gradient API to `vello_api` and the encoding to `vello_common`.
3) I replaced the macros in fine rasterization with generic methods.

Let me know if it's good now. 🙂

sparse_strips/vello_cpu/src/paint.rs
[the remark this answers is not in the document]
waywardmonkeys: I am planning to do a breaking release of `color` and `peniko` this month, so if there's interest in this happening, let's track that and see if that's something that we want to include. If we do include it, we should consider how "classic Vello" would handle it. I don't have any opinion here, just want to facilitate getting the new batch of releases out.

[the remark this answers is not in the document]
LaurenzV: Yeah that would definitely be great, though I don't think we should block the PR on this now. I guess the `transform` attribute would just be ignored in vello for now?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -642,24 +642,27 @@ impl WideTile {
...
-            let Paint::Solid(s) = &paint else {
-                unimplemented!()
+            let bg = if let Paint::Solid(s) = &paint {
+                // Note that we could be more aggressive in optimizing a whole-tile opaque fill
+                // even with a clip stack. It would be valid to elide all drawing commands from
+                // the enclosing clip push up to the fill. Further, we could extend the clip
+                // push command to include a background color, rather than always starting with
+                // a transparent buffer. Lastly, a sequence of push(bg); strip/fill; pop could
+                // be replaced with strip/fill with the color (the latter is true even with a
+                // non-opaque color).
```
DJMcNab, line 652: To clarify, this final optimisation is for the case where there is no "path" set within the clip stack?
Does the previous active path still apply?

This is probably me not fully grokking the current pipeline.

Edit: I see that this comment is just copied as-is from the previous

sparse_strips/vello_common/src/encode.rs
```diff
@@ -0,0 +1,819 @@
...
+use std::borrow::Cow;
+use std::f32::consts::PI;
+use std::iter;
+use vello_api::paint::{Gradient, IndexedPaint, Paint};
+
+const DEGENERATE_THRESHOLD: f32 = 1.0e-6;
+const NUDGE_VAL: f32 = 1.0e-7;
+
+/// A trait for encoding gradients.
+pub trait Encode {
```
DJMcNab, line 20: Can we call this `EncodeExt` or `GradientExt`? I'd also consider making it sealed.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -0,0 +1,819 @@
...
+        let idx = paints.len();
+        paints.push(encoded.into());
+
+        Paint::Indexed(IndexedPaint::new(idx))
+    }
+}
+
+/// Returns a fallback paint in case the gradient is invalid. The paint will be either
+/// black or contain the color of the first stop of the gradient.
+fn validate(gradient: &Gradient) -> Option<Paint> {
```
DJMcNab, line 250:
```suggestion
fn validate(gradient: &Gradient) -> Result<(), Paint> {
```
would be a more idiomatic signature

DJMcNab (commented):
We do need to get this landed. I don't really have the expertise to meaningfully review this.

We can bring this up in office hours tomorrow, unless @grebmeg is willing and able to review it before then.

LaurenzV:
> We do need to get this landed.

Yeah, it would be great so that I can put up the images PR before the next office hours, but let's see whether that works out time-wise.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Yep, I didn't write this one. 😄

sparse_strips/vello_api/src/paint.rs
```diff
@@ -42,3 +44,43 @@ impl From<AlphaColor<Srgb>> for Paint {
...
+    /// Note that the first stop must have an offset of 0.0 and the last stop
+    /// must have an offset of 1.0. In addition to that, the stops must be sorted
+    /// with offsets in ascending order.
+    pub stops: ColorStops,
+    /// A transformation to apply to the gradient.
+    pub transform: Affine,
+    /// The extend of the gradient.
+    pub extend: peniko::Extend,
+    /// An additional opacity to apply to all gradient stops.
+    pub opacity: f32,
```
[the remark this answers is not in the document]
LaurenzV: Yes, that's a good idea.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -42,3 +44,43 @@ impl From<AlphaColor<Srgb>> for Paint {
...
+}
+
+/// A kind of paint used for drawing shapes.
+#[derive(Debug, Clone)]
+pub enum PaintType {
```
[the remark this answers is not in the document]
LaurenzV: I've added some documentation to `Paint`, I'm not sure what more to write about `PaintType`, though. :p Arguably the naming of those two is also not great, but I think that's something that can be refined.

grebmeg (approved):

PR #895, opened 2025-04-04
fix(vello_hybrid): update config on resize

grebmeg:
`velly_hybrid` resizing issue was caused by not updating `Config` when the viewport size changed. As a result, the vertex positions in the shader were being calculated based on the previous config, which led to distorted rendered content.

Before:

https://github.com/user-attachments/assets/561e0bb3-408d-4afc-8946-48bc3e98db15

After:

https://github.com/user-attachments/assets/51ac9591-fa67-4e2d-aa66-dff746680547

https://github.com/user-attachments/assets/963497cf-7061-47d8-ba8a-4723e456d4b5

Images in the videos may flicker during resizing because of the problem seems related to wpu (more details in [#vello > Vello Hybrid: Resizing @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Vello.20Hybrid.3A.20Resizing/near/510114878)):

https://github.com/user-attachments/assets/7feeb008-8f97-42c0-9ab5-da6eb53dec75

### DIFF
--- a/sparse_strips/vello_hybrid/src/render.rs
+++ b/sparse_strips/vello_hybrid/src/render.rs
@@ -23,9 +23,9 @@ use wgpu::{
 
 use crate::scene::Scene;
 
-/// Parameters for the renderer
-#[derive(Debug)]
-pub struct RenderParams {
+/// Dimensions of the rendering target
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct RenderSize {
     /// Width of the rendering target
     pub width: u32,
     /// Height of the rendering target
@@ -34,9 +34,13 @@ pub struct RenderParams {
 
 /// Options for the renderer
 #[derive(Debug)]
-pub struct RendererOptions {
+pub struct RenderTargetConfig {
     /// Format of the rendering target
     pub format: wgpu::TextureFormat,
+    /// Width of the rendering target
+    pub width: u32,
+    /// Height of the rendering target
+    pub height: u32,
 }
 
 /// Contains all GPU resources needed for rendering
@@ -48,6 +52,8 @@ struct GpuResources {
     pub alphas_texture: Texture,
     /// Bind group for rendering
     pub render_bind_group: BindGroup,
+    /// Buffer for config data
+    pub config_buffer: Buffer,
 }
 
 /// GPU renderer for the hybrid rendering system
@@ -62,6 +68,9 @@ pub struct Renderer {
 
     /// Scratch buffer for staging alpha texture data.
     alpha_data: Vec<u8>,
+
+    /// Dimensions of the rendering target
+    render_size: RenderSize,
 }
 
 /// Contains the data needed for rendering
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -306,25 +323,44 @@ impl Renderer {
...
             };
             self.resources = Some(GpuResources {
                 strips_buffer,
                 alphas_texture,
                 render_bind_group,
+                config_buffer,
             });
         };
 
+        if dimensions_changed {
```
DJMcNab, line 347: Does this path still trigger if the config buffer was recreated?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -285,15 +302,15 @@ impl Renderer {
...
                     contents: bytemuck::bytes_of(&Config {
-                        width: render_params.width,
-                        height: render_params.height,
+                        width: new_render_size.width,
+                        height: new_render_size.height,
                         strip_height: Tile::HEIGHT.into(),
                         alphas_tex_width_bits: max_texture_dimension_2d.trailing_zeros(),
                     }),
-                    usage: wgpu::BufferUsages::UNIFORM,
+                    usage: wgpu::BufferUsages::UNIFORM | wgpu::BufferUsages::COPY_DST,
```
DJMcNab, line 313: This is recreated even if there is an already extant config buffer, right?

DJMcNab (approved):
It works, and this buffer lifecycle management is something which we already know needs to change, so from that perspective I'm happy for this to land.

sparse_strips/vello_hybrid/src/render.rs
grebmeg: Thanks for both comments! You were right — those two places were out of sync. The config was being recreated when the alpha texture was regenerated, which then triggered writing the same data to the buffer again. I’ve refactored the code a bit and updated the config management to match how the other textures are managed.

Fixed in 0e5a4d09

sparse_strips/vello_hybrid/src/render.rs
grebmeg: Check the comment above for more details.

Fixed in 0e5a4d09

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -200,6 +209,10 @@ impl Renderer {
             render_pipeline,
             resources: None,
             alpha_data: Vec::new(),
+            render_size: RenderSize {
+                width: render_target_config.width,
+                height: render_target_config.height,
```
DJMcNab, line 214: I'm not sure how much value this adds - these values never be used (because when they would have been used, the render resources will have been created for the first time anyway). I think it would be just as valid for this to be any values (including probably 0x0). If this analysis is right, it being on the `RenderTargetOptions` is misleading.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -306,25 +333,44 @@ impl Renderer {
...
                 strips_buffer,
                 alphas_texture,
                 render_bind_group,
+                config_buffer,
             });
         };
 
+        // Update config buffer if dimensions changed and config buffer exists.
+        // We don't need to initialize a new config buffer because it's fixed size (uniform buffer).
+        if needs_new_config && self.resources.is_some() {
```
DJMcNab, line 359: How would `self.resources` ever be `None` here? We unwrap it in the next statement in this block anyway.
It looks to me that `needs_new_config` will still be true even if the buffer was just created (although I do note that this is only an issue on the first frame, so isn't a huge concern).

PR #896, opened 2025-04-04
When hinting glyphs, round y-offsets to integers

tomcur:
Without this, while glyph outlines will be hinted to integers, the translation can bump the outlines to fractional locations again.

### DIFF
--- a/sparse_strips/vello_common/src/glyph.rs
+++ b/sparse_strips/vello_common/src/glyph.rs
@@ -143,6 +143,8 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
             let mut local_transform = {
                 let mut translation = Vec2::new(glyph.x as f64 * scale, glyph.y as f64 * scale);
                 if hinting_instance.is_some() {
+                    // When hinting, ensure the y-offset is integer. The x-offset doesn't matter,
+                    // as we perform vertical-only hinting.
                     translation.y = translation.y.round();
                 }
                 Affine::translate(translation)
--- a/sparse_strips/vello_common/src/glyph.rs
+++ b/sparse_strips/vello_common/src/glyph.rs
@@ -140,15 +140,8 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
                 continue;
             }
 
-            let mut local_transform = {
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This makes me think of rounding stuff in Parley recently.

tomcur:
> This makes me think of rounding stuff in Parley recently.

This one? https://github.com/linebender/parley/pull/297

waywardmonkeys:
Yes, sir.

DJMcNab (commented):
I don't know if we have any good docs for this, but https://docs.google.com/document/d/1gEqf7ehTzd89Djf_VpkL0B_Fb15e0w5fuv_UzyacAPU/edit?disco=AAABctut6QY is relevant (it's a link to a specific comment) here.

I don't think this is quite the right place to be doing this quantisation; we would need to be rounding the absolute y coordinate, not the "local" y coordinate.

Separately: ideally, there's be some way to warn if this changed frame-by-frame (that is, the baseline isn't rounded). But I don't think there's any way to detect that, because the relationship between `y` and the baseline is not so simple.

tomcur:
> I don't think this is quite the right place to be doing this quantisation; we would need to be rounding the absolute y coordinate, not the "local" y coordinate.

Hmm... I guess you're right, in that depending on `GlyphRenderer`, we don't quite know where the outline is going to be placed. I believe currently using `vello_cpu`, the `y`-coordinate here is the absolute coordinate.

DJMcNab:
> Hmm... I guess you're right, in that depending on GlyphRenderer, we don't quite know where the outline is going to be placed. I believe currently using vello_cpu, the y-coordinate here is the absolute coordinate.

Is it not impacted by the "adding" of the transform on (new) line 163?

tomcur:
> Is it not impacted by the "adding" of the transform on (new) line 163?

I was going to write it is not, but on second thought it is partially impacted. The order of operations is effectively:

```
transform * translate * skew * glyph_transform
```

(Where you have to read `b * a` as `b after a`.) In this code-path `transform` is guaranteed to have a scale of `1` because `take_uniform_scale` has taken the scale out of `transform`.

However, if `transform` itself translates with fractional coordinates, the outline can still be bumped to a fractional coordinate. We can account for that here.

DJMcNab (approved):

taj-p (approved):
LGTM

PR #897, opened 2025-04-04
Generalize glyph hinting: allow glyph transforms and global skews

tomcur:
On top of https://github.com/linebender/vello/pull/896 (that should land first). See c0b778693cc19d4c3520c9313bd46d7d2641ce7f and onwards for the changes here.

This builds on the work landed in https://github.com/linebender/vello/pull/889.

This generalizes our detection of when (vertical-only) hinting can be performed. When the second coefficient (`b`) of the affine transformation matrix is 0 (first column, second row), glyphs are not skewed vertically or rotated, and can be hinted. Because of this generalization, the `horizontal_skew` special case has been removed.

We now calculate the total transformation matrix once, and separately calculate the glyph translations using the run's transformation matrix for global glyph positioning. This reduces the number of affine multiplies per glyph to one-and-a-bit down from three.


sparse_strips/vello_common/src/glyph.rs
```diff
@@ -140,20 +135,20 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
...
-            let mut local_transform =
-                Affine::translate(Vec2::new(glyph.x as f64 * scale, glyph.y as f64 * scale));
-            if let Some(skew) = horizontal_skew {
-                local_transform *= Affine::skew(skew.tan() as f64, 0.0);
-            }
-            if let Some(glyph_transform) = glyph_transform {
-                local_transform *= glyph_transform;
+            let mut translation = translation_transform
+                .pre_translate(Vec2::new(glyph.x as f64, glyph.y as f64))
+                .translation();
```
tomcur, line 140: With proper inlining, the compiler should be able to elide most float calculations here, but I've not verified that. We could also instead store the translation transformation affine as a 2x2 matrix instead of the full 3x2 `Affine` (the translation transform is just the global transform with translation set to 0), and manually calculate the following.

```
translate_x = a * glyph.x + c * glyph.y
translate_y = b * glyph.x + d * glyph.y
```

Thoughts?

tomcur: Actually, we can take the global transform from the run.

```rust
let [a, b, c, d, _, _] = self.run.transform.as_coeffs();
let mut translation = Vec2::new(
    a * glyph.x as f64 + c * glyph.y as f64,
    b * glyph.x as f64 + d * glyph.y as f64,
);
```

tomcur: I've applied this in 1b51ec0112f34528ea1032d43f14e4a864150aca.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -209,66 +205,50 @@ fn prepare_glyph_run<'a>(
...
+        let vertical_font_size = run.font_size * t_d as f32;
+        let size = Size::new(vertical_font_size);
         let hinting_instance =
             HintingInstance::new(outlines, size, run.normalized_coords, HINTING_OPTIONS).ok();
-
-        return PreparedGlyphRun {
-            transform,
-            glyph_transform: run.glyph_transform,
+        PreparedGlyphRun {
+            transform: Affine::new([1., t_b, t_c, 1., t_e, t_f]),
```
taj-p, line 240:
```suggestion
            transform: Affine::new([1., 0., t_c, 1., t_e, t_f]),
```

nit: Setting the second parameter to `0.` signals to the reader that, as per prior lines, we know what this value will end up being; the variables we do specify here are the ones that are variable.

taj-p (approved):
LGTM - thank you - this is so much better!

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -140,20 +134,30 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
...
+            // glyph.
+            let [a, b, c, d, _, _] = self.run.transform.as_coeffs();
+            let translation = Vec2::new(
+                a * glyph.x as f64 + c * glyph.y as f64,
+                b * glyph.x as f64 + d * glyph.y as f64,
+            );
+
+            // When hinting, ensure the y-offset is integer. The x-offset doesn't matter, as we
+            // perform vertical-only hinting.
+            let mut total_transform = transform.then_translate(translation).as_coeffs();
```
DJMcNab, line 151: This feels suspicious to me. In theory, we want to skew, then translate locally, then transform globally, right?
But this ordering seems to me to be translate locally, then skew, then transform globally.

In any case, we need a test which has a long-ish text run skewed using `glyph_transform`.

sparse_strips/vello_cpu/tests/glyph.rs
```diff
@@ -98,13 +98,53 @@ fn glyphs_skewed_unhinted() {
...
+            * Affine::skew(-20_f64.to_radians().tan(), 0.),
+    );
+    ctx.set_paint(REBECCA_PURPLE.with_alpha(0.5).into());
+    ctx.glyph_run(&font)
+        .font_size(font_size)
+        .glyph_transform(Affine::skew(20_f64.to_radians().tan(), 0.))
+        .hint(true)
+        .fill_glyphs(glyphs.into_iter());
+
+    check_ref(&ctx, "glyphs_skewed_unskewed");
```
DJMcNab, line 125: This rendering doesn't look right to me. If we're skewing each glyph individually, that should be a different operation to skewing the global box.
That is, this playground shows what I expect the rendering of this snapshot to be.
https://developer.mozilla.org/en-US/play?id=%2BlgWWSeKM7878QsZJVVId%2Bz0GMEv%2FqgYywTMjfjDnIW9K6nNfCtN5UPeH7%2FhgPFEQloDnLVzXIaKA7fG

(this is the same issue as my other comment)

DJMcNab (changes requested):

DJMcNab (approved):
My previous comments about needing more tests stands, but I have had the logic explained to me, and I largely believe it.

sparse_strips/vello_common/src/glyph.rs
tomcur: For posterity: we apply `glyph_transform`, then translate with the glyph position from shaping/layouting, then apply the global `transform`.

Mathematically that's `transform * glyph_translation * glyph_transform`, where all of these are 3x3 affine matrices.

`glyph_translation` only adds a translation. That means we can equivalently calculate `transform * glyph_transform`, and then add the (transformed) translation component from `glyph_translation` to the total transform. This order is slightly cheaper to calculate.

PR #898, opened 2025-04-05
Fix overflow in coarse rasterization

LaurenzV:
This can overflow for not too big sizes (I encountered it while trying to render on a canvas with 10000x10000 pixels.


tomcur (approved):
Thanks, makes sense!

PR #899, opened 2025-04-05
Be more `no_std`

waywardmonkeys:
Instead of having `std` in scope and on by default, be `no_std` by default and use `std` more explicitly.


waywardmonkeys:
This is a continuation of the changes that I've been making across the whole Linebender stack for all `no_std` support to be configured this way.

PR #900, opened 2025-04-06
Add fast path for stripping rectangles

LaurenzV:
This re-adds my previously implemented fast path for rectangles. In my test scene with 20000 rectangles, this reduces the rendering time from around 37ms to 19ms!

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -341,18 +341,23 @@ pub fn render_rect(
         rect.max_y().min(height as f64) as f32,
     );
 
-    let top_strip_idx = y0 as u16 / Tile::HEIGHT;
+    let top_strip_idx = (y0 as u16).max(0) / Tile::HEIGHT;
     let top_strip_y = top_strip_idx * Tile::HEIGHT;
-    let bottom_strip_idx = y1 as u16 / Tile::HEIGHT;
+    // In the wide tile generation stage, there is an assertion that all strips outside the
+    // viewport must have been culled, so we cull here.
+    //
+    // This index is inclusive, i.e. pixels at row `bottom_strip_idx` 
+    // are still part of the rectangle.
+    let bottom_strip_idx = (y1 as u16).min(height - 1) / Tile::HEIGHT;
     let bottom_strip_y = bottom_strip_idx * Tile::HEIGHT;
 
     let x0_floored = x0.floor();
     let x1_floored = x1.floor();
 
-    let x_start = x0_floored as u16;
+    let x_start = (x0_floored as u16).max(0);
     // Inclusive, i.e. the pixel at column `x_end` is the very right border (possibly only anti-aliased)
     // of the rectangle, which should still be stripped.
-    let x_end = x1_floored as u16;
+    let x_end = (x1_floored as u16).min(width - 1);
 
     // Calculate the vertical/horizontal coverage of a pixel, using a start
     // and end point. The area between the start and end point is considered to be
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -6,6 +6,7 @@
 use vello_api::peniko::Fill;
 
 use crate::flatten::Line;
+use crate::kurbo::Rect;
 use crate::tile::{Tile, Tiles};
 
 /// A strip.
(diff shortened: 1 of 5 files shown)


tomcur (commented):
An initial thought inline, I'll look more in-depth in the coming days.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -241,3 +258,22 @@ impl GlyphRenderer for RenderContext {
...
+
+        coeffs[1] != 0.0 || coeffs[2] != 0.0
+    }
+}
+
+fn transform_non_skewed_rect(rect: &Rect, affine: Affine) -> Rect {
+    let p1 = affine * Point::new(rect.x0, rect.y0);
+    let p2 = affine * Point::new(rect.x1, rect.y1);
+
+    Rect::from_points(p1, p2)
```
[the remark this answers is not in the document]
LaurenzV: Haven't checked what the compiler does, but I think the change makes sense!

tomcur (commented):
Most tests with rectangles call `RenderContext::fill_rect`, which now uses the specialized path for rectangles. It would be good to continue exercising the general case as well.

I don't know what the best solution looks like. Perhaps performing multiple tests against the same snapshot could work, though we'd have to be robust against small rounding errors.

LaurenzV:
I've added a few more test cases which check against the same snapshot as the rectangle case, I feel like this should already be an improvement? As you mentioned, there are few other test cases with rectangles, but there's also only so many ways to draw a rectangle. :P So I think this should already be a good coverage.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -302,3 +303,186 @@ pub fn render(
...
+    alpha_buf: &mut Vec<u8>,
+    width: u16,
+    height: u16,
+) {
+    // The idea for this fast path is as follows:
+    // - We generate strips of width 1 for the left as well as the right side of the rectangle. The
+    //   left side has a winding number of 0, the right side has a winding number of 1.
+    // - We generate a strip of the full rectangle width for the top and bottom part of the rectangle.
+    // - Of course, it's also possible that a rectangle has a width of less than 2 or a height of
+    //   less than 4. The current logic does account for those edge cases.
```
taj-p, line 322: Should this disclaimer form part of the function documentation? Is it up to the consumer to use the general case instead of this fast path for these small rects? Do we want to add a debug assert into this function to ensure that and should we not call this fast path in render context for those small rectangles?

sparse_strips/vello_cpu/tests/basic.rs
```diff
@@ -317,6 +317,17 @@ fn filled_unaligned_rect() {
...
+    let rect = Rect::new(1.5, 1.5, 28.5, 18.5).to_path(0.1);
+
+    ctx.set_paint(REBECCA_PURPLE.with_alpha(0.5).into());
+    ctx.fill_path(&rect);
+
+    check_ref(&ctx, "filled_unaligned_rect_as_path");
+}
+
 #[test]
 fn filled_transformed_rect_1() {
```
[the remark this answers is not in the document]
taj-p: I see you've added more test cases since this comment. I do wonder whether future test cases will continue to exercise the fast path (since it's so easy to create rects compared to other shapes). I defer to your best judgement here!

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -246,17 +263,40 @@ impl GlyphRenderer for RenderContext {
     }
 }
 
+trait AffineExt {
+    fn has_skew(&self) -> bool;
+}
+
+impl AffineExt for Affine {
+    fn has_skew(&self) -> bool {
```
taj-p, line 271: I'm no geometry expert, but when I think of skew I generally think of shearing (i.e. the shape becoming distorted somehow by one or more axes sliding against a fixed axis). But, if I'm reading this correctly, we're also testing for no rotation. WDYT of making this (inversely) `.is_axis_aligned` or (explicit) `.has_rotation_or_skew` or something else?

taj-p (commented):
Nice! This was a great read! Certainly adding a fast path for rects makes sense and I can see how it is so much faster.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -241,3 +258,25 @@ impl GlyphRenderer for RenderContext {
...
+
+fn transform_non_skewed_rect(rect: &Rect, affine: Affine) -> Rect {
+    debug_assert!(
+        !affine.has_skew(),
+        "this method should only be called with non-skewing transforms"
+    );
+    let [a, _, _, d, _, _] = affine.as_coeffs();
+
+    Rect::new(a * rect.x0, d * rect.y0, a * rect.x1, d * rect.y1) + affine.translation()
+}
```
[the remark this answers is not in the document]
LaurenzV: Yeah, sounds reasonable!

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -202,6 +207,18 @@ impl RenderContext {
         self.wide.generate(&self.strip_buf, fill_rule, paint);
     }
 
+    fn render_rect(&mut self, rect: &Rect, paint: Paint) {
```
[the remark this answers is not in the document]
LaurenzV: It would definitely be nice to keep it in lock step, I can see whether I can add it for this case. The main problem is that I know 0 about GPU programming, so for other things it might be harder for me to port them.

sparse_strips/vello_cpu/tests/basic.rs
```diff
@@ -438,6 +449,17 @@ fn strip_inscribed_rect() {
     check_ref(&ctx, "strip_inscribed_rect");
 }
 
+#[test]
+fn filled_vertical_hairline_rect_as_path() {
+    let mut ctx = get_ctx(5, 8, false);
+    let rect = Rect::new(2.25, 0.0, 2.75, 8.0).to_path(0.1);
```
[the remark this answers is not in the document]
LaurenzV: Yeah you are right, using very small test cases is something I did in the very beginning, i've stopped with that by now. 😅 The reason I did it isn't really for storage reason, but because it makes it easier to debug in case there is a failure.

Nevertheless, since this test case is supposed to be the same as `filled_vertical_hairline_rect.png` (it's unfortunately not quite due to a rounding error), I think for this case it makes sense to try to keep the dimensions the same.

sparse_strips/vello_cpu/tests/basic.rs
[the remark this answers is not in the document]
LaurenzV: I mean, the thing is for things like gradients and patterns (where I'm also using `fill_rect`) it shouldn't really be a problem because the strip generation is completely unrelated to fine rasterization. So I feel like as long as we have at least a few cases of filled rectangle-like paths (which I added now), there shouldn't be any unexpected issues. But perhaps I'm wrong...

sparse_strips/vello_common/src/strip.rs
LaurenzV: I'm not sure I understand. 🤔 The disclaimer says that the logic _does_ account for such small rectangles, so you can use it with any arbitrary rectangle, no? 😄

sparse_strips/vello_cpu/src/render.rs
LaurenzV: A rotation is a specific case of skewing (the x and y axes are skewed by the same angle, and there also is a scaling factor), so I think this would also be covered by it? Though I can definitely rename it.

sparse_strips/vello_cpu/tests/basic.rs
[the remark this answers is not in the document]
tomcur: If all rectangles are tested using both the special case and the general case, we exercise both paths, and gain more confidence both implementations are correct. The simplest way is probably to test by using both code paths to render to pixmaps.

I'm thinking especially about tricky cases we've identified that exercise specific edge cases of the general tile generation and strip rendering. Those algorithms have to account for edge cases but need to do it in a performant way; the logic is tricky and easy to get wrong.

E.g., the following two are important tests for the general case, but no longer test it in the PR currently.

- Correct behavior when geometry extends above the viewport.
   https://github.com/linebender/vello/blob/f7b5c642f7f0a33a3443dab2af3b60514d1942c9/sparse_strips/vello_cpu/tests/basic.rs#L135-L145
- Correct winding and coverage calculation when geometry lands exactly on the tile boundary one tile to the left of the viewport.
   https://github.com/linebender/vello/blob/f7b5c642f7f0a33a3443dab2af3b60514d1942c9/sparse_strips/vello_cpu/tests/basic.rs#L168-L179

sparse_strips/vello_common/src/strip.rs
taj-p: I can't read 🤦

sparse_strips/vello_cpu/src/render.rs
taj-p: Oh interesting. I thought rotation is transformation about a point and skew shifting along an axis. TIL that shears + scale can become rotation - I've never seen that before - I guess that makes sense. I'll defer to your judgement on naming here.

waywardmonkeys: I think the trait method should get a doc comment to make this clear.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -302,3 +303,188 @@ pub fn render(
...
+    };
+
+    let top_alphas = vertical_alpha_coverage(top_strip_y);
+    // Create the strip for the top part of the rectangle.
+    horizontal_strip(alpha_buf, strip_buf, &top_alphas, top_strip_y);
+
+    // If rect covers more than one strip vertically, we need to strip the vertical line
+    // segments of the rectangle, and finally the bottom horizontal line segment.
+    if top_strip_idx != bottom_strip_idx {
+        let alphas = [1.0, 1.0, 1.0, 1.0];
```
grebmeg, line 448: Would it make sense to allocate it similarly to how you did above with `let mut buf = [0.0_f32; Tile::HEIGHT as usize];`?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -302,3 +303,188 @@ pub fn render(
...
+    if rect.is_zero_area() {
+        return;
+    }
+
+    // Note that we currently deal with negative-area rects as positive-area rects.
+    // Shouldn't be a problem for solid fill, but might need some tweaking for gradient
+    // and pattern fills.
+    let (x0, x1, y0, y1) = (
+        rect.min_x().max(0.0) as f32,
+        rect.max_x().min(width as f64) as f32,
```
dominikh, line 341: for `min_x() == 30`, `max_x() == 40`, `width == 10`, this sets `x0 = 30` and `x1 = 10`. This propagates through to `x_start` and `x_end`. Later we check that `x_end - x_start >= 1`, but due to wraparound, that'll be true. The end result is that we write a lot of unnecessary alpha values.

LaurenzV:
Maybe we can revisit in the future, but I'm pretty sure that for performant fine rasterization with SIMD, we will want to uphold the assumption that strips are always a multiple of 4 in width, which isn't the case with this fast path, and it might be tricky to change the PR to uphold this. So I'll close this for now.

PR #901, opened 2025-04-06
Reset whole render context in `reset`

LaurenzV:
Ran into an issue where running my demo tool for long enough would overflow the alpha buffer even though I called `reset`. I think everything should be cleared when you call that method.


grebmeg (approved):
[nit]: It might be worth adding a small test.

PR #902, opened 2025-04-06
Do not store output buffer in `Fine`

LaurenzV:
We only need the output buffer when packing but not for just fine rasterization, so there isn't much point in storing it inside of the struct.


tomcur (approved):
Agreed!

PR #908, opened 2025-04-11
Update to wgpu 25

waywardmonkeys:
wgpu 25 adjusted the downlevel limits so directly
setting `max_color_attachments` to 4 is no longer necessary (https://github.com/gfx-rs/wgpu/pull/6994).

wgpu 25 also now does polyfills for `unpack4x8unorm` and friends thanks to @DJMcNab in https://github.com/gfx-rs/wgpu/pull/7408.


waywardmonkeys:
This will need https://github.com/Wumpf/wgpu-profiler/pull/95

waywardmonkeys:
Also, before this lands, we'll need to do a release of Vello that uses wgpu 24 to keep `bevy_vello` happy.

sparse_strips/vello_hybrid/examples/webgl/src/lib.rs
```diff
@@ -65,21 +65,15 @@ impl RendererWrapper {
...
-                        max_color_attachments: 4,
-                        max_texture_dimension_2d: max_texture_size,
-                        ..wgpu::Limits::downlevel_webgl2_defaults()
-                    },
-                    ..Default::default()
+            .request_device(&wgpu::DeviceDescriptor {
+                label: None,
+                required_features: wgpu::Features::empty(),
+                required_limits: wgpu::Limits {
+                    max_texture_dimension_2d: max_texture_size,
```
DJMcNab, line 72: I didn't carefully review this code when it landed the first time. Should this be using the max size from the adapters [`limits`](https://docs.rs/wgpu/latest/wgpu/struct.Adapter.html#method.request_device), instead of getting it manually from the canvas? (cc @taj-p)

sparse_strips/vello_hybrid/examples/winit/src/render_context.rs
```diff
@@ -183,19 +183,17 @@ impl RenderContext {
     async fn new_device(&mut self, compatible_surface: Option<&Surface<'_>>) -> Option<usize> {
         let adapter =
             wgpu::util::initialize_adapter_from_env_or_default(&self.instance, compatible_surface)
-                .await?;
+                .await
+                .expect("No suitable GPU adapters found on the system!");
```
DJMcNab, line 187: I'd expect this to use `.ok()?` like the later one.

sparse_strips/vello_hybrid/shaders/sparse_strip_renderer.wgsl
```diff
@@ -132,16 +132,3 @@ fn unpack_alphas_from_channel(rgba: vec4<u32>, channel_index: u32) -> u32 {
...
-// Downlevel targets do not support native WGSL `unpack4x8unorm`.
-fn unpack4x8unorm(rgba_packed: u32) -> vec4<f32> {
-    // Extract each byte and convert to float in range [0,1]
-    return vec4<f32>(
-        f32((rgba_packed >> 0u) & 0xFFu) / 255.0,  // r
-        f32((rgba_packed >> 8u) & 0xFFu) / 255.0,  // g
-        f32((rgba_packed >> 16u) & 0xFFu) / 255.0, // b
-        f32((rgba_packed >> 24u) & 0xFFu) / 255.0  // a
-    );
-}
```
DJMcNab, line 147: Yay!

DJMcNab (commented):
We should also audit our uses of `BufferSlice`, to make sure they're still needed. In particular, *I think* `.slice(..)` can be removed entirely.

sparse_strips/vello_hybrid/examples/winit/src/render_context.rs
waywardmonkeys: Fixed.

sparse_strips/vello_hybrid/examples/webgl/src/lib.rs
taj-p: Yes - I was under the mistaken assumption that we could avoid all the parameter lookups required to construct `Limits`, but I see that [it's created as part of adapter construction](https://github.com/gfx-rs/wgpu/blob/303b0db00ef14431c75929814702aaef07c8cdbb/wgpu-hal/src/gles/adapter.rs#L677) so this is a needless parameter lookup.

I've created a PR for this change [here](https://github.com/linebender/vello/pull/911).

sagudev:
any progress with this?

waywardmonkeys:
> We should also audit our uses of `BufferSlice`, to make sure they're still needed. In particular, _I think_ `.slice(..)` can be removed entirely.

Sounds like a good follow up for someone that isn't me!

waywardmonkeys:
Okay. I'd like to land this. Bevy 0.17 is coming up and while it is currently on wgpu 25, they're probably going to update to the newly released wgpu 26 before releasing 0.17.

I don't see us publishing a new Vello in the next couple of weeks, so let's just roll this through.

PR #913, opened 2025-04-14
Use u8/u16 for clipping fine rasterization

LaurenzV:
Sorry, I probably should have noticed this in the original PR.  But since everything in the current version of the pipeline uses u8/u16, I think it makes sense for clipping to do that as well (especially since SIMD will also be important for clipping performance)? Maybe I'm missing the reason why it was done using f32, though.


grebmeg (approved):
> since everything in the current version of the pipeline uses u8/u16, I think it makes sense for clipping to do that as well (especially since SIMD will also be important for clipping performance)?

Yeah, I agree — it makes sense! Thanks for the correction!

PR #914, opened 2025-04-15
Fix issue in filling of clips

LaurenzV:
I noticed another issue in filling of clips (which already existed previously, but unfortunately I didn't detect it in my last PR). Currently, if I fill a rectangle with a black fill of opacity 0.5 onto a white background, I get the RGB values 132 134 138 instead of the expected values 127, 127, 127. I believe the issue was that we mistakenly multiplied the alpha of the source color to the source, which is not necessary because the color is already premultiplied.

With this PR, I get the expected values 127, 127, 127. It also fixes an inconsistency with an existing test, where the color was not consistent with the stripped part of the rectangle:

image

Once gradients are merged, we can probably rewrite `clip_fill` and `clip_stroke` to reuse the same method as for normal alpha compositing, to avoid code duplication.

### DIFF
--- a/sparse_strips/vello_cpu/snapshots/clip_rectangle_and_circle.png
+++ b/sparse_strips/vello_cpu/snapshots/clip_rectangle_and_circle.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ba1651c909d6b5b3080cf5c1399e3bb286262a0e21b52d704b26dcea94364f18
+oid sha256:0a3d7e6c32bb0d6a90218f2e02bb66d8873e7c3d35fc1c67e92f7f391934fa53
 size 813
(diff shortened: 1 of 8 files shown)


LaurenzV:
Yeah good point, let me try to understand why first.

LaurenzV:
Ah I mean, I think it makes sense. Because any shape that is within the clip will be inside the `source buffer`, so as soon as there is any opacity involved (e.g. through anti-aliasing), this would have triggered the bug.

LaurenzV:
I just noticed a bigger problem though. That is the current code (that I wrote) for `clip_strip`:

```rs
target_buffer[px_offset + channel_idx] =
                        div_255(dest * inverse_alpha + mask_alpha * source) as u8;
```

I don't think this is correct, as the normalization should only apply to the multiplication, so I think this should be

```rs
target_buffer[px_offset + channel_idx] =
                        (div_255(dest * inverse_alpha) + div_255(mask_alpha * source)) as u8;
```

However, I just realized that I have the exact same thing in our current `strip::src_over`, and changing that behavior leads to differences in pretty much all images. 😬 I guess I will open an issue for this.

grebmeg (commented):
Thanks!

PR #915, opened 2025-04-16
Add support for rendering images

LaurenzV:
Note that this PR builds on top of https://github.com/linebender/vello/pull/893, which is why the diff is so huge. The changes for image rendering are fortunately not as big and can be viewed in isolation here for easier reviewing: https://github.com/LaurenzV/vello/commit/54b90163ad514b060bbede0f398afacca472873e

The reason I'm opening this now already is that I hope that we can already discuss this in today's office hours so that we can land this first version of images soon after #893, as my development branch is starting to diverge a lot and it's getting increasingly hard to keep all the different features in sync. :(

Similarly to my gradients PR, performance was not a focus here, but instead correctness.


LaurenzV:
Will be opening a new PR with a clean commit history.

PR #916, opened 2025-04-16
Prevent overflow in clips that are completely outside of the viewport

LaurenzV:


sparse_strips/vello_common/src/coarse.rs
```diff
@@ -333,7 +333,7 @@ impl Wide {
         let mut cur_wtile_y = clip_bbox.y0();
 
         // Process strips to determine the clipping state for each wide tile
-        for i in 0..n_strips - 1 {
+        for i in 0..n_strips.saturating_sub(1) {
```
grebmeg, line 336: There're two other cases above where `n_strips - 1` is used when calculating bbox. I know there’s a precondition that returns an empty bbox when `n_strips <= 1`, but it might still be worth keeping the approach consistent here.

grebmeg (approved):

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -308,12 +308,12 @@ impl Wide {
         } else {
             // Calculate the y range from first to last strip in wide tile coordinates
             let wtile_y0 = strips[0].strip_y();
-            let wtile_y1 = strips[n_strips - 1].strip_y() + 1;
+            let wtile_y1 = strips[n_strips.saturating_sub(1)].strip_y() + 1;
```
tomcur, line 311: It is worth noting `saturating_sub` doesn't come completely free, but it's not worth worrying about that too much here as this path isn't too hot.

x86: https://godbolt.org/z/qP6r85q7f
AArch64: https://godbolt.org/z/dh91xGbbG

tomcur (approved):

LaurenzV:
Uhhhhm what is that error about?

LaurenzV:
I guess I'll just try again.

PR #920, opened 2025-04-17
Reduce clipping memory usage and allocation

tomcur:
The scratch strip buffer may be quite large when the render context has been used for rendering complex paths. If the clip geometry is simple, storing that buffer in the clip state has two downsides:
- the buffer will have unused capacity, inflating the memory footprint of the renderer,
- as the buffer is moved into the clip state, a new buffer has to be allocated, and for subsequent rendering of complex paths will have to grow large again.

This PR proposes keeping the original allocation for the scratch strip buffer and memcpy'ing the clip's strips to a new allocation of the right size. (As a side-effect the size of the `Clip` struct is reduced from 40 to 32 bytes.)

There are alternatives: the first downside could be alleviated by a `shrink_to_fit`, and allocator thrashing of the second downside could be reduced by allocating the new buffer with the same capacity as the old buffer. It's probably better to just keep the original allocation for the scratch buffer, though.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -9,7 +9,7 @@ use crate::{
     tile::Tile,
 };
 use alloc::vec;
-use alloc::vec::Vec;
+use alloc::{boxed::Box, vec::Vec};
 use vello_api::color::PremulRgba8;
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):
I think that makes sense!

PR #921, opened 2025-04-17
planeshift: Fork of the upstream

waywardmonkeys:
This is a fork of the upstream, without the backends yet, and without the GL integration, and without some of the winit integration.

### DIFF
--- a/planeshift/src/backend.rs
+++ b/planeshift/src/backend.rs
@@ -94,11 +94,10 @@ pub trait Backend: Sized {
 
     // `winit` integration
     #[cfg(feature = "enable-winit")]
-    fn window(&self) -> Option<&Window>;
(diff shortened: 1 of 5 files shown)


waywardmonkeys:
This is part of a conversation starter.

Among questions that should be answered:

* Does this approach work today?
* Can we get good backends in place for Core Animation, Windows.UI.Composition, Android's SurfaceControl, Wayland's APIs (including `viewporter), and a wgpu-based fallback?
* Should we do this here or approach upstream about doing this in the original repo?
* Should we keep this name or rename it?
* What will we need to do for winit integration for subsurfaces and such in Wayland and other APIs?

PoignardAzur:
I think this is superseded by the subduction work?

PR #922, opened 2025-04-17
Musings in the direction of spatio-temporal allocation

DJMcNab:
As discussed in [#checkins > Daniel McNab @ 💬](https://xi.zulipchat.com/#narrow/channel/197474-checkins/topic/Daniel.20McNab/near/512047439), I've started doing some coding on spatio-temporal allocation, as a task for my own interest.

Sharing this as a PR so that there's a stable pointer to it, but I don't expect to drive this to completion.

The current state is that the "pre-scheduling" is done. That is, I have the code which determines what needs to be scheduled, and where it can be scheduled to.

### DIFF
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -3,8 +3,10 @@
 
 //! Basic render operations.
 
+use std::ops::Range;
+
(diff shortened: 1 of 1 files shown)


DJMcNab:
This is horribly outdated now.

PR #923, opened 2025-04-18
Add support for blending, compositing and layers

LaurenzV:
This is a first attempt at an implementation for the layering model, as discussed during the office hours.


sparse_strips/vello_cpu/tests/assets/cowboy.png
DJMcNab, line 1: What license is this used under?

LaurenzV: It's the emoji from Noto Color Emoji, so it should be fine to include? (There already exist images of it, after all)

LaurenzV: (And I generated the image myself.)

LaurenzV:
I believe this should now be ready for review. :)

sparse_strips/vello_cpu/tests/assets/cowboy.png
LaurenzV: See also here, it's from the same font: https://github.com/linebender/vello/blob/b7f82779272a6a8c7e92fabb544233b0ec49c968/vello_tests/snapshots/big_colr.png

PoignardAzur (approved):
This isn't my area of expertise, but I'm not seeing any obvious red flags.

I think it's fine to merge this if not merging is blocking further work.

sparse_strips/vello_api/src/mask.rs
```diff
@@ -0,0 +1,86 @@
...
+use alloc::vec::Vec;
+
+/// A type of mask.
+#[derive(Debug, Copy, Clone)]
+pub enum MaskType {
+    /// A luminance mask.
+    Luminance,
+    /// An alpha mask.
+    Alpha,
+}
```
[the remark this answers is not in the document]
LaurenzV: Sounds reasonable, I will make that change.

DJMcNab (commented):
I'm not entirely convinced by the alpha mask feature here; it feels both:
1) Something which reasonably trivially composes into an additional blending operation
2) Hard to integrate into Vello Hybrid

LaurenzV:
The main reason I added it is because it’s needed for SVG, but if there is an equivalent using just blending then I’m open to removing it. But how exactly do you simulate an alpha mask using blending?

DJMcNab:
I've not thought about this carefully, but I imagine that creating the alpha mask as a (premultiplied white) image, sampling from that image, mix blend your "target" onto that image as source, and then blend with the "intended" mode onto the original source.

PR #924, opened 2025-04-21
sparse_strips: Mark as `no_std` but use `std`

waywardmonkeys:
This can't be enforced in CI yet, but marks these crates as `no_std` and then uses `std` in one spot in `vello_common`.

This can be improved and refined in subsequent PRs.

### DIFF
--- a/sparse_strips/vello_common/src/pico_svg.rs
+++ b/sparse_strips/vello_common/src/pico_svg.rs
@@ -9,7 +9,13 @@
 //! for demonstration purposes. It supports basic SVG features like paths,
 //! fill, stroke, and grouping.
 
-use std::str::FromStr;
+extern crate std;
+
+use alloc::boxed::Box;
+use alloc::vec;
+use alloc::vec::Vec;
+use core::str::FromStr;
+use std::eprintln;
 
 use roxmltree::{Document, Node};
 use vello_api::kurbo::{Affine, BezPath, Point, Size, Vec2};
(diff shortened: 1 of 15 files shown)


tomcur (approved):

sparse_strips/vello_common/src/pico_svg.rs
```diff
@@ -9,7 +9,13 @@
...
 //! fill, stroke, and grouping.
 
-use std::str::FromStr;
+extern crate std;
+
+use alloc::boxed::Box;
+use alloc::vec;
+use alloc::vec::Vec;
+use core::str::FromStr;
+use std::eprintln;
```
DJMcNab, line 18: :laughing:

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -29,6 +29,10 @@
 //!
 //! See the individual module documentation for more details on usage and implementation.
 
+#![no_std]
```
DJMcNab, line 32: We will actually need this for `wasm32-unknown-none`, so it's good to see this.

DJMcNab (approved):

sparse_strips/vello_common/src/pico_svg.rs
waywardmonkeys: This makes it clear now what the offending code is. :)

PR #925, opened 2025-04-21
Refactor the test infrastructure for sparse strips

LaurenzV:
This PR proposes refactoring the test infrastructure for sparse strips rendering. Note that it's not ready for careful review yet as there's still a bit missing (for example, I haven't moved the snapshot tests yet to avoid trouble with git LFS), but feedback on the overall idea in the meanwhile would definitely be appreciated (or we just discuss it on Wednesday).

To summarize, the main change is that the tests do not live in `vello_cpu` anymore, but instead in a custom crate `vello_sparse_tests` (I couldn't take `vello_tests` because that crate already exists). In addition to that, I created a new `vello_macros` crate which contains a custom macro that we use to define tests. The basic idea of how this works is that we annotate each test with this macro (and define some paramters like the height, width, etc.), and then the macro will automatically generate different versions of the same tests at runtime. I think this gives us some nice improvements now, and will gives us a lot of quality-of-life improvements even further down the road:

1) We do not need to always call the `get_ctx` and `check_ref` method, but each method is instead supplied the render context as a parameter, saving some boilerplate code.

2) Right now, our rendering pipeline is based solely on u8/16. Soon (hopefully in the next 1-2 weeks), I will start working on a version of fine rasterization that is based on f32 instead and should serve as the gold standard for correctness (the reference images will be based on that version of the pipeline). By having such a macro for our tests, once we get to that point we can simply adapt the macro to generate versions of the test based on u8 AND f32 and can compare them to the same reference image (for the u8 version, we will obviously need some wiggle room and allow discrepancies of +-1/2). But by having this design, we can simply define one version of the test and automatically test it against both versions of the pipeline, which I think is incredibly useful.

3) The above will become even more relevant once we get to SIMD, where we ideally also reuse each test to run it against that specific SIMD pipeline, which is very easy to do if every test uses that macro. A great thing is that we can also conditionally include/exclude tests based on the target feature we detect, so that we only run SSE tests on a processor supporting SSE, while we also include AVX2 and AVX512 versions of the tests if supported by the architecture.

4) Finally, by extracting the tests into a custom crate, once we have a render trait that is implemented by both, vello_hybrid and vello_cpu, it should be possible to change all occurances of `&mut RenderContext` to that trait, which should allow us to run those same tests against vello_hyrbid (again, with some wiggle room for small discrepancies), which I imagine would be very useful as well. 

We can add lots of configuration options, such as for example being able to temporarily disable specific kinds of tests via environment variables or completely disabling tests for one renderer if they involve features that are not supported by both.


taj-p:
>  once we have a render trait that is implemented by both, vello_hybrid and vello_cpu

I think I'd really like to add snapshot testing to vello_hybrid (especially with the clipping work) in the very near term. How do you see that evolution in relation to this refactor? Should this refactor live alongside a separate snapshot testing infrastructure within vello hybrid or is it possible to somehow combine the two?

taj-p:
> Since we don't have an implementation of the renderer trait yet, we could create a temporary one inside of `vello-tests`, and then we might already be able to integrate tests with vello-hybrid as part of this PR already!

I love this idea!

LaurenzV:
Should be good now. As far as I can tell, the images are again stored in Git LFS.

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Create a new vello snapshot test.
+#[proc_macro_attribute]
+pub fn v_test(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let attrs = parse_macro_input!(attr as AttributeInput);
+
+    let input_fn = parse_macro_input!(item as ItemFn);
+
+    let input_fn_name = input_fn.sig.ident.clone();
+    let u8_fn_name = Ident::new(&format!("{}_cpu_u8", input_fn_name), input_fn_name.span());
+    let hybrid_fn_name = Ident::new(&format!("{}_gpu", input_fn_name), input_fn_name.span());
```
taj-p, line 99:
```suggestion
    let hybrid_fn_name = Ident::new(&format!("{}_hybrid", input_fn_name), input_fn_name.span());
```

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    } else {
+        empty_snippet.clone()
+    };
+    let ignore_cpu = if skip_cpu {
+        ignore_snippet.clone()
+    } else {
+        empty_snippet.clone()
+    };
+    let cpu_threshold = threshold;
+    let gpu_threshold = threshold + 1;
```
taj-p, line 141: Might be worth defining these magic values to the top of the file as constants? 

```
const DEFAULT_CPU_TOLERANCE: u32 = 0;
const DEFAULT_HYBRID_TOLERANCE: u32 = 1;
```

and then refer to them in the documentation above threshold

Cargo.toml
```diff
@@ -132,3 +134,6 @@ criterion = { version = "0.5.1", default-features = false }
 rand = { version = "0.9.0", default-features = false, features = ["std_rng"] }
 usvg = { version = "0.45.0" }
 once_cell = "1.21.3"
+proc-macro2 = "1.0.93"
```
taj-p, line 137:
```suggestion

# Used for testing
proc-macro2 = "1.0.93"
```

sparse_strips/vello_sparse_tests/tests/util.rs
```diff
@@ -178,27 +179,35 @@ pub(crate) fn stops_blue_green_red_yellow() -> ColorStops {
...
+pub(crate) fn check_ref(
+    ctx: &impl Renderer,
+    // The name of the test.
+    test_name: &str,
+    // The name of the specific instance of the test that is being run
+    // (e.g. test_gpu, test_cpu_u8, etc.)
+    specific_name: &str,
+    // Tolerance for pixel differences.
+    threshold: u8,
+    // Whether the test instance the "gold standard" and should be used
```
taj-p, line 203:
```suggestion
    // Whether the test instance is the "gold standard" and should be used
```

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+            use crate::util::{
+                check_ref, get_ctx
+            };
+            use vello_cpu::RenderContext;
+
+            let mut ctx = get_ctx::<RenderContext>(#width, #height, #transparent);
+            #input_fn_name(&mut ctx);
+            if !#no_ref {
+                check_ref(&ctx, #input_fn_name_str, #u8_fn_name_str, #cpu_threshold, true);
+            }
```
taj-p, line 158: Snapshot names can currently clash between test modules. I.e., in the below scenario, we would generate snapshots called `with_foo.png` for both modules. 

There is also inconsistent use of module name prefixing across tests (e.g. clip tests are prefixed with `clip_` but issues and basic are not). I think we can either choose a consistent naming pattern or prefix snapshots as below. I don't mind which!

```
mod clip {
   // ...
   fn with_foo() { ... }
}
mod gradient {
   // ...
   fn with_foo() { ... }
}
```




```suggestion
            if !#no_ref {
                let module_path = module_path!().strip_prefix("tests::").unwrap();
                let ref_name = format!("{}::{}", module_path, #input_fn_name_str);
                check_ref(&ctx, &ref_name, #u8_fn_name_str, #cpu_threshold, true);
            }
```

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+            use crate::util::{
+                check_ref, get_ctx
+            };
+            use vello_hybrid::Scene;
+
+            let mut ctx = get_ctx::<Scene>(#width, #height, #transparent);
+            #input_fn_name(&mut ctx);
+            if !#no_ref {
+                check_ref(&ctx, #input_fn_name_str, #hybrid_fn_name_str, #gpu_threshold, false);
+            }
```
taj-p, line 173: As above

```suggestion
            if !#no_ref {
                let module_path = module_path!().strip_prefix("tests::").unwrap();
                let ref_name = format!("{}::{}", module_path, #input_fn_name_str);
                check_ref(&ctx, &ref_name, #hybrid_fn_name_str, #gpu_threshold, false);
            }
```

taj-p (approved):
LGTM! Nothing blocking - defer to your best judgement on any comment rather than my suggestion

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+struct Arguments {
+    /// The width of the scene.
+    width: u16,
+    /// The height of the scene.
+    height: u16,
+    /// The (additional) maximum tolerance for how much two pixels are allowed to deviate from each other
+    /// when comparing to the reference images. Some renderers already have an existing tolerance
+    /// (currently 0 for the CPU and 1 for the hybrid renderer), this value will simply be added
+    /// to the currently existing threshold
+    threshold: u8,
```
taj-p, line 64: Would it be possible to create separate fields for hybrid + cpu thresholds? I've found some clipping tests require 1 additional pixel deviation for hybrid

LaurenzV: I can do that, yeah. I suspect the higher differences come from the fact that we have the u8 pipeline for tests right now, I think it should be better once we have the f32 pipeline, but that remains to be seen. ^^

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+            skip_cpu: false,
+            skip_gpu: false,
+            no_ref: false,
+        }
+    }
+}
+
+/// Create a new vello snapshot test.
+#[proc_macro_attribute]
+pub fn v_test(attr: TokenStream, item: TokenStream) -> TokenStream {
```
DJMcNab, line 92: I'm not a massive fan of the name `v_test` - it's too terse.
Names like `multi_backend_test`, `vello_sparse_strips_test` and  all go too far the other way. Maybe `sparse_test` or even `vello_test` is the best option?

LaurenzV: Sure, can change that.

sparse_strips/vello_sparse_tests/tests/mod.rs
tomcur, line 1: Note that formatting isn't currently enforced by CI on these files, see https://github.com/linebender/vello/issues/946.

sparse_strips/vello_sparse_tests/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "vello_sparse_tests"
```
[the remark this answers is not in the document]
LaurenzV: Hmm I think it would be better to leave it that way for now, and in the future we might be able to merge it with the existing `vello_tests`.

sparse_strips/vello_macros/src/lib.rs
LaurenzV: You are right about that, I didn't know that you can access the module path like that! I think I will add this as a TODO for now because I don't want to rename all files (in this already very big PR) again, but I agree that we should change that.

sparse_strips/vello_macros/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+                match flag_str.as_str() {
+                    "transparent" => args.transparent = true,
+                    "skip_cpu" => args.skip_cpu = true,
+                    "skip_gpu" => args.skip_gpu = true,
+                    "no_ref" => args.no_ref = true,
+                    "ignore" => {
+                        args.skip_cpu = true;
+                        args.skip_gpu = true;
+                    }
+                    _ => panic!("unknown flag attribute {}", flag_str),
```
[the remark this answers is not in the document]
LaurenzV: There doesn't seem to be a `compile_fail!` macro, and for some reason using `compile_error!` always triggers a compilation failure in my case... I will leave it the way it is for now.

sparse_strips/vello_sparse_tests/tests/mod.rs
LaurenzV: Hmm, I tried running `cargo fmt -- ./**/*.rs` but it still doesn't seem to do anything for me...

LaurenzV: Let's just fix it in a follow-up PR so I can get this version landed.

PR #927, opened 2025-04-22
Store u8 and f32 versions of solid colors

LaurenzV:
This is to enable clients choosing what preciseness they would like to target, without having to recompute the color in each command.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -276,7 +276,6 @@ impl GlyphRenderer for RenderContext {
 #[cfg(test)]
 mod tests {
     use crate::RenderContext;
-    use core::iter;
     use vello_common::kurbo::{Rect, Shape};
 
     #[test]


sparse_strips/vello_api/src/paint.rs
```diff
@@ -26,6 +27,55 @@ impl IndexedPaint {
...
+struct ColorRepr {
+    // We pre-compute u8 and f32 forms of the color, so that clients can choose which one to
+    // use based on the targeted preciseness.
+    premul_u8: PremulRgba8,
+    premul_f32: [f32; 4],
+}
+
+/// A premultiplied color.
+#[derive(Debug, Clone, PartialEq)]
+// Wrap behind an Arc to keep the memory footprint small.
```
DJMcNab, line 40: Hmm. I think that 5 32 bit values isn't that much worse than 1 64 bit value. Especially since the latter also has an extra 2 64 bit values on the heap, plus whatever overhead from the heap itself.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -34,21 +84,17 @@ impl IndexedPaint {
 /// 2) Indexed paints, which can represent any arbitrary, more complex paint that is
 ///    determined by the frontend. The intended way of using this is to store a vector
 ///    of paints and store its index inside `IndexedPaint`.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum Paint {
     /// A premultiplied RGBA8 color.
-    Solid(PremulRgba8),
+    Solid(PremulColor),
```
DJMcNab, line 90: What even is the point of this fast-path now, if it requires going through a pointer dereference?

DJMcNab (commented):
Have you got the benchmarking results for this? It seems extremely expensive, if I understand correctly.

LaurenzV:
> It seems extremely expensive, if I understand correctly.

Is it? I mean the conversion/allocation is only performed once for each path drawn, so I don't think this will have any noticeable impact compared to the other stuff that happens. Or do you mean because of the additional pointer dereference? Which I also don't think will matter much because it only happens once per wide tile command, no? But yeah would be nice to measure somehow, though I'm not sure what the best way is.

sparse_strips/vello_api/src/paint.rs
LaurenzV: The thing is that each command stores a paint, so if you have tens of thousands of commands it does make a difference, but maybe this difference is not large enough to justify that. I was also on the fence about this, I'm open to just storing it on the stack instead.

sparse_strips/vello_api/src/paint.rs
LaurenzV: I would say the advantage is that we don't need to store the solid color inside of the `paints` vector each time the user sets a new paint.

LaurenzV: But it's just one allocation that happens when drawing a path, so this should be completely insignificant, no? There are much more allocations that happen in other parts of the pipeline, so in my view it shouldn't have any considerable impact on performance. But I'm definitely open for other options, the main point is that there needs to be some way to get both, a f32 and u8 version of the color, ideally without having to recalculate it in every command.

tomcur: After having rendered a few scenes, the scratch buffers will have grown and the pipeline doesn't necessarily need to allocate a lot anymore. I think the main remaining offender of allocation is flattening/expanding to `BezPath`s, but that could in principle be reduced or fully removed, too. Clipping also allocates, unfortunately. I'm in favor of trying to keep the hot-path as free of allocations as possible, which is the motivation behind e.g. https://github.com/linebender/vello/pull/920.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -99,6 +99,45 @@ pub struct Image {
...
+
+    /// Return the color as a premultiplied RGBA8 color, packed into a u32 in little-endian order.
+    pub fn rbga_u32(&self) -> u32 {
+        self.premul_u8.to_u32()
+    }
+
+    /// Return the color as a premultiplied RGBA32 color in the range [0.0, 1.0].
+    pub fn rbga_f32(&self) -> [f32; 4] {
+        self.premul_f32.components
+    }
```
tomcur, line 133: Would it make sense to give access to the underlying color values, by making the `PremulColor` fields `pub` or having the following?

```rust
pub fn as_premul_rgba8(&self) -> PremulRgba8;
pub fn as_premul_f32(&self) -> PremulColor<Srgb>
```

That way the normal `color` APIs can be used by consumers.

tomcur (approved):

PR #929, opened 2025-04-22
Initial scaffolding for image filters folder

DJMcNab:
This lays out the goals for the image filters work, and provides some basic scaffolding.

This intentionally does not start implementations, to avoid mixing in feedback for the layout with that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ members = [
     "sparse_strips/vello_hybrid/examples/webgl",
     "sparse_strips/vello_hybrid/examples/winit",
     "sparse_strips/vello_toy",
+
+    "image_filters/vello_filters_cpu",
 ]
 
 [workspace.package]
--- a/image_filters/README.md
+++ b/image_filters/README.md
@@ -1,7 +1,7 @@
 # Vello Image Filters
 
 In 2d Rendering, you often need to process images in ways which aren't just drawing of paths.
-For example, by blurring a background, zooming in, or inverting colours.
+For example, by blurring a background, zooming in, or inverting colors.
 This folder provides implementations of several different image filters, using different backends.
 
 These backends are:
--- a/image_filters/README.md
+++ b/image_filters/README.md
@@ -1,0 +1,44 @@
+# Vello Image Filters
+
+In 2d Rendering, you often need to process images in ways which aren't just drawing of paths.
+For example, by blurring a background, zooming in, or inverting colours.
+This folder provides implementations of several different image filters, using different backends.
+
+These backends are:
+
+- The "reference" CPU backend, found in [`vello_filters_cpu`].
+  We intend to publish this, and use it in Vello CPU.
+- GPU shader code, without a driver.
+  We haven't yet decided how this will be distributed.
+  This is not yet implemented.
+- Utilities and pipelines for using the GPU shader with wgpu for simple cases.
+  We intend to publish this crate, and might use it in Vello Hybrid.
+  This is not yet implemented.
+
+We intend to have GPU implementations which are able to be executed on WebGL, with more optimised versions (e.g. using compute shaders) for more powerful GPUs.
+
+In this README, we use the "Svg" prefix to refer to filters from the Filter Effects Module Level 1, found at <https://drafts.fxtf.org/filter-effects/>.
+The planned filters are (at least), and the implementations we have working are:
+
+| Filter                              | CPU | GPU (planned) | wgpu (planned) |
+| ----------------------------------- | --- | ------------- | -------------- |
+| [Svg Gaussian Blur][feGaussianBlur] |     |               |                |
+
+(Please let us know if this table is out-of-date)
+
+Our implementations have been based on the following sources:
+
+- resvg
+
+Filters which don't operate on multiple pixels are generally out-of-scope for initial versions of this crate.
+These include:
+
+- [Svg Blend][feBlend]. This will generally be implemented as part of your renderer directly, as it is an entirely local operation.
+- [Svg Composite][feComposite]. This will generally be implemented as part of a renderer directly, as it is entirely local. It's not clear how the arithmetic operation should be implemented.
+
+<!-- We expect to also adapt code from `imageproc`, but have not done so yet.
+This will impact the licenses this code is available under. -->
+
+[feGaussianBlur]: https://drafts.fxtf.org/filter-effects/#feGaussianBlurElement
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -21,6 +21,8 @@ members = [
     "sparse_strips/vello_hybrid/examples/webgl",
     "sparse_strips/vello_hybrid/examples/winit",
     "sparse_strips/vello_toy",
+
+    "image_filters/vello_filters_cpu",
```
LaurenzV, line 25: Wouldn't it make sense to put it into `sparse_strips` as well? Or is it going to be used by normal vello, too?

image_filters/vello_filters_cpu/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "vello_filters_cpu"
+version.workspace = true
+categories = ["rendering", "graphics"]
+keywords = ["2d", ""]
```
LaurenzV, line 5: Is the second empty keyword on purpose?

image_filters/vello_filters_cpu/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright 2020 the Vello Authors
```
LaurenzV, line 1: Not 2025?

image_filters/vello_filters_cpu/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2023 the Vello Authors
```
LaurenzV, line 1: Same here?

LaurenzV (approved):

image_filters/README.md
```diff
@@ -0,0 +1,44 @@
+# Vello Image Filters
+
+In 2d Rendering, you often need to process images in ways which aren't just drawing of paths.
+For example, by blurring a background, zooming in, or inverting colours.
```
tomcur, line 4:
```suggestion
For example, by blurring a background, zooming in, or inverting colors.
```

(sorry)

image_filters/vello_filters_cpu/LICENSE-MIT
DJMcNab: This one is right

image_filters/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Our implementations have been based on the following sources:
+
+- resvg
+
+Filters which don't operate on multiple pixels are generally out-of-scope for initial versions of this crate.
+These include:
+
+- [Svg Blend][feBlend]. This will generally be implemented as part of your renderer directly, as it is an entirely local operation.
+- [Svg Composite][feComposite]. This will generally be implemented as part of a renderer directly, as it is entirely local. It's not clear how the arithmetic operation should be implemented.
```
tomcur, line 37: Same for `feMerge` and `feFlood`? Clients can get drop shadows by combining some of these.

image_filters/README.md
```diff
@@ -0,0 +1,44 @@
...
+  This is not yet implemented.
+
+We intend to have GPU implementations which are able to be executed on WebGL, with more optimised versions (e.g. using compute shaders) for more powerful GPUs.
+
+In this README, we use the "Svg" prefix to refer to filters from the Filter Effects Module Level 1, found at <https://drafts.fxtf.org/filter-effects/>.
+The planned filters are (at least), and the implementations we have working are:
+
+| Filter                              | CPU | GPU (planned) | wgpu (planned) |
+| ----------------------------------- | --- | ------------- | -------------- |
+| [Svg Gaussian Blur][feGaussianBlur] |     |               |                |
```
tomcur, line 25: `feOffset` is a candidate for inclusion in an initial version: it's useful for drop shadows and if you're shifting by fractional pixels, you may want may want bi{linear,cubic} filtering.

Cargo.toml
DJMcNab: I hadn't done that for a couple of reasons:
1) This makes sense as its own grouping; that is, you'd probably end up having `sparse_strips/image_filters/vello_filters_cpu`
2) These don't really have anything to do with sparse strips; although the sparse strips renderers would use them

Happy to take feedback if we want to settle on a different solution here

image_filters/README.md
DJMcNab: Yes, I believe so. The thing to notice is that this folder is *not* implementing a render graph; at least, that's my current thinking. Instead, it *only* provides the implementations of filters as "pure" functions into buffers.

That is, this is very explicitly designed to be used inside an existing renderer. At which point, `feFlood` would just be drawing a rectangle using the normal mechanism in your renderer, probably.

DJMcNab: Similarly, I think `feMergeNode` is just doing a porter-duff over blit, which will already be implemented in your renderer.

DJMcNab:
This is a PR which is safe to merge with the extant broken CI, and I think getting it landed is valuable, so I'm expecting to do so once people are happy with what it is.

Cargo.toml
LaurenzV: Sounds reasonable to me.

tomcur (approved):
I agree with the minimal approach proposed, and practically having just the blur is (interestingly) enough to get quite a bit done. It's worth keeping, e.g., convolution in mind when thinking about the API, as a filter with more complicated input.

image_filters/vello_filters_cpu/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "vello_filters_cpu"
```
tomcur, line 2: I missed this in my original review, but I think this doesn't need "Vello" branding: the filters are intended to be a general solution. This might give someone the impression they are to be used together with Vello.

DJMcNab: That's a reasonable claim. I *think* I disagree; I think it's important to have a searchable name, and these are ultimately being developed with Vello as the "target" consumer.

But I don't feel all that strongly about it.

waywardmonkeys: If that's the case, then I guess I need to develop my own. My original discussion was entirely about this being independent of Vello.

tomcur: > I think it's important to have a searchable name

I agree. This could get a proper name of itself (in keeping with the Esperanto convention, throwing a bunch of words through a translation dictionary: "Miksi",  "Efiko", "Adapti").

PR #931, opened 2025-04-22
Add support for color spaces in gradients

LaurenzV:


LaurenzV:
Should be ready for review now.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -3,7 +3,8 @@
 
 //! Types for paints.
 
-use crate::kurbo::Affine;
+use crate::color::{ColorSpaceTag, HueDirection};
```
waywardmonkeys, line 6: Should combine these uses with the ones 2 lines down via `peniko::color`.

sparse_strips/vello_bench/src/cpu_fine.rs
```diff
@@ -96,7 +96,7 @@ pub fn fill(c: &mut Criterion) {
                 },
                 stops: $stops,
                 extend: peniko::Extend::$extend,
-                transform: Affine::default(),
+                ..Default::default()
```
waywardmonkeys, line 99: There's a periodic lint that we do that will ask us to turn this into `Gradient::default()`, so might as well do it now. Below as well (and in other files).

sparse_strips/vello_common/src/encode.rs
```diff
@@ -634,9 +665,9 @@ mod tests {
     use super::{EncodeExt, Gradient};
     use crate::color::DynamicColor;
     use crate::color::palette::css::{BLACK, BLUE, GREEN};
-    use crate::kurbo::{Affine, Point};
+    use crate::kurbo::Point;
+    use crate::peniko::ColorStops;
```
waywardmonkeys, line 669: Merge with line below.

sparse_strips/vello_bench/src/cpu_fine.rs
DJMcNab: The proper style for `..Default::default()` is this one, as far as I know. That is, the editor periodic lint doesn't trigger in this case.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -375,15 +403,15 @@ fn encode_stops(stops: &[ColorStop], start: f32, end: f32, pad: bool) -> Vec<Gra
...
         }
 
-        GradientRange {
+        vec![GradientRange {
             x0,
             x1,
             c0,
             factors,
-        }
+        }]
```
tomcur, line 411: This can return `GradientRange` directly rather than allocating a `Vec`:

```suggestion
        GradientRange {
            x0,
            x1,
            c0,
            factors,
        }
```

sparse_strips/vello_common/src/encode.rs
```diff
@@ -394,25 +422,28 @@ fn encode_stops(stops: &[ColorStop], start: f32, end: f32, pad: bool) -> Vec<Gra
...
         });
 
-        left_range.chain(stop_ranges.chain(right_range)).collect()
+        left_range
+            .chain(stop_ranges.chain(right_range))
+            .flatten()
+            .collect()
     } else {
-        stop_ranges.collect()
+        stop_ranges.flatten().collect()
```
tomcur, line 446: Using a `GradientRange` from `create_range` directly:

```suggestion
        // We handle padding by inserting dummy stops in the beginning and end with a very big
        // range.
        let left_range = iter::once({
            let first_stop = interpolated_stops.first().unwrap();
            let mut encoded_range = create_range(first_stop, first_stop);
            encoded_range.x0 = f32::MIN;

            encoded_range
        });

        let right_range = iter::once({
            let last_stop = interpolated_stops.last().unwrap();

            let mut encoded_range = create_range(last_stop, last_stop);
            encoded_range.x1 = f32::MAX;

            encoded_range
        });

        left_range.chain(stop_ranges.chain(right_range)).collect()
    } else {
        stop_ranges.collect()
```

sparse_strips/vello_common/src/encode.rs
tomcur: (Cont'd in https://github.com/linebender/vello/pull/931#discussion_r2056146844.)

sparse_strips/vello_common/src/encode.rs
```diff
@@ -375,15 +403,15 @@ fn encode_stops(stops: &[ColorStop], start: f32, end: f32, pad: bool) -> Vec<Gra
...
             x0,
             x1,
             c0,
             factors,
-        }
+        }]
     };
 
-    let stop_ranges = stops.windows(2).map(|s| {
+    let stop_ranges = interpolated_stops.windows(2).map(|s| {
```
tomcur, line 414:
```suggestion
    // Note: this could use `Iterator::map_windows` once stabilized, meaning `interpolated_stops`
    // no longer needs to be collected.
    let stop_ranges = interpolated_stops.windows(2).map(|s| {
```

Once [`Iterator::map_windows`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.map_windows) is stabilized, this would no longer need to have an allocated `interpolated_stops`, as it no longer needs `slice::windows`.

On the other hand, as a `Vec` is being built by this function (and not an iterator), the function could also be rewritten to just be a loop over `stops`, rather than building an explicit iterator that is collected at the end.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -343,22 +351,42 @@ fn apply_reflect(stops: &[ColorStop]) -> SmallVec<[ColorStop; 4]> {
...
-fn encode_stops(stops: &[ColorStop], start: f32, end: f32, pad: bool) -> Vec<GradientRange> {
-    let create_range = |left_stop: &ColorStop, right_stop: &ColorStop| {
+fn encode_stops(
+    stops: &[ColorStop],
+    start: f32,
+    end: f32,
+    pad: bool,
+    cs: ColorSpaceTag,
+    hue_dir: HueDirection,
+) -> Vec<GradientRange> {
```
tomcur, line 361: Not yet important, but if this were to take `&mut Vec<GradientRange>`, the caller could decide to reuse an existing allocation if it has one available (rather than collecting at the end, you'd use [`Vec::extend`](https://doc.rust-lang.org/stable/std/iter/trait.Extend.html#tymethod.extend)).

sparse_strips/vello_common/src/encode.rs
LaurenzV: Good catch!

sparse_strips/vello_common/src/encode.rs
tomcur: (This might be a nice candidate for [generators](https://rust-lang.github.io/rfcs/3513-gen-blocks.html).)

sparse_strips/vello_common/src/encode.rs
LaurenzV: Hmm, I don't see how this would be useful given that each encoded gradient owns the vector, though? You'd have to create a new one every time, anyway.

tomcur: That is true for current `vello_cpu`, but may not be true for all consumers. For `vello_cpu`, if allocation for these gradients turns out to be expensive (honestly, it probably isn't, in the grand scheme of things), you could conceivably keep hold of the old gradient buffers and reuse them rather than dropping them. I don't think it's currently important though.

tomcur: That is currently true, but if allocation for these gradients turns out to be expensive (honestly, it probably isn't, in the grand scheme of things), you could conceivably keep hold of the old gradient buffers and reuse them rather than dropping them. I don't think it's currently important though.

LaurenzV:
I think this one should be good to merge, right? I don't remember any concerns being raised during office hours.

tomcur (approved):

PR #934, opened 2025-04-23
Start of spatio-temportal allocation for clipping

raphlinus:
This commit has a sketch of spatio-temporal allocation for clipping, but it is not fully wired up yet. Scenes without clipping should work, but there is a fair amount of TODO remaining for clipping.

There's a fair amount of refactoring here. The biggest change is that draw calls and render passes can be issued from inside the scheduler, as opposed to separate "prepare" and "render" calls. The number of render passes needed will vary by the scene.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render.rs
+++ b/sparse_strips/vello_hybrid/src/render.rs
@@ -17,11 +17,12 @@ use std::fmt::Debug;
 use bytemuck::{Pod, Zeroable};
 use vello_common::tile::Tile;
 use wgpu::{
-    BindGroup, BindGroupLayout, BlendState, Buffer, ColorTargetState, ColorWrites, Device,
-    PipelineCompilationOptions, Queue, RenderPass, RenderPipeline, Texture, util::DeviceExt,
+    BindGroup, BindGroupLayout, BlendState, Buffer, ColorTargetState, ColorWrites, CommandEncoder,
+    Device, PipelineCompilationOptions, Queue, RenderPass, RenderPassColorAttachment,
+    RenderPassDescriptor, RenderPipeline, Texture, TextureView, util::DeviceExt,
 };
 
-use crate::scene::Scene;
+use crate::{scene::Scene, schedule::Schedule};
 
 /// Dimensions of the rendering target
 #[derive(Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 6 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,230 @@
...
+        wide_tile_y: u16,
+        tile: &WideTile,
+        state: &mut TileState,
+    ) {
+        state.stack.clear();
+        state.stack.push(TileEl {
+            slot_ix: !0,
+            round: self.round,
+        });
+        let bg = tile.bg.to_u32();
```
waywardmonkeys, line 113: Is `bg` going to be used outside of this alpha check and then the conditionalized draw?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -203,194 +265,263 @@ impl Renderer {
...
+            multiview: None,
+            cache: None,
+        });
+        let clip_textures = std::array::from_fn(|_| {
+            device.create_texture(&wgpu::TextureDescriptor {
+                label: Some("clip temp texture"),
+                size: wgpu::Extent3d {
+                    width: 256, // TODO: make configurable
+                    height: 1024,
+                    depth_or_array_layers: 1,
```
DJMcNab, line 317: I think it would be valid to use a 2d texture array here, rather than an array of textures.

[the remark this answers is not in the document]
raphlinus: I'm not sure it's going to be better. You're going to have an array of two texture views in any case, where one will be bound to sample from, and the other will be a render attachment. At the Vulkan level, there are going to be image barriers between the render passes. When it's bound as a render attachment, it's going to be `VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL`, and when sampled from, `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` (see the [texture](https://github.com/SaschaWillems/Vulkan/blob/74be818cc6b6fcab23932adbf245f624f4cf4ada/examples/texture/texture.cpp#L278) Vulkan example to see what that barrier looks like). I'm not sure having different layouts for different levels works; if not, I can imagine a smart driver using `VK_IMAGE_LAYOUT_GENERAL`, which might not be as efficient. I haven't investigated what wgpu actually does here. D3D12 has similar considerations.

In short, using a texture array gets you into tricky areas for the mapping to lower level APIs, while using two textures avoids that.

raphlinus:
This landed in other PRs, no point in having this one still open. Thanks Andrew and others!

PR #935, opened 2025-04-23
WIP [vello_hybrid] Image rendering

grebmeg:
This PR introduces image rendering in the Vello Hybrid GPU.

### WIP
The work is still in progress and there are a few remaining tasks to complete, but I’d appreciate any early feedback in the meantime.

**TODO:**
- The current implementation is based on `PaintType::Image`, which is used by `vello_cpu` and relies on `Pixmap`. We should support an opaque image ID reference here additionally, allowing the client, either using the existing `ImageCache` or a custom solution, to manage textures independently and supply only the references (with bind groups). How to implement this in a clever and convenient way is TBD.
- Refactor `GpuStrip` to only handle strip-related data. Image-specific fields such as `x_advance` and `y_advance` should be managed separately.

https://github.com/user-attachments/assets/f3ed11aa-917d-46e0-bfd5-5231151e559f


sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub view: TextureView,
+    /// The bind group for this image
+    pub bind_group: Option<BindGroup>,
+}
+
+/// Manages image resources for the renderer
+#[derive(Debug)]
+pub struct ImageCache {
+    /// Map of image IDs to resources
+    pub images: HashMap<u32, ImageResource>,
```
taj-p, line 22: We might be able to use a cheaper/faster collection type like using indices into a vector with free list or vector of `Occupied/Unoccupied` enums where `Unoccupied` slots point to the next free index.

waywardmonkeys: For the future, I have a work in progress crate where I was doing a resource pool using generational IDs and stuff based on how [Sokol](https://github.com/floooh/sokol) does them, but in Rust rather than C.

I plan to finish it up before RustWeek as I need it ... but happy to talk about this later.

examples/assets/desert.jpeg
taj-p, line 1: ❤️

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -187,12 +198,22 @@ impl Scene {
...
     /// Set the paint for subsequent rendering operations.
-    pub fn set_paint(&mut self, paint: Paint) {
-        self.paint_visible = match &paint {
-            Paint::Solid(color) => !color.is_transparent(),
-            Paint::Indexed(_) => true,
-        };
-        self.paint = paint;
+    pub fn set_paint(&mut self, paint: impl Into<PaintType>) {
+        self.paint = paint.into();
+    }
```
ajakubowicz-canva, line 203: FYI - see this feedback about the public API: https://github.com/linebender/vello/pull/1051#discussion_r2142629296

PR #937, opened 2025-04-24
Add support for rendering bitmap and COLR glyphs

LaurenzV:


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -248,12 +249,49 @@ impl RenderContext {
...
+                    .to_path(0.1),
+                    self.transform,
+                    &mut self.line_buf,
+                );
+
+                let image = Image {
+                    pixmap: Arc::new(glyph.pixmap),
+                    x_extend: peniko::Extend::Pad,
+                    y_extend: peniko::Extend::Pad,
+                    // TODO: Should this be configurable?
```
dfrg, line 283: If not configurable, we should probably default to high quality if we’re downsampling beyond a certain threshold (maybe < 0.5 in either dimension?). Bilinear filtering tends to produce bad results for small glyphs. I believe @DJMcNab described it as “cursed” :)

LaurenzV: Done!

LaurenzV:
Will mark this as draft and also include COLR glyphs on that branch, once I've managed to implement it.

dfrg:
> Will mark this as draft and also include COLR glyphs on that branch, once I've managed to implement it.

Sounds good. I assume this is blocked on a new release  of skrifa as well and I plan on getting one out this week.

LaurenzV:
New release would definitely be great, but I don't think it's a strict blocker as we can just depend on the git version for now.

examples/assets/colr_test_glyphs/test_glyphs-glyf_colr_1.ttf
DJMcNab, line 1: Is there a source for this file? We should link to it if so.

dfrg: https://github.com/googlefonts/color-fonts

LaurenzV:
I think this should be done now. :) 

We will probably have to change the API a bit later down the road to accomodate for proper client-side caching, but I think this shouldn't stop getting this initial version merged to have the functionality. There is a small edge case (radial gradients with negative radius) that we don't support yet, but I don't think this should be blocking. Other than that, I think the abstraction in `vello_common` is pretty nice and should make implementing the same for `vello_hybrid` easy once the require features are there!

nicoburns (approved):
Rebasing this on top of the `blitz` branch and running Blitz against https://www.fileformat.info/info/emoji/browsertest.htm on macos, this branch seems to render the same emoji as the `vello` crate. Some (notably the numbers at the top) are missing on both.

PR #940, opened 2025-04-28
Fix wide tile assignment order

LaurenzV:
I have no idea how this wasn't discovered earlier, but:

In `get` and `get_mut`, we compute the index as follows: `usize::from(y) * usize::from(self.width_tiles()) + usize::from(x)`, i.e. we assume the wide tiles are stored in row major.

However, when creating the wide tiles, we were first iterating over the width and then over the height, meaning that it was actually stored in column major, and because of this the values of `x` and `y` were wrong. I honestly do not really get why this didn't lead to test failures earlier, but I'm pretty sure this is wrong? Added a test to confirm.


taj-p (approved):
LGTM

PR #942, opened 2025-04-30
Update to new `peniko` and `color`

waywardmonkeys:


waywardmonkeys:
Examples for the sparse strips depend on Parley and so we end up in this non-virtuous cycle between Parley, Peniko, and Vello (examples).

There also isn't a great sequence to publish these crates where there isn't a bit of a mess.

Maybe we should prioritize cleaning this up, like my old proposal to have a `raw_resource_handle` crate in https://github.com/linebender/peniko/pull/68

DJMcNab (approved):
For reference, we fixed the Parley/Peniko mismatch with linebender/peniko#107

tomcur (approved):

PR #944, opened 2025-05-01
Pass in `Transform` for paint types separately

tomcur:
This proposes passing in `Transform` for paint types separately through methods on the render context, moving the field out of the `PaintType` types' structs. This allows using existing ecosystem types directly (like `peniko::Gradient`).

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -33,6 +33,7 @@ pub struct RenderContext {
     pub(crate) tiles: Tiles,
     pub(crate) strip_buf: Vec<Strip>,
     pub(crate) paint: PaintType,
+    pub(crate) paint_transform: Affine,
     pub(crate) stroke: Stroke,
     pub(crate) transform: Affine,
     pub(crate) fill_rule: Fill,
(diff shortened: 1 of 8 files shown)


tomcur (commented):
See also the discussion in https://github.com/linebender/vello/pull/893#discussion_r2032671026. Vello Classic passes the paint(/brush) transform in through the paint methods, e.g., [`vello::Scene::fill`](https://docs.rs/vello/0.4.1/vello/struct.Scene.html#method.fill).

nicoburns (approved):
I think this makes sense.

IMO a similar change adopting `peniko::Blob`, `peniko::Image`, and `peniko::Brush` would also make sense. This would mean allocating an `Arc<dyn AsRef<[u8]> + Send + Sync>` (which could be an `Arc<Pixmap>` or an `Arc<Vec<u8>>`), but this would be be a small allocation rather than copying the entire image data. And we could always add an additional paint type(s) to `peniko::Brush` later is avoiding that allocation turns out to be important.

LaurenzV (approved):

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -196,6 +201,16 @@ impl RenderContext {
         self.paint = paint.into();
     }
 
+    /// Set the current paint transform.
```
DJMcNab, line 204: Some more complete docs here might be good. Even just mentioning the use case, noting that the brush is impacted by the global transform, etc. (that is, that this is logically applied "after" the current global transform).

I won't block on this

PR #945, opened 2025-05-01
Format code in vello_cpu/tests

tomcur:
These aren't picked up by `cargo fmt --all` (also not in CI).

We could do the following.

```bash
cargo fmt --check -- ./**/*.rs
```

### DIFF
--- a/sparse_strips/vello_cpu/tests/basic.rs
+++ b/sparse_strips/vello_cpu/tests/basic.rs
@@ -3,9 +3,7 @@
 
 //! Tests for basic functionality.
 
-use crate::util::{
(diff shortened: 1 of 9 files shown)


LaurenzV:
Maybe we can do this after https://github.com/linebender/vello/pull/925 is merged? Otherwise it will just make rebasing harder. 😬

LaurenzV (approved):

tomcur:
> Ah, but if your other PR is based on that then don't worry about it, I will figure it out.

Sorry! I think a `git rebase main -Xours` should get you most of the way there.

PR #947, opened 2025-05-02
vello_hybrid: WebGL2 renderer backend that skips wgpu

ajakubowicz-canva:
### PR Status

This PR is a working proposal that exists to present the benefits of a webgl backend to vello_hybrid. A couple caveats about the implementation upfront:
 - The wgsl shader is transpiled in a quick and dirty way using `naga`. This works for the POC, but I'd be interested to discuss if an approach like linebender/shaders should be taken, where the `wgsl` shader is transpiled into a Rust struct containing the binding metadata.
 - Secondly, everything can be subject to change. This PR is merely a conversation starter that I implemented to validate the approach and binary size wins.

### Context

Currently about ~30% of the internet is not using WebGPU (via [CanIUse](https://caniuse.com/webgpu)), requiring the use of the `webgl` feature on `wgpu`.
This incurs a roughly `3MB` binary size increase as wgpu needs to ship `naga` to the user. This verges on blocking adoption, hence this PR.

This PR adds a `webgl` feature to `vello_hybrid` which:
 1. Transpiles the `wgsl` shader to WebGL2 compatible `glsl`.
 2. Splits the `render.rs` module into a `wgpu` backend and a `WebGL` backend.

The `WebGL` backend doesn't use `wgpu` at all, instead using the native browser `WebGL2Context`.

For testing, this PR also forks the `webgl` example:
 - Example `webgl` now contains the webgl backend. You can see that there aren't many changes from a consumer as they simply need to pass in the webgl 2 context. Note, I didn't clean up the example at all, instead only making the changes required to migrate to the WebGL backend.
 - The prior `webgl` example using `wgpu` has been moved to `wgpu_webgl` and is unchanged.

### Results

The `wgpu_webgl` example running in release mode has a binary size of `5.1MB`.
The `webgl` example running in release mode has a binary size of `1.9MB`.

This is a `3.2MB` savings.

I've also tested this PR internally and it went smoothly and binary size reduction is consistent with the examples.

I'm super excited to discuss this with you all!


DJMcNab:
A few thoughts, not carefully thought through. We discussed this in [#office hours > Renderer 2025-05-07](https://xi.zulipchat.com/#narrow/channel/359642-office-hours/topic/Renderer.202025-05-07/with/516756847). I'd recommend following the track in the transcript:
1) The need for runtime lookups has been raised as unfortunate. The place to look to avoid this is wiring through the information from [`naga::back::glsl::ReflectionInfo`](https://docs.rs/naga/latest/naga/back/glsl/struct.ReflectionInfo.html) to the runtime. This is similar to the approach we take in the extant Vello Shaders.
2) We probably want *specifically* the shader generation to be its own crate. This means that you don't need to run the `build.rs` unless you have the web/opengl backend enabled. This again is as in Vello Shaders.
3) The "binary patching" done needs a really strong motivating comment. I *suspect* this is related to point (1). Deferring doing this in a more principled way *could* be reasonable.
4) Have you considered using glutin here? That would allow the use of OpenGL on non-web platforms, among other considerations. I think the lift from that would be quite low, so it's pretty easy.
5) The impact this has on image resources is potentially concerning. That's not for this PR to sort out - #957 and then #935 would need to handle it. This should probably land before #935.

ajakubowicz-canva:
Superseded by https://github.com/linebender/vello/pull/1011

PR #948, opened 2025-05-02
Add support for drawing blurred, rounded rectangles

LaurenzV:
Since I don't really understand the math behind it I unfortunately cannot add better documentation, but it seems to work at least.


nicoburns:
Seems to produce the desired effect (see outline of search box)

Screenshot 2025-05-02 at 15 06 38

(Unrelated to this PR, but something funky is going on with the leaf image in the footer, which is odd as I believe that's a png!)

sparse_strips/vello_common/src/math.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Mathematical helpfer functions.
```
tomcur, line 4:
```suggestion
//! Mathematical helper functions.
```

sparse_strips/vello_cpu/src/fine/blurred_rect.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::util::scalar::div_255;
+use vello_common::encode::EncodedBlurredRectangle;
+use vello_common::kurbo::Point;
+use vello_common::math::compute_erf7;
+
+#[derive(Debug)]
+pub(crate) struct BlurredRectFiller<'a> {
+    /// The current position that should be processed.
+    cur_pos: Point,
+    /// The underlying gradient.
```
tomcur, line 18:
```suggestion
    /// The underlying encoded blurred rectangle.
```

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -116,6 +117,37 @@ impl RenderContext {
         self.fill_path(&rect.to_path(DEFAULT_TOLERANCE));
     }
 
+    /// Fill a blurred rectangle with the given radius and standard deviation.
+    ///
+    /// Note that this only works properly if the current paint is set to a solid color.
+    /// If not, it will fall back to using black as the fill color.
+    pub fn fill_blurred_rect(&mut self, rect: &Rect, radius: f32, std_dev: f32) {
```
tomcur, line 124:
```suggestion
    pub fn fill_blurred_rounded_rect(&mut self, rect: &Rect, radius: f32, std_dev: f32) {
```

sparse_strips/vello_common/src/encode.rs
```diff
@@ -701,6 +711,113 @@ impl GradientLike for RadialKind {
...
+        let std_dev = self.std_dev.max(1e-6);
+
+        let min_edge = width.min(height);
+        let rmax = 0.5 * min_edge;
+        let r0 = radius.hypot(std_dev * 1.15).min(rmax);
+        let r1 = radius.hypot(std_dev * 2.0).min(rmax);
+
+        let exponent = 2.0 * r1 / r0;
+
+        let std_dev_inv = std_dev.max(1e-6).recip();
```
tomcur, line 784: The `f32::max` is already performed above.

```suggestion
        let std_dev_inv = std_dev.recip();
```

sparse_strips/vello_common/src/encode.rs
```diff
@@ -701,6 +711,113 @@ impl GradientLike for RadialKind {
     }
 }
 
+/// An encoded blurred rectangle.
+#[derive(Debug)]
+pub struct EncodedBlurredRectangle {
```
tomcur, line 716:
```suggestion
/// An encoded blurred, rounded rectangle.
#[derive(Debug)]
pub struct EncodedBlurredRoundedRectangle {
```

sparse_strips/vello_cpu/src/fine/blurred_rect.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Drawing blurred rectangles.
+//!
+//! Implementation is adapted from: <https://git.sr.ht>/~raph/blurrr/tree/master/src/distfield.rs.
```
tomcur, line 6:
```suggestion
//! Implementation is adapted from: <https://git.sr.ht/~raph/blurrr/tree/master/src/distfield.rs>.
```

LaurenzV:
If you have a minimal example I will look into it!

tomcur (approved):
Awesome! I've given the math a very cursory glance, it seems to be correctly adapted from the math at https://git.sr.ht/~raph/blurrr/tree/master/src/distfield.rs, which I'm assuming is correct. That leaves just some nits. In the APIs I think we should call this "blurred, rounded rect", not just "blurred rect".

nicoburns:
> If you have a minimal example I will look into it!

@LaurenzV Not minimal yet, but it's this image:

download

Or as a data URL:

`data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABUAAAAYCAMAAAAiV0Z6AAAAPFBMVEVLoEN0wU6CzFKCzFKCzFKCzFKCzFJSo0MSczNDmkCCzFJPoUMTczNdr0gmgziCzFITczMTczMTczMTczPh00jOAAAAFHRSTlPF/+bIsms8Ad///hX+//5/tXw7aMEAx10AAACaSURBVHgBbc4HDoRQCATQ33tbvf9dF9QxaCT9UQaltLHOh/golXKhMs5Xqa0xU1lyoa2fXFyQOsDG38qsLy4TaV+sFislovyhPzLJJrBu6eQOtpW0LjbJkzTuTDLRVNKa3uxJI+VdiRqXSeu6GW+Qxi29eLIi8H7EsYrT42BD+mQtNO5JMjRuC4lSY8V4hsLX0egGijvUSEP9AbylEsOkeCgWAAAAAElFTkSuQmCC`

Can be run in context by running

```bash
cargo run --release --example screenshot https://google.com --cpu
```

with the branch from this PR: https://github.com/DioxusLabs/blitz/pull/209

I suspect it may be my conversion from `peniko::Image` to `Pixmap` which is at fault. I've just used exactly the same bytes for both:

```rust
Arc::new(Pixmap {
      width: image.width as u16,
      height: image.height as u16,
      buf: unpremultiply(image.data.as_ref()),
})
  ```
  
but https://github.com/linebender/vello/pull/949 suggests a transformation is needed. I've tried adding the following conversion function:

```rust
fn unpremultiply(src: &[u8]) -> Vec<u8> {
    let mut out = Vec::with_capacity(src.len());
    for src in src.chunks_exact(4) {
        let alpha = src[3];
        out.push(((src[0] as u16 * 255) / alpha as u16) as u8);
        out.push(((src[1] as u16 * 255) / alpha as u16) as u8);
        out.push(((src[2] as u16 * 255) / alpha as u16) as u8);
        out.push(alpha);
    }

    out
}
```

But that doesn't seem to help.

LaurenzV:
Will take a look next week.

LaurenzV:
Ah, it's because you were _unpremultiplying_ it instead of premultiplying it. You need something like this:

```rs

            let premultiplied = image.data.as_ref()
                .chunks_exact(4)
                .flat_map(|d| {
                    let alpha = d[3] as u16;
                    let premultiply = |e: u8| ((e as u16 * alpha) / 255) as u8;

                    if alpha == 0 {
                        [0, 0, 0, 0]
                    } else {
                        [
                            premultiply(d[0]),
                            premultiply(d[1]),
                            premultiply(d[2]),
                            d[3],
                        ]
                    }
                })
                .collect::<Vec<_>>();
            
            let pixmap = Pixmap {
                width: image.width as u16,
                height: image.height as u16,
                buf: premultiplied,
            };
```

Though now I am getting this weird greenish border, not sure if that's intended?

image

LaurenzV:
Unfortunately I'm not really able to reproduce the green border when rendering it myself with a minimal vello_cpu example, so not sure what causes this.

LaurenzV:
Nevermind! It's because you are not rendering with extend mode pad, if I manually change it to pad it works. :)

image

PR #949, opened 2025-05-02
Use `peniko::Image` for the image `PaintType`

tomcur:
This is more or less a continuation of https://github.com/linebender/vello/pull/944, which enabled more easily using ecosystem types.

Note `vello_api` allows constructing a `Pixmap` from PNG data, but this PR does not currently implement the same for a `peniko::Image`.

Further, because `Pixmap` holds premultiplied RGBA8, it is not trivial to convert a `Pixmap` to a `peniko::Image`, as the latter for the moment only holds regular RGBA8. Doing PNG data -> `Pixmap` -> `peniko::Image` is lossy because of all the rounding.

For the moment, this does the premultiplication for rendering during sampling pixel values for fine rasterization. That might not be the best option, especially if at some point in the future `peniko::Image` supports more formats than just RGBA8.

### DIFF
--- a/sparse_strips/vello_api/src/paint.rs
+++ b/sparse_strips/vello_api/src/paint.rs
@@ -3,10 +3,8 @@
 
 //! Types for paints.
 
-use crate::pixmap::Pixmap;
-use alloc::sync::Arc;
 use peniko::{
-    Gradient, ImageQuality,
+    Gradient, Image,
     color::{AlphaColor, PremulRgba8, Srgb},
 };
 
@@ -53,19 +51,6 @@ impl From<AlphaColor<Srgb>> for Paint {
         // code.
         Self::Solid(PremulColor::new(value))
     }
-}
-
-/// An image.
-#[derive(Debug, Clone)]
-pub struct Image {
-    /// The underlying pixmap of the image.
-    pub pixmap: Arc<Pixmap>,
-    /// Extend mode in the horizontal direction.
-    pub x_extend: peniko::Extend,
-    /// Extend mode in the vertical direction.
-    pub y_extend: peniko::Extend,
-    /// Hint for desired rendering quality.
-    pub quality: ImageQuality,
 }
 
 /// A premultiplied color.
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -7,15 +7,14 @@ use crate::color::palette::css::BLACK;
 use crate::color::{ColorSpaceTag, HueDirection, PremulColor, Srgb, gradient};
 use crate::encode::private::Sealed;
 use crate::kurbo::{Affine, Point, Vec2};
-use crate::peniko::{ColorStop, Extend, Gradient, GradientKind, ImageQuality};
-use crate::pixmap::Pixmap;
+use crate::peniko::{Blob, ColorStop, Extend, Gradient, GradientKind, Image, ImageQuality};
 use alloc::borrow::Cow;
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::f32::consts::PI;
 use core::iter;
 use smallvec::SmallVec;
-use vello_api::paint::{Image, IndexedPaint, Paint};
+use vello_api::color::PremulRgba8;
+use vello_api::paint::{IndexedPaint, Paint};
 
 const DEGENERATE_THRESHOLD: f32 = 1.0e-6;
 const NUDGE_VAL: f32 = 1.0e-7;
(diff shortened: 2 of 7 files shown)


sparse_strips/vello_api/src/paint.rs
```diff
@@ -127,6 +112,11 @@ impl From<Gradient> for PaintType {
 
 impl From<Image> for PaintType {
     fn from(value: Image) -> Self {
+        assert!(
+            value.width < u16::MAX as u32 && value.height < u16::MAX as u32,
+            "the image is too big, the width and height can be no larger than {}px",
+            u16::MAX,
+        );
```
tomcur, line 119: How do we want to deal with fallible paint types?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -520,6 +528,24 @@ pub struct EncodedImage {
...
+
+        let global_alpha = (self.alpha * 255. + 0.5) as u8;
+        let alpha = ((data[3] as u16 * global_alpha as u16) / 255) as u8;
+        PremulRgba8 {
+            r: ((data[0] as u16 * alpha as u16) / 255) as u8,
+            g: ((data[1] as u16 * alpha as u16) / 255) as u8,
+            b: ((data[2] as u16 * alpha as u16) / 255) as u8,
+            a: alpha,
+        }
+    }
```
tomcur, line 546: This might not be the best place for doing this conversion, and it will get more complicated if `peniko::ImageFormat` supports more than `Rgba8` in the future.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -127,6 +112,11 @@ impl From<Gradient> for PaintType {
...
 impl From<Image> for PaintType {
     fn from(value: Image) -> Self {
+        assert!(
+            value.width < u16::MAX as u32 && value.height < u16::MAX as u32,
+            "the image is too big, the width and height can be no larger than {} pixels",
+            u16::MAX,
+        );
         Self::Image(value)
     }
 }
```
tomcur, line 122: How do we want to deal with fallible paint types?

tomcur: (Note in this case we could instead just forward the width and height as `u32` into the `EncodedImage`, but currently we convert the width and height to `u16` similarly to the other dimensions in the pipeline.)

nicoburns: I think panicking is probably fine for now? u16::MAX is pretty large for a bitmap!

LaurenzV:
While it would be nice to use peniko types throughout the whole code, I think we should not rush this change and maybe revisit the topic again in a few weeks. Having all those calculations in the `sample` function will make it incredibly slow, and I don't think we can afford this performance-wise...

So my suggestion would be to wait until I'm done with scalar performance optimizations + SIMD, and then we can look into whether we can change the API to work with peniko images instead. But I am a bit doubtful.

nicoburns:
> While it would be nice to use peniko types throughout the whole code, I think we should not rush this change and maybe revisit the topic again in a few weeks. Having all those calculations in the `sample` function will make it incredibly slow, and I don't think we can afford this performance-wise...

Possibly this is one to solve with the "pre-registration" API we discussed in office hours. I figure that when the image (could be a `peniko::Image`) is registered, the conversion could happen at that point (for GPU renderers the image could be uploaded as a texture at that point). Then it only needs to happen once per-image rather than once frame / draw.

tomcur (commented):
> Possibly this is one to solve with the "pre-registration" API we discussed in office hours? I figure that when the image (could be a peniko::Image) is registered, the conversion could happen at that point

Yes, something like that seems ideal. There is one possible solution before that, where at the encoding step the image is converted to premultiplied sRGB (or other compositing color space).

nicoburns:
> One thing I'm wondering is whether it makes sense for Peniko to have `ImageFormat::PremulRgba8`. That would allow `Pixmap` to be readily converted to `peniko::Image`. The main question that raises is how we want to deal with image format conversions in Vello CPU, Vello Hybrid, and existing Vello Classic.

Perhaps one for rendering office hours, but supporting both `Rgba8` and `PremulRgba8` formats makes sense to me.
- Images decoded from things like PNG or JPEG files will likely not be in premultiplied format
- Images coming from prior rendering stages may well be

So probably best to let the `Image` type represent both, and then the renderer can do the minimum number of conversions?

nicoburns:
Another thing which occurred to me: perhaps `Pixmap` ought to be considered `vello_cpu`'s "texture" format (analogous to `wgpu::Texture` in the GPU renderer)?

sagudev:
In servo, we have something called [`Snapshot`](https://github.com/servo/servo/blob/52f53f61e15cfe74570473e87aa1b87f8eaf96f2/components/pixels/snapshot.rs#L121), so image that can be (premultilied) RGBA/BGRA. What it is exactly it's determined at runtime (somehow similar to [`DynamicColor`](https://docs.rs/color/latest/color/struct.DynamicColor.html)), this allow us to avoid as many conversations as possible and only do conversions when really needed. It would be useful for servo if peniko/vello had something similar.

PR #950, opened 2025-05-02
Add `RenderContext::render_to_buffer` to vello_cpu

nicoburns:
This adds a more flexible API to `vello_cpu` which allows rendering into an arbitrary `&mut [u8]` with separately provided `width` and `height` rather than specifically a `Pixmap`. This is useful when you don't own/control the buffer.

The underlying code in `fine` was already compatible with this.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -239,14 +239,23 @@ impl RenderContext {
         self.wide.reset();
     }
 
-    /// Render the current context into a pixmap.
-    pub fn render_to_pixmap(&self, pixmap: &mut Pixmap) {
+    /// Render the current context into a buffer.
+    /// The buffer is expected to be in premultiplied RGBA8 format with length `width * height * 4`
+    pub fn render_to_buffer(&self, buffer: &mut [u8], width: u16, height: u16) {
         assert!(
             !self.wide.has_layers(),
             "some layers haven't been popped yet"
         );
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -239,14 +239,19 @@ impl RenderContext {
         self.wide.reset();
     }
 
-    /// Render the current context into a pixmap.
-    pub fn render_to_pixmap(&self, pixmap: &mut Pixmap) {
+    /// Render the current context into a buffer.
+    pub fn render_to_buffer(&self, buffer: &mut [u8], width: u16, height: u16) {
```
LaurenzV, line 243: Might be obvious, but we probably should document the width/height reuirement, as well as that an rgba8 buffer is expected.

tomcur: > that an rgba8 buffer is expected

I agree; and premultiplied RGBA8 specifically.

nicoburns: That makes sense. Do either of you have a good idea about how/whether we might be able to support writing into `&mut [u32]` where `u32` is in `ARGB` format (high bits are A). This is the format the [softbuffer::Buffer](https://docs.rs/softbuffer/latest/softbuffer/struct.Buffer.html) uses (which would be the expected crate to use for rendering `vello_cpu` into a `winit` window)?

Issue #917, opened 2025-04-16
OOB access for clip exceeding viewport on the right


LaurenzV:
For the following code:

```rs
let mut ctx = RenderContext::new(100, 100);

ctx.clip(&Rect::new(0.0, 0.0, 500.0, 10.0).to_path(0.1));
ctx.fill_rect(&Rect::new(0.0, 0.0, 100.0, 100.0));
ctx.finish();
```

I get:

```

attempted to access out-of-bounds wide tile
thread 'empty::colr' panicked at sparse_strips/vello_common/src/coarse.rs:162:9:
attempted to access out-of-bounds wide tile
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/aa8f0fd7163a2f23aa958faed30c9c2b77b934a5/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/aa8f0fd7163a2f23aa958faed30c9c2b77b934a5/library/core/src/panicking.rs:75:14
   2: vello_common::coarse::Wide::get_mut
             at /Users/lstampfl/Programming/GitHub/vello/sparse_strips/vello_common/src/coarse.rs:162:9
   3: vello_common::coarse::Wide::pop_clip
             at /Users/lstampfl/Programming/GitHub/vello/sparse_strips/vello_common/src/coarse.rs:584:25
   4: vello_common::coarse::Wide::pop_clips
             at /Users/lstampfl/Programming/GitHub/vello/sparse_strips/vello_common/src/coarse.rs:423:13
   5: vello_cpu::render::RenderContext::finish
             at ./src/render.rs:190:9
   6: tests::empty::colr
             at ./tests/empty.rs:65:5
   7: tests::empty::colr::{{closure}}
             at ./tests/empty.rs:60:10
   8: core::ops::function::FnOnce::call_once
             at /Users/lstampfl/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/aa8f0fd7163a2f23aa958faed30c9c2b77b934a5/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```

PR #951, opened 2025-05-02
Fix: clear layer and clip stacks in `Wide::reset`

nicoburns:
Without this patch, I get a lot "some layers haven't been popped yet" errors (from the assert in `RenderContext::render_to_buffer`) when reusing a `RenderContext` after resetting it by calling `RenderContext::reset`.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -156,6 +156,8 @@ impl Wide {
             tile.bg = PremulColor::new(TRANSPARENT);
             tile.cmds.clear();
         }
+        self.layer_stack.clear();
+        self.clip_stack.clear();
     }
 
     /// Return the number of horizontal tiles.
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):
Thanks, could you maybe add a test case below, similar to the ones we currently have?

LaurenzV:
Though it does seem a bit suspicious because, assuming no bugs in either your code or vello_cpu, after you are done rendering, the layer count should always be 0 if all `push_layer` calls are followed by a `pop_layer` call at some point. 

But it still makes sense to reset that when resetting the render context.

nicoburns:
> Though it does seem a bit suspicious because, assuming no bugs in either your code or vello_cpu, after you are done rendering, the layer count should always be 0 if all `push_layer` calls are followed by a `pop_layer` call at some point.

I think I've worked out what's happening. This bug occurs in Blitz's WPT test runner which uses `catch_unwind` to isolate individual tests cheaply. So if a test panics during scene building (e.g. due to #917 or some issue in Blitz's code) then the `RenderContext` which will still contain the partially constructed scene is reused for the *next* test, with `RenderContext::reset` being relied upon to correctly reset the state.

I'll add a test.

PR #953, opened 2025-05-03
Update most deps

waywardmonkeys:
This leaves `wgpu` at 24 for this release, skrifa at the same version as it is now, and doesn't update `android_logger`.


examples/scenes/Cargo.toml
tomcur, line 1: I believe the bump to `rand` 0.9.1 is missing here.

tomcur (approved):
One observation about  `scenes/simple`.

examples/scenes/Cargo.toml
waywardmonkeys: Resolved in #954.

waywardmonkeys:
@tomcur Thanks for catching that `rand` issue. :)

PR #957, opened 2025-05-05
[Vello Hybrid]: Clipping (Spatiotemporal Allocation)

taj-p:
## Context

Hooks up and addresses all the TODO's in Raph's sketch (https://github.com/linebender/vello/pull/934). See this [thread](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Spatiotemporal.20allocation.20.28hybrid.29/near/513442829)[#vello > Spatiotemporal allocation (hybrid) @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Spatiotemporal.20allocation.20.28hybrid.29/near/513442829). 

## Notes

-  Made `GpuResources` non-optional - I thought this was more trouble than it was worth.
- Adds in new `clear_slots` pipeline to support fine grained clearing of slots in slot textures (needed for spatiotemporal allocation)
- Regenerates test snapshots
- Enables the clipping test suite for `vello_hybrid`
- There are many performance wins to be had. This PR is pretty big already, so I think these are worth following up separately.
   - Not re-creating buffers for each render pass (re-using the allocations between calls)
   - Using a staging belt (to prevent allocating an extra staging buffer per `write_buffer`)
   - Perhaps allowing more than 1 column of slots per slot texture.

I've copied the documentation from `schedule.rs` below for information about how spatiotemporal allocation works:

# Scheduling

- Draw commands are either issued to the final target or slots in a clip texture.
- Rounds represent a draw in up to 3 render targets (two clip textures and a final target).
- The clip texture stores slots for many clip depths. Once our clip textures are full,
  we flush rounds (i.e. execute render passes) to free up space. Note that a slot refers
  to 1 wide tile's worth of pixels in the clip texture.
- The `free` vector contains the indices of the slots that are available for use in the two clip textures.

## Example

Consider the following scene of drawing a single wide tile with three overlapping rectangles with
decreasing width clipping regions.

```rs
const WIDTH: f64 = 100.0;
const HEIGHT: f64 = Tile::HEIGHT as f64;
const OFFSET: f64 = WIDTH / 3.0;

let colors = [RED, GREEN, BLUE];

for i in 0..3 {
    let clip_rect = Rect::new((i as f64) * OFFSET, 0.0, 100, HEIGHT);
    ctx.push_clip_layer(&clip_rect.to_path(0.1));
    ctx.set_paint(colors[i]);
    ctx.fill_rect(&Rect::new(0.0, 0.0, WIDTH, HEIGHT));
}
for _ in 0..3 {
    ctx.pop_layer();
}
```

This single wide tile scene should produce the below rendering:
```
┌────────────────────────────┌────────────────────────────┌─────────────────────────────
│      ──              ───   │       /     /       /     /│        ──────────────      │
│  ────            ────      │      /     /       /     / │────────                    │
│──           ─────          │     /     /       /     /  │                            │
│        ───Red              │    /     /Green  /     /   │           Blue             │
│    ────                ──  │   /     /       /     /    │                     ───────│
│ ───                ────    │  /     /       /     /     │       ──────────────       │
│                  ──        │ /     /       /     /      │───────                     │
└────────────────────────────└────────────────────────────└────────────────────────────┘
                                                                                        
```
How the scene is scheduled into rounds and draw calls are shown below:

### Round 0

In this round, we don't have any preserved slots or slots that we need to sample from. Simply,
draw unclipped primitives.

### Draw to texture 0:

In Slot N - 1 of texture 0, draw the unclipped green rectangle.

Slot N - 1:
```
┌──────────────────────────────────────────────────────────────────────────────────────┐
│       /     /       /     /        /     /       /     /       /     /       /     / │
│      /     /       /     /        /     /       /     /       /     /       /     /  │
│     /     /       /     /        /     /       /     /       /     /       /     /   │
│    /     /       /     /        /     / Green /     /       /     /       /     /    │
│   /     /       /     /        /     /       /     /       /     /       /     /     │
│  /     /       /     /        /     /       /     /       /     /       /     /      │
│ /     /       /     /        /     /       /     /       /     /       /     /       │
└──────────────────────────────────────────────────────────────────────────────────────┘
```

### Draw to texture 1:

In Slot N - 2 of texture 1, draw unclipped red rectangle and, in slot N - 1, draw the unclipped
blue rectangle.

Slot N - 2:
```
┌──────────────────────────────────────────────────────────────────────────────────────┐
│      ──              ───                            ──              ───              │
│  ────            ────               ──          ────            ────               ──│
│──           ─────               ────          ──           ─────               ────  │
│        ─────                ────        Red           ─────                ────      │
│    ────                 ────                      ────                 ────          │
│ ───                 ────                       ───                 ────              │
│                  ───                                            ───                  │
└──────────────────────────────────────────────────────────────────────────────────────┘
```
Slot N - 1:
```
┌──────────────────────────────────────────────────────────────────────────────────────┐
│                                           ────────────────────────────────────────── │
│───────────────────────────────────────────                                           │
│                                                                                      │
│                                         Blue                          ───────────────│
│                                           ────────────────────────────               │
│               ────────────────────────────                                           │
│───────────────                                                                       │
└──────────────────────────────────────────────────────────────────────────────────────┘
```

### Round 1

At this point, we have three slots that contain our unclipped rectangles. In this round,
we start to sample those pixels to apply clipping (texture 1 samples from texture 0 and
the render target view samples from texture 1).

### Draw to texture 0:

Slot N - 1 of texture 0 contains our unclipped green rectangle. In this draw, we sample
the pixels from slot N - 2 from texture 1 to draw the blue rectangle into this slot.

Slot N - 1:
```
┌─────────────────────────────────────────────────────────┌─────────────────────────────
│        /     /       /     /       /     /       /     /│        ──────────────      │
│       /     /       /     /       /     /       /     / │────────                    │
│      /     /       /     /       /     /       /     /  │                            │
│     /     /       /  Green      /     /       /     /   │           Blue             │
│    /     /       /     /       /     /       /     /    │                     ───────│
│   /     /       /     /       /     /       /     /     │       ──────────────       │
│  /     /       /     /       /     /       /     /      │───────                     │
└─────────────────────────────────────────────────────────└────────────────────────────┘
```

### Draw to texture 1:

Then, into Slot N - 2 of texture 1, which contains our red rectangle, we sample the pixels
from slot N - 1 of texture 0 which contain our green and blue rectangles.
```

┌────────────────────────────┌────────────────────────────┌─────────────────────────────
│      ──              ───   │       /     /       /     /│        ──────────────      │
│  ────            ────      │      /     /       /     / │────────                    │
│──           ─────          │     /     /       /     /  │                            │
│        ───Red              │    /     /Green  /     /   │           Blue             │
│    ────                ──  │   /     /       /     /    │                     ───────│
│ ───                ────    │  /     /       /     /     │       ──────────────       │
│                  ──        │ /     /       /     /      │───────                     │
└────────────────────────────└────────────────────────────└────────────────────────────┘
```

### Draw to render target

At this point, we can sample the pixels from slot N - 1 of texture 1 to draw the final
rendition.

## Nuances

- When there are no clip/blend regions, we can render directly to the final target.
- The above example provides an intuitive explanation for how rounds after 3 clip depths
  are scheduled. At clip depths 1 and 2, we can draw directly to the final target within a
  single round.
- Before drawing into any slot, we need to clear it. If all slots can be cleared or are free,
  we can use a `LoadOp::Clear` operation. Otherwise, we need to clear the dirty slots using
  a fine grained render pass.

For more information about this algorithm, see this [Zulip thread].

[Zulip thread]: https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Spatiotemporal.20allocation.20.28hybrid.29/near/513442829


sparse_strips/vello_sparse_tests/snapshots/image_with_transform_rotate_1.png
taj-p, line 1: I think these snapshots are no longer used.

sparse_strips/vello_sparse_tests/tests/clip.rs
```diff
@@ -67,6 +91,50 @@ fn clip_rectangle_with_star_evenodd(ctx: &mut impl Renderer) {
     ctx.pop_layer();
 }
 
+#[vello_test]
+fn clip_deeply_nested_circles(ctx: &mut impl Renderer) {
```
taj-p, line 95: I would like this test to exercise creating many rounds in the scheduler.

taj-p:
cc @ajakubowicz-canva

sparse_strips/vello_sparse_tests/snapshots/fill_command_respects_clip_bounds.png
LaurenzV, line 1: What happened here?

LaurenzV: As in, why did the reference image change if the CPU version wasn't touched?

sparse_strips/vello_sparse_tests/snapshots/image_with_transform_rotate_1.png
LaurenzV, line 1: Why are those deleted? We do ignore the tests right now because they don't run in Windows CI, but I think we can still keep the reference images.

LaurenzV: They are currently not used because the tests are ignored, but they still exist in the `image` test file.

sparse_strips/vello_sparse_tests/snapshots/fill_command_respects_clip_bounds.png
LaurenzV: Ah, I presume it's because you regenerated all images and oxipng created a slightly different version of the png file.

sparse_strips/vello_sparse_tests/snapshots/image_with_transform_rotate_1.png
taj-p: Ah sorry, would you like me to revert this change?

sparse_strips/vello_sparse_tests/snapshots/fill_command_respects_clip_bounds.png
taj-p: I've reverted these changes in https://github.com/linebender/vello/pull/957/commits/2ab36ec191bf9025c6482b9d4a5be88a2769c89d

sparse_strips/vello_sparse_tests/snapshots/image_with_transform_rotate_1.png
taj-p: I've reverted these changes in https://github.com/linebender/vello/pull/957/commits/2ab36ec191bf9025c6482b9d4a5be88a2769c89d

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -405,10 +405,14 @@ impl Programs {
         );
 
         let max_texture_dimension_2d = device.limits().max_texture_dimension_2d;
-        let alpha_texture_height = 2;
-        let alphas_texture =
-            Self::make_alphas_texture(device, max_texture_dimension_2d, alpha_texture_height);
-        let alpha_data = vec![0; (max_texture_dimension_2d * alpha_texture_height * 16) as usize];
+        const INITIAL_ALPHA_TEXTURE_HEIGHT: u32 = 1;
```
ajakubowicz-canva, line 408: Non actionable – for my own understanding, why is it a safe change to change `INITIAL_ALPHA_TEXTURE_HEIGHT` from `2` to `1`? What does it impact?

taj-p: This was motivated by wanting to keep initial allocations small. If the scene requires more alphas, that can be done in `prepare`.

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -30,18 +30,39 @@
...
+pub enum RenderError {
+    /// No slots available for rendering.
+    ///
+    /// This error is likely to occur if a scene has an extreme number of nested layers
+    /// (clipping, blending, masks, or opacity layers).
+    ///
+    /// TODO: Consider supporting more than a single column of slots in slot textures.
+    #[error("No slots available for rendering")]
+    SlotsExhausted,
+}
```
ajakubowicz-canva, line 64: Are there other errors we will want to add?

Should `unwrap` and `expect` methods actually return a `RenderError`?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+            depth_stencil: None,
+            multisample: wgpu::MultisampleState::default(),
+            multiview: None,
+            cache: None,
+        });
+
+        let slot_texture_views: [TextureView; 2] = core::array::from_fn(|_| {
+            device
+                .create_texture(&wgpu::TextureDescriptor {
+                    label: Some("slot temp texture"),
```
ajakubowicz-canva, line 356: If we're labelling the slot texture descriptors, should the index passed into the closure be used to differentiate slot 0 from slot 1?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+                    size: wgpu::Extent3d {
+                        width: WideTile::WIDTH as u32,
+                        height: Tile::HEIGHT as u32 * slot_count as u32,
+                        depth_or_array_layers: 1,
+                    },
+                    mip_level_count: 1,
+                    sample_count: 1,
+                    dimension: wgpu::TextureDimension::D2,
+                    // TODO: Is this correct or need it be RGBA8Unorm?
+                    format: render_target_config.format,
```
ajakubowicz-canva, line 366: Who sets the rendering target? It looks like it's set by the caller.

In our `render_to_file` example, we appear to pass `Rgba8Unorm` as the render target config format. Whilst for the `winit` example we pass `Bgra8Unorm` as the surface format.

I guess another question is related to the texture definition in the shader. In `wgsl` `clip_input_texture` is typed as `texture_2d<f32>`.

Is it expected that the vello_hybrid renderer works with all wgsl texture formats? (sorry for the naivety).

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+            label: Some("Clear Slots Bind Group"),
+            layout: &clear_bind_group_layout,
+            entries: &[wgpu::BindGroupEntry {
+                binding: 0,
+                resource: clear_config_buffer.as_entire_binding(),
+            }],
+        });
+        let clear_slot_indices_buffer = Self::make_clear_slot_indices_buffer(
+            device,
+            slot_count as u64 * size_of::<u32>() as u64,
```
ajakubowicz-canva, line 395: The `u32` attribute used for the slot `index` that's passed to the clear slots wgsl shader seems brittle in it's definition when configuring the descriptors / pipeline.

A parallel is the `GpuStrip` which encapsulates the attribute passed to the `render_strips` shader. Is it worth using the [New Type Idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) to wrap this `u32` such that it's traceable to all locations that need to use the accurate size?

For example this line of code. But also the `array_stride` in the `clear_pipeline`?
It could also provide the attributes?

Maybe this isn't worth doing for 2 code locations.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+                width: render_target_config.width,
+                height: render_target_config.height,
+            },
+            max_texture_dimension_2d,
+        );
+
+        let slot_bind_groups = Self::make_strip_bind_groups(
+            device,
+            &strip_bind_group_layout,
+            &alphas_texture,
```
ajakubowicz-canva, line 428: Each `make_strip_bind_group` creates its own view of the alpha texture. Is it worth creating the view out here and pass it in (similar to how the slot texture views are passed in)?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
                 // There are 16 1-byte alpha values per texel.
-                let alpha_texture_height =
-                    (u32::try_from(alpha_len).unwrap()).div_ceil(max_texture_dimension_2d * 16);
-
+                .div_ceil(alpha_width * 16);
+            let required_alpha_size = alpha_width * required_alpha_height * 16;
+            let current_alpha_size = {
+                let alphas_texture = &self.resources.alphas_texture;
+                alphas_texture.width() * alphas_texture.height() * 16
+            };
```
ajakubowicz-canva, line 599: Definitely can be punted as a performance optimization, but are these multiplications really required to decide whether the alpha texture needs to be expanded?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+            if required_alpha_size > current_alpha_size {
+                // We need to resize the alpha texture to fit the new alpha data.
                 assert!(
-                    alpha_texture_height <= max_texture_dimension_2d,
+                    required_alpha_height <= alpha_width,
```
ajakubowicz-canva, line 603: In this case I might prefer `alpha_width` remains `max_texture_dimensions_2d` as the name is overloaded.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -415,23 +672,118 @@ impl Renderer {
...
-        let resources = &self
-            .resources
-            .as_ref()
-            .expect("`prepare` should be called before `render`");
-        let render_data = scene.prepare_render_data();
-        render_pass.set_pipeline(&self.render_pipeline);
-        render_pass.set_bind_group(0, &resources.render_bind_group, &[]);
-        render_pass.set_vertex_buffer(0, resources.strips_buffer.slice(..));
-        let strips_to_draw = render_data.strips.len();
+    /// Upload the strip data by appending to `self.resources.strips_buffer`.
```
ajakubowicz-canva, line 675: Is the comment correct regarding _appending_ to `self.resources.strips_buffer`? It looks more like `upload_strips` creates a new buffer and assigns it to `strips_buffer`?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -415,23 +672,118 @@ impl Renderer {
...
+    fn upload_strips(&mut self, device: &Device, queue: &Queue, strips: &[GpuStrip]) {
+        if strips.is_empty() {
+            return;
```
ajakubowicz-canva, line 678: Maybe an incorrect assumption, but I would expect that whatever `strips` slice I pass to `upload_strips` is reflected on the `strips_buffer`, meaning I'd expect an empty buffer if an empty slice is passed. Currently the prior filled strips are preserved.

That may be fine if the buffer is simply not read in the case where `strips` is empty, but from reading `do_strip_render_pass`, the vertex buffer is set from the `strips_buffer` meaning it wouldn't be cleared.

sparse_strips/vello_api/src/mask.rs
```diff
@@ -0,0 +1,83 @@
...
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+
+/// A mask.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Mask {
+    data: Arc<Vec<u8>>,
+    width: u16,
+    height: u16,
+}
```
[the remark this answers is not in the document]
taj-p: I'm not sure how you were able to comment on `vello_api` here. I'm going to ignore these comments since they're out of scope of the PR but they could certainly be applied separately.

sparse_strips/vello_api/src/mask.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Return the height of the mask.
+    pub fn height(&self) -> u16 {
+        self.height
+    }
+
+    /// Sample the value at a specific location.
+    ///
+    /// This function might panic or yield a wrong result if the location
+    /// is out-of-bounds.
```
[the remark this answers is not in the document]
taj-p: I'm not sure how you were able to comment on vello_api here. I'm going to ignore these comments since they're out of scope of the PR but they could certainly be applied separately.

sparse_strips/vello_api/src/mask.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Create a new luminance mask from the pixmap.
+    pub fn new_luminance(pixmap: &Pixmap) -> Self {
+        Self::new_with(pixmap, false)
+    }
+
+    fn new_with(pixmap: &Pixmap, alpha_mask: bool) -> Self {
+        let mut data = Vec::with_capacity(pixmap.width() as usize * pixmap.height() as usize);
+
+        for pixel in pixmap.data().chunks_exact(4) {
```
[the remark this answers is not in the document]
taj-p: I'm not sure how you were able to comment on vello_api here. I'm going to ignore these comments since they're out of scope of the PR but they could certainly be applied separately.

sparse_strips/vello_api/src/mask.rs
```diff
@@ -0,0 +1,83 @@
...
+                if pixel[3] != 0 {
+                    r /= a;
+                    g /= a;
+                    b /= a;
+                }
+
+                // See https://www.w3.org/TR/filter-effects-1/#elementdef-fecolormatrix
+                let luma = r * 0.2126 + g * 0.7152 + b * 0.0722;
+                #[allow(clippy::cast_possible_truncation, reason = "This cannot overflow")]
+                data.push(((luma * a) * 255.0 + 0.5) as u8);
```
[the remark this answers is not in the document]
taj-p: I'm not sure how you were able to comment on vello_api here. I'm going to ignore these comments since they're out of scope of the PR but they could certainly be applied separately.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -48,37 +51,54 @@ impl From<AlphaColor<Srgb>> for Paint {
         // Since we only do that conversion once per path it might not be critical, but should
         // still be measured. This also applies to all other usages of `to_rgba8` in the current
         // code.
-        Self::Solid(value.premultiply().to_rgba8())
+        Self::Solid(PremulColor::new(value))
```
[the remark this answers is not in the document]
taj-p: I'm not sure how you were able to comment on vello_api here. I'm going to ignore these comments since they're out of scope of the PR but they could certainly be applied separately.

sparse_strips/vello_api/src/paint.rs
```diff
@@ -48,37 +51,54 @@ impl From<AlphaColor<Srgb>> for Paint {
...
-            .for_each(|stop| *stop = stop.multiply_alpha(alpha));
-        self
+impl PremulColor {
+    /// Create a new premultiplied color.
+    pub fn new(color: AlphaColor<Srgb>) -> Self {
+        let premul = color.premultiply();
+
+        Self {
+            premul_u8: premul.to_rgba8(),
+            premul_f32: premul,
```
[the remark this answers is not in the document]
taj-p: I'm not sure how you were able to comment on vello_api here. I'm going to ignore these comments since they're out of scope of the PR but they could certainly be applied separately.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -415,23 +672,118 @@ impl Renderer {
...
+                })],
+                depth_stencil_attachment: None,
+                occlusion_query_set: None,
+                timestamp_writes: None,
+            });
+
+            render_pass.set_pipeline(&self.programs.clear_pipeline);
+            render_pass.set_bind_group(0, &resources.clear_bind_group, &[]);
+            render_pass.set_vertex_buffer(0, resources.clear_slot_indices_buffer.slice(..));
+            render_pass.draw(0..4, 0..slot_indices.len() as u32);
```
ajakubowicz-canva, line 786: Should this also use `u32::try_from`?

sparse_strips/vello_hybrid/src/render.rs
taj-p: Yes - nice catch 🙏 . Thank you

sparse_strips/vello_hybrid/shaders/clear_slots.wgsl
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// This shader clears specific slots in slot textures to transparent pixels.
+
+// Assumes this texture consists of a single column of slots of `config.slot_height`, 
+// numbering from 0 to `texture_height / slot_height - 1` from top to bottom.
+
+struct Config {
+    // Width of a slot (typically matching `WideTile::WIDTH`)
+    slot_width: u32,
```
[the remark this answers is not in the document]
taj-p: Fixed in https://github.com/linebender/vello/pull/957/commits/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,484 @@
...
+//! │        ───Red              │    /     /Green  /     /   │           Blue             │
+//! │    ────                ──  │   /     /       /     /    │                     ───────│
+//! │ ───                ────    │  /     /       /     /     │       ──────────────       │
+//! │                  ──        │ /     /       /     /      │───────                     │
+//! └────────────────────────────└────────────────────────────└────────────────────────────┘
+//!
+//! ### Draw to render target
+//!
+//! At this point, we can sample the pixels from slot N - 1 of texture 1 to draw the final
+//! rendition.
```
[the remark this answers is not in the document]
taj-p: Fixed in https://github.com/linebender/vello/commit/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -44,43 +46,95 @@ pub struct RenderTargetConfig {
...
+    pub fn render(
+        &mut self,
+        scene: &Scene,
+        device: &Device,
+        queue: &Queue,
+        encoder: &mut CommandEncoder,
+        render_size: &RenderSize,
+        view: &TextureView,
+    ) -> Result<(), RenderError> {
+        // For the time being, we upload the entire alpha buffer as one big chunk. As a future
```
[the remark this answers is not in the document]
taj-p: Fixed in https://github.com/linebender/vello/commit/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/lib.rs
taj-p: Added in https://github.com/linebender/vello/commit/41f78ba77f528202bf4b247de8e518511ff078d5

ajakubowicz-canva: Sorry I wasn't clear in the initial comment. I meant this as a philosophical question about the future. Do you expect any fallible behaviors to be moved into this enum in the future?

sparse_strips/vello_hybrid/src/render.rs
taj-p: `format` is relatively [bloaty](https://rustwasm.github.io/book/reference/code-size.html#avoid-string-formatting) - I think having any label is sufficient for debugging purposes (but I did make the case of the label consistent with other labels in https://github.com/linebender/vello/pull/957/commits/41f78ba77f528202bf4b247de8e518511ff078d5)

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+    }
 
-                    (strips_too_small, alpha_too_small, dimensions_changed)
-                }
-                // self.resources is None if prepare has not been called yet
-                None => (true, true, true),
-            };
+    fn make_alphas_texture(
+        device: &Device,
+        max_texture_dimension_2d: u32,
```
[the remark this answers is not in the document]
taj-p: Great comment - thank you. Updated in https://github.com/linebender/vello/pull/957/commits/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/lib.rs
taj-p: Yes. For example, I think when there are too many alphas to fit into the device's alpha texture, we should return an appropriate render error. 

The question of returning a `RenderError` for unwraps and expects which we believe are safe is another question entirely. Historically, if I believe I'm smarter than the compiler, I lean towards panicking, but I think this decision should be made by Linebender as a group. I'll ask about it in Utrecht.

taj-p: I've unresolved this so that it gets greater visibility from other reviewers.

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -205,197 +312,347 @@ impl Renderer {
...
+        device: &Device,
+        strip_bind_group_layout: &BindGroupLayout,
+        alphas_texture: &Texture,
+        config_buffer: &Buffer,
+        strip_texture_view: &TextureView,
+    ) -> BindGroup {
+        let alphas_texture_view =
+            alphas_texture.create_view(&wgpu::TextureViewDescriptor::default());
+        device.create_bind_group(&wgpu::BindGroupDescriptor {
+            label: Some("Strip Bind Group"),
```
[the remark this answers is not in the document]
taj-p: format is relatively [bloaty](https://rustwasm.github.io/book/reference/code-size.html#avoid-string-formatting) - I think having any label is sufficient for debugging purposes (but I did make the case of the label consistent with other labels in https://github.com/linebender/vello/commit/41f78ba77f528202bf4b247de8e518511ff078d5)

sparse_strips/vello_hybrid/src/render.rs
taj-p: Very nice! Fixed in https://github.com/linebender/vello/pull/957/commits/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -121,6 +121,53 @@ impl Scene {
...
+            None
+        };
+
+        // Blend mode, opacity, and mask are not supported yet.
+        if blend_mode.is_some() {
+            unimplemented!()
+        }
+        if opacity.is_some() {
+            unimplemented!()
+        }
```
ajakubowicz-canva, line 148: I may figure this out as I keep reviewing - but I am naively confused how opacity is not implemented or even how it's expected to be implemented, when there are tests like `clip_with_opacity`.

Is this assumption accurate? In the test `clip_with_opacity`, a clip layer is pushed, and then a rect is filled with alpha 0.5. Would the equivalent in the future be: push a clip layer with opacity set to `0.5`, and then paint a fully opaque rect. Would the opacity in the clip layer transfer to the child drawn within it and the two cases would render the same?

Also, I just realized I am a little confused about using the name `alpha` vs `opacity`. Are they the same thing, e.g. an alpha of `1` and an opacity of `1` are both opaque. Should terms be consolidated?


Edit: yep, I confirmed opacity here behaves as I expect.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -121,6 +121,53 @@ impl Scene {
...
+        if mask.is_some() {
+            unimplemented!()
+        }
+
+        self.wide.push_layer(
+            clip,
+            BlendMode::new(Mix::Normal, Compose::SrcOver),
+            None,
```
ajakubowicz-canva, line 156: I realize I don't actually fully understand the difference between clipping and masking.
Clarifying: A mask is essentially a pixmap (alpha or luminance), whilst a clip_path lets you clip via a bezier path. It's not immediately intuitive that a clip_path is much more challenging than a mask that's "straightforward".

sparse_strips/vello_hybrid/src/render.rs
taj-p: Very nice! Since the widths are the same we can simply compare height. Great insight. I also changed ` * 16` to ` << 4` in https://github.com/linebender/vello/pull/957/commits/b7dbcecd85b654e76da1de001c0d2dc37ca896c8

sparse_strips/vello_hybrid/src/render.rs
taj-p: Fixed in https://github.com/linebender/vello/pull/957/commits/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/render.rs
taj-p: Fixed in https://github.com/linebender/vello/commit/41f78ba77f528202bf4b247de8e518511ff078d5

sparse_strips/vello_hybrid/src/render.rs
taj-p: Added in https://github.com/linebender/vello/pull/957/commits/e137620194c1a8bad486f124f7df21d346acd734

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -415,23 +672,118 @@ impl Renderer {
...
+        self.programs.upload_strips(self.device, self.queue, strips);
+
+        let mut render_pass = self.encoder.begin_render_pass(&RenderPassDescriptor {
+            label: Some("Render to Texture Pass"),
+            color_attachments: &[Some(RenderPassColorAttachment {
+                view: if ix == 2 {
+                    self.view
+                } else {
+                    &self.programs.resources.slot_texture_views[ix]
+                },
```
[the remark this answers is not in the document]
taj-p: https://github.com/linebender/vello/pull/957/commits/e137620194c1a8bad486f124f7df21d346acd734

sparse_strips/vello_hybrid/src/render.rs
taj-p: Skipped in https://github.com/linebender/vello/pull/957/commits/ccb0ea69b442c7f5581a4da350209700396dfcf7

sparse_strips/vello_sparse_tests/tests/clip.rs
```diff
@@ -6,12 +6,36 @@
...
+    const WIDTH: f64 = 100.0;
+    assert!(WIDTH <= WideTile::WIDTH as f64, "Width larger than a tile");
+    const HEIGHT: f64 = Tile::HEIGHT as f64;
+    const OFFSET: f64 = WIDTH / 3.0;
+
+    let colors = [RED, GREEN, BLUE];
+
+    for (i, color) in colors.iter().enumerate() {
+        let clip_rect = Rect::new((i as f64) * OFFSET, 0.0, WIDTH, HEIGHT);
+        ctx.push_clip_layer(&clip_rect.to_path(0.1));
```
ajakubowicz-canva, line 30: Nit: Worth defining `0.1` as `DEFAULT_TOLERANCE` (similar to vello_cpu)?

sparse_strips/vello_hybrid/src/render.rs
```diff
@@ -165,18 +226,64 @@ impl Renderer {
...
+                        ty: wgpu::BufferBindingType::Uniform,
+                        has_dynamic_offset: false,
+                        min_binding_size: None,
+                    },
+                    count: None,
+                }],
+            });
+
+        let strip_shader = device.create_shader_module(wgpu::ShaderModuleDescriptor {
+            label: Some("Strip Shader"),
```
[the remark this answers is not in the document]
taj-p: I didn't like the `render_*` prefix because `render_` could arguably be put in front of so many classes of things. I also didn't like renaming the shaders to `strips`. I'll sleep on it, but if anyone has a better name, please let me know 🙏 . Maybe "fine" for fine rasterisation (but arguably that includes "clear_slots" too).

sparse_strips/vello_sparse_tests/tests/clip.rs
```diff
@@ -6,12 +6,36 @@
...
+        ctx.set_paint(*color);
+        ctx.fill_rect(&Rect::new(0.0, 0.0, WIDTH, HEIGHT));
+    }
+    for _ in colors.iter() {
+        ctx.pop_layer();
```
ajakubowicz-canva, line 35: So the caller needs to make sure there are exactly the same amount of `pop_layer` calls as `push_clip_layer` calls. 

What if there is a pop layer that has been missed? It looks like if there is an extra pop layer we'll panic.

sparse_strips/vello_hybrid/src/render.rs
taj-p: I don't think this is a naive question. We only support Rgba8Unorm and Bgra8Unorm currently I believe. This could be a separate `RenderError`!

sparse_strips/vello_sparse_tests/tests/renderer.rs
ajakubowicz-canva, line 1: When I naively run `cargo test` within `vello_sparse_tests`, the clip hybrid tests appear to be ignored:

```
test clip::clip_deeply_nested_circles_hybrid ... ignored
test clip::clip_rectangle_and_circle_hybrid ... ignored
test clip::clip_rectangle_with_star_evenodd_hybrid ... ignored
test clip::clip_rectangle_with_star_nonzero_hybrid ... ignored
test clip::clip_single_wide_tile_hybrid ... ignored
test clip::clip_transformed_rect_hybrid ... ignored
test clip::clip_triangle_with_star_hybrid ... ignored
test clip::clip_with_multiple_transforms_hybrid ... ignored
test clip::clip_with_opacity_hybrid ... ignored
test clip::clip_with_rotate_hybrid ... ignored
test clip::clip_with_save_restore_hybrid ... ignored
test clip::clip_with_scale_hybrid ... ignored
test clip::clip_with_translation_hybrid ... ignored
```

Is this intended? Is this because these snapshots don't align between cpu and hybrid approaches?

For example I got diff image:
clip_deeply_nested_circles_hybrid

sparse_strips/vello_sparse_tests/tests/renderer.rs
```diff
@@ -164,16 +163,16 @@ impl Renderer for Scene {
...
     fn push_layer(
         &mut self,
-        _: Option<&BezPath>,
+        clip: Option<&BezPath>,
         _: Option<BlendMode>,
         _: Option<u8>,
         _: Option<Mask>,
     ) {
-        unimplemented!()
+        Self::push_layer(self, clip, None, None, None);
```
ajakubowicz-canva, line 171: Is it worth passing all arguments through and hitting the `unimplemented!` block within `push_layer`?

sparse_strips/vello_sparse_tests/tests/renderer.rs
taj-p: Oops - that was caused via a bad merge. Great pickup - thank you. Re-enabled in https://github.com/linebender/vello/pull/957/commits/47119394dd9d98c24a16808bbfbd3526ab89a69b. We currently don't know why there is a difference between CPU and Hybrid (see https://github.com/linebender/vello/pull/970).

sparse_strips/vello_sparse_tests/tests/clip.rs
taj-p: Yup - we panic 😅 . I think this is OOS of this PR however.

sparse_strips/vello_sparse_tests/tests/clip.rs
taj-p: Yes, definitely. But since that touches every file in `tests/*`, let's leave it for a separate PR.

sparse_strips/vello_sparse_tests/tests/renderer.rs
taj-p: Yes, definitely. Fixed in https://github.com/linebender/vello/pull/957/commits/cbf6ba37c2616140dde73e81b10c339df305edb2

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Clipping is basically a special case of an alpha mask, where everything inside of the clip shape is fully opaque in the alpha mask, and everything outside if fully transparent. So you could in theory emulate it with an alpha mask, but since clipping is such a common case we have a custom logic that is more complex, but faster.

ajakubowicz-canva (commented):
I have now technically read over all the code and it looks great! I will need to give the schedular a second pass as there are still some mysteries in there.

Unfortunately as someone relatively new to this repo, my comments have been focused on nits and syntax, and less on the overall design. However, overall design seems to be working per tests.

Great work!

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,484 @@
...
+            tile_state: Default::default(),
+        }
+    }
+
+    pub(crate) fn do_scene(
+        &mut self,
+        junk: &mut RendererJunk<'_>,
+        scene: &Scene,
+    ) -> Result<(), RenderError> {
+        let mut tile_state = mem::take(&mut self.tile_state);
```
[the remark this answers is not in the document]
taj-p: Empty vectors in rust do not allocate, so taking the memory and returning it later allows us to reuse the allocation without creating more suggestions

sparse_strips/vello_hybrid/shaders/render_strips.wgsl
```diff
@@ -108,18 +113,28 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
-    return alpha * unpack4x8unorm(in.color);
+    // Apply the alpha value to the unpacked RGBA color or slot index
+    let alpha_byte = in.rgba_or_slot >> 24u;
+    if alpha_byte != 0 {
+        // in.rgba_or_slot encodes a color
+        return alpha * unpack4x8unorm(in.rgba_or_slot);
+    } else {
+        // in.rgba_or_slot encodes a slot in the source clip texture
+        let clip_x = u32(in.position.x) & 0xFFu;
+        let clip_y = (u32(in.position.y) & 3) + in.rgba_or_slot * config.strip_height;
```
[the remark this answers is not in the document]
ajakubowicz-canva: I think the answer to this question should also potentially result in a code comment in the wgsl shader.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,484 @@
...
+/// clip/blend buffers, and the third for the actual render target.
+#[derive(Debug, Default)]
+struct Round {
+    /// Draw calls scheduled into the two slot textures (0, 1) and the final target (2).
+    draws: [Draw; 3],
+    /// Slots that will be freed after drawing into the two slot textures [0, 1].
+    free: [Vec<usize>; 2],
+}
+
+// State for a single wide tile.
```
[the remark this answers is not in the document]
taj-p: Fixed in https://github.com/linebender/vello/pull/957/commits/007ae057f3e7b7dd69e64e2d51bd0c79fd405782

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,484 @@
...
+        wide_tile_x: u16,
+        wide_tile_y: u16,
+        tile: &WideTile,
+        state: &mut TileState,
+    ) -> Result<(), RenderError> {
+        state.stack.clear();
+        // Sentinel `TileEl` to indicate the end of the stack where we draw all
+        // commands to the final target.
+        state.stack.push(TileEl {
+            slot_ix: !0,
```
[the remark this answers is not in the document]
taj-p: Changed in https://github.com/linebender/vello/pull/957/commits/007ae057f3e7b7dd69e64e2d51bd0c79fd405782

sparse_strips/vello_hybrid/shaders/render_strips.wgsl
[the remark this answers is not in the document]
taj-p: `in.position` represents the position `builtin`. In a vertex shader, it ranges from -1 to 1 for X and Y. In the context of a fragment shader, it represents the pixel coordinate of where we are drawing (see this [article](https://arc.net/l/quote/pglqfmbt)). The `& 3` is used to constrain the pixel coordinate to 4 since that's the height of our tile. I added a `CAUTION:` note to the `config.strip_height` about the danger in changing its value without updating this logic. 

In time, we will want to make this configurable, but I'm not sure how that will present. We could make the `& 3` configurable, but then we should also make the `& 0xFFu` configurable to wide tile width. I think at this stage we should untangle those concerns when we get to them.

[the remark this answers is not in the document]
ajakubowicz-canva: Oh whoops, I got my x and y's confused! Thanks for the great answer :D

ajakubowicz-canva (approved):
Great work!

LaurenzV (approved):
Just rubber stamping since we discussed in office hours that this is fine to merge, and Andrew doesn't seem to have write permissions yet.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,484 @@
...
+//! │ ───                ────    │  /     /       /     /     │       ──────────────       │
+//! │                  ──        │ /     /       /     /      │───────                     │
+//! └────────────────────────────└────────────────────────────└────────────────────────────┘
+//!
+//! ### Draw to render target
+//!
+//! At this point, we can sample the pixels from slot N - 1 of texture 1 to draw the final
+//! result.
+//!
+//! ## Nuances
```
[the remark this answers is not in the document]
taj-p: Added notes in https://github.com/linebender/vello/pull/957/commits/c1e909727253be0ee5a2362721aec5f0a3de34ff

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -0,0 +1,504 @@
...
+                    let draw = self.draw_mut(el.round, clip_depth);
+                    let color = match fill.paint {
+                        Paint::Solid(color) => color,
+                        Paint::Indexed(_) => unimplemented!(),
+                    };
+                    let rgba = color.as_premul_rgba8().to_u32();
+                    debug_assert!(
+                        has_non_zero_alpha(rgba),
+                        "Color fields with 0 alpha are reserved for clipping"
+                    );
```
ajakubowicz-canva, line 378: Raised in [Zulip](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Vello.20Hybrid.20Crashing.20when.20alpha.20.3D.200/near/516919551)[#vello > Vello Hybrid Crashing when alpha = 0 @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Vello.20Hybrid.20Crashing.20when.20alpha.20.3D.200/near/516919551).

What is the recommendation if someone wants to pass a fully transparent fill? Additionally, is there a use-case of painting something fully transparent?

PR #963, opened 2025-05-06
Round vertical hinting offset in Vello Classic

DJMcNab:
See [#vello > Hinting: y-alignment](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Hinting.3A.20y-alignment/with/515265682).

This matches the code in Vello Common.

### DIFF
--- a/vello_encoding/src/resolve.rs
+++ b/vello_encoding/src/resolve.rs
@@ -422,6 +422,8 @@ impl Resolver {
                             scale = transform.matrix[0];
                             font_size *= scale;
                             transform.matrix = [1.0, 0.0, 0.0, 1.0];
+                            // Hinting assumes that the baseline is rounded; otherwise, we are missing the point of hinting.
+                            transform.translation[1] = transform.translation[1].round();
                         } else {
                             hint = false;
                         }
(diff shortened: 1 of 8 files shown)


tomcur (approved):
Assuming this is the final, global transform, this should indeed be doing the same as Vello Common: 

https://github.com/linebender/vello/blob/570b780d9203a53e6a6a2402f28564f88fff7d3e/sparse_strips/vello_common/src/glyph.rs#L149-L154

vello_encoding/src/resolve.rs
```diff
@@ -422,6 +422,8 @@ impl Resolver {
                             scale = transform.matrix[0];
                             font_size *= scale;
                             transform.matrix = [1.0, 0.0, 0.0, 1.0];
+                            // Hinting assumes that the baseline is rounded; otherwise, we are missing the point of hinting.
+                            transform.translation[1] = transform.translation[1].round();
```
dfrg, line 426: The glyph transform is applied later (when emitting the transform stream in resolve) and the translation is also scaled there so I think that’s the proper place to apply rounding.

PR #964, opened 2025-05-06
Improve `Pixmap` API to uphold conditions

tomcur:
Don't allow direct field access, make it clearer the data is premultiplied RGBA8.


sparse_strips/vello_api/src/pixmap.rs
DJMcNab, line 1: Is it still possible to make a PixMap, given a `Vec<u8>`, its width and its height? It doesn't look like there's a suitable constructor function to me.

sparse_strips/vello_api/src/pixmap.rs
```diff
@@ -6,15 +6,17 @@
...
 pub struct Pixmap {
     /// Width of the pixmap in pixels.  
-    pub width: u16,
+    width: u16,
     /// Height of the pixmap in pixels.
-    pub height: u16,
+    height: u16,
     /// Buffer of the pixmap in RGBA format.
-    pub buf: Vec<u8>,
+    buf: Vec<u8>,
```
DJMcNab, line 19: I'd love to make this `Vec<PremulRgba8>` :upside_down_face:

sparse_strips/vello_hybrid/examples/render_to_file.rs
```diff
@@ -157,28 +157,26 @@ async fn run() {
...
-    for row in texture_copy_buffer
+    let mut pixmap = Pixmap::new(width, height);
+    for (row, buf) in texture_copy_buffer
         .slice(..)
         .get_mapped_range()
         .chunks_exact(bytes_per_row as usize)
+        .zip(pixmap.data_mut().chunks_exact_mut(usize::from(width) * 4))
     {
-        img_data.extend_from_slice(&row[0..width as usize * 4]);
+        buf.copy_from_slice(&row[0..usize::from(width) * 4]);
```
DJMcNab, line 167: This now starts zeroed and writes over, right? Rather than before where it started uninitialised, and writes once.

Given the context of this example, the difference is probably fine?

DJMcNab (commented):
I worry that the lack of replacement for the old struct constructor is a meaningful downgrade for some use cases.

I'll defer to Laurenz for approval

sparse_strips/vello_api/src/pixmap.rs
tomcur: Not yet, I was thinking of having something like `into_parts` and `from_parts`, but the constructor is fallible as it has to check the size of the `Vec`. Perhaps `try_from_parts`?

sparse_strips/vello_api/src/pixmap.rs
tomcur: I agree, I think we can do that; that's part of the motivation of this PR. I first wanted us to decide on the API direction more generally.

sparse_strips/vello_hybrid/examples/render_to_file.rs
tomcur: That's right. For a one-shot case like this that may be a bit of a performance hit. With a constructor as in https://github.com/linebender/vello/pull/964/files#r2075398135 we could do the same as before but have it checked.

tomcur (commented):
(Note Bytemuck is already used in `vello_common`.)

With the idea of using `Vec<PremulRgba8>` as in https://github.com/linebender/vello/pull/964/files#r2075400876, we may want methods to/from `&[u8]` as well.

sparse_strips/vello_api/src/pixmap.rs
tomcur: Proposed a constructor in https://github.com/linebender/vello/pull/964/commits/bd986e7481c608a06a9f1b05ae5c5244acb172f8.

sparse_strips/vello_hybrid/examples/render_to_file.rs
tomcur: Changed back with such a constructor, https://github.com/linebender/vello/pull/964/commits/f7bfd3787442413452ca396ffc01ae3e8d285f31.

sparse_strips/vello_api/src/pixmap.rs
```diff
@@ -122,38 +140,60 @@ impl Pixmap {
...
-    #[expect(
-        clippy::cast_possible_truncation,
-        reason = "cannot overflow in this case"
-    )]
-    pub fn unpremultiply(&mut self) {
+    ///
+    /// The pixels are in row-major order. Each pixel consists of four bytes in the order
+    /// `[r, g, b, a]`.
+    pub fn take_unpremultiplied(mut self) -> Vec<u8> {
         for rgba in self.buf.chunks_exact_mut(4) {
```
DJMcNab, line 186: (Idle thought: I wonder if
```rust
        for [r, g, b, a] in self.buf.chunks_exact_mut(4) {
```
can work. I don't have a good intuition for when the exclusive reference is decomposed in patterns)

tomcur: I think that would work with [`slice::as_chunks_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.as_chunks_mut), which returns a mutable slice of const-sized arrays (`&mut [[T; N]]`). That's nightly-only. [`slice::chunks_exact_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.chunks_exact_mut) is an iterator over slices.

DJMcNab: Ah, of course. You'd need to use `TryFrom` to make the array of size 4

sparse_strips/vello_api/src/pixmap.rs
```diff
@@ -24,6 +26,22 @@ impl Pixmap {
...
 
+    /// Create a new pixmap with the given underlying data interpreted as premultiplied RGBA8.
+    ///
+    /// If `data` is not of the right length (`width * height * 4`). If `data` has to grow, the new
+    /// color components are filled with `0`.
+    ///
+    /// The pixels are in row-major order. Each pixel consists of four bytes in the order
+    /// `[r, g, b, a]`.
+    pub fn from_parts(mut data: Vec<u8>, width: u16, height: u16) -> Self {
+        data.resize(usize::from(width) * usize::from(height) * 4, 0);
```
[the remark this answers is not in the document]
tomcur: Agreed. I've made the `from_parts` constructor fallible (with a panic), otherwise we should probably go with a `try_from_parts`. Use-case 2 can be achieved by the user with this API without providing the specialized constructor (and becomes relatively simple once the pixmap is a buffer over `PremulRgba8`).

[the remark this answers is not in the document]
tomcur: Some additional methods for managing the pixmap buffer can be useful, I'll open a separate PR once this one is agreed on.

sparse_strips/vello_api/src/pixmap.rs
```diff
@@ -125,38 +150,60 @@ impl Pixmap {
...
+    /// The pixel data is [premultiplied RGBA8][PremulRgba8].
     #[inline(always)]
-    pub fn sample(&self, x: u16, y: u16) -> &[u8] {
+    pub fn sample(&self, x: u16, y: u16) -> PremulRgba8 {
         let idx = 4 * (self.width as usize * y as usize + x as usize);
-        &self.buf[idx..][..4]
+
+        // The conversion here is a no-op, as `PremulRgba8`'s fields are in the same memory order
+        // as the pixel data.
+        PremulRgba8::from_u8_array(self.buf[idx..][..4].try_into().unwrap())
```
LaurenzV, line 177: Did you actually check that in godbolt or so?

tomcur: Yes, not Godbolt, but cargo-show-asm. You can use an invocation like the following.

```
cargo asm -p vello_api --lib vello_api::pixmap::Pixmap::sample --rust --this-workspace
```

tomcur: Yes, not Godbolt, but cargo-show-asm. You can use an invocation like the following.

```
cargo asm -p vello_api --lib vello_api::pixmap::Pixmap::sample --rust --this-workspace
```

PR #965, opened 2025-05-06
Let `Pixmap::from_png` use `Read` directly

tomcur:
This allows decoding from a `Read` directly without first having to copy to memory.

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/image.rs
+++ b/sparse_strips/vello_sparse_tests/tests/image.rs
@@ -15,7 +15,7 @@ use vello_dev_macros::vello_test;
 
 pub(crate) fn load_image(name: &str) -> Arc<Pixmap> {
     let path = Path::new(env!("CARGO_MANIFEST_DIR")).join(format!("tests/assets/{name}.png"));
-    Arc::new(Pixmap::from_png(&std::fs::read(path).unwrap()).unwrap())
+    Arc::new(Pixmap::from_png(std::fs::File::open(path).unwrap()).unwrap())
 }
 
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -15,7 +15,7 @@ use vello_dev_macros::vello_test;
 
 pub(crate) fn load_image(name: &str) -> Arc<Pixmap> {
     let path = Path::new(env!("CARGO_MANIFEST_DIR")).join(format!("tests/assets/{name}.png"));
-    Arc::new(Pixmap::from_png(&std::fs::read(path).unwrap()).unwrap())
+    Arc::new(Pixmap::from_png(std::fs::File::open(path).unwrap()).unwrap())
```
DJMcNab, line 18: I don't know if this is a use case where we should use `BufReader`. It looks like png's docs don't suggest that, so it's probably fine. And this is only for tests anyway.

DJMcNab (approved):

sparse_strips/vello_sparse_tests/tests/image.rs
tomcur: `png` seems to use `BufReader` under the hood, but it's not documented: https://docs.rs/png/0.17.16/src/png/decoder/read_decoder.rs.html#20-23.

nicoburns:
I don't object to this API, but we shouldn't assume that this will be faster. I've had a number of experiences where reading into memory then decoding was faster, even though it implies an extra copy, and even compared to using a `BufReader`.

PR #968, opened 2025-05-06
Fix the wgpu-profiler integration

DJMcNab:
This accidentally bitrotted when we implemented #803

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -497,6 +497,18 @@ impl Renderer {
             #[cfg(feature = "wgpu-profiler")]
             &mut self.profiler,
         )?;
+        // N.B. This is horrible; this integration of wgpu-profiler really needs some work...
+        #[cfg(feature = "wgpu-profiler")]
+        {
+            self.profiler.end_frame().unwrap();
+            if let Some(result) = self
+                .profiler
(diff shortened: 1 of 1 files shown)


waywardmonkeys (commented):
Really dislike that this is here since it makes it useless for anyone outside of the sample code, but maybe this feature already isn't useful outside of the sample code?

(Useless because in many cases, you'd have people rendering multiple textures with Vello that aren't ending a frame...)

DJMcNab:
FWIW, the use of the term "frame" here in the wgpu-profiler docs doesn't necessarily mean "frame" as in a vsync interval, it more means frame as in packet of information. Each render being a single wgpu-profiler "frame" I think is the most useful here, because it lets them be directly compared. And you can always merge multiple frames if you wanted to.

I've found it useful in Xilem, but only for isolating which stage was being super slow (unsurprisingly, it was flatten).

Cleaning up this API was the motivation for [#vello > Making wgpu-profiler non-optional](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Making.20wgpu-profiler.20non-optional/with/502099078), but that's something to reason about later (the correct solution might be to have an internal never-enabled profiler, and use that when there isn't a user provided one).

PR #969, opened 2025-05-06
`cargo update && cargo upgrade`

DJMcNab:
Of note, we depend explicitly on [wgpu v24.0.3](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md#v2403-2025-03-19) now.

This is most important because of the surface drop order change in v24.0.2


nicoburns (approved):
Removing the drop order requirement is nice!

PR #967, opened 2025-05-06
Add f32 fine-rasterization pipeline

LaurenzV:
Hopefully the last time we have to regenerate all images, but I can't promise it. :)

Overall, while I don't think the `FineType` is very polished it, as you can see it does work pretty well and makes it possible to define each pipeline once for both, u8 and f32 without having to duplicate code. I am optimistic that this design could be further evolved to also support SIMD, though that remains to be seen.

I would also like to test whether we can make do with a default threshold of 0 for the  hybrid renderer, but since the hybrid renderer currently converts colors to u8 (and then back to f32 on the GPU), this currently is not possible. If anyone has a git diff of changes I need to apply to get that to work, I'm happy to try it.

PR #970, opened 2025-05-06
Add support for float colours on the GPU

DJMcNab:
As requested by @LaurenzV in #967.

This probably shouldn't be merged. Note that a tolerance of zero still doesn't work, for unknown reasons.

One potential reason is that the GPU is getting better precision; because it might do blending in f32. But I don't know.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -478,6 +478,7 @@ version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "010263546cea9f9f8385a5b7aad534b9e6448e62a0d3bf9da29d583308dd11bb"
 dependencies = [
+ "bytemuck",
  "libm",
 ]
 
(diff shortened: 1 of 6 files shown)


DJMcNab:
(The messages for those test failures could be a lot more illuminating...)

LaurenzV:
Using my branch I'm only getting 14 failures with 0 tolerance, and they are mostly at the edges, e.g.
image

However, there is one image where the difference is quite big:
image

Will see whether I can investigate what is going on here.

LaurenzV:
Guess we just got unlucky... At some point we end up with the number 0.7509804 on the CPU, which multiplies to 191.500002, and therefore gets rounded up to 192. On the GPU, it gets rounded down apparently. Not sure there is much point in investigating this further though, so I think the best thing would be to just stick with the default tolerance threshold of 1.

It might still be nice to merge the change in this PR at some point, for the sake of consistency across the renderers.

DJMcNab:
This isn't fundamentally broken, but isn't going anywhere anytime soon

PR #971, opened 2025-05-06
Add test for image roundtrip

sagudev:
Repro for #972. It's currently set to should fail as fixup is still in the works.

Why that particular image? Because it makes vello not work (due too math errors). It was generated by WPT (while testing servo) and has encoded JSON about cross origin stuff.


vello_tests/snapshots/data_image_roundtrip.png
DJMcNab, line 1: Why does this exist? I think we should just be able to compare against the "source of truth".

vello_tests/tests/data_image.png
DJMcNab, line 1: I think I'd like this to be in a `vello_tests/assets` folder.

DJMcNab, line 1: What license is this user under? Do we need to follow any conditions from that license?

vello_tests/tests/regression.rs
```diff
@@ -24,3 +25,22 @@ fn rounded_rectangle_watertight() {
         .unwrap()
         .assert_mean_less_than(0.001);
 }
+
+const DATA_IMAGE_PNG: &[u8] = include_bytes!("./data_image.png");
+
+#[test]
+#[should_panic(expected = "Expected mean to be less than 0.001, got 0.3072978")]
```
DJMcNab, line 32: I'm surprised by that expected number being so specific

sagudev: I just copy&pasted full panic msg because I forget that panic does not need to match fully.

vello_tests/tests/data_image.png
sagudev: I think the best would be to simply use snapshot as src, so we keep it in snapshot folder, like I did in https://github.com/linebender/vello/pull/971/commits/ece0dfdb414fc671ac55c75b3c7a43908b9a9200.

vello_tests/snapshots/data_image_roundtrip.png
DJMcNab, line 1: Two things:
1) How did you push this file? It is in the LFS managed folder, but isn't an LFS file(?)
2) This shouldn't be an LFS file, so should probably go in the smoke snapshots folder (test input should never be in LFS, because we want to be able to migrate our test suite away from LFS if needed, given GitHub's willingness to create cliffs around LFS).

vello_tests/tests/regression.rs
```diff
@@ -24,3 +25,23 @@ fn rounded_rectangle_watertight() {
         .unwrap()
         .assert_mean_less_than(0.001);
 }
+
+const DATA_IMAGE_PNG: &[u8] = include_bytes!("../snapshots/data_image_roundtrip.png");
+
+/// Test for <https://github.com/linebender/vello/issues/972>
+#[test]
+#[should_panic(expected = "Expected mean to be less than 0.001, got 0.3")]
```
DJMcNab, line 33:
```suggestion
// Note that this isn't exactly 0.3, it's a number starting with "0.3"
#[should_panic(expected = "Expected mean to be less than 0.001, got 0.3")]
```

vello_tests/snapshots/data_image_roundtrip.png
sagudev: Right, I am not that familiar with vello test structure yet.

DJMcNab (commented):
Sorry to keep coming back with nits - this should be the last one.

This test should actually be in `known_issues.rs`, not `regression.rs`. Once it becomes fixed (and stops panicking), we'd move it to regression.rs.

DJMcNab (approved):
Thank you!

Issue #972, opened 2025-05-06
Images do not roundtrip


sagudev:
When drawing image without any transformations one would expect that output image would be same as input image, but that's not the case for vello right now.

The problem lies in math errors of: https://github.com/linebender/vello/blob/48e658ab0acc00659f0940cd750a2e3af4775230/vello_shaders/shader/fine.wgsl#L1166-L1167

[#vello > Servo 2D canvas backend @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Servo.202D.20canvas.20backend/near/516345856)

DJMcNab:
To update this issue: in #974, this has been fixed for [`Pad`](https://docs.rs/peniko/0.4.0/peniko/enum.Extend.html#variant.Pad), but not repeat or reflect.

DJMcNab:
#974 closing this was a false-positive of GitHub's detection

Keavon:
I took a shot at this since it has been a problem for us in Graphite. Here is the PR: #1499.

PR #973, opened 2025-05-07
Add snapshot tests for image_extend_modes test scenes

sagudev:
They will be needed for fixing #972.


sagudev:
Is there any reason why all test scenes do not have snapshot tests?

DJMcNab:
For image extend mode specifically, we disabled them in #766, because we weren't getting consistent results cross-platform. I don't know why that is no longer applying.

My best guess for the other test scenes not being there is sloth - I got bored partway through adding them.

DJMcNab (approved):
I think this makes sense to have, even with the known deficiencies.

PR #974, opened 2025-05-07
`extend_mode_normalized` and `extend_mode`

### LINKED ISSUE #972: Images do not roundtrip

sagudev:
Current `extend_mode` is renamed to `extend_mode_normalized` (because it require normalized coordinates). To avoid rounding errors we introduce `extend_mode` that relaxes normalization requirement.

Fixes #972


vello_tests/tests/known_issues.rs
```diff
@@ -77,24 +76,3 @@ fn many_bins_gpu() {
...
-        .from_bytes(0, DATA_IMAGE_PNG)
-        .unwrap()
-        .with_quality(ImageQuality::Low);
-    scene.draw_image(&image, Affine::IDENTITY);
-    let mut params = TestParams::new("data_image_roundtrip", image.width, image.height);
-    params.anti_aliasing = AaConfig::Area;
-    smoke_snapshot_test_sync(scene, &params)
-        .unwrap()
-        .assert_mean_less_than(0.001);
-}
```
DJMcNab, line 100: We could consider duplicating this test for the other extend modes

DJMcNab (commented):
One thought on testing. Once that's addressed, I'll be happy to land. I don't think this is going to be meaningfully reviewed by anyone else anytime soon.

sagudev:
It works on my local ubuntu machine but not in CI. Is this the case for `#[cfg_attr(skip_gpu_tests, ignore)]`?

DJMcNab:
The windows failure certainly would be caused by the lack of GPU support for those runners. I'm not sure what the ubuntu failure is. Unfortunately, #892 means that the "current version" uploading of the snapshots is currently broken. I'll open a PR to fix that today.

DJMcNab:
And if you can rebase https://github.com/linebender/vello/pull/1018, that would make debugging the failure on ubuntu more plausible to debug

DJMcNab:
I'm assuming that the failure on Ubuntu is because we're using a fallback CPU implementation of Vulkan, and therefore the precision is too high. What isn't at all clear to me is why the similar test from #971 passed on that version.

I think just marking the new tests as `#[ignore]` unconditionally is the least bad option to unblock us. We should however setup the usual `skip_gpu_tests` for them, so it doesn't fail on windows.

vello_tests/tests/known_issues.rs
```diff
@@ -82,15 +82,36 @@ const DATA_IMAGE_PNG: &[u8] = include_bytes!("../snapshots/smoke/data_image_roun
 
 /// Test for <https://github.com/linebender/vello/issues/972>
 #[test]
-// Note that this isn't exactly 0.3, it's a number starting with "0.3"
-#[should_panic(expected = "Expected mean to be less than 0.001, got 0.3")]
-fn test_data_image_roundtrip() {
+#[ignore]
```
DJMcNab, line 85: I'd like to see an ignore reason here. Something like "CI runs these tests on a CPU, leading to them having unrealistic precision".

DJMcNab (approved):
One small detail then we can land this. Thanks!

PR #977, opened 2025-05-07
Remove hack for tolerance of stroke expansion

LaurenzV:
This was needed in the initial prototype for `vello_cpu` where we were using the stroke expansion implementation of the GPU stroke expansion paper, which approximates the stroke by line segments. We had to adjust the tolerance here, because if you draw a small stroke which is then scaled up, you can very clearly notice the line approximation.

However, the current (and future) kurbo stroke do cubic-to-cubic stroke expansion, so I don't think this should be necessary here?

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -111,10 +111,7 @@ pub fn fill(path: &BezPath, affine: Affine, line_buf: &mut Vec<Line>) {
 
 /// Flatten a stroked bezier path into line segments.
 pub fn stroke(path: &BezPath, style: &Stroke, affine: Affine, line_buf: &mut Vec<Line>) {
-    // TODO: Temporary hack to ensure that strokes are scaled properly by the transform.
-    let tolerance = TOL / affine.as_coeffs()[0].abs().max(affine.as_coeffs()[3].abs());
-
-    let expanded = kurbo::stroke(path.iter(), style, &StrokeOpts::default(), tolerance);
+    let expanded = kurbo::stroke(path.iter(), style, &StrokeOpts::default(), TOL);
     fill(&expanded, affine, line_buf);
 }
 


raphlinus:
Cubic-to-cubic stroke expansion is still technically dependent on the scale factor, but the sensitivity is greatly less than when lowering to lines. Thus, it should be possible to retain stroke-expanded paths over a fairly wide range of scale factors. I don't have my head completely around whether it's a good idea to merge this.

LaurenzV:
The main reason I wanted to remove it is because it's causing trouble with transforms where the scaling factor is close to 0 and the skewing factor is instead pretty large. But maybe we need to come up with a more principled approach here.

dominikh:
Would it make sense to use the [determinant](https://docs.rs/kurbo/latest/kurbo/struct.Affine.html#method.determinant) here? That would consider both the scale's and the skew's effect on the area.

LaurenzV:
Perhaps, not an expert on this though. 😄

tomcur (commented):
I've opened #1180 for consideration, though the current PR can absolutely land if it solves some immediate problems.

> Would it make sense to use the [determinant](https://docs.rs/kurbo/latest/kurbo/struct.Affine.html#method.determinant) here? That would consider both the scale's and the skew's effect on the area.

I believe as a conservative metric we could use the Frobenius norm, ignoring translation, as `sqrt(a^2 + b^2 + c^2 + d^2)`.

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -111,10 +111,7 @@ pub fn fill(path: &BezPath, affine: Affine, line_buf: &mut Vec<Line>) {
 
 /// Flatten a stroked bezier path into line segments.
 pub fn stroke(path: &BezPath, style: &Stroke, affine: Affine, line_buf: &mut Vec<Line>) {
-    // TODO: Temporary hack to ensure that strokes are scaled properly by the transform.
-    let tolerance = TOL / affine.as_coeffs()[0].abs().max(affine.as_coeffs()[3].abs());
-
-    let expanded = kurbo::stroke(path.iter(), style, &StrokeOpts::default(), tolerance);
+    let expanded = kurbo::stroke(path.iter(), style, &StrokeOpts::default(), TOL);
```
tomcur: > The main reason I wanted to remove it is because it's causing trouble with transforms where the scaling factor is close to 0 and the skewing factor is instead pretty large. But maybe we need to come up with a more principled approach here.

@LaurenzV Would the issues still be fixed calculating the following?

```suggestion
    let tolerance = TOL / affine.as_coeffs()[0].abs().max(affine.as_coeffs()[3].abs()).max(1.);
    
    let expanded = kurbo::stroke(path.iter(), style, &StrokeOpts::default(), tolerance);
```

This doubles down on the current hack, ensuring the tolerance is always at most `TOL`, only allowing to become less tolerant for scaled-up geometry.

LaurenzV: As far as I can tell that does seem to work as well, so will adopt, thanks!

PR #980, opened 2025-05-08
Feature flag `pico_svg`

nicoburns:


tomcur (approved):
Perhaps `pico_svg` should just be moved to the example code using it (it doesn't really belong in `vello_common`, at least not in its current form), but this is an improvement.

PR #981, opened 2025-05-08
Add tests for Apple Color Emoji

LaurenzV:


sparse_strips/vello_dev_macros/src/lib.rs
```diff
@@ -253,6 +253,12 @@ fn parse_args(attribute_input: &AttributeInput) -> Arguments {
                     "transparent" => args.transparent = true,
                     "skip_cpu" => args.skip_cpu = true,
                     "skip_hybrid" => args.skip_hybrid = true,
+                    "macos_only" => {
+                        if !cfg!(target_os = "macos") {
+                            args.skip_cpu = true;
+                            args.skip_hybrid = true;
+                        }
+                    }
```
DJMcNab, line 261: This isn't used, right? I'd probably remove it.

In theory, this also has issues with cross-compilation of tests. I think you'd actually want to use `cfg_attr(target_os = "macos", ignore)` in the output for this case

DJMcNab (approved):
See also [#vello > wgpu 25 plans @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/wgpu.2025.20plans/near/516873186)

tomcur (approved):

sparse_strips/vello_dev_macros/src/lib.rs
LaurenzV: Ah whoops... Initially I thought `#[cfg(target_os = "macos")]` wouldn't work so i thought I need to do it as a flag, but seems like that's not the case.

vello/src/scene.rs
```diff
@@ -643,10 +644,24 @@ impl<'a> DrawGlyphs<'a> {
...
+                    // rendering if Apple ever does encode the offset directly in
+                    // the font.
+                    let bearing_y = if bitmap.bearing_y == 0.0
+                        && bitmaps.format() == Some(BitmapFormat::Sbix)
+                    {
+                        100.0
+                    } else {
+                        bitmap.bearing_y
+                    };
+
```
nicoburns, line 660: @LaurenzV Does this need to go in `vello_cpu` too?

PR #985, opened 2025-05-08
Fix `clippy::cast_lossless`

tomcur:
This makes it clear that these are lossless conversions.


LaurenzV:
I'm not a 100% sold on this as I think it makes the code much more verbos, but if others think it's good to have then I'm fine with it.

DJMcNab:
I lead towards being in favour here. I wish `as` had better behaviour, but the unfortunate fact is that it's generally best practise to avoid as as often as possible.

LaurenzV (approved):

tomcur (commented):
`as` when `From` is implemented will also be lossless, but using `From` makes that fact absolutely clear. Every `as` causes some overhead when reading as it needs to be scrutinized. I agree it is unfortunate.

PR #987, opened 2025-05-08
Don't unpremultiply alpha for `Mask` calculation

tomcur:
We were multiplying by alpha directly after again.

### DIFF
--- a/sparse_strips/vello_api/src/mask.rs
+++ b/sparse_strips/vello_api/src/mask.rs
@@ -33,21 +33,15 @@ impl Mask {
             if alpha_mask {
                 data.push(pixel.a);
             } else {
-                let mut r = pixel.r as f32 / 255.0;
-                let mut g = pixel.g as f32 / 255.0;
-                let mut b = pixel.b as f32 / 255.0;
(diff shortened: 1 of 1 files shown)


LaurenzV:
Hmm but the specification says:

> The calculations are performed on non-premultiplied color values.

But I guess it makes sense that since you can just extract the /255 they cancel each other out, but then I'm wondering why they have that sentence in the spec.

LaurenzV (approved):

tomcur (commented):
> > The calculations are performed on non-premultiplied color values.
>
> But I guess it makes sense that since you can just extract the /255 they cancel each other out, but then I'm wondering why they have that sentence in the spec.

For the calculations in that section generally it matters, and the luminance-to-alpha as calculated in that section only takes luminance into account, ignoring alpha, which does require the color to be non-premultiplied.

However, to use that value for calculating a mask, you multiply that value by the original alpha again ([Masking Module Level 1 § 7.10.1](https://www.w3.org/TR/css-masking-1/#MaskValues)). In this case that works out to be the same as doing the entire thing with the premultiplied color we already had.

PR #990, opened 2025-05-08
vello_cpu: re-export types used in public api

nicoburns:
These types from `vello_common` are needed in order to use `vello_cpu`s API. Re-exporting them here means consumers of`vello_cpu` don't have to explicitly depend on `vello_common`.


tomcur (approved):
Makes sense!

PR #996, opened 2025-05-08
Implement `Image::from_peniko_image`

tomcur:
Probably a temporary method until there's better resource binding; in the meantime this is useful for clients already using `Peniko::Image`.

Perhaps `Pixmap` should get a method to do the conversion to premultiplied RGBA8 from `&[Rgba8]`, or perhaps `Vec<Rgba8>` using Bytemuck with the `extern_crate_alloc` feature (though the latter doesn't help with `Peniko::Image`, which stores data in an `Arc`).

### DIFF
--- a/sparse_strips/vello_common/src/paint.rs
+++ b/sparse_strips/vello_common/src/paint.rs
@@ -64,6 +64,58 @@ pub struct Image {
     pub quality: ImageQuality,
 }
 
+impl Image {
+    /// Convert a [`peniko::Image`] to an [`Image`].
+    ///
+    /// This is a somewhat lossy conversion, as the image data data is transformed to
+    /// [premultiplied RGBA8](`PremulRgba8`).
+    ///
+    /// # Panics
+    ///
+    /// This panics if `image` has a `width` or `height` greater than `u16::MAX`.
+    pub fn from_peniko_image(image: &peniko::Image) -> Self {
+        if image.format != peniko::ImageFormat::Rgba8 {
+            unimplemented!("Unsupported image format: {:?}", image.format);
+        }
+
+        assert!(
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_api/src/paint.rs
```diff
@@ -68,6 +68,58 @@ pub struct Image {
...
+    ///
+    /// This is a somewhat lossy conversion, as the image data data is transformed to
+    /// [premultiplied RGBA8](`PremulRgba8`).
+    ///
+    /// # Panics
+    ///
+    /// This panics if `image` has a `width` or `height` greater than `u16::MAX`.
+    pub fn from_peniko_image(image: &peniko::Image) -> Self {
+        if image.format != peniko::ImageFormat::Rgba8 {
+            unimplemented!("Unsupported image format: {:?}", image.format);
```
LaurenzV, line 82: Can't we return an Option instead and return `None` in this case?

LaurenzV (approved):
Seems fine as an intermediate solution!

sparse_strips/vello_api/src/paint.rs
tomcur: I'm a bit torn. I think we do ideally have a non-fallible function for this and we should try to make sure to support all the Peniko image formats (currently there's only `Rgba8`). Perhaps `peniko::ImageFormat` growing should be considered breaking, not least because there's no existing mechanism in Peniko to convert it to something known.

tomcur: (Vello Classic also has infallible methods taking `Image` (through `Brush`), https://docs.rs/vello/0.5.0/vello/struct.Scene.html#method.fill).

PR #998, opened 2025-05-09
Skip painting transparent draw commands

taj-p:
Context: [#vello > Vello Hybrid Crashing when alpha = 0 @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Vello.20Hybrid.20Crashing.20when.20alpha.20.3D.200/near/516919551)


taj-p:
cc @LaurenzV - I didn't implement this for CPU because I wanted to get the fix in quickly. Could you let me know if you'd like a similar treatment in `vello_cpu`? I suspect it might be more impacting if we also want to check whether every gradient color stop is transparent.

LaurenzV (approved):
Since it doesn't crash on the CPU I think it's fine for now!

PR #1001, opened 2025-05-09
Test another claimed property of `div_255`

tomcur:
And test the entire domain we claim the properties for.

### DIFF
--- a/sparse_strips/vello_cpu/src/util.rs
+++ b/sparse_strips/vello_cpu/src/util.rs
@@ -36,14 +36,16 @@ pub(crate) mod scalar {
         use crate::util::scalar::div_255;
 
         #[test]
-        fn division() {
-            for i in 0_u16..=(255 * 255) {
+        fn div_255_properties() {
+            for i in 0_u16..=65279 {
                 let expected = i / 255;
                 let actual = div_255(i);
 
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/util.rs
```diff
@@ -36,14 +36,16 @@ pub(crate) mod scalar {
         use crate::util::scalar::div_255;
 
         #[test]
-        fn division() {
-            for i in 0_u16..=(255 * 255) {
+        fn div_255_properties() {
+            for i in 0_u16..=65279 {
```
DJMcNab, line 40: Would `0..(256 * 255)` potentially be clearer here?

sparse_strips/vello_cpu/src/util.rs
```diff
@@ -36,14 +36,16 @@ pub(crate) mod scalar {
...
                 let expected = i / 255;
                 let actual = div_255(i);
 
-                let diff = expected.abs_diff(actual);
+                // In case of a discrepancy, the division always yields a value higher than the original.
+                assert!(expected <= actual);
```
DJMcNab, line 45: These comments could be assert reasons

DJMcNab (approved):
Seems reasonable

PR #1002, opened 2025-05-09
Get rid of Vello API

DJMcNab:
See primarily [#vello > Clearing out Vello API](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Clearing.20out.20Vello.20API/with/517105581). Its purpose was never all that clear, and so publishing it is likely to lead to an incoherent story in the future.


sparse_strips/vello_common/src/lib.rs
```diff
@@ -36,12 +36,18 @@ pub mod blurred_rounded_rect;
...
+pub mod mask;
 pub mod math;
+pub mod paint;
 #[cfg(feature = "pico_svg")]
 pub mod pico_svg;
+pub mod pixmap;
 pub mod strip;
 pub mod tile;
 
-pub use vello_api::*;
```
tomcur, line 47: 🌬️

PR #1003, opened 2025-05-09
Fix `vello_sparse_tests` SIGSEGV when using `cargo test`

tomcur:
On some platforms, running the GPU tests in parallel crashes the testing process more often than not. This "fixes" it for `cargo test` and leaves `cargo nextest` mostly unimpacted.

```bash
$ cargo test --package vello_sparse_tests
# ...
error: test failed, to rerun pass `-p vello_sparse_tests --test tests`

Caused by:
  process didn't exit successfully: `[snip]/vello/target/debug/deps/tests-704ef390c4f325bc` (signal: 11, SIGSEGV: invalid memory reference)
```

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/renderer.rs
+++ b/sparse_strips/vello_sparse_tests/tests/renderer.rs
@@ -223,9 +223,9 @@ impl Renderer for Scene {
         // ensuring only one set of device resources is alive at the same time. This slows down
         // testing when `cargo test` is used.
         //
-        // Testing with `cargo nextest` is not meaningfully slown down. `nextest` runs each test in
-        // its own process (<https://nexte.st/docs/design/why-process-per-test/>), meaning there is
-        // no contention on this mutex.
+        // Testing with `cargo nextest` (as on CI) is not meaningfully slowed down. `nextest` runs
+        // each test in its own process (<https://nexte.st/docs/design/why-process-per-test/>),
+        // meaning there is no contention on this mutex.
         let _guard = {
             use std::sync::Mutex;
             static M: Mutex<()> = Mutex::new(());
--- a/sparse_strips/vello_sparse_tests/tests/renderer.rs
+++ b/sparse_strips/vello_sparse_tests/tests/renderer.rs
@@ -217,6 +217,21 @@ impl Renderer for Scene {
     }
 
     fn render_to_pixmap(&self, pixmap: &mut Pixmap, _: RenderMode) {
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_sparse_tests/tests/renderer.rs
```diff
@@ -217,6 +217,21 @@ impl Renderer for Scene {
     }
 
     fn render_to_pixmap(&self, pixmap: &mut Pixmap, _: RenderMode) {
```
DJMcNab, line 219: Whilst we're here, can we have a comment that this method implementation is awful and should never be used outside of tests? (to be clear, this is as true before this PR as after)

sparse_strips/vello_sparse_tests/tests/renderer.rs
```diff
@@ -217,6 +217,21 @@ impl Renderer for Scene {
...
+        // On some platforms using `cargo test` triggers segmentation faults in wgpu when the GPU
+        // tests are run in parallel (likely related to the number of device resources being
+        // requested simultaneously). This is "fixed" by putting a mutex around this method,
+        // ensuring only one set of device resources is alive at the same time. This slows down
+        // testing when `cargo test` is used.
+        //
+        // Testing with `cargo nextest` is not meaningfully slown down. `nextest` runs each test in
```
DJMcNab, line 226: Hmm. `slown` is not a word I'm familiar with, and I don't think it's standard. It's borderline for inclusion in typos; if I were in charge, I'd include it, but it's the weakest one I've seen recently.

DJMcNab, line 226:
```suggestion
        // Testing with `cargo nextest` (as on CI) is not meaningfully slown down. `nextest` runs each test in
```

DJMcNab (approved):

sparse_strips/vello_sparse_tests/tests/renderer.rs
tomcur: Oops

nicoburns:
The correct fix here is probably to:
1. Have a separate "RenderContext" that holds a single GPU instance and reuses it across scenes
2. Implement Renderer on that type rather than `Scene`

Blitz's WPT runner takes this approach with a "RenderContext" equivalent per-thread stored in a thread local, and we can reliable run ~20k tests in ~40s without issue using this approach.

But this seems reasonable as short-term fix.

PR #1006, opened 2025-05-09
Use `core::ops` instead of `std::ops` in `vello_cpu`

nicoburns:
This ought to make `vello_cpu` no_std compatible with the `png` feature disabled.


nicoburns:
@DJMcNab Maybe just include this in #1008?

DJMcNab:
Can do

PR #1011, opened 2025-05-15
vello_hybrid: add native WebGL backend

ajakubowicz-canva:
### Context

This PR follows the conversation had about https://github.com/linebender/vello/pull/947 . I made this PR separately as it also incorporates the clipping changes https://github.com/linebender/vello/pull/957 .

In short, this PR adds a native WebGL backend when targeting `wasm32` and if using the `"webgl"` feature on `vello_hybrid`.
The **primary motivation** of using a custom webgl renderer is binary size, allowing 3mb to be removed when targeting WebGL2 natively. This is achieved by omitting `wgpu` from the binary when the architecture is `wasm32` and the `"webgl"` feature flag is set on `vello_hybrid`.

#### vello_hybrid examples

- The `webgl` example has been renamed to `wgpu_webgl`. Now it's more clear that it leverages `wgpu`'s WebGL backend.
- A `native_webgl` example has been added which uses the new WebGL renderer backend.
- `ci.yml` tests both the `wgpu_webgl` example and the `native_webgl` example - smoke testing both webgl techniques.
- A new `ClipScene` has been added for manually viewing and testing deeply nested clipping. ([file](https://github.com/linebender/vello/pull/1011/files#diff-ef57b226886dac928b079c4743d6ed1c86ced27637edca1b60c496c95f03479b))

The PR can be manually tested by locally pulling the branch and running the two examples:

- `cargo run_wasm -p wgpu_webgl --release`: Test original example
- `cargo run_wasm -p native_webgl --release`: Test new backend

#### New `vello_sparse_shaders` package added

This new package contains the WGSL shaders as a source of truth. `vello_hybrid` optionally depends on this library which triggers a build step generating a compiled module. The module contains GLSL shader source code, as well as mappings from the WGSL identifiers to the naga-mangled identifiers in the GLSL.

<details><summary>The generated code:</summary>

```rs
// Generated code by `vello_sparse_shaders` - DO NOT EDIT
/// Build time GLSL shaders derived from wgsl shaders.
/// Compiled glsl for `clear_slots.wgsl`
pub mod clear_slots {
    #![allow(missing_docs, reason="No metadata to generate precise documentation forgenerated code.")]

    pub const VERTEX_SOURCE: &str = r###"#version 300 es

precision highp float;
precision highp int;

struct Config {
    uint slot_width;
    uint slot_height;
    uint texture_height;
    uint _padding;
};
uniform Config_block_0Vertex { Config _group_0_binding_0_vs; };

layout(location = 0) in uint _p2vs_location0;

void main() {
    uint vertex_index = uint(gl_VertexID);
    uint index = _p2vs_location0;
    float x = float((vertex_index & 1u));
    float y = float((vertex_index >> 1u));
    uint _e10 = _group_0_binding_0_vs.slot_height;
    float slot_y_offset = float((index * _e10));
    uint _e15 = _group_0_binding_0_vs.slot_width;
    float pix_x = (x * float(_e15));
    uint _e20 = _group_0_binding_0_vs.slot_height;
    float pix_y = (slot_y_offset + (y * float(_e20)));
    uint _e28 = _group_0_binding_0_vs.slot_width;
    float ndc_x = (((pix_x * 2.0) / float(_e28)) - 1.0);
    uint _e37 = _group_0_binding_0_vs.texture_height;
    float ndc_y = (1.0 - ((pix_y * 2.0) / float(_e37)));
    gl_Position = vec4(ndc_x, ndc_y, 0.0, 1.0);
    gl_Position.yz = vec2(-gl_Position.y, gl_Position.z * 2.0 - gl_Position.w);
    return;
}

"###;

    pub mod vertex {
        pub const CONFIG: &str = "Config_block_0Vertex";
    }
    pub const FRAGMENT_SOURCE: &str = r###"#version 300 es

precision highp float;
precision highp int;

struct Config {
    uint slot_width;
    uint slot_height;
    uint texture_height;
    uint _padding;
};
layout(location = 0) out vec4 _fs2p_location0;

void main() {
    vec4 position = gl_FragCoord;
    _fs2p_location0 = vec4(0.0, 0.0, 0.0, 0.0);
    return;
}

"###;
}
/// Compiled glsl for `render_strips.wgsl`
pub mod render_strips {
    #![allow(missing_docs, reason="No metadata to generate precise documentation forgenerated code.")]

    pub const VERTEX_SOURCE: &str = r###"#version 300 es

precision highp float;
precision highp int;

struct Config {
    uint width;
    uint height;
    uint strip_height;
    uint alphas_tex_width_bits;
};
struct StripInstance {
    uint xy;
    uint widths;
    uint col;
    uint rgba_or_slot;
};
struct VertexOutput {
    vec2 tex_coord;
    uint dense_end;
    uint rgba_or_slot;
    vec4 position;
};
uniform Config_block_0Vertex { Config _group_0_binding_1_vs; };

layout(location = 0) in uint _p2vs_location0;
layout(location = 1) in uint _p2vs_location1;
layout(location = 2) in uint _p2vs_location2;
layout(location = 3) in uint _p2vs_location3;
smooth out vec2 _vs2fs_location0;
flat out uint _vs2fs_location1;
flat out uint _vs2fs_location2;

uint unpack_alphas_from_channel(uvec4 rgba, uint channel_index) {
    switch(channel_index) {
        case 0u: {
            return rgba.x;
        }
        case 1u: {
            return rgba.y;
        }
        case 2u: {
            return rgba.z;
        }
        case 3u: {
            return rgba.w;
        }
        default: {
            return rgba.x;
        }
    }
}

vec4 unpack4x8unorm(uint rgba_packed) {
    return vec4((float(((rgba_packed >> 0u) & 255u)) / 255.0), (float(((rgba_packed >> 8u) & 255u)) / 255.0), (float(((rgba_packed >> 16u) & 255u)) / 255.0), (float(((rgba_packed >> 24u) & 255u)) / 255.0));
}

void main() {
    uint in_vertex_index = uint(gl_VertexID);
    StripInstance instance = StripInstance(_p2vs_location0, _p2vs_location1, _p2vs_location2, _p2vs_location3);
    VertexOutput out_ = VertexOutput(vec2(0.0), 0u, 0u, vec4(0.0));
    float x = float((in_vertex_index & 1u));
    float y = float((in_vertex_index >> 1u));
    uint x0_ = (instance.xy & 65535u);
    uint y0_ = (instance.xy >> 16u);
    uint width = (instance.widths & 65535u);
    uint dense_width = (instance.widths >> 16u);
    out_.dense_end = (instance.col + dense_width);
    float pix_x = (float(x0_) + (float(width) * x));
    uint _e31 = _group_0_binding_1_vs.strip_height;
    float pix_y = (float(y0_) + (y * float(_e31)));
    uint _e39 = _group_0_binding_1_vs.width;
    float ndc_x = (((pix_x * 2.0) / float(_e39)) - 1.0);
    uint _e48 = _group_0_binding_1_vs.height;
    float ndc_y = (1.0 - ((pix_y * 2.0) / float(_e48)));
    out_.position = vec4(ndc_x, ndc_y, 0.0, 1.0);
    uint _e65 = _group_0_binding_1_vs.strip_height;
    out_.tex_coord = vec2((float(instance.col) + (x * float(width))), (y * float(_e65)));
    out_.rgba_or_slot = instance.rgba_or_slot;
    VertexOutput _e71 = out_;
    _vs2fs_location0 = _e71.tex_coord;
    _vs2fs_location1 = _e71.dense_end;
    _vs2fs_location2 = _e71.rgba_or_slot;
    gl_Position = _e71.position;
    gl_Position.yz = vec2(-gl_Position.y, gl_Position.z * 2.0 - gl_Position.w);
    return;
}

"###;

    pub mod vertex {
        pub const CONFIG: &str = "Config_block_0Vertex";
    }
    pub const FRAGMENT_SOURCE: &str = r###"#version 300 es

precision highp float;
precision highp int;

struct Config {
    uint width;
    uint height;
    uint strip_height;
    uint alphas_tex_width_bits;
};
struct StripInstance {
    uint xy;
    uint widths;
    uint col;
    uint rgba_or_slot;
};
struct VertexOutput {
    vec2 tex_coord;
    uint dense_end;
    uint rgba_or_slot;
    vec4 position;
};
uniform Config_block_0Fragment { Config _group_0_binding_1_fs; };

uniform highp usampler2D _group_0_binding_0_fs;

uniform highp sampler2D _group_0_binding_2_fs;

smooth in vec2 _vs2fs_location0;
flat in uint _vs2fs_location1;
flat in uint _vs2fs_location2;
layout(location = 0) out vec4 _fs2p_location0;

uint unpack_alphas_from_channel(uvec4 rgba, uint channel_index) {
    switch(channel_index) {
        case 0u: {
            return rgba.x;
        }
        case 1u: {
            return rgba.y;
        }
        case 2u: {
            return rgba.z;
        }
        case 3u: {
            return rgba.w;
        }
        default: {
            return rgba.x;
        }
    }
}

vec4 unpack4x8unorm(uint rgba_packed) {
    return vec4((float(((rgba_packed >> 0u) & 255u)) / 255.0), (float(((rgba_packed >> 8u) & 255u)) / 255.0), (float(((rgba_packed >> 16u) & 255u)) / 255.0), (float(((rgba_packed >> 24u) & 255u)) / 255.0));
}

void main() {
    VertexOutput in_ = VertexOutput(_vs2fs_location0, _vs2fs_location1, _vs2fs_location2, gl_FragCoord);
    float alpha = 1.0;
    uint alphas_index = uint(floor(in_.tex_coord.x));
    if ((alphas_index < in_.dense_end)) {
        uint y = uint(floor(in_.tex_coord.y));
        uvec2 tex_dimensions = uvec2(textureSize(_group_0_binding_0_fs, 0).xy);
        uint alphas_tex_width = tex_dimensions.x;
        uint texel_index = (alphas_index / 4u);
        uint channel_index_1 = (alphas_index % 4u);
        uint tex_x = (texel_index & (alphas_tex_width - 1u));
        uint _e25 = _group_0_binding_1_fs.alphas_tex_width_bits;
        uint tex_y = (texel_index >> _e25);
        uvec4 rgba_values = texelFetch(_group_0_binding_0_fs, ivec2(uvec2(tex_x, tex_y)), 0);
        uint _e31 = unpack_alphas_from_channel(rgba_values, channel_index_1);
        alpha = (float(((_e31 >> (y * 8u)) & 255u)) * 0.003921569);
    }
    uint alpha_byte = (in_.rgba_or_slot >> 24u);
    if ((alpha_byte != 0u)) {
        float _e45 = alpha;
        vec4 _e47 = unpack4x8unorm(in_.rgba_or_slot);
        _fs2p_location0 = (_e45 * _e47);
        return;
    } else {
        uint clip_x = (uint(in_.position.x) & 255u);
        uint _e62 = _group_0_binding_1_fs.strip_height;
        uint clip_y = ((uint(in_.position.y) & 3u) + (in_.rgba_or_slot * _e62));
        vec4 clip_in_color = texelFetch(_group_0_binding_2_fs, ivec2(uvec2(clip_x, clip_y)), 0);
        float _e69 = alpha;
        _fs2p_location0 = (_e69 * clip_in_color);
        return;
    }
}

"###;
    pub mod fragment {
        pub const CONFIG: &str = "Config_block_0Fragment";
        pub const ALPHAS_TEXTURE: &str = "_group_0_binding_0_fs";
        pub const CLIP_INPUT_TEXTURE: &str = "_group_0_binding_2_fs";
    }
}
```

</details>

The generated code can then be imported with:
`use vello_sparse_shaders::{clear_slots, render_strips};`

#### `vello_hybrid` changes

- A new `render` subdirectory has been added that contains:
  - `common.rs`: All the shared render logic.
  - `wgpu.rs`: The original renderer leveraging `wgpu`.
  - `webgl.rs`: The new WebGL native backend renderer.

- The `Scheduler` has been made backend-agnostic by operating on a new `RendererBackend` trait. Both the `wgpu` and `webgl` renderer backends implement `RendererBackend`.

#### Feature flag changes

Feature flags in `vello_hybrid` are additive. By default the `wgpu` feature is enabled. If the compile target is `wasm32` and the `webgl` feature is enabled on `vello_hybrid`, then the native WebGL renderer will be enabled.

#### Warnings

A runtime warning has been added that will trigger once on either renderer being instantiated, if both:
 - `wgpu` with its WebGL backend is active.
 - The `WebGlRenderer` is also active.

The warning is:

```
Both WebGL and wgpu with the "webgl" feature are enabled.
For optimal performance and binary size on web targets, use only the dedicated WebGL renderer.
```

### Screen recording

> [!NOTE]
> The screen recording below is slightly stale – I've since changed the background to be dark so the white text scene can be read.

webgl_native

Left side is `native_webgl` example (using native WebGL2)
Right side is the existing `webgl` example which uses `wgpu` with the `webgl` feature flag.

### Test plan

To scope down this PR, there are no automated tests for the renderer except for the single browser test introduced in the example. The shader compilation has some unit tests.

This PR was manually tested via the new native webgl example: `cargo run_wasm -p native_webgl`. This example can be tested against the original `cargo run_wasm -p wgpu_webgl`.

### Risks

The only risk I'm uncertain about is the addition of the `wgpu` feature flag, that is used as a default feature. Could this be a breaking change for users that specify "no default features". They'd have to add the `wgpu` feature explicitly. This seems minor.

### Followup work

This PR is huge, because it implements all the existing vello_hybrid features in the WebGL backend. Similarly it also includes build-time shader compilation. Instead of making this change completely impenetrable, I'm splitting test infrastructure into a separate change. This PR must be manually tested in the interim.

The example has been added to CI so that it must compile and run.


sparse_strips/vello_hybrid_shaders/src/types.rs
```diff
@@ -0,0 +1,349 @@
...
+        let fragment_reflection_name = format!("{}_fragment_reflection", name);
+        generated_code.push_str(
+            &self
+                .fragment
+                .reflection_map
+                .to_generated_code(&fragment_reflection_name),
+        );
+        generated_code.push('\n');
+
+        // Generate the main function to construct CompiledGlsl.
```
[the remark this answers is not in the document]
ajakubowicz-canva: This is a very good improvement – maybe something we can punt for a followup. I used the vello_shaders build directory as inspiration which used strings. `quote` would be better for this use-case.

sparse_strips/vello_hybrid_shaders/src/types.rs
```diff
@@ -0,0 +1,349 @@
...
+                "  uniforms.insert(\"{}\", \"{}\");\n",
+                wgsl_name, glsl_name
+            ));
+        }
+        generated_code.push_str(
+            "  let mut texture_mapping = alloc::collections::btree_map::BTreeMap::new();\n",
+        );
+        for (wgsl_name, glsl_name) in &self.texture_mapping {
+            generated_code.push_str(&format!(
+                "  texture_mapping.insert(\"{}\", \"{}\");\n",
```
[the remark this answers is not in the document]
ajakubowicz-canva: This case is interesting and it's a great catch by you. This provides no compile time safety to the identifiers, meaning if you change the identifier in wgsl, it's a runtime failure for render_webgl.

If it's acceptable, I'd prefer to punt this in favor of landing the webgl renderer more quickly. But I'm very happy to address this in a standalone PR. Maybe I can make an issue to track improvements that I can fast-follow in a scoped way.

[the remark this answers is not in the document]
ajakubowicz-canva: Just to expand on the rationale – I think it would be useful for any other features in flight to need to consider the webgl renderer in parallel, such as image rendering. Or exploring filter architecture.

Fixing this generated code and the usage of the generated code can be updated as an implementation detail.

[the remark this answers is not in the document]
taj-p: > I think it would be useful for any other features in flight to need to consider the webgl renderer in parallel, such as image rendering. Or exploring filter architecture.

Alex and I are away so I don't believe there's urgency to ship this. I also think that the most important part of this PR (alongside the WebGL render file) is this mapping. I don't believe this idea necessitates much of a time sink. I POC'd a module and const based approach, which I believe is a significant improvement on the hash map approach because:

1. Startup doesn't require creating and accessing a hashmap
2. The consumer doesn't need to remember WGSL resource names (i.e. there's no string literals in consumer code)

You can see it [here](https://github.com/linebender/vello/pull/1013)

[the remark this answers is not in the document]
ajakubowicz-canva: That POC actually really helps cause I would have gone a totally different direction. You've cleared up how we can map the source identifiers. I really like that POC and can definitely integrate it in this PR. One clarifying question, do you think it's enough to just capitalise the `wgsl` identifier (rather than adding the `_UNIFORM` suffix)?

`config` would become `CONFIG`
`alphas_texture` would become `ALPHAS_TEXTURE`

We could add a suffix that is related to the reflection type, but then we'd get:


`config` would become `CONFIG_UNIFORM`
`alphas_texture` would become `ALPHAS_TEXTURE_TEXTURE` (just cause we already use texture in the wgsl name).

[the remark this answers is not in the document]
taj-p: > I'm slightly leaning towards maintaining the wgsl identifiers except capitalised (without any suffix

This SGTM! 

> alphas_texture would become ALPHAS_TEXTURE_TEXTURE

This is a very compelling example 😅 !

sparse_strips/vello_hybrid_shaders/src/types.rs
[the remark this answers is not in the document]
taj-p: Could we please add a `// TODO` in `build.rs` if we don't add it in this PR?

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+struct WebGlPrograms {
+    /// Program for rendering wide tile commands.
+    strip_program: WebGlProgram,
+    /// Uniform locations for the strip program
+    strip_uniforms: StripUniforms,
+    /// Program for clearing slots in slot textures.
+    clear_program: WebGlProgram,
+    /// Uniform locations for the `clear_program`.
+    clear_uniforms: ClearUniforms,
+    /// WebGL resources for rendering (created during prepare).
```
taj-p, line 170:
```suggestion
    /// WebGL resources for rendering.
```

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+
+        gl.bind_buffer(
+            WebGl2RenderingContext::ARRAY_BUFFER,
+            Some(&self.resources.strips_buffer),
+        );
+
+        let strips_data = bytemuck::cast_slice(strips);
+        let strips_array = js_sys::Uint8Array::from(strips_data);
+
+        gl.buffer_data_with_array_buffer_view(
```
taj-p, line 482: Should this be:

```suggestion
        gl.buffer_data_with_js_u8_array(
```

Do you know whether it's more advantageous to do the below (and skip creating the `Uint8Array`)?

```suggestion
        gl.buffer_data_with_u8_array(
```

(Same goes for other buffer uploads)

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+/// Create a WebGL shader program from vertex and fragment sources.
+fn create_shader_program(
+    gl: &WebGl2RenderingContext,
+    vertex_src: &str,
+    fragment_src: &str,
+) -> WebGlProgram {
+    // Compile vertex shader.
+    let vertex_shader = gl
+        .create_shader(WebGl2RenderingContext::VERTEX_SHADER)
+        .expect("Failed to create vertex shader");
```
taj-p, line 499: The docs recommend an expect message specifying why you expect this not to fail as opposed to describing the failure itself

https://doc.rust-lang.org/src/core/option.rs.html#907-908

(Same goes for other expects)

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+            );
+        } else {
+            self.gl.bind_framebuffer(
+                WebGl2RenderingContext::FRAMEBUFFER,
+                Some(&self.programs.resources.slot_framebuffers[ix]),
+            );
+            // Set viewport to match slot framebuffer.
+            // TODO: Remove the slot height texture calculation.
+            let total_slots: usize =
+                (get_max_texture_dimension_2d(self.gl) / u32::from(Tile::HEIGHT)) as usize;
```
taj-p, line 901: `get_max_texture_dimension_2d` is a blocking WebGL call. Are we able to use a stored value instead?

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+        );
+
+        // Bind framebuffer and setup viewport.
+        self.gl.bind_framebuffer(
+            WebGl2RenderingContext::FRAMEBUFFER,
+            Some(&self.programs.resources.slot_framebuffers[ix]),
+        );
+        // TODO: Remove the slot height texture calculation.
+        let total_slots: usize =
+            (get_max_texture_dimension_2d(self.gl) / u32::from(Tile::HEIGHT)) as usize;
```
taj-p, line 992: `get_max_texture_dimension_2d` is a blocking WebGL call. Are we able to use a stored value instead?

sparse_strips/vello_hybrid/src/render_webgl.rs
taj-p: Alternatively, can we store `total_slots`?

sparse_strips/vello_hybrid/src/render_webgl.rs
taj-p: Alternatively, can we store `total_slots`?

sparse_strips/vello_hybrid/examples/scenes/src/lib.rs
```diff
@@ -75,6 +76,7 @@ pub fn get_example_scenes() -> Box<[AnyScene]> {
         AnyScene::new(svg::SvgScene::tiger()),
         AnyScene::new(text::TextScene::new("Hello, Vello!")),
         AnyScene::new(simple::SimpleScene::new()),
+        AnyScene::new(clip::ClipScene::new()),
```
taj-p, line 79: Should we add this scene to the non-wasm32 `get_example_scenes` function?

taj-p (commented):
Initial comments on other parts of code. I ran the demos (native webgl, webgl, and winit). They look great 🥳

sparse_strips/vello_hybrid_shaders/src/types.rs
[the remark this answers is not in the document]
taj-p: Oh, in my POC, I realise I've used structs for Vertex and Fragment where they could be another nested module 😅

sparse_strips/vello_hybrid_shaders/build.rs
taj-p, line 1: Could we please also add a TODO to format the generated code? See [here](https://docs.rs/bindgen/0.51.1/src/bindgen/lib.rs.html#1968) for an example on how to do this

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -24,26 +24,37 @@
...
 extern crate alloc;
 
 mod render;
+#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+mod render_webgl;
+#[cfg(feature = "wgpu")]
+mod render_wgpu;
+
 mod scene;
+#[cfg(any(all(target_arch = "wasm32", feature = "webgl"), feature = "wgpu"))]
```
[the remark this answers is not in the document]
ajakubowicz-canva: Great pickup. This cfg is necessary (and was added) to pass CI. Specifically this check:
 - `cargo clippy --locked --no-default-features -- -D warnings`

This `cfg` basically says, include the `scheduler` if any of the renderer modules are being used. Without this the `clippy` invocation fails with:

```
error: trait `RendererBackend` is never used
   --> sparse_strips/vello_hybrid/src/schedule.rs:190:18
    |
190 | pub(crate) trait RendererBackend {
    |                  ^^^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`

...
error: struct `Scheduler` is never constructed
...
error: struct `Round` is never constructed
...
```

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -24,26 +24,37 @@
...
 mod schedule;
 pub mod util;
 
-pub use render::{Config, GpuStrip, RenderSize, RenderTargetConfig, Renderer};
+pub use render::{Config, GpuStrip, RenderSize};
+#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+pub use render_webgl::WebGlRenderer;
```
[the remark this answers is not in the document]
ajakubowicz-canva: I like this suggestion, and it's what I initially started with. Initially I treated the `wgpu` and `webgl` crate features as mutually exclusive. However, this was working against the grain and was a lot of trouble ([ref: mutually exclusive features](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features)).

Now features are additive as advised in [The Cargo book - Feature Unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification). This also helps with CI which uses [`cargo hack`](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--each-feature) to test compilation and clippy across all feature flag combinations.

> [NOTE!]
> I changed the `webgl` and `wgpu` features to be additive due to the Cargo book & existing CI that I didn't want to change. I'm now more familiar with the CI, and we could use `--mutually-exclusive-features` with `cargo hack` to inform it of the mutually exclusive features, preventing it from testing them together and thus working around the issue below. That is an option I'm happy to pursue.

Due to being additive, renaming this results in:

`cargo clippy --locked --target wasm32-unknown-unknown --all-features -- -D warnings` failing with:

```
error[E0252]: the name `Renderer` is defined multiple times
  --> sparse_strips/vello_hybrid/src/lib.rs:57:43
   |
55 | pub use render_webgl::Renderer;
   |         ---------------------- previous import of the type `Renderer` here
56 | #[cfg(feature = "wgpu")]
57 | pub use render_wgpu::{RenderTargetConfig, Renderer};
   |                                           ^^^^^^^^ `Renderer` reimported here
   |
   = note: `Renderer` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
57 | pub use render_wgpu::{RenderTargetConfig, Renderer as OtherRenderer};
```

Meta-question here is, should `webgl` and `wgpu` features remain _additive_, or be made _mutually exclusive_, or some other option?

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+        };
+        self.scheduler.do_scene(&mut ctx, scene)?;
+
+        // Blit the view framebuffer to the default framebuffer (canvas element), reflecting the
+        // image along the Y axis to complete the WebGPU to WebGL2 coordinate transform.
+        self.gl.bind_framebuffer(
+            WebGl2RenderingContext::READ_FRAMEBUFFER,
+            Some(&self.programs.resources.view_framebuffer),
+        );
+        {
```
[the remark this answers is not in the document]
ajakubowicz-canva: I can do this. I was being implicit and hoping that the `debug_assert_eq!` would be stripped in `--release` thus leaving `status` as dead-code, which finally would cause it to be removed.

That's a bit too implicit and I can wrap this explicitly in a `cfg`. Great call out.

sparse_strips/vello_hybrid/src/render_webgl.rs
ajakubowicz-canva: TIL!

So this should be something like: "Vertex Shader creation should succeed".
That message is so banal I think I'll change it to an `unwrap`. In the future it could be integrated as a returned error via `Result`.

sparse_strips/vello_hybrid/src/render_wgpu.rs
```diff
@@ -0,0 +1,782 @@
...
+                        min_binding_size: None,
+                    },
+                    count: None,
+                }],
+            });
+
+        let strip_shader = device.create_shader_module(wgpu::ShaderModuleDescriptor {
+            label: Some("Strip Shader"),
+            source: wgpu::ShaderSource::Wgsl(
+                include_str!("../../vello_hybrid_shaders/shaders/render_strips.wgsl").into(),
```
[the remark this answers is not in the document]
ajakubowicz-canva: I don't actually have a good opinion on this decision. My decision was based on the comment in the POC that build should only run as part of the optional dependency. Hence, I've not integrated any of the wgsl shaders into the package output. They are treated as data files that can be used by `vello_hybrid`.

I could be easily swayed to an alternative here. Currently this works.

sparse_strips/README.md
```diff
@@ -22,6 +22,7 @@ This directory contains the core crates for the Vello rendering. Each crate serv
 - **`vello_common`** – Provides shared data structures and utilities for rendering.
 - **`vello_cpu`** – Implements a CPU-based renderer optimized for multithreading and SIMD.
 - **`vello_hybrid`** – A hybrid CPU/GPU renderer, balancing workload between CPU and GPU.
+- **`vello_hybrid_shaders`** – Provide compilation of wgsl to glsl to support the WebGL `vello_hybrid` backend.
```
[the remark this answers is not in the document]
ajakubowicz-canva: Renamed to vello_sparse_shaders in https://github.com/linebender/vello/pull/1011/commits/71c9a45cbf432a9a8cd52f1d99899799978c5156#diff-c788699c721fbed5cdd64ad64925eba015ee5fc1958c0893f6b286b939854112

sparse_strips/vello_hybrid/examples/native_webgl/README.md
```diff
@@ -0,0 +1,21 @@
+## WebGL Demo
+
+Uses Vello Hybrid with a native WebGL2 backend in the browser. This example does not use wgpu.
+
+## Development
+
+Run with `cargo run_wasm -p native_webgl`.
```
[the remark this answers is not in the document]
ajakubowicz-canva: Added to both examples:
 - native_webgl: https://github.com/linebender/vello/pull/1011/commits/71c9a45cbf432a9a8cd52f1d99899799978c5156
 - wgpu_webgl: https://github.com/linebender/vello/pull/1011/commits/a3f6cf3534a8a9f6b6b1243b3a218e4bef80cb74#diff-f3e964af5013c23529a4ee1d56f92f8d9a791db4adfc9298e97bbeba1d37654dR7

sparse_strips/vello_hybrid/src/render.rs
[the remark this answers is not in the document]
ajakubowicz-canva: New structure: https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6#diff-d146599fda479559d167132f5441a375d66aa5c945f0e125c7f921a9781e6c4f

sparse_strips/vello_hybrid/src/render_webgl.rs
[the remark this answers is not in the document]
ajakubowicz-canva: Explicit debug_assertion cfg added: https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6

sparse_strips/vello_hybrid/src/render_webgl.rs
ajakubowicz-canva: I wasn't able to figure out if there's any cost, but there probably is some cost in creating the wrapped JavaScript object.  I simplified all usages of `buffer_data_with_array_buffer_view` to instead use `buffer_data_with_u8_array`.

Great find!

sparse_strips/vello_hybrid/src/render_webgl.rs
ajakubowicz-canva: I've audited all my `expect` usages. Most of them should never fail because they are WebGL calls, and the failure would happen within the WebGL server. Most have been changed to `unwrap`.

sparse_strips/vello_hybrid/src/render_webgl.rs
ajakubowicz-canva: Because we use `max_texture_dimension_2d` in a couple places, I've opted to store `max_texture_dimension_2d` explicitly (which was already implicitly stored as [`alpha_texture_width`](https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6).

Now [`max_texture_dimension_2d`](https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6) is cached, and then it's used as an `alpha_texture_width` where required. I've also removed all the blocking `get_max_texture_dimension_2d()` calls post-initialization.

sparse_strips/vello_hybrid/src/render_webgl.rs
```diff
@@ -0,0 +1,1034 @@
...
+                .check_framebuffer_status(WebGl2RenderingContext::READ_FRAMEBUFFER);
+            debug_assert_eq!(
+                status,
+                WebGl2RenderingContext::FRAMEBUFFER_COMPLETE,
+                "read framebuffer not complete"
+            );
+        }
+        self.gl
+            .bind_framebuffer(WebGl2RenderingContext::DRAW_FRAMEBUFFER, None);
+        {
```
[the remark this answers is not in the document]
ajakubowicz-canva: Done :D
https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6

sparse_strips/vello_hybrid/src/render_webgl.rs
ajakubowicz-canva: Yes done: https://github.com/linebender/vello/pull/1011#discussion_r2094731471

sparse_strips/vello_hybrid/examples/scenes/src/lib.rs
ajakubowicz-canva: Great suggestion: https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6

sparse_strips/vello_hybrid_shaders/build.rs
ajakubowicz-canva: Done: https://github.com/linebender/vello/pull/1011/commits/529067ad66fed7c8ed96427b8cad9f5c491225d6

sparse_strips/vello_hybrid_shaders/src/types.rs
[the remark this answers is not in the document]
ajakubowicz-canva: Thank you so much for this suggestion. I've added these fixes with https://github.com/linebender/vello/pull/1011/commits/994ffc782570e53d07e7a1bb41f8a48ef78e7f1f

Now the webgl render backend is compile time safe!

sparse_strips/vello_hybrid/examples/native_webgl/README.md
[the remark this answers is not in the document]
ajakubowicz-canva: Done in this commit: https://github.com/linebender/vello/pull/1011/commits/ef903a9044cd5938952a526fc47278338689ea05#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R23

ajakubowicz-canva:
Note, the `cargo doc` CI issue is related to a recent nightly toolchain release. I have a mitigation PR that is separate: https://github.com/linebender/vello/pull/1014

.github/workflows/ci.yml
```diff
@@ -331,7 +331,7 @@ jobs:
 
       - name: check vello_hybrid is webgl compatible
```
taj-p, line 332: Did you want to add a step for `sparse_strips/vello_hybrid/examples/native_webgl` or is that accounted for in the next PR?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -0,0 +1,996 @@
...
+    gl.bind_vertex_array(None);
+}
+
+/// Context for WebGL rendering operations.
+struct WebGlRendererContext<'a> {
+    programs: &'a mut WebGlPrograms,
+    gl: &'a WebGl2RenderingContext,
+}
+
+impl WebGlRendererContext<'_> {
```
taj-p, line 825: Architecturally, this differs significantly from the WGPU implementation in how buffers are managed. In the WGPU variant, we allocate a new buffer per pass. This uses more memory but allows the GPU to not be blocked. For example, if multiple render passes are requested by this variant, then the GPU will need to finish drawing before being able to execute `upload_strips` (effectively pausing execution) because there is only 1 buffer in play, if I'm reading this correctly. 

I think we could use a buffer pool with some state management or simply create a new buffer per pass. I don't think we need to handle this now but could we please leave a TODO for this issue?

taj-p (approved):
LGTM

sparse_strips/vello_hybrid/src/render/webgl.rs
ajakubowicz-canva: You are reading it correctly, I can definitely leave a TODO and will also do some investigating and thinking around the buffer pool proposal.

.github/workflows/ci.yml
ajakubowicz-canva: Great catch – will fix.

sparse_strips/vello_hybrid/src/render/webgl.rs
ajakubowicz-canva: TODO added: https://github.com/linebender/vello/pull/1011/files#diff-d3d335ee2962ebe7ad03595115c4bf5d6458d30f1c03c5dd90ab95f11ce8a597R822-R824

.github/workflows/ci.yml
ajakubowicz-canva: Done: https://github.com/linebender/vello/pull/1011/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL332-R338

ajakubowicz-canva:
All clear to merge from Raph: [#vello > vello_hybrid: WebGL2 renderer backend that skips wgpu #947 @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/vello_hybrid.3A.20WebGL2.20renderer.20backend.20that.20skips.20wgpu.20.23947/near/519293074)

sparse_strips/vello_hybrid/src/render_wgpu.rs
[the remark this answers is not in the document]
DJMcNab: I don't think this can work with a crates.io release, if this hasn't been changed.

In Vello Shaders, we exported it as a Rust string, which worked fine.

Crimitorii:
Gratz! Big job done.

PR #1012, opened 2025-05-16
ci: update to stable Rust 1.87, typos 1.32.0

waywardmonkeys:


LaurenzV (approved):

ajakubowicz-canva:
This PR will be unblocked by https://github.com/linebender/vello/pull/1021

ajakubowicz-canva:
I redact my prior comment. This PR bumps the Rust version, which seems to change the size of the `Result` calculation such that `clippy` is triggered.

This PR will need to include the change in: https://github.com/linebender/vello/pull/1021

vello_shaders/src/compile/mod.rs
```diff
@@ -85,6 +85,13 @@ pub struct ShaderInfo {
 }
 
 impl ShaderInfo {
+    #[cfg_attr(
+        not(target_arch = "wasm32"),
```
waywardmonkeys, line 89: I think you want to check against `target_pointer_width` here.

ajakubowicz-canva: TIL

xStrom (approved):

PR #1013, opened 2025-05-16
Alternative name encoding

taj-p:
Encodes the shader like:

```rs
pub mod shaders {
    #![allow(unused_mut, reason = "Increases code gen complexity")]
    /// Compiled glsl for `clear_slots.wgsl`
    pub mod ClearSlots {
        pub struct Vertex {}
        impl Vertex {
            pub const CONFIG_UNIFORM: &'static str = "Config_block_0Vertex";
        }

        pub struct Fragment {}
        impl Fragment {}

        pub const VERTEX_SOURCE: &'static str = r###"#version 300 es
....
"###;
        pub const FRAGMENT_SOURCE: &'static str = r###"#version 300 es
...
"###;
    }


ajakubowicz-canva:
This is really good stuff. Thank you for the reference :D

ajakubowicz-canva:
I've rolled this change into #1011

PR #1016, opened 2025-05-20
export WGSL shader source code directly from `vello_sparse_shaders`

ajakubowicz-canva:
# Context

Addresses comment https://github.com/linebender/vello/pull/1011#discussion_r2097657320 from @DJMcNab . Replaces `include_str!` with a direct import from `vello_sparse_shaders` to get access to the wgsl source code.

### Changes

 - Adds a `wgsl` module to the build-time module generated by `vello_sparse_shaders`.

This allows `render_strips.wgsl` to be accessed via: `vello_sparse_shaders::wgsl::RENDER_STRIPS` instead of a cross-package `include_str!`.

 - Adds a `glsl` feature flag to `vello_sparse_shaders` so the default compiled module does not include `glsl` unless the `glsl` feature is used. This makes builds much leaner for wgsl without glsl.

### Test plan

Reviewed generated code, and tested both examples manually:
 - `cargo run_wasm -p wgpu_webgl --release --port 8001`
 - `cargo run_wasm -p native_webgl --release --port 8000`

Finally, CI is very thorough.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -227,16 +227,12 @@ impl Programs {
 
         let strip_shader = device.create_shader_module(wgpu::ShaderModuleDescriptor {
             label: Some("Strip Shader"),
-            source: wgpu::ShaderSource::Wgsl(
-                include_str!("../../../vello_sparse_shaders/shaders/render_strips.wgsl").into(),
-            ),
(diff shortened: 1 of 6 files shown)


sparse_strips/vello_hybrid/src/render/wgpu.rs
ajakubowicz-canva, line 1: Most important change of this PR is replacing these `include_str!` calls.

PR #1017, opened 2025-05-20
Replace uses of `eprintln!` with `log`

### LINKED ISSUE #1010: Use configurable logging instead of `eprintln!`
Currently, vello does a lot of logging using `eprintln!`

This is unfortunate:
- GUI programs using vello can't easily show these messages in a developer console or log viewer UI without intercepting stderr.
- Headless programs (example use case: SVG to PNG converter) can't hide messages the user doesn't care about, like `Initialising in parallel using 14 threads`.

It would be nicer if vello provided some way to configure where to log messages to. Using an existing logging crate like `log` or `tracing` would be nice. But if there is some reason to not add those dependencies, providing a way set which `Write` instance it should log to would also be sufficient.

DJMcNab:
Fixes #1010

### DIFF
--- a/vello_shaders/build.rs
+++ b/vello_shaders/build.rs
@@ -13,11 +13,14 @@ mod types;
 
 use std::env;
 use std::fmt::Write;
+use std::io::{Write as _, stdout};
 use std::path::Path;
 
 use compile::ShaderInfo;
+use log::Level;
 
 fn main() {
+    log::set_logger(&BUILD_SCRIPT_LOGGER).unwrap();
     let out_dir = env::var_os("OUT_DIR").unwrap();
     let dest_path = Path::new(&out_dir).join("shaders.rs");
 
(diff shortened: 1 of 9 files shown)


vello_shaders/build.rs
```diff
@@ -136,3 +139,36 @@ fn write_msl(buf: &mut String, info: &ShaderInfo) -> Result<(), std::fmt::Error>
     writeln!(buf, "            }},")?;
     Ok(())
 }
+
+struct BuildScriptLog;
```
waywardmonkeys, line 143: It is semi obvious after reading more code, but having a doc comment here would be nice.

vello_shaders/build.rs
```diff
@@ -136,3 +139,36 @@ fn write_msl(buf: &mut String, info: &ShaderInfo) -> Result<(), std::fmt::Error>
...
+    fn enabled(&self, _: &log::Metadata<'_>) -> bool {
+        true
+    }
+
+    fn log(&self, record: &log::Record<'_>) {
+        let mut target = String::new();
+        if record.level() >= Level::Warn {
+            write!(LineAdapter(&mut target), "{}", record.args()).unwrap();
+        }
+        println!("{target}");
```
tomcur, line 157: Should this be inside the branch?

nicoburns (approved):

PR #1018, opened 2025-05-21
Update artifact uploading

DJMcNab:
This should have been changes in #892.

I'm not overly happy with the difference between `vello_tests/current` and `vello_sparse_tests/diffs`, but that isn't for this PR to solve.


LaurenzV (approved):
Can't judge whether it's right for `vello_test`, but looks right for sparse strips at least.

xStrom (approved):

PR #1021, opened 2025-05-23
Ignore new `result_large_error` clippy lint

ajakubowicz-canva:
### Context

New PRs being opened against Vello are failing CI due to the new [clippy lint `result_large_err`](https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err).

E.g. Blocked PRs:
 - https://github.com/linebender/vello/pull/1020
 - https://github.com/linebender/vello/pull/1012

### Fix

Temporarily silence this clippy lint explicitly. Happy to change the reason to something else or a TODO if that would be preferred and more explicit.

### Test plan

Passing CI.

### Risk

I don't think this should have any impact on end users as this is a clippy attribute addition.

### DIFF
--- a/vello_shaders/src/compile/mod.rs
+++ b/vello_shaders/src/compile/mod.rs
@@ -85,7 +85,7 @@ pub struct ShaderInfo {
 }
 
 impl ShaderInfo {
-    #[allow(clippy::result_large_err, reason = "Large error expected.")]
+    #[expect(clippy::result_large_err, reason = "Deferred: This is a cold code path.")]
     pub fn new(name: &str, source: String, entry_point: &str) -> Result<Self> {
         let module = wgsl::parse_str(&source).map_err(|error| Error::new(&source, name, error))?;
         let module_info = naga::valid::Validator::new(ValidationFlags::all(), Capabilities::all())
--- a/vello_shaders/src/compile/mod.rs
+++ b/vello_shaders/src/compile/mod.rs
@@ -85,7 +85,10 @@ pub struct ShaderInfo {
 }
 
 impl ShaderInfo {
-    #[expect(clippy::result_large_err, reason = "Deferred: This is a cold code path.")]
+    #[expect(
+        clippy::result_large_err,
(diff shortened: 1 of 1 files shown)


ajakubowicz-canva:
I've rebased off this branch, but I can rebase off main if you want to see the repro.

The other "live" example is: https://github.com/linebender/vello/actions/runs/15064714129/job/42346870528?pr=1012 in #1012

vello_shaders/src/compile/mod.rs
```diff
@@ -85,6 +85,7 @@ pub struct ShaderInfo {
 }
 
 impl ShaderInfo {
+    #[allow(clippy::result_large_err, reason = "Large error expected.")]
```
DJMcNab, line 88:
```suggestion
    #[expect(clippy::result_large_err, reason = "Deferred: This is a cold code path.")]
```

The old `reason` just restated the lint name, whereas the reason should explain why we're ignoring the error here. I believe that our MSRV is high enough to use `expect`.

ajakubowicz-canva: Thank you so much! Lets see if it passes CI 😄

DJMcNab:
Oh, so the size calculation for result_large_error is different in 1.86 and 1.87? That's the only reasonable explanation I can see for this happening.

ajakubowicz-canva:
Yeah – I think you're right. I've just been checking against CI. I have:

```
> rustc --version
rustc 1.87.0 (17067e9ac 2025-05-09)
```

And I think CI is using `1.86.0`.

DJMcNab:
https://github.com/linebender/vello/actions/runs/15162566157/job/42787918862 is a run on main which I've just re-ran - it isn't running into this

DJMcNab:
I think ultimately this will need to be merged into #1012.

ajakubowicz-canva:
yeah. Agree. Closing this.

I think I got confused that this would happen on the webgl test PR because I was running `cargo hack` locally to predict CI, but I'm on the next version.

PR #1023, opened 2025-05-26
Improve encoding of gradients

LaurenzV:
This PR makes some changes to the way gradients are encoded. The motivation for this is two-fold:
- Improve the overall performance of gradient rendering.
- Make gradient rendering more SIMD-friendly.

Overall, the PR contains four changes in total:
- The encoding has been adapted so that the parametric variable `t` for each gradient is always between 0.0 and 1.0. This will make it easier to implement an efficient version of the `extend` method. Right now, I still left the while loop, but this can now be replaced with `(val - val.floor()).fract()`, which is much more SIMD-friendly.
- The encoding of linear gradients has been simplified, such that the calculation of the current position along the gradient line is basically a no-op, giving much better performance.
- Encoding of radial gradients has been completely rewritten. Previously, we would use a quadratic solver to find the parametric value `t` for a certain position, which had some issues and also was not really SIMDifiable. Now, I have instead rewritten it to use the same approach that is used by Skia, which is faster and will be much faster once we have SIMD, since the approach is very SIMD-friendly.
- It fixes a bug where the offset for sampling at the pixel center was applied wrongly.

I _think_ this is pretty much as good as encoding will get, so while I don't want to make promises, I expect this to be the final form of gradient encoding and don't think it will change significantly in the future, meaning that it should also be safe to start implementing this for the GPU (do note that this only applies to the encoding itself and not for example the `cur_pos` function, that one will probably change because the current approach with using `Option` for radial gradients won't work for SIMD.


LaurenzV:
Unfortauntely, there are a few test cases where there are slight differences in the pixel values in CI. Just increasing the pixel threshold does not solve this problem, because the problem is that the difference results in a color being sampled from a different gradient range at the boundary. To circumvent this, I have now added an attribute that allows to specify the number of pixels that are allowed to deviate completely from the reference image.

DJMcNab:
To clarify, these are failures where the hybrid renderer doesn't match up with the CPU only renderer?

LaurenzV:
No, as the hybrid renderer doesn’t support gradients yet. It’s where the CPU renderer has slightly different results depending on which OS CI runs on.

sparse_strips/vello_sparse_tests/snapshots/glyphs_colr_test_glyphs.png
nicoburns, line 1: The difference in the sweep gradients look surprisingly big to me. Can someone who know what they are talking about check that this is an expected level of difference? ("swipe" view is easiest to see the difference)

sparse_strips/vello_sparse_tests/snapshots/gradient_sweep_with_transform_rotate_2.png
dfrg, line 1: Is this one hitting some edge case? The other rotated sweeps look clean.

sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -29,6 +29,11 @@ struct Arguments {
     skip_cpu: bool,
     /// Whether the test should not be run on the GPU (`vello_hybrid`).
     skip_hybrid: bool,
+    /// The maximum number of pixels that are allowed to completely deviate from the reference
+    /// images. This attribute mainly exists because there are some test cases (like gradients),
+    /// where, due to floating point inaccuracies, some pixels might land on a different color
+    /// stop and thus yield a different value in CI.
+    diff_pixels: u16,
```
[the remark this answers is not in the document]
LaurenzV: I don't think so, I don't think this is an issue in the `GradientFiller::advance` implementation (after all we are already using f64 for that), but simply a result of the fact that floating point computations don't seem to be 100% consistent on different systems.

For the `gradient_with_color_spaces_2`, `gradient_sweep_with_transform_rotate_1` and `gradient_sweep_with_transform_rotate_2` test cases, i could probably change them slightly so that they don't hit this edge case.

For the `glyphs_colr_test_glyphs` test case, I don't think there is much I can do about it because the sweep gradients are defined as part of the font file.

dominikh:
At the end of the day, if you want the gradients to look good in any scenario that isn't a perfectly vertical or horizontal linear gradient, you'll need some form of anti aliasing or supersampling, anyway. I'm not sure how much the artifacts matter in light of that.

LaurenzV:
So I've gone ahead and adjusted some test cases so they don't hit that edge case anymore, but I don't think we'll be able to get rid of the CI issues for the COLR test glyphs, unfortunately.

grebmeg (approved):
Alright, I think I’m done with this PR! 😊 Really nice work, @LaurenzV  — and thanks for walking me through some parts that weren’t immediately clear. I have a feeling either I or someone else will revisit this later when we start implementing gradients for `vello_hybrid`.

PR #1029, opened 2025-05-29
Align image rasterization more with Skia implementation

LaurenzV:
On the one hand, to make it more SIMD-compatible, but on the other hand for some nice performance improvements already!

```
Before:
fine/image/transform/none_u8: 893.79 ns
fine/image/transform/none_f32: 901.79 ns
fine/image/extend/pad_u8: 902.01 ns
fine/image/extend/pad_f32: 947.86 ns
fine/image/extend/repeat_u8: 2.2883 µs
fine/image/extend/repeat_f32: 2.3449 µs
fine/image/extend/reflect_u8: 2.3390 µs
fine/image/extend/reflect_f32: 2.4345 µs

After:
fine/image/transform/none_u8_scalar: 438.70 ns
fine/image/transform/none_f32_scalar: 626.66 ns
fine/image/extend/pad_u8_scalar: 464.97 ns
fine/image/extend/pad_f32_scalar: 801.11 ns
fine/image/extend/repeat_u8_scalar: 500.76 ns
fine/image/extend/repeat_f32_scalar: 682.00 ns
fine/image/extend/reflect_u8_scalar: 876.11 ns
fine/image/extend/reflect_f32_scalar: 1.1661 µs
```


sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -11,39 +11,44 @@ use vello_common::tile::Tile;
...
+    width_inv: f32,
 }
 
 impl<'a> ImageFiller<'a> {
     pub(crate) fn new(image: &'a EncodedImage, start_x: u16, start_y: u16) -> Self {
+        let width = image.pixmap.width() as f32;
+        let height = image.pixmap.height() as f32;
+
         Self {
             // We want to sample values of the pixels at the center, so add an offset of 0.5.
```
ajakubowicz-canva, line 45: Nit: I think this comment can be removed, or, maybe it's helpful if this comment is added on the `new` method as documentation that `start_x` and `start_y` should be centered on the pixel?

Or, is `ImageFiller` no longer sampling the center of pixels?

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -155,12 +161,12 @@ impl<'a> ImageFiller<'a> {
                 // Nearest neighbor filtering.
                 // Simply takes the nearest pixel to our current position.
                 ImageQuality::Low => {
-                    let point = extend_point(pos);
+                    let p = extend_point(pos);
```
ajakubowicz-canva, line 164: Nit: I sort of prefer non single letter names. But optional.

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -135,14 +139,16 @@ impl<'a> ImageFiller<'a> {
         let extend_point = |mut point: Point| {
             // For the same reason as mentioned above, we always floor.
```
ajakubowicz-canva, line 140: Nit: This comment seems no longer relevant.

sparse_strips/vello_cpu/src/fine/image.rs
ajakubowicz-canva: So to clarify, `start_x` and `start_y` are no longer sampling the center?

LaurenzV: We are sampling the center due to this line:

image

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -278,21 +263,26 @@ impl<'a> ImageFiller<'a> {
...
+            let bias_in_ulps = s.trunc();
 
-            if val_mod < max {
-                val_mod
-            } else {
-                (period - 1.0) - val_mod
-            }
+            let m_bits = m.to_bits();
+            let biased_bits = m_bits.wrapping_sub(bias_in_ulps as u32);
+            f32::from_bits(biased_bits)
```
ajakubowicz-canva, line 285: This is so damn clever. If I am understanding, by subtracting by 1 bit in a u32 representation of `f32`, then you are basically shifting the number by the smallest possible amount.

I can see this is based on Skia, so I have high confidence this is correct. My main question is could there be any edge cases? For example what if `m` is `0f32` ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=66d5f9329e793e6f8410b14f200effee))?

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -278,21 +263,26 @@ impl<'a> ImageFiller<'a> {
     }
 }
 
-fn extend(val: f32, extend: Extend, max: f32) -> f32 {
+#[inline(always)]
+fn extend(val: f32, extend: Extend, max: f32, inv_max: f32) -> f32 {
+    const BIAS: f32 = 0.01;
```
ajakubowicz-canva, line 268: How was this value chosen? Why not `f32::EPSILON`?

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -11,39 +11,43 @@ use vello_common::tile::Tile;
...
 pub(crate) struct ImageFiller<'a> {
     /// The current position that should be processed.
     cur_pos: Point,
     /// The underlying image.
     image: &'a EncodedImage,
+    // We precompute these values so we don't need to recalculate them for every pixel.
+    height: f32,
+    height_inv: f32,
+    width: f32,
+    width_inv: f32,
```
ajakubowicz-canva, line 36: nit that can be ignored: This third comment explains why and is in first person plural. Could be more consistent. Maybe something like:

`Precomputed values reused in per-pixel calculations.`

ajakubowicz-canva (approved):
I learnt a lot reviewing this. Thank you for excellent references to Skia.

I'll leave to your best judgement if someone else should review.

sparse_strips/vello_cpu/src/fine/image.rs
LaurenzV: That doesn't work, because for example `30.0f32 - f32::EPSILON` is still 30.0, so for larger values the bias wouldn't have any effect. This value was admittedly chosen somewhat arbitrarily, but it should be enough for all values that we support (at least for u16::MAX, I tried that). Will add a comment explaining this, though.

sparse_strips/vello_cpu/src/fine/image.rs
LaurenzV: If `m` is 0.0, then `s` and `u` must be an integer number (since our `max` is always an integer as well), and thus `m_bits` is also 0, and no subtraction happens. Will add a comment for that.

Issue #1034, opened 2025-06-01
Another panic in clipping


LaurenzV:
With the following test case:

```
#[vello_test(width = 612, height = 792)]
fn nested_clip_path_panic_2(ctx: &mut impl Renderer) {
    let path1 = BezPath::from_svg("M268.2662811279297,458.7216339111328 C268.2662811279297,455.4202194213867 265.5900192260742,452.74398040771484 262.2886199951172,452.74398040771484 C258.9872055053711,452.74398040771484 256.3109664916992,455.4202194213867 256.3109664916992,458.7216339111328 C256.3109664916992,462.02303314208984 258.9872055053711,464.6992950439453 262.2886199951172,464.6992950439453 C265.5900192260742,464.6992950439453 268.2662811279297,462.02303314208984 268.2662811279297,458.7216339111328 Z M262.2886199951172,458.7216339111328").unwrap();
    ctx.push_clip_layer(&path1);
    let path2 = BezPath::from_svg("M250.33160400390625,470.6786193847656 L274.2456043958664,470.6786193847656 L274.2456043958664,446.7646189928055 L250.33160400390625,446.7646189928055 Z").unwrap();
    ctx.push_clip_layer(&path2);
    ctx.fill_path(&BezPath::from_svg("M250.33160400390625,470.6786193847656 L274.2456043958664,470.6786193847656 L274.2456043958664,446.7646189928055 L250.33160400390625,446.7646189928055 Z").unwrap());
    ctx.pop_layer();
    ctx.pop_layer();
}
```

LaurenzV:
Reduced to: 

```rs
#[vello_test(width = 612, height = 792)]
fn nested_clip_path_panic_3(ctx: &mut impl Renderer) {
    let path1 = Rect::new(256.0, 452.0, 268.0, 464.0).to_path(0.1);
    ctx.push_clip_layer(&path1);
    let path2 = &Rect::new(0.0, 0.0, 612.0, 792.0).to_path(0.1);
    ctx.push_clip_layer(&path2);
    ctx.fill_path(&Rect::new(0.0, 0.0, 612.0, 792.0).to_path(0.1));
    ctx.pop_layer();
    ctx.pop_layer();
}
```

PR #1035, opened 2025-06-01
Fix panic in nested clip paths

LaurenzV:
This is an attempt at fixing #1034, and is based on #1033, which should be merged first. I am not 100% confident that this is the correct solution, so a careful review would be appreciated.

The problem is as follows: The first clip path is a very small rectangle, which means that the clip bbox will only cover the wile (1, 0). The second clip path is much bigger and starts at the wide tile (0, 0).

Because of this, once we pop the second clip path, the assignment
```rs
let mut x1 = x0 + strip_width;
```

will end up yielding 4 (since the second, bigger clip path starts at x = 0, and it's strip will be 4 wide). However, since the second clip path starts outside of the clip bbox,

```rs
let clip_x = clip_bbox.x0() * WideTile::WIDTH;
if clip_x > x {
    col += u32::from(clip_x - x);
    x1 = clip_x.max(x1);
}
```

The `x` value will be adjusted to 256 (since the clip bbox starts at the first wide tile in the x direction), but note that our x1 is still 4! Because of this

```rs
 let x2 = next_strip.x;
let clipped_x2 = x2.min((cur_wtile_x + 1) * WideTile::WIDTH);
let width = clipped_x2.saturating_sub(x1);
```

the width calculated here will be much bigger than 256, and thus lead to a panic.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -615,7 +615,7 @@ impl Wide {
             let next_strip = &strips[i + 1];
             let strip_width =
                 ((next_strip.alpha_idx - strip.alpha_idx) / u32::from(Tile::HEIGHT)) as u16;
-            let x1 = x0 + strip_width;
+            let mut x1 = x0 + strip_width;
             let wtile_x0 = (x0 / WideTile::WIDTH).max(clip_bbox.x0());
             let wtile_x1 = x1.div_ceil(WideTile::WIDTH).min(clip_bbox.x1());
 
@@ -626,6 +626,7 @@ impl Wide {
             if clip_x > x {
                 col += u32::from(clip_x - x);
                 x = clip_x;
+                x1 = clip_x.max(x1);
             }
 
             // Render clip strips for each affected tile and mark for popping
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -615,9 +615,9 @@ impl Wide {
             let next_strip = &strips[i + 1];
             let strip_width =
                 ((next_strip.alpha_idx - strip.alpha_idx) / u32::from(Tile::HEIGHT)) as u16;
-            let mut x1 = x0 + strip_width;
+            let mut clipped_x1 = x0 + strip_width;
             let wtile_x0 = (x0 / WideTile::WIDTH).max(clip_bbox.x0());
-            let wtile_x1 = x1.div_ceil(WideTile::WIDTH).min(clip_bbox.x1());
+            let wtile_x1 = clipped_x1.div_ceil(WideTile::WIDTH).min(clip_bbox.x1());
 
             // Calculate starting position and column for alpha mask
             let mut x = x0;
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -250,3 +251,27 @@ fn out_of_viewport_clip(ctx: &mut impl Renderer) {
...
+fn nested_clip_path_panic(ctx: &mut impl Renderer) {
+    let path1 = Rect::new(256.0, 0.0, 257.0, 2.0).to_path(0.1);
+    ctx.push_clip_layer(&path1);
+    let path2 = Rect::new(181.0, -200.0, 760.0, 618.0).to_path(0.1);
+    ctx.push_clip_layer(&path2);
+    ctx.pop_layer();
+    ctx.pop_layer();
+}
+
+#[vello_test(width = 612, height = 4)]
```
grebmeg: Should it be this?

```suggestion
#[vello_test(width = 512, height = 4)]
```

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -626,6 +626,7 @@ impl Wide {
             if clip_x > x {
                 col += u32::from(clip_x - x);
                 x = clip_x;
+                x1 = clip_x.max(x1);
```
grebmeg: Yeah, I think it makes sense to adjust the current right-edge coordinate given the case you ran into. It might also be worth explicitly naming this variable to highlight that it can be adjusted to fit the clip area — something like `clipped_x1`. I’ll leave that decision up to you.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -615,7 +615,7 @@ impl Wide {
             let next_strip = &strips[i + 1];
```
grebmeg, line 615: I noticed that one of the strips has an x-coordinate of 508, while the rect’s x1 is 511. I thought the strip width was always 4 for rects — is that not the case, or am I missing something?

grebmeg (approved):

sparse_strips/vello_sparse_tests/tests/issues.rs
LaurenzV: Just tried it and yeah, this triggers the issue as well. Will use that.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Yeah, exactly. The rectangle starts at 511, but since strips always start at a offset that is a multiple of 4, the strip itself will start at 508 instead, and since it is 4 wide it will also cover the x coordinate 511.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Changed the naming.

sparse_strips/vello_common/src/coarse.rs
grebmeg: Ah, yeah, it includes the last coordinate — I missed that piece! Thanks for clarifying! 😊

PR #1038, opened 2025-06-02
Return GL from WebGLRenderer

taj-p:
I don't know how different browsers and their versions handle creating WebGL2 contexts with differing context options (e.g. enabling/disabling `antialias`).

The specification states that the [first call wins](https://arc.net/l/quote/otbncojz): " The WebGLContextAttributes object is only used on the first call to getContext. No facility is provided to change the attributes of the drawing buffer after its creation.".

For this reason, I think we should return the GL context created for the renderer for use by external consumers.

### DIFF
--- a/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
@@ -30,7 +30,7 @@ struct RendererWrapper {
 impl RendererWrapper {
     #[cfg(target_arch = "wasm32")]
     fn new(canvas: web_sys::HtmlCanvasElement) -> Self {
-        let renderer = vello_hybrid::WebGlRenderer::new(&canvas);
+        let (renderer, _) = vello_hybrid::WebGlRenderer::new(&canvas);
 
         Self { renderer }
     }
@@ -380,7 +380,7 @@ pub async fn render_scene(scene: vello_hybrid::Scene, width: u16, height: u16) {
         .append_child(&canvas)
         .unwrap();
 
-    let mut renderer = vello_hybrid::WebGlRenderer::new(&canvas);
+    let (mut renderer, _) = vello_hybrid::WebGlRenderer::new(&canvas);
 
     let render_size = vello_hybrid::RenderSize {
         width: width as u32,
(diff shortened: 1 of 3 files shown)


ajakubowicz-canva (commented):
I am hesitant to expand the API surface in a way that's incompatible with the `wgpu` backend – as I imaging a future where we delete the `WebGlRenderer` backend when 99.9% of users support `WebGPU`, and then we provide the 3mb wgpu WebGL polyfill to support the `0.1%` of tail users. We do not want to make a more challenging migration.

`wgpu` does not expose the `WebGL2Context` as a public API. `wgpu` does expose a OpenGL context for native ([see windows docs here](https://docs.rs/wgpu-hal/latest/x86_64-pc-windows-msvc/wgpu_hal/gles/struct.Adapter.html#method.new_external)).

But, I also don't think this is necessary. Per the [web spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext), it is spec compliant for subsequent calls to `getContext` to return the first created context. See the table, where a `webgl2` context is requested when the canvas context is already `webgl` or `webgl2`:

> Return the same object as was returned the last time the method was invoked with this same first argument.

If a user really needs access to the context, they can retrieve the exact same context via `getContext`.

taj-p:
> I am hesitant to expand the API surface in a way that's incompatible with the `wgpu` backend – as I imaging a future where we delete the `WebGlRenderer` backend when 99.9% of users support `WebGPU`, and then we provide the 3mb wgpu WebGL polyfill to support the `0.1%` of tail users. We do not want to make a more challenging migration.
> 
> `wgpu` does not expose the `WebGL2Context` as a public API. `wgpu` does expose a OpenGL context for native ([see windows docs here](https://docs.rs/wgpu-hal/latest/x86_64-pc-windows-msvc/wgpu_hal/gles/struct.Adapter.html#method.new_external)).
> 
> But, I also don't think this is necessary. Per the [web spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext), it is spec compliant for subsequent calls to `getContext` to return the first created context. See the table, where a `webgl2` context is requested when the canvas context is already `webgl` or `webgl2`:
> 
> > Return the same object as was returned the last time the method was invoked with this same first argument.
> 
> If a user needs access to the context, they can retrieve the exact same context via `getContext` on the canvas element.

Thank you for the great explanation 🙏 . Agree!

PR #1039, opened 2025-06-02
vello_hybrid: use canvas extents for webgl coordinate y-axis flip

ajakubowicz-canva:
### Context

When integrating the WebGL backend internally, @taj-p found a bug with the y-axis coordinate flip.

We were _only_ flipping the height of the scene, when we should be flipping the full canvas height ([as per wgpu's implementation](https://github.com/gfx-rs/wgpu/blob/f139e223e218f3bbe74e84645c3d47d209f7b755/wgpu-hal/src/gles/web.rs#L266-L277)).

### Fix

Use the `drawingBufferWidth` and `drawingBufferHeight` per the first point on https://webgl2fundamentals.org/webgl/lessons/webgl-anti-patterns.html .

### Test plan

No test plan currently – I'm conflicted about how to add a test for this as the `vello_test` macro does not provide a way to decouple surface dimensions and scene dimensions.

Should I add a bespoke hand-rolled test, or extend the `vello_test` macro?

### DIFF
--- a/sparse_strips/vello_dev_macros/src/test.rs
+++ b/sparse_strips/vello_dev_macros/src/test.rs
@@ -225,7 +225,7 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
             let mut ctx = get_ctx::<Scene>(#width, #height, #transparent);
             #input_fn_name(&mut ctx);
             if !#no_ref {
-                check_ref(&ctx, #input_fn_name_str, #webgl_fn_name_str, #hybrid_tolerance, false, RenderMode::OptimizeSpeed, #reference_image_name);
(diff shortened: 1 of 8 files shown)


ajakubowicz-canva:
> Should I add a bespoke hand-rolled test, or extend the vello_test macro?

cc/ @taj-p

taj-p (commented):
> Should I add a bespoke hand-rolled test, or extend the vello_test macro?

@ajakubowicz-canva , I suspect whatever's easiest 😅 . I'd imagine passing a `surface_[width|height]` attribute through `vello_test` and into `get_ctx` might be more maintainable.

ajakubowicz-canva:
I've updated to draft as I don't think I am understanding the bug properly. I cannot write a red test that is fixed by updating the `webgl` flip logic. Both tests added in this draft PR pass with either logic.

ajakubowicz-canva:
The confusion was that if `RenderSize` !== the logical surface dimensions, rendering gets weird and there is no error.

PR #1041, opened 2025-06-03
Use `min` and `max` instead of `clamp`

LaurenzV:
Has a very nice impact on performance:

```
fine/image/transform/none_u8_scalar
                        time:   [420.06 ns 421.11 ns 422.19 ns]
                        change: [-6.1518% -5.8343% -5.5324%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  1 (1.00%) high severe

fine/image/transform/none_f32_scalar
                        time:   [550.21 ns 551.17 ns 552.26 ns]
                        change: [-11.889% -11.423% -10.971%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
```

I think the reason is that `clamp` as additional handling for `NaN`, but we don't need this.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/image.rs
@@ -272,7 +272,7 @@ fn extend(val: f32, extend: Extend, max: f32, inv_max: f32) -> f32 {
     match extend {
         // Note that max should be exclusive, so subtract a small bias to enforce that.
         // Otherwise, we might sample out-of-bounds pixels.
-        Extend::Pad => val.clamp(0.0, max - BIAS),
+        Extend::Pad => val.min(max - BIAS).max(0.0),
         Extend::Repeat => val - floor(val * inv_max) * max,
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -272,7 +272,7 @@ fn extend(val: f32, extend: Extend, max: f32, inv_max: f32) -> f32 {
     match extend {
         // Note that max should be exclusive, so subtract a small bias to enforce that.
         // Otherwise, we might sample out-of-bounds pixels.
-        Extend::Pad => val.clamp(0.0, max - BIAS),
+        Extend::Pad => val.min(max - BIAS).max(0.0),
```
waywardmonkeys, line 277: Might be worth commenting here about this choice to avoid someone changing it in the future, since it is more common to see a clamp than a min/max.

taj-p (approved):

PR #1042, opened 2025-06-03
Lift image extend calculation out of the hot loop for the simple case

LaurenzV:
Based on top of #1041. This PR lifts the check for the image extend out of the loop, which (as shown by the benchmarks) seems to take up a significant chunk of time for the `Pad` case. Since the code is relatively short, I don't think there are any concers about having this method be generic. 

For bilinear/bicubic filtering, this is not worth having because it would bloat the code size and will not be the bottleneck.

Those are the benchmark results I'm getting for the `Pad` case now (compared to #1041):
```
fine/image/transform/none_u8_scalar
                        time:   [326.32 ns 326.70 ns 327.11 ns]
                        change: [-25.214% -25.074% -24.931%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  2 (2.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe

fine/image/transform/scale_u8_scalar
                        time:   [343.22 ns 343.56 ns 343.91 ns]
                        change: [-24.942% -24.780% -24.619%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 22 outliers among 100 measurements (22.00%)
  4 (4.00%) low severe
  4 (4.00%) low mild
  3 (3.00%) high mild
  11 (11.00%) high severe

fine/image/transform/rotate_u8_scalar
                        time:   [2.4611 µs 2.4640 µs 2.4673 µs]
                        change: [-0.4451% -0.1810% +0.0581%] (p = 0.16 > 0.05)
Found 13 outliers among 100 measurements (13.00%)
  4 (4.00%) low severe
  3 (3.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe

fine/image/extend/pad_u8_scalar
                        time:   [347.47 ns 348.00 ns 348.58 ns]
                        change: [-27.252% -26.488% -25.857%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  6 (6.00%) high mild
  2 (2.00%) high severe
```

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/image.rs
@@ -83,7 +83,7 @@ impl<'a> ImageFiller<'a> {
                     self.height_inv,
                 );
             }
-            
+
             match self.image.extends.0 {
                 peniko::Extend::Pad => self.run_simple::<F, Pad>(target, &y_positions),
                 peniko::Extend::Repeat => self.run_simple::<F, Repeat>(target, &y_positions),


sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -109,29 +84,48 @@ impl<'a> ImageFiller<'a> {
...
+        let mut x_pos = self.cur_pos.x;
+        let x_advance = self.image.x_advance.x;
+        let y_advance = self.image.y_advance.y;
+
+        let mut y_positions = [0.0; Tile::HEIGHT as usize];
+
+        for (idx, pos) in y_positions.iter_mut().enumerate() {
+            *pos = extend(
+                (self.cur_pos.y + y_advance * idx as f64) as f32,
+                self.image.extends.1,
```
taj-p: Did we want to pay the price of matching on this value redundantly when we can determine it once? I'm doubtful it'll have much impact on performance, so I'll leave it to your judgement

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -109,29 +84,48 @@ impl<'a> ImageFiller<'a> {
...
-                        .pixmap
-                        .sample(x_pos as u16, *y_pos as u16)
-                        .to_u8_array()[..],
-                ),
-                ImageQuality::Medium | ImageQuality::High => unimplemented!(),
-            };
-
-            pixel.copy_from_slice(&sample);
+    fn run_simple<F: FineType, E: Extend>(&mut self, target: &mut [F]) {
+        // Fast path. Each step in the x/y direction only updates x/y component of the
```
taj-p: Nit: should this be function documentation instead of a comment in the body?

taj-p: You could minimise impact on binary size by keeping `run_column_simple`, but making it generic on E and keeping these preliminary calculations (`y_positions`) inline and not generic

taj-p (approved):
Nice improvement in performance!! No comments are blocking

sparse_strips/vello_cpu/src/fine/image.rs
LaurenzV: I'm not sure I understand. 🤔 If we don't use generics (like on current main), then we have to do the matching inside of the loop. I initially thought that branch prediction would take care of this, but apparently not.

sparse_strips/vello_cpu/src/fine/image.rs
LaurenzV: I've lifted the `y_positions` out of function, but I will keep the whole loop in there now, so we only really need to do the matching once.

sparse_strips/vello_cpu/src/fine/image.rs
taj-p: I implemented a quick naive implementation of this idea (pulling the match out of the loop):

```rs
            match self.image.extends.1 {
                peniko::Extend::Pad => {
                    for (idx, pos) in y_positions.iter_mut().enumerate() {
                        *pos = Pad::extend(
                            (self.cur_pos.y + y_advance * idx as f64) as f32,
                            self.height,
                            self.height_inv,
                        );
                    }
                }
                peniko::Extend::Repeat => {
                    for (idx, pos) in y_positions.iter_mut().enumerate() {
                        *pos = Repeat::extend(
                            (self.cur_pos.y + y_advance * idx as f64) as f32,
                            self.height,
                            self.height_inv,
                        );
                    }
                }
                peniko::Extend::Reflect => {
                    for (idx, pos) in y_positions.iter_mut().enumerate() {
                        *pos = Reflect::extend(
                            (self.cur_pos.y + y_advance * idx as f64) as f32,
                            self.height,
                            self.height_inv,
                        );
                    }
                }
            }
```

I ran a couple benchmarks and it didn't show any improvement, which is probably expected since it's only run a few times and probably correctly branch predicted.

PR #1044, opened 2025-06-03
Add support for multi-threaded fine rasterization

LaurenzV:
This PR adds support for multi-threaded fine rasterization for `vello_cpu`. See here for a small demo, as you can see, we get near-linear speedup for large scenes where fine rasterization is the main bottleneck (note the frame count in the top right):

https://github.com/user-attachments/assets/bc1daab3-f1cf-48a7-b1bb-0c59cfd135e5

I also wanted to make sure that there is no significant performance impact from doing the splitting up into regions. Because of this, I rendered two simple scenes (with just a single rect) multiple times and then compared the timings:

10000 iterations to 600x600:

Before: 1.106366416s
After: 1.16696825s

1000 iterations to 3840x2160:

Before: 2.161953042s
After: 2.226217167s

So while there is a small slowdown, I don't think it's enough to justify creating a separate path for the single thread case. It's also a bit hard to test this automatically since it's based on feature flags, but I ran the tests manually and the diverging code is small enough that I think it should be manageable, at least for now. If making strip-rendering multi-threaded turns out to be more complicated perhaps we have to figure something out, though.

### DIFF
--- a/sparse_strips/vello_bench/src/fine/pack.rs
+++ b/sparse_strips/vello_bench/src/fine/pack.rs
@@ -1,13 +1,21 @@
 // Copyright 2025 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// See the comment in `Cargo.toml` for why we have those feature gates.
+
 use criterion::Bencher;
+#[cfg(not(feature = "multithreading"))]
 use vello_common::coarse::WideTile;
+#[cfg(not(feature = "multithreading"))]
 use vello_common::tile::Tile;
-use vello_cpu::fine::{Fine, FineType, SCRATCH_BUF_SIZE};
+#[cfg(not(feature = "multithreading"))]
+use vello_cpu::fine::SCRATCH_BUF_SIZE;
+use vello_cpu::fine::{Fine, FineType};
+#[cfg(not(feature = "multithreading"))]
 use vello_cpu::region::Regions;
 use vello_dev_macros::vello_bench;
 
+#[cfg(not(feature = "multithreading"))]
 #[vello_bench]
 pub fn pack<F: FineType>(b: &mut Bencher<'_>, fine: &mut Fine<F>) {
     let mut buf = vec![0; SCRATCH_BUF_SIZE];
@@ -19,4 +27,10 @@ pub fn pack<F: FineType>(b: &mut Bencher<'_>, fine: &mut Fine<F>) {
             std::hint::black_box(&r);
         });
     });
+}
+
+#[cfg(feature = "multithreading")]
+#[vello_bench]
+pub fn pack<F: FineType>(_: &mut Bencher<'_>, _: &mut Fine<F>) {
+    unimplemented!()
 }
--- a/sparse_strips/vello_bench/src/fine/pack.rs
+++ b/sparse_strips/vello_bench/src/fine/pack.rs
@@ -1,3 +1,6 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
 use criterion::Bencher;
 use vello_common::coarse::WideTile;
 use vello_common::tile::Tile;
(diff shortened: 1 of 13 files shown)


sparse_strips/vello_bench/src/fine/pack.rs
ajakubowicz-canva, line 1: Can this entire module be enabled/disabled using `#[cfg(not(feature = "multithreading"))]` [here](https://github.com/linebender/vello/pull/1044/files#diff-97f8dd49a3256b42392698f6cf0a351fa25752c1b2009d7a85ba91b6bae0bf87R20). Then this module doesn't need any `cfg` attributes as the whole module is only included if `multithreading` is enabled?

sparse_strips/vello_cpu/src/region.rs
```diff
@@ -0,0 +1,93 @@
...
+        self.regions.par_iter_mut().for_each(func);
+    }
+
+    #[cfg(not(feature = "multithreading"))]
+    pub fn update_regions(&mut self, func: impl FnMut(&mut Region<'_>)) {
+        self.regions.iter_mut().for_each(func);
+    }
+}
+
+/// A region that covers the pixel that fall into a specific wide tile.
```
ajakubowicz-canva: I'm not sure I understand this comment. Isn't the region equivalent to a wide tile?
Could this doc me updated (if it's accurate) to something like:

`A rectangular region containing the pixels from one wide tile.`

sparse_strips/vello_cpu/src/region.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Default, Debug)]
+pub struct Region<'a> {
+    /// The x coordinate of the wide tile this region covers.
+    pub(crate) x: usize,
+    /// The y coordinate of the wide tile this region covers.
+    pub(crate) y: usize,
```
ajakubowicz-canva: No change required – just curious why `x` and `y` are stored as `usize` instead of `u16` as per their usage in `render`:

```
                let wtile = self.wide.get(x as u16, y as u16);
                fine.set_coords(x as u16, y as u16);
```

My assumption is `u16` would result in a smaller struct.

sparse_strips/vello_bench/src/fine/pack.rs
LaurenzV: That would be the better solution, but the problem is that is that in `main.rs`, we cannot add the feature gates to the macro call. So this is kind of the only workaround I could come up with. This is definitely not ideal, though, and I hope we can come up with something better in the future.

sparse_strips/vello_cpu/src/region.rs
LaurenzV: They are equivalent for the general case, however, for the right/bottom edges, they might contain less pixels if the pixmap isn't a multiple of the width/height.

Will update the comment.

sparse_strips/vello_cpu/Cargo.toml
```diff
@@ -28,6 +31,7 @@ std = ["vello_common/std"]
 libm = ["vello_common/libm", "dep:libm"]
 # Allow loading Pixmap from PNG, and drawing png glyphs.
 png = ["vello_common/png"]
+multithreading = ["std", "dep:rayon", "dep:thread_local"]
```
DJMcNab, line 34: This feature should be documented in `src/lib.rs`, and therefore in the package README. It should also have a doc comment here

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -109,6 +114,14 @@ impl RenderContext {
         self.render_path(self.fill_rule, paint);
     }
 
+    /// Set the thread pool that should be used for multi-threaded rendering.
+    ///
+    /// If not set, the global rayon thread pool will be used instead.
+    #[cfg(feature = "multithreading")]
+    pub fn set_thread_pool(&mut self, pool: rayon::ThreadPool) {
```
DJMcNab, line 121: This is how you configure how many threads are used, right?

LaurenzV: Yes, though this will most likely change with the next PR which parallelized everything.

PR #1046, opened 2025-06-09
Document the `multithreading` feature in `vello_cpu`

LaurenzV:


sparse_strips/vello_cpu/README.md
```diff
@@ -92,6 +92,7 @@ assert_eq!(&result, expected_render);
 - `libm`: Use floating point implementations from [libm].
 - `png`(enabled by default): Allow loading [`Pixmap`]s from PNG images.
   Also required for rendering glyphs with an embedded PNG.
+- `multithreading`: Enable multi-threaded rendering (currently only a part of the pipeline can be parallelized).
```
DJMcNab: We could mention that this requires the standard library to be available (but skipping that now would be fine, as e.g. png also doesn't document that)

(Although rayon might be no_std compatible these days, if you use the right features and provide your own thread spawn function)

DJMcNab (approved):
Other than the readme sync caught by CI, looks good to me

sparse_strips/vello_cpu/README.md
LaurenzV: I think things will get a lot more complicated when the whole pipeline is multi-threading compatible, so I'm doubtful that we will be able to support `no_std` for that, but we shall see. 😄

nicoburns: I'd probably ditch the "(currently only a part of the pipeline can be parallelized)" part. That's not really relevant for users?

PR #1047, opened 2025-06-10
Add initial support for full multi-threaded rendering

LaurenzV:
# Overview

This PR is a first attempt at adding multi-threading support across the whole pipeline. The emphasis is on “first attempt”, because the current state is far from ideal, and I will definitely be refining and improving it in the near future, but I think it’s good enough to be merged as a first version, as it also puts many of the necessary building blocks in place. 

From my initial benchmarks, the gains are not impressive when drawing many simple shapes (in my case rectangles, especially small ones), with 8 threads I’m only getting a speedup for 2x-3x, for very small rectangles (8x8), multithreading is even slower. I will be doing more extensive profiling to understand the source of those slowdowns. However, as soon as more complex geometry like curves are involved, I’m currently seeing speedups ranging from 5x-6x, which I would say is at least a good start. There are some obvious things that need to be optimized (coarse rasterization currently always happens on the main thread, there is a lot of cloning happening to pass information around between threads), so hopefully there is room for improvement in the future. There's also some parts of the code I wouldn't exactly call beautiful (like the way we handle sending alphas from worker threads) and will probably rethink in the future.

I had to make some small changes to `vello_common`, but other than that nearly everything is contained in `vello_cpu`. Perhaps some of the abstractions I introduced (like the `StripGenerator`) could later on be moved to `vello_common` and be reused by `vello_hybrid`, but I'm not sure how useful they would be.

# Changes
Most of the code has been commented, but to provide a small summary of the main changes:
- In addition to holding an index to an alpha buffer, alpha strips also hold an identifier that identifies the thread the alpha buffer belongs to.
- All of the tile/strip generation code has been encapsulated into a `StripGenerator` struct, that provide a callback to the generated strips which can then be processed by clients.
- A `Dispatcher` trait is introduced, that has a very similar interface to the current render context, but explicitly takes a reference to all the arguments it needs. The main motivation for this is that multi-threaded and single-threaded rendering are different enough that it's not really possible to implement them in the same struct (for example, for multi-threaded rendering we need an owned version of its arguments, while for single-threaded rendering this would be needless). In addition, by abstracting this away behind the `Dispatcher` trait, even if you enable the `multithreading` feature, you can still opt to just using the single-threaded mode without any overhead.

Conceptually, the current way multi-threading is not too complex: The main thread sends paths to worker threads, which will then turn the path into a strips representation. Those worker threads then send those strips back to the main thread. Order is obviously important here, which is why each operation is assigned a `cmd_idx`, and when reading from the ordered channel, the results will be received in the correct order by the main thread. Then, we simply do coarse rasterization on the main thread and finally multi-threaded fine rasterization, as it already exists on main.

An important detail is that we also do batching of tasks: If we draw a lot of small paths, sending one path per thread would lead to an incredible overhead due to frequent context switching. Because of this, we try to very roughly estimate how long it will take to process the path, and based on that batch multiple paths and then send them all together to one thread. The current heuristic seems to be working at least okay, but when I get to fine-tuning I will definitely do more careful profiling for that to see whether it can be improved.

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -20,10 +20,6 @@ pub struct Strip {
     pub y: u16,
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
-    /// The index of the thread that contains the alpha values
-    /// pointed to by `alpha_idx`.
-    #[cfg(feature = "multithreading")]
-    pub thread_idx: u8,
     /// The winding number at the start of the strip.
     pub winding: i32,
 }
@@ -44,29 +40,7 @@ pub fn render(
     fill_rule: Fill,
     lines: &[Line],
 ) {
-    render_inner(
-        tiles,
-        strip_buf,
-        alpha_buf,
-        #[cfg(feature = "multithreading")]
-        0,
-        fill_rule,
-        lines,
-    );
-}
-
-/// Render the tiles stored in `tiles` into the strip and alpha buffer.
-/// The strip buffer will be cleared in the beginning.
-#[cfg(feature = "multithreading")]
-pub fn render_with_thread_idx(
-    tiles: &Tiles,
-    strip_buf: &mut Vec<Strip>,
-    alpha_buf: &mut Vec<u8>,
-    thread_idx: u8,
-    fill_rule: Fill,
-    lines: &[Line],
-) {
-    render_inner(tiles, strip_buf, alpha_buf, thread_idx, fill_rule, lines);
+    render_inner(tiles, strip_buf, alpha_buf, fill_rule, lines);
 }
 
 /// Render the tiles stored in `tiles` into the strip and alpha buffer.
@@ -75,7 +49,6 @@ pub fn render_inner(
     tiles: &Tiles,
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
-    #[cfg(feature = "multithreading")] thread_idx: u8,
     fill_rule: Fill,
     lines: &[Line],
 ) {
@@ -107,8 +80,6 @@ pub fn render_inner(
         x: prev_tile.x * Tile::WIDTH,
         y: prev_tile.y * Tile::HEIGHT,
         alpha_idx: alpha_buf.len() as u32,
-        #[cfg(feature = "multithreading")]
-        thread_idx,
         winding: 0,
     };
 
@@ -171,8 +142,6 @@ pub fn render_inner(
                         x: u16::MAX,
                         y: prev_tile.y * Tile::HEIGHT,
                         alpha_idx: alpha_buf.len() as u32,
-                        #[cfg(feature = "multithreading")]
-                        thread_idx,
                         winding: winding_delta,
                     });
                 }
@@ -194,8 +163,6 @@ pub fn render_inner(
                 x: tile.x * Tile::WIDTH,
                 y: tile.y * Tile::HEIGHT,
                 alpha_idx: alpha_buf.len() as u32,
-                #[cfg(feature = "multithreading")]
-                thread_idx,
                 winding: winding_delta,
             };
             // Note: this fill is mathematically not necessary. It provides a way to reduce
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -41,6 +41,40 @@ pub fn render(
     tiles: &Tiles,
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
+    fill_rule: Fill,
+    lines: &[Line],
+) {
+    render_inner(
+        tiles,
+        strip_buf,
+        alpha_buf,
+        #[cfg(feature = "multithreading")]
+        0,
+        fill_rule,
+        lines,
+    );
+}
+
+/// Render the tiles stored in `tiles` into the strip and alpha buffer.
+/// The strip buffer will be cleared in the beginning.
+#[cfg(feature = "multithreading")]
+pub fn render_with_thread_idx(
+    tiles: &Tiles,
+    strip_buf: &mut Vec<Strip>,
+    alpha_buf: &mut Vec<u8>,
+    thread_idx: u8,
+    fill_rule: Fill,
+    lines: &[Line],
+) {
+    render_inner(tiles, strip_buf, alpha_buf, thread_idx, fill_rule, lines);
+}
+
+/// Render the tiles stored in `tiles` into the strip and alpha buffer.
+/// The strip buffer will be cleared in the beginning.
+pub fn render_inner(
+    tiles: &Tiles,
+    strip_buf: &mut Vec<Strip>,
+    alpha_buf: &mut Vec<u8>,
     #[cfg(feature = "multithreading")] thread_idx: u8,
     fill_rule: Fill,
     lines: &[Line],
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -20,6 +20,10 @@ pub struct Strip {
     pub y: u16,
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
+    /// The index of the thread that contains the alpha values
+    /// pointed to by `alpha_idx`. Can be ignored if multi-threaded rendering
+    /// is not enabled/supported.
+    pub thread_idx: u16,
     /// The winding number at the start of the strip.
     pub winding: i32,
 }
@@ -37,6 +41,7 @@ pub fn render(
     tiles: &Tiles,
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
+    thread_idx: u16,
     fill_rule: Fill,
     lines: &[Line],
 ) {
@@ -68,6 +73,7 @@ pub fn render(
         x: prev_tile.x * Tile::WIDTH,
         y: prev_tile.y * Tile::HEIGHT,
         alpha_idx: alpha_buf.len() as u32,
+        thread_idx,
         winding: 0,
     };
 
@@ -130,6 +136,7 @@ pub fn render(
                         x: u16::MAX,
                         y: prev_tile.y * Tile::HEIGHT,
                         alpha_idx: alpha_buf.len() as u32,
+                        thread_idx,
                         winding: winding_delta,
                     });
                 }
@@ -151,6 +158,7 @@ pub fn render(
                 x: tile.x * Tile::WIDTH,
                 y: tile.y * Tile::HEIGHT,
                 alpha_idx: alpha_buf.len() as u32,
+                thread_idx,
                 winding: winding_delta,
             };
             // Note: this fill is mathematically not necessary. It provides a way to reduce
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -40,18 +40,6 @@ pub fn render(
     fill_rule: Fill,
     lines: &[Line],
 ) {
-    render_inner(tiles, strip_buf, alpha_buf, fill_rule, lines);
-}
-
-/// Render the tiles stored in `tiles` into the strip and alpha buffer.
-/// The strip buffer will be cleared in the beginning.
-pub fn render_inner(
-    tiles: &Tiles,
-    strip_buf: &mut Vec<Strip>,
-    alpha_buf: &mut Vec<u8>,
-    fill_rule: Fill,
-    lines: &[Line],
-) {
     strip_buf.clear();
 
     if tiles.is_empty() {
@@ -317,3 +305,4 @@ pub fn render_inner(
         }
     }
 }
+
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -21,7 +21,7 @@ pub struct Strip {
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
     /// The index of the thread that contains the alpha values
-    /// pointed to by `alpha_idx`. 
+    /// pointed to by `alpha_idx`.
     #[cfg(feature = "multithreading")]
     pub thread_idx: u8,
     /// The winding number at the start of the strip.
@@ -41,8 +41,7 @@ pub fn render(
     tiles: &Tiles,
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
-    #[cfg(feature = "multithreading")]
-    thread_idx: u8,
+    #[cfg(feature = "multithreading")] thread_idx: u8,
     fill_rule: Fill,
     lines: &[Line],
 ) {
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -305,4 +305,3 @@ pub fn render(
         }
     }
 }
-
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -23,7 +23,7 @@ pub struct Strip {
     /// The index of the thread that contains the alpha values
     /// pointed to by `alpha_idx`. 
     #[cfg(feature = "multithreading")]
-    pub thread_idx: u16,
+    pub thread_idx: u8,
     /// The winding number at the start of the strip.
     pub winding: i32,
 }
@@ -42,7 +42,7 @@ pub fn render(
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
     #[cfg(feature = "multithreading")]
-    thread_idx: u16,
+    thread_idx: u8,
     fill_rule: Fill,
     lines: &[Line],
 ) {
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -21,8 +21,8 @@ pub struct Strip {
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
     /// The index of the thread that contains the alpha values
-    /// pointed to by `alpha_idx`. Can be ignored if multi-threaded rendering
-    /// is not enabled/supported.
+    /// pointed to by `alpha_idx`. 
+    #[cfg(feature = "multithreading")]
     pub thread_idx: u16,
     /// The winding number at the start of the strip.
     pub winding: i32,
@@ -41,6 +41,7 @@ pub fn render(
     tiles: &Tiles,
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
+    #[cfg(feature = "multithreading")]
     thread_idx: u16,
     fill_rule: Fill,
     lines: &[Line],
@@ -73,6 +74,7 @@ pub fn render(
         x: prev_tile.x * Tile::WIDTH,
         y: prev_tile.y * Tile::HEIGHT,
         alpha_idx: alpha_buf.len() as u32,
+        #[cfg(feature = "multithreading")]
         thread_idx,
         winding: 0,
     };
@@ -136,6 +138,7 @@ pub fn render(
                         x: u16::MAX,
                         y: prev_tile.y * Tile::HEIGHT,
                         alpha_idx: alpha_buf.len() as u32,
+                        #[cfg(feature = "multithreading")]
                         thread_idx,
                         winding: winding_delta,
                     });
@@ -158,6 +161,7 @@ pub fn render(
                 x: tile.x * Tile::WIDTH,
                 y: tile.y * Tile::HEIGHT,
                 alpha_idx: alpha_buf.len() as u32,
+                #[cfg(feature = "multithreading")]
                 thread_idx,
                 winding: winding_delta,
             };
--- a/sparse_strips/vello_cpu/src/dispatch/mod.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/mod.rs
@@ -1,3 +1,6 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
 #[cfg(feature = "multithreading")]
 pub(crate) mod multi_threaded;
 pub(crate) mod single_threaded;
--- a/sparse_strips/vello_cpu/src/dispatch/mod.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/mod.rs
@@ -1,0 +1,23 @@
+pub(crate) mod single_threaded;
+#[cfg(feature = "multithreading")]
+pub(crate) mod multi_threaded;
+
+use core::fmt::Debug;
+use vello_common::coarse::Wide;
+use vello_common::encode::EncodedPaint;
+use vello_common::mask::Mask;
+use vello_common::paint::Paint;
+use crate::kurbo::{Affine, BezPath, Stroke};
+use crate::peniko::{BlendMode, Fill};
+use crate::RenderMode;
+
+pub(crate) trait Dispatcher: Debug {
+    fn wide(&self) -> &Wide;
+    fn fill_path(&mut self, path: &BezPath, fill_rule: Fill, transform: Affine, paint: Paint);
+    fn stroke_path(&mut self, path: &BezPath, stroke: &Stroke, transform: Affine, paint: Paint);
+    fn push_layer(&mut self, clip_path: Option<&BezPath>, fill_rule: Fill, clip_transform: Affine, blend_mode: BlendMode, opacity: f32, mask: Option<Mask>);
+    fn pop_layer(&mut self);
+    fn reset(&mut self);
+    fn flush(&mut self);
+    fn rasterize(&self, buffer: &mut [u8], render_mode: RenderMode, width: u16, height: u16, encoded_paints: &[EncodedPaint]);
+}
--- a/sparse_strips/vello_cpu/src/dispatch/mod.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/mod.rs
@@ -1,23 +1,38 @@
-pub(crate) mod single_threaded;
 #[cfg(feature = "multithreading")]
 pub(crate) mod multi_threaded;
+pub(crate) mod single_threaded;
 
+use crate::RenderMode;
+use crate::kurbo::{Affine, BezPath, Stroke};
+use crate::peniko::{BlendMode, Fill};
 use core::fmt::Debug;
 use vello_common::coarse::Wide;
 use vello_common::encode::EncodedPaint;
 use vello_common::mask::Mask;
 use vello_common::paint::Paint;
-use crate::kurbo::{Affine, BezPath, Stroke};
-use crate::peniko::{BlendMode, Fill};
-use crate::RenderMode;
 
 pub(crate) trait Dispatcher: Debug {
     fn wide(&self) -> &Wide;
     fn fill_path(&mut self, path: &BezPath, fill_rule: Fill, transform: Affine, paint: Paint);
     fn stroke_path(&mut self, path: &BezPath, stroke: &Stroke, transform: Affine, paint: Paint);
-    fn push_layer(&mut self, clip_path: Option<&BezPath>, fill_rule: Fill, clip_transform: Affine, blend_mode: BlendMode, opacity: f32, mask: Option<Mask>);
+    fn push_layer(
+        &mut self,
+        clip_path: Option<&BezPath>,
+        fill_rule: Fill,
+        clip_transform: Affine,
+        blend_mode: BlendMode,
+        opacity: f32,
+        mask: Option<Mask>,
+    );
     fn pop_layer(&mut self);
     fn reset(&mut self);
     fn flush(&mut self);
-    fn rasterize(&self, buffer: &mut [u8], render_mode: RenderMode, width: u16, height: u16, encoded_paints: &[EncodedPaint]);
+    fn rasterize(
+        &self,
+        buffer: &mut [u8],
+        render_mode: RenderMode,
+        width: u16,
+        height: u16,
+        encoded_paints: &[EncodedPaint],
+    );
 }
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -174,19 +174,21 @@ impl MultiThreadedDispatcher {
             match result_receiver.try_recv() {
                 Ok(cmd) => match cmd {
                     CoarseCommand::Render {
+                        thread_id,
                         strips,
                         fill_rule,
                         paint,
-                    } => self.wide.generate(&strips, fill_rule, paint.clone()),
+                    } => self.wide.generate(&strips, fill_rule, paint, thread_id),
                     CoarseCommand::PushLayer {
+                        thread_id,
                         clip_path,
                         blend_mode,
                         mask,
                         opacity,
                     } => self
                         .wide
-                        .push_layer(clip_path.clone(), blend_mode, mask.clone(), opacity),
-                    CoarseCommand::PopLayer => self.wide.pop_layer(),
+                        .push_layer(clip_path, blend_mode, mask, opacity, thread_id),
+                    CoarseCommand::PopLayer { thread_id } => self.wide.pop_layer(),
                 },
                 Err(e) => match e {
                     TryRecvError::Empty => {
@@ -425,17 +427,21 @@ impl RenderTask {
 
 enum CoarseCommand {
     Render {
+        thread_id: u8,
         strips: Box<[Strip]>,
         fill_rule: Fill,
         paint: Paint,
     },
     PushLayer {
+        thread_id: u8,
         clip_path: Option<(Box<[Strip]>, Fill)>,
         blend_mode: BlendMode,
         mask: Option<Mask>,
         opacity: f32,
     },
-    PopLayer,
+    PopLayer {
+        thread_id: u8,
+    },
 }
 
 #[derive(Debug)]
@@ -452,7 +458,7 @@ impl Worker {
         thread_id: u8,
         alpha_storage: Arc<Mutex<HashMap<u8, Arc<Vec<u8>>>>>,
     ) -> Self {
-        let strip_generator = StripGenerator::new(width, height, thread_id);
+        let strip_generator = StripGenerator::new(width, height);
 
         Self {
             strip_generator,
@@ -477,6 +483,7 @@ impl Worker {
                 } => {
                     let func = |strips: &[Strip]| {
                         let coarse_command = CoarseCommand::Render {
+                            thread_id: self.thread_id,
                             strips: strips.into(),
                             fill_rule,
                             paint,
@@ -496,6 +503,7 @@ impl Worker {
                 } => {
                     let func = |strips: &[Strip]| {
                         let coarse_command = CoarseCommand::Render {
+                            thread_id: self.thread_id,
                             strips: strips.into(),
                             fill_rule: Fill::NonZero,
                             paint,
@@ -529,6 +537,7 @@ impl Worker {
                     };
 
                     let coarse_command = CoarseCommand::PushLayer {
+                        thread_id: self.thread_id,
                         clip_path: clip,
                         blend_mode,
                         mask,
@@ -539,7 +548,12 @@ impl Worker {
                 }
                 RenderTask::PopLayer { task_idx } => {
                     result_sender
-                        .send(task_idx, CoarseCommand::PopLayer)
+                        .send(
+                            task_idx,
+                            CoarseCommand::PopLayer {
+                                thread_id: self.thread_id,
+                            },
+                        )
                         .unwrap();
                 }
             }
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -16,7 +16,7 @@ use crossbeam_channel::TryRecvError;
 use rayon::{ThreadPool, ThreadPoolBuilder};
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::sync::Mutex;
+use std::sync::{Barrier, Mutex};
 use std::sync::atomic::{AtomicU8, AtomicU32, Ordering};
 use thread_local::ThreadLocal;
 use vello_common::coarse::{Cmd, Wide};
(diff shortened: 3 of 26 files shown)


sparse_strips/vello_common/src/strip.rs
```diff
@@ -20,6 +20,10 @@ pub struct Strip {
     pub y: u16,
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
+    /// The index of the thread that contains the alpha values
+    /// pointed to by `alpha_idx`. Can be ignored if multi-threaded rendering
+    /// is not enabled/supported.
+    pub thread_idx: u16,
```
taj-p:
```suggestion
    #[cfg(feature = "multithreading")]
    pub thread_idx: u16,
```

I'm not sure we want to pay the extra 16 bytes unnecessarily for single threaded applications for each strip and `Cmd` (`CmdClipAlphaFill` seems to be the largest variant, so would increase the cost of storing each cmd by 16 bytes). 

I think this struct is very close to being 64 bytes (to fit into a single cache line), but this thread_idx makes it a bit harder for us to compress.

What are your thoughts of feature gating this field?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,625 @@
...
+
+pub(crate) struct MultiThreadedDispatcher {
+    wide: Wide,
+    thread_pool: ThreadPool,
+    task_batch: Vec<RenderTask>,
+    batch_cost: f32,
+    task_sender: Option<RenderTasksSender>,
+    workers: Arc<ThreadLocal<RefCell<Worker>>>,
+    result_receiver: Option<CoarseCommandReceiver>,
+    alpha_storage: Arc<Mutex<HashMap<u16, Arc<Vec<u8>>>>>,
```
taj-p: Nit: Is this a mapping from thread index to alphas? Do we need to pay the fairly large `HashMap` cost or would a sparse array (or simply a vec where each index refers to some thread) provide the same functionality without going through hashing?

sparse_strips/vello_common/src/strip.rs
taj-p: Given `thread_idx` isn't actually used that frequently (once in generation and again in dispatch), I actually wonder whether we'll see greater performance (via improved CPU cache utilisation at earlier stages of the pipeline) if `thread_idx` is moved out of this struct into a separate store. (This one is totally optional, just an observation, and can be investigated outside of this PR).

edit: We would need to get Strip down to 64 bytes for this to be effective

taj-p: Super nit and question: I think only super computers/really expensive servers support more than 255 threads. If we deployed Vello to a server, I would think you'd scale whatever program uses Vello before allocating all available threads to a single program that uses Vello (and you probably wouldn't want all those threads on rendering?). Should our `thread_idx` be a `u8`?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Yeah, this is one of the design parts I'm not happy with. Since the thread IDs should be contiguous, perhaps we can just use a vector. But I hope I I can come up with some fundamentally different way to handle the alphas in the future.

sparse_strips/vello_common/src/strip.rs
LaurenzV: Sure, I think feature-gating would be fine!

I'm not sure how the store would solve the problem, since we still need to somehow associate each command with the location in the store, no?

LaurenzV: Yes, good point, I think `u8` should be more than enough!

taj-p: > I'm not sure how the store would solve the problem, since we still need to somehow associate each command with the location in the store, no?

Yes, I think so. I haven't read through the code in a manner that makes me confident with this answer (I can take a closer look tomorrow?), but perhaps the `thread_idx` can be returned by the thread one time for the N strips it generates

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,625 @@
...
+        &self,
+        buffer: &mut [u8],
+        width: u16,
+        height: u16,
+        encoded_paints: &[EncodedPaint],
+    ) {
+        let mut buffer = Regions::new(width, height, buffer);
+        let fines = ThreadLocal::new();
+        let wide = &self.wide;
+        let alpha_map = self.alpha_storage.lock().unwrap().clone();
```
nicoburns: Are you intending to clone the HashMap inside the `Arc<Mutex>` here?

nicoburns:
> there is a lot of cloning happening to pass information around between threads

It definitely feels like a lot of the cloning (and perhaps even some of the channels?) could in principle be implemented with just references? Perhaps with a slightly different backing data structures such that the locations in memory can be more easily guaranteed to stick around for the duration of the thread. It may be that some kind of "scoped task" abstraction (mirroring scoped threads, but with a thread pool) could also help here.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Yes, the reason is that we need to Access the Alphas for each wide tile command, so always having to access it through a mutex will likely be very expensive. Since the alphas in the hashmap are wrapped in an Arc, this shouldn’t be too expensive.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
ajakubowicz-canva: Is it worth documenting this in the code via a TODO or some other comment? Specifically document that thread IDs should be contiguous and a better data structure should be evaluated.

LaurenzV: I mean, there is _a lot_ of TODOs for improving the performance/design of this, so I'm not sure it's worth adding a TODO for each of them. :p

Will be addressing the most pressing reviews today, though.

ajakubowicz-canva: Fair :P

sparse_strips/vello_sparse_tests/tests/util.rs
```diff
@@ -32,8 +32,17 @@ static DIFFS_PATH: std::sync::LazyLock<PathBuf> = std::sync::LazyLock::new(|| {
...
-pub(crate) fn get_ctx<T: Renderer>(width: u16, height: u16, transparent: bool) -> T {
-    let mut ctx = T::new(width, height);
+pub(crate) fn get_ctx<T: Renderer>(
+    width: u16,
+    height: u16,
+    transparent: bool,
+    multi_threaded: bool,
+) -> T {
+    let mut ctx = if multi_threaded {
+        T::new_multithreaded(width, height, 1)
```
ajakubowicz-canva: Out of curiosity and naivety, why test with only a single thread vs 2, or more? Would it run into the issues described in the [cargo-nextest docs](https://nexte.st/docs/configuration/threads-required/#heavy-tests-and-threads-required)?

LaurenzV: Wow good catch... No idea what happened, I could've sworn I put 4 threads there. 😅 Thanks, will fix this, we should definitely be running with multiple threads to ensure there are no race conditions.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -39,6 +43,41 @@ pub fn render(
...
+    render_inner(tiles, strip_buf, alpha_buf, thread_idx, fill_rule, lines);
+}
+
+/// Render the tiles stored in `tiles` into the strip and alpha buffer.
+/// The strip buffer will be cleared in the beginning.
+pub fn render_inner(
+    tiles: &Tiles,
+    strip_buf: &mut Vec<Strip>,
+    alpha_buf: &mut Vec<u8>,
+    #[cfg(feature = "multithreading")] thread_idx: u8,
```
ajakubowicz-canva: Nit: Is there a convention about conditionally compiled arguments position? E.g. would it be easier (for the caller) if `thread_idx` was in the last position. Then if switching between features you have less of an awkward change.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -109,29 +119,36 @@ impl RenderContext {
...
-        self.render_path(Fill::NonZero, paint);
+        self.dispatcher
+            .stroke_path(path, &self.stroke, self.transform, paint);
     }
 
     /// Fill a rectangle.
     pub fn fill_rect(&mut self, rect: &Rect) {
-        self.fill_path(&rect.to_path(DEFAULT_TOLERANCE));
+        // Don't use `rect.to_path` here, because it will perform a new allocation, which
+        // profiling showed can become a bottleneck for many small rectangles.
```
ajakubowicz-canva, line 137: [No action for this PR] Doesn't this situation also happen with the blurred rectangle example below? Is it worth providing an alternative or modified API to `to_path` such that an existing `temp_path` can be passed in and used instead of allocating?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -281,11 +299,16 @@ impl RenderContext {
...
+    }
+
+    /// Flush any pending operations.
+    ///
+    /// This is a no-op when using the single-threaded render mode, and can be ignored.
+    /// For multi-threaded rendering, you _have_ to call this before rasterizing, otherwise
+    /// the program will panic.
+    pub fn flush(&mut self) {
+        self.dispatcher.flush();
     }
```
ajakubowicz-canva, line 315: Am I correct to assume that there is no way to achieve multithreaded without extending the public API with `flush`?

Couldn't rasterizing call `flush` before rasterising, hence ensuring that flush is called? Or is it valuable to provide this control to the caller?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -4,59 +4,73 @@
...
     pub(crate) paint: PaintType,
     pub(crate) paint_transform: Affine,
     pub(crate) stroke: Stroke,
     pub(crate) transform: Affine,
     pub(crate) fill_rule: Fill,
+    pub(crate) temp_path: BezPath,
     pub(crate) encoded_paints: Vec<EncodedPaint>,
-    #[cfg(feature = "multithreading")]
-    pub(crate) thread_pool: Option<rayon::ThreadPool>,
+    dispatcher: Box<dyn Dispatcher>,
```
ajakubowicz-canva, line 43: Not blocking, just curious: Isn't there a performance penalty for using `dyn`. In this case where there are two private internal types that implement `Dispatcher`, could this avoid the `Box` and `dyn` by being an enum containing one of the two concrete types?
Although, I guess then the enum will become as wide as the larger multithreaded variant. Out of interest, have you benchmarked the difference?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+        self.thread_pool.spawn_broadcast(move |_| {
+            let worker = workers.get().unwrap();
+            let mut borrowed = worker.borrow_mut();
+            borrowed.reset();
+            t.fetch_sub(1, Ordering::SeqCst);
+        });
+
+        // TODO: Maybe there is a better way of doing this? Something like a `WaitGroup`, but that
+        // can be used in conjunction with `spawn_broadcast`.
+        while thread_counter.load(Ordering::SeqCst) > 0 {}
```
ajakubowicz-canva: It feels like the _simplest_ alternative would be to use a [Barrier](https://doc.rust-lang.org/std/sync/struct.Barrier.html).

Set up the barrier outside the spawn_broadcast:
`let barrier = Arc::new(Barrier::new(self.num_threads + 1));`
Then place a `b.wait()` after the `borrowed.reset();` and a `b.wait()` here. Once the threads + the main thread all count down the barrier then they'll all continue together.

ajakubowicz-canva: I'm not certain of the performance characteristics, I feel like that should inform the change.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+            path_length,
+        }
+    }
+}
+
+// This formula was derived by recording benchmarks of how long paths with a certain set of properties
+// take to render and then use "machine learning" to derive a formula which approximates the runtime.
+// The result will be far from accurate, but all we need is a ballpark range, and it's important
+// that the calculation is relatively fast, since it will be run on the main thread for each path.
+// TODO: We will need to update the formula once SIMD + faster stroke expansion landed.
```
ajakubowicz-canva, line 641: It feels like this `estimate_runtime_in_micros` is going to be tricky to keep synced as changes happen to the codebase. It's also possible that different compile targets may have different runtime performance characteristics across the various features. This may become more pronounced when there is SIMD, as different SIMD architecture targets (and the fallback) may cause this estimate to be significantly different based on the compilation.

Maybe tracked in some issue, is there some alternative where this estimate can be figured out at runtime. E.g. batches are created and their contents recorded. Then maybe they are timed, and some cost is updated so subsequent batches are better estimated. This sounds like a whole complex feature though.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
ajakubowicz-canva: Something to consider in a followup, I think the using a `Vec` instead of a `HashMap` should surely be faster, as it provides direct access to the alphas.

I also churned a bunch of thoughts around this `alpha_map` clone. I experimented using a `RWLock` or just holding the lock over the entire raster method. But, in the end I've come to agree that cloning a small alpha_storage containing ~8 Arcs should surely be better than multiple traversals through locks.

I guess my question is, is there anyone else contending with this Mutex if we hold the alpha_storage mutex for the entirety of `rasterize` to avoid the clone? E.g. remove `.clone()` and share the alpha_storage directly via immutable reference to `update_regions_par`.


Edit: Provided each thread only writes to some disjoint part of the alpha_storage collection, technically we could utilise unsafe to avoid locks entirely. Maybe? But that would be `unsafe`.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+    thread_pool: ThreadPool,
+    task_batch: Vec<RenderTask>,
+    batch_cost: f32,
+    task_sender: Option<RenderTasksSender>,
+    workers: Arc<ThreadLocal<RefCell<Worker>>>,
+    result_receiver: Option<CoarseCommandReceiver>,
+    alpha_storage: Arc<Mutex<HashMap<u8, Arc<Vec<u8>>>>>,
+    task_idx: usize,
+    num_threads: u16,
+    flushed: RefCell<bool>,
```
ajakubowicz-canva: I don't think this needs to be a `RefCell`. `flushed` is only used in the context of `&self mut` methods. I was able to replace this with `bool` without other changes.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+type CoarseCommandReceiver = ordered_channel::Receiver<CoarseCommand>;
+
+pub(crate) struct MultiThreadedDispatcher {
+    wide: Wide,
+    thread_pool: ThreadPool,
+    task_batch: Vec<RenderTask>,
+    batch_cost: f32,
+    task_sender: Option<RenderTasksSender>,
+    workers: Arc<ThreadLocal<RefCell<Worker>>>,
+    result_receiver: Option<CoarseCommandReceiver>,
```
ajakubowicz-canva, line 44: Extreme ignorable bike-shedding nit. I think the prefix `task_sender` and `result_receiver` made me think they were quite separate concepts. Really, they are tightly bound, with `task_sender` causing the `result_receiver` to receive the resulting course commands.

Maybe these can be simplified to `sender` and `receiver` making it more clear that they're linked?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+            let worker = workers.get().unwrap();
+            let mut worker = worker.borrow_mut();
+
+            while let Ok(tasks) = render_task_receiver.recv() {
+                worker.run_tasks(tasks, &mut coarse_command_sender);
+            }
+
+            // If we reach this point, it means the task_sender has been dropped by the main thread
+            // and no more tasks are available. So we are done, and just need to place the alphas
+            // of the worker thread in the hashmap. Then, we drop the `result_sender`. Once all
```
ajakubowicz-canva, line 121: Nit:
```suggestion
            // of the worker thread in the hashmap. Then, we drop the `coarse_command_sender`. Once all
```

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+            // worker thread have dropped `result_sender`, the main thread knows that all alphas
+            // have been placed, and it's safe to proceed.
```
ajakubowicz-canva: Nit: 
```suggestion
            // If we reach this point, it means the task_sender has been dropped by the main thread
            // and no more tasks are available. So we are done, and just need to place the alphas
            // of the worker thread in the hashmap. Then, we drop the `coarse_command_sender`. Once all
            // worker thread have dropped `coarse_command_sender`, the main thread knows that all alphas
            // have been placed, and it's safe to proceed.
```

Although maybe this comment is related to the [naming comment](https://github.com/linebender/vello/pull/1047/files#r2144088222). Maybe `receiver` and `sender` is enough.

ajakubowicz-canva: Because the dropping of the channel is semantically important, might it be useful to explicitly call `drop(coarse_command_sender)` at the end of the scope so it's super clear and not implicit?

sparse_strips/vello_cpu/src/render.rs
ajakubowicz-canva: There are "batched internal flushes" done if the `COST_THRESHOLD` is reached, done implicitly. This further makes me wonder why we don't make this `flush` implicit as well? Especially as it's assumed the it'll have a cost under the `COST_THRESHOLD`.
I may have also misunderstood.

ajakubowicz-canva (commented):
Awesome! This is looking great!

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -541,38 +497,3 @@ impl ColrRenderer for RenderContext {
...
-        let mut ctx = RenderContext::new(100, 100);
-        let rect = Rect::new(0.0, 0.0, 100.0, 100.0);
-
-        ctx.fill_rect(&rect);
-
-        assert!(!ctx.line_buf.is_empty());
-        assert!(!ctx.strip_buf.is_empty());
-        assert!(!ctx.alphas.is_empty());
-
-        ctx.reset();
```
ajakubowicz-canva, line 561: After removing this test `reset` is no longer exercised anywhere.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -4,59 +4,73 @@
...
-    pub(crate) line_buf: Vec<Line>,
-    pub(crate) tiles: Tiles,
-    pub(crate) strip_buf: Vec<Strip>,
     pub(crate) paint: PaintType,
     pub(crate) paint_transform: Affine,
     pub(crate) stroke: Stroke,
     pub(crate) transform: Affine,
     pub(crate) fill_rule: Fill,
+    pub(crate) temp_path: BezPath,
     pub(crate) encoded_paints: Vec<EncodedPaint>,
```
ajakubowicz-canva, line 42: I am not clear on the lifecycle of `RenderContext`. If rendering an animation with vello_cpu, would you create a new `RenderContext` between each frame, or would you maintain a single `RenderContext` doing the following:
 - Call stateful APIs.
 - Rasterize
 - Reset

After calling `reset`, the `RenderContext` should be ready to render the next frame from a clean slate.
If this is the case, then I think `encoded_paints` is currently not being reset and would leak memory? I may have missed where `encoded_paints` is cleared.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: I want to avoid `unsafe` in this first prototype, but will definitely experiment with this and am open to adding `unsafe` if it gives real benefits later on.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: The only reason flushing is separate and doesn't happen in `render_to_pixmap` is that currently the function takes `&self`, but to flush we need to have `&mut self`. We could change `render_to_pixmap` to take a `&mut self`, but somehow it feels wrong to me, because usually a user probably wouldn't expect this function to mutate the state of the render context.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: I haven't benchmarked the difference, but I highly doubt it makes a difference. It's not like the dynamic dispatching happens for every pixel, but it only happens once for each `fill_path`/`stroke_path` call, and there are a lot of other things happening during those calls that I don't think it would make any difference. If it does, I will hopefully notice once I do more careful profiling.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
nicoburns: `Vec<ArcSwap>` (https://docs.rs/arc-swap/latest/arc_swap/) might work without unsafe?

sparse_strips/vello_cpu/src/render.rs
nicoburns: `reset()` is definitely intended to allow reuse. Although I think just for the "render the next frame" use case you ought to be able to reuse a `RenderContext` even without calling `reset()`.

There was previously a bug with this that only manifested when reusing a `RenderContext` after catching a `panic!()` (in Blitz's test runner) left it in an inconsistent state. It might be a useful thing to test.

LaurenzV: The reason I has to delete it is that the data is now not directly accessible anymore through RenderContext, but I’ll see whether I can maybe add it to the Dispatcher classes.

sparse_strips/vello_cpu/src/render.rs
nicoburns: I don't think an enum would help here as branching on the enum is essentially the same as dynamic dispatch (just manual). But I do wonder whether this ought to just always use the multithreaded Dispatcher if the feature is enabled.

LaurenzV: I think it's a really good advantage to be able to still fall back to the single-threaded dispatcher if 0 threads are passed, on the one hand because there is some overhead to using the multi-threaded dispatcher with just one thread, on the other hand also because it makes it much easier to test both in `vello_sparse_tests`. So I would strongly prefer to keep that behavior.

sparse_strips/vello_cpu/src/dispatch/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+use crate::RenderMode;
+use crate::kurbo::{Affine, BezPath, Stroke};
+use crate::peniko::{BlendMode, Fill};
+use core::fmt::Debug;
+use vello_common::coarse::Wide;
+use vello_common::encode::EncodedPaint;
+use vello_common::mask::Mask;
+use vello_common::paint::Paint;
+
+pub(crate) trait Dispatcher: Debug {
```
taj-p, line 17: [Not actionable]: I'm interested if you've had any thoughts around how the methods defined in `Dispatcher` are hiding how its implementations would like to best receive its arguments (specifically, by borrowing or taking ownership).

Through this trait, both the single and multi threaded implementations take a `&BezPath` even though the multi threaded implementation must always `.clone()` it. This means there are cases where we [take a reference](https://github.com/linebender/vello/blob/07c1416e337282ee25936aaa4653698e22967d53/sparse_strips/vello_cpu/src/render.rs#L191) of an owned value just prior to invocation only to [clone it later](https://github.com/linebender/vello/blob/07c1416e337282ee25936aaa4653698e22967d53/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L272).

I suspect this could end up being somewhat costly for glyph rendering if it isn't optimised away.

Are we hoping the compiler takes care of that optimisation for us? Or is this a known limitation that we're aware of but it isn't pragmatic to action at this moment?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
taj-p: I do appreciate the work that has gone into estimating the cost of a render task but I tend to agree with Andrew that this might be difficult to maintain and be very much environment dependent.

I wonder whether targeting path memory instead of estimated runtime cost might tend towards a simpler initial approach and easier maintenance (and then target/investigate estimating dynamic runtime cost separately).

```rs
if  self.batch_path_complexity > PATH_COMPLEXITY_THRESHOLD {
    self.flush_tasks();
}
```

That said, Laurenz has the most experience and intuition about `vello_cpu` and it appears substantial work has gone into deriving this formula, so I certainly defer to your judgement here.

---

I also suggest accumulating complexity within each command method rather than "all at once" in register_task:

```rs
    fn fill_path(&mut self, path: &BezPath, fill_rule: Fill, transform: Affine, paint: Paint) {
        let task_idx = self.bump_task_idx();

        self.batch_path_complexity += compute_complexity(path); // or, perhaps, just, path.elements().len(); 

        self.register_task(RenderTask::FillPath {
            path: path.clone(),
            transform,
            paint,
            fill_rule,
            task_idx,
        });
    }
```

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+        // TODO: Maybe there is a better way of doing this? Something like a `WaitGroup`, but that
+        // can be used in conjunction with `spawn_broadcast`.
+        while thread_counter.load(Ordering::SeqCst) > 0 {}
+
+        self.init();
+    }
+
+    fn flush(&mut self) {
+        self.flush_tasks();
+        core::mem::take(&mut self.task_sender);
```
taj-p: Nit: Should we add a comment here that this "unblocks" our threads allowing them to place their alphas?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+        let result_receiver = self.result_receiver.as_mut().unwrap();
+
+        loop {
+            match result_receiver.try_recv() {
+                Ok(cmd) => match cmd {
+                    CoarseCommand::Render {
+                        strips,
+                        fill_rule,
+                        paint,
+                    } => self.wide.generate(&strips, fill_rule, paint.clone()),
```
taj-p:
```suggestion
                    } => self.wide.generate(&strips, fill_rule, paint),
```

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,627 @@
...
+                    CoarseCommand::PushLayer {
+                        clip_path,
+                        blend_mode,
+                        mask,
+                        opacity,
+                    } => self
+                        .wide
+                        .push_layer(clip_path.clone(), blend_mode, mask.clone(), opacity),
```
taj-p:
```suggestion
                        .push_layer(clip_path, blend_mode, mask, opacity),
```

sparse_strips/vello_common/src/strip.rs
taj-p: > I'm not sure how the store would solve the problem, since we still need to somehow associate each command with the location in the store, no?

Taking a closer look, this should be possible? I suspect we can store the `thread_id` on the `CoarseCommand` as it's returned from the worker. 

This would prevent us needing to keep track of `thread_id` in `StripGenerator` and in `Strip`. I suspect this would have some performance wins. Perhaps we can add a `TODO` at the minimum so we know it's tracked somewhere.

taj-p: > Taking a closer look, this should be possible? I suspect we can store the thread_id on the CoarseCommand as it's returned from the worker.

@LaurenzV / @ajakubowicz-canva (for confirmation) - here's a quick POC https://github.com/LaurenzV/vello/pull/1 that passes `vello_sparse_tests`.  I understand this PR is already well reviewed and being dragged along by my drip fed comments. I'm happy to polish up that PR separately so we can move this along.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
taj-p, line 1: The really nice design choice of this module is that we have two distinct phases. The transition from phase 1 to phase 2 results in a single placing of alphas on each thread. This means we only ever lock the alpha storage for writing _once_ in a given issue commands/flush cycle.

Can I please propose an approach using `OnceLock` that eliminates the need for `Arc`/mutex and should be significantly faster, since, IIUC, after initialisation, a `OnceLock` is just a single memory load operation without thread blocking? Something like the below? Notably, this also removes the hash map, hash lookup costs, and needing to clone the hashmap.

```rs
pub struct OnceLockAlphaStorage {
    slots: Vec<OnceLock<Vec<u8>>>,
}

impl OnceLockAlphaStorage {
    pub fn new(num_threads: u16) -> Self {
        Self {
            slots: (0..num_threads).map(|_| OnceLock::new()).collect(),
        }
    }

    /// Store alpha data for a specific thread (called once per thread)
    pub fn store(&self, thread_id: u8, data: Vec<u8>) -> Result<(), Vec<u8>> {
        self.slots[thread_id as usize].set(data)
    }

    /// Get alpha data for a specific thread
    pub fn get(&self, thread_id: u8) -> Option<&Vec<u8>> {
        self.slots.get(thread_id as usize)?.get()
    }

    // Probably worth adding a take all method and being able to reuse vec allocations somehow
}
```

Does this make sense @ajakubowicz-canva / @LaurenzV ? I'd appreciate your thoughts on whether this is feasible and can appreciably improve performance.

Again, this is an idea that I've[ drafted up here](https://github.com/LaurenzV/vello/pull/2/files) and it passes our tests (like the other idea, I'm happy to polish this up separately to speed along this PR, but I'm also happy for it to be incorporated as part of this PR).

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
taj-p: I suspect this [comment](https://github.com/linebender/vello/pull/1047#discussion_r2146314239) is related to this thread.

taj-p:
The design is so elegant!! It's really impressive. It was a pleasure to read 🙏

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Since this is not in a hot loop or anything, I think it should be safe to use. But will definitely analyze this more carefully in the near future!

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: How about for now I just do a hard-coded limit of like 10 paths per patch for now for simplicity, and refine it in the future? 

I do think there is much value in having some kind of path estimation, I'm aware that runtimes will be different on different targets, but it's also not about getting an accurate runtime estimate, but just a rough approximation so we don't end up sending one path to each thread if we for example if we are drawing a lot of small paths.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Thanks, it had to be that previously but seems like as part of some refactors this became redundant.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Hmm, while it's true that they are kind of dependent on each other, they are from two separate queues, so I feel like just calling them `sender` and `receiver` might make people think that they are part of the same queue?

sparse_strips/vello_cpu/src/dispatch/mod.rs
LaurenzV: Yeah, I also noticed that. I'm not sure how to best deal with that, I might try an approach where each thread has its own `BezPath` buffer and then we copy the path into that instead of always creating a new allocation, though I'm not sure how much faster that will be.

In the worst case, this is simply an overhead that we have to accept for multi-threaded rendering and hope that the gains of multi-threadingn largely outweight this cost. That's another reason why I think it's good that you can still use the `SingleThreadedDispatcher` with multithreading enabled, so you don't have to pay that cost. 

Will add a TODO, though.

LaurenzV:
Thanks for all the comments!! Will be slowly addressing the remaining ones over the weekend, so that it's hopefully ready by Monday.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
ajakubowicz-canva: Yeah good point! Happy with it as it is. :)

sparse_strips/vello_cpu/src/render.rs
LaurenzV: > If this is the case, then I think encoded_paints is currently not being reset and would leak memory?

Yeah, you are right, thanks! This needs to be cleared as well.

sparse_strips/vello_common/src/strip.rs
LaurenzV: I integrated your proposed changes, thanks a lot!

sparse_strips/vello_cpu/src/dispatch/mod.rs
LaurenzV: Added a TODO comment to `MultiThreadedDispatcher`.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: I re-added the test to `StripGenerator`. While this still doesn't really test it inside of `RenderContext`,  properly doing this without adding manual methods just for testing in `Dispatcher` is going to be a bit hard, so I'm not sure it's worth it.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Added the `Barrier` now, thanks for the hint!

sparse_strips/vello_common/src/strip.rs
LaurenzV: It's in the last position for all methods now.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: I adapted the comment and also added an explicit `drop` statement.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: > after initialisation, a OnceLock is just a single memory load operation without thread blocking

I'm not sure whether that's the case, as each access presumably still needs to check whether it's already been initialized? 

In any case, I agree that this is much nicer overall so I will incoroporate this, thanks a lot!

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: I went with Taj's design prposal for now, but will definitely revisit in the near future!

sparse_strips/vello_cpu/src/render.rs
LaurenzV: It probably doesn't matter as much for blurred rectangles since those are much slower to render, but I agree that it would be nice to have a more general method for this, will add a TODO!

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: > I also suggest accumulating complexity within each command method rather than "all at once" in register_task:

Hmm, not sure I understand the advantage of doing so, doesn't this just lead to more clutter in the other methods?

LaurenzV:
I think nearly everything should be addressed now, except for the task runtime estimation thing. Does just setting a fixed size (like 10) sound good to you for now?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
ajakubowicz-canva: I am happy for a TODO and punt cost estimation forward. I am not fussed with what cost estimation you land since it's going to be subject to change.

Edit: There is already a TODO so I'm happy with no change for this PR.

ajakubowicz-canva (approved):
Very very nice. Regarding the question about path estimation I'll defer to @taj-p .

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// Store alpha data for a specific thread (called once per thread)
+    pub(crate) fn store(&self, thread_id: u8, data: Vec<u8>) -> Result<(), Vec<u8>> {
+        self.slots[thread_id as usize].set(data)
+    }
+
+    fn slots(&self) -> &[OnceLock<Vec<u8>>] {
+        &self.slots
+    }
+
+    // Probably worth adding a take all method and being able to reuse vec allocations somehow
```
taj-p:
```suggestion
    // TODO: Probably worth adding a take all method and being able to reuse vec allocations somehow
```

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
taj-p: > Hmm, not sure I understand the advantage of doing so, doesn't this just lead to more clutter in the other methods?

🤦 - sorry I can't read sometimes.

> I do think there is much value in having some kind of path estimation, I'm aware that runtimes will be different on different targets, but it's also not about getting an accurate runtime estimate, but just a rough approximation so we don't end up sending one path to each thread if we for example if we are drawing a lot of small paths.

I think the current approach is fine. Don't get my original suggestion wrong. I think you're in the driver's seat here and we're happy to defer to your best judgement. Happy for this to go in as is if you prefer 😄 .

taj-p (approved):
LGTM! This is an amazing piece of work. Kudos Laurenz!

LaurenzV:
Thanks a lot to both of you!

PR #1048, opened 2025-06-10
Refactor `vello::util`

nicoburns:
- A lot of general cleanup, adding comments, etc.
- Split out a `find_existing_device` method
- Store a copy of `DeviceHandle` inside `RenderSurface`
- Move methods from `RenderContext` to `RenderSurface`
- Allow users to customise `Features` and `Limits` (possibly these should be less global)

(the latter was the initial motivation for me interacting with this code - I need to be able to configure these for the `override_image` demo (both vello and blitz))

Some questions:

- Should `RenderContext` be global? (actually static? suggested to be Arc<Mutex>>?)
- Should `RenderContext` also hold `VelloRenderer`s? Why is `with_winit` reusing `VelloRenderer` instances?
- Should util / `RenderContext` provide utility methods to help with pipeline caching?

### DIFF
--- a/vello/src/util.rs
+++ b/vello/src/util.rs
@@ -4,43 +4,66 @@
 //! Simple helpers for managing wgpu state and surfaces.
 
 use std::future::Future;
-
 use wgpu::{
-    Adapter, Device, Instance, Limits, MemoryHints, Queue, Surface, SurfaceConfiguration,
+    Adapter, Device, Features, Instance, Limits, MemoryHints, Queue, Surface, SurfaceConfiguration,
     SurfaceTarget, Texture, TextureFormat, TextureView, util::TextureBlitter,
 };
 
-use crate::{Error, Result};
+use crate::Error as VelloError;
 
 /// Simple render context that maintains wgpu state for rendering the pipeline.
 pub struct RenderContext {
+    /// A WGPU `Instance`. This only needs to be created once per application.
     pub instance: Instance,
-    pub devices: Vec<DeviceHandle>,
+    /// A pool of already-created devices so that we can avoid recreating devices
+    /// when we already have a suitable one available
+    pub device_pool: Vec<DeviceHandle>,
+
+    // Config
+    extra_features: Option<Features>,
+    override_limits: Option<Limits>,
 }
 
+/// A wgpu `Device`, it's associated `Queue`, and the adapter used to create it.
+/// Q: could we drop the adapter here? wgpu docs say adapters do not need to be kept around...
+#[derive(Clone, Debug)]
 pub struct DeviceHandle {
     adapter: Adapter,
     pub device: Device,
     pub queue: Queue,
 }
 
+impl DeviceHandle {
+    /// Returns the adapter associated with the device.
+    pub fn adapter(&self) -> &Adapter {
+        &self.adapter
+    }
+}
+
+impl Default for RenderContext {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl RenderContext {
-    #[expect(
-        clippy::new_without_default,
-        reason = "Creating a wgpu Instance is something which should only be done rarely"
-    )]
     pub fn new() -> Self {
-        let backends = wgpu::Backends::from_env().unwrap_or_default();
-        let flags = wgpu::InstanceFlags::from_build_config().with_env();
-        let backend_options = wgpu::BackendOptions::from_env_or_default();
-        let instance = Instance::new(&wgpu::InstanceDescriptor {
-            backends,
-            flags,
-            backend_options,
-        });
+        Self::with_features_and_limits(None, None)
+    }
+
+    pub fn with_features_and_limits(
+        extra_features: Option<Features>,
+        override_limits: Option<Limits>,
+    ) -> Self {
         Self {
-            instance,
-            devices: Vec::new(),
+            instance: Instance::new(&wgpu::InstanceDescriptor {
+                backends: wgpu::Backends::from_env().unwrap_or_default(),
+                flags: wgpu::InstanceFlags::from_build_config().with_env(),
+                backend_options: wgpu::BackendOptions::from_env_or_default(),
+            }),
+            device_pool: Vec::new(),
+            extra_features,
+            override_limits,
         }
     }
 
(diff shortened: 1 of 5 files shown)


vello/src/util.rs
```diff
@@ -51,142 +74,89 @@ impl RenderContext {
...
-                },
-                None,
-            )
-            .await
-            .ok()?;
+
+        // Determine features to request
+        // The user may request additional features
+        let extra_features = self.extra_features.unwrap_or(Features::empty());
+        let vello_features = wgpu::Features::CLEAR_TEXTURE | wgpu::Features::PIPELINE_CACHE;
```
DJMcNab, line 124: We probably should pass these in using the new mechanism, now

DJMcNab (commented):
A few thoughts, without having specifically carefully reviewed the code:
1) `vello::util` should never have been our job to expose, and should be removed entirely (maybe inlined as appropriate into `with_winit`, and the other examples should just manage this themselves[^1]). If you want to make a crate for something like this, the best place for that would probably be in the wgpu repository itself. I think we could land this first, if you wanted. However, I don't want to make another release with breaking changes here before removing it, so we need a plan for someone to do the requisite work. That might involve just slapping a `docs(hidden)` on the whole module; I wouldn't like that solution, but it does address our needs here.
2) `vello::util`'s is something which should be pretty independent of the rest of Vello - it's pretty much a generic wgpu utility. The addition of the blitter was explicitly as an intermediate state.
3) We have another copy of this code in `sparse_strips`, which probably should be updated similarly.
4) The updates generally look like a reasonable improvement to the code quality here
5) This needs a changelog entry.

It's clear to me that Vello owning all of this stuff is not fit for purpose, and will require more and more special cased APIs.

[^1]: This is a question about what the role of the "simple" example is; I think it's reasonable for that example to not care about multiple windows. See the discussion in https://github.com/linebender/vello/pull/227, where it seems like this multi-device support was never actually necessary, at least for single `Surface` cases. For headless and the tests, there is no surface, so it definitely doesn't need all this complexity.

nicoburns:
@DJMcNab It seems that this infrastructure is mostly for caching/reusing resources? Otherwise we could just recreate everything on every "resume"? Do you have a read on which operations are expensive?

- Creating a `wgpu::Instance` (wgpu docs say this doesn't necessarily need to be persisted)
- Creating a `wgpu::Adapter` (wgpu docs say this doesn't necessarily need to be persisted)
- Creating a `wgpu::Device` (should be relatively cheap?)
- Creating a `vello::Renderer` (I'm guessing is where the cost is and why we're caching devices?)
- How does pipeline caching interact with all of this?

---

> maybe inlined as appropriate into with_winit, and the other examples should just manage this themselves[1](https://github.com/linebender/vello/pull/1048#user-content-fn-1-6e737c5c6259601aa0ca8a5753260f7b)

I would really like to make `with_winit` simpler if at all possible. It feels very spaghetti and under-abstracted to me. Although this could be an abstraction that lives within the example if need be.

---

> vello::util's is something which should be pretty independent of the rest of Vello - it's pretty much a generic wgpu utility.

I think Vello also has quite specific requirements around things like features, limits, texture format? I guess we might be able to pass those in as parameters.

---

> We have another copy of this code in sparse_strips, which probably should be updated similarly.

It would be nice to unify these into one if possible. Although I guess the intermediate texture isn't great in that regard.

DJMcNab:
> It seems that this infrastructure is mostly for caching/reusing resources? Otherwise we could just recreate everything on every "resume"? Do you have a read on which operations are expensive?

I don't think that's the case? It just some utilities around creating wgpu utilities, because there's a lot of boilerplate there. In particular, this is boilerplate around supporting multiple devices, which is only "necessary" if you're going to have multiple surfaces (an unoptimal surface will keep chugging along with bad performance, which is fine - we don't even both recreating a more optimal surface)

I don't see how any part of my suggestion suggests to not maintain these resources, just not have them be managed by code in Vello - that's the wrong abstraction level.

> I would really like to make `with_winit` simpler if at all possible. It feels very spaghetti and under-abstracted to me. Although this could be an abstraction that lives within the example if need be.

Yeah, this would live inside of the `with_winit` example in that case.

> I think Vello also has quite specific requirements around things like features, limits, texture format? I guess we might be able to pass those in as parameters.

Ish? We basically just require baseline wgpu, but can make use of the clear texture ability. We really don't want to be in charge of creating your device.
 
> It would be nice to unify these into one if possible. Although I guess the intermediate texture isn't great in that regard.

Writing this out yourself in each app is basically the idiomatic way to write wgpu code. My suggestion for where a utility for this should live is the wgpu repository - the maintainers there are better placed to determine how to write this type of code idiomatically. See for example the examples for vello_hybrid: https://github.com/linebender/vello/blob/36c6659b7a03f8b6afe51c8cc770948f32e9d9da/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs#L38-L71

DJMcNab:
See also in #879, where we recreate the device etc in resume if and only if the old device was not compatible:
https://github.com/linebender/vello/pull/879/files#diff-566ea9ba31e45a5ffe1a87c559ccd0e12eb967c5160d4ce65ef0651e3384999cR303-R314

nicoburns:
> I don't think that's the case? It just some utilities around creating wgpu utilities, because there's a lot of boilerplate there. In particular, this is boilerplate around supporting multiple devices, which is only "necessary" if you're going to have multiple surfaces 

If it's just boilerplate, then I don't understand why we persist the devices and the instance (i.e. why the code is stateful). It seems like you could just have a function that goes through the instance -> adapter -> device -> surface dance from scratch every time which would be much simpler (never bother with `find_existing_device` just always `create_new_device`).

My assumption from the structure of the code was that there was some reason why devices must be reused, and I assumed performance (I'm still confused).

DJMcNab:
I really don't follow what you're asking; your suggests seem completely unrelated to what I'm suggesting.

But as I say, if you want to just move on from this, we can land this PR, with adding `doc(hidden)` to the util module (so it's clear that this isn't to be treated as a public API) and have removing this utility as follow-up.

eliheuer:
On the "allow users to customize Features and Limits" subject:

A full-screen WebGPU + Vello application on a retina display needs more than the 8192 `max_texture_dimension_2d` that `RenderContext::new_device` hardcodes via `Limits::default()`.

With no public override, the only fix I could think of was to abandon `RenderContext` and replicate its `wgpu` setup so I could pass the actual max I need.

nicoburns:
I ended up creating the `wgpu_context` crate which can be used instead of vello::util. It's far from perfect, but it does allow for Features/Limits customisation and has a few other improvements.

PR #1049, opened 2025-06-11
First sketch of porting strip rendering to SIMD

LaurenzV:
Here are the changes I needed to make: https://github.com/raphlinus/fearless_simd/commit/7a999662e99bce6d7b16d2d128fe4bfc3a8eb141

And here are the benchmark results before vs. after:

```
render_strips/Ghostscript_Tiger
                        time:   [181.39 µs 182.40 µs 183.90 µs]
                        change: [-18.710% -18.391% -18.040%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (2.00%) high mild
  1 (2.00%) high severe
render_strips/paris_30k time:   [17.009 ms 17.160 ms 17.321 ms]
                        change: [-17.442% -16.607% -15.729%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (4.00%) high mild
render_strips/coat_of_arms
                        time:   [1.5429 ms 1.5453 ms 1.5481 ms]
                        change: [-15.918% -15.404% -14.926%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (4.00%) high mild
  1 (2.00%) high severe
```

This might not look like a lot, but the reason for this is that the previous implementation already auto-vectorized on ARM, so I would say this is still a very significant speed boost! Tests also pass, with some minor differences due to our usage of fused multiply-add. But it shows that `fearless_simd` works at least. :)

Currently this works with explicit f32x4, but it's also adaptable fo f32x8 and f32x16. In the end, we will probably want to choose f32x8 for now.

Just a draft PR because it's not intended to be merged (we will need to implement fallback in `fearless_simd` first), but only for showcase.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -800,6 +800,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "fearless_simd"
+version = "0.2.0"
+
+[[package]]
 name = "file-id"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3200,6 +3204,7 @@ name = "vello_common"
 version = "0.0.1"
 dependencies = [
  "bytemuck",
+ "fearless_simd",
  "libm",
  "peniko",
  "png",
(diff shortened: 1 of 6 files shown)


LaurenzV:
Will open another PR once it's done.

PR #1050, opened 2025-06-12
[vello_dev_macros] enable vello_hybrid `clip` and `opacity` tests

ajakubowicz-canva:
This is a small test-only housekeeping PR enabling more tests for vello_hybrid.

Vello hybrid has landed clipping and opacity seems to pass tests. I also checked the other ignored tests but they still have failures.


taj-p (approved):
LGTM!!

PR #1051, opened 2025-06-12
[vello_hybrid] update `set_paint` to conform with `vello_cpu`'s set_paint API

ajakubowicz-canva:
### Context

I am currently working on gradients, and Alex is working on Images. In both cases, being passed a `Paint` seems insufficient for expressing Images and Gradients.

While drafting gradients I noticed we had duplicated this API change, so I'm submitting it standalone. See the same change in the image draft PR: https://github.com/linebender/vello/pull/935/files#diff-cf782a179aa0fd954852f73458fb40f9a2e608c3b20eec0265acfc3e4dc284ccR201-R203

### Why

In order to enable image and gradient in vello_hybrid, I believe an API change is required.

A nice benefit of this change is that our API more closely resembles `vello_cpu`.

### Test plan

CI and existing unit tests.

### DIFF
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -10,7 +10,7 @@ use vello_common::flatten::Line;
 use vello_common::glyph::{GlyphRenderer, GlyphRunBuilder, GlyphType, PreparedGlyph};
 use vello_common::kurbo::{Affine, BezPath, Cap, Join, Rect, Shape, Stroke};
 use vello_common::mask::Mask;
-use vello_common::paint::Paint;
+use vello_common::paint::{Paint, PaintType};
 use vello_common::peniko::Font;
 use vello_common::peniko::color::palette::css::BLACK;
 use vello_common::peniko::{BlendMode, Compose, Fill, Mix};
@@ -25,7 +25,7 @@ pub(crate) const DEFAULT_TOLERANCE: f64 = 0.1;
 /// the current transform.
 #[derive(Debug)]
 struct RenderState {
-    pub(crate) paint: Paint,
+    pub(crate) paint: PaintType,
     pub(crate) stroke: Stroke,
     pub(crate) transform: Affine,
     pub(crate) fill_rule: Fill,
@@ -45,7 +45,7 @@ pub struct Scene {
     pub(crate) line_buf: Vec<Line>,
     pub(crate) tiles: Tiles,
     pub(crate) strip_buf: Vec<Strip>,
-    pub(crate) paint: Paint,
+    pub(crate) paint: PaintType,
     paint_visible: bool,
     pub(crate) stroke: Stroke,
     pub(crate) transform: Affine,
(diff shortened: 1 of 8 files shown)


sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -96,13 +96,26 @@ impl Scene {
...
+            PaintType::Solid(s) => s.into(),
+            PaintType::Gradient(_) => {
+                unimplemented!("gradient not implemented")
+            }
+            PaintType::Image(_) => {
+                unimplemented!("images not implemented")
+            }
+        }
+    }
+
```
ajakubowicz-canva, line 110: This stub mimics the implementation in `vello_cpu` and is required once we start encoding images and gradients.

vello_cpu code ref: https://github.com/linebender/vello/blob/main/sparse_strips/vello_cpu/src/render.rs#L93-L108

This code was also added in the Image draft PR: https://github.com/linebender/vello/pull/935/files#diff-cf782a179aa0fd954852f73458fb40f9a2e608c3b20eec0265acfc3e4dc284ccR280-R292

LaurenzV (approved):

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -187,12 +201,10 @@ impl Scene {
     }
 
     /// Set the paint for subsequent rendering operations.
-    pub fn set_paint(&mut self, paint: Paint) {
-        self.paint_visible = match &paint {
-            Paint::Solid(color) => !color.is_transparent(),
-            Paint::Indexed(_) => true,
-        };
-        self.paint = paint;
+    pub fn set_paint(&mut self, paint: impl Into<PaintType>) {
```
DJMcNab, line 207: I'm worried by this signature - supporting images from a pixmap is explicitly out of scope - instead they need to be from a wgpu texture (possibly a sub-area of such a texture?). And that texture would normally be expected to be "late-bound"

As an intermediate solution to unblock work this is probably fine, and I suspect this API is *fine* for gradients. But it might be worth having a comment that this is not the permant API for Images here.

ajakubowicz-canva: Just for my own understanding, is your comment also applicable to `vello_cpu`'s `set_paint` API? I'm just trying to understand if the difference from CPU vs GPU necessitates the signature difference.

Is the following suggestion ok? 
```suggestion
    // TODO: This API is not final. Supporting images from a pixmap is out of scope. Instead images
    //              should be passed via a wgpu texture (or region of a texture), and be "late-bound".
    pub fn set_paint(&mut self, paint: impl Into<PaintType>) {
```

Final question, that may change the comment above. `Scene` may be rendered by a different backend (e.g. `WebGLRenderer`), I assume a `wgpu texture` might need to be abstracted to something that's either agnostic of the backend, or aware of the backend. You can probably tell I have very naive thoughts about the image feature.

Also, thank you so much for commenting!

DJMcNab: Yeah, that's a reasonable comment. Effectively, we want to use a backend-agnostic id for images, and then when we're at render time, those can be hydrated with the real value. So it's kind of both... it's something which is unaware the backend, but code which produces the "something" needs to be aware of the backend.

Using a pixmap in Vello CPU isn't unreasonable there, as there's only one reasonable type, but will likely also change for consistency (and changing it could also improve parallelism in some niche cases/improve handling of image filters?)

It's not worth spending too much energy on this now, so long as there is a todo comment

ajakubowicz-canva: I appreciate you knowledge sharing. Thank you! Will add a comment then merge (once I get back to computer). :D

ajakubowicz-canva: Comment added in: https://github.com/linebender/vello/pull/1051/commits/e19a35275d905e1c33f84270a43fdb490ed8936a

PR #1052, opened 2025-06-14
Add a more ergonomic API for texture overrides

### LINKED ISSUE #664: Better API for already GPU-resident images
There have been repeated requests for supporting using a wgpu `Texture` within a Vello `Scene`, such as #660.

We did implement a targeted fix for this in #636.

In [#gpu > vello adding wgpu texture buffers to scene](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/vello.20adding.20wgpu.20texture.20buffers.20to.20scene), Chad suggested a new variant of [Brush](https://docs.rs/peniko/latest/peniko/enum.Brush.html) for this use case.
I think that is likely to be the best API, due to the other constraints on the `Scene` API (namely that we have chosen to make it independent of wgpu, for any consumers who use Vello without using wgpu directly)

nicoburns:
- Fixes #664 
- This replaces the `override_image` API in Vello Classic
- It follows the same overall strategy as the `override_image` but wraps it in a more ergonomic public API that handles the creation of dummy images for users.
- It marks `override_image` as deprecated rather than removing it, with the idea being that this could be released in a patch update and we'd then remove it in the next breaking release.
- It would be nice if a future refinement could avoid the need to allocate an `Arc` for the `Blob`. I think this would require giving texture's their own dedicate slot in the coarse encoding.

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -312,6 +312,15 @@ pub enum Error {
 )]
 pub(crate) type Result<T, E = Error> = std::result::Result<T, E>;
 
+/// An opaque handle to a texture registered with [`Renderer::register_texture`] that
+/// can be passed to [`Scene::draw_texture`] to draw the texture.
+#[derive(Copy, Clone, PartialEq, Hash)]
+pub struct TextureHandle {
+    pub(crate) id: u64,
+    pub(crate) width: u32,
+    pub(crate) height: u32,
+}
+
 /// Renders a scene into a texture or surface.
 ///
 /// Currently, each renderer only supports a single surface format, if it
@@ -520,6 +529,7 @@ impl Renderer {
     /// dimensions would be rendered.
     ///
     /// [data]: peniko::Image::data
+    #[deprecated = "Use register_texture and unregister_texture methods instead"]
     pub fn override_image(
         &mut self,
         image: &peniko::Image,
@@ -529,6 +539,40 @@ impl Renderer {
             Some(texture) => self.engine.image_overrides.insert(image.data.id(), texture),
             None => self.engine.image_overrides.remove(&image.data.id()),
         }
+    }
+
+    /// Register a [`wgpu::Texture`] with Vello. You will receive a [`TextureHandle`] which
+    /// can be used to draw the registered texture using [`Scene::draw_texture`]
+    ///
+    /// If the texture is no longer active then it should be unregistered using [`unregister_texture`](Self::unregister_texture)
+    pub fn register_texture(&mut self, texture: wgpu::Texture) -> TextureHandle {
+        // Generate a unique ID using peniko::Blob to guarantee it won't clash
+        // with a user-generated image Blob
+        let id = peniko::Blob::<()>::new(std::sync::Arc::new(&[])).id();
+
+        let handle = TextureHandle {
+            id,
+            width: texture.width(),
+            height: texture.height(),
+        };
+
+        // Create a texture base for the texture
+        let texture_base = wgpu::TexelCopyTextureInfoBase {
+            texture,
+            mip_level: 0,
+            origin: wgpu::Origin3d::ZERO,
+            aspect: wgpu::TextureAspect::All,
+        };
+
+        // Insert it into the overrides map
+        self.engine.image_overrides.insert(id, texture_base);
+
+        handle
+    }
+
+    /// Unregister a [`wgpu::Texture`] that was registered with [`register_texture`](Self::register_texture)
+    pub fn unregister_texture(&mut self, handle: TextureHandle) {
+        self.engine.image_overrides.remove(&handle.id);
     }
 
     /// Reload the shaders. This should only be used during `vello` development
--- a/vello/src/scene.rs
+++ b/vello/src/scene.rs
@@ -23,6 +23,8 @@ use skrifa::{
 use vello_encoding::BumpAllocatorMemory;
 use vello_encoding::{Encoding, Glyph, GlyphRun, NormalizedCoord, Patch, Transform};
 
+use crate::TextureHandle;
+
 // TODO - Document invariants and edge cases (#470)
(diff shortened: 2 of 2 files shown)


vello/src/scene.rs
```diff
@@ -306,6 +308,58 @@ impl Scene {
...
+        transform: Affine,
+        alpha: f32,
+        handle: TextureHandle,
+        brush_transform: Option<Affine>,
+        shape: &impl Shape,
+        x_extend: Extend,
+        y_extend: Extend,
+    ) {
+        let dummy_image = Image {
+            data: Blob::from_raw_parts(Arc::new([]), handle.id),
```
DJMcNab, line 344: Why can't `TextureHandle` just be a thin wrapper around `Image`?
Having to allocate here is a bit weird.

See my other comments for where this naturally leads

vello/src/lib.rs
```diff
@@ -531,6 +541,40 @@ impl Renderer {
...
+            height: texture.height(),
+        };
+
+        // Create a texture base for the texture
+        let texture_base = wgpu::TexelCopyTextureInfoBase {
+            texture,
+            mip_level: 0,
+            origin: wgpu::Origin3d::ZERO,
+            aspect: wgpu::TextureAspect::All,
+        };
```
DJMcNab, line 565: Hmm, so because we infer the size, we can't configure the x and y of the Origin3d (but it would be meaningful to configure the `z`).

This is the main reason to keep `override_texture`. That being said, I don't think we properly make the guarantees about this working...

vello/src/lib.rs
```diff
@@ -531,6 +541,40 @@ impl Renderer {
         }
     }
 
+    /// Register a [`wgpu::Texture`] with Vello. You will receive a [`TextureHandle`] which
+    /// can be used to draw the registered texture using [`Scene::draw_texture`]
+    ///
+    /// If the texture is no longer active then it should be unregistered using [`unregister_texture`](Self::unregister_texture)
+    pub fn register_texture(&mut self, texture: wgpu::Texture) -> TextureHandle {
```
DJMcNab, line 548: This API is a good idea; the other API is hard to use. Here's a thought though, why not:
```suggestion
    pub fn register_texture(&mut self, texture: wgpu::Texture) -> Image {
```
This seems to me to have a number of advantages:
1) It removes the `Arc` allocation each time you want to use the texture.
2) It means we don't need a new API on the `Scene` side
3) It allows customising the image quality, alpha, and extend modes in a much less noisy way, as well as allowing the quality to be updated.

vello/src/lib.rs
```diff
@@ -520,6 +529,7 @@ impl Renderer {
     /// dimensions would be rendered.
     ///
     /// [data]: peniko::Image::data
+    #[deprecated = "Use register_texture and unregister_texture methods instead"]
```
DJMcNab, line 532: Even with this new API, `override_image` is still useful:

- It allows sampling from an atlas (i.e. making an image which only copies from a "subset" of the base texture). I think it's correct for the `register_texture` API to not accept the mip levels, etc.
- It allows using the same handle between `Renderers` (with different underlying textures for each Device). Admittedly, this is very niche.

vello/src/scene.rs
nicoburns: Hmm... I think my take on this is that `peniko::Image` is a bit of a weird group of properties, even for regular images. If we look at `peniko::Image`:

```rust
pub struct Image {
    pub data: Blob,
    pub format: ImageFormat,
    pub width: u32,
    pub height: u32,
    pub x_extend: Extend,
    pub y_extend: Extend,
    pub quality: ImageQuality,
    pub alpha: f32,
}
```

then I think I'd argue that only `data`, `format`, `width`, and `height` are inherent properties of an image, and that `x_extend`, `y_extend`, `quality`, and `alpha` are things that you might reasonably want to change on a per-render basis (perhaps even rendering the same underlying image resource multiple times in the same scene with different alpha / extend values).

I think I'd like to see something analogous to the texture/texture_view split where this is split into two structs (with one containing the other and/or a reference to the other).

vello/src/lib.rs
nicoburns: How you would feel about replacing `override_image` with a variant of `register_texture` (say `register_texture_info`) that takes a `TexelCopyTextureInfoBase` instead of a `Texture`? The difference would be that users wouldn't have to generate their own `Image`. Also just a better name, as "override_image" is a bit of an implementation detail and probably not a name that users will look for.

DJMcNab: That seems pretty reasonable. Maybe `register_texture_subset` would be a better name?

vello/src/scene.rs
DJMcNab: All of that is true, but it's not clear to me if you're suggesting a difference course for this PR, and if so what that is.

Ultimately, Vello Classic is in maintenance mode whilst the sparse strips work continues, so the less confusion there is in our APIs, the better. That split of per-image vs per-render properties is a real one which is causing pain elsewhere, but I don't think it can meaningfully block this PR.

vello/src/lib.rs
nicoburns: >  Maybe `register_texture_subset` would be a better name?

Sure. I don't fully understand what this is for, so I'm not in a good position to suggest a name.

vello/src/lib.rs
nicoburns: If we can land https://github.com/linebender/peniko/pull/117 then that would help with a lot of the ergonomic concerns here.

I still think a dedicated method (and perhaps making `TextureHandle` a newtype wrapper around `Image`) would make sense. The idea is that the fact that you need to pass a fake image with an ID that represents a texture in order to render a texture is an implementation detail, and that the API would be much clearer/cleaner if we hide that implementation detail and make `TextureHandle` opaque.

PR #1054, opened 2025-06-20
[vello_cpu] reset transforms when resetting the `RenderContext`

ajakubowicz-canva:
Small change, but I think it's intended that all state resets when calling `.reset` on the RenderContext.


ajakubowicz-canva:
@LaurenzV Attached a screenshot of what I see:

Screenshot 2025-06-20 at 4 59 34 pm

Thank you!

PR #1059, opened 2025-06-21
Attempt to port strip rendering to f32x8

LaurenzV:
This is a bit awkward, because the original implementation has an accumulator that need to operate on a column-basis, so we need to do some trickery to get it to work for f32x8. Unfortunately, that does seem to have a non-negligible impact on the benchmarks on NEON, especially for `paris_30k`:

```
                        time:   [195.67 µs 196.05 µs 196.50 µs]
                        change: [+0.6444% +1.5014% +2.3170%] (p = 0.00 < 0.05)
  2 (4.00%) high mild
  1 (2.00%) high severe
render_strips/paris_30k_simd
                        time:   [20.271 ms 20.574 ms 20.908 ms]
                        change: [+5.3170% +7.1051% +8.9524%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (4.00%) high mild
render_strips/coat_of_arms_simd
                        time:   [1.6535 ms 1.6550 ms 1.6567 ms]
                        change: [+1.1521% +2.0065% +2.6064%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (4.00%) high mild
  2 (4.00%) high severe
```

Though obviously the most interesting thing will be to try this implementation with fp16. If we see a considerable speed boost here, then it probably is still worth merging (perhaps we have to benchmark WASM as well, though). If anyone has a more elegant solution to the accumulator problem, I'm all ears. 😄 

Marking as draft because I'm not sure whether this should be merged.

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -65,6 +65,8 @@ fn render_impl<S: Simd>(
         return;
     }
 
+    const WIDTH: usize = Tile::WIDTH as usize / 2;
+
     // The accumulated tile winding delta. A line that crosses the top edge of a tile
     // increments the delta if the line is directed upwards, and decrements it if goes
     // downwards. Horizontal lines leave it unchanged.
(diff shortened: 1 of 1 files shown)


ajakubowicz-canva:
Our of curiosity do you have an intuition for why the time has regressed? Could it be the cost of splitting and combining?

LaurenzV:
No idea, I suppose that the way its written just confuses the compiler a bit and makes it generate worse code, but hard to say.

LaurenzV:
I think a better approach will be needed for larger vectors, so I'll close this for now.

PR #1060, opened 2025-06-22
Attempt to make `text` feature optional

LaurenzV:
Unfortunately I'm stuck because just enabling the "std" feature will also enable the "skrifa/std" feature, and because of this skrifa will be pulled in as a dependency even though the text feature is disabled. I'm not sure how to best solve this without adding two `skrifa-std`/`skrifa-libm` features that need to be enabled manually by the user. Suggestions are welcome.

The main motivation is that it would allow using the renderer in contexts where text rendering is already implemented with a different method, without having to pull in skrifa as a dependency (e.g. resvg).

### DIFF
--- a/sparse_strips/vello_common/Cargo.toml
+++ b/sparse_strips/vello_common/Cargo.toml
@@ -22,12 +22,12 @@ peniko = { workspace = true, features = ["bytemuck"] }
 fearless_simd = { workspace = true }
 png = { workspace = true, optional = true }
 roxmltree = { version = "0.20.0", optional = true }
-skrifa = { workspace = true }
+skrifa = { workspace = true, optional = true }
 smallvec = { workspace = true }
 libm = { version = "0.2.15", optional = true }
 
 [features]
-default = ["std", "png"]
+default = ["std", "png", "text"]
 # Enable using SIMD instructions for rendering
 simd = []
 # Get floating point functions from the standard library (likely using your target’s libc).
@@ -38,6 +38,8 @@ libm = ["peniko/libm", "skrifa/libm", "dep:libm", "fearless_simd/libm"]
 png = ["std", "dep:png"]
 # Enable multi-threaded rendering.
 multithreading = []
+# Add support for text rendering
+text = ["skrifa"]
 
 # Development only features
 
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -15,17 +15,17 @@ use alloc::vec;
 use alloc::vec::Vec;
 use vello_common::blurred_rounded_rect::BlurredRoundedRectangle;
 use vello_common::color::{AlphaColor, Srgb};
+#[cfg(feature = "text")]
 use vello_common::colr::{ColrPainter, ColrRenderer};
 use vello_common::encode::{EncodeExt, EncodedPaint};
 use vello_common::fearless_simd::Level;
+#[cfg(feature = "text")]
 use vello_common::glyph::{GlyphRenderer, GlyphRunBuilder, GlyphType, PreparedGlyph};
 use vello_common::kurbo::{Affine, BezPath, Cap, Join, Rect, Shape, Stroke};
 use vello_common::mask::Mask;
-use vello_common::paint::{Image, Paint, PaintType};
-use vello_common::peniko;
+use vello_common::paint::{Paint, PaintType};
 use vello_common::peniko::color::palette::css::BLACK;
-use vello_common::peniko::{BlendMode, Compose, Fill, Gradient, Mix};
-use vello_common::peniko::{Font, ImageQuality};
+use vello_common::peniko::{BlendMode, Compose, Fill, Mix};
 use vello_common::pixmap::Pixmap;
 
 pub(crate) const DEFAULT_TOLERANCE: f64 = 0.1;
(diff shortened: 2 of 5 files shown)


nicoburns:
I think there is a `skrifa?/std` syntax for this  case. Which will enable the std feature if skrifa is enabled.

LaurenzV:
Perfect, thanks! :)

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -41,6 +43,7 @@ pub struct RenderContext {
     pub(crate) fill_rule: Fill,
     pub(crate) temp_path: BezPath,
     pub(crate) encoded_paints: Vec<EncodedPaint>,
+    #[allow(dead_code, reason = "used when the `text` feature is enabled")]
```
nicoburns:
```suggestion
    #[cfg(feature = "text")]
```

nicoburns: Or alternatively:

nicoburns:
```suggestion
    #[cfg_attr(not(feature = "text")), allow(dead_code))]
```

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -10,22 +10,24 @@ use crate::dispatch::multi_threaded::MultiThreadedDispatcher;
...
 use alloc::vec;
 use alloc::vec::Vec;
 use vello_common::blurred_rounded_rect::BlurredRoundedRectangle;
+#[cfg(feature = "text")]
 use vello_common::color::{AlphaColor, Srgb};
+#[cfg(feature = "text")]
 use vello_common::colr::{ColrPainter, ColrRenderer};
 use vello_common::encode::{EncodeExt, EncodedPaint};
 use vello_common::fearless_simd::Level;
+#[cfg(feature = "text")]
```
nicoburns: Would be nice to combine the optional imports from the same crate into one. Also, I would pull the `#[cfg(feature = "text")]` imports into their own block below the others (separated by a newline so the formatter doesn't mix them).

nicoburns (approved):
Some nits in comments. But nothing major (or that I would consider blocking).

sparse_strips/vello_common/Cargo.toml
```diff
@@ -22,22 +22,24 @@ peniko = { workspace = true, features = ["bytemuck"] }
...
+std = ["peniko/std", "skrifa?/std", "fearless_simd/std"]
 # Use floating point implementations from libm.
-libm = ["peniko/libm", "skrifa/libm", "dep:libm", "fearless_simd/libm"]
+libm = ["peniko/libm", "skrifa?/libm", "dep:libm", "fearless_simd/libm"]
 # Allow loading Pixmap from PNG, and drawing png glyphs.
 png = ["std", "dep:png"]
 # Enable multi-threaded rendering.
 multithreading = []
+# Add support for text rendering
+text = ["dep:skrifa"]
```
DJMcNab, line 42: The proliferation of different feature variants is a bit of a concern, both for testing and in terms of making contributions more difficult. If `png` were `no_std`, I could see a reasonable argument for making that mandatory if text were enabled.

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -17,7 +17,7 @@ workspace = true
 [dependencies]
 bytemuck = { workspace = true, features = ["derive"] }
 thiserror = { workspace = true }
-vello_common = { workspace = true, features = ["std"] }
+vello_common = { workspace = true, features = ["std", "text"] }
```
DJMcNab, line 20: Hmm, this is a bit of a surprise

LaurenzV: Surprise as in? We can also make it an optional feature there, but I thought the motivation to do so is much stronger for vello_cpu, because even without text vello_hybrid is a pretty chunky dependency already.

DJMcNab: This whole thing is something I'd expect us to have spent some time on it in office hours, where we could make this decision.

I also don't know how this will interact with Vello API, when we have that; because the different renderers would have to enable/disable their own glyph features, which isn't possible in general to coordinate.

LaurenzV: Oh okay, sorry, I didn't think it would be a big deal, I can revert it if you want. But I think it would be useful to have, especially if we want to use this renderer for example in `resvg` in the future, being able to remove the dependency on skrifa might be useful.

DJMcNab: I don't think we should revert this (at least not without discussion). But I do think that this will require significant thinking work to behave nicely with Vello API.

nicoburns: For Vello API I'd expect:
- That the API (trait) itself isn't feature flagged
- The feature flag would be in the backend crate
- The implementation would become a noop in the case that the feature is disabled

Not ideal that it becomes a runtime error, but I think some level of compromise on this kind of thing is inevitable when abstracting over backends with different capabilities.

sparse_strips/vello_common/Cargo.toml
DJMcNab, line 42: Is this new feature intentionally not documented in lib.rs/the README, or is it just an oversight?

LaurenzV: Just an oversight

Issue #1061, opened 2025-06-22
Layer shapes are wrong


sagudev:
As reported in [#vello > Servo 2D canvas backend @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Servo.202D.20canvas.20backend/near/525153593) vello currently chooses to big size (and they are also off) for layers, I suspect something is wrong with binning.

For this code:
```rust
scene.fill(
        vello::peniko::Fill::NonZero,
        Affine::IDENTITY,
        vello::peniko::color::AlphaColor::from_rgb8(0, 255, 0),
        None,
        &Rect::from_origin_size((0.0, 0.0), (60., 60.)),
    );
    scene.push_layer(
        vello::peniko::Compose::Clear,
        1.0,
        vello::kurbo::Affine::IDENTITY,
        &Rect::from_origin_size((20.0, 20.0), (20., 20.)),
    );
    scene.pop_layer();
    scene.fill(
        vello::peniko::Fill::NonZero,
        Affine::IDENTITY,
        vello::peniko::color::AlphaColor::from_rgb8(255, 0, 0),
        None,
        &Rect::from_origin_size((20.0, 20.0), (20., 20.)),
    );
```
one would expect to get 

![Image](https://github.com/user-attachments/assets/f1ad8501-510a-4a5a-868b-cd925814283e)

but what we actually get is 

![Image](https://github.com/user-attachments/assets/e2aa2d98-43a9-4889-8331-b49d990bc4f6)

sagudev:
>Presumably the final erroneous areas line up with the 16x16 tile boundary. I've not checked that, but it shouldn't be surprising.
>
>Fixing this is pretty hard; the data about clip regions/layer boundaries is, I think, already discarded by the time that we're in fine >rasterisation (and is all folded into the paths to be filled).

_Originally posted by @DJMcNab in https://github.com/linebender/vello/pull/1062#pullrequestreview-2949863109_

And I checked and erroneous areas do line up with the 16x16 tile boundary.

PR #1062, opened 2025-06-22
Add smoke test for layer size

sagudev:
This is for https://github.com/linebender/vello/issues/1061


vello_tests/tests/known_issues.rs
```diff
@@ -78,6 +78,38 @@ fn many_bins_cpu() {
...
+        &Rect::from_origin_size((20.0, 20.0), (20., 20.)),
+    );
+    scene.pop_layer();
+    scene.fill(
+        vello::peniko::Fill::NonZero,
+        Affine::IDENTITY,
+        vello::peniko::color::AlphaColor::from_rgb8(255, 0, 0),
+        None,
+        &Rect::from_origin_size((20.0, 20.0), (20., 20.)),
+    );
```
DJMcNab, line 99: Can we maybe make this a hollow shape? That would actually then be testing that the clear compose is actually doing anything. But possibly disregard this if I'm misunderstanding what the clear compose does

DJMcNab (commented):
Presumably the final erroneous areas line up with the 16x16 tile boundary. I've not checked that, but it shouldn't be surprising.

Fixing this is pretty hard; the data about clip regions/layer boundaries is, I think, already discarded by the time that we're in fine rasterisation (and is all folded into the paths to be filled).

vello_tests/tests/known_issues.rs
sagudev: I seen many WPT test that did exactly like I did and it's really painful to dbg, so in https://github.com/linebender/vello/pull/1062/commits/739e6a18619ffe631d67b3af9ca1ab4efcff7e8d I draw red on top of green, and then expect layer to clear whole red area (we are currently clearing too much).

sagudev:
> Presumably the final erroneous areas line up with the 16x16 tile boundary. I've not checked that, but it shouldn't be surprising.

I checked and it is.

> Fixing this is pretty hard

I got the same impression after throwing a glance at shaders, this is definitely issue for someone more experienced then me.

vello_tests/tests/known_issues.rs
```diff
@@ -78,6 +78,38 @@ fn many_bins_cpu() {
     many_bins(true);
 }
 
+/// Test for <https://github.com/linebender/vello/issues/1061>
+#[test]
+#[should_panic]
+fn test_layer_size() {
```
DJMcNab, line 84: Maybe?

```suggestion
fn compose_clear() {
```

sagudev:
I discovered this issue with `Compose::Copy`, but `Compose::DestIn` also does not work correctly.

sagudev:
gentle ping @DJMcNab

DJMcNab:
Hmm, I was under the impression that this PR is blocked on you - you should either incorporate or push back against the suggestion to rename the test. Either one is fine, but the PR is waiting on your decision there.

vello_tests/tests/known_issues.rs
sagudev: Given that layout size is not respected by many modes, I think we should keep name as is.

PR #1064, opened 2025-06-24
[vello_hybrid] Image rendering

grebmeg:
### Image Rendering in Sparse Strips

Adds image rendering support to `vello_hybrid` with texture atlas management, multiple quality levels, and extend modes.

https://github.com/user-attachments/assets/25298ebe-70d6-4e7d-a347-58595e6cbfd0

**Image Features:**
- Quality Levels: Low (nearest), Medium (bilinear), High (bicubic) filtering
- Extend Modes: Pad, Repeat, Reflect for image boundaries
- Transform Support: Full affine transformations (rotation, scaling, skewing)

**Key Components:**
- Image Cache (`ImageCache`): texture atlas management with automatic slot reuse using `guillotiere` packing
- GPU Encoding (`GpuEncodedImage`): stores transform, sampling, and quality data for GPU shaders
- Paint System: treats images as indexed paints
- Shader Pipeline: WGSL shaders handle sampling with extend modes (pad/repeat/reflect)

To render an image, the corresponding texture must first be registered in `ImageCache`. During this process, the image is "packed" into an atlas and assigned an `ImageId` along with its atlas offset coordinates. This `ImageId` serves as an opaque reference and is later used as the image source during scene rendering.

Images are represented as `EncodedPaint::Image`, which include transformation and quality parameters. The CPU precomputes sampling advances (`x_advance` and `y_advance`) for efficient rendering. This data is used to generate `GpuEncodedImage`s, which store the transform matrix and translation, image dimensions, atlas offset, quality settings, and extend modes (such as pad, repeat, or reflect).

The path is divided into `GpuStrip`s. Each strip now includes a paint type (0 for solid color, 1 for alpha, and 2 for image) and an associated paint texture Id. This texture Id is used to locate the corresponding `GpuEncodedImage` data within the `encoded_paints_texture` in the shader.

The vertex shader is responsible for computing texture coordinates and applying the necessary transformations. The fragment shader then samples from the atlas texture, taking into account the specified quality settings.

**Notes:**
- WebGL is currently not supported. I’ll continue working on it once this PR is reviewed.
- Four tests are failing: three involve bicubic interpolation with minor pixel color differences, and one shows pixel discrepancies when skew is applied. These issues are likely due to incorrect floating-point rounding or may require a bias adjustment. I’m not certain yet, but I’ll address them in a later step.


LaurenzV:
> and one shows pixel discrepancies when skew is applied.

I expected the skewing test cases to cause trouble, because their edges cross exactly through the pixel boundary. If I remember correctly, there even are two test cases currently disabled? Anyway, you can try adding a small translation offset in both directions so the edges don't cross the pixel center anymore, and then try whether this fixes the issue.

For bicubic scaling, I haven't looked into how it works on the GPU, but it also was more or less expected that there will be larger differences. Either we can get away by just increasing the threshold, or we introduce some new kind of macro that allows us to generate separate images for hybrid and the CPU renderer.

examples/assets/cowboy.png
LaurenzV, line 1: Don't we already have this file somewhere int he repo? Is there a reason for adding it a second time?

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -53,6 +60,16 @@ impl<'a> ImageFiller<'a> {
...
     pub(super) fn run<F: FineType>(mut self, target: &mut [F]) {
+        // Get the pixmap from the image source
+        let pixmap = match &self.image.source {
+            ImageSource::Pixmap(pixmap) => pixmap,
+            ImageSource::OpaqueId(_) => {
+                panic!(
+                    "CPU renderer does not support OpaqueId image sources. Use ImageSource::Pixmap instead."
+                );
+            }
+        };
```
LaurenzV: Alternatively you can do

```rs
let ImageSource::Pixmap(pixmap) = &self.image.source else { panic!() }
```

sparse_strips/vello_cpu/src/fine/image.rs
```diff
@@ -104,6 +121,14 @@ impl<'a> ImageFiller<'a> {
...
         let mut x_pos = self.cur_pos.x;
         let x_advance = self.image.x_advance.x;
+        let pixmap = match &self.image.source {
+            ImageSource::Pixmap(pixmap) => pixmap,
+            ImageSource::OpaqueId(_) => {
+                panic!(
+                    "CPU renderer does not support OpaqueId image sources. Use ImageSource::Pixmap instead."
+                );
+            }
+        };
```
LaurenzV: Same as above

sparse_strips/vello_sparse_tests/tests/util.rs
```diff
@@ -58,7 +58,7 @@ pub(crate) fn get_ctx<T: Renderer>(
     ctx
 }
 
-pub(crate) fn render_pixmap(ctx: &impl Renderer, render_mode: RenderMode) -> Pixmap {
+pub(crate) fn render_pixmap(ctx: &mut impl Renderer, render_mode: RenderMode) -> Pixmap {
```
LaurenzV: Hmm, what's the reason we need mutable access here? Could we use interior mutability in `vello_hybrid` instead?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -163,3 +314,169 @@ fn unpack4x8unorm(rgba_packed: u32) -> vec4<f32> {
...
+            return clamp(t, 0.0, 1.0);
+        }
+        case EXTEND_REPEAT: {
+            return fract(t);
+        }
+        case EXTEND_REFLECT, default: {
+            return abs(t - 2.0 * round(0.5 * t));
+        }
+    }
+}
```
LaurenzV: Hmm, interesting, I also tried using normalized values on the CPU but ran into a lot of precision issues... But since it seems to work here, maybe I should revisit.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -163,3 +314,169 @@ fn unpack4x8unorm(rgba_packed: u32) -> vec4<f32> {
...
+    let offsets = array<f32, 4>(-1.5, -0.5, 0.5, 1.5);
+    for (var x_idx = 0; x_idx < 4; x_idx++) {
+        for (var y_idx = 0; y_idx < 4; y_idx++) {
+            let offset = vec2<f32>(offsets[x_idx], offsets[y_idx]);
+            let sample_pos_f = coords + offset;
+            let local_xy = sample_pos_f - image_offset;
+            let clamped_pos = vec2<f32>(
+                extend_mode(local_xy.x, extend_modes.x, image_size.x),
+                extend_mode(local_xy.y, extend_modes.y, image_size.y)
+            );
```
LaurenzV: I'm not sure if that matters for the GPU, but I realized that the CPU version is a lot faster if you precalculate the extended positions (see https://github.com/LaurenzV/vello/blob/2482925e7db77574f78574ebb17e32addab8c28f/sparse_strips/vello_cpu/src/fine2/shaders/image.rs#L180-L321 my current SIMD version). But anyway, not urgent, just as a note.

examples/assets/cowboy.png
grebmeg: You’re right — no particular reason for adding it again. Thanks for pointing it out! I’ll update it to use the one from `vello_sparse_tests`.

sparse_strips/vello_cpu/src/fine/image.rs
grebmeg: Fixed

sparse_strips/vello_cpu/src/fine/image.rs
grebmeg: Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Maybe it’s because I’m still using `extend_mode` directly? Also, both of these functions are almost copy-pasted from the `fine.wgsl` code in Vello Classic.

sparse_strips/vello_sparse_tests/tests/util.rs
grebmeg: Certainly, we can use interior mutability in that case.

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Thanks for pointing that out! I’ve refactored the bilinear and bicubic sampling — hopefully it’s more readable now and should be faster on GPU, though I haven’t benchmarked it yet.

Do you have any idea why, in your case with precomputed extended positions, it ends up being significantly faster?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -97,6 +101,55 @@ impl Renderer {
 
         self.scheduler.do_scene(&mut junk, scene)
     }
+
+    /// Upload an image to the atlas texture.
+    pub fn copy_texture_to_atlas(
+        &mut self,
+        encoder: &mut CommandEncoder,
+        texture: &Texture,
```
taj-p: This API necessitates the image be uploaded to a texture prior to rendering even if that image is only ever used in the Atlas.

I wonder whether instead of passing a `wgpu::Texture` directly, we could pass a trait that implements an ability to write to the atlas? 

Then we could implement that trait for `Pixmap` or `Texture` ourselves so that, for users with a pixmap, they can avoid 1 copy (and instead write directly into the atlas with `queue.write_texture` or similar). The pixmap might need to be Arc'd but that's likely cheaper than paying the extra copy and paying for more texture memory.

It also allows consumers to implement that trait if `Pixmap` or `Texture` isn't sufficient.

sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -396,76 +412,83 @@ fn image_bilinear_10x_scale_2(ctx: &mut impl Renderer) {
...
         Affine::scale(5.0) * Affine::translate((10.0, 10.0)),
-        rgb_img_2x2(),
+        image_source,
         ImageQuality::High,
         Extend::Reflect,
     );
 }
 
-#[vello_test]
+#[vello_test(hybrid_tolerance = 7)]
```
grebmeg, line 479: > For bicubic scaling, I haven't looked into how it works on the GPU, but it also was more or less expected that there will be larger differences. Either we can get away by just increasing the threshold, or we introduce some new kind of macro that allows us to generate separate images for hybrid and the CPU renderer.

@LaurenzV While iterating through `hybrid_tolerance`, I found the values where the tests stopped failing. Essentially, we need to add `DEFAULT_HYBRID_TOLERANCE` (equals to 1) added on top to get the right tolerance for pixel differences.

Do these values seem reasonable to you? Also, do you have any ideas on where I might tweak things further to bring the results closer to the CPU version?

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Thanks, @taj-p! That’s a great suggestion. I’ve implemented your idea as I understood it, and the API now feels more ergonomic, customizable, and performant.

Fixed.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Probably because the compiler isn't able to recognize that the y values in the inner loop are the same for each x iteration

sparse_strips/vello_sparse_tests/tests/image.rs
LaurenzV: That's definitely higher than it should be considering that the GPU version is more or less an exact port, but in the interest of moving this PR forward it's probably fine to keep for now? I'll see whether I can take a closer look, but unfortunately since I'm not very familiar with GPU programming it might be hard for me to find out what's going on.

For bilinear there aren't any issues, right? This would at least narrow it down to the bicubic calculations.

grebmeg: Yep, that’s correct — as far as I know, I haven’t run into any edge cases (though I haven’t actively looked for them).

sparse_strips/vello_common/src/encode.rs
```diff
@@ -443,18 +441,30 @@ impl EncodeExt for Image {
         let transform = transform.inverse() * Affine::translate((0.5, 0.5));
         // TODO: This is somewhat expensive for large images, maybe it's not worth optimizing
         // non-opaque images in the first place..
```
DJMcNab: This comment is now dangling; it should have moved with the line below

sparse_strips/vello_common/src/paint.rs
```diff
@@ -51,11 +51,24 @@ impl From<AlphaColor<Srgb>> for Paint {
...
 
+/// Opaque GPU handle
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+pub struct ImageId(pub u32);
+
+/// Bitmap source used by `Image`.
+#[derive(Debug, Clone)]
+pub enum ImageSource {
+    /// Pixmap pixels travel with the scene packet.
+    Pixmap(Arc<Pixmap>),
```
DJMcNab, line 75: I'm generally hesitant to force us to maintain our own atlas; it makes our wgpu behaviour more opaque, forces additional copies, means that we have to guess at image lifecycles if we want to retain things, etc.

Having first-class support for `Texture`s built-in somewhat alleviates this concern. And we can always revisit this later; I won't deny that it's ergonomically helpful to have direct support for images here.
Having these be behind an `Arc` does help, as it guarantees that the pixel content won't change underneath us.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -427,7 +429,7 @@ impl GlyphRenderer for RenderContext {
                 };
 
                 let image = vello_common::paint::Image {
-                    pixmap: Arc::new(glyph.pixmap),
+                    source: ImageSource::Pixmap(Arc::new(glyph.pixmap)),
```
DJMcNab, line 432: I think we really want to cache these arced bitmaps if possible. That's not for this PR, though; it's pre-existing, and is probably part of a general glyph caching feature

sparse_strips/vello_hybrid/examples/scenes/Cargo.toml
```diff
@@ -12,10 +12,12 @@ workspace = true
 [dependencies]
 bytemuck = { workspace = true, features = [] }
 vello_hybrid = { workspace = true }
-vello_common = { workspace = true, features = ["pico_svg"] }
+vello_common = { workspace = true, features = ["pico_svg", "png"] }
```
DJMcNab, line 15: I *think* that `vello_hybrid` should have a pass through `png` feature, which is enabled by default.

Needing to add a dependency on `vello_common` to get png emoji working would not be a good state. (In writing this comment, I'm presuming without checking that this example depends on `vello_common` only for the `pico_svg` feature)

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -15,6 +15,8 @@ use std::rc::Rc;
 #[cfg(target_arch = "wasm32")]
 use vello_common::kurbo::{Affine, Point};
 #[cfg(target_arch = "wasm32")]
+use vello_hybrid::ImageCache;
+#[cfg(target_arch = "wasm32")]
```
DJMcNab: Every single item in this crate having `cfg(target_arch = "wasm32")` is not the right solution here. A top-level `#![cfg(target_arch = "wasm32")]` should be [valid](https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.attr.crate-level-attrs).

Again, not for this PR, but we should make this into an issue. cc @ajakubowicz-canva

(Ha, on new line 36, it's even nested!)

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// The offset of the image in the atlas
+    pub offset: [u32; 2],
+    /// The atlas allocation ID for deallocation
+    pub(crate) atlas_alloc_id: AllocId,
+}
+
+/// Manages image resources for the renderer
+#[allow(
+    missing_debug_implementations,
+    reason = "AtlasAllocator doesn't implement Debug"
```
DJMcNab: We used a manual debug impl in Vello Classic to work around this (I think).
Even if we didn't, I think it's worth having

sparse_strips/vello_hybrid/examples/scenes/src/image.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn render(&mut self, scene: &mut Scene, root_transform: Affine) {
+        scene.set_transform(
+            root_transform
+                * Affine::translate((200.0, 50.0))
+                * Affine::rotate(PI / 4.0)
+                * Affine::scale(0.5),
+        );
+        scene.set_paint_transform(Affine::translate((0.0, 0.0)));
+        scene.set_paint(Image {
+            source: ImageSource::OpaqueId(ImageId(0)),
```
DJMcNab: This feels incredibly hard to coordinate on larger scales. This is `0` only because we "know" it will be the first image allocated, right?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -778,3 +1115,145 @@ impl RendererBackend for RendererContext<'_> {
...
+                    z: 0,
+                },
+                aspect: wgpu::TextureAspect::All,
+            },
+            wgpu::Extent3d {
+                width,
+                height,
+                depth_or_array_layers: 1,
+            },
+        );
```
DJMcNab, line 1223: In theory, part of the sparse strips/spatiotemporal allocation design was that we wouldn't need to atlas wgpu Textures, because switching pipelines is relatively cheap.

However, getting a first version which works with an atlas here is important. I believe that the experiments with different forms here can be follow-ups.

sparse_strips/vello_common/src/encode.rs
grebmeg: Ah, I missed that — thanks!

Fixed

sparse_strips/vello_common/src/paint.rs
grebmeg: I haven’t found a good way to support `Texture`s properly, as each strip would require binding its own texture. While it may be possible, it would add considerable overhead and complexity — managing draw call splitting and ensuring the correct texture is bound at the right time. By contrast, a texture atlas, though a bit more opaque, can offer much better performance, simpler support, and allow for late binding directly in the shader. With a thoughtfully designed and ergonomic API, I’m hopeful we can mitigate its drawbacks.

sparse_strips/vello_cpu/src/render.rs
grebmeg: Are you referring to caching specifically for the CPU case? Because for the GPU case, the atlas copy seems sufficient to me?

sparse_strips/vello_hybrid/examples/scenes/Cargo.toml
grebmeg: Are you saying that `vello_hybrid` should handle PNG/JPEG image decoding directly, without requiring users to manually enable a feature or add a dependency like `vello_common`, so that things like PNG emojis just work out of the box?

sparse_strips/vello_hybrid/examples/scenes/src/image.rs
grebmeg: Yes, right. I think the step I’m skipping (and just using integers) is registering my scene elements in the storage. For elements that use images, I should upload the images first, then assign the corresponding `ImageId`s. That way, when building the scene alongside other data, `ImageId` would already be available for use. Also, I think there should be a mapping between image data and `ImageId` somewhere in the system. This is my current understanding of how it should work, but I’d be interested in seeing real-world use cases to validate it.

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: I don’t think we have [that](https://github.com/linebender/vello/blob/8a2b6b4658e7b873024038dffa7ea709f21107cf/vello_encoding/src/image_cache.rs#L19), but I agree it’s worth having.

Fixed

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: I agree with exploring follow-ups and finding the most efficient and user-friendly solution here. That said, as I mentioned in the comment above, I don’t currently have a better idea — though it’s possible I’m overlooking something.

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
ajakubowicz-canva: I can take this cleanup task. Thank you!

ajakubowicz-canva: Done here: https://github.com/linebender/vello/pull/1070

sparse_strips/vello_cpu/src/render.rs
DJMcNab: What I mean is that we seem to be allocating a new `Arc` around the pixmap data for each time you render an emoji, even though the data is always the same for that same emoji (for the correct definition of the same emoji).

As I say, this doesn't need to happen in this PR

sparse_strips/vello_common/src/paint.rs
DJMcNab: My comment here is also looking at:
> The paint type is different
> The image being sampled is different and we don’t have bindless

From the [Potato design document](https://docs.google.com/document/d/1gEqf7ehTzd89Djf_VpkL0B_Fb15e0w5fuv_UzyacAPU/edit?tab=t.0). That is, the claim being made originally was that switching pipelines is relatively cheap, if you can batch the calls to a given pipeline.

I've not put too much further thought into this, and non-compute rendering pipelines is an area where I have a lot to learn

sparse_strips/vello_hybrid/examples/scenes/Cargo.toml
DJMcNab: Yes, exactly. As far as I know, we don't need to handle JPEGs for anything, but we do want to support PNG emoji by default.
 `vello_cpu` has a `"png"` default feature, which enables `vello_common/png`. I think that `vello_hybrid` probably should have the same. It should be possible to use vello_hybrid without png decoding (for example, for a future `wasm32v1-none` webgl which the png crate currently doesn't support)

sparse_strips/vello_hybrid/src/image_cache.rs
DJMcNab: Thanks for checking. I know I've written a few manual debug impls, but I guess it was for different types

grebmeg:
Added the WebGL implementation

sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -162,13 +162,14 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
...
         input_fn_name_str.contains("compose")
             || input_fn_name_str.contains("gradient")
-            || input_fn_name_str.contains("image")
             || input_fn_name_str.contains("layer")
             || input_fn_name_str.contains("mask")
             || input_fn_name_str.contains("mix")
             || input_fn_name_str.contains("blurred_rounded_rect")
     };
 
+    let skip_hybrid_webgl = skip_hybrid || input_fn_name_str.contains("image");
```
ajakubowicz-canva: Does webgl still need to be skipped?

sparse_strips/vello_hybrid/examples/winit/src/main.rs
```diff
@@ -44,12 +45,13 @@ struct App<'s> {
...
     mouse_down: bool,
     last_cursor_position: Option<Point>,
+    image_cache: ImageCache,
 }
 
 fn main() {
     #[cfg(not(target_arch = "wasm32"))]
     let (scenes, start_scene_index) = {
-        let mut start_scene_index = 0;
+        let mut start_scene_index = 4;
```
taj-p:
```suggestion
        let mut start_scene_index = 0;
```

I think this was leftover from development - but I think most people expect to see ghost tiger on load

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -50,9 +51,21 @@ impl AppState {
...
+            .get_parameter(web_sys::WebGl2RenderingContext::MAX_TEXTURE_SIZE)
+            .unwrap()
+            .as_f64()
+            .unwrap() as u32;
+        image_cache.resize(max_texture_dimension_2d, max_texture_dimension_2d);
+
+        let mut app_state = Self {
             scenes,
-            current_scene: 0,
+            current_scene: 4,
```
taj-p:
```suggestion
            current_scene: 0,
```

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -50,9 +51,21 @@ impl AppState {
...
+        let mut image_cache = vello_hybrid::ImageCache::new();
+        // Get the max texture dimension from the WebGL renderer
+        let max_texture_dimension_2d = renderer_wrapper
+            .renderer
+            .gl_context()
+            .get_parameter(web_sys::WebGl2RenderingContext::MAX_TEXTURE_SIZE)
+            .unwrap()
+            .as_f64()
+            .unwrap() as u32;
+        image_cache.resize(max_texture_dimension_2d, max_texture_dimension_2d);
```
taj-p: I think it might be easier to use if `ImageCache::new` took a `width` and `height`

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,283 @@
...
+            free_idxs: Vec::new(),
+        }
+    }
+
+    /// Resize the image cache
+    pub fn resize(&mut self, width: u32, height: u32) {
+        self.atlas = AtlasAllocator::new(size2(width as i32, height as i32));
+        self.slots.clear();
+        self.free_idxs.clear();
+    }
```
taj-p: In practice, how often do you think a user will need to resize an image cache? The reason I ask is because the name `resize` doesn't indicate to the user that this is essentially a `self.clear` followed by a re-initialisation of the atlas. I wonder whether we could omit this API entirely, make `width` and `height` immutable, and pass them via `new` instead. Perhaps we can implement a `default` that uses the default sizes.

If you think it's worth keeping, then what do you think about renaming it to `reset_with_size` to make its behavior clearer?

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Get an image resource by its Id
+    pub fn get(&self, id: ImageId) -> Option<&ImageResource> {
+        let index = id.0 as usize;
+        if index < self.slots.len() {
+            self.slots[index].as_ref()
+        } else {
+            None
+        }
```
taj-p:
```suggestion
        self.slots.get(id.0 as usize)?.as_ref()
```

I think this is equivalent

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,283 @@
...
+        let image_resource = ImageResource {
+            id: image_id,
+            width,
+            height,
+            offset: [x, y],
+            atlas_alloc_id,
+        };
+
+        self.slots[index] = Some(image_resource);
+        image_id
```
taj-p:
```suggestion
        let alloc = self
            .atlas
            .allocate(size2(width as i32, height as i32))
            .expect("Failed to allocate texture");

        let slot_idx = self.free_idxs.pop().unwrap_or_else(|| {
            // No free slots, append to vector
            let index = self.slots.len();
            // Placeholder, will be replaced
            self.slots.push(None);
            index
        });

        #[expect(
            clippy::cast_possible_truncation,
            reason = "u32 is enough for the index"
        )]
        let image_id = ImageId(slot_idx as u32);
        self.slots[slot_idx] = Some(ImageResource {
            id: image_id,
            width,
            height,
            offset: [alloc.rectangle.min.x as u32, alloc.rectangle.min.y as u32],
            atlas_alloc_id: alloc.id,
        });
        image_id
```

Ultra nit: I would generally write it like this

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,283 @@
...
+                // Deallocate from the atlas using the stored allocation ID
+                self.atlas.deallocate(image_resource.atlas_alloc_id);
+
+                // Mark slot as free and add to free list
+                self.slots[index] = None;
+                self.free_idxs.push(index);
+                return true;
+            }
+        }
+        false
```
taj-p:
```suggestion
        if let Some(image_resource) = self.slots.get_mut(index).and_then(Option::take) {
            // Deallocate from the atlas using the stored allocation ID
            self.atlas.deallocate(image_resource.atlas_alloc_id);
            self.free_idxs.push(index);
            true
        } else {
            false
        }
```

We can eliminate the manual bounds checks, reduce the nesting, and reduce the branching by `Option::take` and using `get_mut`

sparse_strips/vello_common/src/paint.rs
```diff
@@ -51,11 +51,24 @@ impl From<AlphaColor<Srgb>> for Paint {
     }
 }
 
+/// Opaque GPU handle
```
taj-p:
```suggestion
/// Opaque image handle
```

I'm not sure whether we should add the `GPU` part because this code is used in common with CPU

sparse_strips/vello_common/src/paint.rs
```diff
@@ -51,11 +51,24 @@ impl From<AlphaColor<Srgb>> for Paint {
...
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+pub struct ImageId(pub u32);
```
taj-p: Is this an opaque ID if consumers are free to create their own ImageIds and its internals are public?

I was expecting something like this:

```rs
pub struct ImageId(u32);

impl ImageId {
    pub(crate) fn new(value: u32) -> Self {
        Self(value)
    }
}
```

Or maybe

```rs
pub struct ImageId(u32);

impl ImageId {
    pub(crate) fn new(value: u32) -> Self {
        Self(value)
    }
    
    pub(crate) fn as_u32(&self) -> u32 {
        self.0
    }
}
```

But, I suspect that this comment is a lot to change. So, happy to just leave a TODO or similar to make this truly opaque in the future.

sparse_strips/vello_dev_macros/src/test.rs
taj-p: I removed the skip on image tests and it passed locally. Does this mean `ignore_hybrid_webgl` and other parts of this file can be reverted?

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -21,6 +21,7 @@ vello_common = { workspace = true, features = ["std", "text"] }
 wgpu = { workspace = true, optional = true }
 vello_sparse_shaders = { workspace = true, optional = true }
 log = { workspace = true }
+guillotiere = "0.6.2"
```
taj-p, line 24: I'm interested in how you landed on `guillotiere` for the allocator. In [Nical's blog post](https://nical.github.io/posts/etagere.html), it seemed that the shelf allocator was best:

> Overall the best compromise was the simple shelf allocator which I ended up using in Firefox for both glyph and image textures in the cache (in both cases configured to have two columns per texture).

And Guillotine is used for short lived atlases

> In other parts of WebRender, for short lived atlases (a single frame), the guillotine allocation algorithm is used.

That said, our characteristics may differ from Nical's because our allocator is for both images and glyphs. I wonder whether a future activity should split the two - one texture and allocator for images and another for glyphs.

This sentiment is expressed in [etagere's README](https://github.com/nical/etagere):

> The shelf packing algorithm works very well when there is a high number of items with similar sizes, for example for dynamic glyph atlases.

So, perhaps, until we have separate atlases for images and glyphs, we'll need to stick with this?

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,283 @@
...
+#[derive(Debug)]
+pub struct ImageResource {
+    /// The ID of the image
+    pub id: ImageId,
+    /// The width of the image
+    pub width: u32,
+    /// The height of the image
+    pub height: u32,
+    /// The offset of the image in the atlas
+    pub offset: [u32; 2],
```
taj-p:
```suggestion
    pub width: u16,
    /// The height of the image
    pub height: u16,
    /// The offset of the image in the atlas
    pub offset: [u16; 2],
```

Do you think u16 is enough?

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+
+impl Default for ImageCache {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ImageCache {
```
taj-p, line 68: I'm a little unsure of the access modifiers on `ImageCache`. I am wondering whether `allocate` and `get` need to be exposed to consumers. For example, to correctly upload some image into the system, we need to call `renderer.upload_image` and pass in the cache. That method then orchestrates the allocation. Since this is internal, can we make `allocate` and `get` `pub(crate)`?

I am also wondering what benefit ImageResource being public to consumers has. 

---

I am also unsure about the API ergonomics. To correctly allocate an image, a consumer has to go through `renderer.upload_image` but to correctly deallocate an image, a consumer has to `image_cache.deallocate`. 

I wonder whether we can instead make `ImageCache` public only to the crate and for consumers to use the `renderer` as the facade into image management so we aren't splitting responsibility between the renderer and `ImageCache`:

```rs
renderer.upload_image(...);
renderer.destroy_image(...);
```

Beyond API ergonomics, I think this API is safer for consumers because currently this flow is allowed which would cause clobbering of images in our internal atlas texture. So, by binding the cache to the renderer, we are also binding the cache to the texture it is managing:

```rs
let mut image_cache_1 = ImageCache::new();
let mut image_cache_2 = ImageCache::new();
let image_id_1 = renderer.upload_image(&mut image_cache_1, ...); // upload [100,100] in atlas texture
let image_id_2 = renderer.upload_image(&mut image_cache_2, ...); // overwrite [100,100] in atlas texture

// drawing image_id_1 will draw image_id_1
```

I think the importance of this becomes more evident when we start to consider storing an array of atlas textures than just one too.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -526,6 +527,37 @@ impl Scheduler {
...
+                let rgba = color.as_premul_rgba8().to_u32();
+                debug_assert!(
+                    has_non_zero_alpha(rgba),
+                    "Color fields with 0 alpha are reserved for clipping"
+                );
+                (alpha_col, 0, rgba)
+            }
+            Paint::Indexed(indexed_paint) => {
+                let paint_id = indexed_paint.index();
+                // 16 bytes per textel: Rgba32Uint (4 bytes) * 4 (4 textels)
```
```suggestion
                // 16 bytes per texel: Rgba32Uint (4 bytes) * 4 (4 texels)
```

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -44,9 +44,37 @@ pub struct GpuStrip {
...
+#[allow(
+    dead_code,
+    reason = "Clippy fails when --no-default-features because webgl impl doesn't use this struct"
+)]
+pub(crate) struct GpuEncodedImage {
+    // 1st 16 bytes
+    /// The rendering quality of the image.
+    pub quality: u32,
+    /// The extends in the horizontal and vertical direction.
+    pub extend_modes: [u32; 2],
```
taj-p: Is it possible to pack quality and extend modes into a single u32? From what I can tell, `quality` is a 3 variant enum and so too is `extend_modes`?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -44,9 +44,37 @@ pub struct GpuStrip {
...
+    pub _padding0: u32,
+    // 2nd 16 bytes
+    /// The size of the image in pixels.
+    pub image_size: [u32; 2],
+    /// The offset of the image in pixels.
+    pub image_offset: [u32; 2],
```
taj-p:
```suggestion
    pub image_size: [u16; 2],
    /// The offset of the image in pixels.
    pub image_offset: [u16; 2],
```

sparse_strips/vello_hybrid/src/render/common.rs
taj-p: After the changes to `GpuEncodedImage`, we might be able to remove some padding.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
taj-p, line 1: It would be nice in the future if we had some vello hybrid benchmarks so we can better evaluate the impact of shader updates on performance

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -44,9 +44,37 @@ pub struct GpuStrip {
...
+    // 1st 16 bytes
+    /// The rendering quality of the image.
+    pub quality: u32,
+    /// The extends in the horizontal and vertical direction.
+    pub extend_modes: [u32; 2],
+    pub _padding0: u32,
+    // 2nd 16 bytes
+    /// The size of the image in pixels.
+    pub image_size: [u32; 2],
+    /// The offset of the image in pixels.
```
taj-p:
```suggestion
    /// The offset of the image in the atlas texture in pixels.
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -219,6 +331,12 @@ struct WebGlResources {
     alphas_texture: WebGlTexture,
     /// Height of alpha texture.
     alpha_texture_height: u32,
+    /// Atlas texture for image data.
+    atlas_texture: WebGlTexture,
```
taj-p, line 357: Do we need a TODO/issue/some way to track allowing for more than 1 texture's worth of images? I think this is quite a limitation which would prevent our use of image rendering at Canva but happy to do this separately in the future.

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -44,9 +44,37 @@ pub struct GpuStrip {
     /// Column-index into the alpha texture where this strip's alpha values begin.
     ///
     /// There are [`Config::strip_height`] alpha values per column.
-    pub col: u32,
-    /// RGBA color value
-    pub rgba: u32,
+    pub col_idx: u32,
+    /// Paint data
+    pub rgba_or_slot: u32,
+    /// Paint type: 0 = solid, 1 = alpha, 2 = image
```
taj-p, line 51: Should we also mention this is a packed type with the paint type and idx together

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -44,9 +44,37 @@ pub struct GpuStrip {
     /// Column-index into the alpha texture where this strip's alpha values begin.
     ///
     /// There are [`Config::strip_height`] alpha values per column.
-    pub col: u32,
-    /// RGBA color value
-    pub rgba: u32,
+    pub col_idx: u32,
+    /// Paint data
```
taj-p:
```suggestion
    /// Color value or slot index when alpha is 0
```

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -44,9 +44,37 @@ pub struct GpuStrip {
...
+    pub rgba_or_slot: u32,
```
taj-p, line 49: I actually think you might be able to use this `u32` for `paint` instead of creating a new field for it. 

Currently:
- `alpha` == 0 -> this is a slot index (slots range from `0 < n < (max_texture_dimension_2d / u32::from(Tile::HEIGHT)`
- `alpha` != 0 -> this is a color value

Since slots only require ~10 bits and `paint` encodes an id, presumably, there are enough bits leftover to cover all the images one would want to paint. So something like:


- `alpha` == 0 -> 
    - 9th bit == 1 this is a slot index (slots range from `0 < n < (max_texture_dimension_2d / u32::from(Tile::HEIGHT)`
    - 9th bit == 0, bits 23 and 22 are paint type and the rest are the index
- `alpha` != 0 -> this is a color value

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -167,6 +178,99 @@ impl WebGlRenderer {
...
+                                image_size: [image_resource.width, image_resource.height],
+                                image_offset: image_resource.offset,
+                                transform: transform.as_coeffs().map(|x| x as f32),
+                                _padding1: [0, 0],
+                            });
+                        }
+                    }
+                }
+                _ => {
+                    unimplemented!("For now, we don't support gradient and rounded rectangle")
```
```suggestion
                    unimplemented!("Gradient and rounded rectangle unsupported")
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -458,6 +605,62 @@ impl WebGlPrograms {
...
+            gl.tex_parameteri(
+                WebGl2RenderingContext::TEXTURE_2D,
+                WebGl2RenderingContext::TEXTURE_WRAP_S,
+                WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
+            );
+            gl.tex_parameteri(
+                WebGl2RenderingContext::TEXTURE_2D,
+                WebGl2RenderingContext::TEXTURE_WRAP_T,
+                WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
+            );
```
taj-p: Question: Do we need to reassign these parameters if we've already set them when we initially created the texture?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -167,6 +178,99 @@ impl WebGlRenderer {
...
+    }
+
+    /// Write image data to the atlas texture using any type that implements `AtlasWriter`.
+    ///
+    /// This is the WebGL analogue of the wgpu Renderer's `write_to_atlas` method.
+    /// This method allows efficient uploading from different sources:
+    /// - `Pixmap`: Direct upload without intermediate texture
+    /// - `ImageData`: Browser `ImageData` objects
+    /// - Custom implementations for other image sources
+    pub fn write_to_atlas<T: WebGlAtlasWriter>(
```
taj-p: What is the purpose of this public method? Will there be consumers who want to write to the atlas directly without associating it with a given image ID?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -167,6 +178,99 @@ impl WebGlRenderer {
...
+            offset,
+            width,
+            height,
+        );
+    }
+
+    /// Get a reference to the underlying WebGL context.
+    ///
+    /// This allows direct access to WebGL operations for advanced use cases like texture creation.
+    pub fn gl_context(&self) -> &WebGl2RenderingContext {
```
taj-p, line 215:
```suggestion
    pub fn gl(&self) -> &WebGl2RenderingContext {
```

I think it's customary in opengl to refer to the GL ctx as just gl

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -97,6 +102,102 @@ impl Renderer {
...
+        image_id
+    }
+
+    /// Write image data to the atlas texture using any type that implements `AtlasWriter`.
+    ///
+    /// This method allows efficient uploading from different sources:
+    /// - `Pixmap`: Direct upload without intermediate texture
+    /// - `Texture`: Texture-to-texture copy
+    /// - Custom implementations for other image sources
+    pub fn write_to_atlas<T: AtlasWriter>(
```
taj-p: As in the WebGL renderer, I'm not sure how this would be used - does a consumer need to be aware of which `ImageId` is mapped to which location in the atlas?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -28,24 +38,39 @@ struct Config {
...
     @location(3) rgba_or_slot: u32,
+    // Packed paint type (2 bits) and paint texture id (30 bits)
+    // Paint type: 0 = solid, 1 = alpha, 2 = image
+    // Paint texture id locates the encoded image data `EncodedImage` in the encoded_paints_texture
+    @location(4) paint: u32,
 }
 
 struct VertexOutput {
+    // Render type for the strip
+    @location(0) @interpolate(flat) paint_type: u32,
```
taj-p: Separating `in.paint` into `out.paint_type` and `out.paint_tex_id` and passing more memory from the vertex to fragment shader is likely more expensive than adding the bitshift in the fragment shader and passing a single `out.paint`. 

But this might change if we combine paint and `rgba_or_slot`

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -28,24 +38,39 @@ struct Config {
...
-    @location(0) tex_coord: vec2<f32>,
+    @location(1) tex_coord: vec2<f32>,
+    // UV coordinates for the current fragment, used for image sampling
+    @location(2) sample_xy: vec2<f32>,
     // Ending x-position of the dense (alpha) region
-    @location(1) @interpolate(flat) dense_end: u32,
+    @location(3) @interpolate(flat) dense_end: u32,
     // Color value or slot index when alpha is 0
-    @location(2) @interpolate(flat) rgba_or_slot: u32,
+    @location(4) @interpolate(flat) rgba_or_slot: u32,
```
taj-p, line 71: As mentioned in a prior comment, I suspect this `u32` can be combined with paint.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -234,6 +259,7 @@ impl Scene {
         self.line_buf.clear();
         self.tiles.reset();
         self.strip_buf.clear();
+        self.encoded_paints.clear();
 
         let render_state = Self::default_render_state();
         self.transform = render_state.transform;
```
taj-p, line 265: Do we need to set the paint transform?

sparse_strips/vello_dev_macros/src/test.rs
grebmeg: Ah, yes — my bad. I think it should generally follow `skip_hybrid` option, and if needed during development, we should be able to disable it for certain tests in the webgl implementation. I’ve removed the image skipping I accidentally left in.

taj-p (commented):
I haven't quite finished the review, but I see you're committing so I'll submit early and finish up the review by EOD

sparse_strips/vello_common/src/paint.rs
grebmeg: Yes, agreed

Fixed

sparse_strips/vello_common/src/paint.rs
grebmeg: I took your second example as the basis for now. I think it’s better to leave the true purpose implicit for the moment and adjust it later as needed.

sparse_strips/vello_hybrid/Cargo.toml
grebmeg: I decided to use the same allocator as Vello Classic for now. Rather than focusing on which allocator is best, I’ve chosen to defer that decision and concentrate on image rendering first. That said, I’m happy to explore which allocator would be most suitable for our use case. 

Regarding a separate glyph atlas — I believe the plan is to try using a sparse strip representation, right? In that case, only a small number of glyphs (like emojis) might actually need to be placed in the atlas/texture.

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/examples/winit/src/main.rs
grebmeg: Ah, yes — I missed that. Thanks!

Fixed

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: u16 is sufficient, but the WGPU API expects u32, so some additional conversions are needed — though the WebGL API expects i32.

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: Thank you! That’s a great suggestion — I’ll implement it.

Fixed

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: I don’t think it’s worth keeping for now. If we do end up needing it, we can add it later when the feature is actually requested.

Fixed

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: I initially took a similar approach, but eventually ended up adding an extra field — though I don’t exactly remember why, so let me address this in a follow-up.

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Removed

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Sure, it’s something I plan to add as a future improvement. I’ve added TODOs in both implementations.

Fixed

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Correct, it’s not needed.

Fixed

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Removed it from both implementations.

Fixed

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Good point — we do need that.

Fixed

sparse_strips/vello_hybrid/src/schedule.rs
grebmeg: Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Yes, it’s something I’ve just started evaluating and implementing because I believe it’s definitely worth pursuing. So far, based on profiling on macOS, I’ve observed that rendering `ImageScene` takes approximately 72µs in the vertex shader and 197µs in the fragment shader. Creating a dedicated benchmarking tool would definitely be worthwhile.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Yep, agree

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Yes, I’ll address it as a follow-up.

sparse_strips/vello_hybrid/src/image_cache.rs
taj-p: Ah yup! I was thinking it'd be similar to how `GpuStrip` uses u16 for x and y and unpacks in shader

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -167,6 +177,81 @@ impl WebGlRenderer {
         }
         Ok(())
     }
+
+    /// Upload image to cache and atlas in one step. Returns the `ImageId`.
+    ///
+    /// This is the WebGL analogue of the wgpu Renderer's `upload_image` method.
+    /// It allocates space in the image cache and uploads the image data to the atlas texture.
+    pub fn upload_image<T: WebGlAtlasWriter>(
```
taj-p, line 185: Do we need a function to destroy an image by `ImageId`?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -97,6 +105,83 @@ impl Renderer {
...
+
+    /// Upload image to cache and atlas in one step. Returns the `ImageId`.
+    ///
+    /// It's used when an image is not already in the cache.
+    ///
+    /// This is a convenience method that:
+    /// 1. Reserves space in the image cache
+    /// 2. Writes the image data directly to the atlas
+    /// 3. Returns the `ImageId` for use in rendering
+    pub fn upload_image<T: AtlasWriter>(
```
taj-p, line 117: Do we need a function to destroy an image by ImageId?

taj-p (approved):
LGTM! Great work @grebmeg - really impressive!!! So good to see images in hybrid 🎉 

---

The shader is getting a bit worrisome with the amount of divergence it's seeing for image rendering. IIUC warps/wavefronts correctly then each thread in the group has to perform all branches of a condition if at least 1 thread in the group needs to enter that branch. So, if 1 pixel needs to do bicubic sampling, then all threads need to do 16 texture loads (but with their results masked out).

I don't actually know whether splitting up the pipeline will see significantly improved performance - I suspect so (with some rocket science scheduling), but I'm not 100% sure. In any case, this does seem like the right strategy - go with the simple option first and then improve from there (especially when we don't have benchmarking in place).

sparse_strips/vello_common/src/paint.rs
```diff
@@ -51,11 +51,36 @@ impl From<AlphaColor<Srgb>> for Paint {
     }
 }
 
+/// Opaque image handle
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+pub struct ImageId(u32);
+
+impl ImageId {
+    /// Create a new image id from a u32.
+    pub fn new(value: u32) -> Self {
```
taj-p, line 61: Do we want to add a TODO / note somewhere that we'll want to make this private in future?

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Yes, we need it — I’ve included an initial implementation in this PR and will add further handling and examples in a follow-up PR.

Fixed

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: ditto

sparse_strips/vello_common/src/paint.rs
grebmeg: Added

sparse_strips/vello_hybrid/examples/scenes/src/image.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn render(&mut self, scene: &mut Scene, root_transform: Affine) {
+        scene.set_transform(
+            root_transform
+                * Affine::translate((200.0, 50.0))
+                * Affine::rotate(PI / 4.0)
+                * Affine::scale(0.5),
+        );
+        scene.set_paint_transform(Affine::translate((0.0, 0.0)));
+        scene.set_paint(Image {
+            source: ImageSource::OpaqueId(ImageId::new(0)),
```
ajakubowicz-canva: Nit: Could be more readable to have at the top the image ids defined to names:
 - `let splash_flower_id = ImageId::new(0)`

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -170,6 +174,77 @@ impl AppState {
...
+        self.renderer_wrapper.renderer.upload_image(&pixmap1);
+
+        // 2nd example — uploading from a WebGL texture
+        let pixmap2 = ImageScene::read_cowboy_image();
+        let texture2 = self.pixmap_to_webgl_texture(&pixmap2);
+        self.renderer_wrapper.renderer.upload_image(&texture2);
+    }
+
+    /// Convert a pixmap to WebGL texture
+    fn pixmap_to_webgl_texture(
```
ajakubowicz-canva, line 194: Is this being done for demonstrative purposes that a WebGL texture can be passed into `upload_image`? Because, IIUC the pixmaps could be uploaded directly.

grebmeg: Yes, it was only added for demonstration purposes.

sparse_strips/vello_hybrid/examples/scenes/src/image.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/examples/scenes/src/lib.rs
```diff
@@ -66,6 +67,7 @@ pub fn get_example_scenes(svg_paths: Option<Vec<&str>>) -> Box<[AnyScene]> {
     scenes.push(AnyScene::new(text::TextScene::new("Hello, Vello!")));
     scenes.push(AnyScene::new(simple::SimpleScene::new()));
     scenes.push(AnyScene::new(clip::ClipScene::new()));
+    scenes.push(AnyScene::new(image::ImageScene {}));
```
ajakubowicz-canva, line 70: Nit: I feel like `ImageScene` should also have a `::new()` for consistency.

grebmeg:
> In any case, this does seem like the right strategy - go with the simple option first and then improve from there (especially when we don't have benchmarking in place).

I agree with the points above — this is something I’ve been concerned about too. I’m also unsure which option would perform better, especially without a benchmarking setup in place, which makes it harder to reason about. With the current implementation in place, we can now start benchmarking and use that data to guide improvements going forward.

> edit: Are we able to ensure those follow up suggestions are tracked somewhere (even if it's just in a personal note)?

Sure 👌

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -477,6 +702,75 @@ impl Programs {
         })
     }
 
+    fn make_atlas_texture(
```
ajakubowicz-canva, line 705: I turned off the `webgl` polyfill from the `wgpu_webgl` example and tried to run the example. The example crashes with:

```
Buffer size (1073741824) exceeds the max buffer size limit (268435456). This adapter supports a higher maxBufferSize of 4294967296, which can be specified in requiredLimits when calling requestDevice(). Limits differ by hardware, so always check the adapter limits prior to requesting a higher limit.
 - While calling [Queue].WriteTexture([TexelCopyTextureInfo texture: [Texture "Atlas Texture"], mipLevel: 0, origin: [Origin3D x:0, y:0, z:0], aspect: TextureAspect::All], (1228800 bytes), [TexelCopyBufferLayout offset:0, bytesPerRow:2560, rowsPerImage:480], [Extent3D width:640, height:480, depthOrArrayLayers:1])

```


I think using max texture for both width and height hits a secondary limit for max buffer size.

I could get past this error via the very hacky:

```rs
        let atlas_texture = Self::make_atlas_texture(
            device,
            max_texture_dimension_2d,
            max_texture_dimension_2d / 4, // Division by 4 just to shrink buffer (not proposing this)
            wgpu::TextureFormat::Rgba8Unorm,
        );
```

sparse_strips/vello_hybrid/examples/scenes/src/lib.rs
grebmeg: Addressed in #1077

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Addressed in #1077

PR #1065, opened 2025-06-24
[vello_sparse_tests] Add vello_cpu scalar test coverage on WASM

ajakubowicz-canva:
### Context

This is a test only PR - adding additional test coverage for scalar vello_cpu tests running on the browser.

**Before this PR:**

```sh
$ wasm-pack test --firefox --headless --features=webgl
test result: ok. 85 passed; 0 failed; 165 ignored; 0 filtered out; finished in 7.49s
```

**After this PR:**

```sh
$ wasm-pack test --firefox --headless --features=webgl
test result: ok. 581 passed; 0 failed; 169 ignored; 0 filtered out; finished in 22.18s
```

This can also be seen [in CI](https://github.com/linebender/vello/actions/runs/15842375361/job/44657293874?pr=1065#step:11:790) of this PR.

### Changes 

 - `load_image` has been moved into `utils` and has become a macro. This is required such that WASM can embed the image assets into the compiled WASM binary (as wasm doesn't have access to filesystem).
 - Most changes are to the `vello_test` proc macro, which now generates two additional tests, a u8 and f32 scalar wasm cpu test.

### Test plan

This whole PR is the test.

### DIFF
--- a/sparse_strips/vello_dev_macros/src/test.rs
+++ b/sparse_strips/vello_dev_macros/src/test.rs
@@ -80,6 +80,14 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
         &format!("{}_cpu_f32_neon", input_fn_name),
         input_fn_name.span(),
     );
+    let u8_fn_name_scalar_wasm = Ident::new(
+        &format!("{}_cpu_u8_scalar_wasm", input_fn_name),
+        input_fn_name.span(),
+    );
+    let f32_fn_name_scalar_wasm: Ident = Ident::new(
+        &format!("{}_cpu_f32_scalar_wasm", input_fn_name),
+        input_fn_name.span(),
+    );
     let multithreaded_fn_name = Ident::new(
         &format!("{}_cpu_multithreaded", input_fn_name),
         input_fn_name.span(),
@@ -99,6 +107,8 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
     let f32_fn_name_str_scalar = f32_fn_name_scalar.to_string();
     let u8_fn_name_str_neon = u8_fn_name_neon.to_string();
     let f32_fn_name_str_neon = f32_fn_name_neon.to_string();
+    let u8_fn_name_scalar_wasm_str = u8_fn_name_scalar_wasm.to_string();
+    let f32_fn_name_scalar_wasm_str = f32_fn_name_scalar_wasm.to_string();
     let multithreaded_fn_name_str = multithreaded_fn_name.to_string();
     let hybrid_fn_name_str = hybrid_fn_name.to_string();
     let webgl_fn_name_str = webgl_fn_name.to_string();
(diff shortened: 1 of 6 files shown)


sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -80,6 +80,14 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
...
         input_fn_name.span(),
     );
+    let u8_fn_name_scalar_wasm = Ident::new(
+        &format!("{}_cpu_u8_scalar_wasm", input_fn_name),
+        input_fn_name.span(),
+    );
+    let f32_fn_name_scalar_wasm: Ident = Ident::new(
+        &format!("{}_cpu_f32_scalar_wasm", input_fn_name),
+        input_fn_name.span(),
+    );
```
LaurenzV, line 90: In the long term, are we planning to run scalar AND SIMD on WASM, or just SIMD?

sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -2,44 +2,39 @@
...
 use vello_dev_macros::vello_test;
 
-pub(crate) fn load_image(name: &str) -> Arc<Pixmap> {
-    let path = Path::new(env!("CARGO_MANIFEST_DIR")).join(format!("tests/assets/{name}.png"));
-    Arc::new(Pixmap::from_png(std::fs::File::open(path).unwrap()).unwrap())
-}
-
 fn rgb_img_10x10() -> Arc<Pixmap> {
-    load_image("rgb_image_10x10")
+    load_image!("rgb_image_10x10")
```
LaurenzV, line 17: Just FYI there will probably be conflicts with the image rendering PR for vello_hybrid, but I guess not much we can do for now.

sparse_strips/vello_dev_macros/src/test.rs
AndrewJakubowicz: Once I've implemented feature parity in fearless simd, I'd like us to test both scalar and SIMD. What do you think?

LaurenzV: Sure that's fine, just was curious!

sparse_strips/vello_sparse_tests/tests/image.rs
ajakubowicz-canva: Great callout! @grebmeg would you like me to hold this PR back until images has landed, or are you ok with the rebase and replacing `load_image` with `load_image!`? (I think it should be a syntactic only change).

ajakubowicz-canva:
Ooops, this PR has revealed that there is a bug in fearless_simd in the WASM architecture. fearless_simd on WASM should only add SIMD instructions if the target is compiled with the `+simd128` target feature. It looks like I stuffed something up because the test binary contains lots of simd instructions even though the `fallback` level is expected, and the WASM is not being built with `+simd128`.

sparse_strips/vello_sparse_tests/tests/image.rs
grebmeg: @AndrewJakubowicz No need to wait on me — feel free to go ahead and get this landed.

ajakubowicz-canva:
After further investigation, this PR is safe to land. The SIMD in the WASM binary would not be encountered during the tests, and is instead instructions included due to `target_feature` declarations. That is resolved in: https://github.com/raphlinus/fearless_simd/pull/16

Issue #1066, opened 2025-06-24
Negative dash offsets are rendered wrong


sagudev:
Bigger the absolute value => bigger mistake.

Expected:
![Expected](https://github.com/user-attachments/assets/87a01e75-06e2-4d6f-b091-40b74baec06d)
Actual:
![Actual](https://github.com/user-attachments/assets/d57ee48b-42e8-4f85-82cb-331d6e9afda8)

Code:
```rust
let stroke = vello::kurbo::Stroke::new(1.0).with_dashes(-30., [4., 2.]);
    scene.stroke(
        &stroke,
        Affine::IDENTITY,
        vello::peniko::Color::from_rgb8(0, 255, 0),
        None,
        &Rect::from_origin_size(Point::new(1., 1.), Size::new(28., 28.)),
    );
```

sagudev:
I think it is possible to hack around this, by taking cycles into account and set cyclically equivalent (positive) offset.

sagudev:
Given that vello splits dashing into more path shapes: https://github.com/linebender/vello/blob/d9d3586652ac1f340fa92b4b4e39a60c6a77e4db/vello/src/scene.rs#L264

this is probably kurbo bug.

sagudev:
> this is probably kurbo bug

It is, fix is https://github.com/linebender/kurbo/pull/454

PR #1067, opened 2025-06-24
Add test for negative dash offset

sagudev:
repro for https://github.com/linebender/vello/issues/1066

### DIFF
--- a/vello_tests/tests/regression.rs
+++ b/vello_tests/tests/regression.rs
@@ -106,5 +106,9 @@ fn test_negative_dash_offset() {
     params.anti_aliasing = AaConfig::Area;
     smoke_snapshot_test_sync(scene, &params)
         .unwrap()
-        .assert_mean_less_than(0.001);
+        .assert_mean_less_than(if cfg!(target_os = "macos") {
+            0.005
+        } else {
+            0.001
+        });
 }


Hywan:
Now https://github.com/linebender/kurbo/pull/454 is merged, can we merge this one?

sagudev:
The test is currently set for old kurbo, so it expects a panic. When we will have updated kurbo I will make this into regression test.

tomcur (dismissed):
LGTM (once the test passes)!

vello_tests/tests/regression.rs
```diff
@@ -89,3 +89,26 @@ fn text_stroke_width_zero() {
...
+    );
+    let mut params = TestParams::new("negative_dash_offset", 30, 30);
+    params.anti_aliasing = AaConfig::Area;
+    smoke_snapshot_test_sync(scene, &params)
+        .unwrap()
+        .assert_mean_less_than(if cfg!(target_os = "macos") {
+            0.005
+        } else {
+            0.001
+        });
```
tomcur, line 113: Do you know why there is a difference between macOS and other systems here? Is it because of FastMath, or is it because macOS is tested on the GPU in CI?

sagudev: IDK, I do not own mac system to check this.

PR #1068, opened 2025-06-25
Use LUT for rendering gradients

### LINKED ISSUE #1058: Use LUT for gradients
Currently, one of the main bottlenecks of gradient rendering is that for each pixel we are rendering, we need to search the corresponding gradient range the pixel belongs to, and then do the interpolation. This is the technique Skia uses, and while it works decently well, it's still pretty slow.

What Blend2D does instead is that it precomputes a lookup table of a specific size (256, 512, 1024) depending on the number of stops in the gradient, and then when calculating the actual pixel values, we just need to compute the parametric `t` value and then do a lookup in the lookup table. While this might lead to banding/artifacts for very large gradients, I think that this should be more than good enough for the general case and this should be much faster, so I would definitely like to try this out.

The only problem I see is how to handle this in `vello_hybrid`, as we probably don't want to leave the computation of the lookup table to the CPU. Perhaps the computation of such a table can be done similarly on the GPU? Alternatively, we could use different techniques on CPU and GPU, but this will probably make testing harder as the results might be slightly different.

LaurenzV:
This reduces the frame time of rendering a linear gradient to 4k from ~17.5ms to ~5.7ms. :)

Fixes #1058.


sparse_strips/vello_common/src/encode.rs
```diff
@@ -717,6 +723,20 @@ pub struct EncodedGradient {
...
     /// Whether the gradient requires `source_over` compositing.
     pub has_opacities: bool,
+    u8_lut: OnceCell<GradientLut<u8>>,
+    f32_lut: OnceCell<GradientLut<f32>>,
+}
+
+impl EncodedGradient {
+    /// Get the lookup table for sampling u8-based gradient values.
+    pub fn u8_lut(&self) -> &GradientLut<u8> {
+        self.u8_lut.get_or_init(|| GradientLut::new(&self.ranges))
```
DJMcNab, line 733: This doesn't seem right to me; since `ranges` is `pub`, this could be invalidated by any caller, leading to a stale LUT being used.

LaurenzV: Hmm yeah, good point, we should probably make that private.

LaurenzV:
Since this unfortunately requires regenerating most test cases, if there are concerns about the sampling resolution, now is the time to bring them up. :p

LaurenzV:
I think this will require some more thinking on how to make this SIMD-compatible, so let's not merge this for now.

LaurenzV:
Will include this in my SIMD PR.

PR #1071, opened 2025-06-27
fix: sparse_strips scalar WASM build containing SIMD instructions

ajakubowicz-canva:
### Context

This is a bit of a subtle fix that follows https://github.com/raphlinus/fearless_simd/pull/16

The important excerpts from the WebAssembly spec.

> "[Modules](https://www.w3.org/TR/wasm-core-2/index.html#syntax-module) are valid when all the components they contain are valid."
> ~[link](https://www.w3.org/TR/wasm-core-2/#modules%E2%91%A2)

Thus, for a browser (or WebAssembly runtime) that doesn't support SIMD, the module validation check would fail because:
 - `v128` type would be invalid and wouldn't exist in that WebAssembly context.
 - `expr` must also be valid, so instructions over f32x4 would fail validation.

###  Why

Prior to the fearless_simd fix, it was possible via `target_feature` to accidentally leak SIMD instructions into the binary. Even if these instructions are never executed, their mere presence would result in the WASM binary failing to validate.

### Changes

This PR was implemented in a test-driven style.

First I implemented the test:
 - The test only runs if the `simd128` target feature is not requested in the rustc flags.
 - The test self-introspects the test wasm binary and validates the binary against a validator that doesn't contain SIMD support.

This test failed (because of the leaking SIMD instructions). Hence, on a browser that didn't support SIMD we would crash. (Not good for consumers of vello using WASM asking for a scalar binary).

Example test failure:

Screenshot 2025-06-27 at 3 21 59 pm

Then I bumped up fearless_simd, and the test passes!

### Test plan

This PR is effectively a test to prevent regressions.

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/wasm_binary_invariants.rs
+++ b/sparse_strips/vello_sparse_tests/tests/wasm_binary_invariants.rs
@@ -10,7 +10,7 @@ async fn no_simd_instruction_inclusion() {
     // is imperative that there isn't a single SIMD instruction in the resulting binary. These can
     // accidentally creep into the binary due to usage of `#![cfg(target_feature = "simd128")]`. Any
     // inclusion of a SIMD instruction in a non-SIMD WASM binary can invalidate the whole binary for
-    // browsers that do not have SIMD support.
+    // browsers (or WebAssembly runtimes) that do not have SIMD support.
     //
     // This test runs when simd128 is not enabled, and self-introspects the binary to ensure no SIMD
     // instructions are included.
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_sparse_tests/tests/wasm_binary_invariants.rs
```diff
@@ -0,0 +1,41 @@
...
+    //
+    // This test runs when simd128 is not enabled, and self-introspects the binary to ensure no SIMD
+    // instructions are included.
+
+    use wasm_bindgen::JsCast;
+    use wasm_bindgen_futures::JsFuture;
+    use wasmparser::{Validator, WasmFeatures};
+
+    let window = web_sys::window().unwrap();
+    let url = "/wasm-bindgen-test_bg.wasm";
```
LaurenzV, line 23: Where is tha URL coming from?

LaurenzV (approved):
Seems OK to me.

sparse_strips/vello_sparse_tests/tests/wasm_binary_invariants.rs
ajakubowicz-canva: That's a great question. It's the module name + `_bg.wasm`. (Code search: https://github.com/search?q=repo%3Arustwasm%2Fwasm-bindgen+%22_bg.wasm%22&type=code).

That `wasm-bindgen-test` module name comes from [`wasm_bindgen_test_runner`](https://github.com/rustwasm/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs#L183). It looks like it's the hard coded module name.

PR #1073, opened 2025-06-30
Fix `unused_mut` lint

waywardmonkeys:
This starts showing up in Rust 1.88.


waywardmonkeys:
Is this a breaking change?

LaurenzV (approved):

LaurenzV:
I believe it's not breaking, as it only means that the variable _within_ the function is mutable.

PR #1074, opened 2025-06-30
Fix `clippy::uninlined_format_args` lints

waywardmonkeys:


LaurenzV:
Perhaps we should wait until #1064 landed to avoid merge conflicts there.

waywardmonkeys:
Sure. This is blocking updating the stable Rust CI to 1.88.

nicoburns (approved):
Will defer to others on merge order, but the changes themselves look good.

LaurenzV:
I will have my SIMD PR up soon with many changes, so maybe we can wait until then.

DJMcNab:
I think just getting something landed here is worthwhile; it's only a small PR, so rebasing past it won't be too bad.
There'll always be something coming down the pipeline

waywardmonkeys:
I am going to rebase this and land it within 2 hours.

PR #1076, opened 2025-07-02
[vello_cpu]: add WASM vello_cpu examples

ajakubowicz-canva:
### Context

To help with performance profiling and inspecting vello_cpu example renders on the browser, this PR adds a vello_cpu examples.

Run with `cargo run_wasm -p wasm_cpu --release`

### Scenes added

We don't yet have a rendering abstraction between `vello_cpu` and `vello_hybrid`, so the example scenes are pretty similar, but not quite identical. I simplified each one to use `RenderContext` instead of `Scene`, and also made them very specific to running on the browser.

I added:
 - simple
 - tiger svg
 - cowboy image
 - text
 - clipping

No new assets are added in this PR - only existing assets in the repo are used. See the screen recording to see the scenes added.

### Test plan

> [!Note]
> Passing the +simd128 rustc flags spams the console with `(ignoring feature)` logs. This is because the run-wasm tool is built for your system architecture with the same global flags – and because it's not wasm it doesn't recognize the feature. Everything still works.

Test it by running it with `cargo run_wasm -p wasm_cpu --release` and test explicit SIMD by running `RUSTFLAGS=-Ctarget-feature=+simd128 cargo run_wasm -p wasm_cpu --release`.

### Screen recording

https://github.com/user-attachments/assets/80867362-bce5-4a9e-b303-88b2c9233972

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4116,6 +4116,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "wasm_cpu"
+version = "0.5.0"
+dependencies = [
+ "bytemuck",
+ "console_error_panic_hook",
+ "console_log",
+ "js-sys",
+ "log",
+ "parley",
+ "vello_common",
+ "vello_cpu",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "wasm-bindgen-test",
+ "web-sys",
+ "wgpu",
+]
+
+[[package]]
 name = "wasmparser"
 version = "0.214.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


LaurenzV (approved):
I've played around with and encountered no issues, it seems to work fine, thanks! I've only skimmed the actual code and not thought about it deeply, but since most of it is ported from existing code, it should be fine!

DJMcNab:
Yeah - `cargo run_wasm` is insufficiently flexible a lot of the time. I don't really have good solutions here at the moment.

taj-p (approved):
LGTM! I think once we have our rendering abstraction we can consider combining these examples

PR #1077, opened 2025-07-02
Fix(vello_hybrid): Exceeded WebGPU buffer size limit

grebmeg:
This is a WebGPU buffer size limit error. Here's what's happening:
- Current limit: WebGPU device is configured with a maximum buffer size of 256MB
- Attempted usage: The application is trying to use a buffer of 1GB
- Hardware capability: Graphics adapter actually supports up to 4GB

### DIFF
--- a/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
@@ -54,6 +54,7 @@ impl RendererWrapper {
                         // (8). Some devices (including CI) support only up to 4.
                         max_color_attachments: 4,
                         max_texture_dimension_2d: adapter.limits().max_texture_dimension_2d,
+                        max_buffer_size: adapter.limits().max_buffer_size,
                         ..wgpu::Limits::downlevel_webgl2_defaults()
                     },
                     ..Default::default()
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -54,6 +54,7 @@ impl RendererWrapper {
                         // (8). Some devices (including CI) support only up to 4.
                         max_color_attachments: 4,
                         max_texture_dimension_2d: adapter.limits().max_texture_dimension_2d,
+                        max_buffer_size: adapter.limits().max_buffer_size,
```
ajakubowicz-canva, line 57: I'm not completely clear, how does this resolve the issue?

ajakubowicz-canva (approved):
Didn't manually test, very nice small fix!

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
grebmeg: The original code uses WebGL2’s default `max_buffer_size`, which is 256 MB. However, the new code calculates a required buffer size for the atlas texture based on `max_texture_dimension_2d` that exceeds 256 MB, causing a failure. That said, the adapter actually supports a higher limit of 4 GB, so we can safely use that here.
I think a similar approach is already applied in the line above for `max_texture_dimension_2d`.

ajakubowicz-canva: Ah I understand! Gotcha!

PR #1078, opened 2025-07-02
Rewrite `vello_cpu` to be more SIMD-friendly

LaurenzV:
This is a first attempt at rewriting `vello_cpu` to be more SIMD-friendly. 

As can be seen in the code, there are still quite a few TODOs left. The usage of SIMD types is also not very consistent (in some cases we are using f32x16, in some cases we are limited to f32x4), but it should be a good enough basis for any platform that provides 128-bit SIMD capabilities, which I think is a good start.

This PR should fix the performance regression that was introduced when adding the f32 pipeline. There are certain parts of the pipeline (namely opaque filling and alpha compositing) which only marginally improve as part of this PR on NEON because it already auto-vectorized very well before, but other parts like gradients, blurred rects and images (especially images with bilinear and bicubic filtering) are _much_ faster now.

It also incorporates the gradient LUT change from a previous PR.

I've adjusted a gradient test case so that it doesn't trigger the floating point impreciseness we have already had before.


LaurenzV:
@ajakubowicz-canva Any idea what the WASM issue is about?

LaurenzV:
@nicoburns if you feel like trying, with this branch performance should be back to normal like before f32 was introduced, and hopefully even faster. If not, then something is definitely fishy.

LaurenzV:
(Assuming you have a device that supports NEON).

nicoburns:
@LaurenzV It seems that there was something fishy going on. And it was that I was still multiplying the alpha input to `push_layer` by `255` even though it's `f32` not `u8` now. For some reason that causes significant slowdown (on all of this branch, main, and 0.0.1) even though it successfully renders. I'm seeing very good results with this branch (but also decent with `main` now that I've fixed the alpha issue).

See: [#vello > vello_cpu performance @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/vello_cpu.20performance/near/526825491) for more and some numbers from Blitz with this branch.

nicoburns:
@LaurenzV any chance that you could rebase this on top of the Send+Sync fix?

LaurenzV:
@nicoburns Done, also added the `Clip` fix (haven't validated it yet, though).

ajakubowicz-canva:
@LaurenzV Regarding the [WASM error in CI](https://github.com/linebender/vello/actions/runs/16033534236/job/45239701907?pr=1078#step:11:34), I've reproduced it locally.
It looks like wasmparser has a [function locals limit](https://github.com/bytecodealliance/wasm-tools/blob/b5c3d98e40590512a3b12470ef358d5c7b983b15/crates/wasmparser/src/limits.rs#L29).

This is apparently a limit in browsers as well?

One possible fix is to add optimizations or `--release` to the test binary. This resolves the issue because the number of locals drops in the offending function. Another possible fix is to split the function that is tripping this issue.


I did some digging and it looks like the offending function is `<_ZN9vello_cpu4fine17Fine$LT$S$C$T$GT$7run_cmd17hc0f5ddbe307cb113E>` – with 59,555 locals.

To figure that out I used `wasm-objdump -d <path_to_test_wasm>`, saving the result to a temp file. Then ran a script to count the number of locals between functions. (Note first number is the total number of locals).

```
59555 locals: 12e439e func[40867] <_ZN9vello_cpu4fine17Fine$LT$S$C$T$GT$7run_cmd17hc0f5ddbe307cb113E>:
33522 locals: 1115375 func[40515] <_ZN120_$LT$vello_cpu..fine..common..image..FilteredImagePainter$LT$S$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$4next17h70dc3fee6f36c1ecE>:
27529 locals: 104b0ed func[40484] <_ZN104_$LT$vello_cpu..fine..common..gradient..GradientPainter$LT$S$GT$$u20$as$u20$vello_cpu..fine..Painter$GT$8paint_u817h61acd92ec3c6cccfE>:
26339 locals: 13a66b1 func[40870] <_ZN9vello_cpu4fine17Fine$LT$S$C$T$GT$7run_cmd17he9eceec63ec2557dE>:
26298 locals: f07a52 func[38737] <_ZN9vello_cpu4fine4lowp7compose4Plus7compose17h5930bdb72cdb25a0E>:
21502 locals: ec7b03 func[38734] <_ZN9vello_cpu4fine4lowp7compose8DestAtop7compose17ha51e5d3bdc0768a0E>:
```

Running the same test on the `--release` binary gives:

```
131 locals: 33b541 func[8947] <_ZN9vello_cpu4fine17Fine$LT$S$C$T$GT$7run_cmd17hd204ef72cf5b1012E>:
109 locals: 3b3464 func[9735] <_ZN6skrifa7outline4glyf4hint6engine8dispatch61_$LT$impl$u20$skrifa..outline..glyf..hint..engine..Engine$GT$3run17habda427752f54333E.llvm.1517942214237201949>:
98 locals: 35f64b func[9213] <_ZN88_$LT$vello_cpu..fine..lowp..U8Kernel$u20$as$u20$vello_cpu..fine..FineKernel$LT$S$GT$$GT$22alpha_composite_buffer17h193da8f92a13bf92E>:
98 locals: 32e8bc func[8862] <_ZN118_$LT$vello_cpu..fine..lowp..image..BilinearImagePainter$LT$S$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$4next17h30b8723ea0fa6970E>:
97 locals: 33f550 func[8957] <_ZN9vello_cpu4fine4lowp4fill15alpha_composite17h5573303672b72203E>:
97 locals: 336980 func[8940] <_ZN104_$LT$vello_cpu..fine..common..gradient..GradientPainter$LT$S$GT$$u20$as$u20$vello_cpu..fine..Painter$GT$8paint_u817hd53e10e32909004fE>:
96 locals: 35fd6c func[9214] <_ZN9vello_cpu4fine4lowp3mix17h2c9bb87567627f07E>:
94 locals: 32a20e func[8852] <_ZN9vello_cpu4fine4lowp7compose7SrcAtop7compose17hbcc2a2287c1cc456E>:
93 locals: 32b5fa func[8854] <_ZN9vello_cpu4fine4lowp7compose4Plus7compose17hcccbb1890845449aE>:
93 locals: 32ac14 func[8853] <_ZN9vello_cpu4fine4lowp7compose8DestAtop7compose17hafc9cff29ac7eed0E>:
93 locals: 327f5a func[8847] <_ZN9vello_cpu4fine4lowp7compose3Xor7compose17hd6716a092601259cE>:
93 locals: 3273ac func[8845] <_ZN9vello_cpu4fine4lowp7compose8DestOver7compose17h37e05c34a078b958E>:
```

We go from `59555` locals to `131` locals in the largest function.

I defer the solution to you, but I think we probably just want to add some optimization config to the build instead of finding and splitting the function.

LaurenzV:
Oh wow that's a lot of locals... Yeah I think just using `release` is the best option here. Should I include this in my PR here or do you want to create a new, separate one for it?

ajakubowicz-canva:
I think it's fine to include here to unblock CI.

ajakubowicz-canva:
I tested this with `RUSTFLAGS=-Ctarget-feature=+simd128 cargo run_wasm -p wasm_cpu --release` and compared it with the non SIMD build against **main**.

Timings of a frame of the cowboy emoji zoomed in:
 - main branch: no-simd: ~180ms.
 - main branch: simd: ~175ms.
 - this PR: no-simd: ~ 76ms.
 - this PR: simd: ~ 76ms.


Because the measurements are not scientific, I can't differentiate between the SIMD and non SIMD WASM builds. However, there is a huge improvement in image rendering compared to main!

LaurenzV:
This is with all of the SIMD operations necessary for image rendering already implemented?

ajakubowicz-canva:
I don't know, maybe `Level::fallback()` was used, but I ran `RUSTFLAGS=-Ctarget-feature=+simd128 cargo run_wasm -p wasm_cpu --release` on this PR so I assumed that this PR must only be using the implemented WASM SIMD operations. I may have made a mistake or the Level might have been incorrectly set.

The CI for Vello does run SIMD WASM tests, so I think the subset of operations is not hitting any `todo!()` or `unimplemented!()` parts of fearless simd!

LaurenzV:
> I don't know, maybe Level::fallback() was used

Oh yes, for fine rasterization I did not implement anything for WASM and let it fallback. I just thought maybe you already implemented it, that's why I wasn't sure. So this means that we will hopefully see even better gains once it's implemented. :)

Cargo.toml
```diff
@@ -115,7 +115,7 @@ rayon = { version = "1.10.0" }
 thread_local = "1.1.8"
 crossbeam-channel = "0.5.15"
 ordered-channel = { version = "1.2.0", features = ["crossbeam-channel"] }
-fearless_simd = { git = "https://github.com/raphlinus/fearless_simd", rev = "8c22f7c", default-features = false }
+fearless_simd = { git = "https://github.com/raphlinus/fearless_simd", rev = "8a1cea7", default-features = false }
```
ajakubowicz-canva: https://github.com/raphlinus/fearless_simd/pull/25 can be merged and commit can be on main branch.

LaurenzV: Sorry, I missed that yesterday!

.github/workflows/ci.yml
```diff
@@ -356,11 +356,11 @@ jobs:
         working-directory: sparse_strips/vello_hybrid/examples/native_webgl
 
       - name: Run vello_sparse_tests on Chrome
-        run: wasm-pack test --headless --chrome --features webgl
+        run: wasm-pack test --headless --chrome --features webgl --release
```
ajakubowicz-canva, line 361: Should a comment be added to describe why `--release` is being used?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -742,76 +701,6 @@ pub struct GradientRange {
...
-    }
-
-    fn is_defined(&self, pos: Point) -> bool {
-        self.pos_inner(pos).is_some()
-    }
-}
-
 /// An encoded blurred, rounded rectangle.
 #[derive(Debug)]
 pub struct EncodedBlurredRoundedRectangle {
```
ajakubowicz-canva, line 706: While reviewing this PR I found this unrelated bug in `BlurredRoundedRectangle:: encode_into`:

```
            if self.rect.y0 > self.rect.y1 {
                core::mem::swap(&mut rect.x0, &mut rect.x1);
            }
```

I think it should be swapping the `y` fields.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -12,8 +12,10 @@ use crate::paint::{Image, ImageSource, IndexedPaint, Paint, PremulColor};
 use crate::peniko::{ColorStop, Extend, Gradient, GradientKind, ImageQuality};
 use alloc::borrow::Cow;
 use alloc::vec::Vec;
-use core::f32::consts::PI;
-use core::iter;
+#[cfg(not(feature = "multithreading"))]
+use core::cell::OnceCell;
+#[cfg(feature = "multithreading")]
+use once_cell::sync::OnceCell;
```
ajakubowicz-canva, line 18: Why can't `core::cell::OnceCell` be used for both?

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -124,12 +127,44 @@ impl Dispatcher for SingleThreadedDispatcher {
...
+                    );
+                }
+                _ => self.rasterize_with::<Fallback, F32Kernel>(
+                    Fallback::new(),
+                    buffer,
+                    width,
+                    height,
+                    encoded_paints,
+                ),
+            },
```
ajakubowicz-canva: Nit: Could WASM be added now 😄 ?

(Otherwise I can do it in a followup)

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+}
+
+#[inline(always)]
+pub(crate) fn extend<S: Simd>(val: f32x8<S>, pad: bool) -> f32x8<S> {
+    if pad {
+        val.max(0.0).min(1.0)
+    } else {
+        (val - val.floor()).fract()
```
ajakubowicz-canva, line 165: Can this be simplified to `val.fract()` or am I missing something?

sparse_strips/vello_cpu/src/fine/common/image.rs
```diff
@@ -0,0 +1,543 @@
...
+        // locations (-0.5, -0.5) and (0.5, 0.5), and weight them by the fractional
+        // x/y position using simple linear interpolation in both dimensions.
+        // In bicubic filtering, we instead span a 4x4 grid around the
+        // center of the location we are sampling, and sample those points
+        // using a cubic filter to weight each location's contribution.
+
+        #[inline(always)]
+        fn fract<S: Simd>(val: f32x4<S>) -> f32x4<S> {
+            val - val.floor()
+        }
```
ajakubowicz-canva, line 201: Is there a `fract` method implemented on `f32x4`?

sparse_strips/vello_cpu/src/fine/common/image.rs
```diff
@@ -0,0 +1,543 @@
...
+    x_positions: f32x4<S>,
+    y_positions: f32x4<S>,
+) -> u8x16<S> {
+    let idx = x_positions.cvt_u32() + y_positions.cvt_u32() * data.width_u32;
+
+    macro_rules! sample {
+        ($idx:expr) => {
+            data.pixmap.sample_idx(idx[$idx]).to_u32()
+        };
+    }
```
ajakubowicz-canva: Nit: Just curious if it's advantageous to use macros to reduce some repetitive typing? I do admit that it does make the line below much simpler, however it does slightly raise the complexity.

sparse_strips/vello_cpu/src/fine/highp/compose.rs
```diff
@@ -0,0 +1,154 @@
...
+            Compose::SrcIn => SrcIn::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::DestIn => DestIn::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::SrcOut => SrcOut::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::DestOut => DestOut::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::SrcAtop => SrcAtop::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::DestAtop => DestAtop::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::Xor => Xor::compose(simd, src_c, bg_c, alpha_mask),
+            Compose::Plus => Plus::compose(simd, src_c, bg_c, alpha_mask),
+            // Have not been able to find a formula for this, so just fallback to Plus.
+            Compose::PlusLighter => SrcOver::compose(simd, src_c, bg_c, alpha_mask),
```
ajakubowicz-canva: Should this be:

```suggestion
            Compose::PlusLighter => Plus::compose(simd, src_c, bg_c, alpha_mask),
```

sparse_strips/vello_cpu/src/fine/highp/mod.rs
ajakubowicz-canva, line 1: Instead of manually declaring inline always or inline never – is it worth considering if there is some way ["Profile Guided Optimization"](https://doc.rust-lang.org/rustc/profile-guided-optimization.html) can help?

sparse_strips/vello_cpu/src/util.rs
```diff
@@ -56,3 +66,89 @@ pub(crate) mod scalar {
...
+
+impl<S: Simd> NormalizedMulExt for u8x32<S> {
+    #[inline(always)]
+    fn normalized_mul(self, other: Self) -> Self {
+        let divided = (self.simd.widen_u8x32(self) * other.simd.widen_u8x32(other)).div_255();
+        self.simd.narrow_u16x32(divided)
+    }
+}
+
+pub(crate) trait Div255Ext {
```
ajakubowicz-canva, line 82: Could this be documented?
IIUC this trait is handling the optimisation that `x / 255` is equivalent to `(x >> 8)`?

sparse_strips/vello_dev_macros/src/bench.rs
```diff
@@ -35,15 +37,40 @@ pub(crate) fn vello_bench_inner(_: TokenStream, item: TokenStream) -> TokenStrea
...
+            //     run_float(b, vello_common::fearless_simd::Fallback::new());
+            // });
+
+            // Uncomment this to enable f32_neon benchmarks.
+            // #[cfg(target_arch = "aarch64")]
+            // if let Some(neon) = Level::new().as_neon() {
+            //     c.bench_function(&get_bench_name(&#input_fn_name_str, "f32_neon"), |b| {
+            //         run_float(b, neon);
+            //     });
+            // }
```
ajakubowicz-canva, line 73: Why not rely on the conditional compilation?

sparse_strips/vello_dev_macros/src/lib.rs
```diff
@@ -15,7 +15,7 @@ mod test;
 // means that the absolute difference of each component of a pixel must not be higher
 // than 1. For example, if the target pixel is (233, 43, 64, 100), then permissible
 // values are (232, 43, 65, 101) or (233, 42, 64, 100), but not (231, 43, 64, 100).
-const DEFAULT_CPU_U8_TOLERANCE: u8 = 1;
+const DEFAULT_CPU_U8_TOLERANCE: u8 = 2;
```
ajakubowicz-canva: Is it worth documenting or elaborating on this tolerance change?

sparse_strips/vello_sparse_tests/snapshots/gradient_linear_spread_method_pad.png
ajakubowicz-canva, line 1: haha this threw me through a loop - but I can see that you changed this test.

sparse_strips/vello_sparse_tests/snapshots/gradient_linear_with_y_repeat.png
ajakubowicz-canva, line 1: I assume the differences in this image are expected due to precision?

sparse_strips/vello_sparse_tests/tests/mix.rs
```diff
@@ -89,7 +89,7 @@ fn mix_lighten(ctx: &mut impl Renderer) {
     mix(ctx, BlendMode::new(Mix::Lighten, Compose::SrcOver));
 }
 
-#[vello_test(cpu_u8_tolerance = 1)]
+#[vello_test(cpu_u8_tolerance = 4)]
```
ajakubowicz-canva, line 92: Is this increase in tolerance due to some documented reason?

sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -412,399 +607,203 @@ impl<F: FineType> Fine<F> {
...
-}
-
-impl Widened<Self> for f32 {
+impl<S: Simd> PosExt<S> for f32x4<S> {
     #[inline(always)]
-    fn clamp(self) -> Self {
-        Self::clamp(self, Self::ZERO, Self::ONE)
-    }
+    fn splat_pos(simd: S, pos: f32, _: f32, y_advance: f32) -> Self {
+        let column_mask: Self = [0.0, 1.0, 2.0, 3.0].simd_into(simd);
```
ajakubowicz-canva: Can this be made more explicit that this depends on a tile height of 4? E.g. could this even use the tile height constant somehow?

Cargo.toml
```diff
@@ -115,7 +115,7 @@ rayon = { version = "1.10.0" }
 thread_local = "1.1.8"
 crossbeam-channel = "0.5.15"
 ordered-channel = { version = "1.2.0", features = ["crossbeam-channel"] }
-fearless_simd = { git = "https://github.com/raphlinus/fearless_simd", rev = "8c22f7c", default-features = false }
+fearless_simd = { git = "https://github.com/raphlinus/fearless_simd", rev = "42e01d2", default-features = false }
```
ajakubowicz-canva: Nit: Update to point to the new linebender repo: https://github.com/linebender/fearless_simd

sparse_strips/vello_cpu/src/fine/common/gradient/sweep.rs
```diff
@@ -0,0 +1,59 @@
...
+    let a = f32x8::splat(simd, 2.476_102e-2).madd(f32x8::splat(simd, -7.054_738_2e-3), s);
+    let b = f32x8::splat(simd, -5.185_397e-2).madd(a, s);
+    let c = f32x8::splat(simd, 0.159_121_17).madd(b, s);
+
+    let mut phi = slope * c;
+
+    phi = simd.select_f32x8(simd.simd_lt_f32x8(x_abs, y_abs), c2 - phi, phi);
+    phi = simd.select_f32x8(simd.simd_lt_f32x8(x, c0), c3 - phi, phi);
+    phi = simd.select_f32x8(simd.simd_lt_f32x8(y, c0), c1 - phi, phi);
+    phi = simd.select_f32x8(phi.simd_eq(phi), phi, c0);
```
ajakubowicz-canva, line 57: haha what is `phi.simd_eq(phi)`? Is this using the fact that NaN != NaN to remove NaN? Worth commenting?

LaurenzV:
No worries, wasn't expecting you to take a detailed look at everything! I understand it's hard to review, but splitting this up into multiple smaller PRs will be close to impossible I think. Will be addressing your comments and then merge.

sparse_strips/vello_common/src/encode.rs
LaurenzV: Because `core::cell::OnceCell` doesn't implement `Sync`, which is required for multi-threading. And `OnceLock` is only available in `std`. So either we do this, or we don't require `once_cell` but then with one feature we use `OnceCell`, while with the other feature we use `OnceLock`. Not sure what's better.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Sure, let me try to add it.

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
LaurenzV: 🤔 I think you are right. 😄

LaurenzV: Ah, nevermind, no you can't because otherwise the value van be negative as well.

sparse_strips/vello_cpu/src/fine/highp/compose.rs
LaurenzV: Yeah, good catch. Since I haven't really found anything on that blend mode, I haven't added any test for it, that's why I didnt catch it.

sparse_strips/vello_sparse_tests/snapshots/gradient_linear_with_y_repeat.png
LaurenzV: Yeah.

sparse_strips/vello_sparse_tests/tests/mix.rs
LaurenzV: There is documentation for another method, but not for this one in particular. Unfortunately for blending, it just can happen in certain circumstances that specific pixels have a much larger divergence in the low-precision pipeline.

sparse_strips/vello_cpu/src/fine/common/image.rs
LaurenzV: This is not exactly the same, because it has different behavior for negative numbers. Will add a comment.

sparse_strips/vello_cpu/src/fine/highp/mod.rs
LaurenzV: Might be worth exploring in the future!

sparse_strips/vello_cpu/src/util.rs
LaurenzV: It is documented further up.

sparse_strips/vello_dev_macros/src/bench.rs
LaurenzV: I mainly did it because performance-wise, we only really care about the u8 SIMD version. It was a bit annoying to always manually comment it out when running the benchmarks, so I just left it commented out. I could change it back, but I doubt that anyone would want to run benchmarks for the other versions.

sparse_strips/vello_dev_macros/src/lib.rs
LaurenzV: If I change it back to 1 those are the failures I get:

image

So it's mainly because the u8 version of bilinear rendering works differently, so there's a bigger divergence. I can add a comment for that.

sparse_strips/vello_common/src/encode.rs
nicoburns: I think using std::OnceLock is better. It's the same thing as the sync OnceCell from the once_cell crate but avoids bringing in a dependency.

You could import it with an aliases name to avoid having to conditionally compile other code.

PR #1084, opened 2025-07-03
[vello_hybrid]: add example for image destruction

grebmeg:
This is a follow-up to [the comment](https://github.com/linebender/vello/pull/1064#discussion_r2178903924) in #1064.

Extended the examples to include support for destroying images:

https://github.com/user-attachments/assets/36a1ed37-dd08-4c1f-b83f-9528bf63c87b

### DIFF
--- a/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
@@ -11,8 +11,11 @@
 
 use std::cell::RefCell;
 use std::rc::Rc;
-use vello_common::kurbo::{Affine, Vec2};
-use vello_hybrid_scenes::AnyScene;
+use vello_common::{
+    kurbo::{Affine, Vec2},
+    paint::ImageId,
+};
+use vello_hybrid_scenes::{AnyScene, SceneResources};
 use wasm_bindgen::prelude::*;
 use web_sys::{Event, HtmlCanvasElement, KeyboardEvent, MouseEvent, WheelEvent};
 
@@ -41,6 +44,8 @@ struct AppState {
     renderer_wrapper: RendererWrapper,
     need_render: bool,
     canvas: HtmlCanvasElement,
+    scene_resources: SceneResources,
+    uploaded_images: Vec<ImageId>,
 }
 
 impl AppState {
@@ -62,6 +67,8 @@ impl AppState {
             renderer_wrapper,
             need_render: true,
             canvas,
+            scene_resources: SceneResources::new(),
+            uploaded_images: Vec::new(),
         };
 
         app_state.upload_images_to_atlas();
@@ -77,7 +84,11 @@ impl AppState {
         self.scene.reset();
 
         // Render the current scene with transform
-        self.scenes[self.current_scene].render(&mut self.scene, self.transform);
+        self.scenes[self.current_scene].render(
+            &mut self.scene,
+            self.transform,
+            &self.scene_resources,
+        );
 
         let render_size = vello_hybrid::RenderSize {
             width: self.width,
@@ -118,8 +129,9 @@ impl AppState {
         self.need_render = true;
     }
 
-    fn reset_transform(&mut self) {
+    fn reset_scene(&mut self) {
         self.transform = Affine::IDENTITY;
+        self.upload_images_to_atlas();
         self.need_render = true;
     }
 
@@ -180,14 +192,27 @@ impl AppState {
     fn upload_images_to_atlas(&mut self) {
         use vello_hybrid_scenes::image::ImageScene;
 
+        self.scene_resources.images.clear();
+        self.uploaded_images.clear();
+
         // 1st example — uploading pixmap directly to WebGL atlas
         let pixmap1 = ImageScene::read_flower_image();
-        self.renderer_wrapper.renderer.upload_image(&pixmap1);
+        let image_id1 = self.renderer_wrapper.renderer.upload_image(&pixmap1);
 
         // 2nd example — uploading from a WebGL texture
         let pixmap2 = ImageScene::read_cowboy_image();
         let texture2 = self.pixmap_to_webgl_texture(&pixmap2);
-        self.renderer_wrapper.renderer.upload_image(&texture2);
+        let image_id2 = self.renderer_wrapper.renderer.upload_image(&texture2);
+
+        self.scene_resources.images = vec![image_id1, image_id2];
+        self.uploaded_images = vec![image_id2, image_id1];
+    }
+
+    fn handle_destroy_image(&mut self) {
+        if let Some(image_id) = self.uploaded_images.pop() {
+            self.renderer_wrapper.renderer.destroy_image(image_id);
+        }
+        self.need_render = true;
     }
 
     /// Convert a pixmap to WebGL texture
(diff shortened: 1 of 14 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -546,9 +556,12 @@ impl Scheduler {
...
                 match scene.encoded_paints.get(paint_id) {
                     Some(EncodedPaint::Image(encoded_image)) => match &encoded_image.source {
-                        ImageSource::OpaqueId(_) => {
+                        ImageSource::OpaqueId(image_id) => {
                             let paint_type = 2_u32;
-                            let paint_packed = (paint_type << 30) | (paint_tex_id & 0x3FFFFFFF);
+                            let skip_paint = !image_cache.contains(*image_id) as u32;
+                            let paint_packed = (paint_type << 30)
+                                | (skip_paint << 29)
+                                | (paint_tex_id & 0x1FFFFFFF);
```
grebmeg, line 564: To ensure correctness, we must maintain stable references between uploaded images and their corresponding encoded paints. This means that a given uploaded image should always map to the same `paint_tex_id` and metadata block in `encoded_paints_texture`, regardless of when or how it is accessed.

The shader relies on `paint_tex_id` to fetch the associated `EncodedImage` metadata, which in turn assumes that the image data is available in the texture atlas at a specific offset and size. If an image is destroyed but its `ImageId` is still referenced in the scene (i.e., the image is cleaned without cleaning up its usage in the scene), the renderer may sample from invalid memory.

To mitigate this, `prepare_gpu_encoded_paints` resets the width, height, and offset fields to zero for destroyed images. Additionally, a `skip_paint` flag is introduced to explicitly prevent the shader from rasterizing these invalid entries.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -179,7 +178,12 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
     let paint_type = in.paint >> 30u;
 
     if paint_type == PAINT_TYPE_IMAGE {
-        let paint_tex_id = in.paint & 0x3FFFFFFF;
+        let paint_tex_id = in.paint & 0x1FFFFFFFu;
+        let skip_paint = in.paint >> 29u & 1u;
+        if skip_paint == 1u {
+            discard;
+        }
```
ajakubowicz-canva, line 185: Is it worth an inline comment here describing that skip_paint signals that an image cannot be sampled from?

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Technically we could put the onus on the caller that they must only use valid opaque ids and allow sampling from invalid memory. However, this is probably better unless there's some large performance benefit for pushing that concern up to the caller.

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -180,14 +192,27 @@ impl AppState {
...
+        self.scene_resources.images = vec![image_id1, image_id2];
+        self.uploaded_images = vec![image_id2, image_id1];
+    }
+
+    fn handle_destroy_image(&mut self) {
+        if let Some(image_id) = self.uploaded_images.pop() {
+            self.renderer_wrapper.renderer.destroy_image(image_id);
+        }
+        self.need_render = true;
     }
```
ajakubowicz-canva, line 216: Could a comment be added to mention that this method is intentionally destroying the image resource whilst maintaining the dangling image pointer in the scene? This is to demonstrate resilience to invalid image points being used. Ditto for this method in the other example.

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -396,7 +422,7 @@ pub async fn run_interactive(canvas_width: u16, canvas_height: u16) {
     let document = web_sys::window().unwrap().document().unwrap();
     let instructions = document.create_element("div").unwrap();
     instructions.set_inner_html(
-        "Left/Right Arrow: Change scene | Space: Reset view | Mouse Drag: Pan | Mouse Wheel: Zoom",
+        "Left/Right Arrow: Change scene | Space: Reset scene | D: Destroy image | Mouse Drag: Pan | Mouse Wheel: Zoom",
```
ajakubowicz-canva, line 425: I kind of feel like this "destroy image" functionality fits more into a test than an example. Is it worth adding this feature into a test and maybe even consider removing it from the examples?

The "easy" case for destroying images is to expect the caller to destroy a resource and then no longer use that opaque identifier. The edge case is testing that invalid pointers don't result in artifacts. That feels like a test case that would be valuable.

ajakubowicz-canva (commented):
Overall LGTM! The only actionable comment I have is that I think we really want to add this case into a test case instead of just tested manually via example.

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -80,6 +80,14 @@ impl ImageCache {
         self.slots.get(id.as_u32() as usize)?.as_ref()
     }
 
+    /// Check if an image is in the cache
+    pub(crate) fn contains(&self, id: ImageId) -> bool {
```
taj-p, line 84: I think if we wanted to make this more resilient we could add a generation to `ImageId`:

```rs
let id1 = renderer.upload_image(...);
renderer.destroy_image(id1);
let id2 = renderer.upload_image(...);
image_cache.contains(id1); // returns true because id2 uses the same u32 as id1 due to the free list
```

If we incremented a generation whenever we destroy an image, we'd be protected.

More practically, this would also protect from double destroys:

```rs
let id1 = renderer.upload_image(...);
renderer.destroy_image(id1);
let id2 = renderer.upload_image(...);
renderer.destroy_image(id1); // destroys id2
```

This also overcomes that comment in `schedule.rs`

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: > To mitigate this, prepare_gpu_encoded_paints resets the width, height, and offset fields to zero for destroyed images. Additionally, a skip_paint flag is introduced to explicitly prevent the shader from rasterizing these invalid entries.

Could we use generation in our `ImageId` instead?

sparse_strips/vello_hybrid/src/image_cache.rs
taj-p: If it's unclear, I am proposing `ImageId` include a `generation` field (maybe u32?) that is used in corresponding `get` / `contains` methods. For example (quick + rough code):

```rs
    pub fn get(&self, id: ImageId) -> Option<&ImageResource> {
        let index = id.0 as usize;
        if index < self.slots.len() {
            let info = self.slots[index].as_ref()?;
            if info.id.generation() == id.generation() {
                Some(info)
            } else {
                None
            }
        } else {
            None
        }
    }
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -546,9 +556,12 @@ impl Scheduler {
 
                 match scene.encoded_paints.get(paint_id) {
                     Some(EncodedPaint::Image(encoded_image)) => match &encoded_image.source {
-                        ImageSource::OpaqueId(_) => {
+                        ImageSource::OpaqueId(image_id) => {
                             let paint_type = 2_u32;
-                            let paint_packed = (paint_type << 30) | (paint_tex_id & 0x3FFFFFFF);
+                            let skip_paint = !image_cache.contains(*image_id) as u32;
+                            let paint_packed = (paint_type << 30)
+                                | (skip_paint << 29)
```
taj-p, line 563: I think this means that we will render something the consumer doesn't expect silently. I suspect at this point we know the consumer hasn't called our APIs correctly, but we're silently rendering their commands anyway - to create a result they probably don't want. I think we might want to propagate an error or panic instead.

PR #1086, opened 2025-07-04
feat(vello_hybrid): Add clipping support for image rendering

grebmeg:
This PR adds clipping support for image rendering in `vello_hybrid`, allowing images to be correctly clipped by multiple nested clip layers.

SCR-20250704-owuq

The x, y coordinates used for sampling depend on clip depth. If a strip is not in clipped space, it receives global strip view coordinates. Otherwise, it uses local slot coordinates, which results in incorrect image sampling — since sampling must occur relative to the global strip position. To address this, position data is now passed for all strips associated with image paints.

Additionally, the logic for determining whether to sample color from a solid rgba value, an image texture, or a slot texture has been refined. The flag used for this distinction was reduced from an 8-bit field to a 1-bit discriminator, enabling the `rgba_or_slot` field to be reused as a `payload` across different paint modes.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -46,7 +46,7 @@ pub struct GpuStrip {
     /// There are [`Config::strip_height`] alpha values per column.
     pub col_idx: u32,
     /// Color value or slot index when alpha is 0
-    pub rgba_or_slot: u32,
+    pub payload: u32,
     /// Packed paint type (2 bits) and paint texture id (30 bits)
     /// Paint type: 0 = solid, 1 = alpha, 2 = image
     /// Paint texture id locates the encoded image data `EncodedImage`
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -29,27 +29,26 @@ pub struct Config {
     pub alphas_tex_width_bits: u32,
 }
 
-/// Represents a GPU strip for rendering
+/// Represents a GPU strip for rendering.
+///
+/// This struct corresponds to the `StripInstance` struct in the shader.
+/// See the `StripInstance` documentation in `render_strips.wgsl` for detailed field descriptions.
 #[repr(C)]
 #[derive(Debug, Clone, Copy, Zeroable, Pod)]
 pub struct GpuStrip {
-    /// X coordinate of the strip
+    /// See `StripInstance::xy` documentation in `render_strips.wgsl`.
     pub x: u16,
-    /// Y coordinate of the strip
+    /// See `StripInstance::xy` documentation in `render_strips.wgsl`.
     pub y: u16,
-    /// Width of the strip
+    /// See `StripInstance::widths` documentation in `render_strips.wgsl`.
     pub width: u16,
-    /// Width of the portion where alpha blending should be applied.
+    /// See `StripInstance::widths` documentation in `render_strips.wgsl`.
     pub dense_width: u16,
-    /// Column-index into the alpha texture where this strip's alpha values begin.
-    ///
-    /// There are [`Config::strip_height`] alpha values per column.
+    /// See `StripInstance::col_idx` documentation in `render_strips.wgsl`.
     pub col_idx: u32,
-    /// Color value or slot index when alpha is 0
+    /// See `StripInstance::payload` documentation in `render_strips.wgsl`.
     pub payload: u32,
-    /// Packed paint type (2 bits) and paint texture id (30 bits)
-    /// Paint type: 0 = solid, 1 = alpha, 2 = image
-    /// Paint texture id locates the encoded image data `EncodedImage`
+    /// See `StripInstance::paint` documentation in `render_strips.wgsl`.
     pub paint: u32,
 }
 
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -567,9 +567,9 @@ impl Scheduler {
                                 ((scene_strip_y as u32) << 16) | (scene_strip_x as u32);
                             (scene_strip_xy, paint_packed)
                         }
-                        _ => unimplemented!("unsupported image source"),
+                        _ => unimplemented!("Unsupported image source"),
                     },
-                    _ => (0, 0),
+                    _ => unimplemented!("Unsupported paint type"),
                 }
             }
         }
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -389,7 +389,7 @@ impl Scheduler {
                     width: WideTile::WIDTH,
                     dense_width: 0,
                     col_idx: 0,
-                    rgba_or_slot: bg,
+                    payload: bg,
                     paint: 0,
                 });
             }
@@ -402,10 +402,12 @@ impl Scheduler {
                     let el = state.stack.last().unwrap();
                     let draw = self.draw_mut(el.round, clip_depth);
 
-                    let (col_idx, rgba_or_slot, paint) = Self::process_paint(&fill.paint, scene, 0);
+                    let (scene_strip_x, scene_strip_y) = (wide_tile_x + fill.x, wide_tile_y);
+                    let (payload, paint) =
+                        Self::process_paint(&fill.paint, scene, (scene_strip_x, scene_strip_y));
 
                     let (x, y) = if clip_depth == 1 {
-                        (wide_tile_x + fill.x, wide_tile_y)
+                        (scene_strip_x, scene_strip_y)
                     } else {
                         (fill.x, el.slot_ix as u16 * Tile::HEIGHT)
                     };
@@ -415,8 +417,8 @@ impl Scheduler {
                         y,
                         width: fill.width,
                         dense_width: 0,
-                        col_idx,
-                        rgba_or_slot,
+                        col_idx: 0,
+                        payload,
                         paint,
                     });
                 }
@@ -424,15 +426,19 @@ impl Scheduler {
                     let el = state.stack.last().unwrap();
                     let draw = self.draw_mut(el.round, clip_depth);
 
-                    let alpha_col = (alpha_fill.alpha_idx / usize::from(Tile::HEIGHT))
+                    let col_idx = (alpha_fill.alpha_idx / usize::from(Tile::HEIGHT))
                         .try_into()
                         .expect("Sparse strips are bound to u32 range");
 
-                    let (col_idx, rgba_or_slot, paint) =
-                        Self::process_paint(&alpha_fill.paint, scene, alpha_col);
+                    let (scene_strip_x, scene_strip_y) = (wide_tile_x + alpha_fill.x, wide_tile_y);
+                    let (payload, paint) = Self::process_paint(
+                        &alpha_fill.paint,
+                        scene,
+                        (scene_strip_x, scene_strip_y),
+                    );
 
                     let (x, y) = if clip_depth == 1 {
-                        (wide_tile_x + alpha_fill.x, wide_tile_y)
+                        (scene_strip_x, scene_strip_y)
                     } else {
                         (alpha_fill.x, el.slot_ix as u16 * Tile::HEIGHT)
                     };
@@ -443,7 +449,7 @@ impl Scheduler {
                         width: alpha_fill.width,
                         dense_width: alpha_fill.width,
                         col_idx,
-                        rgba_or_slot,
+                        payload,
                         paint,
                     });
                 }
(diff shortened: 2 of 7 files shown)


sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -46,7 +46,7 @@ pub struct GpuStrip {
     /// There are [`Config::strip_height`] alpha values per column.
     pub col_idx: u32,
     /// Color value or slot index when alpha is 0
```
taj-p: Do we need to update this comment?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -10,14 +10,19 @@
 // The alpha values are stored in a texture and sampled during fragment shading.
 // This approach optimizes memory usage by only storing alpha data where needed.
 //
-// The `StripInstance`'s `rgba_or_slot` field can either encode a color or a slot index.
+// The `StripInstance`'s `payload` field can either encode a color or a slot index.
```
taj-p: Do we need to mention paint in this comment?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -547,19 +560,28 @@ impl Scheduler {
...
+                                | (PAINT_TYPE_IMAGE << 29)
+                                | (paint_tex_id & 0x1FFFFFFF);
+                            let scene_strip_xy =
+                                ((scene_strip_y as u32) << 16) | (scene_strip_x as u32);
+                            (scene_strip_xy, paint_packed)
                         }
                         _ => unimplemented!("unsupported image source"),
                     },
-                    _ => (0, 0, 0),
+                    _ => (0, 0),
```
taj-p: Should we actually error here? Or are you tackling that in the other PR? It shouldn't be too hard since `schedule` already can return `RenderError`.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -40,8 +45,9 @@ struct Config {
 
 // Strip instance data
 //
-// `rgba_or_slot` field can either encode a color or a slot index.
-// If the alpha value is non-zero, the fragment shader samples the alpha texture.
+// `payload` field can either encode a color, [x, y] for image sampling or a slot index
+// If color source is payload and the paint type is solid, the fragment shader uses the color directly.
+// If color source is payload and the paint type is image, the fragment shader samples the image.
 // Otherwise, the fragment shader samples the source clip texture using the given slot index.
```
taj-p:
```suggestion
// The `paint` field is packed with metadata that controls how `payload` is interpreted:
//
// `paint` bit layout:
//   - Bit 31:     `color_source`      0 = use payload, 1 = use slot texture
//   - Bits 29-30: `paint_type`        0 = solid, 1 = image
//   - Bits 0-28:  `paint_texture_id`  if paint_type = PAINT_TYPE_IMAGE, index of `EncodedImage` 
//
// Decision tree for paint/payload interpretation:
//
// ```
// color_source = 0 (COLOR_SOURCE_PAYLOAD) - Use payload data directly
// ├── paint_type = 0 (PAINT_TYPE_SOLID) - Solid color rendering
// │   └── payload = [r, g, b, a] RGBA (packed as u8s)
// │
// └── paint_type = 1 (PAINT_TYPE_IMAGE) - Image rendering
//     └── payload = [x, y] scene coordinates (packed as u16s)
//
// color_source = 1 (COLOR_SOURCE_SLOT) - Use slot texture
// └── payload = slot_index (u32)
```

I wrote this out so that I could understand what's going on. What exists is fine but this might be an improvement over the text (I find the tree easier to understand).

Then instead of duplicating the information in the `payload` and `paint` fields, we can write `// See `StripInstance` documentation.`

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -50,9 +56,16 @@ struct StripInstance {
...
-    @location(3) rgba_or_slot: u32,
-    // Packed paint type (2 bits) and paint texture id (30 bits)
+    // Packed data:
+    // - [r, g, b, a] packed as u8's (when color source is payload and paint type is solid)
+    // - [x, y] packed as u16's (when color source is payload and paint type is image)
+    // - a slot index (when color source is slot)
+    @location(3) payload: u32,
+    // Packed data:
+    // - color source (1 bit)
+    // - paint type (2 bits)
```
taj-p: Do we use 2 bits instead of 1 for paint type in preparation for gradients?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -174,57 +188,58 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
-        
-        if encoded_image.quality == IMAGE_QUALITY_HIGH {
-            let final_xy = image_offset + extended_xy;
-            let sample_color = bicubic_sample(
-                atlas_texture,
-                final_xy,
-                image_offset,
-                image_size,
-                encoded_image.extend_modes
+    let alpha_byte = in.payload >> 24u;
```
taj-p:
```suggestion
```

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -174,57 +188,58 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
+    let color_source = (in.paint >> 31u) & 0x1u;
+    var final_color = unpack4x8unorm(in.payload);
+    let paint_type = (in.paint >> 29u) & 0x3u;
+
+    if color_source == COLOR_SOURCE_PAYLOAD {
+        // in.payload encodes a color for PAINT_TYPE_SOLID or sample_xy for PAINT_TYPE_IMAGE
+        final_color = alpha * final_color;
+
+        if paint_type == PAINT_TYPE_IMAGE {
```
taj-p:
```suggestion
        if paint_type == PAINT_TYPE_SOLID {
            final_color = alpha * final_color;
        } else {
            // paint_type == PAINT_TYPE_IMAGE
```

Was it on purpose to calculate `final_color` outside the if and again in the nested image quality ifs

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -174,57 +188,58 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
-            let sample_color = bicubic_sample(
-                atlas_texture,
-                final_xy,
-                image_offset,
-                image_size,
-                encoded_image.extend_modes
+    let alpha_byte = in.payload >> 24u;
+    let color_source = (in.paint >> 31u) & 0x1u;
+    var final_color = unpack4x8unorm(in.payload);
+    let paint_type = (in.paint >> 29u) & 0x3u;
```
taj-p: We only need to evaluate this if `color_source == COLOR_SOURCE_PAYLOAD` (so we can move this to within the `if` statement)

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -174,57 +188,58 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
-            let final_xy = image_offset + extended_xy;
-            let sample_color = bicubic_sample(
-                atlas_texture,
-                final_xy,
-                image_offset,
-                image_size,
-                encoded_image.extend_modes
+    let alpha_byte = in.payload >> 24u;
+    let color_source = (in.paint >> 31u) & 0x1u;
+    var final_color = unpack4x8unorm(in.payload);
```
taj-p: This value is always reassigned and only used when paint_type is solid. I think we can:

```wgsl
    var final_color: vec4<f32>; // <--- don't assign value

    if color_source == COLOR_SOURCE_PAYLOAD {
         let paint_type = (in.paint >> 29u) & 0x3u;
        // in.payload encodes a color for PAINT_TYPE_SOLID or sample_xy for PAINT_TYPE_IMAGE
        if paint_type == PAINT_TYPE_SOLID {
            final_color = alpha * unpack4x8unorm(in.payload); // <-- inline the unpack
        } else {
```

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -40,8 +45,9 @@ struct Config {
 
 // Strip instance data
 //
-// `rgba_or_slot` field can either encode a color or a slot index.
-// If the alpha value is non-zero, the fragment shader samples the alpha texture.
+// `payload` field can either encode a color, [x, y] for image sampling or a slot index
+// If color source is payload and the paint type is solid, the fragment shader uses the color directly.
+// If color source is payload and the paint type is image, the fragment shader samples the image.
 // Otherwise, the fragment shader samples the source clip texture using the given slot index.
 struct StripInstance {
```
taj-p, line 76: Let's not do this now - I think the current approach is fine, but now that I've been thinking about this a long time, I thought I should at least document some thinking:

I think we needed to increase `StripInstance` by 32 bits solely to support images in clipping regions due to needing to support passing scene coordinates. As mentioned in the first PR, it's likely there is a form of `StripInstance` where we can reduce its size by 1 `u32`.

In other words, if we didn't support images in clipping regions, we likely wouldn't need to increase `StripInstance` by 32 bits. This will be the case for wide tiles containing text or SVGs.

It's hard to say on average what is better, but I suspect that many consumers aren't benefiting from these 32 bits. Instead, I wonder whether we could remove those 32 bits if we instead added a lookup texture from slot_idx to scene coordinates which is only checked when we detect we're in a slot.

I think it might be better to _not_ action this until we've at least added gradients (which may consume more bits of `StripInstance` without necessarily increasing its size).

sparse_strips/vello_hybrid/src/render/common.rs
taj-p: I wonder whether it's easier to simply reference the documentation in the shader than try to keep this documentation up to date

taj-p (approved):
LGTM! Great work Alex! Really clever use of those `StripInstance` bits 🎉

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Good point! I often forget to update both places, so I’ve decided to go with your second suggestion: adding a reference from the `GPUStrip` documentation to the shader `StripInstance` one.

Fixed

sparse_strips/vello_hybrid/src/schedule.rs
grebmeg: That’s a good idea, though I think it’d be more appropriate to place `unimplemented!` here. These lines will be changed when we add `Gradient`s support.

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Yep, actually that comment’s outdated now.

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: It’s definitely an improvement! I also find it easier to understand. Thank you!

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: I completely agree that keeping `StripInstance` as small as possible is important — ultimately, it determines how many strips we can support. For example, assuming a total vertex buffer size of 256 MB in WebGL, and with each `StripInstance` currently using 20 bytes, we can fit a `N` instances within that limit.

Another factor to keep in mind is `max_vertex_buffer_array_stride`, which is guaranteed to be at least 256 bytes and typically goes up to 2 KB. This limit likely exists due to alignment considerations with L1/L2 GPU caches, which optimize for data that is read once per draw. Given that internal GPU throughput is highly efficient, once data is uploaded, processing it on-GPU is generally not a bottleneck.

In light of that, I believe this becomes a tradeoff discussion. Using indirect data access via texture fetches could negatively impact performance and increase implementation complexity. In comparison, keeping the data directly in the vertex buffer is simpler and may be more efficient overall.

One possible bottleneck right now is that we’re uploading all strips every frame, which could be expensive on some devices.

Anyway, I’m happy to revisit this once we add gradient support.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Yes, there’s one more — `EncodedPaint::BlurredRoundedRect`. I’m not exactly sure how it fits in yet, but in general, it should be possible to implement it using the current approach.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Oh, yes, thanks!

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: I agree, initially I just wanted to avoid any additional branches.

Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Fixed

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Yes, it was intentional, though I don’t have a strong reason for it. My thinking was to avoid adding another branch for a simple calculation like unpacking a value that would eventually be overwritten during image rendering.

PR #1092, opened 2025-07-10
Add support for SIMD flattening

LaurenzV:
A first, sketch, those are the results I'm getting:

```
flatten/Ghostscript_Tiger
                        time:   [457.71 µs 458.66 µs 459.65 µs]
                        change: [-33.258% -33.001% -32.734%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
flatten/paris_30k       time:   [29.686 ms 29.734 ms 29.783 ms]
                        change: [-13.721% -13.544% -13.363%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  1 (1.00%) high mild
flatten/coat_of_arms    time:   [2.3983 ms 2.4182 ms 2.4524 ms]
                        change: [-29.976% -29.057% -27.928%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  1 (1.00%) high severe
Benchmarking flatten/heraldry: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.5s, enable flat sampling, or reduce sample count to 50.
flatten/heraldry        time:   [1.6884 ms 1.6905 ms 1.6926 ms]
                        change: [-28.858% -27.328% -25.972%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
```

I was hoping for a bit more, but it's not nothing!

TODOs:
- Merge https://github.com/linebender/fearless_simd/pull/33 and also add WASM support.
- Fix build.


ajakubowicz-canva:
@LaurenzV I hope you don't mind that I pushed https://github.com/linebender/vello/pull/1092/commits/71a6981247474c60715985a8249464f19f161191 to test that it passes CI.

Eventually this PR will need to point to Fearless SIMD `main` once WASM has landed.

Cargo.toml
ajakubowicz-canva, line 1: A note that this PR is blocked on Cargo.toml pointing to a commit on fearless_simd `main` branch.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This is a temporary module that contains a SIMD version of flattening of cubic curves, as
+//! well as some code that was copied from kurbo, which is needed to reimplement the
+//! full `flatten` method.
```
ajakubowicz-canva, line 6: Is there a discussion about upstreaming this change kurbo? Is there a goal to eventually use `fearless_simd` in `kurbo` as well as `vello`?
Are there any risks that prevent making this change in kurbo?


If we do go this path, could we create a TODO tied to an issue to clean up this duplication/debt?

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
...
+use alloc::vec::Vec;
+use bytemuck::{Pod, Zeroable};
+use fearless_simd::*;
+
+#[cfg(not(feature = "std"))]
+use crate::kurbo::common::FloatFuncs as _;
+
+// Unlike kurbo, which takes a closure with a callback for outputting the lines, we use a trait
+// instead. The reason is that this way the callback can be inlined, which is not possible with
+// a closure and turned out to have a noticeable overhead.
```
ajakubowicz-canva, line 19: This looks like more valuable feedback for kurbo.

sparse_strips/vello_common/src/flatten_simd.rs
ajakubowicz-canva: Given this module:
 - Changes the callback mechanism to use a trait.
 - Uses `f32` instead of `f64`
 - Integrates SIMD git sketches into a new SIMD flatten function.

I think it may be worth expecting some of these functions to not be that temporary. Could we add a little more documentation?

For example, please document `flatten` ([even if it copies some or most of the original documentation](https://github.com/linebender/kurbo/blob/b0b50b3ea0e4080f9120c6141306dd7eee81f273/kurbo/src/bezpath.rs#L521-L586)).

Is there also relevant documentation that can be added to describe the difference between `kurbo`'s flatten compared to the SIMD sketches that have been integrated? It may also be helpful to describe the differences to kurbo and why we are maintaining a copy.

ajakubowicz-canva (commented):
Looking very great!

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: > Is there a discussion about upstreaming this change kurbo? Is there a goal to eventually use fearless_simd in kurbo as well as vello?

I imagine it will be upstreamed _eventually_, but probably not anytime soon. Can add a TODO.

> I think it may be worth expecting some of these functions to not be that temporary. Could we add a little more documentation?

I can try to add some more documentation, but considering that it's really just for internal use, I'm not a 100% sure whether it's really worth it to copy all of the original documentation? 😄 I

I can definitely document some of the main differences, but I won't be able to document the difference in the mathematics because I don't really understand them myself.

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: I opened a issue: https://github.com/linebender/kurbo/issues/470

Cargo.toml
LaurenzV: Done.

sparse_strips/vello_common/src/flatten_simd.rs
ajakubowicz-canva: I think what is tricky is determining the correctness bar for this PR and figuring out a path forward, if we're forking a part of kurbo with some mathematics that isn't completely understood. What bar of correctness are we aiming for in `sparse_strips` (vs what bar of correctness does `kurbo` uphold)? Is correctness for this PR upheld because we have confidence that:

1. We trust Raph's gist is correct and was migrated over accurately.
2. We pass the set of tests that are currently written. (Do we have full coverage of flattening edge cases checked into `sparse_strips`?). Are there any test cases we should migrate from kurbo to add confidence as part of this PR while we maintain a fork in flattening logic?

To be very explicit, I'm mostly just asking these questions so I'm understanding if there's any risk introduced by this change. I ultimately think the SIMD benefits are super worth it! There are three options off the top of my head for aiding me in reviewing this and moving forward.

1. We do nothing and lean on existing tests which pass. My perspective is the tests are very comprehensive.
2. Maybe we port some additional tests from kurbo (if there are any flattening specific tests) that lock in a bit more correctness around the flattening implementation we have.
3. Maybe someone with more domain expertise like Raph could more carefully review some of the mathematical soundness.

I left this comment hastily before signing off for the weekend, because I think this will be an issue for me with reviewing this PR on Monday since I love the performance improvement but struggle myself in reviewing this kind of mathematical code. I'm pretty happy to review it leveraging the current test corpus, but it might be a nice addition to port over some kurbo flattening tests if there are any to pull from.


Once again, this is awesome! I'm just trying to figure out a strategy to ensure this PR gets the review it deserves.

LaurenzV: Regarding testing, I have
 1. Tested this in an interactive tool where there is a single curve and I dragged the various points around to ensure that the flattening result at least _looks_ correct.
 2. Integrated this into my PDF rendering library (with 600+ PDFs) and rendered them to make sure there are no notable regressions, and there weren't any.

But yeah, I do understand that and we probably should discuss it during the next office hours. With that said, I don't think there are problems with merging this as is, and in case we encounter issues during production, either we find the bug or we can always revert to the `kurbo` implementation.

raphlinus:
I haven't had a chance to review this in detail, but agree with Laurenz. If there are failures, I think they will surface in the course of ordinary testing.

sparse_strips/vello_common/src/flatten_simd.rs
ajakubowicz-canva: Thank you for the great comment!

I am unblocked reviewing it on Monday, and I'll try read over the mathematics but feel excellent that you have done some additional manual testing.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
...
+pub(crate) trait Callback {
+    fn callback(&mut self, el: PathEl);
+}
+
+pub(crate) fn flatten<S: Simd>(
```
ajakubowicz-canva, line 28: Optional nit: It could be helpful for documentation to link to the [kurbo docs for `flatten`](https://docs.rs/kurbo/latest/kurbo/fn.flatten.html).
That can maybe avoid duplicate docs.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
...
+struct FlattenParams {
+    a0: f64,
+    a2: f64,
+    u0: f64,
+    uscale: f64,
+    /// The number of `subdivisions * 2 * sqrt_tol`.
+    val: f64,
+}
+
+const MAX_QUADS: usize = 16;
```
ajakubowicz-canva, line 183: Optional: Interesting! Is it worth making `MAX_QUADS` configurable? It looks like the original scalar code in kurbo doesn't have a limit (and instead uses dynamic `Vec` allocations).

Worth adding a comment as to why `16` was chosen? Is it a value that is rarely reached? This brief comment might also be useful on `estimate_num_quads` which is clamped to `MAX_QUADS`.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
...
+        let approxint = approx_parabola_integral_simd_x4(xmin);
+        let cusp = (sqrt_tol * da_abs) / approxint;
+        let val_raw = simd.select_f32x4(mask, noncusp, cusp);
+        let val = f32x4::from_bytes(
+            simd.and_mask32x4(
+                is_finite_simd(val_raw),
+                mask32x4::from_bytes(val_raw.to_bytes()),
+            )
+            .to_bytes(),
+        );
```
ajakubowicz-canva: Optional: I think this can be made more readable by replacing with:

```rs
let finite_mask = is_finite_simd(val_raw);
let val = simd.select_f32x4(finite_mask, val_raw, f32x4::splat(simd, 0.0));
```

I also tested the flatten benchmark and I don't think this changes performance:

```
Benchmarking flatten/Ghostscript_Tiger: Collecting 100 samples in estimated 6.0047 s (15k iterations
flatten/Ghostscript_Tiger
                        time:   [390.01 µs 391.21 µs 392.62 µs]
                        change: [-0.6446% -0.0784% +0.5759%] (p = 0.81 > 0.05)
  5 (5.00%) high mild
  2 (2.00%) high severe

```

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
...
+    let uscale = f32x8::splat(simd, ctx.uscale[i]);
+
+    for j in 0..n.div_ceil(4) {
+        let u = approx_parabola_inv_integral_simd(a);
+        let t = (-ctx.u0[i] * uscale).madd(u, uscale);
+        let mt = 1.0 - t;
+        let z = p0 * mt * mt;
+        let z1 = z.madd(p1, 2.0 * t * mt);
+        let p = z1.madd(p2, t * t);
+
```
ajakubowicz-canva, line 431: (Non actionable)

I spent a bit of time trying to optimize this function by trying different approaches. I can't beat what is in this PR :D 

(I read over this [random resource](https://www.gamedev.net/articles/programming/math-and-physics/practical-guide-to-bezier-curves-r3166/) and tried to replace this with Horner's algorithm.

I could only tie with the current implementation. If you're interested, this is what I tried:

```
#[inline]
fn output_lines_simd<S: Simd>(
    simd: S,
    ctx: &FlattenCtx,
    i: usize,
    x0: f32,
    dx: f32,
    n: usize,
    out: &mut [f32],
) {
    let p0 = pt_splat_simd(simd, ctx.even_pts[i]);
    let p1 = pt_splat_simd(simd, ctx.odd_pts[i]);
    let p2 = pt_splat_simd(simd, ctx.even_pts[i + 1]);

    let two_p1_minus_p0 = 2.0 * (p1 - p0);
    let p0_minus_2p1_plus_p2 = p0 - 2.0 * p1 + p2;
    
    const IOTA2: [f32; 8] = [0., 0., 1., 1., 2., 2., 3., 3.];
    let iota2 = f32x8::from_slice(simd, &IOTA2);
    let x = f32x8::splat(simd, x0) + dx * iota2;
    let da = f32x8::splat(simd, ctx.da[i]);
    let mut a = f32x8::splat(simd, ctx.a0[i]) + da * x;
    let a_inc = f32x8::splat(simd, 4.0 * dx * ctx.da[i]);
    
    let uscale = f32x8::splat(simd, ctx.uscale[i]);
    let u0_uscale = f32x8::splat(simd, ctx.u0[i] * ctx.uscale[i]);
    
    for j in 0..n.div_ceil(4) {
        let u = approx_parabola_inv_integral_simd(a);
        let t = u * uscale - u0_uscale;
        let p = p0 + t * (two_p1_minus_p0 + t * p0_minus_2p1_plus_p2);
        
        out[j * 8..][..8].copy_from_slice(&p.val);
        a = a + a_inc;
    }
}
```

It also fails two tests due to what I assume are floating point differences.

ajakubowicz-canva (approved):
This is amazing!
I reviewed this by going function by function in the `flatten_simd` module, and I played with each one. I tried to find alternative implementations and tried to figure out the algorithms.

I couldn't beat what is implemented in this PR. This is great! I learnt a lot!!!

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -0,0 +1,497 @@
...
+fn estimate_num_quads(c: CubicBez, accuracy: f32) -> usize {
+    let q_accuracy = (accuracy * TO_QUAD_TOL) as f64;
+    let max_hypot2 = 432.0 * q_accuracy * q_accuracy;
+    let p1x2 = c.p1.to_vec2() * 3.0 - c.p0.to_vec2();
+    let p2x2 = c.p2.to_vec2() * 3.0 - c.p3.to_vec2();
+    let err = (p2x2 - p1x2).hypot2();
+    let n_quads = ((err / max_hypot2).powf(1. / 6.0).ceil() as usize).max(1);
+
+    n_quads.min(MAX_QUADS)
+}
```
ajakubowicz-canva, line 497: Since there is some pretty complex math here, do you think there's an `accuracy` value that could benefit from a simpler fast path?

Maybe the branching cost would ruin the optimization.

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: From Raph: 
image

I can add a comment, though.

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: Hmm, perhaps, but I definitely don't have the maths knowledge to figure it out. 😄

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: Done.

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: Good catch, thanks!

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV: Added a short comment.

LaurenzV:
@ajakubowicz-canva Thanks a lot for the reviews!

PR #1093, opened 2025-07-12
No shader runtime checks

sagudev:
In servo I observed performance regression with #908: https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Servo.202D.20canvas.20backend/with/528398019

but disabling shader runtime checks gives it back on pre #908.

### DIFF
--- a/vello/src/wgpu_engine.rs
+++ b/vello/src/wgpu_engine.rs
@@ -834,10 +834,15 @@ impl WgpuEngine {
         entries: Vec<wgpu::BindGroupLayoutEntry>,
         cache: Option<&PipelineCache>,
     ) -> WgpuShader {
-        let shader_module = device.create_shader_module(wgpu::ShaderModuleDescriptor {
-            label: Some(label),
-            source: wgpu::ShaderSource::Wgsl(wgsl),
-        });
+        let shader_module = unsafe {
+            device.create_shader_module_trusted(
+                wgpu::ShaderModuleDescriptor {
+                    label: Some(label),
+                    source: wgpu::ShaderSource::Wgsl(wgsl),
+                },
+                wgpu::ShaderRuntimeChecks::unchecked(),
+            )
+        };
         let bind_group_layout = device.create_bind_group_layout(&wgpu::BindGroupLayoutDescriptor {
             label: None,
             entries: &entries,
--- a/vello/src/wgpu_engine.rs
+++ b/vello/src/wgpu_engine.rs
@@ -834,6 +834,7 @@ impl WgpuEngine {
         entries: Vec<wgpu::BindGroupLayoutEntry>,
         cache: Option<&PipelineCache>,
     ) -> WgpuShader {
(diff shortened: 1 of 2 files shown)


sagudev:
It's https://docs.rs/wgpu/latest/wgpu/struct.ShaderRuntimeChecks.html#structfield.force_loop_bounding that makes mac fail.

waywardmonkeys (approved):
Yes, I was going to suggest the same thing.

sagudev:
There are more `create_shader_module` invocations in hybrid, that we could potentially trust, but this PR is focused on vello classic.

waywardmonkeys:
> There are more `create_shader_module` invocations in hybrid, that we could potentially trust, but this PR is focused on vello classic.

I think I it is fine to file an issue for them to track this in hybrid.

vello/src/wgpu_engine.rs
```diff
@@ -834,10 +834,15 @@ impl WgpuEngine {
         entries: Vec<wgpu::BindGroupLayoutEntry>,
         cache: Option<&PipelineCache>,
     ) -> WgpuShader {
-        let shader_module = device.create_shader_module(wgpu::ShaderModuleDescriptor {
-            label: Some(label),
-            source: wgpu::ShaderSource::Wgsl(wgsl),
-        });
+        let shader_module = unsafe {
```
DJMcNab, line 838: This needs a `SAFETY` comment. In particular, this function not being public is now part of the crate's soundness boundary. I do think the tradeoff of only having the unsafe be here is reasonable enough.

(I'm also a little bit surprised that this crate isn't `#![forbid(unsafe_code)]` or at least `#![deny(unsafe_code)]` - not to say that this use of unsafe isn't justified).

vello/src/wgpu_engine.rs
```diff
@@ -834,10 +834,15 @@ impl WgpuEngine {
...
+            device.create_shader_module_trusted(
+                wgpu::ShaderModuleDescriptor {
+                    label: Some(label),
+                    source: wgpu::ShaderSource::Wgsl(wgsl),
+                },
+                wgpu::ShaderRuntimeChecks::unchecked(),
```
DJMcNab, line 844: There are reasonable arguments for making this `force_loop_bounding: false; bounds_checks: cfg!(debug_assertions))`.

But OTOH, a bounds check failure never gets back-reported, so practically I agree that this is fine.

sagudev: I was also thinking this, but decided against it for similar reasons.

vello/src/wgpu_engine.rs
sagudev: Done in https://github.com/linebender/vello/pull/1093/commits/1c32b7c2bd52014dd3bc2de527363b6d05d45795, LMK if we can word it better.

DJMcNab (approved):
Thanks!

PR #1095, opened 2025-07-12
Remove `once_cell` dependency in `vello_common`

LaurenzV:


ajakubowicz-canva (dismissed):

sparse_strips/vello_common/src/lib.rs
```diff
@@ -51,7 +51,7 @@
     reason = "We temporarily ignore those because the casts\
 only break in edge cases, and some of them are also only related to conversions from f64 to f32."
 )]
-#![no_std]
+#![cfg_attr(not(feature = "multithreading"), no_std)]
```
waywardmonkeys: No ... this shouldn't do this.

I think it should instead be making `multithreading` enable the `std` feature.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -14,9 +14,9 @@ use alloc::borrow::Cow;
 use alloc::vec::Vec;
 #[cfg(not(feature = "multithreading"))]
 use core::cell::OnceCell;
-#[cfg(feature = "multithreading")]
-use once_cell::sync::OnceCell;
 use smallvec::SmallVec;
+#[cfg(feature = "multithreading")]
+use std::sync::OnceLock as OnceCell;
```
waywardmonkeys, line 20: This seems like it'll lead to confusion and the usages should be fixed.

waywardmonkeys: I see why this was done this way now ... I don't know what should be done, but I wouldn't block on this part. I would still want to see the other part fixed though.

sparse_strips/vello_common/src/lib.rs
waywardmonkeys: basically, if something from `std` is needed, then we'd do what we do in other files in this crate and conditionally `extern crate std`.

sparse_strips/vello_common/src/encode.rs
LaurenzV: Yep will do, just had breakfast that’s why I only made the one change so far

LaurenzV: I added a comment explainining it instead, I think it's still nicer than duplicating the selection logic everywhere.

PR #1097, opened 2025-07-14
Don't make `std` prelude available in `vello_cpu`

LaurenzV:


sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -59,7 +59,7 @@ impl MultiThreadedDispatcher {
             .unwrap();
         let alpha_storage = Arc::new(OnceLockAlphaStorage::new(num_threads));
         let workers = Arc::new(ThreadLocal::new());
-        let task_batch = vec![];
+        let task_batch = std::vec![];
```
DJMcNab:
```suggestion
        let task_batch = alloc::vec![];
```

(Although I'd probably suggest just importing it alongside where you import `Vec`)

tomcur (approved):

PR #1098, opened 2025-07-14
Performance improvements for gradients

LaurenzV:
This PR contains a number of changes to improve the performance of complex paints:
- Instead of wrapping each painter in a `Box`, we use generics. I checked the changes in my C wrapper library and it didn't have any noticeable impact on binary size, which makes sense because the `paint_u8` and `paint_f32` methods are probably never inlined. However, it does feel like it might have impacted compilation time a bit, but I haven't measured it. But this is definitely a necessary change, as the `Box` allocations really showed up in the profiles (~20%).
- I implemented an optimization for SRGB gradients so we don't create additional stops which are unnecessary.
- I slightly reduced the size of the LUT table for certain number of stops, but this should still be more than sufficient (and also matches Blend2D).
- I SIMDified and optimized the actual LUT computation.

Here is a comparison of rendering times for linear gradients, the gains are especially noticeable for smaller sizes, but also the biggest one has gotten quite a bit faster thanks to the removal of `Box`.:

Before:
image

After:
image

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -913,7 +913,7 @@ impl<T: FromF32Color> GradientLut<T> {
             2 => 512,
             _ => 1024,
         };
-        
+
         // Add a bit of padding since we always process in blocks of 4, even though less might be
         // needed.
         let mut lut = vec![[T::ZERO, T::ZERO, T::ZERO, T::ZERO]; lut_size + 3];
@@ -934,15 +934,14 @@ impl<T: FromF32Color> GradientLut<T> {
         };
 
         let inv_lut_size = f32x4::splat(simd, 1.0 / lut_size as f32);
+        let add_factor = f32x4::from_slice(simd, &[0.0, 1.0, 2.0, 3.0]) * inv_lut_size;
 
         for (ramp_range, range) in ramps {
             let biases = f32x16::block_splat(f32x4::from_slice(simd, &range.bias));
             let scales = f32x16::block_splat(f32x4::from_slice(simd, &range.scale));
 
             ramp_range.step_by(4).for_each(|idx| {
-                let t_vals = (f32x4::splat(simd, idx as f32)
-                    + f32x4::from_slice(simd, &[0.0, 1.0, 2.0, 3.0]))
-                    * inv_lut_size;
+                let t_vals = add_factor.madd(f32x4::splat(simd, idx as f32), inv_lut_size);
 
                 let t_vals = element_wise_splat(simd, t_vals);
 


sparse_strips/vello_common/src/simd.rs
```diff
@@ -0,0 +1,97 @@
...
+pub trait Splat4thExt<S> {
+    /// Splat every 4th element of the vector.
+    fn splat_4th(self) -> Self;
+}
+
+impl<S: Simd> Splat4thExt<S> for f32x4<S> {
+    #[inline(always)]
+    fn splat_4th(self) -> Self {
+        let zip1 = self.zip_high(self);
+        zip1.zip_high(zip1)
```
ajakubowicz-canva, line 20: Did you benchmark any alternatives to using `zip_high` twice?

sparse_strips/vello_common/src/simd.rs
```diff
@@ -0,0 +1,97 @@
...
+        p2 = p2.splat_4th();
+
+        self.simd.combine_f32x8(p1, p2)
+    }
+}
+
+impl<S: Simd> Splat4thExt<S> for u8x16<S> {
+    #[inline(always)]
+    fn splat_4th(self) -> Self {
+        // TODO: SIMDify
```
ajakubowicz-canva, line 49: Is this being tracked in an issue in fearless_simd?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -840,69 +866,105 @@ fn unit_to_line(p0: Point, p1: Point) -> Affine {
...
+    fn new<S: Simd>(simd: S, ranges: &[GradientRange]) -> Self {
+        // Inspired by Blend2D.
+        let lut_size = match ranges.len() {
+            1 => 256,
+            2 => 512,
+            _ => 1024,
+        };
 
-        let inv_lut_size = 1.0 / lut_size as f32;
+        let mut lut = vec![[T::ZERO, T::ZERO, T::ZERO, T::ZERO]; lut_size + 4];
```
ajakubowicz-canva: Is it worth adding an implementation comment for why `+4` is being added here? I _think_ this is being used to avoid any bounds checking in the following LUT calculation. (As SIMD must operate on groups of 4)

Could this over-allocation be smaller - e.g. `3`?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -840,69 +866,105 @@ fn unit_to_line(p0: Point, p1: Point) -> Affine {
...
+        let inv_lut_size = f32x4::splat(simd, 1.0 / lut_size as f32);
+
+        for (ramp_range, range) in ramps {
+            let biases = f32x16::block_splat(f32x4::from_slice(simd, &range.bias));
+            let scales = f32x16::block_splat(f32x4::from_slice(simd, &range.scale));
+
+            ramp_range.step_by(4).for_each(|idx| {
+                let t_vals = (f32x4::splat(simd, idx as f32)
+                    + f32x4::from_slice(simd, &[0.0, 1.0, 2.0, 3.0]))
+                    * inv_lut_size;
```
ajakubowicz-canva: Would there be a performance benefit to either:
 1. Pull the constant multiplication out of the loop, and only adding in the loop?
 2. Using `madd` if we want to keep the multiplication in the loop?

ajakubowicz-canva (approved):
Awesome!

sparse_strips/vello_common/src/simd.rs
LaurenzV: I have not, will add this as a TODO. But since it only affects the high-precision pipeline it's at least not super critical.

sparse_strips/vello_common/src/simd.rs
LaurenzV: Nope, I'm not really sure how you would SIMDify this best, though.

sparse_strips/vello_common/src/encode.rs
LaurenzV: The addition is performed before the multiplication (hence the brackets), so unfortunately this isn't possible. But I tried doing one multiplication outside of the loop and then `madd` in the loop, and it seems to at least have a small effect.

sparse_strips/vello_common/src/encode.rs
LaurenzV: Done.

PR #1102, opened 2025-07-15
Partly SIMDify the `f32_to_u8` method

LaurenzV:
This gives me -7% runtime on the bilinear image rendering benchmark on NEON:

```
     Running benches/main.rs (/Users/lstampfl/Programming/GitHub/vello/target/release/deps/main-011ad84cd531a979)
                        time:   [3.4155 µs 3.4190 µs 3.4231 µs]
                        change: [-7.3623% -7.1971% -7.0174%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  5 (5.00%) high severe
```

On WASM, it drops the frame time from 40ms to 30ms, so a huge difference! I'm not sure if it's possible to easily SIMDify the u32 to u8 conversion (at least from what I can tell `tiny-skia` also only uses SIMD for the f32 to u32 conversion), open to suggestions, though.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/mod.rs
+++ b/sparse_strips/vello_cpu/src/fine/mod.rs
@@ -91,24 +91,26 @@ impl<S: Simd> NumericVec<S> for u8x16<S> {
 
 #[inline(always)]
 pub(crate) fn f32_to_u8<S: Simd>(val: f32x16<S>) -> u8x16<S> {
-    // TODO: SIMDify
+    let simd = val.simd;
+    let converted = simd.cvt_u32_f32x16(val);
+    
     [
-        val.val[0] as u8,
-        val.val[1] as u8,
-        val.val[2] as u8,
-        val.val[3] as u8,
-        val.val[4] as u8,
-        val.val[5] as u8,
-        val.val[6] as u8,
-        val.val[7] as u8,
-        val.val[8] as u8,
-        val.val[9] as u8,
-        val.val[10] as u8,
-        val.val[11] as u8,
-        val.val[12] as u8,
-        val.val[13] as u8,
-        val.val[14] as u8,
-        val.val[15] as u8,
+        converted[0] as u8,
+        converted[1] as u8,
+        converted[2] as u8,
+        converted[3] as u8,
+        converted[4] as u8,
+        converted[5] as u8,
+        converted[6] as u8,
+        converted[7] as u8,
+        converted[8] as u8,
+        converted[9] as u8,
+        converted[10] as u8,
+        converted[11] as u8,
+        converted[12] as u8,
+        converted[13] as u8,
+        converted[14] as u8,
+        converted[15] as u8,
     ]
     .simd_into(val.simd)
 }
--- a/sparse_strips/vello_cpu/src/fine/mod.rs
+++ b/sparse_strips/vello_cpu/src/fine/mod.rs
@@ -92,8 +92,11 @@ impl<S: Simd> NumericVec<S> for u8x16<S> {
 #[inline(always)]
 pub(crate) fn f32_to_u8<S: Simd>(val: f32x16<S>) -> u8x16<S> {
     let simd = val.simd;
+    // Note that converting to u32 first using SIMD and then u8
+    // is much faster than converting directly from f32 to u8.
     let converted = simd.cvt_u32_f32x16(val);
 
+    // TODO: Maybe we can also do this using SIMD?
     [
         converted[0] as u8,
         converted[1] as u8,
--- a/sparse_strips/vello_cpu/src/fine/mod.rs
+++ b/sparse_strips/vello_cpu/src/fine/mod.rs
@@ -93,7 +93,7 @@ impl<S: Simd> NumericVec<S> for u8x16<S> {
 pub(crate) fn f32_to_u8<S: Simd>(val: f32x16<S>) -> u8x16<S> {
     let simd = val.simd;
     let converted = simd.cvt_u32_f32x16(val);
-    
+
     [
         converted[0] as u8,
         converted[1] as u8,


sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -91,24 +91,26 @@ impl<S: Simd> NumericVec<S> for u8x16<S> {
 
 #[inline(always)]
 pub(crate) fn f32_to_u8<S: Simd>(val: f32x16<S>) -> u8x16<S> {
-    // TODO: SIMDify
```
ajakubowicz-canva, line 94: Maybe maintain a comment that this is only partly SIMD-ified.

Also - that this change would result in a performance improvement is completely unintuitive. May be worth a small inline comment that the performance improvement is due to the SIMD-ifying from f32 -> u32, is faster than a scalar conversion from f32 -> u8.

sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -91,24 +91,26 @@ impl<S: Simd> NumericVec<S> for u8x16<S> {
...
+    let simd = val.simd;
+    let converted = simd.cvt_u32_f32x16(val);
```
grebmeg, line 97: Would we get a further performance gain by optimizing with `reinterpret_u8_u32x16`? Something like below:

```rust
pub(crate) fn f32_to_u8<S: Simd>(val: f32x16<S>) -> u8x16<S> {
    let simd = val.simd;
    let converted = simd.cvt_u32_f32x16(val);
    let reinterpret_u8_u32x16 = simd.reinterpret_u8_u32x16(converted);

    [
        reinterpret_u8_u32x16.val[0],
        reinterpret_u8_u32x16.val[4],
        reinterpret_u8_u32x16.val[8],
        reinterpret_u8_u32x16.val[12],
        reinterpret_u8_u32x16.val[16],
        reinterpret_u8_u32x16.val[20],
        reinterpret_u8_u32x16.val[24],
        reinterpret_u8_u32x16.val[28],
        reinterpret_u8_u32x16.val[32],
        reinterpret_u8_u32x16.val[36],
        reinterpret_u8_u32x16.val[40],
        reinterpret_u8_u32x16.val[44],
        reinterpret_u8_u32x16.val[48],
        reinterpret_u8_u32x16.val[52],
        reinterpret_u8_u32x16.val[56],
        reinterpret_u8_u32x16.val[60],
    ]
    .simd_into(val.simd)
}
```

LaurenzV: Let me try.

LaurenzV: Doesn't seem to have any impact, so I'll keep the current code since it's a bit more compact.

grebmeg: Huh, interesting — I’m getting a consistent ~11.5% speedup on my M1 Max with that code. Perhaps there’s something I’m missing?

```
                        time:   [3.4874 µs 3.5206 µs 3.5596 µs]
                        change: [-12.086% -11.569% -10.929%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) high mild
  2 (2.00%) high severe
```

LaurenzV: Maybe I am doing something wrong, but I cannot reproduce this. :( However, if I use either of our versions _and_ remove the `inline` attribute I'm seeing another 6% boost, is that maybe where yours is coming from?

grebmeg: I don’t think so — I only modified the function body 🤔 if I remove `inline`, I see a 44.485% performance regression 😅 I’m using the following command to run the code:
```shell
cargo bench -p vello_bench -- fine/image/quality/medium_u8_neon
```

LaurenzV: That's really weird, it really doesn't reproduce on my machine, seems weird because I also have an M1 chip... Maybe a third person could try.

grebmeg: Mine’s an M1 Max — not sure if that’s the reason for the difference. I noticed you’ve already updated the code in another draft PR — nice work! Still, I’m really curious about why we’re seeing different results.

Hey @ajakubowicz-canva, if you’ve got a moment, would you mind running the code above on your laptop? It’d be great to have a third data point.

ajakubowicz-canva: On my machine (M2 Max), I get:

```sh
                        time:   [3.0263 µs 3.0335 µs 3.0422 µs]
                        change: [-0.6361% -0.0671% +0.5254%] (p = 0.83 > 0.05)
```

When switching between the two cases (what is merged, vs using `reinterpret_u8_u32x16`).


Just to see what happens on my machine, I made the function in this PR `inline(never)`, such that I could reference the function in the compiled assembly, and ran: `cargo asm -p vello_cpu "vello_cpu::fine::f32_to_u8" --rust 0`. This prints out the function as it's compiled.

Result (as merged currently):

```asm
.globl  vello_cpu::fine::f32_to_u8
        .p2align        2
vello_cpu::fine::f32_to_u8:
Lfunc_begin175:
        .cfi_startproc
        ldp q0, q1, [x0, #32]
        ldp q2, q3, [x0]
        fcvtzu.4s v3, v3
        fcvtzu.4s v2, v2
        fcvtzu.4s v1, v1
        fcvtzu.4s v0, v0
        uzp1.8h v0, v0, v1
        uzp1.8h v1, v2, v3
        uzp1.16b v0, v1, v0
        str q0, [x8]
        ret
```

If I applied @grebmeg 's change we get identical assembly.
However, if I make the function scalar again (e.g. revert all SIMD changes):

```asm
        .globl  vello_cpu::fine::f32_to_u8
        .p2align        2
vello_cpu::fine::f32_to_u8:
Lfunc_begin175:
        .cfi_startproc
        ldp q0, q1, [x0]
        ldp q2, q3, [x0, #32]
        fcvtzu.4s v3, v3
        movi.2d v4, #0x0000ff000000ff
        umin.4s v3, v3, v4
        fcvtzu.4s v2, v2
        umin.4s v2, v2, v4
        uzp1.8h v2, v2, v3
        fcvtzu.4s v1, v1
        umin.4s v1, v1, v4
        fcvtzu.4s v0, v0
        umin.4s v0, v0, v4
        uzp1.8h v0, v0, v1
        uzp1.16b v0, v0, v2
        str q0, [x8]
        ret
```

We can look at the compiled assembly to understand the performance improvement:

- Both SIMD variants produce identical assembly - the compiler optimizes away the `reinterpret_u8_u32x16` abstraction difference.
- SIMD gains performance by removing bounds checking - the scalar version clamps values to 255 using `umin` instructions, while SIMD assumes inputs are already in valid range.
- Instruction count: 11 (SIMD) vs 17 (scalar) - SIMD saves 4 `umin` instructions, and 1 constant load.

This performance improvement comes at a trade-off where if the input exceeds `255`, SIMD may produce undefined results (since explicit clamping no longer happens).

PR #1103, opened 2025-07-15
Optimize alpha coverage calculation in strip rendering

LaurenzV:
Instead of iterating over the windings in chunks of 4 and pushing out the alpha values in each iteration, first calculate the alpha coverage as f32 for all locations, then convert them in one step to u8 and finally push them to the alpha buffer at the same time.

This PR is dependent on #1102, so marking this as draft.

Gives the following improvements:

```
                        time:   [163.37 µs 163.62 µs 163.94 µs]
                        change: [-8.0006% -7.7978% -7.6215%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 50 measurements (12.00%)
  3 (6.00%) high mild
  3 (6.00%) high severe
render_strips/paris_30k_simd
                        time:   [18.553 ms 18.679 ms 18.819 ms]
                        change: [-4.3391% -3.3364% -2.3418%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (4.00%) high mild
render_strips/coat_of_arms_simd
                        time:   [1.3592 ms 1.3608 ms 1.3628 ms]
                        change: [-14.658% -14.333% -14.020%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 50 measurements (12.00%)
  2 (4.00%) high mild
  4 (8.00%) high severe
render_strips/heraldry_simd
                        time:   [912.09 µs 912.53 µs 913.00 µs]
                        change: [-11.924% -11.736% -11.570%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (4.00%) high mild
  2 (4.00%) high severe
```

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -115,7 +115,7 @@ fn render_impl<S: Simd>(
                         let area = location_winding[x];
                         let coverage = area.abs();
                         let mulled = p1.madd(coverage, p2);
-                        // Note that we are not storing the location winding here but the actual 
+                        // Note that we are not storing the location winding here but the actual
                         // alpha value as f32, so we reuse the variable as a temporary storage.
                         // Also note that we need the `min` here because the winding can be > 1
                         // and thus the calculated alpha value need to be clamped to 255.


sparse_strips/vello_common/src/strip.rs
```diff
@@ -114,12 +115,7 @@ fn render_impl<S: Simd>(
                         let area = location_winding[x];
                         let coverage = area.abs();
                         let mulled = p1.madd(coverage, p2);
-                        let slice = mulled.val;
-                        // TODO: Improve this
-                        alpha_buf.push(slice[0] as u8);
-                        alpha_buf.push(slice[1] as u8);
-                        alpha_buf.push(slice[2] as u8);
-                        alpha_buf.push(slice[3] as u8);
+                        location_winding[x] = mulled.min(p2);
```
taj-p, line 122: Do we need a comment somewhere describing that this acts as both `location_winding` and an intermediate alpha `u8` slice (hence the `.min`)?

taj-p (approved):
LGTM! Clever use of reusing `location_winding`! 🧠

sparse_strips/vello_common/src/strip.rs
LaurenzV: Good point!

tomcur (approved):
Looks good!

PR #1105, opened 2025-07-15
Ignore NaN paths

LaurenzV:


sparse_strips/vello_common/src/flatten.rs
```diff
@@ -76,36 +76,46 @@ pub fn fill(path: &BezPath, affine: Affine, line_buf: &mut Vec<Line>) {
...
+            }
         }
     });
 
     if !closed {
         close_path(start, p0, line_buf);
     }
+
+    // A path that contains NaN is ill-defined, so ignore it.
+    if is_nan {
```
grebmeg, line 117: Would it make sense to log the presence of a NaN (maybe only in debug)? In this case, I think it would be helpful to log which exact value is NaN within the closure above?

LaurenzV: I thought about this too, but how would we log this? Do we really want to add the log dependency just for that?

DJMcNab: I think that it would be reasonable. `log` is a reasonably small crate, silent failures are really unfriendly, and most reasonable programs will already be including log anyway.

One other thing which could be worth considering is keeping a source [`Location`](https://doc.rust-lang.org/stable/core/panic/struct.Location.html) (probably gated behind a feature or rustflag?). Because tracking down where `NaN` paths come from is a pain otherwise.

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -76,36 +77,48 @@ pub fn fill(path: &BezPath, affine: Affine, line_buf: &mut Vec<Line>) {
...
     let mut closed = false;
+    let mut is_nan = false;
 
-    kurbo::flatten(iter, TOL, |el| match el {
-        kurbo::PathEl::MoveTo(p) => {
-            if !closed && p0 != start {
-                close_path(start, p0, line_buf);
-            }
+    kurbo::flatten(iter, TOL, |el| {
+        is_nan |= el.is_nan();
```
nicoburns, line 83: Have we investigated whether the `NaN`s are being passed into `kurbo::flatten` or generated by `kurbo::flatten`?

If the former, then might we want to:
- Detect `NaN` before passing them to flatten (perhaps in an `iter.map(..)`?)
- Bail early in case of `NaN`.
?

The latter would be a bigger issue...

LaurenzV: I'm pretty sure when I checked yesterday the NaNs where passed from `blitz`. I feel like iterating twice over the path just to check for NaN would be a bit wasteful since it's really an edge case (and I feel like it's on the client to not pass such paths in the first place, so this is just an additional sanity check, really), and we can't abort early since it's just a callback. The slowdown did not come from the flattening but from later stages, so this should be enough to fix the issue you were facing.

nicoburns: > I'm pretty sure when I checked yesterday the NaNs where passed from `blitz`

This does not surprise me at all. But it's good to check.

> I feel like iterating twice over the path just to check for NaN would be a bit wasteful since it's really an edge case
> we can't abort early since it's just a callback

A `.map` wouldn't do that due to Rust iterators being lazy, it would just push the `NaN` check prior to the flattening stage. With that approach I think we could abort early by using `filter_map` rather than `map`, but it's entirely possible that would introduce overhead on the happy (non-`NaN`) path. If don't anticipate `NaN` causing problems (e.g. slowdowns) earlier in the pipeline than this then this seems fine.

> so this should be enough to fix the issue you were facing.

I can confirm that it does.

sparse_strips/vello_common/src/flatten.rs
nicoburns: I wonder if we should consider an opt-in panic on `NaN`. IME `NaN` values almost always indicate a programmer error (one absolutely shouldn't be passing `NaN`s into Vello), and a `panic` would give us a stack trace without any additional effort.

Morally this would be a debug_assertion, but as we generally want to run Vello in release mode it would probably be better to use a feature flag rather than an actual `cfg(debug_assertions)`

AustinMReppert: I think log makes more sense here. For downstream vello apps, it would make diagnosing performance issues easier.

DJMcNab: > a panic would give us a stack trace without any additional effort.

Oh cool! I thought the multithreading feature would mess with that, but if this is still in the main thread, then opt-in panicking here would be cleaner than storing our own `Location`s.

Either way, anything further than logging can be made into a follow-up

nicoburns (approved):
Per discussion at office hours, it seems that this is likely acceptable at least in the short term. A more principled stance on dealing with `NaN`s may follow later.

PR #1106, opened 2025-07-15
Optimize blending with default blend mode

LaurenzV:
While playing with a profiler and blitz, I noticed that a non-insignificant amount of time is going into compositing/blending, even though no special blend modes are used. For the default blend mode, we can simply use the fast path of doing alpha compositing instead of taking the slow route.


nicoburns (approved):
Not an expert on blending/compositing. But it is consistent with my understanding that it should be possible to optimise the default blend mode(s).

How does this relate to the `Clip` blend mode? Is that included in `is_default()`, and if not, should it be?

LaurenzV:
> How does this relate to the Clip blend mode? Is that included in is_default(), and if not, should it be?

Yep, it's included there.

PR #1108, opened 2025-07-16
Add a very basic SVG renderer to `vello_toy`

LaurenzV:
Picking up this previous attempt to add this to `vello_toy`. The main intention is to be able to easily benchmark changes on more complex inputs instead of just relying on the microbenchmarks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /target
 /debug.svg
+/svg.png
 /trace.json
 .DS_Store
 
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 /target
 /debug.svg
+/svg.png
```
grebmeg, line 3: Is this file meant to be committed?

LaurenzV: Nope, that's why I put it in the .gitignore, or what do you mean? 🤔  The binary will produce a PNG file with that name, so in order to avoid accidentally committing it, it's added here.

PR #1109, opened 2025-07-18
[vello_hybrid] add debug assertion to enforce correct WebGl context attributes

ajakubowicz-canva:
This PR adds an optional debug assertion to catch a configuration error (that cost me a couple hours of debugging).

**Context**

When using `vello_hybrid`, it's possible to accidentally create a WebGL context on the `HTMLCanvasElement` before passing it to `WebGlRenderer::new(canvas)`. [Per the WebGL specification](https://html.spec.whatwg.org/#dom-canvas-getcontext), if the canvas already has a context, that cached context is returned and any new context attributes are ignored.

This is a problem when the cached context has `antialias: true` (default), as `vello_hybrid` requires `antialias: false` to render anything.

**User impact**

Previously: `vello_hybrid` would fail silently, rendering nothing and optionally logging GL operation warnings to the console.

Now (debug mode only): The renderer throws on instantiation with a clear error message indicating that antialias must be false.

### DIFF
--- a/sparse_strips/vello_hybrid/Cargo.toml
+++ b/sparse_strips/vello_hybrid/Cargo.toml
@@ -29,6 +29,7 @@ js-sys = { version = "0.3.77", optional = true }
 web-sys = { version = "0.3.77", features = [
     "HtmlCanvasElement",
     "WebGl2RenderingContext",
+    "WebGlContextAttributes",
     "WebGlProgram",
     "WebGlUniformLocation",
     "WebGlBuffer",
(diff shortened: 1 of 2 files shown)


taj-p (approved):
LGTM

grebmeg:
I’ll just leave this link below here in case it’s helpful for anyone wondering why `antialias` should be set to false.

~~Read more in [3.7.4 Framebuffer objects](https://registry.khronos.org/webgl/specs/latest/2.0/) under `blitFramebuffer`~~

Actually, a more detailed explanation can be found in the [OpenGL specification](https://registry.khronos.org/OpenGL/specs/es/3.0/es_spec_3.0.pdf) under section 4.3.2 Reading Pixels.

grebmeg:
Actually, a more detailed explanation can be found in the [OpenGL specification](https://registry.khronos.org/OpenGL/specs/es/3.0/es_spec_3.0.pdf) under section 4.3.2 Reading Pixels.

PR #1110, opened 2025-07-18
Update to kurbo@0.11.3

sagudev:
Depends on https://github.com/linebender/kurbo/pull/467


DJMcNab (approved):
I guess we can land this? Does it have any value for us?

sagudev:
> Does it have any value for us?

Yes, it will fix #1066.

DJMcNab:
I'd like some clarification on that then - this PR has not been marked as fixing that issue; is that because you want to leave it open to track adding a regression test (i.e. #1067)?

I don't want to leave that as an orphan issue, and it's a nicer UX if the close event is correctly linked to the PR which addresses the issue.

sagudev:
> I'd like some clarification on that then - this PR has not been marked as fixing that issue; is that because you want to leave it open to track adding a regression test (i.e. https://github.com/linebender/vello/pull/1067)?
> 
> I don't want to leave that as an orphan issue, and it's a nicer UX if the close event is correctly linked to the PR which addresses the issue.

Yeah, I was planning to close in it with regression test, but I agree it's better to close it here.

Do land if you think it's okay now.

PR #1112, opened 2025-07-21
Add getters for `RenderContext` options

### LINKED ISSUE #1111: Add getters on `RenderingContext` for options
Currently there is no way to get current transform and other options (basically all struct fields) from `RenderingContext`. So I think we should expose these by impl getter for each setter.

usecase: Temporary overwrite one of options.

sagudev:
Fixes #1111.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -305,9 +305,19 @@ impl RenderContext {
         self.stroke = stroke;
     }
 
+    /// Get the current stroke
+    pub fn stroke(&self) -> &Stroke {
+        &self.stroke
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):
Seems reasonable to me as a temporary measure until we switch to a non-stateful API.

sagudev:
> Seems reasonable to me as a temporary measure until we switch to a non-stateful API.

Can't wait for stateless, why don't we simply wrote stateless wrapper that would be vello-classic compatible? I think that would be doable (and something like this already exists as https://github.com/DioxusLabs/blitz/blob/main/packages/anyrender_vello_cpu/src/scene.rs).

PR #1114, opened 2025-07-21
Add a dynamically calculated granularity when rasterizing with multi-threading

LaurenzV:
See the comments in the code for an explanation. I probably should note that this change is not based on some measurement. Unfortunately, the rendering times vary very wildly (I still need to investigate whether there is something I can do to alleviate that), which makes it close to impossible to make precise measurements. But I think that at least conceptually this is a reasonable thing, but I'm open for feedback!

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -234,15 +234,22 @@ impl MultiThreadedDispatcher {
         // However, if the granularity is too small, we will end up with many
         // context switches if our drawing area is large (for a screen size of 1920x1080,
         // we will end up with around ~2000 regions). We therefore aim to choose a granularity such
-        // that no more than 50 chunks need to be processed by a single thread.
+        // that no more than 50 chunks need to be processed by a single thread. However, we also
+        // don't want to put too many regions in a single group (in this case 8) to prevent
         let granularity = {
             const CHUNKS_PER_THREAD: u32 = 50;
-            debug_assert_ne!(self.num_threads, 0);
+            const MIN_GRANULARITY: u32 = 1;
+            const MAX_GRANULARITY: u32 = 8;
 
             let num_regions = buffer.len() as u32;
+            // `MultiThreadedDispatcher` should never be created with just 0 threads.
+            assert!(self.num_threads > 0);
             let regions_per_thread = num_regions / self.num_threads as u32;
 
-            regions_per_thread.div_ceil(CHUNKS_PER_THREAD).min(1)
+            regions_per_thread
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -224,8 +224,38 @@ impl MultiThreadedDispatcher {
...
+        // be fine enough so in case the amount of work per region is very different, the work
+        // can still be distributed in a way so that threads don't end up being idle waiting for
+        // other threads to finish.
+        //
+        // However, if the granularity is too small, we will end up with many
+        // context switches if our drawing area is large (for a screen size of 1920x1080,
+        // we will end up with around ~2000 regions). We therefore aim to choose a granularity such
+        // that no more than 50 (more or less arbitrarily chosen) chunks need to be processed
+        // by a single thread. However, we also don't want to put too many regions in a
+        // single group (the currently chosen values i 8) to prevent stalls in case some of the
```
sagudev:
```suggestion
        // single group (the currently chosen values is 8) to prevent stalls in case some of the
```

LaurenzV:
I'm probably leaning towards not merging this as I'm really not able to empirically see a performance improvement, but still leaving this so we can discuss during office hours.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -224,8 +224,38 @@ impl MultiThreadedDispatcher {
...
+        // context switches if our drawing area is large (for a screen size of 1920x1080,
+        // we will end up with around ~2000 regions). We therefore aim to choose a granularity such
+        // that no more than 50 (more or less arbitrarily chosen) chunks need to be processed
+        // by a single thread. However, we also don't want to put too many regions in a
+        // single group (the currently chosen values is 8) to prevent stalls in case some of the
+        // later regions contain more work than the previous ones.
+        let granularity = {
+            const CHUNKS_PER_THREAD: u32 = 50;
+            const MIN_GRANULARITY: u32 = 1;
+            const MAX_GRANULARITY: u32 = 8;
```
grebmeg, line 244: Please consider my comment as just some thoughts — I haven’t looked deeply into the multithreading implementation yet.

First, would it make sense to make those parameters configurable? We could then benchmark with different values to find a good balance.

Second, as I understand, we already have some estimation for task workloads in path processing. However, since different regions may have varying computational complexity (as you mentioned), some groups could take significantly longer to process. Would it be feasible to combine regions into groups based on estimated complexity, up to a certain threshold, similar to how we use `COST_THRESHOLD` for tasks?

LaurenzV:
I think this is not worth having, so I'll close. Thanks for the comments, though!

PR #1116, opened 2025-07-21
Hoist `NaN` check prior to flattening

nicoburns:
I'm not sure if this is better (probably it makes no difference at all?). But @LaurenzV if you have an quick/easy way to check perf it might be worth checking it?


nicoburns:
Ack. I didn't make this against latest main.

PR #1118, opened 2025-07-22
[vello_cpu] Make layers symmetry assertions optional

### LINKED ISSUE #1117: Allow unpoped layers when rendering
What's the rational behind current limitation? This can commonly happen in 2d canvas.

usecase: Allow rendering scene while scene is still building.

sagudev:
fixes #1117

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -51,6 +51,7 @@ pub struct RenderContext {
         allow(dead_code, reason = "used when the `text` feature is enabled")
     )]
     pub(crate) level: Level,
+    pub(crate) ensure_all_layers_are_popped: bool,
     dispatcher: Box<dyn Dispatcher>,
 }
 
@@ -62,6 +63,8 @@ pub struct RenderSettings {
     /// The number of worker threads that should be used for rendering. Only has an effect
     /// if the `multithreading` feature is active.
     pub num_threads: u16,
+    /// Enable layers symmetry assertions
+    pub ensure_all_layers_are_popped: bool,
 }
 
 impl Default for RenderSettings {
@@ -75,6 +78,7 @@ impl Default for RenderSettings {
                 .saturating_sub(1) as u16,
             #[cfg(not(feature = "multithreading"))]
             num_threads: 0,
+            ensure_all_layers_are_popped: true,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -62,6 +63,8 @@ pub struct RenderSettings {
     /// The number of worker threads that should be used for rendering. Only has an effect
     /// if the `multithreading` feature is active.
     pub num_threads: u16,
+    /// Enable layers symmetry assertions
```
DJMcNab: Missing full stop. This is also extremely oblique - I'd like some more complete docs here.

For example:

You should generally set this to `true`.
By default, Vello CPU panics if you have not correctly matched each [`push_layer`] with a [`pop_layer`].
As missing pops can have non-local side effects (for example, clipping all content after a clip because you missed closing the clip context), this avoids users getting into hard to debug states.
Using `false` here disables that check, which can be useful for implementing APIs which don't put as much care into precise semantics, such as the 2d web canvas.
When this check is disabled, all open layers will be automatically closed at render time.

Obviously the intra-doc-links will need to be made to point to the right place if you use that wording.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -392,7 +415,9 @@ impl RenderContext {
     ) {
         // TODO: Maybe we should move those checks into the dispatcher.
         let wide = self.dispatcher.wide();
-        assert!(!wide.has_layers(), "some layers haven't been popped yet");
+        if self.ensure_all_layers_are_popped {
+            assert!(!wide.has_layers(), "some layers haven't been popped yet");
+        }
```
DJMcNab: I'd like there to be a test or two that this makes sense (i.e. performing some rendering with mismatched layer counts, then rendering again with the same context).

Because my assumption would be that this check has the potential for impact on rendering (for example, I was under the impression that certain blends would only happen on layer close).

DJMcNab: To expand on that - if this was always valid, I would probably lean towards making this a `debug_panic`, like we see in Masonry; i.e. panicking when debug assertions are enabled, and logging an error when not. That needn't happen in this PR, however.

(To clarify, I mean `debug_panic` when the new setting is `true`)

sparse_strips/vello_cpu/src/render.rs
sagudev: > When this check is disabled, all open layers will be automatically closed at render time.

I think this is not true, well at least per 2d spec it should not be. Unclosed are just treated as closed in render time, but that does not mean they are actually closed (it's still valid to do pop).

DJMcNab: Oh, so it's actually so that you can do a render halfway through creating the content, and then still keep going? And so you still need to pop all of the layers before your next frame or w/e?

sagudev: > Oh, so it's actually so that you can do a render halfway through creating the content, and then still keep going?

Yes, that's how 2d canvas works. One can build scene slower than it is rendered.

DJMcNab: Okay, so the doc comment should maybe say something like "cases where rendering an incomplete scene must be supported, primarily 2d canvas"?

sagudev:
Hm, okay sparse strips needs balanced (poped) layers (because that's where work is happening). It might be possible to migrate this (raqote does not need this), but that's too deep for me. So we will need to simulate this in servo by poping and pushing (we only need clips really).

Issue #1119, opened 2025-07-23
Nested clip blend does not work


sagudev:
`/html/canvas/element/compositing/2d.composite.clip.clear.html` and other `/html/canvas/element/compositing/2d.composite.clip.*.html` fail in servo because vello_cpu does not handle nested clip,blend correctly:
```rust
let rect = Rect::new(0.0, 0.0, 30.0, 30.0);
    // initial coloring
    ctx.set_paint(GREEN);
    ctx.fill_rect(&rect);
    // clip to rect
    {
        ctx.push_clip_layer(&Rect::new(-20., -20., 0., 0.).to_path(0.1));
        // clear outside the clip, this should be nop
        {
            ctx.push_blend_layer(Compose::Clear.into());
            ctx.fill_rect(&rect);
            ctx.pop_layer();
        }
        ctx.pop_layer();
    }
```

Outer clip should affect inner blend.

sagudev:
Actually even joining in one layer does not work:
```rust
ctx.push_layer(
            Some(&Rect::new(-20., -20., 0., 0.).to_path(0.1)),
            Some(Compose::Clear.into()),
            None,
            None,
        );
        ctx.fill_rect(&rect);
        ctx.pop_layer();
```

sagudev:
```rust
let rect = Rect::new(0.0, 0.0, 30.0, 30.0);
    // initial coloring
    ctx.set_paint(GREEN);
    ctx.fill_rect(&rect);
    // clip to rect
    {
        ctx.push_layer(
            Some(&Rect::new(-20., -20., 0., 0.).to_path(0.1)),
            None, //Some(Compose::Clear.into()),
            None,
            None,
        );
        ctx.set_paint(RED);
        ctx.fill_rect(&rect);
        ctx.pop_layer();
    }
```
works tho, so there is something sus about interaction blend<->clip.

LaurenzV:
I think there are multiple things going on here. I also had discussions with @dominikh before and I think there are definitely issues with the interaction between our clipping algorithm and blending.

However, in this particular case, the issue (or at least part of the issue) seems to be in the clipping algorithm: When determining which wide tiles have zero clipping, we currently only consider the bounding box of the current clipping path (which is zero in this case since it's completely outside of the viewport). However, I _think_ what should instead be used for determining the zero clipped wide tiles is the parent bbox (which in this case is the whole viewport, and thus all wide tiles would get a zero clip)? CC @grebmeg Perhaps my understanding is wrong, though.

[...]

PR #1120, opened 2025-07-23
Add test for (destructive) blend clip

sagudev:
Repro for https://github.com/linebender/vello/issues/1119

### DIFF
--- a/sparse_strips/vello_sparse_tests/snapshots/clip_clear.png
+++ b/sparse_strips/vello_sparse_tests/snapshots/clip_clear.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:5df0da52aaba923344202c1e30e377e11cce5ceef303d4dc1c460078637e908c
+size 98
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_sparse_tests/tests/clip.rs
```diff
@@ -305,6 +306,26 @@ fn clip_with_opacity(ctx: &mut impl Renderer) {
     ctx.pop_layer();
 }
 
+/// <https://github.com/linebender/vello/issues/1119>
+#[vello_test(ignore = "Currently fails")]
```
sagudev: vello_test does not support should_panic

sparse_strips/vello_sparse_tests/snapshots/clip_clear.png
LaurenzV, line 1: Wait why is there an image if the test fails?

sagudev: I generated correct image. So one knows what we expect when it will work correctly.

LaurenzV:
If it's okay, my preference would be to remove the additional clip path from the compose test cases (since tests are failing anyway), and just add a specific test case for that one once we get to that. But adding the new failing test case seems fine to me.

sagudev:
> If it's okay, my preference would be to remove the additional clip path from the compose test cases (since tests are failing anyway), and just add a specific test case for that one once we get to that. But adding the new failing test case seems fine to me.

Yeah, I was just interested to see if all others are really also bad.

LaurenzV (approved):

sagudev:
just rebase

PR #1121, opened 2025-07-23
Add Luminance based clipping

mTvare6:
### Requires [peniko#116](https://github.com/linebender/peniko/pull/116)

SVG has support for Luminance based clipping but vello doesn't support luminance based clipping. If the PR mentioned is merged, then this would allow the vello to use that.
https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/mask-type

### DIFF
--- a/vello_shaders/shader/shared/blend.wgsl
+++ b/vello_shaders/shader/shared/blend.wgsl
@@ -19,6 +19,7 @@ const MIX_HUE = 12u;
 const MIX_SATURATION = 13u;
 const MIX_COLOR = 14u;
 const MIX_LUMINOSITY = 15u;
+const MIX_LUMINANCE_CLIP = 64u;
 const MIX_CLIP = 128u;
 
 fn screen(cb: vec3<f32>, cs: vec3<f32>) -> vec3<f32> {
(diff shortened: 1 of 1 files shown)


LaurenzV:
In `vello_cpu`, you can already use `Mask::new_luminance`, I don't think more is needed here to implement what's required for SVG?

mTvare6:
> In `vello_cpu`, you can already use `Mask::new_luminance`, I don't think more is needed here to implement what's required for SVG?

I'm mainly concerned with classic vello as it in required in [GraphiteEditor/Graphite](https://github.com/GraphiteEditor/Graphite/pull/2808), so we would still require a new mix mode for classic vello, don't we?

LaurenzV:
I'm not familiar with vello classic, but that's possible, yes. Then I would suggest removing the parts from `vello_cpu`and just keeping the ones in `vello`.

nicoburns:
> In `vello_cpu`, you can already use `Mask::new_luminance`, I don't think more is needed here to implement what's required for SVG?

@LaurenzV In `vello_cpu`, what is the different between specifying a mask using the `mask` paramter to a layer vs. creating an extra later and using blend modes to achieve a masking effect (which I believe is the only available approach in vello classic)? Is there an advantage (performance?) to the `mask` parameter, or is it just a convenience?

LaurenzV:
I think this was discussed in the Zulip before, basically I wasn't aware you can use blending to simulate this effect, but in any case blending is super slow on the CPU, so I think having masks as an explicit special case makes sense there.

Keavon:
I haven't followed closely as to the technical aspects of this and how it aligns with the desired approach within Vello's engineering direction, but I did want to pop in and just mention that this is a blocker for Graphite in its current form (unless we can use a workaround of some sort).

PR #1122, opened 2025-07-24
[vello_hybrid] Add opacity layers

ajakubowicz-canva:
### Context

Add opacity handling in `vello_hybrid` layers!

### Discussion

It's probably best to look at the documentation in the shader code, but this PR leverages the fact that we have a huge amount of bits that do nothing. I've packed opacity into bits 0-7.

### Test plan

Tested via the existing opacity tests.

### DIFF
--- a/sparse_strips/vello_dev_macros/src/test.rs
+++ b/sparse_strips/vello_dev_macros/src/test.rs
@@ -163,7 +163,7 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
     skip_hybrid |= {
         input_fn_name_str.contains("compose")
             || input_fn_name_str.contains("gradient")
-            || input_fn_name_str.contains("layer")
+            || input_fn_name_str.contains("layer_multiple_properties")
             || input_fn_name_str.contains("mask")
             || input_fn_name_str.contains("mix")
             || input_fn_name_str.contains("blurred_rounded_rect")
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -562,8 +562,6 @@ impl Scheduler {
                         (0, nos.slot_ix as u16 * Tile::HEIGHT)
                     };
 
-                    // TODO: Quantizing the opacity. Maybe we shouldn't do this quantization? Test
-                    // pass.
                     let opacity_u8 = (tos.opacity * 255.0) as u32;
                     let paint = (COLOR_SOURCE_SLOT << 31) | (opacity_u8 & 0xFF);
 
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -181,6 +181,7 @@ use crate::{GpuStrip, RenderError, Scene};
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
 use core::mem;
+use vello_common::peniko::{BlendMode, Compose, Mix};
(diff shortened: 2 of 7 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -529,6 +533,50 @@ impl Scheduler {
...
+                    let round = nos.round.max(tos.round + usize::from(next_round));
+                    let draw = self.draw_mut(round, clip_depth - 1);
+                    let (x, y) = if clip_depth <= 2 {
+                        (wide_tile_x, wide_tile_y)
+                    } else {
+                        (0, nos.slot_ix as u16 * Tile::HEIGHT)
+                    };
+
+                    // TODO: Quantizing the opacity. Maybe we shouldn't do this quantization? Test
+                    // pass.
```
ajakubowicz-canva: I can probably delete this comment, since tests are passing.

sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -163,7 +163,7 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
     skip_hybrid |= {
         input_fn_name_str.contains("compose")
             || input_fn_name_str.contains("gradient")
-            || input_fn_name_str.contains("layer")
+            || input_fn_name_str.contains("layer_multiple_properties")
```
ajakubowicz-canva, line 166: By being more specific here, it enables tests to run such as:

 - `opacity_on_layer`
 - `opacity_nested_on_layer`

These tests fail without the changes in this PR.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva:
```suggestion
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -529,6 +535,49 @@ impl Scheduler {
...
+                    assert!(
+                        matches!(
+                            mode,
+                            BlendMode {
+                                mix: Mix::Normal,
+                                compose: Compose::SrcOver
+                            }
+                        ),
+                        "Changing blend mode is unsupported"
+                    );
```
ajakubowicz-canva, line 553: Note: I am fairly confident that supporting the various blend modes is basically going to be packing some more bits into the `paint`, and using those to choose the blend mode within the shader.

I can do that as a followup if desired.

grebmeg (approved):
LGTM

So good! 🎉 I played around a bit with the tests and different layers, and didn’t find any discrepancies with cpu. I’d suggest maybe adding an additional test for an image inside an opacity layer. Otherwise, the implementation looks super neat!

sparse_strips/vello_hybrid/src/schedule.rs
grebmeg: 👍 I think we can reuse a lot of code from [blend.wgsl](https://github.com/linebender/vello/blob/main/vello_shaders/shader/shared/blend.wgsl) in Vello Classic?

ajakubowicz-canva: I time-boxed 1 hour in attempting compositing / mix and there is an additional wrinkle. We can't exclusively pack more bits into `paint` because we currently implicitly are compositing using `SrcOver`. This means that we don't need to sample the destination to decide on the new destination result.

However, adding custom composite & mix requires being able to sample the src & destination to create a new destination. So there will need to be a bit more work done (than opacity) in order to accomplish this.

Issue #1124, opened 2025-07-24
Radial gradients do not render correctly


sagudev:
https://github.com/web-platform-tests/wpt/blob/master/html/canvas/element/fill-and-stroke-styles/2d.gradient.radial.inside3.html worked in vello (classic) but not in vello_cpu

classic code:
```rust
let g = Gradient::new_two_point_radial((50., 25.), 200., (50., 25.), 100.).with_stops([
        ColorStop {
            offset: 0.,
            color: Color::from_rgb8(255, 0, 0).into(),
        },
        ColorStop {
            offset: 0.993,
            color: Color::from_rgb8(255, 0, 0).into(),
        },
        ColorStop {
            offset: 1.,
            color: Color::from_rgb8(0, 255, 0).into(),
        },
    ]);
    scene.fill(
        vello::peniko::Fill::NonZero,
        Affine::IDENTITY,
        &g,
        None,
        &Rect::new(0., 0., 100., 50.),
    );
```
renders all green:

<img width="101" height="47" alt="Image" src="https://github.com/user-attachments/assets/5a4ca1e1-550c-4091-b6b9-c5bf23649379" />

but vello_cpu makes:

<img width="100" height="50" alt="Image" src="https://github.com/user-attachments/assets/54398a57-095b-4376-8557-73cc8fe62820" />

vello cpu code is in #1125

sagudev:
There are actually more radial failures in wpt: https://github.com/sagudev/servo/actions/runs/16462236523/job/46535424734, but some look like precision problems, although they might be fixed with this too.

Radial grad code needs to be rechecked anyway.

sagudev:
code: https://github.com/linebender/vello/blob/daf940230a24cbb123a458b6de95721af47aef98/sparse_strips/vello_common/src/encode.rs#L89

LaurenzV:
I have not validated this 100% but this patch seems to fix it (and makes sense to me): https://github.com/linebender/vello/compare/main...laurenz/grad_fix_attempt

Unfortunately, this requires regenerating nearly all of the gradient snapshot tests, so I definitely need to think through this more carefully before opening a PR.

sagudev:
> I have not validated this 100% but this patch seems to fix it (and makes sense to me): [main...laurenz/grad_fix_attempt](https://github.com/linebender/vello/compare/main...laurenz/grad_fix_attempt)
> 
[...]

PR #1125, opened 2025-07-24
Add test for radial gradient

sagudev:
repro for #1124

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/issues.rs
+++ b/sparse_strips/vello_sparse_tests/tests/issues.rs
@@ -287,8 +287,11 @@ fn intersected_clip_bbox_with_x0_gt_x1(ctx: &mut impl Renderer) {
 }
 
 // https://github.com/web-platform-tests/wpt/blob/master/html/canvas/element/fill-and-stroke-styles/2d.gradient.radial.inside3.html
-/// <https://github.com/linebender/vello/issues/1124>
-#[vello_test(no_ref, width = 100, height = 50)]
+#[vello_test(
+    ignore = "https://github.com/linebender/vello/issues/1124",
+    width = 100,
+    height = 50
+)]
 fn gradient_radial_wrong_color(ctx: &mut impl Renderer) {
     ctx.set_paint(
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -284,3 +286,26 @@ fn intersected_clip_bbox_with_x0_gt_x1(ctx: &mut impl Renderer) {
     ctx.pop_layer();
     ctx.pop_layer();
 }
+
+// https://github.com/web-platform-tests/wpt/blob/master/html/canvas/element/fill-and-stroke-styles/2d.gradient.radial.inside3.html
+/// <https://github.com/linebender/vello/issues/1124>
+#[vello_test(no_ref, width = 100, height = 50)]
+fn radial_grad(ctx: &mut impl Renderer) {
```
LaurenzV: Maybe a bit of a more descriptive name would be good, like `gradient_radial_wrong_color`?

sagudev:
> I _think_ we shouldn't include the reference images at all if they are wrong in the first place. Having a comment should be enough to indicate that the test case is currently not working. And is there a reason you can't just use `ignore` instead of `no_ref`? This would make it more explicit that we are ignoring it because it's not working, instead of it being a passing test but we don't need a reference image to test it.

I mistakenly committed an image. I checked how other failing test are doing and it seems to me that convention is to make them no_ref or wrong snapshot. Unfortunately vello_test does not support should_panic as we do in vello classic.

LaurenzV:
Indeed, there is no `should_panic`. I mean there aren't any specific guidelines, but my intention was for `no_ref` to be used for tests that don't require any reference images (because they mainly test that a specific input doesn't lead to a crash). I've used the `ignore` flag (see for example the test `image_with_transform_rotate_1`) to indicate that a test is purposefully being ignored for some reason.

sagudev:
Yeah, what bugs me is that ignore and should_panic (so test is expected to fail) are conceptually different. Vello classic has: https://github.com/linebender/vello/blob/main/vello_tests/tests/known_issues.rs.

LaurenzV:
I'll close this since I already intergrated the changes into the other PR. Thanks for the test cases!

PR #1127, opened 2025-07-24
Clampt vertical lines on the right of the viewport to the border

### LINKED ISSUE #1126: Large path produces (non-sensical?) out-of-viewport tiles
example:

```rs
#[test]
    fn large_path() {
        let path = BezPath::from_svg("M261,0 L78848,0 L78848,4 L261,4 Z").unwrap();
        let mut line_buf = vec![];
        fill(Level::new(), &path, Affine::IDENTITY, &mut line_buf);

        let mut tiles = Tiles::new();
        tiles.make_tiles(&line_buf, 10, 10);
        
        eprintln!("{:?}", tiles);
        panic!();
        
    }
```

Result:
```
Tiles { tile_buf: [Tile { packed_winding_line_idx: 2147483649, x: 19712, y: 0 }, Tile { packed_winding_line_idx: 2147483651, x: 65, y: 0 }], sorted: false }
```

LaurenzV:
Fixes #1126.

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -268,7 +268,10 @@ impl Tiles {
                 let y_top_tiles = (line_top_y as u16).min(tile_rows);
                 let y_bottom_tiles = (line_bottom_y.ceil() as u16).min(tile_rows);
 
-                let x = line_left_x as u16;
+                // Clamp all tiles that are strictly on the right of the viewport to the tile x coordinate
+                // right next to the outside of the viewport. If we don't do this, we might end up
+                // with too big tile coordinates, which will cause overflows in strip rendering.
+                let x = (line_left_x as u16).min(tile_columns + 1);
                 for y_idx in y_top_tiles..y_bottom_tiles {
                     let y = f32::from(y_idx);
 
@@ -326,7 +329,10 @@ impl Tiles {
 
 #[cfg(test)]
 mod tests {
-    use crate::flatten::{Line, Point};
+    use std::vec;
+    use fearless_simd::Level;
+    use crate::flatten::{fill, Line, Point};
+    use crate::kurbo::{Affine, BezPath};
     use crate::tile::{Tile, Tiles};
 
     #[test]
@@ -546,5 +552,19 @@ mod tests {
 
         let mut tiles = Tiles::new();
         tiles.make_tiles(&[line], 600, 600);
+    }
+
+    #[test]
+    fn vertical_path_on_the_right_of_viewport() {
+        let path = BezPath::from_svg("M261,0 L78848,0 L78848,4 L261,4 Z").unwrap();
+        let mut line_buf = vec![];
+        fill(Level::new(), &path, Affine::IDENTITY, &mut line_buf);
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&line_buf, 10, 10);
+        tiles.sort_tiles();
+
+        let tile = tiles.get(1);
+        assert_eq!(tile.x, 4)
     }
 }


sparse_strips/vello_common/src/tile.rs
```diff
@@ -547,4 +553,18 @@ mod tests {
...
+        let path = BezPath::from_svg("M261,0 L78848,0 L78848,4 L261,4 Z").unwrap();
+        let mut line_buf = vec![];
+        fill(Level::new(), &path, Affine::IDENTITY, &mut line_buf);
+
+        let mut tiles = Tiles::new();
+        tiles.make_tiles(&line_buf, 10, 10);
+        tiles.sort_tiles();
+
+        let tile = tiles.get(1);
+        assert_eq!(tile.x, 4);
```
tomcur:
```suggestion
        assert!(tiles.is_empty());
```

sparse_strips/vello_common/src/tile.rs
```diff
@@ -268,7 +268,10 @@ impl Tiles {
                 let y_top_tiles = (line_top_y as u16).min(tile_rows);
                 let y_bottom_tiles = (line_bottom_y.ceil() as u16).min(tile_rows);
 
-                let x = line_left_x as u16;
+                // Clamp all tiles that are strictly on the right of the viewport to the tile x coordinate
+                // right next to the outside of the viewport. If we don't do this, we might end up
+                // with too big tile coordinates, which will cause overflows in strip rendering.
+                let x = (line_left_x as u16).min(tile_columns + 1);
```
tomcur:
```suggestion
                // Do not emit tiles that are strictly on the right of the viewport. They do not
                // impact winding, and if we don't do this, we might end up with too big tile
                // coordinates, which will cause overflows in strip rendering.
                if line_left_x as u16 >= tile_columns {
                    continue;
                }

                let y_top_tiles = (line_top_y as u16).min(tile_rows);
                let y_bottom_tiles = (line_bottom_y.ceil() as u16).min(tile_rows);

                let x = line_left_x as u16;
```

tomcur (changes requested):
We can cull these tiles rather than clamping, as they do not impact winding. The branch below (the general case) culls these kinds of tiles. See also e.g. the `cull_line_at_right` test in this file.

sparse_strips/vello_common/src/tile.rs
LaurenzV: I could've sworn I tried this and it didn't work, but I guess I did something wrong 😄 thanks!

tomcur: I wasn't able to run all tests in `vello_sparse_tests` due to a SIGSEGV (one cause was fixed in https://github.com/linebender/vello/pull/1003 but apparently it has come back), but the ones I could run succeed.

tomcur (approved):

LaurenzV:
@tomcur Is it possible that `line_left_x as u16 >= tile_columns` should be `line_left_x as u16 > tile_columns` instead? Given the following code:

```
fn main() {
    let settings = RenderSettings {
        level: Level::new(),
        num_threads: 0,
        render_mode: RenderMode::OptimizeSpeed,
    };
    
    let mut ctx = RenderContext::new_with(556, 612, settings);
    let mut pixmap = Pixmap::new(556, 612);
    
    ctx.push_layer(Some(&BezPath::from_svg("M0,612 L556,612 L556,0 L0,0 Z").unwrap()), None, None, None);
    ctx.fill_path(&BezPath::from_svg("M0.35199928283691406,-0.72198486328125 L288.49700927734375,-0.72198486328125 L288.49700927734375,613.5119934082031 L0.35199928283691406,613.5119934082031 L0.35199928283691406,-0.72198486328125").unwrap());
    ctx.pop_layer();
    
    ctx.flush();
    
    // Encode the pixmap into a PNG file.
    let png = pixmap.into_png().unwrap();
    
    std::fs::write("output.png", png).unwrap();
}
```

Right now, I'm getting the following (note how in the very last row there are black pixels missing):
output

If I change the condition to `>`, I get the right result instead.

LaurenzV:
This fixes a lot of regressions in my integration tests, so i think this is the right change.

LaurenzV:
Hmm nope, I'm still seeing failures with that change. It does seem to work fine with my original implementation, however.

PR #1128, opened 2025-07-28
Two improvements to gradient handling.

LaurenzV:
This PR addresses the issues in #1124 and supersedes #1125.

There are two independent issues that were going on.

Firstly, there was an off-by-one error in our scaling factor for the t values. If our LUT size is 512, we want the values to be scaled between 0 and 511, and not 0 and 512. I actually already considered this when calculating the scale factor used in fine rasterization, but when computing the actual look-up table, we were using the wrong scale value. Unfortunately, this required regenerating nearly all gradient reference images, but I think that this should be more correct now (and we have an additional test case now that verifies this).

Secondly, we based the calculation of the lut size solely on the number of stops, but we didn't consider the position of the stops. If two stops are positioned very close to each other, we might have to increase the resolution. I still hardcoded a maximum limit of 4096 because I don't think it makes sense to go arbitrarily high there, but hopefully this should be good enough for all intents and purposes.


sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -284,3 +284,49 @@ fn intersected_clip_bbox_with_x0_gt_x1(ctx: &mut impl Renderer) {
     ctx.pop_layer();
     ctx.pop_layer();
 }
+
+// https://github.com/web-platform-tests/wpt/blob/master/html/canvas/element/fill-and-stroke-styles/2d.gradient.radial.inside3.html
```
grebmeg: [nit]: Would it be better to reference this (and below) using permalink?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -992,6 +987,42 @@ impl<T: FromF32Color> GradientLut<T> {
...
+
+    for x1 in ranges.iter().map(|e| e.x1) {
+        // For example, if the first stop is at 0.001, then we need a resolution of at least 1000
+        // so that we can still safely capture the first stop.
+        let res = ((1.0 / (x1 - last_x1)).ceil() as usize)
+            .min(MAX_LEN)
+            .next_power_of_two();
+        min_size = min_size.max(res);
+        last_x1 = x1;
+    }
```
grebmeg, line 1020: Just to clarify: would it be possible to calculate each range’s resolution, map it to the ramp range within a single function, and have that function return ramps? Essentially, combine this logic above with [that code](https://github.com/linebender/vello/blob/9aad729a92a28fbb96bc8ca3a824a648a597c2fa/sparse_strips/vello_common/src/encode.rs#L917-L929).

grebmeg: Feel free to ignore my previous question, because actually I don’t think it’s a good idea, since the current approach has a clear separation and performance difference is negligible.

sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -284,3 +284,49 @@ fn intersected_clip_bbox_with_x0_gt_x1(ctx: &mut impl Renderer) {
...
+// See <https://github.com/linebender/vello/issues/1124>.
+#[vello_test(width = 100, height = 50)]
+fn gradient_radial_wrong_color(ctx: &mut impl Renderer) {
```
grebmeg: Would it be possible to use slightly more descriptive names for both of tests?

PR #1133, opened 2025-07-29
Improvements to multi-threaded path rendering

LaurenzV:
This PR contains a number of improvements to multi-threaded rendering, including:

-  A much simpler and more effective cost estimation for path rendering.
- An optimization to prevent excessive allocations in case we are drawing many small paths.
- Splitting up the code into multiple files.
- Reusing of the alpha buffers.
- Adding more doc comments.

I have run a couple more experiments with regards to multi-threading, including:
- Doing coarse rasterization on worker threads by putting it behind a mutex.
- Doing coarse rasterization completely in parallel by splitting the canvas into "bands".

Unfortunately, at least from the experiments I've made so far, it only made things worse. I will do more experiments in that direction, but this PR contains a number of improvements that should be possible to land as is, and have a positive effect on performance (validated via benchmarking).

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -587,6 +587,11 @@ impl<T: Default> MaybePresent<T> {
     }
 
     pub(crate) fn take(&self) -> T {
+        assert!(
+            self.present.load(Ordering::SeqCst),
+            "Tried to access `MaybePresent` before initialization."
+        );
+
         let mut locked = self.value.lock().unwrap();
         self.present.store(false, Ordering::SeqCst);
         std::mem::take(&mut *locked)


sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -557,197 +554,41 @@ enum CoarseCommand {
...
-            }
-        }
-
-        Self {
-            num_line_segments,
-            num_curve_segments,
-            path_length,
-        }
+    pub(crate) fn take(&self) -> T {
+        let mut locked = self.value.lock().unwrap();
```
ajakubowicz-canva, line 595: Should `take` also have an assert?

```
assert!(
            self.present.load(Ordering::SeqCst),
            "Tried to access `MaybePresent` before initialization."
        );
```

sparse_strips/vello_cpu/src/dispatch/multi_threaded/small_path.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::kurbo::{BezPath, PathEl, PathSeg, segments};
+use smallvec::SmallVec;
+
+const SMALL_PATH_THRESHOLD: usize = 10;
```
ajakubowicz-canva, line 7: Why 10?

Would something like `4` still provide the benefit and reduce the "large_enum_variant" tradeoff? Non blocking - just querying this magic number especially as it has memory implications.

sparse_strips/vello_cpu/src/dispatch/multi_threaded/worker.rs
```diff
@@ -0,0 +1,162 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::Level;
+use crate::dispatch::multi_threaded::{CoarseTask, CoarseTaskSender, Path, RenderTask};
+use crate::peniko::Fill;
+use crate::strip_generator::StripGenerator;
+use std::prelude::rust_2015::Vec;
```
ajakubowicz-canva: Why are you using the `2015` version of the Rust standard library `Vec`?

sparse_strips/vello_cpu/src/dispatch/multi_threaded/worker.rs
```diff
@@ -0,0 +1,162 @@
...
+                RenderTask::PopLayer => {
+                    task_buf.push(CoarseTask::PopLayer);
+                }
+            }
+        }
+
+        result_sender.send(task_idx as usize, task_buf).unwrap();
+    }
+
+    pub(crate) fn finalize(&mut self) -> Vec<u8> {
```
ajakubowicz-canva, line 159: Is there any mechanism to enforce that `finalize` is only called once before `reset` is called?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -612,6 +612,7 @@ mod tests {
 
         ctx.push_clip_layer(&Rect::new(20.0, 20.0, 180.0, 180.0).to_path(0.1));
         ctx.pop_layer();
+        ctx.flush();
```
ajakubowicz-canva, line 615: Why was this added? Also – where are the assertions in these tests? Is it just that they don't crash?

ajakubowicz-canva (commented):
Awesome! I took a first pass, and am only really blocked on the `2015` Vec. As a side note, do you have documentation on the performance improvements this PR introduces?

sparse_strips/vello_cpu/src/render.rs
LaurenzV: This was added because otherwise it can happen that the test finishes before the worker threads actually finished, causing a panic when calling `send.unwrap()` in the worker thread. I can add a comment for this.

LaurenzV: Or I guess perhaps we shouldn't unwrap in the first place... Not sure.

sparse_strips/vello_cpu/src/dispatch/multi_threaded/small_path.rs
LaurenzV: Mainly so that for example rounded rectangles (which have curves in the corner) are also covered by this. 4 seems a bit too low to me.

LaurenzV:
> As a side note, do you have documentation on the performance improvements this PR introduces?

I haven't documented the improvements for this specific PR, but you can view the current numbers at https://laurenzv.github.io/vello_chart/.

sparse_strips/vello_cpu/src/dispatch/multi_threaded/worker.rs
LaurenzV: Auto-import messed up it seems 😅

sparse_strips/vello_cpu/src/dispatch/multi_threaded/worker.rs
LaurenzV: This shouldn't be a problem, as this is only an internal method, and we call this in `Dispatcher::init` only after we have initialized the worker once before.

PR #1136, opened 2025-07-29
Add the ability to disable anti-aliasing

LaurenzV:
This doesn't bring any performance benefit (though I'm pretty certain a better strip rendering algorithm could be developed for that), but still useful for anyone that needs the functionality.

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -38,9 +38,18 @@ pub fn render(
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
     fill_rule: Fill,
+    anti_aliasing: bool,
     lines: &[Line],
 ) {
-    render_dispatch(level, tiles, strip_buf, alpha_buf, fill_rule, lines);
+    render_dispatch(
+        level,
+        tiles,
+        strip_buf,
+        alpha_buf,
+        fill_rule,
+        anti_aliasing,
+        lines,
+    );
 }
 
 simd_dispatch!(render_dispatch(
@@ -49,6 +58,7 @@ simd_dispatch!(render_dispatch(
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
     fill_rule: Fill,
+    anti_aliasing: bool,
     lines: &[Line],
 ) = render_impl);
 
@@ -58,6 +68,7 @@ fn render_impl<S: Simd>(
     strip_buf: &mut Vec<Strip>,
     alpha_buf: &mut Vec<u8>,
     fill_rule: Fill,
+    anti_aliasing: bool,
     lines: &[Line],
 ) {
     strip_buf.clear();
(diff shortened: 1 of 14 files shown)


sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -96,8 +110,13 @@ impl Dispatcher for SingleThreadedDispatcher {
             // So just assign a dummy value here.
             let mut strip_buf = &[][..];
 
-            self.strip_generator
-                .generate_filled_path(c, fill_rule, clip_transform, |strips| strip_buf = strips);
+            self.strip_generator.generate_filled_path(
+                c,
+                fill_rule,
+                clip_transform,
+                true,
```
grebmeg: Should `anti_alias` also be passed here?
Below are the results of `running layer_multiple_properties_1` test with true/false on that line:

- `anti_alias: true`:
image

- `anti_alias: false`:
image

sparse_strips/vello_common/src/strip.rs
```diff
@@ -143,7 +154,17 @@ fn render_impl<S: Simd>(
             let p1 = s.combine_f32x4(location_winding[0], location_winding[1]);
             let p2 = s.combine_f32x4(location_winding[2], location_winding[3]);
 
-            alpha_buf.extend_from_slice(&f32_to_u8(s.combine_f32x8(p1, p2)).val);
+            let mut u8_vals = f32_to_u8(s.combine_f32x8(p1, p2));
+
+            if !anti_aliasing {
+                u8_vals = s.select_u8x16(
+                    u8_vals.simd_ge(u8x16::splat(s, 128)),
```
grebmeg, line 161: I’m wondering — was the 128 threshold chosen based on something specific? Would it be correct to say that setting the threshold to 1 would effectively give us conservative rasterization? More precisely, this would result in overestimated conservative rasterization, whereas to achieve underestimated rasterization, you’d need to consider whether the pixel center lies inside or outside the shape.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -143,7 +154,17 @@ fn render_impl<S: Simd>(
...
+                    u8x16::splat(s, 255),
+                    u8x16::splat(s, 0),
```
grebmeg, line 163: Is this approach more performant than computing those coverages earlier (after line 108) with branches based on `anti_aliasing`?

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Probably makes sense, yeah!

sparse_strips/vello_common/src/strip.rs
LaurenzV: yes, 128 was chosen because it's the middle, so everything with >= 50% is considered fully covered, while anything below is considered not covered at all.

Setting it to 1 would basically mean as long as the pixel is touched somehow, we consider it fully rasterized.

I'm not sure what's best here, but the current behavior seems sensible I think?

sparse_strips/vello_common/src/strip.rs
LaurenzV: If you mean within the for loop, then I would say yes, because this way we only have one branch, while if we put it above we have one branch for each iteration.

sparse_strips/vello_common/src/strip.rs
grebmeg: What I’m trying to say is that, from my understanding, there are two main categories: standard (precise) rasterization, which in our case means calculating coverage, and conservative rasterization, which can be of two types. One is overestimated, where a pixel is covered if any part of it intersects the primitive (effectively a threshold of 1), and the other is underestimated, where a pixel is covered only if its center lies inside the primitive. There may be other approaches as well I believe. So, what I’m suggesting here is to make the threshold (currently 128) configurable.

sparse_strips/vello_common/src/strip.rs
grebmeg: I meant somewhere in [this block](https://github.com/linebender/vello/pull/1136/files#diff-5eb613f8eeca1b6a6c023088e6d8dc600cc6dd481fbb32b72eb80a7b48d598fcR126-R133) of code. Though, if we want to support conservative rasterization as mentioned in the comment above, you’ll need to know whether the pixel intersects the primitive — so I think your current approach makes sense.

grebmeg (approved):
LGTM

I think it might make sense to make the aliasing threshold (which decides whether a pixel should be colored) configurable, to support conservative rasterization cases if needed.

sparse_strips/vello_common/src/strip.rs
LaurenzV: Yeah, that makes sense to me! If it's okay I would prefer adding this as a TODO though until someone actually needs this, because I'm not sure if anyone would really need this kind of granular control over this.

PR #1137, opened 2025-07-30
Path sparse strips caching / Recording API

grebmeg:
This PR follows up on our discussion during [this Renderer Office Hours](https://docs.google.com/document/d/1r28-uhSl3w9-MhPIfgt7uQXDxPBBVzkw25tlS3b6VMM/edit?tab=t.15ff7tiilrpq) (23/07/2025) regarding path sparse strips caching. We agreed to explore a potential API that would allow consumers to manage cached sparse strip data directly. As discussed in [another Renderer Office Hours](https://docs.google.com/document/d/1r28-uhSl3w9-MhPIfgt7uQXDxPBBVzkw25tlS3b6VMM/edit?tab=t.qiqbtp181ovc#heading=h.gsd3qibwjgrp) (30/07/25), we agreed to proceed with this.

This PR introduces the ability to cache path sparse strips, built on top of a new abstraction for capturing rendering commands and their associated data. This enables the ability to record and later replay rendering commands.

At the core of this system is a `Recorder`, a lightweight facade provided to user that just pushes rendering commands into a `Recording`. `Recording` owns the command list plus an optional `CachedStrips` (pre-baked strip geometry + alpha buffer). 

`Scene`/`RenderContext` gains recording support: it can create a recording (`Recordable::record`) and/or replay one (`Recordable::render_recording`). On later replays it re-uses those cached strips, only patching alpha indices and appending to the current alpha buffer.

**Not addressed in this PR (to be handled in follow-ups):**
- Multithreaded support
- Integer translation for sparse strips
The data below demonstrates this case, but there’s an issue with Y-misalignment that still needs to be resolved.

### Vello Hybrid
In the svg example, recording the svg rendering can be done by wrapping `render_svg` in `record_and_render` and passing `Recorder` instead of `Scene`. This captures `Recording`, which is then stored and reused on subsequent redraws.

During redraws, `try_reuse_recording` checks whether the current transform matches the one from the original recording: 
- If the transform is identical, the recording is reused as-is (`Identical`).
- If the transform is a pure translation (**currently experimental**), the system applies the translation to the sparse strips and reuses the updated recording (`Translated`).
- Otherwise (e.g., scaling, skewing), the recording is not reused, and a fresh render is required (`Fresh`).

Below are rough benchmarks using the SVG example in vello_hybrid, tested with the Ghostscript Tiger SVG. The test involves multiple consecutive redraws to demonstrate:
1. `Fresh` Rendering – the initial full render
2. `Identical` Replay – same transform (reuses cached data)
3. `Translated` Replay (**currently experimental**) – tiger is moved across the screen using a translation transform (reuses cached data)

- dev profile
image  

- release profile
image  

### Issues/Notes:
- Strips use u16 coordinates, so if a translation shifts coordinates into negative values, it causes a panic. Clipping cached strips may be needed to prevent this?
- It would be great to have a rendering abstraction and implement it for `Recorder`.


sparse_strips/vello_hybrid/src/recording.rs
LaurenzV, line 1: Is there anything speaking against moving it into `vello_common` so it can be used by `vello_cpu` later on as well?

grebmeg: Yeah, definitely — it’s part of the plan to support both renderers. I just started with `vello_hybrid` since it was the easiest place to implement first.

LaurenzV: Oh yeah, no worries, I can take care of implementing this on the `vello_cpu` side. But since we plan to do that, we might as well already move the file to vello_common I think. 😄

nicoburns:
Would it make sense to make path caching optional when recording a Recording (would this be difficult with your current design)? I can imagine there might be use cases where it is known that the cache won't be useful (e.g. creating a recording to be serialized to disk) and users might want to avoid the overhead.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+
+use crate::kurbo::{Affine, BezPath, Rect, Stroke};
+use crate::mask::Mask;
+use crate::paint::PaintType;
+use crate::peniko::{BlendMode, Fill};
+use crate::strip::Strip;
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+
+/// Cached sparse strip data
```
LaurenzV: [nit] but for consistency maybe we could preserve finishing doc comments with a full stop? 👀

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+
+/// A recording of rendering commands that can cache generated strips
+#[derive(Debug, Clone)]
+pub struct Recording {
+    /// Recorded commands
+    pub commands: Vec<RenderCommand>,
+    /// Cached sparse strips
+    pub cached_strips: Option<CachedStrips>,
+    /// Strip ranges for each geometry command: `(start_index, count)`
+    pub strip_ranges: Option<Vec<(usize, usize)>>,
```
LaurenzV: Would it make sense to have that field be part of `CachedStrips`.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+use alloc::vec::Vec;
+
+/// Cached sparse strip data
+#[derive(Debug, Clone)]
+pub struct CachedStrips {
+    /// The cached sparse strips
+    pub strips: Box<[Strip]>,
+    /// The alpha buffer data
+    pub alphas: Box<[u8]>,
+}
```
LaurenzV: Since there are methods that will clone the recording (e.g. `render_or_record`) and could be called multiple times, I wonder whether we should use an `Arc`  here instead?

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+}
+
+/// A recording of rendering commands that can cache generated strips
+#[derive(Debug, Clone)]
+pub struct Recording {
+    /// Recorded commands
+    pub commands: Vec<RenderCommand>,
+    /// Cached sparse strips
+    pub cached_strips: Option<CachedStrips>,
```
LaurenzV: I'm also wondering whether it's really necessary to store this in an Option. Can't we just handle it like `commands`, where if the vectors are empty, it meens it hasn't been initialized yet?

LaurenzV: I guess the problem is it's possible a sequence of instructions doesn't actually produce any visible output (and thus everything is empty),not sure if that's an edge case worth considering, though.

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
```diff
@@ -64,7 +64,7 @@ impl<'a, S: Simd> GradientPainter<'a, S> {
     }
 }
 
-impl<'a, S: Simd> Iterator for GradientPainter<'a, S> {
+impl<S: Simd> Iterator for GradientPainter<'_, S> {
```
LaurenzV, line 67: Is that from a clippy lint or?

LaurenzV: Same for the other changes

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -596,6 +599,253 @@ impl ColrRenderer for RenderContext {
...
+
+    /// Save the current rendering state
+    fn save_current_state(&self) -> RenderState {
+        RenderState {
+            transform: self.transform,
+            fill_rule: self.fill_rule,
+            stroke: self.stroke.clone(),
+            paint: self.paint.clone(),
+            paint_transform: self.paint_transform,
+            alphas: self.dispatcher.alpha_buf().to_vec(),
```
LaurenzV: I think that this is very problematic, because we don't want to clone the whole alpha buffer every time we create a new recording. Could we not do a `mem::swap` of the alpha buffer? This will require more creativity for multi-threading, but it should work for single-threaded.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -596,6 +599,253 @@ impl ColrRenderer for RenderContext {
...
+                    }
+                    RenderCommand::SetTransform(transform) => {
+                        self.set_transform(*transform);
+                    }
+                    RenderCommand::SetFillRule(fill_rule) => {
+                        self.set_fill_rule(*fill_rule);
+                    }
+                    RenderCommand::SetStroke(stroke) => {
+                        self.set_stroke(stroke.clone());
+                    }
```
LaurenzV: This does mean that replaying a recording will "pollute" the state of the renderer, right? If so and this is intentional, this should probably be documented.

LaurenzV (commented):
Just a couple of notes, but otherwise the CPU-side of things looks good to me.

Is my understanding correct though that this current version does not at all allow for caching objects with different translation factors? Was this discarded because it was too complicated?

sparse_strips/vello_cpu/src/render.rs
nicoburns: Are we still considering making the `RenderContext` non-stateful like vello classic (with regard to paint, transform, stroke, etc)?

LaurenzV: I think so.

sparse_strips/vello_common/src/recording.rs
grebmeg: I don’t mind that. Also, I think it would be nice to enforce all those kinds of style guide principles via CI clippy checks, right? — though I have a feeling that for this one in particular, it might require writing a custom rule.

Fixed.

sparse_strips/vello_common/src/recording.rs
grebmeg: Good question! The `Option` serves an important semantic purpose here - it distinguishes between "strips haven't been computed yet" (None) vs "strips have been computed and can be legitimately empty" (some with empty strips). Unlike **commands** where empty just means "no commands yet", cached strips need this three-state logic: uncomputed, computed-empty, computed-non-empty. This allows the `Recording` to be lazily evaluated from just the list of commands. If you’re okay with it and don’t see major concerns, I’d go with the current version. Open to discussion though.

sparse_strips/vello_common/src/recording.rs
grebmeg: Absolutely, thanks so much!

Fixed

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -368,3 +372,237 @@ impl GlyphRenderer for Scene {
...
+            paint: self.paint.clone(),
+            paint_transform: self.paint_transform,
+            stroke: self.stroke.clone(),
+            transform: self.transform,
+            fill_rule: self.fill_rule,
+            blend_mode: self.blend_mode,
+            strip_buf: self.strip_buf.clone(),
+            alphas: self.alphas.clone(),
+        }
+    }
```
```suggestion
    fn save_current_state(&mut self) -> RenderState {
        RenderState {
            paint: self.paint.clone(),
            paint_transform: self.paint_transform,
            stroke: std::mem::take(&mut self.stroke),
            transform: self.transform,
            fill_rule: self.fill_rule,
            blend_mode: self.blend_mode,
            strip_buf: std::mem::take(&mut self.strip_buf),
            alphas: std::mem::take(&mut self.alphas),
        }
    }
```

We don't need to pay clone cost. We can simply take the memory. If we take the memory, we might want to change the name to reflect that (perhaps `take_current_state` or similar)?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -368,3 +372,237 @@ impl GlyphRenderer for Scene {
...
+        alphas: &mut Vec<u8>,
+    ) {
+        self.make_strips(fill_rule);
+        // Adjust alpha indices for combined buffer
+        let alpha_offset = alphas.len() as u32;
+        for mut strip in self.strip_buf.iter().cloned() {
+            strip.alpha_idx += alpha_offset;
+            strips.push(strip);
+        }
+        alphas.extend_from_slice(&self.alphas);
```
```suggestion
        alphas.extend_from_slice(&self.alphas);
        self.alphas.clear();
```

Alphas isn't cleared like how strips are cleared before strip rendering as shown below. Instead they accumulate, so we need to clear here after each iteration otherwise we see over 3 million alphas for ghost tiger

https://github.com/taj-p/vello/blob/46182c5c17cff44bfec5fc32b7accc70ca8571ba/sparse_strips/vello_common/src/strip.rs#L74

image

taj-p: This reduces the time to render ghost tiger from ~23 ms back to ~13ms with 300k alphas (roughly the same time as without recording)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -368,3 +372,237 @@ impl GlyphRenderer for Scene {
...
+        let mut collected_strips = Vec::new();
+        let mut collected_alphas = Vec::new();
+        let mut strip_ranges = Vec::new();
+
+        for command in commands {
+            let start_index = collected_strips.len();
+
+            match command {
+                RenderCommand::FillPath(path) => {
+                    self.generate_fill_strips(path, &mut collected_strips, &mut collected_alphas);
```
taj-p: These `generate_*` commands are creating the strips _and_ copying the alphas into `collected_alphas`. This means that every command suffers a copy from `self.alphas` into `collected_alphas`.

Could we please instead return `self.alphas` as `collected_alphas` to avoid all those copies? So, for example:

```rs
        let mut collected_strips = Vec::new();
        let mut strip_ranges = Vec::new();

        for command in commands {
            let start_index = collected_strips.len();

            match command {
                RenderCommand::FillPath(path) => {
                    self.generate_fill_strips(path, &mut collected_strips, &mut collected_alphas);
                    let count = collected_strips.len() - start_index;
                    strip_ranges.push((start_index, count));
                }
                ...
            }
        }

        let alphas = std::mem::take(&mut self.alphas);
        self.restore_state(saved_state);
        (collected_strips, alphas, strip_ranges)
```

sparse_strips/vello_hybrid/src/scene.rs
taj-p: See above comment to avoid this per-command-copy entirely

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -368,3 +372,237 @@ impl GlyphRenderer for Scene {
...
+    fn make_strips_into_buffers(
+        &mut self,
+        fill_rule: Fill,
+        strips: &mut Vec<Strip>,
+        alphas: &mut Vec<u8>,
+    ) {
+        self.make_strips(fill_rule);
+        // Adjust alpha indices for combined buffer
+        let alpha_offset = alphas.len() as u32;
+        for mut strip in self.strip_buf.iter().cloned() {
```
taj-p: Can we drain or take the memory from `strip_buf` instead of cloning?

sparse_strips/vello_hybrid/src/scene.rs
taj-p: Hmmm I think this will also mean we don't need to `strip.alpha_idx += alpha_offset;` in `make_strips_into_buffers`? I think we could simply take the strips.

i.e., I was able to get this working:

```rs
    fn make_strips_into_buffers(&mut self, fill_rule: Fill, strips: &mut Vec<Strip>) {
        self.make_strips(fill_rule);
        strips.extend(self.strip_buf.drain(..));
    }
```

There's probably additional refactoring opportunities after this change.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -368,3 +372,237 @@ impl GlyphRenderer for Scene {
...
+                                let paint = self.encode_current_paint();
+                                self.wide.generate(
+                                    &adjusted_strips[start..end],
+                                    self.fill_rule,
+                                    paint,
+                                    0,
+                                );
+                            }
+
+                            range_index += 1;
```
```suggestion
                        assert!(range_index < strip_ranges.len());
                        let (start, count) = strip_ranges[range_index];
                        assert!(start < adjusted_strips.len() && count > 0);

                        let end = start + count;
                        let paint = self.encode_current_paint();
                        self.wide
                            .generate(&adjusted_strips[start..end], self.fill_rule, paint, 0);

                        range_index += 1;
```

It's better to panic than to silently fail when an invariant is disobeyed. Rust has bounds checks (unless we do an unchecked lookup), so I think it's better to just `start + count` instead of do the minimum check (and potentially silently fail).

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -368,3 +372,237 @@ impl GlyphRenderer for Scene {
...
+    ) -> Vec<Strip> {
+        // Calculate offset for alpha indices based on current buffer size
+        let alpha_offset = self.alphas.len() as u32;
+        // Create adjusted strips with corrected alpha indices
+        let mut adjusted_strips = Vec::with_capacity(cached_strips.len());
+        for strip in cached_strips {
+            let mut adjusted_strip = *strip;
+            adjusted_strip.alpha_idx += alpha_offset;
+            adjusted_strips.push(adjusted_strip);
+        }
```
taj-p: Rust will automatically reserve the right capacity if you use an iterator IIUC - that might be preferred to the manual capacity setting here

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    /// Get alphas as slice
+    pub fn alphas(&self) -> &[u8] {
+        &self.alphas
+    }
+}
+
+/// A recording of rendering commands that can cache generated strips
+#[derive(Debug, Clone)]
+pub struct Recording {
```
taj-p, line 82: Do we have a strategy so that consumers of `vello_cpu` and `vello_hybrid` cannot access internal-to-vello fields? Since `Recording`'s fields are public, we can access them entirely by consumers (this is a way for consumers to break invariants).

For example, a silly consumer can do this:

image


---

Don't need to solve this here - I'm just wondering whether it's something we've already thought about.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+    pub cached_strips: Option<CachedStrips>,
+    /// Strip ranges for each geometry command: `(start_index, count)`
+    pub strip_ranges: Option<Vec<(usize, usize)>>,
+    /// Last recorded transform
+    pub transform: Affine,
+}
+
+/// Individual rendering commands that can be recorded
+#[derive(Debug, Clone)]
+pub enum RenderCommand {
```
taj-p, line 93: Do we need to add glyph run?

sparse_strips/vello_common/src/recording.rs
taj-p: Is `render_or_record` the only place we clone `CachedStrips` or `Recording`? I'm wondering whether we should even allow `Recording` to be `clone`? When would it make sense as a consumer to do that? Maybe, we should push the consumer to `Arc` and instead remove this `Clone`?

Separately, I think `render_or_record` has pretty unexpected performance implications. If you're rendering an existing recording, the consumer will have to re-clone the entire recording (which could be huge) to satisfy the function signature:

```rs
    fn render_or_record<F>(&mut self, recording: Option<&mut Recording>, f: F) -> Recording
    where
        F: FnOnce(&mut Recorder<'_>),
    {
        if let Some(existing) = recording {
            self.render_recording(existing);
            existing.clone() // just cloning what is passed in
        } else {
            self.record_and_render(f)
        }
    }
```

I'm wondering whether we should omit this function or adjust its signature so consumers aren't surprised by it being relatively more expensive than expected.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// This method provides a convenient way to handle optional recordings.
+    /// If a recording is provided, it will be rendered. Otherwise, it will
+    /// record and render the provided commands.
+    fn render_or_record<F>(&mut self, recording: Option<&mut Recording>, f: F) -> Recording
+    where
+        F: FnOnce(&mut Recorder<'_>),
+    {
+        if let Some(existing) = recording {
+            self.render_recording(existing);
+            existing.clone()
```
taj-p: I think this line is unexpected from the consumers POV. This could be cloning 3 million alphas and cause significant performance implications for the consumer. I'm wondering whether we could adjust the signature to take an `Option<Recording>` so this isn't required or remove it completely (can consumers emulate this function themselves?)

sparse_strips/vello_hybrid/src/scene.rs
taj-p: If this isn't super clear, [these](https://github.com/linebender/vello/pull/1151/files#diff-cf782a179aa0fd954852f73458fb40f9a2e608c3b20eec0265acfc3e4dc284ccR389) are my scratch notes while I was reviewing this PR

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,410 @@
...
+    ///
+    /// # Example
+    /// ```ignore
+    /// let recording = scene.record(|ctx| {
+    ///     ctx.fill_rect(&Rect::new(0.0, 0.0, 100.0, 100.0));
+    ///     ctx.set_paint(Color::RED);
+    ///     ctx.stroke_path(&some_path);
+    /// });
+    /// ```
+    fn record<F>(&mut self, f: F) -> Recording
```
taj-p: Would it be possible to optionally overwrite an existing `Recording` so that we can re-use the allocations for these methods?

sparse_strips/vello_common/src/recording.rs
grebmeg: I think I can fix the issue by changing the function signature in `render_or_record` to take an owned `Recording`. That should resolve the problem, at least for now. That said, the function itself doesn’t seem particularly meaningful — I don’t see any compelling use cases for it. So instead, I’ll just remove it for now. Does that address your concern?

Updated.

sparse_strips/vello_common/src/recording.rs
grebmeg: I think that’s a good idea! I’ve added specific methods like `record_into` to support that kind of reuse.

Fixed.

sparse_strips/vello_common/src/recording.rs
grebmeg: I realized that as well after Laurenz’s comment and removed it. This can actually be handled on the consumer side, and to be honest, the method didn’t make much sense to me anyway.

Fixed

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
grebmeg: Yes, that’s correct — I’m not entirely sure why, but my local checks are flagging those lines as errors, while on github they seem to be just warnings. Would you like me to roll the changes back?

sparse_strips/vello_cpu/src/render.rs
grebmeg: I’ve added a note to `Recordable` trait.

Updated.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: All the suggestions above make sense, and after applying them, everything works as expected. Really nice fix — thanks for pointing it out, Taj!

Fixed.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Totally agree, that way’s better!

Fixed.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: It make sense, thanks!

Fixed

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Fixed

sparse_strips/vello_cpu/src/render.rs
grebmeg: It makes sense — Taj left a similar comment on `vello_hybrid` method as well, so I updated both to take the memory first and set it later using the currently available API in `StripGenerator`. We can update this in the future once multithreaded support is implemented.

Fixed.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Ah, now I understand what you meant on Friday! At the time, I was honestly surprised where all those alphas came from — but turns out they were from my own PR! 😅

I was aware that strips were cleared in the code, but during some cleanup, I accidentally removed the line that cleared alphas and didn’t notice it. You can see in the PR description that I initially had around 13ms per fresh render on Tiger.

Thanks a lot for pointing that out! I think with the current approach, where we take ownership of the alpha memory after generating all strips, we no longer need to explicitly clear alphas.

Fixed.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: This code has changed a bit — we’re now using `strips.append(&mut self.strip_buf)` here, which seems to give us good performance?

Alternatively, the most performant approach (for cpu implementation as well) might be to generate strips directly into an external buffer. That could be achieved by removing [`strip_buf.clear()`](https://github.com/linebender/vello/blob/b0e2e598ac62c7b3d04d8660e7b1b7659b596970/sparse_strips/vello_common/src/strip.rs#L74) inside `strip::render`, making the clearing an explicit step elsewhere, and updating `make_strips` signature to accept an external buffer.

I think this is feasible, but it would require a fair number of changes. Let me know if you’d like me to take that path. cc @LaurenzV

sparse_strips/vello_common/src/recording.rs
grebmeg: I agree — would making them private to be sufficient? I’ve gone ahead and made them private.

Updated.

sparse_strips/vello_common/src/recording.rs
grebmeg: Yes, we need that — I’ll address it in PR #1153

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
LaurenzV: Fine from my side. Perhaps those are from a new version which isn't part of CI yet.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: I'm fine leaving this as is for now if it makes it easier for now.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,506 @@
...
+
+/// Cached sparse strip data.
+#[derive(Debug)]
+pub struct CachedStrips {
+    /// The cached sparse strips.
+    strips: Box<[Strip]>,
+    /// The alpha buffer data.
+    alphas: Box<[u8]>,
+    /// Strip ranges for each geometry command: `(start_index, count)`.
+    strip_ranges: Box<[(usize, usize)]>,
```
LaurenzV: Theoretically it should be possible to just store the `start_index` and then infer the count from the next start index, no? But I'm fine with leaving this as is.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,506 @@
...
+/// # State Modification During Replay
+///
+/// **Important:** When replaying recordings using methods like `render_recording()`,
+/// the renderer's state (transform, paint, fill rule, stroke settings, etc.) will be
+/// modified to match the state changes captured in the recording. The renderer will
+/// be left in the final state after all commands have been executed.
+///
+/// If you need to preserve the original renderer state, save it before replaying
+/// recordings and restore it afterward.
+pub trait Recordable {
```
LaurenzV, line 253: Perhaps we can add a comment that this shouldn't be used (for now) when using vello_cpu with the multithreading feature?

sparse_strips/vello_common/src/recording.rs
grebmeg: Yes, you’re right — we can simply use the start indices.

Fixed.

sparse_strips/vello_common/src/recording.rs
grebmeg: Added.

sparse_strips/vello_common/src/recording.rs
taj-p: That use case allows the re-use of `Recording::commands` but not the cached_strips which are arguably larger and more costly to allocate. The alpha buffer and strips allocations are lost when we [clear](https://github.com/linebender/vello/blob/c5cef0b9f933c4ac89d0e07ac06e8d242da95988/sparse_strips/vello_common/src/recording.rs#L181).

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,509 @@
...
+    ///     ctx.fill_rect(&Rect::new(0.0, 0.0, 100.0, 100.0));
+    /// });
+    ///
+    /// // Reuse the same recording (preserves allocations)
+    /// scene.record_into(&mut recording, |ctx| {
+    ///     ctx.fill_rect(&Rect::new(50.0, 50.0, 150.0, 150.0));
+    ///     ctx.set_paint(Color::BLUE);
+    /// });
+    /// ```
+    fn record_into<F>(&mut self, recording: &mut Recording, f: F)
```
taj-p: I'm wondering also whether it's preferable to double the API surface area or require the consumer to create a recording before `record`.

There are currently a lot of recording methods. 10 is a lot. 

```rs
fn record(...)
fn record_into(...)
fn prepare_recording(...)
fn render_recording(...)
fn record_and_prepare(...)
fn record_and_prepare_into(...)
fn record_and_render(...)
fn record_and_render_into(...)
fn execute_recording(...) // should this be name "render_recording" to mimic "record_and_render"?
fn generate_strips_from_commands(...) // should this one be public?
```

I'm wondering whether we can reduce and simplify the surface area with just the following:

```rs
fn record(recording: &mut Recording)
fn prepare_recording(recording: &mut Recording)
fn render_recording(recording: &mut Recording)
```

So, a consumer needs to do the below to render:

```rs
let recording = Recording::new();
scene.record(&mut recording, |ctx| { ... });
scene.render_recording(recording);
```

And the below to prepare for later rendering:

```
```rs
let recording = Recording::new();
scene.record(&mut recording, |ctx| { ... });
scene.prepare_recording(recording);

// sometime later

scene.render_recording(recording);
```

---

Alternatively, we could use phantom types to ensure that a recording is prepared before it's rendered (and for any mutation of a prepared recording to return an unprepared recording)? But, it might be simpler to just `prepare` implicitly if a recording hasn't yet been prepared. I'm not sure.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,509 @@
...
+    ///     ctx.fill_rect(&Rect::new(0.0, 0.0, 100.0, 100.0));
+    /// });
+    /// // Strips are now generated and cached, allocations preserved
+    /// ```
+    fn record_and_prepare_into<F>(&mut self, recording: &mut Recording, f: F)
+    where
+        F: FnOnce(&mut Recorder<'_>),
+    {
+        self.record_into(recording, f);
+        self.prepare_recording(recording);
```
taj-p: There is an error with the `*_into` methods. If we render a first set of commands into a `Recording`, prepare it, then try to `record_into` a second set of commands into `Recording`, we will only render the first set I believe.

When I tried to test this with the below, I was met with a panic - I'm not sure if this is related or another issue:

```rs
    let mut recording = Recording::new();

    ctx.record_into(&mut recording, |ctx| {
        let mut path = BezPath::new();
        path.move_to((10.0, 10.0));
        path.line_to((180.0, 20.0));
        path.line_to((30.0, 40.0));
        path.close_path();

        // Use a combined transform that includes the root transform
        let scene_transform = Affine::scale(5.0);
        ctx.set_transform(root_transform * scene_transform);

        ctx.set_paint(palette::css::REBECCA_PURPLE);
        ctx.fill_path(&path);
        let stroke = Stroke::new(1.0);
        ctx.set_paint(palette::css::DARK_BLUE);
        ctx.set_stroke(stroke);
        ctx.stroke_path(&path);
    });

    ctx.record_into(&mut recording, |ctx| {
        let mut path = BezPath::new();
        path.move_to((10.0, 10.0));
        path.line_to((180.0, 20.0));
        path.line_to((30.0, 40.0));
        path.close_path();

        // Use a combined transform that includes the root transform
        let scene_transform = Affine::scale(10.0);
        ctx.set_transform(root_transform * scene_transform);

        ctx.set_paint(palette::css::REBECCA_PURPLE);
        ctx.fill_path(&path);
        let stroke = Stroke::new(1.0);
        ctx.set_paint(palette::css::DARK_BLUE);
        ctx.set_stroke(stroke);
        ctx.stroke_path(&path);
    });

```

image

(The first "Fresh" frame was from ghost tiger - I modified the "simple" scene)

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// Generate strips from strip commands and capture ranges.
+    ///
+    /// Returns:
+    /// - `collected_strips`: The generated strips.
+    /// - `collected_alphas`: The generated alphas.
+    /// - `strip_start_indices`: The start indices of strips for each geometry command.
+    fn generate_strips_from_commands(
+        &mut self,
+        commands: &[RenderCommand],
+    ) -> (Vec<Strip>, Vec<u8>, Vec<usize>);
```
taj-p: Should we just ask implementers to implement `prepare_recording` entirely so that we don't need to expose this method to consumers or do we think there's value in this method for public consumption?

sparse_strips/vello_common/src/recording.rs
taj-p: (Could we please add a unit test for this 🙏 )?

taj-p (changes requested):
Looks brilliant!

I've RC'd just to make sure we figure out the panic before proceeding. 

I've also left a suggestion re the API surface and on a seemingly logic error with the `*_into` methods.

sparse_strips/vello_common/src/recording.rs
taj-p: For the phantom data approach, if we want to prevent recording to a `Recording` after a prepare, then we can do something like:

```rs
struct Mutable;
struct Prepared;

struct Recording<State = Mutable> {
    _state: PhantomData<State>,
}

type PreparedRecording = Recording<Prepared>;

/// Record function that accepts a mutable reference to a Recording and returns a Recording
pub fn record(recording: &mut Recording<Mutable>) -> Recording<Mutable> {
    // ...

    // Return a new recording (could also return a clone or modified version)
    Recording {
        _state: PhantomData,
    }
}

/// Prepare function that takes a Recording and returns a PreparedRecording
pub fn prepare(recording: Recording<Mutable>) -> PreparedRecording {
    // ...

    Recording {
        _state: PhantomData,
    }
}

/// Render function that takes a PreparedRecording
pub fn render(prepared: &PreparedRecording) {
    //...
}

impl PreparedRecording {
    /// Clear function that returns a Recording (back to mutable state)
    pub fn clear(self) -> Recording<Mutable> {
        Recording {
            _state: PhantomData,
        }
    }
}
```

But, again, I'm honestly not too sure about this. I'll defer to you on whether using a pattern like this is worthwhile.

grebmeg: I agree with narrowing down the API to just the methods that are currently needed. We can always add more later if it makes sense to do so.

Regarding the pattern — to be honest, I’m also not fully convinced it’s the right fit here. It feels like it introduces extra complexity without providing clear benefits, especially when simple implicit calls might be sufficient. Personally, I’d prefer starting with a simpler approach for now, and only consider introducing the pattern if it proves necessary after we’ve used the API a bit more.

sparse_strips/vello_common/src/recording.rs
grebmeg: I’ve updated `CachedStrips` to use `Vec` instead of `Box` so that its allocated memory can be reused after clearing the data — hopefully this aligns with what you meant.

sparse_strips/vello_common/src/recording.rs
grebmeg: I think it would be more straightforward if consumers to implement it themselves.

Fixed.

sparse_strips/vello_common/src/recording.rs
grebmeg: Oh, yes — that was because in `prepare_recording` I had a condition to prepare the data only if `cached_strips` was empty. I’ve updated it and added a test, so it should work now.

Fixed.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,396 @@
...
+///
+/// # State Modification During Replay
+///
+/// **Important:** When replaying recordings using methods like `render_recording()`,
+/// the renderer's state (transform, paint, fill rule, stroke settings, etc.) will be
+/// modified to match the state changes captured in the recording. The renderer will
+/// be left in the final state after all commands have been executed.
+///
+/// If you need to preserve the original renderer state, save it before replaying
+/// recordings and restore it afterward.
```
taj-p: Is this part still true? I thought we save state before we create the cached strips and restore it afterwards?

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// scene.record(&mut recording, |ctx| {
+    ///     ctx.fill_rect(&Rect::new(0.0, 0.0, 100.0, 100.0));
+    /// });
+    ///
+    /// // Generate strips explicitly
+    /// scene.prepare_recording(&mut recording);
+    /// ```
+    fn prepare_recording(&mut self, recording: &mut Recording);
+
+    /// Render using a recording (caches strips on first use).
```
taj-p:
```suggestion
    /// Render using a recording.
```

That part isn't true - I think we pay the cost of re-preparing on each call

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,396 @@
...
+        self.prepare_recording(recording);
+        self.execute_recording(recording);
+    }
+
+    /// Execute a recording.
+    ///
+    /// This method executes a previously recorded sequence of operations.
+    /// It will generate and cache the necessary rendering data if it hasn't been done yet.
+    /// Subsequent calls will reuse the cached data for better performance.
+    fn execute_recording(&mut self, recording: &Recording);
```
taj-p:
```suggestion
    /// Render a recording (NOTE: recording must be previously prepared).
    fn render_recording(&mut self, recording: &Recording);
```

I think it's a bit confusing now. `render_recording` IIUC renders from scratch each call, but `execute_recording` only works if the strips were previously prepared via `prepare_recording` or `render_recording`. 

I think the existing `render_recording` method might introduce more confusion than convenience.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// Subsequent calls will reuse the cached data for better performance.
+    fn render_recording(&mut self, recording: &mut Recording) {
+        self.prepare_recording(recording);
+        self.execute_recording(recording);
+    }
+
+    /// Execute a recording.
+    ///
+    /// This method executes a previously recorded sequence of operations.
+    /// It will generate and cache the necessary rendering data if it hasn't been done yet.
```
taj-p:
```suggestion
```

I don't think this part is true anymore

sparse_strips/vello_common/src/recording.rs
taj-p: Not really - sorry I wasn't clearer. We still require allocating a large vector when preparing a recording to hold the alphas even though we probably have one already available within the recording itself. I was thinking something like this: https://github.com/linebender/vello/pull/1158/files

sparse_strips/vello_common/src/recording.rs
grebmeg: Yes, that’s correct. I was trying to say that you could do it as a consumer too, but with the current changes, you can’t — and I’m not sure it’s even necessary. I’ll remove that sentence.

Fixed.

sparse_strips/vello_common/src/recording.rs
grebmeg: Updated.

sparse_strips/vello_common/src/recording.rs
grebmeg: Updated.

sparse_strips/vello_common/src/recording.rs
grebmeg: Right, that makes sense — in my mind it felt like a convenient shortcut, but I can see how it could cause confusion given the current behavior. I’ll remove it for now and consider reintroducing it later if it becomes necessary.

Removed.

sparse_strips/vello_common/src/recording.rs
grebmeg: Got it, thanks for the suggestion! That makes sense now.

Updated.

grebmeg:
> cLGTM if we remove the helper method and re-use the alpha buf allocation via something like https://github.com/linebender/vello/pull/1158/files (if that makes sense! If it doesn't, let's chat!)

I removed the helper and re-used the alpha buf allocation as suggested.

PR #1141, opened 2025-07-30
Clarify docs of override image

### LINKED ISSUE #1138: Image overrides copy textures
When using image overrides I expected vello use texture directly, not to make copy of it.

https://github.com/linebender/vello/blob/65975d7612e38c199afaa547268763e1b4b826c9/vello/src/wgpu_engine.rs#L487

Is there any special reason for this?

DJMcNab:
Fixes #1138

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -508,7 +508,10 @@ impl Renderer {
 
     /// Overwrite `image` with `texture`.
     ///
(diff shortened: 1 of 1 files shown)


sagudev (commented):
I think we should also mention that because of this, it is also required for textures to have `TextureUsages::COPY_SRC`.

sagudev (approved):

PR #1142, opened 2025-07-31
Add tests for current state of alpha gradients

sagudev:


vello_tests/tests/known_issues.rs
DJMcNab, line 1: It isn't clear to me why this is a "known issue"; the test rendering is correct...

sagudev: Not if you are expecting unpremultiplied interpolation, but yeah it's not really good placement of test.
Anyway this test will be split into two tests. One for premultiplied and one for straight interpolation, when we will support both: https://github.com/linebender/vello/pull/1145 so I will fix it then.

PR #1143, opened 2025-07-31
Add a `take_png` method for pixmaps

LaurenzV:


sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -192,6 +192,18 @@ impl Pixmap {
         Ok(pixmap)
     }
 
+    /// Return the current content of the pixmap as a PNG.
+    #[cfg(feature = "png")]
+    pub fn take_png(self) -> Result<Vec<u8>, png::EncodingError> {
```
grebmeg: [nit]: What do you think about using `into_png` instead as the counterpart to `from_png` above? Personally, it sounds more idiomatic compared to "take", which leans toward `Option::take()` or `RefCell::take()`, implying "extract a component from a container leaving something" rather than "convert to another format".

grebmeg (approved):

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: Sounds good to me.

LaurenzV:
Ah whoops, I forgot to rename the PR. 😅 But I did rename the method.

PR #1140, opened 2025-07-30
Towards interpolating gradients in non-premultiplied space

LaurenzV:
This doesn't solve the issue for gradients with non-SRGB color space because the `color` crate will always provide a linear approximation in premultiplied space, but at least we can resolve the issue for SRGB color spaces, which seems like a good first step.

Fixes #1129.

PR #1145, opened 2025-08-02
[sparse strips] Support gradient interpolation in unpremultiplied alpha space

### LINKED ISSUE #1056: Gradient interpolation is done with premultiplied alpha
Currently, we are first premultiplying the colors of the gradients stops and then interpolating between those. Instead, what we need to do is to do interpolation using unpremultiplied colors, and then premultiply the result.

sagudev:
Some SIMD stuff is from https://github.com/linebender/vello/pull/1140

Depends on https://github.com/linebender/peniko/pull/115

This currently includes only Vello CPU/Hybrid implementation, but I plan to do classic Vello too.

Fixes #1056

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -17,8 +17,9 @@ use alloc::vec::Vec;
 #[cfg(not(feature = "multithreading"))]
 use core::cell::OnceCell;
 use core::hash::{Hash, Hasher};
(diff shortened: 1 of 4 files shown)


sagudev:
Diffing with firefox 2D canvas:
slika

Single dots are probably resist fingerprinting stuff, vertical lines are probably actual errors, although by the looks of it just computational.

PR #1149, opened 2025-08-06
Add support for SSE4.2 and AVX2

LaurenzV:
Note that we currently still don't have proper AVX2 instructions, but we do have SSE4 + FMA.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/common/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/common/image.rs
@@ -267,8 +267,8 @@ impl<S: Simd> Iterator for FilteredImagePainter<'_, S> {
             }
             ImageQuality::High => {
                 // Compare to <https://github.com/google/skia/blob/84ff153b0093fc83f6c77cd10b025c06a12c5604/src/opts/SkRasterPipeline_opts.h#L5030-L5075>.
-                let cx = weights(x_fract);
-                let cy = weights(y_fract);
+                let cx = self.simd.vectorize(|| weights(x_fract));
+                let cy = self.simd.vectorize(|| weights(y_fract));
 
                 const OFFSETS: [f32; 4] = [-1.5, -0.5, 0.5, 1.5];
 
--- a/sparse_strips/vello_cpu/src/fine/common/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/common/image.rs
@@ -443,41 +443,44 @@ pub(crate) fn extend<S: Simd>(
 
 /// Calculate the weights for a single fractional value.
 fn weights<S: Simd>(simd: S, fract: f32x4<S>) -> [f32x4<S>; 4] {
-    simd.vectorize(#[inline(always)] || {
-        let s = fract.simd;
-        const MF: [[f32; 4]; 4] = mf_resampler();
+    simd.vectorize(
+        #[inline(always)]
+        || {
+            let s = fract.simd;
+            const MF: [[f32; 4]; 4] = mf_resampler();
 
-        [
-            single_weight(
-                fract,
-                f32x4::splat(s, MF[0][0]),
-                f32x4::splat(s, MF[0][1]),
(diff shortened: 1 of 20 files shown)


DJMcNab:
> see Zulip discussion

My best guess is [#simd > target-feature enable in simd_dispatch macro](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/target-feature.20enable.20in.20simd_dispatch.20macro/with/533049372)? When referring to discussions on other platforms, it's a helpful habit to link to them.

LaurenzV:
This should mostly be ready, just need https://github.com/linebender/fearless_simd/pull/62.

Also note that while the title mentions "AVX2", it's really onlz SSE4 + FMA (assuming the compiler can't optimize those into AVX2), since we haven't actually implemented AVX2 in fearless_simd.

sparse_strips/vello_cpu/src/fine/common/image.rs
```diff
@@ -267,8 +267,14 @@ impl<S: Simd> Iterator for FilteredImagePainter<'_, S> {
             }
             ImageQuality::High => {
                 // Compare to <https://github.com/google/skia/blob/84ff153b0093fc83f6c77cd10b025c06a12c5604/src/opts/SkRasterPipeline_opts.h#L5030-L5075>.
-                let cx = weights(x_fract);
-                let cy = weights(y_fract);
+                let cx = self.simd.vectorize(
+                    #[inline(always)]
+                    || weights(x_fract),
+                );
```
DJMcNab: This still seems suspect to me. The closure which calls `weights` is still `inline(always)`, but the actual function `weights` isn't marked as that. (So I don't think that the body of weights will be inlined here).

Have you validated that the SIMD methods within weights get inlined properly? Any function you call in a `vectorize` call needs to itself be `inline(always)`.

I think what you might end up needing to do is to move this vectorize call to *inside* weights (see my other comment)

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -27,47 +27,63 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
...
+                for y in 0..Tile::HEIGHT {
+                    for (x, pixel) in region
+                        .row_mut(y)
+                        .chunks_exact_mut(COLOR_COMPONENTS)
+                        .enumerate()
+                    {
+                        let idx =
+                            COLOR_COMPONENTS * (usize::from(Tile::HEIGHT) * x + usize::from(y));
+                        let start = &blend_buf[idx..];
+                        // TODO: Use SIMD
```
DJMcNab:
```suggestion
                        // TODO: Use explicit SIMD
```

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -27,47 +27,63 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
...
     fn copy_solid(simd: S, dest: &mut [Self::Numeric], src: [Self::Numeric; 4]) {
-        let color = f32x16::block_splat(src.simd_into(simd));
-
-        for el in dest.chunks_exact_mut(16) {
-            el.copy_from_slice(&color.val);
-        }
+        simd.vectorize(
+            #[inline(always)]
+            || {
+                let color = f32x16::block_splat(src.simd_into(simd));
```
DJMcNab, line 63: Oh, interesting. This is a neat pattern I hadn't appreciated before.

sparse_strips/vello_cpu/src/fine/common/image.rs
LaurenzV: Will move it, good catch.

LaurenzV:
I've updated the benchmarks with results from my x86 machine: https://laurenzv.github.io/vello_chart/

All of the Rust crates were compiled with `target-feature=+avx2` to make the comparison fair.

PR #1153, opened 2025-08-11
[feat] Add glyph runs recording commands

grebmeg:


sparse_strips/vello_common/src/recording.rs
```diff
@@ -114,6 +120,38 @@ pub enum RenderCommand {
...
+        font_size: f32,
+        /// Per-glyph transform.
+        glyph_transform: Option<Affine>,
+        /// Whether font hinting is enabled.
+        hint: bool,
+        /// Normalized variation coordinates for variable fonts.
+        normalized_coords: Vec<NormalizedCoord>,
+        /// The glyphs to stroke.
+        glyphs: Vec<Glyph>,
+    },
```
nicoburns: Would it make sense to:

- Merge these two variants
- Add a `Style` field to disambiguate fill vs. stroke

?

nicoburns: I would also recommend creating a dedicated struct for these fields and making the enum variant a simple wrapper for that struct (rather than using a struct-like enum variant). Struct-like enum variants are hard to work with as you can't easily pass the values from the just variant around, take references to them, etc.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -114,6 +120,38 @@ pub enum RenderCommand {
...
+        /// Font to use for the glyphs.
+        font: Font,
+        /// Font size in pixels per em.
+        font_size: f32,
+        /// Per-glyph transform.
+        glyph_transform: Option<Affine>,
+        /// Whether font hinting is enabled.
+        hint: bool,
+        /// Normalized variation coordinates for variable fonts.
+        normalized_coords: Vec<NormalizedCoord>,
```
LaurenzV: Does it make sense to use a smallvec here?

sparse_strips/vello_common/src/recording.rs
grebmeg: Thank you, that all makes sense!

Fixed.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -412,4 +448,91 @@ impl<'a> Recorder<'a> {
...
+    #[cfg(feature = "text")]
+    pub fn glyph_run(&mut self, font: &Font) -> RecorderGlyphRunBuilder<'_> {
+        RecorderGlyphRunBuilder::new(font.clone(), self.recording)
+    }
+}
+
+/// A glyph run builder specifically for recording.
+#[cfg(feature = "text")]
+#[derive(Debug)]
+pub struct RecorderGlyphRunBuilder<'a> {
```
LaurenzV: I haven't tried this, but couldn't we just implement `GlyphRenderer` for `Recording` and then reuse the existing `GlyphRunBuilder`?

grebmeg: I didn’t go this route because the current `GlyphRenderer` processes glyphs immediately, whereas recording needs to capture the glyph commands for later execution. But let me actually give that a try.

sparse_strips/vello_common/src/recording.rs
grebmeg: I think it’s resolved now?

sparse_strips/vello_common/src/recording.rs
grebmeg: Your suggestion works perfectly and fits better into the system. Thank you for pointing that out!

Fixed.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -3,9 +3,19 @@
...
 //! Recording API for caching sparse strips
 
+#[cfg(feature = "text")]
+use crate::glyph::GlyphRenderer;
+#[cfg(feature = "text")]
+use crate::glyph::GlyphRunBuilder;
+#[cfg(feature = "text")]
+use crate::glyph::GlyphType;
+#[cfg(feature = "text")]
+use crate::glyph::PreparedGlyph;
```
LaurenzV: Maybe we can group those imports?

sparse_strips/vello_common/src/recording.rs
```diff
@@ -124,6 +138,12 @@ pub enum RenderCommand {
     SetPaintTransform(Affine),
     /// Reset the paint transform.
     ResetPaintTransform,
+    /// Render a fill outline glyph.
+    #[cfg(feature = "text")]
+    FillOutlineGlyph((BezPath, Affine)),
+    /// Render a stroke outline glyph.
+    #[cfg(feature = "text")]
+    StrokeOutlineGlyph((BezPath, Affine)),
```
LaurenzV, line 140: It should be possible to also support other types of glyph by storing a  `Vec` of render commands, right? But doesn't have to be as part of this PR, just wondering.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -124,6 +138,12 @@ pub enum RenderCommand {
     SetPaintTransform(Affine),
     /// Reset the paint transform.
     ResetPaintTransform,
+    /// Render a fill outline glyph.
+    #[cfg(feature = "text")]
+    FillOutlineGlyph((BezPath, Affine)),
+    /// Render a stroke outline glyph.
```
LaurenzV, line 138: Maybe `filled` and `stroked`?

sparse_strips/vello_hybrid/examples/scenes/src/text.rs
```diff
@@ -91,6 +130,83 @@ impl Default for TextScene {
...
+}
+
+/// Check if two transforms are identical (within tolerance)
+fn transforms_are_identical(a: Affine, b: Affine) -> bool {
+    let a_coeffs = a.as_coeffs();
+    let b_coeffs = b.as_coeffs();
+    let tolerance = 1e-6;
+
+    for i in 0..6 {
+        if (a_coeffs[i] - b_coeffs[i]).abs() > tolerance {
```
LaurenzV, line 191: I think you could use `abs_diff` if you want.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -481,21 +465,37 @@ impl Scene {
...
+                        self.transform * *transform,
+                    );
+                    strip_start_indices.push(start_index);
+                }
+                RenderCommand::StrokeOutlineGlyph((path, transform)) => {
+                    self.generate_stroke_strips(
+                        path,
+                        &mut collected_strips,
+                        self.transform * *transform,
+                    );
```
LaurenzV: I think it should be possible to write:

```rs
#[cfg(feature = "text")]
RenderCommand::FillOutlineGlyph((path, transform)) | RenderCommand::StrokeOutlineGlyph((path, transform))  => {
    let glyph_transform = self.transform * *transform;
    self.generate_fill_strips(
        path,
        &mut collected_strips,
        glyph_transform,
        &mut strip_generator,
    );
    strip_start_indices.push(start_index);
}
```

LaurenzV (approved):

sparse_strips/vello_common/src/recording.rs
grebmeg: Good call! I’ll group them.

Fixed.

sparse_strips/vello_common/src/recording.rs
grebmeg: From my perspective, the current naming is more appropriate because it describes commands using verbs that express actions. For example, `RenderCommand::FillOutlineGlyph` reads as an instruction, with associated data specifying what’s needed to execute the command unlike `RenderCommand::FilledOutlineGlyph`, which sounds like a description of an already completed action.

sparse_strips/vello_common/src/recording.rs
grebmeg: Yeah, I think so. We could store it as a flattened list of commands, or as rendering commands within a single command. Alternatively, we might come up with a solution for nested recordings, which would let you record a COLR glyph into it. Another option could be to skip storing the list altogether, pre-render it to an image, and then store COLR and bitmap glyphs in the same way. I believe it will be much clearer once we add support for Bitmap and COLR glyphs to `vello_hybrid`.

sparse_strips/vello_hybrid/examples/scenes/src/text.rs
grebmeg: Doesn’t the [abs_diff](https://doc.rust-lang.org/std/primitive.u32.html?search=abs_diff) method only exist for integer types in Rust?

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Fixed.

sparse_strips/vello_hybrid/examples/scenes/src/text.rs
nicoburns: Correct https://doc.rust-lang.org/stable/std/?search=abs_diff

taj-p (approved):
Nice!

PR #1155, opened 2025-08-11
[WIP] Spatiotemporal Blending for vello_hybrid

ajakubowicz-canva:
This is still very much a work in progress.

Technically this currently passes all the existing tests, but it currently has a serious limitation. It is too parallel, and will place dependent blends into the same draw call.

There is also a kitchen sink test that I am using to weed out bugs. I will probably end up removing that test once I've split it up into smaller test cases. A test fuzzer would be awesome.

Status:
 - All existing tests for compositing pass.
 - Clip tests pass with the new system.
 - I've left Mix out, because I want to focus on getting scheduler change rock solid before doing color math.

Feeling good with this draft. What is quite funny is I think the solution to my problem is to use Taj's design of pausing processing tiles for a given round at a Blend, flushing a round then iterating over the rest of the wide tiles, detailed in [#vello > Hybrid Blending @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Hybrid.20Blending/near/532618735)

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -900,24 +900,24 @@ impl WideTile {
                 let (_, tail) = self.cmds.split_at(self.cmds.len() - 3);
 
                 let updated = match tail {
-                    [Cmd::PushBuf, Cmd::AlphaFill(a), Cmd::Blend(b)] => {
-                        if !b.is_destructive() && a.blend_mode.is_none() {
-                            let mut blended = a.clone();
-                            blended.blend_mode = Some(*b);
-                            Some(Cmd::AlphaFill(blended))
-                        } else {
-                            None
-                        }
-                    }
-                    [Cmd::PushBuf, Cmd::Fill(a), Cmd::Blend(b)] => {
-                        if !b.is_destructive() && a.blend_mode.is_none() {
-                            let mut blended = a.clone();
-                            blended.blend_mode = Some(*b);
-                            Some(Cmd::Fill(blended))
-                        } else {
-                            None
-                        }
-                    }
+                    // [Cmd::PushBuf, Cmd::AlphaFill(a), Cmd::Blend(b)] => {
+                    //     if !b.is_destructive() && a.blend_mode.is_none() {
+                    //         let mut blended = a.clone();
+                    //         blended.blend_mode = Some(*b);
+                    //         Some(Cmd::AlphaFill(blended))
+                    //     } else {
+                    //         None
+                    //     }
+                    // }
+                    // [Cmd::PushBuf, Cmd::Fill(a), Cmd::Blend(b)] => {
+                    //     if !b.is_destructive() && a.blend_mode.is_none() {
+                    //         let mut blended = a.clone();
+                    //         blended.blend_mode = Some(*b);
+                    //         Some(Cmd::Fill(blended))
+                    //     } else {
+                    //         None
+                    //     }
+                    // }
                     _ => None,
                 };
 
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -837,18 +837,8 @@ impl Scheduler {
                     state.stack.last_mut().unwrap().opacity = *opacity;
                 }
                 Cmd::Blend(mode) => {
-                    // This blend mode is implicitly supported. Currently no other blend mode is
-                    // supported in `vello_hybrid`.
-                    assert!(
-                        matches!(
-                            mode,
-                            BlendMode {
-                                mix: Mix::Normal,
-                                compose: Compose::SrcOver
-                            }
-                        ),
-                        "Changing blend mode is unsupported"
-                    );
+                    // Some blend modes and mixes are not supported.
+                    assert!(matches!(mode.mix, Mix::Normal), "Only Mix::Normal is supported currently");
 
                     let tos = state.stack.last().unwrap();
                     let nos = &state.stack[state.stack.len() - 2];
@@ -878,10 +868,13 @@ impl Scheduler {
                             tos.dest_slot.get_idx() as u32 | ((temp_slot.get_idx() as u32) << 16);
 
                         // Pack opacity, mix_mode, and compose_mode into paint
-                        let paint = (COLOR_SOURCE_BLEND << 30)
-                            | (opacity_u8 << 16)  // opacity (bits 16-23)
-                            | (0 << 8)            // mix_mode = MIX_NORMAL (bits 8-15)
-                            | 3; // compose_mode = COMPOSE_SRC_OVER (bits 0-7)
+                        // Extract the actual mix and compose values from mode
+                        let mix_mode = mode.mix as u32; // Assuming Mix has repr(u8) like Compose
+                        let compose_mode = mode.compose as u32;
+
+                        let paint = (COLOR_SOURCE_BLEND << 30) | (opacity_u8 << 16)           // opacity (bits 16-23)
+                            | (mix_mode << 8)              // mix_mode (bits 8-15)
+                            | compose_mode; // compose_mode (bits 0-7)
 
                         draw.0.push(GpuStrip {
                             x,
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -241,9 +241,37 @@ struct TileState {
     stack: Vec<TileEl>,
 }
 
+#[derive(Debug, Copy, Clone)]
+enum ClaimedSlot {
+    Texture0(usize),
+    Texture1(usize),
+}
+
+impl ClaimedSlot {
+    fn get_idx(&self) -> usize {
+        match self {
+            ClaimedSlot::Texture0(idx) => *idx,
+            ClaimedSlot::Texture1(idx) => *idx,
+        }
+    }
+
+    fn get_texture(&self) -> usize {
+        match self {
+            ClaimedSlot::Texture0(_) => 0,
+            ClaimedSlot::Texture1(_) => 1,
+        }
+    }
+}
+
 #[derive(Clone, Copy, Debug)]
 struct TileEl {
-    slot_ix: usize,
+    /// `dest_slot` represents the final location of where the contents of this tile will end up. If
+    /// there are layers, then painting happens to `temporary_slot` such that blending can be done
+    /// to the dest_slot with another slot.
+    dest_slot: ClaimedSlot,
+    /// Temporary slot only used if we know there is another layer pushed above this tile element.
+    /// INVARIANT: This slot and the dest_slot are always on opposite textures.
+    temporary_slot: Option<ClaimedSlot>,
     round: usize,
     opacity: f32,
 }
@@ -264,6 +292,17 @@ impl Scheduler {
             clear,
             rounds_queue: Default::default(),
             tile_state: Default::default(),
+        }
+    }
+
+    fn claim_free_slot(&mut self, texture: usize) -> ClaimedSlot {
+        let slot_ix = self.free[texture].pop().unwrap();
+        self.clear[texture].push(slot_ix as u32);
+
+        match texture {
+            0 => ClaimedSlot::Texture0(slot_ix),
+            1 => ClaimedSlot::Texture1(slot_ix),
+            _ => panic!("invalid slot texture"),
         }
     }
 
@@ -377,7 +416,8 @@ impl Scheduler {
         // Sentinel `TileEl` to indicate the end of the stack where we draw all
         // commands to the final target.
         state.stack.push(TileEl {
-            slot_ix: usize::MAX,
+            dest_slot: ClaimedSlot::Texture0(usize::MAX),
+            temporary_slot: None,
             round: self.round,
             opacity: 1.,
         });
@@ -397,7 +437,7 @@ impl Scheduler {
                 });
             }
         }
-        for cmd in &tile.cmds {
+        for (cmd_idx, cmd) in tile.cmds.iter().enumerate() {
             // Note: this starts at 1 (for the final target)
             let clip_depth = state.stack.len();
             match cmd {
@@ -412,7 +452,7 @@ impl Scheduler {
                     let (x, y) = if clip_depth == 1 {
                         (scene_strip_x, scene_strip_y)
                     } else {
-                        (fill.x, el.slot_ix as u16 * Tile::HEIGHT)
+                        (fill.x, el.dest_slot.get_idx() as u16 * Tile::HEIGHT)
                     };
 
                     draw.0.push(GpuStrip {
@@ -443,7 +483,7 @@ impl Scheduler {
                     let (x, y) = if clip_depth == 1 {
                         (scene_strip_x, scene_strip_y)
                     } else {
-                        (alpha_fill.x, el.slot_ix as u16 * Tile::HEIGHT)
+                        (alpha_fill.x, el.dest_slot.get_idx() as u16 * Tile::HEIGHT)
                     };
 
                     draw.0.push(GpuStrip {
@@ -464,10 +504,10 @@ impl Scheduler {
                         }
                         self.flush(renderer);
                     }
-                    let slot_ix = self.free[ix].pop().unwrap();
-                    self.clear[ix].push(slot_ix as u32);
+                    let slot = self.claim_free_slot(ix);
                     state.stack.push(TileEl {
-                        slot_ix,
+                        dest_slot: slot,
+                        temporary_slot: None,
                         round: self.round,
                         opacity: 1.,
                     });
@@ -484,8 +524,9 @@ impl Scheduler {
                         round - self.round < self.rounds_queue.len(),
                         "round must be in queue"
                     );
-                    self.rounds_queue[round - self.round].free[1 - clip_depth % 2]
-                        .push(tos.slot_ix);
+
+                    self.rounds_queue[round - self.round].free[tos.dest_slot.get_texture()]
+                        .push(tos.dest_slot.get_idx());
                 }
                 Cmd::ClipFill(clip_fill) => {
                     let tos = &state.stack[clip_depth - 1];
@@ -496,17 +537,20 @@ impl Scheduler {
                     let (x, y) = if clip_depth <= 2 {
                         (wide_tile_x + clip_fill.x as u16, wide_tile_y)
                     } else {
-                        (clip_fill.x as u16, nos.slot_ix as u16 * Tile::HEIGHT)
+                        (
+                            clip_fill.x as u16,
+                            nos.dest_slot.get_idx() as u16 * Tile::HEIGHT,
+                        )
                     };
                     // Opacity packed into the first 8 bits – pack full opacity (0xFF).
-                    let paint = COLOR_SOURCE_SLOT << 31 | 0xFF;
+                    let paint = COLOR_SOURCE_SLOT << 30 | 0xFF;
                     draw.0.push(GpuStrip {
                         x,
                         y,
                         width: clip_fill.width as u16,
                         dense_width: 0,
                         col_idx: 0,
-                        payload: tos.slot_ix as u32,
+                        payload: tos.dest_slot.get_idx() as u32,
                         paint,
                     });
                 }
@@ -519,10 +563,13 @@ impl Scheduler {
                     let (x, y) = if clip_depth <= 2 {
                         (wide_tile_x + clip_alpha_fill.x as u16, wide_tile_y)
                     } else {
-                        (clip_alpha_fill.x as u16, nos.slot_ix as u16 * Tile::HEIGHT)
+                        (
+                            clip_alpha_fill.x as u16,
+                            nos.dest_slot.get_idx() as u16 * Tile::HEIGHT,
+                        )
                     };
                     // Opacity packed into the first 8 bits – pack full opacity (0xFF).
-                    let paint = COLOR_SOURCE_SLOT << 31 | 0xFF;
+                    let paint = COLOR_SOURCE_SLOT << 30 | 0xFF;
                     draw.0.push(GpuStrip {
                         x,
                         y,
@@ -531,7 +578,7 @@ impl Scheduler {
                         col_idx: (clip_alpha_fill.alpha_idx / usize::from(Tile::HEIGHT))
                             .try_into()
                             .expect("Sparse strips are bound to u32 range"),
-                        payload: tos.slot_ix as u32,
+                        payload: tos.dest_slot.get_idx() as u32,
                         paint,
                     });
                 }
@@ -561,12 +608,12 @@ impl Scheduler {
                     let (x, y) = if clip_depth <= 2 {
                         (wide_tile_x, wide_tile_y)
                     } else {
-                        (0, nos.slot_ix as u16 * Tile::HEIGHT)
+                        (0, nos.dest_slot.get_idx() as u16 * Tile::HEIGHT)
                     };
 
                     // Opacity packed into the first 8 bits.
                     let opacity_u8 = (tos.opacity * 255.0) as u32;
-                    let paint = (COLOR_SOURCE_SLOT << 31) | opacity_u8;
+                    let paint = (COLOR_SOURCE_SLOT << 30) | opacity_u8;
 
                     draw.0.push(GpuStrip {
                         x,
@@ -574,7 +621,7 @@ impl Scheduler {
                         width: WideTile::WIDTH,
                         dense_width: 0,
                         col_idx: 0,
-                        payload: tos.slot_ix as u32,
+                        payload: tos.dest_slot.get_idx() as u32,
                         paint,
                     });
                 }
@@ -598,7 +645,7 @@ impl Scheduler {
                     has_non_zero_alpha(rgba),
                     "Color fields with 0 alpha are reserved for clipping"
                 );
-                let paint_packed = (COLOR_SOURCE_PAYLOAD << 31) | (PAINT_TYPE_SOLID << 29);
+                let paint_packed = (COLOR_SOURCE_PAYLOAD << 30) | (PAINT_TYPE_SOLID << 29);
                 (rgba, paint_packed)
             }
             Paint::Indexed(indexed_paint) => {
@@ -609,9 +656,9 @@ impl Scheduler {
                 match scene.encoded_paints.get(paint_id) {
                     Some(EncodedPaint::Image(encoded_image)) => match &encoded_image.source {
                         ImageSource::OpaqueId(_) => {
-                            let paint_packed = (COLOR_SOURCE_PAYLOAD << 31)
-                                | (PAINT_TYPE_IMAGE << 29)
-                                | (paint_tex_id & 0x1FFFFFFF);
+                            let paint_packed = (COLOR_SOURCE_PAYLOAD << 30)
+                                | (PAINT_TYPE_IMAGE << 28)
+                                | (paint_tex_id & 0x0FFFFFFF);
                             let scene_strip_xy =
                                 ((scene_strip_y as u32) << 16) | (scene_strip_x as u32);
                             (scene_strip_xy, paint_packed)
@@ -623,6 +670,18 @@ impl Scheduler {
             }
         }
     }
+}
+
+/// Look ahead to see if it's possible that blending will happen.
+fn should_use_temporary_slot(cmds: &[Cmd], current_idx: usize) -> bool {
+    for cmd in &cmds[current_idx + 1..] {
+        match cmd {
+            Cmd::PushBuf => return true,
+            Cmd::PopBuf => return false, // End of this buffer's scope
+            _ => continue,
+        }
+    }
+    false
 }
 
 const COLOR_SOURCE_PAYLOAD: u32 = 0;
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -286,7 +286,11 @@ struct TileEl {
 impl TileEl {
     fn get_draw_texture(&self, buffer_depth: usize) -> usize {
         if buffer_depth > 1 {
-            self.dest_slot.get_texture()
+            if let Some(temp_slot) = self.temporary_slot {
+                temp_slot.get_texture()
+            } else {
+                self.dest_slot.get_texture()
+            }
         } else {
             2
         }
@@ -374,7 +378,7 @@ impl Scheduler {
     ///
     /// The rounds queue must not be empty.
     fn flush<R: RendererBackend>(&mut self, renderer: &mut R) {
-        let round = self.rounds_queue.pop_front().unwrap();
+        let round = dbg!(self.rounds_queue.pop_front().unwrap());
 
         #[cfg(debug_assertions)]
         {
@@ -391,8 +395,17 @@ impl Scheduler {
                     2 => "Final Output",
                     _ => "Unknown",
                 };
+                let sampling = match target_idx {
+                    0 => "Texture 1 (odd clip buffer)",
+                    1 => "Texture 0 (even clip buffer)",
+                    2 => "Texture 1 (odd clip buffer)",
+                    _ => "Unknown",
+                };
 
-                println!("  Rendering to: {}", target);
+                println!(
+                    "  Rendering to: {} by sampling slots from {}",
+                    target, sampling
+                );
 
                 for (_, strip) in draw.0.iter().enumerate() {
                     if target_idx < 2 {
@@ -514,10 +527,6 @@ impl Scheduler {
             println!();
         }
         for (i, draw) in round.draws.iter().enumerate() {
-            if draw.0.is_empty() {
-                continue;
-            }
-
             let load = {
                 if i == 2 {
                     // We're rendering to the view, don't clear.
@@ -534,6 +543,11 @@ impl Scheduler {
                     LoadOp::Load
                 }
             };
+
+            if draw.0.is_empty() {
+                continue;
+            }
+
             renderer.render_strips(&draw.0, i, load);
         }
         for i in 0..2 {
@@ -572,7 +586,7 @@ impl Scheduler {
         });
         {
             // If the background has a non-zero alpha then we need to render it.
-            let bg = tile.bg.as_premul_rgba8().to_u32();
+            let bg = dbg!(tile).bg.as_premul_rgba8().to_u32();
             if has_non_zero_alpha(bg) {
                 let draw = self.draw_mut(self.round, 2);
                 draw.0.push(GpuStrip {
@@ -601,7 +615,12 @@ impl Scheduler {
                     let (x, y) = if clip_depth == 1 {
                         (scene_strip_x, scene_strip_y)
                     } else {
-                        (fill.x, el.dest_slot.get_idx() as u16 * Tile::HEIGHT)
+                        let slot_idx = if let Some(temp_slot) = el.temporary_slot {
+                            temp_slot.get_idx()
+                        } else {
+                            el.dest_slot.get_idx()
+                        };
+                        (fill.x, slot_idx as u16 * Tile::HEIGHT)
                     };
 
                     draw.0.push(GpuStrip {
@@ -632,7 +651,12 @@ impl Scheduler {
                     let (x, y) = if clip_depth == 1 {
                         (scene_strip_x, scene_strip_y)
                     } else {
-                        (alpha_fill.x, el.dest_slot.get_idx() as u16 * Tile::HEIGHT)
+                        let slot_idx = if let Some(temp_slot) = el.temporary_slot {
+                            temp_slot.get_idx()
+                        } else {
+                            el.dest_slot.get_idx()
+                        };
+                        (alpha_fill.x, slot_idx as u16 * Tile::HEIGHT)
                     };
 
                     draw.0.push(GpuStrip {
@@ -681,6 +705,8 @@ impl Scheduler {
 
                     self.rounds_queue[round - self.round].free[tos.dest_slot.get_texture()]
                         .push(tos.dest_slot.get_idx());
+                    // If a TileEl was not used for blending the temporary slot may still be in use
+                    // and needs to be cleared.
                     if let Some(temp_slot) = tos.temporary_slot {
                         self.rounds_queue[round - self.round].free[temp_slot.get_texture()]
                             .push(temp_slot.get_idx());
@@ -689,11 +715,33 @@ impl Scheduler {
                 Cmd::ClipFill(clip_fill) => {
                     let tos: &TileEl = &state.stack[clip_depth - 1];
                     let nos = &state.stack[clip_depth - 2];
+
                     // Basically if we are writing onto the even texture, we need to go up a round
                     // to target it.
                     let next_round = clip_depth % 2 == 0 && clip_depth > 2;
                     let round = nos.round.max(tos.round + usize::from(next_round));
-                    let draw = self.draw_mut(round, nos.get_draw_texture(clip_depth - 1));
+                    if let Some(temp_slot) = nos.temporary_slot {
+                        let draw = self.draw_mut(round, nos.dest_slot.get_texture());
+                        let paint: u32 = COLOR_SOURCE_SLOT << 30 | 0xFF; // Full opacity copy
+                        draw.0.push(GpuStrip {
+                            x: 0,
+                            y: nos.dest_slot.get_idx() as u16 * Tile::HEIGHT,
+                            width: WideTile::WIDTH,
+                            dense_width: 0,
+                            col_idx: 0,
+                            payload: temp_slot.get_idx() as u32,
+                            paint,
+                        });
+                    }
+
+                    let draw = self.draw_mut(
+                        round,
+                        if (clip_depth - 1) <= 1 {
+                            2
+                        } else {
+                            nos.dest_slot.get_texture()
+                        },
+                    );
                     let (x, y) = if clip_depth <= 2 {
                         (wide_tile_x + clip_fill.x as u16, wide_tile_y)
                     } else {
@@ -713,13 +761,45 @@ impl Scheduler {
                         payload: tos.dest_slot.get_idx() as u32,
                         paint,
                     });
+
+                    let nos_ptr = state.stack.len() - 2;
+                    let _ = &mut state.stack[nos_ptr].temporary_slot.take().iter().for_each(
+                        |temp_slot| {
+                            self.rounds_queue[round - self.round].free[temp_slot.get_texture()]
+                                .push(temp_slot.get_idx());
+                        },
+                    );
                 }
                 Cmd::ClipStrip(clip_alpha_fill) => {
                     let tos = &state.stack[clip_depth - 1];
                     let nos = &state.stack[clip_depth - 2];
+
                     let next_round = clip_depth % 2 == 0 && clip_depth > 2;
                     let round = nos.round.max(tos.round + usize::from(next_round));
-                    let draw = self.draw_mut(round, nos.get_draw_texture(clip_depth - 1));
+
+                    // If nos has a temporary slot, copy it to dest_slot first
+                    if let Some(temp_slot) = nos.temporary_slot {
+                        let draw = self.draw_mut(round, nos.dest_slot.get_texture());
+                        let paint = COLOR_SOURCE_SLOT << 30 | 0xFF;
+                        draw.0.push(dbg!(GpuStrip {
+                            x: 0,
+                            y: nos.dest_slot.get_idx() as u16 * Tile::HEIGHT,
+                            width: WideTile::WIDTH,
+                            dense_width: 0,
+                            col_idx: 0,
+                            payload: temp_slot.get_idx() as u32,
+                            paint,
+                        }));
+                    }
+
+                    let draw = self.draw_mut(
+                        round,
+                        if (clip_depth - 1) <= 1 {
+                            2
+                        } else {
+                            nos.dest_slot.get_texture()
+                        },
+                    );
                     let (x, y) = if clip_depth <= 2 {
                         (wide_tile_x + clip_alpha_fill.x as u16, wide_tile_y)
                     } else {
@@ -730,7 +810,7 @@ impl Scheduler {
                     };
                     // Opacity packed into the first 8 bits – pack full opacity (0xFF).
                     let paint = COLOR_SOURCE_SLOT << 30 | 0xFF;
-                    draw.0.push(GpuStrip {
+                    draw.0.push(dbg!(GpuStrip {
                         x,
                         y,
                         width: clip_alpha_fill.width as u16,
@@ -740,7 +820,15 @@ impl Scheduler {
                             .expect("Sparse strips are bound to u32 range"),
                         payload: tos.dest_slot.get_idx() as u32,
                         paint,
-                    });
+                    }));
+
+                    let nos_ptr = state.stack.len() - 2;
+                    let _ = &mut state.stack[nos_ptr].temporary_slot.take().iter().for_each(
+                        |temp_slot| {
+                            self.rounds_queue[round - self.round].free[temp_slot.get_texture()]
+                                .push(temp_slot.get_idx());
+                        },
+                    );
                 }
                 Cmd::Opacity(opacity) => {
                     state.stack.last_mut().unwrap().opacity = *opacity;
@@ -761,30 +849,72 @@ impl Scheduler {
 
                     let tos = state.stack.last().unwrap();
                     let nos = &state.stack[state.stack.len() - 2];
+                    dbg!(tos, nos);
 
-                    let next_round = clip_depth % 2 == 0 && clip_depth > 2;
+                    let next_round: bool = clip_depth % 2 == 0 && clip_depth > 2;
                     let round = nos.round.max(tos.round + usize::from(next_round));
 
-                    let draw = self.draw_mut(round, nos.get_draw_texture(clip_depth - 1));
-                    let (x, y) = if clip_depth <= 2 {
-                        (wide_tile_x, wide_tile_y)
+                    if let Some(temp_slot) = nos.temporary_slot {
+                        // let draw = self.draw_mut(round, tos.get_draw_texture(clip_depth - 1));
+
+                        // I think this can be nos.dest_slot
+                        let draw = self.draw_mut(
+                            round,
+                            if clip_depth <= 2 {
+                                2
+                            } else {
+                                nos.dest_slot.get_texture()
+                            },
+                        );
+                        let (x, y) = if clip_depth <= 2 {
+                            (wide_tile_x, wide_tile_y)
+                        } else {
+                            (0, nos.dest_slot.get_idx() as u16 * Tile::HEIGHT)
+                        };
+
+                        // Blend tos.dest_slot with nos.temporary_slot into nos.dest_slot
+                        let paint = (COLOR_SOURCE_BLEND << 30) | ((temp_slot.get_idx() as u32) << 16)  // dest_slot (14 bits)
+                            | (0 << 8)  // mix_mode = MIX_NORMAL (8 bits)  
+                            | 3; // compose_mode = COMPOSE_SRC_OVER (8 bits)
+
+                        draw.0.push(GpuStrip {
+                            x,
+                            y,
+                            width: WideTile::WIDTH,
+                            dense_width: 0,
+                            col_idx: 0,
+                            payload: tos.dest_slot.get_idx() as u32, // src_slot
+                            paint,
+                        });
+
+                        // Invalidate the temporary slot after use
+                        let nos_ptr = state.stack.len() - 2;
+                        let temporary_slot =
+                            &mut state.stack[nos_ptr].temporary_slot.take().unwrap();
+                        self.rounds_queue[round - self.round].free[temporary_slot.get_texture()]
+                            .push(temporary_slot.get_idx());
                     } else {
-                        (0, nos.dest_slot.get_idx() as u16 * Tile::HEIGHT)
-                    };
+                        // let draw = self.draw_mut(round, tos.get_draw_texture(clip_depth - 1));
+                        let draw = self.draw_mut(round, tos.get_draw_texture(clip_depth - 1));
+                        let (x, y) = if (clip_depth - 1) <= 2 {
+                            (wide_tile_x, wide_tile_y)
+                        } else {
+                            (0, nos.dest_slot.get_idx() as u16 * Tile::HEIGHT)
+                        };
+                        // No temporary slot - just copy with opacity
+                        let opacity_u8 = (tos.opacity * 255.0) as u32;
+                        let paint = (COLOR_SOURCE_SLOT << 30) | opacity_u8;
 
-                    // Opacity packed into the first 8 bits.
-                    let opacity_u8 = (tos.opacity * 255.0) as u32;
-                    let paint = (COLOR_SOURCE_SLOT << 30) | opacity_u8;
-
-                    draw.0.push(GpuStrip {
-                        x,
-                        y,
-                        width: WideTile::WIDTH,
-                        dense_width: 0,
-                        col_idx: 0,
-                        payload: tos.dest_slot.get_idx() as u32,
-                        paint,
-                    });
+                        draw.0.push(GpuStrip {
+                            x,
+                            y,
+                            width: WideTile::WIDTH,
+                            dense_width: 0,
+                            col_idx: 0,
+                            payload: tos.dest_slot.get_idx() as u32,
+                            paint,
+                        });
+                    }
                 }
                 _ => unimplemented!(),
             }
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -181,7 +181,7 @@ use crate::{GpuStrip, RenderError, Scene};
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
 use core::mem;
-use vello_common::peniko::{BlendMode, Compose, Mix};
+use vello_common::peniko::Mix;
 use vello_common::{
     coarse::{Cmd, WideTile},
     encode::EncodedPaint,
@@ -885,6 +885,10 @@ impl Scheduler {
                             payload,
                             paint,
                         });
+                        // TODO(DO NOT SUBMIT): There is a problem with blend dependencies. Currently
+                        // the system has no problem parallelizing dependent blends. Instead, I
+                        // think we need to signal after a blend that we need a new round to
+                        // continue working on the tile. After flushing we can continue.
 
                         // Invalidate the temporary slot after use
                         let nos_ptr = state.stack.len() - 2;
@@ -893,6 +897,7 @@ impl Scheduler {
                         self.rounds_queue[round - self.round].free[temporary_slot.get_texture()]
                             .push(temporary_slot.get_idx());
                     } else {
+                        // TODO(DO NOT SUBMIT): This path is very obviously wrong - and needs to not exist.
                         let draw = self.draw_mut(round, tos.get_draw_texture(clip_depth - 1));
                         let (x, y) = if (clip_depth - 1) <= 2 {
                             (wide_tile_x, wide_tile_y)
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -189,6 +189,13 @@ use vello_common::{
     tile::Tile,
 };
 
+const COLOR_SOURCE_PAYLOAD: u32 = 0;
+const COLOR_SOURCE_SLOT: u32 = 1;
+const COLOR_SOURCE_BLEND: u32 = 2;
+
+const PAINT_TYPE_SOLID: u32 = 0;
+const PAINT_TYPE_IMAGE: u32 = 1;
+
 /// Trait for abstracting the renderer backend from the scheduler.
 pub(crate) trait RendererBackend {
     /// Clear specific slots in a texture.
@@ -276,6 +283,16 @@ struct TileEl {
     opacity: f32,
 }
 
+impl TileEl {
+    fn get_draw_texture(&self, buffer_depth: usize) -> usize {
+        if buffer_depth > 1 {
+            self.dest_slot.get_texture()
+        } else {
+            2
+        }
+    }
+}
+
 #[derive(Debug, Default)]
 struct Draw(Vec<GpuStrip>);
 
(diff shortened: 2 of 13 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -340,6 +505,11 @@ impl Scheduler {
                     LoadOp::Load
                 }
             };
+
+            if draw.0.is_empty() {
+                continue;
+            }
```
ajakubowicz-canva: We must still be able to clear slots even if no draws are scheduled.

ajakubowicz-canva:
Update on this PR. It's close, but there are still a bunch of details to figure out. For the test cases (plus the additional ones), this works. However, we are still failing at:
 - What if we run out of slots? Can suspended tiles complete their work efficiently.
 - Blending directly onto the view - this requires pushing an extra buffer because we cannot blend more than once onto the final view.

Then there are some easy low effort performance wins - like I think I can remove the linear scan of wide tiles. Due to the two dot points above, I've demoted this back to a draft.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -894,43 +894,6 @@ impl WideTile {
             // we can just pop it instead.
             self.cmds.pop();
         } else {
-            if self.cmds.len() >= 3 {
```
ajakubowicz-canva, line 897: Removing this optimization matches what was done in https://github.com/linebender/vello/pull/1159/files#diff-5e16077335b3a3c3ed1845456ec5716c0368870c15a5cea692c03ad428d207a5L897-L933

I've got this in this PR as well because it's essential for vello_hybrid blending to work.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1069,68 @@ impl Scheduler {
...
     rgba >= 0x1_00_00_00
 }
+
+/// Does a single linear scan over the wide tile commands to prepare them for `do_tile`. Notably
+/// this function:
+///  - Precomputes the layers that require temporary slots due to blending.
+///
+/// TODO: Can be triggered via a const generic on coarse draw cmd generation which will avoid
+/// linear scans.
+fn prepare_cmds<'a>(cmds: &'a [Cmd]) -> Vec<AnnotatedCmd<'a>> {
```
ajakubowicz-canva, line 1134: This function _can_ be totally removed, but I'd almost call that a performance optimization. If acceptable it might be nice to leave it as-is and addressed later once I have a better understanding of:
 - Why webgl is not working.
 - A solution to deadlocks.

sparse_strips/vello_cpu/examples/wasm_cpu/src/scenes/mod.rs
```diff
@@ -55,6 +56,7 @@ pub(crate) fn get_example_scenes() -> Box<[AnyScene]> {
         AnyScene::new(text::TextScene::new("Hello World from vello_cpu")),
         AnyScene::new(clip::ClipScene::new()),
         AnyScene::new(image::ImageScene::new()),
+        AnyScene::new(blend::BlendScene::new()),
```
taj-p, line 58: nit: Maybe best to place this before/after the clip scene

sparse_strips/vello_hybrid/examples/scenes/src/lib.rs
```diff
@@ -81,6 +83,7 @@ pub fn get_example_scenes() -> Box<[AnyScene]> {
         AnyScene::new(simple::SimpleScene::new()),
         AnyScene::new(clip::ClipScene::new()),
         AnyScene::new(image::ImageScene {}),
+        AnyScene::new(blend::BlendScene::new()),
```
taj-p, line 106: nit: ditto above

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -215,24 +224,43 @@ pub(crate) struct Scheduler {
...
     rounds_queue: VecDeque<Round>,
-    /// State for a single wide tile.
-    tile_state: TileState,
+    #[cfg(debug_assertions)]
+    clear: [Vec<u32>; 2],
+}
+
+/// The information required to resume a wide tile draw operation. Note that the suspended tile owns
+/// global slot resources, so must be resumed to free the owned slots. Tile drawing must suspend to
+/// allow blend layers to be merged.
```
taj-p: "Tile drawing must suspend to allow blend layers to be merged." I think this part is a little vague and stumped me for a little. I think this is speaking to how we cannot execute all the wide tile commands without a draw call in between when a blend exists (because we need to ensure the two slots we are blending are in the same texture)

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +269,234 @@ struct TileState {
...
+        // Round has flushed. Instead of flushing to wait for a blend, we suspend the tile's work,
+        // and continue iterating through the tiles. Slots claimed from the slot textures are a
+        // shared resource, and it's possible for all the tiles to not be able to "resume" from
+        // their "suspended" state due to no slots being available.
+        while !self.rounds_queue.is_empty() || !pending_work.is_empty() {
+            if !self.rounds_queue.is_empty() {
+                self.flush(renderer);
+            }
+
+            let pending: Vec<PendingTileWork<'scene>> = mem::take(&mut pending_work);
```
taj-p: Each iteration through the while allocates a new `PendingTileWork` vec. We could instead use 2 scratch buffers owned by the scheduler to prevent those allocations

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +269,234 @@ struct TileState {
...
+
+    pub(crate) fn do_scene<'scene, R: RendererBackend>(
+        &mut self,
+        renderer: &mut R,
+        scene: &'scene Scene,
     ) -> Result<(), RenderError> {
-        let mut tile_state = mem::take(&mut self.tile_state);
         let wide_tiles_per_row = scene.wide.width_tiles();
         let wide_tiles_per_col = scene.wide.height_tiles();
+        let mut pending_work: Vec<PendingTileWork<'scene>> = Default::default();
```
taj-p: As we do for other scheduler allocations, should the scheduler own this so it isn't allocated each `do_scene`?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -215,24 +224,43 @@ pub(crate) struct Scheduler {
...
+    /// Used to reference the wide tile.
+    wide_tile_row: u16,
+    /// On resuming the cmd index to start processing draws from.
+    next_cmd_idx: usize,
+    /// The suspended tile state stack representing active layers.
+    stack: TileState,
+    /// Round at which this work was suspended. Used to calculate round offsets.
+    suspended_at_round: usize,
+    /// The draw commands being iterated to draw the wide tile.
+    annotated_cmds: Vec<AnnotatedCmd<'a>>,
```
taj-p, line 256: I'm mildly concerned about all the allocations introduced in this PR. 

Each frame:
-  each wide tile creates an allocation equal or larger in size to its `wide_tile.cmds` [here](https://github.com/taj-p/vello/blob/b090ce9c05c88e15f0814b6ae0e00e71f23ca040/sparse_strips/vello_hybrid/src/schedule.rs#L1081)
- each wide tile creates a minor allocation [here](https://github.com/taj-p/vello/blob/b090ce9c05c88e15f0814b6ae0e00e71f23ca040/sparse_strips/vello_hybrid/src/schedule.rs#L1107)
- creates a new wide tile in length [`PendingTileWork`](https://github.com/taj-p/vello/blob/b090ce9c05c88e15f0814b6ae0e00e71f23ca040/sparse_strips/vello_hybrid/src/schedule.rs#L424) vec

Even though in the Vec of annotated commands, we're only referencing most commands, since we have the `Generated` variant, Rust will allocate enough memory in each slot to populate an entire command.

Depending on how long we think this `AnnotatedCmd` will live in `PendingTileWork`, there are a range of options available:
- At a minimum, store more of these allocations in `Scheduler` for re-use (at least the `Vec<PendingTileWork>`)
- Re-use the `Vec<AnnotatedCmd>` allocations in the `Scheduler` by detaching it from `PendingTileWork` somehow
- Not reference commands that are in `wide_tile.cmds` duplicately in `Vec<AnnotatedCmds>`
- something else

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +269,234 @@ struct TileState {
...
+            Self::Texture1(_) => 1,
+        }
+    }
+}
+
+/// Annotated commands lets the scheduler be smarter about what work to do per command by preparing
+/// some work upfront in `O(N)` time, where `N` is the length of the wide tile draw commands.
+///
+/// TODO: In the future these annotations could be optionally enabled in coarse.rs directly avoiding
+/// the need to do a linear scan.
```
```suggestion
/// the need to do linear scans.
```

There are multiple scans occurring

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1066,68 @@ impl Scheduler {
...
+    }
+    if seen_blend_to_canvas {
+        // We need to wrap the draw commands with an extra layer - preventing blending directly into
+        // the canvas. Linear time.
+        annotated_commands.insert(0, AnnotatedCmd::Generated(Cmd::PushBuf));
+        annotated_commands.push(AnnotatedCmd::Generated(Cmd::Blend(BlendMode {
+            mix: Mix::Normal,
+            compose: Compose::SrcOver,
+        })));
+        annotated_commands.push(AnnotatedCmd::Generated(Cmd::PopBuf));
```
taj-p: We might be able to reduce the total allocation size of `AnnotatedCmd` by instead storing a variant that means these commands (scratch code below):

```rs
#[derive(Debug)]
enum AnnotatedCmd<'a> {
    IdentityBorrowed(&'a Cmd),
    PushBuf,
    BlendWithPop,
    PushBufWithTemporarySlot,
}
```

Note that this removes the `Generated` variant which is much larger than all other variants.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1000,68 @@ impl Scheduler {
...
+            _ => 0,
+        };
+        // A blend command is issued directly to the viewport.
+        seen_blend_to_canvas |= matches!(cmd, Cmd::Blend(_)) && depth == 2;
+        annotated_commands.push(AnnotatedCmd::IdentityBorrowed(cmd));
+    }
+    if seen_blend_to_canvas {
+        // We need to wrap the draw commands with an extra layer - preventing blending directly into
+        // the canvas. Linear time.
+        annotated_commands.insert(0, AnnotatedCmd::Generated(Cmd::PushBuf));
```
taj-p: We can skip this `O(Vec::len)` insertion if we initialise annotated_commands with a replacement element like a `AnnotatedCmd::Empty` variant or `Option` and replace it here (via `annotated_commands[0] = ... `)

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1000,68 @@ impl Scheduler {
...
+/// this function:
+///  - Precomputes the layers that require temporary slots due to blending.
+///
+/// TODO: Can be triggered via a const generic on coarse draw cmd generation which will avoid
+/// linear scans.
+fn prepare_cmds<'a>(cmds: &'a [Cmd]) -> Vec<AnnotatedCmd<'a>> {
+    let mut annotated_commands: Vec<AnnotatedCmd<'a>> = Default::default();
+    // vello_hybrid cannot support non destructive in-place composites. Expand out to explicit blend
+    // layers.
+    let mut seen_blend_to_canvas = false;
```
```suggestion
    let mut seen_blend_to_surface = false;
```

I think (?) surface is generally what is referred to as the presentable surface (instead of canvas)

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1000,68 @@ impl Scheduler {
...
+        annotated_commands.push(AnnotatedCmd::Generated(Cmd::Blend(BlendMode {
+            mix: Mix::Normal,
+            compose: Compose::SrcOver,
+        })));
+        annotated_commands.push(AnnotatedCmd::Generated(Cmd::PopBuf));
+    }
+
+    let mut pointer_to_push_buf_stack: Vec<usize> = Default::default();
+
+    // Second pass: Precompute which buffers will need temporary slots for blending.
```
taj-p: Is the second pass necessary? What in this second pass cannot be done in the first pass?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -215,24 +222,43 @@ pub(crate) struct Scheduler {
...
-    tile_state: TileState,
+    #[cfg(debug_assertions)]
+    clear: [Vec<u32>; 2],
+}
+
+/// The information required to resume a wide tile draw operation. Note that the suspended tile owns
+/// global slot resources, so must be resumed to free the owned slots. Tile drawing must suspend to
+/// allow blend layers to be merged.
+#[derive(Debug)]
+struct PendingTileWork<'a> {
```
```suggestion
struct PendingWideTileWork<'a> {
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +267,235 @@ struct TileState {
...
+    IdentityBorrowed(&'a Cmd),
+    /// An owned generated command. Allows adding additional commands.
+    Generated(Cmd),
+    /// A `Cmd::PushBuf` that will be blended into by the layer above it. Will need a
+    /// `temporary_slot` created for it to enable compositing.
+    PushBufWithTemporarySlot,
+}
+
+impl<'a> AnnotatedCmd<'a> {
+    fn unwrap<'b: 'a>(&'b self) -> &'a Cmd {
```
```suggestion
    fn as_cmd<'b: 'a>(&'b self) -> &'a Cmd {
```

I think `unwrap` clashes too hard with `[Result|Option]::unwrap`

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +267,235 @@ struct TileState {
...
+
+impl TemporarySlot {
+    fn invalidate(&mut self) {
+        match self {
+            Self::Invalid(_) => {}
+            Self::None => {}
+            Self::Valid(slot) => {
+                *self = Self::Invalid(*slot);
+            }
+        };
```
```suggestion
        match self {
            Self::Valid(slot) => *self = Self::Invalid(*slot),
            _ => {}
        }
```

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -794,38 +834,51 @@ impl WideTile {
 
     pub(crate) fn fill(&mut self, x: u16, width: u16, paint: Paint) {
         if !self.is_zero_clip() {
```
ajakubowicz-canva, line 836: Compile time `MODE` had to be added because this optimization cannot work for vello_hybrid with blends.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done thank you :D

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Changed to `as_cmd`. Great comment!

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done :D

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: I made it a single pass.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Great suggestions!

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: This remains `a linear scan` because I fixed the implementation. Much better to be `O(N)` than `O(3N)` in the worst case.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: This is a great suggestion - I considered it initially but was being confused by other things. I've now added this.

It is an optimization in both memory, but also allows us to avoid a linear scan.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Did this excellent suggestion!

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: I completely agree with you.

Maybe with the introduction of [`MODE`](https://github.com/linebender/vello/pull/1155/commits/f6b75817a012a7ca4a138850fe2c383a1b313e3e) we can do something here. I time-boxed an attempt to clean up some of these things and it's tricky (and I might simply just not have the right knowledge).

For example. Getting rid of `PendingTileWork` is tricky due to the `'scene` lifetime. I think I know of an `unsafe` way to get around this, but I'm interested if there is a safe solution.

With the introduction of `MODE` we could roll the annotated commands changes _into_ `WideTile` somehow. Maybe as a sparse array of metadata relating to the existing commands.

I've applied your other excellent code review feedback, but might need a bit more guidance with this comment.

ajakubowicz-canva:
Nit to self: It's not encoded when a suspended tile can resume. I think it's not as implicit as "on the next round". It may be in N+2 rounds where N is the suspended round.

~TODO: Explore this.~
Edit: Explored - there is a forced flush between iterations over the tiles. This enforces that at least 1 flush happens before suspension and work resumption.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done :D Great suggestion.

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: > For example. Getting rid of PendingTileWork is tricky due to the 'scene lifetime.

We don't necessarily need the `'scene` lifetime. Since `WideTile::cmds` are immutable across the schedule, we can store the index into `WideTile::cmds` like this: https://github.com/linebender/vello/commit/38f2f00d8eb266301c417fcec68b476c563617ab

---

After `PendingWideTileWork`, we also have to think about the `AnnotatedCmds` vector which probably needs some close considersation.

I took a profile ([profile.json.gz](https://github.com/user-attachments/files/21868615/profile.json.gz)) and found that maybe 10% of the time is spent allocating/deallocating vectors. You can load it with [samply](https://github.com/mstange/samply/) via `samply load profile.json.gz`. Also note that this profile was taken after the comment that pre-allocates the `AnnotatedCmd` vec with the expected capacity and by zooming and panning the complex blend example.

I suspect Web performance will be worse than this native profile.

That said, I'm wondering whether we should simply punt this as a performance optimisation to be looked at separately so we can land the feature sooner. I think it'll be easier to optimise these allocations in isolation than within this larger PR (which builds on top of the existing pattern of allocating-on-demand).

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1025,149 @@ impl Scheduler {
...
 }
+
+/// Does a single linear scan over the wide tile commands to prepare them for `do_tile`. Notably
+/// this function:
+///  - Precomputes the layers that require temporary slots due to blending.
+///
+/// TODO: Can be triggered via a const generic on coarse draw cmd generation which will avoid
+/// a linear scan.
+fn prepare_cmds<'a>(cmds: &'a [Cmd]) -> Vec<AnnotatedCmd<'a>> {
+    let mut annotated_commands: Vec<AnnotatedCmd<'a>> = Default::default();
```
```suggestion
    let mut annotated_commands: Vec<AnnotatedCmd<'a>> = Vec::with_capacity(cmds.len() + 3);
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1025,149 @@ impl Scheduler {
...
+    width: u16,
+    dense_width: u16,
+    col_idx: u32,
+}
 
-const PAINT_TYPE_SOLID: u32 = 0;
-const PAINT_TYPE_IMAGE: u32 = 1;
+impl GpuStripBuilder {
+    /// Position at canvas/scene coordinates.
+    fn at_canvas(x: u16, y: u16, width: u16) -> Self {
```
```suggestion
    fn at_surface(x: u16, y: u16, width: u16) -> Self {
```

I think it's best to be consistent with WGPU with what a surface is.

```suggestion
    fn at_render_target(x: u16, y: u16, width: u16) -> Self {
```

Or we could choose to align on `render_target`? I don't mind either

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +272,251 @@ struct TileState {
...
+
+    pub(crate) fn do_scene<'scene, R: RendererBackend>(
+        &mut self,
+        renderer: &mut R,
+        scene: &'scene Scene,
     ) -> Result<(), RenderError> {
-        let mut tile_state = mem::take(&mut self.tile_state);
         let wide_tiles_per_row = scene.wide.width_tiles();
         let wide_tiles_per_col = scene.wide.height_tiles();
+        let mut pending_work: Vec<PendingWideTileWork<'scene>> = Default::default();
```
taj-p: Can we pre-allocate this vector since we know how many elements we'll need ahead of time?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +272,251 @@ struct TileState {
...
+        // Round has flushed. Instead of flushing to wait for a blend, we suspend the tile's work,
+        // and continue iterating through the tiles. Slots claimed from the slot textures are a
+        // shared resource, and it's possible for all the tiles to not be able to "resume" from
+        // their "suspended" state due to no slots being available.
+        while !self.rounds_queue.is_empty() || !pending_work.is_empty() {
+            if !self.rounds_queue.is_empty() {
+                self.flush(renderer);
+            }
+
+            let pending: Vec<PendingWideTileWork<'scene>> = mem::take(&mut pending_work);
```
taj-p: Instead of taking the allocation of `pending_work` and freeing `pending` at the end of the block, we could use a scratch buffer to preserve the allocation. For example, something like the below:

```rs
        let mut scratch: Vec<PendingWideTileWork<'scene>> = Default::default();

        ...
        while !self.rounds_queue.is_empty() || !pending_work.is_empty() {
            ...

            let mut pending: Vec<PendingWideTileWork<'scene>> = mem::replace(&mut pending_work, scratch);
            for work in pending.drain(..) {
                ...
            }
            scratch = pending;

            ...
        }
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +272,251 @@ struct TileState {
...
+            if !pending_work.is_empty() {
+                // TODO: This can be tested by arbitrarily limiting the total slots that the
+                // Scheduler is instantiated with. For example, tests deadlock at a value of `20`.
+                // Ideally we improve this system such that it does not deadlock given a situation
+                // where a single tile can always be completed given all resources. This may require
+                // reserving the slots required for a wide tile to complete upfront.
+                assert!(
+                    !self.rounds_queue.is_empty(),
+                    "deadlock in scheduler detected"
+                );
```
taj-p, line 510: Should we instead return a `RenderError` variant here?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -440,34 +695,106 @@ impl Scheduler {
...
+                    //
+                    // This works well _once_. Unfortunately, to blend to `nos` again we need to
+                    // copy the contents of `dest_slot` back to `temporary_slot` such that it can be
+                    // composited again. Without this, the tile will not accumulate blends, and
+                    // instead will have repeated blends clobber the prior blends. Hence, if a
+                    // buffer is being pushed that will be blended back to `tos`, copy contents from
+                    // `tos.dest_slot` to `tos.temporary_slot` ready for future blending.
+                    {
+                        let tos: &mut TileEl = state.stack.last_mut().unwrap();
+                        if let TemporarySlot::Invalid(temp_slot) = tos.temporary_slot {
```
taj-p, line 736: Is it possible to work out in the linear scan whether this is required (or whether we can skip the copying back to `temp_slot` since there are no further blends?)

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Wouldn't returning a `RenderError::SchedulerDeadlock` variant make this public API? In the future I hope to wholly eliminate this class of error.

Removing a panic seems easier than removing a `RenderError` variant that could be observed elsewhere. Happy to add the RenderError though, just checking my understanding.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Yes - however I'll track this as a followup. https://github.com/linebender/vello/issues/1165

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Tracking in https://github.com/linebender/vello/issues/1165 so we can punt from this PR.

sparse_strips/vello_common/src/coarse.rs
ajakubowicz-canva, line 1: cc/ @LaurenzV I don't expect you to review this PR, however it may be worth taking a look at the changes to `vello_common`, specifically regarding to this const generic.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done in https://github.com/linebender/vello/pull/1155/commits/b3cf355219bb9d9b4c1725c83a2e38a8d337d165#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR431-R434

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Great suggestion: https://github.com/linebender/vello/pull/1155/commits/b3cf355219bb9d9b4c1725c83a2e38a8d337d165#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR1133

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done: https://github.com/linebender/vello/pull/1155/commits/b3cf355219bb9d9b4c1725c83a2e38a8d337d165#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR1042

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done: https://github.com/linebender/vello/pull/1155/commits/b3cf355219bb9d9b4c1725c83a2e38a8d337d165#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR498

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Done: https://github.com/linebender/vello/pull/1155/commits/b3cf355219bb9d9b4c1725c83a2e38a8d337d165#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR431-R434

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Discussion explored in the [other comment](https://github.com/linebender/vello/pull/1155#discussion_r2286379214).
This is something tracked in #1165

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Hopefully clarified in https://github.com/linebender/vello/pull/1155/files#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR234-R238

taj-p (approved):
LGTM! Let's goooooooooooooo 🚀 ! Amazing work 🙌

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +272,250 @@ struct TileState {
...
+
+                let tile_state =
+                    self.initialize_tile_state(wide_tile, wide_tile_x, wide_tile_y, scene);
+                let annotated_cmds = prepare_cmds(&wide_tile.cmds);
+                pending_work.push(PendingWideTileWork {
+                    wide_tile_col,
+                    wide_tile_row,
+                    next_cmd_idx: 0,
+                    stack: tile_state,
+                    suspended_at_round: self.round,
```
grebmeg, line 456: Would the initial round be 0 here for all tile works?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -349,85 +587,101 @@ impl Scheduler {
...
+        wide_tile_y: u16,
+        cmds: &[AnnotatedCmd<'a>],
+        state: &mut TileState,
+        start_cmd_idx: usize,
+    ) -> Result<Option<usize>, RenderError> {
+        let mut has_blended = false;
+        for (offset_cmd_idx, annotated_cmd) in cmds[start_cmd_idx..].iter().enumerate() {
+            let cmd_idx = start_cmd_idx + offset_cmd_idx;
             // Note: this starts at 1 (for the final target)
             let clip_depth = state.stack.len();
```
grebmeg: Since it’s not only clip-related, should we rename it?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -440,34 +694,106 @@ impl Scheduler {
...
+                            tos.temporary_slot = TemporarySlot::Valid(temp_slot);
+
+                            // Make sure the destination slot and temporary slot are cleared
+                            // appropriately.
+                            let dest_slot = tos.dest_slot;
+                            let round = self.get_round(tos.round - 1);
+                            round.clear[temp_slot.get_texture()].push(temp_slot.get_idx() as u32);
+                            #[cfg(debug_assertions)]
+                            self.clear[temp_slot.get_texture()].push(temp_slot.get_idx() as u32);
+                            if dest_slot.get_idx() as u32 != u32::MAX {
```
grebmeg: Should it be `usize::MAX` here?
```suggestion
                            if dest_slot.get_idx() != usize::MAX {
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -484,105 +810,178 @@ impl Scheduler {
...
-                        "Changing blend mode is unsupported"
+                        matches!(mode.mix, Mix::Normal),
+                        "Only Mix::Normal is supported currently"
                     );
 
                     let tos = state.stack.last().unwrap();
                     let nos = &state.stack[state.stack.len() - 2];
 
-                    let next_round = clip_depth % 2 == 0 && clip_depth > 2;
+                    let next_round: bool = clip_depth % 2 == 0 && clip_depth > 2;
```
grebmeg: [nit]: This is repeated several times, maybe we should add `requires_next_round` method or similar?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +272,250 @@ struct TileState {
...
+impl TileEl {
+    fn get_draw_texture(&self, buffer_depth: usize) -> usize {
+        if buffer_depth > 1 {
+            if let TemporarySlot::Valid(temp_slot) = self.temporary_slot {
+                temp_slot.get_texture()
+            } else {
+                self.dest_slot.get_texture()
+            }
+        } else {
+            2
```
grebmeg, line 378: I’m familiar with those numbers, but for someone who isn’t, the code may be harder to follow? Would it make sense to add an additional enum to describe the targets?

```rust
enum Target {
    Even = 0,
    Odd = 1,
    Surface = 2,
}
```

And then we could Pprovide a helper `Target::for_draw(buffer_depth, temporary_slot_present)` to centralize the logic currently in `TileEl::get_draw_texture`.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: I'm not sure I'm a fan of introducing a const generic for this, since this seems to be the only usage, can't we just take an additional flag in the function?

sparse_strips/vello_common/src/coarse.rs
LaurenzV: I left a comment.

sparse_strips/vello_common/src/coarse.rs
ajakubowicz-canva: It is true that an argument can be fed through `vello_hybrid/scene` -> `self.wide.generate` to `fill`.

I can roll it back if you prefer, however, this refactor may still be required in the followup when the required annotated commands change is rolled up into wide tile cmd generation.

Specifically [this linear scan](https://github.com/linebender/vello/pull/1155/files#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR1122-R1172) can be removed by rolling it into cmd generation (gated behind const generic). This allows removal of an unnecessary  linear scan & allocation of `annotated_commands`.


Would you prefer I roll this back in this PR?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -484,105 +810,178 @@ impl Scheduler {
...
+                            wide_tile_y,
+                            clip_fill.width as u16,
+                        )
                     } else {
-                        (clip_fill.x as u16, nos.slot_ix as u16 * Tile::HEIGHT)
+                        GpuStripBuilder::at_slot(
+                            nos.dest_slot.get_idx(),
+                            clip_fill.x as u16,
+                            clip_fill.width as u16,
+                        )
```
grebmeg: It looks like the logic "copy temp to dest if temp exists and is valid" is the same as in `Cmd::ClipStrip` and somewhat similar to `Cmd::Blend(mode)`. Would it make sense to encapsulate this into a separate method?

sparse_strips/vello_common/src/coarse.rs
ajakubowicz-canva: cc/ @taj-p Are you also happy with a patch that looks like what is shared below?

Below is a patch that can be applied to this PR (that passes tests) which rolls back the const generic change and replaces it with a `skip_background_optimization` argument that is fed through `vello_cpu` and `vello_hybrid`.

I am easy either way. The const generic method would be compiled out whilst the diff below would be "constant" in the context of `vello_hybrid` or `vello_cpu` but still incur branching cost.

If we go in this direction we will probably need to consider the const generic change when addressing followups.

<details><summary> Expand for diff:</summary>

```diff
diff --git a/sparse_strips/vello_common/src/coarse.rs b/sparse_strips/vello_common/src/coarse.rs
index fd306d0a..96f7724a 100644
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -38,22 +38,15 @@ impl Layer {
     }
 }
 
-/// `MODE_CPU` allows compile time optimizations to be applied to wide tile draw command generation
-/// specific to `vello_cpu`.
-pub const MODE_CPU: u8 = 0;
-/// `MODE_HYBRID` allows compile time optimizations to be applied to wide tile draw command
-/// generation specific for `vello_hybrid`.
-pub const MODE_HYBRID: u8 = 1;
-
 /// A container for wide tiles.
 #[derive(Debug)]
-pub struct Wide<const MODE: u8 = MODE_CPU> {
+pub struct Wide {
     /// The width of the container.
     pub width: u16,
     /// The height of the container.
     pub height: u16,
     /// The wide tiles in the container.
-    pub tiles: Vec<WideTile<MODE>>,
+    pub tiles: Vec<WideTile>,
     /// The stack of layers.
     layer_stack: Vec<Layer>,
     /// The stack of active clip regions.
@@ -131,33 +124,16 @@ impl Bbox {
     }
 }
 
-impl Wide<MODE_CPU> {
+impl Wide {
     /// Create a new container for wide tiles.
     pub fn new(width: u16, height: u16) -> Self {
-        Self::new_internal(width, height)
-    }
-}
-
-impl Wide<MODE_HYBRID> {
-    /// Create a new container for wide tiles.
-    pub fn new(width: u16, height: u16) -> Self {
-        Self::new_internal(width, height)
-    }
-}
-
-impl<const MODE: u8> Wide<MODE> {
-    /// Create a new container for wide tiles.
-    fn new_internal(width: u16, height: u16) -> Self {
         let width_tiles = width.div_ceil(WideTile::WIDTH);
         let height_tiles = height.div_ceil(Tile::HEIGHT);
         let mut tiles = Vec::with_capacity(usize::from(width_tiles) * usize::from(height_tiles));
 
         for h in 0..height_tiles {
             for w in 0..width_tiles {
-                tiles.push(WideTile::<MODE>::new_internal(
-                    w * WideTile::WIDTH,
-                    h * Tile::HEIGHT,
-                ));
+                tiles.push(WideTile::new(w * WideTile::WIDTH, h * Tile::HEIGHT));
             }
         }
 
@@ -198,7 +174,7 @@ impl<const MODE: u8> Wide<MODE> {
     /// Get the wide tile at a certain index.
     ///
     /// Panics if the index is out-of-range.
-    pub fn get(&self, x: u16, y: u16) -> &WideTile<MODE> {
+    pub fn get(&self, x: u16, y: u16) -> &WideTile {
         assert!(
             x < self.width_tiles() && y < self.height_tiles(),
             "attempted to access out-of-bounds wide tile"
@@ -210,7 +186,7 @@ impl<const MODE: u8> Wide<MODE> {
     /// Get mutable access to the wide tile at a certain index.
     ///
     /// Panics if the index is out-of-range.
-    pub fn get_mut(&mut self, x: u16, y: u16) -> &mut WideTile<MODE> {
+    pub fn get_mut(&mut self, x: u16, y: u16) -> &mut WideTile {
         assert!(
             x < self.width_tiles() && y < self.height_tiles(),
             "attempted to access out-of-bounds wide tile"
@@ -221,7 +197,7 @@ impl<const MODE: u8> Wide<MODE> {
     }
 
     /// Return a reference to all wide tiles.
-    pub fn tiles(&self) -> &[WideTile<MODE>] {
+    pub fn tiles(&self) -> &[WideTile] {
         self.tiles.as_slice()
     }
 
@@ -237,7 +213,14 @@ impl<const MODE: u8> Wide<MODE> {
     ///    - Generate alpha fill commands for the intersected wide tiles
     /// 2. For active fill regions (determined by fill rule):
     ///    - Generate solid fill commands for the regions between strips
-    pub fn generate(&mut self, strip_buf: &[Strip], fill_rule: Fill, paint: Paint, thread_idx: u8) {
+    pub fn generate(
+        &mut self,
+        strip_buf: &[Strip],
+        fill_rule: Fill,
+        paint: Paint,
+        thread_idx: u8,
+        skip_background_optimization: bool,
+    ) {
         if strip_buf.is_empty() {
             return;
         }
@@ -338,8 +321,12 @@ impl<const MODE: u8> Wide<MODE> {
                     let x_wtile_rel = x % WideTile::WIDTH;
                     let width = x2.min((wtile_x + 1) * WideTile::WIDTH) - x;
                     x += width;
-                    self.get_mut(wtile_x, strip_y)
-                        .fill(x_wtile_rel, width, paint.clone());
+                    self.get_mut(wtile_x, strip_y).fill(
+                        x_wtile_rel,
+                        width,
+                        paint.clone(),
+                        skip_background_optimization,
+                    );
                 }
             }
         }
@@ -782,7 +769,7 @@ impl<const MODE: u8> Wide<MODE> {
 
 /// A wide tile.
 #[derive(Debug)]
-pub struct WideTile<const MODE: u8 = MODE_CPU> {
+pub struct WideTile {
     /// The x coordinate of the wide tile.
     pub x: u16,
     /// The y coordinate of the wide tile.
@@ -802,25 +789,9 @@ pub struct WideTile<const MODE: u8 = MODE_CPU> {
 impl WideTile {
     /// The width of a wide tile in pixels.
     pub const WIDTH: u16 = 256;
-}
 
-impl WideTile<MODE_CPU> {
     /// Create a new wide tile.
-    pub fn new(width: u16, height: u16) -> Self {
-        Self::new_internal(width, height)
-    }
-}
-
-impl WideTile<MODE_HYBRID> {
-    /// Create a new wide tile.
-    pub fn new(width: u16, height: u16) -> Self {
-        Self::new_internal(width, height)
-    }
-}
-
-impl<const MODE: u8> WideTile<MODE> {
-    /// Create a new wide tile.
-    fn new_internal(x: u16, y: u16) -> Self {
+    pub fn new(x: u16, y: u16) -> Self {
         Self {
             x,
             y,
@@ -832,45 +803,41 @@ impl<const MODE: u8> WideTile<MODE> {
         }
     }
 
-    pub(crate) fn fill(&mut self, x: u16, width: u16, paint: Paint) {
+    pub(crate) fn fill(
+        &mut self,
+        x: u16,
+        width: u16,
+        paint: Paint,
+        skip_background_optimization: bool,
+    ) {
         if !self.is_zero_clip() {
-            match MODE {
-                MODE_CPU => {
-                    let bg = if let Paint::Solid(s) = &paint {
-                        // Note that we could be more aggressive in optimizing a whole-tile opaque fill
-                        // even with a clip stack. It would be valid to elide all drawing commands from
-                        // the enclosing clip push up to the fill. Further, we could extend the clip
-                        // push command to include a background color, rather than always starting with
-                        // a transparent buffer. Lastly, a sequence of push(bg); strip/fill; pop could
-                        // be replaced with strip/fill with the color (the latter is true even with a
-                        // non-opaque color).
-                        //
-                        // However, the extra cost of tracking such optimizations may outweigh the
-                        // benefit, especially in hybrid mode with GPU painting.
-                        let can_override = x == 0
-                            && width == WideTile::WIDTH
-                            && s.is_opaque()
-                            && self.n_clip == 0
-                            && self.n_bufs == 0;
-                        can_override.then_some(*s)
-                    } else {
-                        // TODO: Implement for indexed paints.
-                        None
-                    };
-
-                    if let Some(bg) = bg {
-                        self.cmds.clear();
-                        self.bg = bg;
-                    } else {
-                        self.cmds.push(Cmd::Fill(CmdFill {
-                            x,
-                            width,
-                            paint,
-                            blend_mode: None,
-                        }));
-                    }
-                }
-                MODE_HYBRID => {
+            if !skip_background_optimization {
+                let bg = if let Paint::Solid(s) = &paint {
+                    // Note that we could be more aggressive in optimizing a whole-tile opaque fill
+                    // even with a clip stack. It would be valid to elide all drawing commands from
+                    // the enclosing clip push up to the fill. Further, we could extend the clip
+                    // push command to include a background color, rather than always starting with
+                    // a transparent buffer. Lastly, a sequence of push(bg); strip/fill; pop could
+                    // be replaced with strip/fill with the color (the latter is true even with a
+                    // non-opaque color).
+                    //
+                    // However, the extra cost of tracking such optimizations may outweigh the
+                    // benefit, especially in hybrid mode with GPU painting.
+                    let can_override = x == 0
+                        && width == Self::WIDTH
+                        && s.is_opaque()
+                        && self.n_clip == 0
+                        && self.n_bufs == 0;
+                    can_override.then_some(*s)
+                } else {
+                    // TODO: Implement for indexed paints.
+                    None
+                };
+
+                if let Some(bg) = bg {
+                    self.cmds.clear();
+                    self.bg = bg;
+                } else {
                     self.cmds.push(Cmd::Fill(CmdFill {
                         x,
                         width,
@@ -878,7 +845,13 @@ impl<const MODE: u8> WideTile<MODE> {
                         blend_mode: None,
                     }));
                 }
-                _ => unreachable!(),
+            } else {
+                self.cmds.push(Cmd::Fill(CmdFill {
+                    x,
+                    width,
+                    paint,
+                    blend_mode: None,
+                }));
             }
         }
     }
@@ -1086,7 +1059,7 @@ impl BlendModeExt for BlendMode {
 
 #[cfg(test)]
 mod tests {
-    use crate::coarse::{MODE_CPU, Wide, WideTile};
+    use crate::coarse::{Wide, WideTile};
     use crate::color::AlphaColor;
     use crate::color::palette::css::TRANSPARENT;
     use crate::paint::{Paint, PremulColor};
@@ -1096,7 +1069,7 @@ mod tests {
 
     #[test]
     fn optimize_empty_layers() {
-        let mut wide = WideTile::<MODE_CPU>::new(0, 0);
+        let mut wide = WideTile::new(0, 0);
         wide.push_buf();
         wide.pop_buf();
 
@@ -1105,17 +1078,19 @@ mod tests {
 
     #[test]
     fn basic_layer() {
-        let mut wide = WideTile::<MODE_CPU>::new(0, 0);
+        let mut wide = WideTile::new(0, 0);
         wide.push_buf();
         wide.fill(
             0,
             10,
             Paint::Solid(PremulColor::from_alpha_color(TRANSPARENT)),
+            false,
         );
         wide.fill(
             10,
             10,
             Paint::Solid(PremulColor::from_alpha_color(TRANSPARENT)),
+            false,
         );
         wide.pop_buf();
 
@@ -1129,10 +1104,10 @@ mod tests {
         )));
         let blend_mode = BlendMode::new(Mix::Lighten, Compose::SrcOver);
 
-        let mut wide = WideTile::<MODE_CPU>::new(0, 0);
+        let mut wide = WideTile::new(0, 0);
         wide.push_buf();
-        wide.fill(0, 10, paint.clone());
-        wide.fill(10, 10, paint.clone());
+        wide.fill(0, 10, paint.clone(), false);
+        wide.fill(10, 10, paint.clone(), false);
         wide.blend(blend_mode);
         wide.pop_buf();
 
@@ -1146,9 +1121,9 @@ mod tests {
         )));
         let blend_mode = BlendMode::new(Mix::Lighten, Compose::Clear);
 
-        let mut wide = WideTile::<MODE_CPU>::new(0, 0);
+        let mut wide = WideTile::new(0, 0);
         wide.push_buf();
-        wide.fill(0, 10, paint.clone());
+        wide.fill(0, 10, paint.clone(), false);
         wide.blend(blend_mode);
         wide.pop_buf();
 
@@ -1157,7 +1132,7 @@ mod tests {
 
     #[test]
     fn tile_coordinates() {
-        let wide = Wide::<MODE_CPU>::new(1000, 258);
+        let wide = Wide::new(1000, 258);
 
         let tile_1 = wide.get(1, 3);
         assert_eq!(tile_1.x, 256);
@@ -1172,7 +1147,7 @@ mod tests {
     fn reset_clears_layer_and_clip_stacks() {
         type ClipPath = Option<(Box<[Strip]>, Fill)>;
 
-        let mut wide = Wide::<MODE_CPU>::new(1000, 258);
+        let mut wide = Wide::new(1000, 258);
         let no_clip_path: ClipPath = None;
         wide.push_layer(no_clip_path, BlendMode::default(), None, 0.5, 0);
 
diff --git a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
index 02d874b6..effd89b5 100644
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -21,7 +21,7 @@ use std::cell::RefCell;
 use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use std::sync::{Barrier, Mutex};
 use thread_local::ThreadLocal;
-use vello_common::coarse::{Cmd, MODE_CPU, Wide};
+use vello_common::coarse::{Cmd, Wide};
 use vello_common::encode::EncodedPaint;
 use vello_common::fearless_simd::{Fallback, Level, Simd};
 use vello_common::mask::Mask;
@@ -101,7 +101,7 @@ pub(crate) struct MultiThreadedDispatcher {
 
 impl MultiThreadedDispatcher {
     pub(crate) fn new(width: u16, height: u16, num_threads: u16, level: Level) -> Self {
-        let wide = Wide::<MODE_CPU>::new(width, height);
+        let wide = Wide::new(width, height);
         let thread_pool = ThreadPoolBuilder::new()
             .num_threads(num_threads as usize)
             .build()
@@ -247,7 +247,7 @@ impl MultiThreadedDispatcher {
                                 strips,
                                 fill_rule,
                                 paint,
-                            } => self.wide.generate(&strips, fill_rule, paint, thread_id),
+                            } => self.wide.generate(&strips, fill_rule, paint, thread_id, false),
                             CoarseTask::PushLayer {
                                 thread_id,
                                 clip_path,
diff --git a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
index 147fee68..7f0d24c3 100644
--- a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
@@ -9,7 +9,7 @@ use crate::peniko::{BlendMode, Fill};
 use crate::region::Regions;
 use crate::strip_generator::StripGenerator;
 use alloc::vec::Vec;
-use vello_common::coarse::{MODE_CPU, Wide};
+use vello_common::coarse::Wide;
 use vello_common::encode::EncodedPaint;
 use vello_common::fearless_simd::{Fallback, Level, Simd};
 use vello_common::mask::Mask;
@@ -24,7 +24,7 @@ pub(crate) struct SingleThreadedDispatcher {
 
 impl SingleThreadedDispatcher {
     pub(crate) fn new(width: u16, height: u16, level: Level) -> Self {
-        let wide = Wide::<MODE_CPU>::new(width, height);
+        let wide = Wide::new(width, height);
         let strip_generator = StripGenerator::new(width, height, level);
 
         Self {
@@ -81,7 +81,7 @@ impl Dispatcher for SingleThreadedDispatcher {
     ) {
         let wide = &mut self.wide;
 
-        let func = |strips| wide.generate(strips, fill_rule, paint, 0);
+        let func = |strips| wide.generate(strips, fill_rule, paint, 0, false);
         self.strip_generator
             .generate_filled_path(path, fill_rule, transform, anti_alias, func);
     }
@@ -96,7 +96,7 @@ impl Dispatcher for SingleThreadedDispatcher {
     ) {
         let wide = &mut self.wide;
 
-        let func = |strips| wide.generate(strips, Fill::NonZero, paint, 0);
+        let func = |strips| wide.generate(strips, Fill::NonZero, paint, 0, false);
         self.strip_generator
             .generate_stroked_path(path, stroke, transform, anti_alias, func);
     }
diff --git a/sparse_strips/vello_cpu/src/render.rs b/sparse_strips/vello_cpu/src/render.rs
index a7c01f07..48cbdbfc 100644
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -837,7 +837,7 @@ impl RenderContext {
         };
         self.dispatcher
             .wide_mut()
-            .generate(&adjusted_strips[start..end], fill_rule, paint, 0);
+            .generate(&adjusted_strips[start..end], fill_rule, paint, 0, false);
     }
 
     /// Prepare cached strips for rendering by adjusting indices.
diff --git a/sparse_strips/vello_hybrid/src/scene.rs b/sparse_strips/vello_hybrid/src/scene.rs
index cf24acc9..b3a12982 100644
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -5,7 +5,7 @@
 
 use alloc::vec;
 use alloc::vec::Vec;
-use vello_common::coarse::{MODE_HYBRID, Wide};
+use vello_common::coarse::Wide;
 use vello_common::encode::{EncodeExt, EncodedPaint};
 use vello_common::fearless_simd::Level;
 use vello_common::flatten::{FlattenCtx, Line};
@@ -46,7 +46,7 @@ struct RenderState {
 pub struct Scene {
     pub(crate) width: u16,
     pub(crate) height: u16,
-    pub(crate) wide: Wide<MODE_HYBRID>,
+    pub(crate) wide: Wide,
     pub(crate) alphas: Vec<u8>,
     pub(crate) line_buf: Vec<Line>,
     pub(crate) tiles: Tiles,
@@ -71,7 +71,7 @@ impl Scene {
         Self {
             width,
             height,
-            wide: Wide::<MODE_HYBRID>::new(width, height),
+            wide: Wide::new(width, height),
             anti_alias: true,
             alphas: vec![],
             level: Level::fallback(),
@@ -312,7 +312,7 @@ impl Scene {
     // Assumes that `line_buf` contains the flattened path.
     fn render_path(&mut self, fill_rule: Fill, paint: Paint) {
         self.make_strips(fill_rule);
-        self.wide.generate(&self.strip_buf, fill_rule, paint, 0);
+        self.wide.generate(&self.strip_buf, fill_rule, paint, 0, true);
     }
 
     fn make_strips(&mut self, fill_rule: Fill) {
@@ -538,7 +538,7 @@ impl Scene {
             _ => Fill::NonZero,
         };
         self.wide
-            .generate(&adjusted_strips[start..end], fill_rule, paint, 0);
+            .generate(&adjusted_strips[start..end], fill_rule, paint, 0, true);
     }
 
     /// Prepare cached strips for rendering by adjusting alpha indices and extending alpha buffer.
diff --git a/sparse_strips/vello_hybrid/src/schedule.rs b/sparse_strips/vello_hybrid/src/schedule.rs
index d99665b2..b492f18e 100644
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -181,7 +181,6 @@ use crate::{GpuStrip, RenderError, Scene};
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
 use core::mem;
-use vello_common::coarse::MODE_HYBRID;
 use vello_common::peniko::{BlendMode, Compose, Mix};
 use vello_common::{
     coarse::{Cmd, WideTile},
@@ -602,7 +601,7 @@ impl Scheduler {
 
     fn initialize_tile_state(
         &mut self,
-        tile: &WideTile<MODE_HYBRID>,
+        tile: &WideTile,
         wide_tile_x: u16,
         wide_tile_y: u16,
         scene: &Scene,
diff --git a/sparse_strips/vello_toy/src/debug.rs b/sparse_strips/vello_toy/src/debug.rs
index 689a6beb..8b79d9c2 100644
--- a/sparse_strips/vello_toy/src/debug.rs
+++ b/sparse_strips/vello_toy/src/debug.rs
@@ -14,7 +14,7 @@ use std::path;
 use svg::node::element::path::Data;
 use svg::node::element::{Line as SvgLine, Path, Rectangle};
 use svg::{Document, Node};
-use vello_common::coarse::{Cmd, MODE_CPU, Wide, WideTile};
+use vello_common::coarse::{Cmd, Wide, WideTile};
 use vello_common::color::palette::css::BLACK;
 use vello_common::fearless_simd::Level;
 use vello_common::flatten::{FlattenCtx, Line};
@@ -34,7 +34,7 @@ fn main() {
     let mut tiles = Tiles::new();
     let mut strip_buf = vec![];
     let mut alpha_buf = vec![];
-    let mut wide = Wide::<MODE_CPU>::new(args.width, args.height);
+    let mut wide = Wide::new(args.width, args.height);
 
     let stages = &args.stages;
 
@@ -86,7 +86,7 @@ fn main() {
     }
 
     if stages.iter().any(|s| s.requires_wide_tiles()) {
-        wide.generate(&strip_buf, args.fill_rule, BLACK.into(), 0);
+        wide.generate(&strip_buf, args.fill_rule, BLACK.into(), 0, false);
     }
 
     draw_grid(&mut document, args.width, args.height);

```

</details>

sparse_strips/vello_common/src/coarse.rs
ajakubowicz-canva: You are awesome. Responded [over there](https://github.com/linebender/vello/pull/1155/files#r2287117988).

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -625,13 +1024,149 @@ impl Scheduler {
...
 }
+
+/// Does a single linear scan over the wide tile commands to prepare them for `do_tile`. Notably
+/// this function:
+///  - Precomputes the layers that require temporary slots due to blending.
+///
+/// TODO: Can be triggered via a const generic on coarse draw cmd generation which will avoid
+/// a linear scan.
+fn prepare_cmds<'a>(cmds: &'a [Cmd]) -> Vec<AnnotatedCmd<'a>> {
+    let mut annotated_commands: Vec<AnnotatedCmd<'a>> = Vec::with_capacity(cmds.len() + 3);
```
grebmeg, line 1136: Is `3` here due to potential additional commands later? Maybe we could add a short comment for quicker understanding?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -241,69 +272,250 @@ struct TileState {
...
+        renderer: &mut R,
+        scene: &'scene Scene,
     ) -> Result<(), RenderError> {
-        let mut tile_state = mem::take(&mut self.tile_state);
         let wide_tiles_per_row = scene.wide.width_tiles();
         let wide_tiles_per_col = scene.wide.height_tiles();
+        let mut pending_work: Vec<PendingWideTileWork<'scene>> =
+            Vec::with_capacity((wide_tiles_per_row as usize) * (wide_tiles_per_col as usize));
+        let mut pending_work_scratch: Vec<PendingWideTileWork<'scene>> =
+            Vec::with_capacity((wide_tiles_per_row as usize) * (wide_tiles_per_col as usize));
```
grebmeg, line 439: Would it make sense to preserve both allocations by moving them into `Scheduler` struct fields, so they can be reused across calls?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -215,24 +225,45 @@ pub(crate) struct Scheduler {
...
     /// The slots that are free to use in each slot texture.
     free: [Vec<usize>; 2],
-    /// Slots that require clearing before subsequent draws for each slot texture.
-    clear: [Vec<u32>; 2],
     /// Rounds are enqueued on push clip commands and dequeued on flush.
     rounds_queue: VecDeque<Round>,
-    /// State for a single wide tile.
-    tile_state: TileState,
+    #[cfg(debug_assertions)]
+    clear: [Vec<u32>; 2],
```
grebmeg: I see there are some comments about the usage in `flush` method, but I think it would be helpful to add a comment here as well.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: If you think it's needed in the future, then I'm fine leaving it as is. I just thought it was not necessary if it's only for that one thing.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -349,85 +587,101 @@ impl Scheduler {
...
         }
-        for cmd in &tile.cmds {
+        state
+    }
+
+    /// Iterates over wide tile commands and schedules them for rendering.
+    ///
+    /// Returns `Some(command_idx)` if there is more work to be done. Returns `None` if the wide
+    /// tile has been fully consumed.
+    fn do_tile<'a, R: RendererBackend>(
```
grebmeg, line 645: While debugging implementation, I came up with the debugging output below. It might be useful, though you may already have something similar? If not, it could be worth adding later. It definitely lacks some information, but for now I’m just checking whether it would be useful. Where:
_: free
D: dest slot
T: temporary slot (valid)
t: temporary slot (invalid)
C: scheduled clear

image

ajakubowicz-canva: I like it! During development I used this pretty intense print out:
https://github.com/linebender/vello/pull/1155/commits/489d6db2d265c1d2cb9813bee051e237006454f5#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fL489-L653

I think we could consider some kind of debug information that would make the scheduler easier to understand. It could even be some method directly on the gpu strip, enabled via a feature flag.

I've omitted it any of those changes from this PR simply because it adds extra overhead and review burden. But I support adding something like that as an opt-in for understandability.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Yes! Tracked in https://github.com/linebender/vello/issues/1165

Specific comment is [this one](https://github.com/linebender/vello/pull/1155#discussion_r2286379214).

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Oh my golly - that's a great catch. Probably good to pull that out as a constant, e.g.

```rs
const SENTINEL_SLOT_IDX = usize::MAX
```

Will fix thank you.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Yes!

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Tracked in issue: https://github.com/linebender/vello/issues/1165

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Added: https://github.com/linebender/vello/pull/1155/files/2c37c28c1e850be60f30b96eeb12c0f845fe4e6f..cb29a620b951ab7bcee1bcc916ba4d32d39cd120#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR234

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Comment added: https://github.com/linebender/vello/pull/1155/files/2c37c28c1e850be60f30b96eeb12c0f845fe4e6f..cb29a620b951ab7bcee1bcc916ba4d32d39cd120#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR1134

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Tracked in https://github.com/linebender/vello/issues/1165

If you don't mind I'd like to address in a followup.

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Fixed in: https://github.com/linebender/vello/pull/1155/files/2c37c28c1e850be60f30b96eeb12c0f845fe4e6f..cb29a620b951ab7bcee1bcc916ba4d32d39cd120#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR201-R202

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Tracked for followup: https://github.com/linebender/vello/issues/1165

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: Renamed to `depth`: https://github.com/linebender/vello/pull/1155/files/2c37c28c1e850be60f30b96eeb12c0f845fe4e6f..cb29a620b951ab7bcee1bcc916ba4d32d39cd120#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR658

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: For now I've addressed with a comment: https://github.com/linebender/vello/pull/1155/commits/d5a80b53aa41263b11e4ec8b65dc9c62d8b44e2c#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR377

sparse_strips/vello_common/src/coarse.rs
ajakubowicz-canva: Awesome - will leave as-is. Thank you for taking a look :D

grebmeg (approved):
Great to see blending and compositing land in vello_hybrid — excellent work, Andrew! 🎉

LGTM!

sparse_strips/vello_hybrid/src/schedule.rs
ajakubowicz-canva: As a quick note: In most cases this loose assumption should still hold, as this copy-up is done in a `PushBuf` cmd.

However, you are right that this is dependent on how the layer is flattened, and if the layer alternates between being a clip layer and blend layer, then we'll waste time on the clip layers.

PR #1156, opened 2025-08-12
Add a getter method for the current render settings in place

LaurenzV:


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -53,6 +53,7 @@ pub struct RenderContext {
         allow(dead_code, reason = "used when the `text` feature is enabled")
     )]
     pub(crate) level: Level,
+    pub(crate) num_threads: u16,
```
sagudev: Why not just store `RenderSettings` here?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -84,33 +84,30 @@ impl Default for RenderSettings {
...
     }
 
     /// Create a new render context with specific settings.
     pub fn new_with(width: u16, height: u16, settings: &RenderSettings) -> Self {
-        Self::new_inner(width, height, settings.num_threads, settings.level)
+        Self::new_inner(width, height, *settings)
     }
 
-    fn new_inner(width: u16, height: u16, num_threads: u16, level: Level) -> Self {
+    fn new_inner(width: u16, height: u16, settings: RenderSettings) -> Self {
```
sagudev: I think we should just make this public:

```suggestion
    /// Create a new render context with specific settings.
    pub fn new_with(width: u16, height: u16, settings: RenderSettings) -> Self {
```

sparse_strips/vello_common/src/recording.rs
sagudev, line 1: This is why I prefer rebases.

sagudev (approved):

sparse_strips/vello_common/src/recording.rs
LaurenzV: Oops sorry, didn't notice this when pushing. 😅 Lemme fix it

PR #1157, opened 2025-08-12
Force the user to specify the render mode in the beginning

LaurenzV:
I think that's reasonable to do, one the one hand it would allow us making decisions based on the render mode in earlier parts of the pipeline (though I'm unsure we will ever need it), but more importantly, it allows us to get the render mode while we are still processing commands. 

For example, up until now, when rendering COLR glyphs, we always used `OptimizeQuality`, but now we can use whatever mode the user selected. Another motivation is that for SVG rendering, there are various situations where we wil need to create sub pixmaps (for rendering patterns or masks), and it would be good if we can use the same rendering mode for those.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -67,6 +67,13 @@ pub struct RenderSettings {
     /// The number of worker threads that should be used for rendering. Only has an effect
     /// if the `multithreading` feature is active.
     pub num_threads: u16,
+    /// Whether to prioritize speed or quality when rendering.
+    ///
+    /// For most cases (especially for real-time rendering), it is highly recommented to set
+    /// this to `OptimizeSpeed`. If accuracy is a more significant concern (for example for visual
+    /// regression testing), then you can set this to `OptimizeQuality`.
+    ///
+    pub render_mode: RenderMode,
 }
 
 impl Default for RenderSettings {
@@ -80,6 +87,7 @@ impl Default for RenderSettings {
                 .saturating_sub(1) as u16,
             #[cfg(not(feature = "multithreading"))]
             num_threads: 0,
+            render_mode: RenderMode::OptimizeSpeed,
         }
     }
 }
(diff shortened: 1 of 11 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -64,6 +64,13 @@ pub struct RenderSettings {
     /// The number of worker threads that should be used for rendering. Only has an effect
     /// if the `multithreading` feature is active.
     pub num_threads: u16,
+    /// Whether to prioritize speed or quality when rendering.
+    ///
+    /// For most cases (especially for real-time rendering), it is highly recommented to set
+    /// this to `OptimizeSpeed`. If accuracy is a more significant concern (for example for visual
+    /// regression testing), then you can set this to `OptimizeQuality`.
+    ///
+    pub render_mode: RenderMode,
```
nicoburns: I recieved feedback from sagudev that we ought to be more clear about exactly the difference in the implementations of the different modes is. It was not clear to them that they ought to be using `OptimiseSpeed`. I'm thinking we ought to be specific that it's u8 vs f32 (but perhaps with wording which allows for that changing in future).

LaurenzV: Makes sense, will expand on that.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -67,6 +67,16 @@ pub struct RenderSettings {
...
     pub num_threads: u16,
+    /// Whether to prioritize speed or quality when rendering.
+    ///
+    /// For most cases (especially for real-time rendering), it is highly recommended to set
+    /// this to `OptimizeSpeed`. If accuracy is a more significant concern (for example for visual
+    /// regression testing), then you can set this to `OptimizeQuality`.
+    ///
+    /// Currently, the only difference this makes is that when choosing `OptimizeSpeed`, rasterization
+    /// will happen using u8/u16, while `OptimizeQuality` will use a f32-based pipeline.
+    ///
```
tomcur: Trailing newline.

```suggestion
```

tomcur (approved):
Makes sense to me!

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -419,10 +429,15 @@ impl RenderContext {
...
-    pub fn render_to_pixmap(&self, pixmap: &mut Pixmap, render_mode: RenderMode) {
+    pub fn render_to_pixmap(&self, pixmap: &mut Pixmap) {
         let width = pixmap.width();
         let height = pixmap.height();
-        self.render_to_buffer(pixmap.data_as_u8_slice_mut(), width, height, render_mode);
+        self.render_to_buffer(
+            pixmap.data_as_u8_slice_mut(),
+            width,
+            height,
+            self.render_settings.render_mode,
```
T-256, line 439: I think it could move to `render_to_buffer`. In my usecase I'm switching from creating `Pixmap` (because it doesn't accept custom buffer via reference pointer) to make it cheaper by directly pass `&mut buffer`. here is before and after:

- before:
```rs
render_context.render_to_pixmap(pixmap);
```
- after:
```rs
                render_context.render_to_buffer(
                    buffer,
                    render_context.width(),
                    render_context.height(),
                    vello_cpu::RenderMode::OptimizeSpeed,
                );
```

PR #1159, opened 2025-08-13
Add support for non-isolated blending in `vello_cpu`

LaurenzV:
I also removed the blend optimization since IMO it causes more troubles than it brings benefits.

Reference images match the ones in `tiny-skia`.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -354,6 +354,11 @@ impl RenderContext {
     pub fn paint(&self) -> &PaintType {
         &self.paint
     }
+    
+    /// Set the blend mode that should be used when drawing objects.
+    pub fn set_blend_mode(&mut self, blend_mode: BlendMode) {
+        self.blend_mode = blend_mode;
+    } 
 
     /// Set the current paint transform.
     ///
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -47,6 +47,7 @@ pub struct RenderContext {
     pub(crate) stroke: Stroke,
     pub(crate) transform: Affine,
     pub(crate) fill_rule: Fill,
+    pub(crate) blend_mode: BlendMode,
     pub(crate) temp_path: BezPath,
     // TODO: Consider taking a configurable threshold instead of just a boolean value here.
     pub(crate) anti_alias: bool,
@@ -136,6 +137,7 @@ impl RenderContext {
             dispatcher,
             transform,
             anti_alias,
+            blend_mode: BlendMode::new(Mix::Normal, Compose::SrcOver),
             paint,
             paint_transform,
             fill_rule,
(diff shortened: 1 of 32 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -339,6 +355,11 @@ impl RenderContext {
         &self.paint
     }
 
+    /// Set the blend mode that should be used when drawing objects.
+    pub fn set_blend_mode(&mut self, blend_mode: BlendMode) {
+        self.blend_mode = blend_mode;
```
sagudev, line 385: We also need to reset it on reset.

sagudev (commented):
I am testing this on servo WPT and I did a quick look. I think this should be reviewed by someone with deeper understanding of sparse strips.

sagudev:
Hm, canvas does not agree with alpha handling I think: https://github.com/sagudev/servo/actions/runs/16936941963
I need to investigate more.

sparse_strips/vello_sparse_tests/snapshots/compose_non_isolated_copy.png
sagudev, line 1: This looks wrong. Here only source should be present (other should be cleared).

sparse_strips/vello_sparse_tests/snapshots/compose_non_isolated_dest_in.png
sagudev, line 1: This  look wrong because the result is the same as next one.

sagudev:
New proper results: https://github.com/sagudev/servo/actions/runs/16947826831/attempts/1#summary-48036644654.
So while canvas has actually isolated blending, this PR allows us to use less layers (more perf) and properly respect clipping.

We should still fix #1119 to match Vello classic.

sparse_strips/vello_cpu/src/render.rs
sagudev: Can we also get getter for this like we have for the rest (for consistency)?

LaurenzV:
@sagudev Do you think you would have time to do a high-level review and approve if it looks okay? Otherwise I feel like this PR will remain in limbo forever. 😄 It's not that big of a change so I don't think landing this should cause any trouble, especially since existing blending tests work just fine.

sagudev (approved):
Looks alright from higher perspective.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -425,6 +442,7 @@ impl RenderContext {
         self.reset_paint_transform();
         #[cfg(feature = "text")]
         self.glyph_caches.as_mut().unwrap().maintain();
+        self.blend_mode = BlendMode::new(Mix::Normal, Compose::SrcOver);
```
T-256, line 445: Couldn't use `BlendMode::default()` consistently in here and other places?

LaurenzV: Nope, because right now the default is Mix::Clip which I want to avoid using.

sagudev: Mix::Clip is not default anymore: https://docs.rs/peniko/latest/src/peniko/blend.rs.html#184-191, see https://github.com/linebender/peniko/pull/144.

LaurenzV: Oh! I guess we can change it then!

grebmeg:
Hi @LaurenzV, would you mind updating the `CHANGELOG.md` files in both `vello_common` and `vello_cpu` to reflect the recent updates? It’ll ensure everything is documented properly before release. Really appreciate it! 🙏

LaurenzV:
Oh yeah sure, will try to do that soon!

LaurenzV:
I opened a PR.

PR #1161, opened 2025-08-14
Alternative API for image overrides

### LINKED ISSUE #664: Better API for already GPU-resident images

DJMcNab:
Fixes #664

Supersedes (and integrates) #1052

The proposal in #1052 is an ergonomics improvement, but has several shortfalls. In particular, it required special case methods for every API which used an external texture. As these only had partial coverage, you couldn't use an external texture to render text (or any stroked path).

This is a much smaller proposal based on the core of that PR, to instead return a (fake) Peniko `Image`, which is then compatible with every Scene method which would draw an image.

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -312,6 +312,15 @@ pub enum Error {
 )]
 pub(crate) type Result<T, E = Error> = std::result::Result<T, E>;
 
+/// An opaque handle to a texture registered with [`Renderer::register_texture`] that
+/// can be passed to [`Scene::draw_texture`] to draw the texture.
+#[derive(Copy, Clone, PartialEq, Hash)]
+pub struct TextureHandle {
+    pub(crate) id: u64,
+    pub(crate) width: u32,
+    pub(crate) height: u32,
+}
+
 /// Renders a scene into a texture or surface.
 ///
 /// Currently, each renderer only supports a single surface format, if it
@@ -522,6 +531,7 @@ impl Renderer {
     /// dimensions would be rendered.
     ///
     /// [data]: peniko::Image::data
+    #[deprecated = "Use register_texture and unregister_texture methods instead"]
     pub fn override_image(
         &mut self,
         image: &peniko::Image,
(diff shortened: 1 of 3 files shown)


sagudev (approved):
This is great.

nicoburns (approved):
This seems like a good minimal change to improve ergonomics.

Medium-to-longer term, I am looking at the API `vello_hybrid` exposes for images where they require *all* images (not just textures) to be pre-registered like this, and then the image data passed to the draw calls is just an ID. I suspect we might want to look at doing similar for Vello Classic.

sagudev:
> Medium-to-longer term, I am looking at the API `vello_hybrid` exposes for images where they require _all_ images (not just textures) to be pre-registered like this, and then the image data passed to the draw calls is just an ID. I suspect we might want to look at doing similar for Vello Classic.

Would this also avoid atlas texture?

nicoburns:
> Would this also avoid atlas texture?

I believe not yet. For that I believe we need https://github.com/gfx-rs/wgpu/issues/3637 which works today on native platforms, but vello is also targeting web and "bindless" isn't yet available there.

For CPU-side (non-changing) images I believe it does enable the image to be upload to the Atlas once and then referenced directly. @grebmeg could confirm the details.

DJMcNab:
Yeah - we know that this needs to change, and an explicit "upload" step is the right way for that to happen.

This is just dealing with the API we have now, to simplify use cases without going overboard on complexity.

grebmeg:
> For CPU-side (non-changing) images I believe it does enable the image to be upload to the Atlas once and then referenced directly. @grebmeg could confirm the details.

If this is about `vello_cpu`, I don’t think we support atlases there — but we do in `vello_hybrid`. The main reason is that, at the moment, the only way to use external wgpu textures is by creating multiple bind groups with attached textures and then issuing multiple draw calls with the relevant bindings. I suspect this approach will prove less performant and add unnecessary complexity to the scheduler. For webgl2, I don’t think there’s a more efficient way to render thousands of images than with an atlas? Therefore, my view is that bindless support in wgpu would allow us to work with external textures much more efficiently, though this benefit wouldn’t extend to WebGL.

PR #1163, opened 2025-08-15
Extract `StripGenerator` into `vello_common` and hook it up into `vello_hybrid`

### LINKED ISSUE #1154: Move `StripGenerator` into `vello_common`?
I believe this abstraction is general enough that it should also be useful for `vello_hybrid`, so maybe we can move it into `vello_common`?

LaurenzV:
Fixes #1154.

### DIFF
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -18,8 +18,7 @@ use vello_common::peniko::color::palette::css::BLACK;
 use vello_common::peniko::{BlendMode, Compose, Fill, Mix};
 use vello_common::recording::{PushLayerCommand, Recordable, Recording, RenderCommand};
 use vello_common::strip::Strip;
-use vello_common::tile::Tiles;
-use vello_common::{flatten, strip};
+use vello_common::strip_generator::StripGenerator;
 
 /// Default tolerance for curve flattening
 pub(crate) const DEFAULT_TOLERANCE: f64 = 0.1;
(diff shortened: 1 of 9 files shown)


sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -134,32 +120,46 @@ impl Scene {
...
-            &mut self.line_buf,
-            &mut self.flatten_ctx,
-        );
+
         let paint = self.encode_current_paint();
-        self.render_path(self.fill_rule, paint);
+        self.fill_path_with(path, self.transform, self.fill_rule, paint, self.anti_alias);
+    }
+
+    fn fill_path_with(
```
grebmeg, line 129: [nit]: It might be helpful to add short comments for `fill_path_with` and `stroke_path_with`.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -588,21 +558,15 @@ impl Scene {
...
-            self.level,
+        self.strip_generator.generate_stroked_path(
             path,
             &self.stroke,
             transform,
-            &mut self.line_buf,
-            &mut self.flatten_ctx,
+            self.anti_alias,
+            |generated_strips| {
+                strips.extend_from_slice(generated_strips);
```
grebmeg, line 566: Would it make sense to pass `generated_strips` as an owned value to the generation strips functions? The benefit is that we could use append?

grebmeg (approved):
Nice! 🎉 LGTM

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: I don't think so, because then we can't reuse the strip buffer allocation, which is pretty important I think.

PR #1164, opened 2025-08-19
Fix hybrid examples to not use `std::time` in Wasm

taj-p:
std::time isn't implemented in wasm, so we cannot access `std::time::Instant` on these platforms:

image


DJMcNab:
In https://github.com/linebender/vello/pull/595, we used `web-time` for Vello Classic's `with_winit`.

PR #1170, opened 2025-08-22
Limit number of threads to 8 by default in multi-threaded rendering

LaurenzV:
Still not sure what's going on, but when trying 16 threads on my Intel machine instead of 8 I'm seeing a sudden very heavy performance drop. Let's limit it to 8 by default for now until we have figured this out.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -77,7 +77,8 @@ impl Default for RenderSettings {
             num_threads: (std::thread::available_parallelism()
                 .unwrap()
                 .get()
-                .saturating_sub(1) as u16).min(8),
+                .saturating_sub(1) as u16)
+                .min(8),
             #[cfg(not(feature = "multithreading"))]
             num_threads: 0,
         }


sagudev:
> Still not sure what's going on, but when trying 16 threads on my Intel machine instead of 8 I'm seeing a sudden very heavy performance drop. Let's limit it to 8 by default for now until we have figured this out.

Which processor you have? Is it possible you only have 8 physical cores (each acting as 2 logical cores - hyper threading)?

LaurenzV:
Intel i9-14900K

sagudev:
> Intel i9-14900K

I guess you are out of performance cores:

>Total Cores 24
>Performance-cores 8
>Efficient-cores 16
>Total Threads 32

LaurenzV:
Hmm I re-ran the benchmarks now and the crazy slowdown isn't there anymore (it's not much faster either though, but that makes sense given the reason you mentioned). But still, I think it doesn't make much sense to use more than 8 cores because at some point, coarse rasterization is going to be a bottleneck, so using 8 as the maximum _by default_ seems sensible to me.

tomcur:
Is it possible there was so much synchronization going on, with not enough actual work, that the processor clocked down, if you're in a balanced or efficiency power governor mode?

LaurenzV:
Hard to say. 😄 As I mentioned,  haven't looked into it too deeply.

LaurenzV:
Can we merge this then or? I don't think we got to discuss this during office hours.

taj-p (approved):
LGTM - are you able to open an issue documenting that we need to research why this is happening (if you haven't already 🙏 )

---

I took a look awhile ago, but thought it was already approved, sorry!

raphlinus (approved):

PR #1171, opened 2025-08-22
[draft] vello_hybrid compositing without suspending but using pre-planning of rounds

ajakubowicz-canva:
### Context

In rendering office hours it was mentioned that instead of suspending we could try and pre-plan all the future rounds.

Here is my attempt!

### Tradeoffs

Pros:
 - This solution can't deadlock.
 - This solution passes all the tests.

Cons:
 - In this state the performance is terrible. It's an extreme performance regression.

This is a performance regression because it doesn't track the claiming of slots across rounds. This means that one `do_tile` can claim slots over ~30 rounds, and then start forcing flushes as new slots cannot be claimed. The round flush then ends up containing extremely sparse work.

I haven't yet figured out a performant solution.


ajakubowicz-canva:
Ha! Turns out there was a bug in round calculation... This might just work! Another PR coming.

PR #1173, opened 2025-08-23
Support BGRA images in Vello Classic

sagudev:
This is possible because of https://github.com/linebender/peniko/pull/120.


vello_shaders/shader/fine.wgsl
```diff
@@ -1198,7 +1212,7 @@ fn main(
                                 let atlas_uv_clamped = clamp(atlas_uv, image.atlas_offset, atlas_max);
                                 // Nearest neighbor sampling
                                 let fg_rgba = premul_alpha(textureLoad(image_atlas, vec2<i32>(atlas_uv_clamped), 0));
-                                let fg_i = fg_rgba * area[i] * image.alpha;
+                                var fg_i = pixel_format(fg_rgba * area[i] * image.alpha, image.format);
```
DJMcNab: Why is this now changed to a `var`?

vello_shaders/shader/fine.wgsl
```diff
@@ -830,6 +831,19 @@ fn read_end_clip(cmd_ix: u32) -> CmdEndClip {
     return CmdEndClip(blend, alpha);
 }
 
+const PIXEL_FORMAT_RGBA: u32 = 0u;
+const PIXEL_FORMAT_BGRA: u32 = 1u;
+fn pixel_format(pixel: vec4f, format: u32) -> vec4f {
```
DJMcNab, line 837:
```suggestion
// Normalises subpixel order loaded from an image, based on the image's format.
fn pixel_format(pixel: vec4f, format: u32) -> vec4f {
```

vello_shaders/shader/fine.wgsl
```diff
@@ -830,6 +831,19 @@ fn read_end_clip(cmd_ix: u32) -> CmdEndClip {
...
+    switch format {
+        case PIXEL_FORMAT_BGRA: {
+            return pixel.bgra;
```
DJMcNab, line 841:
```suggestion
            // The conversion from RGBA to BGRA is its own inverse.
            return pixel.bgra;
```

DJMcNab (commented):
This implementation looks reasonable to me.

We will need to bring this up at renderer office hours, but I think we can merge it optimistically before then.

However, I would like to see at least one test which uses this before landing.
The changelog update will need to be folded in with #1231.

vello_shaders/shader/fine.wgsl
sagudev: This was from before I introduced `pixel_format` function.

sagudev:
> To clarify the state, this is still lightly blocked on a test.

I know, I am currently on vacation so progress is slow. I will rerequest review when ready.

> This only needs to be similar to: 

This is great example, thanks.

DJMcNab:
Yeah, there's no rush! I'm just noting it e.g. for office hours, so we can keep track of an accurate state

vello_encoding/src/encoding.rs
```diff
@@ -449,10 +446,11 @@ impl Encoding {
             .extend_from_slice(bytemuck::cast_slice(bytemuck::bytes_of(&DrawImage {
                 xy: 0,
                 width_height: (brush.image.width << 16) | (brush.image.height & 0xFFFF),
-                sample_alpha: ((quality as u32) << 12)
+                sample_alpha: ((brush.image.format as u32) << 13
```
DJMcNab: Should this be `14`, since `ImageQuality` has three variants?

vello_shaders/shader/fine.wgsl
```diff
@@ -813,15 +813,16 @@ fn read_image(cmd_ix: u32) -> CmdImage {
     let width_height = info[info_offset + 7u];
     let sample_alpha = info[info_offset + 8u];
     let alpha = f32(sample_alpha & 0xFFu) / 255.0;
-    let quality = sample_alpha >> 12u;
+    let format = sample_alpha >> 13u;
+    let quality = (sample_alpha >> 12u) & 0x1u;
```
DJMcNab: Hmm, so we actually just treat ImageQuality::High as ImageQuality::Low?

vello_tests/tests/smoke_snapshots.rs
```diff
@@ -112,3 +112,31 @@ fn two_emoji_gpu() {
 fn two_emoji_cpu() {
     two_emoji(true);
 }
+
+#[test]
+#[cfg_attr(skip_gpu_tests, ignore)]
+fn test_bgra() {
```
DJMcNab:
```suggestion
fn bgra_image() {
```

vello_tests/tests/smoke_snapshots.rs
```diff
@@ -112,3 +112,31 @@ fn two_emoji_gpu() {
...
+    let image = ImageBrush::new(ImageData {
+        data: blob.into(),
+        format: ImageFormat::Bgra8,
+        width: 2,
+        height: 2,
+        alpha_type: ImageAlphaType::Alpha,
+    });
+    scene.draw_image(&image, Affine::IDENTITY);
+    smoke_snapshot_test_sync(scene, &TestParams::new("bgra", 2, 2))
+        .unwrap()
```
DJMcNab: The scene size being 2x2 is a little bit unfortunate for reading the test. KDE's desktop zoom functionality is best in class, so I can review the tests, but it could be worth making the test larger (or maybe redesigning it to be a property test, i.e. which performs readback).

This isn't blocking

vello_tests/tests/smoke_snapshots.rs
```diff
@@ -112,3 +112,31 @@ fn two_emoji_gpu() {
...
+    let mut blob: Vec<u8> = Vec::new();
+    [
+        palette::css::RED,
+        palette::css::LIME,
+        palette::css::BLUE,
+        palette::css::WHITE,
+    ]
+    .iter()
+    .for_each(|c| {
+        blob.extend(c.to_rgba8().to_u8_array());
```
DJMcNab: Is there a reason we have not swapped the `r` and `b` here?

We seem to currently be testing the ability to render "actually rgba" images wrong, rather than testing the ability to render bgra images correctly.

vello_tests/tests/smoke_snapshots.rs
```diff
@@ -112,3 +112,31 @@ fn two_emoji_gpu() {
...
+        data: blob.into(),
+        format: ImageFormat::Bgra8,
+        width: 2,
+        height: 2,
+        alpha_type: ImageAlphaType::Alpha,
+    });
+    scene.draw_image(&image, Affine::IDENTITY);
+    smoke_snapshot_test_sync(scene, &TestParams::new("bgra", 2, 2))
+        .unwrap()
+        .assert_mean_less_than(0.01);
```
DJMcNab: The CI failure seems somewhat spurious. I suspect it's because we use the closeness detection hashing thing, rather than the pixel-wise comparison operator used by the sparse strip tests.

In this case, just increasing the allowed mean is a reasonable thing; I've manually checked the images, and they aren't perceptually different.

DJMcNab (changes requested):
Thank you - this looks good. There are just a few details we should fix before landing this.

The concern about `ImageQuality` is blocking (plus CI, but as noted, just doing a hack for that is fine).

vello_encoding/src/encoding.rs
sagudev: Hm, I think it should. (Un)packing code feels unsafe, as there are some assumptions of types that can change. We should at least have some const asserts here for size assumptions (by looking using https://doc.rust-lang.org/std/primitive.u32.html#method.leading_zeros on max of types).

vello_shaders/shader/fine.wgsl
sagudev: Hm, `IMAGE_QUALITY_HIGH` is mapped to `IMAGE_QUALITY_MEDIUM` in shader, but I guess because of bitmasking this never happens. I will open issue about this.

DJMcNab: I'm confused why you're not just masking with `0x3u` here, so that the same passthrough is maintained.

sagudev: This is fixed in https://github.com/linebender/vello/pull/1173/commits/a83f1fd3bf39dbf4ad10808b05a974910b4e526e, honestly I forget I wrote that code so I just though it was on main.

vello_tests/tests/smoke_snapshots.rs
sagudev: I also did this in https://github.com/linebender/vello/pull/1173/commits/6dca9c59faa846357096849d50c560da1b7cb639, hopefully the test passes.

DJMcNab (approved):
LGTM. Thanks!

PR #1175, opened 2025-08-23
Remove unnecessary lifetime annotation

parasyte:
These lifetimes are `'static` because the wgpu `Surface` owns `Arc<Window>`.

### DIFF
--- a/examples/simple/src/main.rs
+++ b/examples/simple/src/main.rs
@@ -22,7 +22,6 @@ use vello::wgpu;
 enum RenderState {
     /// `RenderSurface` and `Window` for active rendering.
     Active {
-        // The `RenderSurface` and the `Window` must be in this order, so that the surface is dropped first.
         surface: Box<RenderSurface<'static>>,
         window: Arc<Window>,
     },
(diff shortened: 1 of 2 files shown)


examples/simple/src/main.rs
```diff
@@ -19,18 +19,18 @@ use winit::window::Window;
 use vello::wgpu;
 
 #[derive(Debug)]
-enum RenderState<'s> {
+enum RenderState {
     /// `RenderSurface` and `Window` for active rendering.
     Active {
         // The `RenderSurface` and the `Window` must be in this order, so that the surface is dropped first.
```
parasyte: I'm pretty sure this comment is false. The drop order does not matter because `Arc` always does the right thing.

DJMcNab: It used to be true until https://github.com/gfx-rs/wgpu/pull/6997. But yes, I think we can remove that now

DJMcNab (approved):
This seems reasonable. Let me know when you're ready for this to land (i.e. whether or not you want to remove the comment).

parasyte:
@DJMcNab Sounds good! I have removed the comment, and this is ready to go.

PR #1178, opened 2025-08-27
vello_hybrid: add `mix` blend layer support

ajakubowicz-canva:
Part of: https://github.com/linebender/vello/issues/1165

### Context

This PR extends the `render_strips.wgsl` shader to also support `mix` with compositing.

### How

Functionally this is a very simple PR as we're leveraging the Vello classic shader math. The shaders are taken from [`blend.wgsl`](https://github.com/linebender/vello/blob/8acbf60b6d9271030a710b1075147bf842938c27/vello_shaders/shader/shared/blend.wgsl#L142).

### Test plan

All the existing `mix` tests currently depend on gradients (which `vello_hybrid` doesn't support). For this reason I've added two new "matrix" tests which cover all the cases in a single test without using gradient. This enables early test coverage for `vello_hybrid` without blocking on gradients.

Once gradients are supported we can enable all the rest of the `mix` tests.

### DIFF
--- a/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
@@ -73,7 +73,6 @@ const MIX_HUE = 12u;
 const MIX_SATURATION = 13u;
 const MIX_COLOR = 14u;
 const MIX_LUMINOSITY = 15u;
-const MIX_CLIP = 128u;
 
 struct Config {
     // Width of the rendering target
--- a/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
@@ -56,6 +56,25 @@ const COMPOSE_XOR: u32 = 11u;
 const COMPOSE_PLUS: u32 = 12u;
 const COMPOSE_PLUS_LIGHTER: u32 = 13u;
 
+// Mix modes
+const MIX_NORMAL = 0u;
+const MIX_MULTIPLY = 1u;
+const MIX_SCREEN = 2u;
+const MIX_OVERLAY = 3u;
+const MIX_DARKEN = 4u;
+const MIX_LIGHTEN = 5u;
+const MIX_COLOR_DODGE = 6u;
+const MIX_COLOR_BURN = 7u;
+const MIX_HARD_LIGHT = 8u;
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_sparse_tests/snapshots/mix_compose_combined_test_matrix.png
ajakubowicz-canva, line 1: This test was added specifically to exercise https://github.com/linebender/vello/pull/1178/files#diff-7edfea1a687804789f874a2ff3dd501393d9ffdef9433249c57e4c4e9a61a529R369 which requires a composite that _is not_ `SrcOver` with a non-normal mix. It expands the test coverage for all renderers.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -56,6 +56,25 @@ const COMPOSE_XOR: u32 = 11u;
...
+const MIX_COLOR_BURN = 7u;
+const MIX_HARD_LIGHT = 8u;
+const MIX_SOFT_LIGHT = 9u;
+const MIX_DIFFERENCE = 10u;
+const MIX_EXCLUSION = 11u;
+const MIX_HUE = 12u;
+const MIX_SATURATION = 13u;
+const MIX_COLOR = 14u;
+const MIX_LUMINOSITY = 15u;
+const MIX_CLIP = 128u;
```
taj-p: Is `MIX_CLIP` ever used? If it isn't, we could reduce the number of bits reserved for mixes from 7 to 4 IIUC

taj-p (approved):
LGTM 🚀

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
taj-p: Edit: If it isn't used, is it also worth keeping around?

ajakubowicz-canva: Removed in: https://github.com/linebender/vello/pull/1178/commits/6647a27286cd66852e586fba04f326449df1be7f

Treating the bit packing as an optimization exercise for the future.

taj-p: Are you able to add a TODO (or similar) if it's a far future exercise so we don't forget?

PR #1180, opened 2025-08-27
Improve tolerance calculation for stroke expansion

tomcur:
Motivated by the discussion in and around #977.

We'd want to mark the `Ellipse::radii` method as `#[inline(always)]` in Kurbo.

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -133,8 +133,14 @@ pub fn stroke(
     line_buf: &mut Vec<Line>,
     flatten_ctx: &mut FlattenCtx,
 ) {
-    // TODO: Temporary hack to ensure that strokes are scaled properly by the transform.
-    let tolerance = TOL / affine.as_coeffs()[0].abs().max(affine.as_coeffs()[3].abs());
+    // A transform can be decomposed to a rotation, a scale along the x and y axes, and another
+    // rotation. We are interested in knowing the max scale along those axes. This is the same as
+    // finding the semi-major axis of the ellipse created by transforming the unit circle.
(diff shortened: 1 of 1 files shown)


tomcur (commented):
I believe we can calculate a simpler conservative bound, perhaps something like `max(|a| + |b|, |c| + |d|)`, but I haven't fully thought it through yet. Alternatively, as mentioned in https://github.com/linebender/vello/pull/977, the Frobenius norm may also work: `sqrt(a^2 + b^2 + c^2 + d^2)`.

It would be nice to have a quicker-to-calculate bound than the ellipse (which does a [singular value decomposition](https://github.com/linebender/kurbo/blob/c0cefe49fd7cff0c8994e220053efa5b3363e2b9/kurbo/src/affine.rs#L371-L415)), though it has to be balanced with the additional downstream work a conservative bound causes.

sparse_strips/vello_sparse_tests/snapshots/tricky_strokes.png
tomcur, line 1: Under the old calculation (after the hack in https://github.com/linebender/vello/pull/977), tolerance would always be `TOL` or smaller (i.e., could only become stricter). With the new calculation, `TOL` can become greater as well (i.e., looser). That appears to happen here, causing some differences in pixel coverage.

For example, the first stroke has a scale of approximately `0.591` and on `main` stroke expansion is performed with a tolerance of `0.25` (the value of `TOL`), whereas in this change the tolerance is approximately `0.423`.

tomcur:
This is ready for review, as we're now on Kurbo 0.13, which [optimized the required singular value decomposition](https://github.com/linebender/kurbo/pull/499) and [added the `Ellipse::radii` method](https://github.com/linebender/kurbo/pull/499).

PR #1183, opened 2025-08-29
Support a special "Luminance mask" layer type

DJMcNab:
Supersedes https://github.com/linebender/vello/pull/1121

cc @keavon does this address your need/use case?

cc @mTvare6, as this impacts your PRs

### DIFF
--- a/vello_encoding/src/draw.rs
+++ b/vello_encoding/src/draw.rs
@@ -198,7 +198,10 @@ pub struct DrawBeginClip {
 impl DrawBeginClip {
     /// The `blend_mode` used to indicate that a layer should be
     /// treated as a luminance mask.
-    pub const LUMINANCE_MASK_LAYER: u32 = 0xffffffff;
+    ///
+    /// The least significant 16 bits are reserved for Mix + Compose
+    /// combinations.
+    pub const LUMINANCE_MASK_LAYER: u32 = 0x10000;
 
     /// Creates new clip draw data for a Porter-Duff blend mode.
     pub fn new(blend_mode: BlendMode, alpha: f32) -> Self {
--- a/vello_shaders/shader/fine.wgsl
+++ b/vello_shaders/shader/fine.wgsl
@@ -29,7 +29,7 @@ const IMAGE_QUALITY_LOW = 0u;
 const IMAGE_QUALITY_MEDIUM = 1u;
 const IMAGE_QUALITY_HIGH = 2u;
 
-const LUMINANCE_MASK_LAYER = 0xffffffffu;
+const LUMINANCE_MASK_LAYER = 0x10000u;
 
 @group(0) @binding(2)
 var<storage> ptcl: array<u32>;


nicoburns:
Would it make sense to generalise this new method to allow alpha masks too? I suspect that might be quite convenient for implementing https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image...

DJMcNab:
I suppose so; it's not entirely clear to me whether that is already covered by an existing blend mode. But this operation has been discussed as blocking for Graphite, whereas alpha masks have not been.

Either way, that should be a separate PR to this (but before 0.6.0, probably).

mTvare6:
> Supersedes #1121
> 
> cc @Keavon does this address your need/use case?
> 
> cc @mTvare6, as this impacts your PRs

It does address the use case and works correctly, thanks.

vello_shaders/shader/fine.wgsl
```diff
@@ -1017,7 +1019,21 @@ fn main(
                     }
                     let bg = unpack4x8unorm(bg_rgba);
                     let fg = rgba[i] * area[i] * end_clip.alpha;
-                    rgba[i] = blend_mix_compose(bg, fg, end_clip.blend);
+                    if end_clip.blend == LUMINANCE_MASK_LAYER {
+                        // TODO: Does this case apply more generally?
+                        // See https://github.com/linebender/vello/issues/1061
+                        // TODO: How do we handle anti-aliased edges here?
```
dfrg, line 1025: Presumably they're accounted for in the multiplication of `rgba[i]` and `area[i]` above?

vello_shaders/shader/fine.wgsl
```diff
@@ -1226,6 +1242,8 @@ fn main(
         let coords = xy_uint + vec2(i, 0u);
         if coords.x < config.target_width && coords.y < config.target_height {
             let fg = rgba[i];
+            // TODO: Do we want the semantics of base_color to actually be a porter-duff over?
```
dfrg: I thought `rgba` was initialized to `base_color` in the entry point.

vello_shaders/shader/shared/blend.wgsl
```diff
@@ -218,11 +223,12 @@ fn blend_compose(
     cs: vec3<f32>,
     ab: f32,
     as_: f32,
-    mode: u32
+    compose_mode: u32,
+    mix_mode: u32
```
dfrg: It doesn't seem like we actually use `mix_mode` in this function.

vello_shaders/shader/shared/blend.wgsl
```diff
@@ -283,20 +289,25 @@ fn blend_compose(
...
+    // Max with a small epsilon to avoid NaNs.
+    let inv_alpha = 1.0 / max(color.a, EPSILON);
+    return color.rgb * inv_alpha;
+}
+
 // Apply color mixing and composition. Both input and output colors are
 // premultiplied RGB.
 fn blend_mix_compose(backdrop: vec4<f32>, src: vec4<f32>, mode: u32) -> vec4<f32> {
-    let BLEND_DEFAULT = ((MIX_NORMAL << 8u) | COMPOSE_SRC_OVER);
     let EPSILON = 1e-15;
```
dfrg: I think `EPSILON` can be removed now.

vello_encoding/src/draw.rs
```diff
@@ -196,13 +196,25 @@ pub struct DrawBeginClip {
 }
 
 impl DrawBeginClip {
-    /// Creates new clip draw data.
+    /// The `blend_mode` used to indicate that a layer should be
+    /// treated as a luminance mask.
+    pub const LUMINANCE_MASK_LAYER: u32 = 0xffffffff;
```
dfrg: Any reason for this particular value? If we want to add more special blend modes, seems like a better choice for this would be something like `1 << 16` (assuming the current modes occupy the low 16 bits).

dfrg (approved):
I didn't do a deep dive on the math but the pipeline changes LGTM and everything else seems reasonable with a few nits.

vello_shaders/shader/fine.wgsl
DJMcNab: Yes. This comment is #1184, and I'll remove it from here. The difference again arises with destructive layer kinds, including luminance masks.

vello_shaders/shader/fine.wgsl
DJMcNab: Kind of; `area[i]` indicates "how much the blending should apply to this pixel".

However, we're instead using it as an alpha mask over the foreground. For Porter-Duff Over, this is equivalent. However, for the "destructive" blend modes, that doesn't give anywhere near the same results.
In the extreme case, consider the case where `area[i]` is zero; that is the cause of #1061.

vello_shaders/shader/shared/blend.wgsl
DJMcNab: Good catch! This was due to wanting to credit `mTvare6 `, so I built off their commit.

vello_encoding/src/draw.rs
DJMcNab: Not particularly; I'm happy to change it to `0x10000`

vello_shaders/shader/fine.wgsl
dfrg: Makes sense. That seems unrelated to this PR and shouldn't block landing IMO. Let's try to resolve that separately.

dfrg: As an aside, that issue is interesting because I don't think we anticipated that someone would push/pop an empty layer simply for the side effects of the blend operation.

DJMcNab: Ahah:

> Note: Shape is defined by the mathematical description of the shape. A particular point is either inside the shape or it is not. There are no gradations.

https://drafts.fxtf.org/compositing/#whatiscompositing

PR #1184, opened 2025-08-29
Vello Classic: Draw "over" the base colour

DJMcNab:
Previously, the `base_color` is used to initialise the backdrop, but if e.g. the backdrop were overwritten (such as by a destructive blend operation - see #1183), then it was possible to "punch through" the background, which effectively looked like drawing onto a transparent black layer.

The biggest impact here is around transparency, as the default base colour *normally* ensures that no transparency reaches the render target, but with destructive layers, it is possible for it to. This means that detail in the alpha channel can be lost.

(I'm not completely sure what the right behaviour here is; this is open to discussion. Note that the old version of `base_color` can still be cheaply emulated, whereas emulating this behaviour requires adding a "dummy" layer around the entire scene)

### DIFF
--- a/vello_shaders/shader/fine.wgsl
+++ b/vello_shaders/shader/fine.wgsl
@@ -943,10 +943,6 @@ fn main(
     let xy = vec2(f32(global_id.x * PIXELS_PER_THREAD), f32(global_id.y));
     let local_xy = vec2(f32(local_id.x * PIXELS_PER_THREAD), f32(local_id.y));
     var rgba: array<vec4<f32>, PIXELS_PER_THREAD>;
-    let base_color = unpack4x8unorm(config.base_color);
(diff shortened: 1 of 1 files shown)


DJMcNab:
[There is some evidence in the spec](https://drafts.fxtf.org/compositing/#rootgroup) in favour of this reading:

> Browsers often use an infinite [white, 100% opaque](https://www.w3.org/TR/css-color-4/#sample) root group, for final compositing, but this is not required.

dfrg:
Seems like we might want to differentiate between a base color that is equivalent to an initial fill vs a base color that is “sticky” and survives destructive blends. Maybe a config option?

DJMcNab:
I'm not sure there are all that many use cases for such a non-sticky base colour; certainly in the web blending spec, every context seems to be either rendered *Over* the background, or into a fully transparent context (which is in turn rendered over the background).

PR #1189, opened 2025-09-04
Fix off-by-one issue with tiles that are to the right of the viewport

LaurenzV:
Maybe we can at some point figure out what was wrong with the original solution, but for now using the original fix which should be less error prone is a good start I think.

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -268,6 +268,10 @@ impl Tiles {
                 // Clamp all tiles that are strictly on the right of the viewport to the tile x coordinate
                 // right next to the outside of the viewport. If we don't do this, we might end up
                 // with too big tile coordinates, which will cause overflows in strip rendering.
+                // TODO: in principle it is possible to cull right-of-viewport tiles, but it was causing some
+                // issues, and we are choosing to do the less efficient but working thing for now.
+                // See <https://github.com/linebender/vello/pull/1189> and
+                // <https://github.com/linebender/vello/issues/1126>.
                 let x = (line_left_x as u16).min(tile_columns + 1);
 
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_common/src/tile.rs
```diff
@@ -262,17 +262,14 @@ impl Tiles {
...
-                }
-
                 let y_top_tiles = (line_top_y as u16).min(tile_rows);
                 let y_bottom_tiles = (line_bottom_y.ceil() as u16).min(tile_rows);
 
-                let x = line_left_x as u16;
+                // Clamp all tiles that are strictly on the right of the viewport to the tile x coordinate
+                // right next to the outside of the viewport. If we don't do this, we might end up
+                // with too big tile coordinates, which will cause overflows in strip rendering.
+                let x = (line_left_x as u16).min(tile_columns + 1);
```
tomcur: Perhaps add a TODO note to "track" this.

```suggestion
                // Clamp all tiles that are strictly on the right of the viewport to the tile x coordinate
                // right next to the outside of the viewport. If we don't do this, we might end up
                // with too big tile coordinates, which will cause overflows in strip rendering.
                // TODO: in principle it is possible to cull right-of-viewport tiles, but it was causing some
                // issues, and we are choosing to do the less efficient but working thing for now.
                // See <https://github.com/linebender/vello/pull/1189> and
                // <https://github.com/linebender/vello/issues/1126>.
                let x = (line_left_x as u16).min(tile_columns + 1);
```

LaurenzV:
> Do you know whether the issue always occurs on the bottom-most row of tiles?

I believe that for each row in the pixmap one pixel goes missing, so if the pixmap is high enough it can happen that multiple rows go missing.

PR #1190, opened 2025-09-04
Allow to use the `Recording` type with the multi-threading feature ac…

LaurenzV:
Essentially the laziest solution that I could find. All recordings will be done on the main thread, which means that if you render just recordings, there won't be any speed benefits (apart from the usual gains you get from caching in subsequent rendering operations). However, this solution works and ensures that enabling the `multithreading` feature doesn't suddenly cause the library to crash, which I think is what matters most here.

Figuring out a way to make the execution of recordings themselves multi-threaded can be future work.

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -27,6 +27,7 @@ use vello_common::fearless_simd::{Level, Simd, simd_dispatch};
 use vello_common::mask::Mask;
 use vello_common::paint::Paint;
 use vello_common::strip::Strip;
+use vello_common::strip_generator::StripGenerator;
 
 mod cost;
 mod small_path;
@@ -95,6 +96,8 @@ pub(crate) struct MultiThreadedDispatcher {
     task_idx: u32,
     /// The number of threads active in the thread pool.
     num_threads: u16,
+    /// The strip generator for the main thread, only used for recordings.
+    strip_generator: StripGenerator,
     level: Level,
     flushed: bool,
 }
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -488,6 +498,28 @@ impl Dispatcher for MultiThreadedDispatcher {
             }
         }
     }
+
+    fn generate_wide_cmd(
+        &mut self,
+        strip_buf: &[Strip],
+        fill_rule: Fill,
+        paint: Paint,
+        thread_idx: u8,
```
grebmeg: Can this ever be not 0? If it will always be 0 atm, should we disallow setting that value for now and just hardcode 0 below when registering the task?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -488,6 +498,28 @@ impl Dispatcher for MultiThreadedDispatcher {
...
+    ) {
+        // Note that we are essentially round-tripping here: The wide container is inside of the
+        // main thread, but we first send a render task to a child thread which basically just
+        // forwards it back to the main thread again. We cannot apply the wide command directly
+        // here because there might be other paths that are currently being processed in a child
+        // thread that should be rendered _before_ this wide command. Therefore, we treat it like
+        // any other render task so that we can utilize the same mechanism of assigning a task ID
+        // to ensure that they are executed in order.
```
grebmeg, line 509: My main question is how the order is preserved: is it due to using an ordered channel with task id assignments, which ensures the generated wide tiles maintain execution order?

grebmeg (approved):
I think this's a good start toward supporting recordings in `multithreading` feature! LGTM

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Yes exactly, the ordered_channel returns the render commands ordered by their task ID, which means that once we do coarse rasterization we are always preserving the order.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Fixed.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
grebmeg: Gotcha, thanks for confirming! This is why we use time cost estimations, to minimize situations where threads are waiting on each other, right? It’s really interesting to see how recordings will be properly supported in a multithreaded environment.

PR #1193, opened 2025-09-05
Enforce matching dimensions between render context and pixmap when rasterizing

### LINKED ISSUE #1191: Creating a `RenderContext` and `Pixmap` with differing dimensions can result in a crash
```
let mut ctx = RenderContext::new(700, 600);
let mut pixmap = Pixmap::new(700, 700);
let rect = Rect::new(0.0, 0.0, 700.0, 700.0);

ctx.push_layer(Some(&rect.to_path(0.1)), None, None, None);
ctx.pop_layer();

ctx.flush();
```

Maybe we can somehow enforce them to have the same dimension?

LaurenzV:
Fixes #1191.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -15,6 +15,7 @@ use alloc::boxed::Box;
 use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
+use log::warn;
 use vello_common::blurred_rounded_rect::BlurredRoundedRectangle;
 use vello_common::encode::{EncodeExt, EncodedPaint};
 use vello_common::fearless_simd::Level;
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -448,9 +449,19 @@ impl RenderContext {
     }
 
     /// Render the current context into a pixmap.
+    /// 
+    /// Note that the pixmap needs to have the same dimensions as the render context.
     pub fn render_to_pixmap(&self, pixmap: &mut Pixmap) {
         let width = pixmap.width();
         let height = pixmap.height();
+        
+        if width != self.width || height != self.height {
```
taj-p, line 458: Elsewhere, when invariants like this aren't obeyed, we typically panic:

https://github.com/linebender/vello/blob/main/sparse_strips%2Fvello_common%2Fsrc%2Fpixmap.rs#L43-L47

I'm wondering whether we should either return an Error or similarly panic (to at least be consistent).

LaurenzV: Good question, I’m not sure.

taj-p: I think the best option is allowing the consumer to decide what to do when _they_ break an invariant (i.e. return an Error). Vello hybrid can return an Error type IIRC when spatiotemporal allocation suffers slot exhaustion.

Otherwise, I think it's fine to defer that to a larger review of this question and simply panic here (to be consistent with the rest of the system).

LaurenzV:
Will close this until we revisit to come up with a more principled solution for this problem (in other places as well).

PR #1194, opened 2025-09-06
Fix an bug with clipping of sparse fills

### LINKED ISSUE #1186: Bug in clipping logic
The following code:
```rs
    let mut ctx = RenderContext::new(595, 20);
    let mut pixmap = Pixmap::new(595, 20);

    ctx.set_transform(Affine::translate((0.0, -700.0)));
    ctx.push_clip_layer(&BezPath::from_svg("M551.704,721.115 C465.024,716.424 375.466,706.552 289.699,688.737 C290.316,688.60205 290.935,688.466 291.55,688.33 C377.059,705.978 466.259,715.75 552.629,720.39 C552.32,720.632 552.013,720.87305 551.704,721.115").unwrap());
    ctx.push_clip_layer(&BezPath::from_svg("M-133.795,680.40704 C390.292,801.45905 763.166,503.67102 666.575,258.86005 C1031.16,797.18604 -452.803,1197.37 -133.795,680.40704").unwrap());
    ctx.fill_path(&Rect::new(0.0, 0.0, 595.0, 808.0).to_path(0.1));
    ctx.pop_layer();
    ctx.pop_layer();
    ctx.flush();

    // Encode the pixmap into a PNG file.
    let png = pixmap.into_png().unwrap();

    std::fs::write(format!("clip.png").to_ascii_lowercase(), png).unwrap();
```

results in the following picture:

Image

The black rectangle should not be there.

LaurenzV:
The basic problem is that as an optimization the `push_clip` function does not create `push_buf` commands for any wide tiles that would be completely filled. However, in `pop_clip` it can happen that due to the clamping of `x1`, we end up pushing a `ClipFill` command across the whole wide tile width. I don't think that this is the cleanest solution, but unfortunately I haven't come up with a different method that doesn't cause regressions in other cases. This fix works and doesn't cause any regressions in my integration tests.

Fixes #1186.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -719,7 +719,8 @@ impl<const MODE: u8> Wide<MODE> {
                 let width = clipped_x2.saturating_sub(clipped_x1);
 
                 // If there's a gap, fill it. Only do this if the fill wouldn't cover the
-                // whole tile, as such clips are skipped by the `push_clip` function.
+                // whole tile, as such clips are skipped by the `push_clip` function. See
+                // <https://github.com/linebender/vello/blob/de0659e4df9842c8857153841a2b4ba6f1020bb0/sparse_strips/vello_common/src/coarse.rs#L504-L516>
                 if width > 0 && width < WideTile::WIDTH {
                     let x_rel = u32::from(clipped_x1 % WideTile::WIDTH);
                     self.get_mut(cur_wtile_x, cur_wtile_y)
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_common/src/coarse.rs
```diff
@@ -718,8 +718,9 @@ impl<const MODE: u8> Wide<MODE> {
                 let clipped_x2 = x2.min((cur_wtile_x + 1) * WideTile::WIDTH);
                 let width = clipped_x2.saturating_sub(clipped_x1);
 
-                // If there's a gap, fill it
-                if width > 0 {
+                // If there's a gap, fill it. Only do this if the fill wouldn't cover the
+                // whole tile, as such clips are skipped by the `push_clip` function.
+                if width > 0 && width < WideTile::WIDTH {
```
grebmeg, line 724: I think this fix is fine and makes sense, since we treat long, fully filled wide tiles as a no-op for clipping when they cover the entire tile. The only thing I’d suggest is adding a reference to the relevant lines in `push_clip` where the skipping happens ([link](https://github.com/linebender/vello/blob/de0659e4df9842c8857153841a2b4ba6f1020bb0/sparse_strips/vello_common/src/coarse.rs#L504-L516), maybe this link is enough).

PR #1195, opened 2025-09-06
Optimize flattening of small curve segments

LaurenzV:
@tomcur Does this look to you like a valid optimization? 

Benchmarks before vs. after:
image
image
image


LaurenzV:
I'll close this and combine this with the PR for (maybe) introducing better flattening for small curves.

PR #1196, opened 2025-09-08
[`vello_hybrid`]: Gradient rendering

grebmeg:
This PR implements a comprehensive gradient rendering system for `vello_hybrid` that combines CPU-side preprocessing with GPU-accelerated rendering. The implementation supports three gradient types (linear, radial, and sweep) with efficient caching and shader processing.

image&nbsp;

Gradient rendering in `vello_hybrid` uses a two-phase approach combining CPU preprocessing with GPU acceleration. On the CPU side, gradients are processed through a caching system (`GradientRampCache`) that generates lookup tables (LUTs) for color interpolation. When a gradient is encountered, the system creates a cache key based on color stops, interpolation color space, and hue direction. If not cached, it generates a SIMD-optimized LUT containing pre-computed colors and stores it in a flat byte buffer. All gradient LUTs are packed into a single buffer and uploaded to a 2d `Rgba8Unorm` GPU texture for efficient sampling.

During GPU rendering, the fragment shader handles three gradient types (linear, radial, sweep) through a unified pipeline. Each fragment calculates its position in gradient space using transform matrices, then computes a parameter t specific to the gradient type: linear gradients use the x-coordinate, radial gradients calculate distance-based values (with sub-types for standard, strip, and focal gradients), and sweep gradients use fast polynomial approximation for angle calculations (faster than trigonometric functions). The t value is then used to sample the appropriate position in the gradient LUT texture, with extend modes applied for values outside [0,1].

The system has several optimizations: LRU caching prevents redundant LUT generation, SIMD LUT generation accelerates color interpolation, packed texture storage minimizes GPU memory usage, and the unified sampling function reduces shader complexity.

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -16,7 +16,9 @@ use alloc::vec;
 use alloc::vec::Vec;
 #[cfg(not(feature = "multithreading"))]
 use core::cell::OnceCell;
+use core::hash::{Hash, Hasher};
 use fearless_simd::{Simd, SimdBase, SimdFloat, f32x4, f32x16};
+use peniko::color::cache_key::{BitEq, BitHash, CacheKey};
 use smallvec::{SmallVec, ToSmallVec};
 // So we can just use `OnceCell` regardless of which feature is activated.
 #[cfg(feature = "multithreading")]
@@ -229,15 +231,19 @@ impl EncodeExt for Gradient {
         // step delta for a step in the x/y direction.
         let (x_advance, y_advance) = x_y_advances(&transform);
 
+        let cache_key = CacheKey(GradientCacheKey {
+            stops: self.stops.clone(),
+            interpolation_cs: self.interpolation_cs,
+            hue_direction: self.hue_direction,
+        });
+
         let encoded = EncodedGradient {
+            cache_key,
             kind,
             transform,
             x_advance,
             y_advance,
             ranges,
-            stops: ColorStops(stops.into_owned()),
-            interpolation_cs: self.interpolation_cs,
-            hue_direction: self.hue_direction,
             pad,
             has_opacities,
             u8_lut: OnceCell::new(),
@@ -702,6 +708,8 @@ pub enum EncodedKind {
 /// An encoded gradient.
 #[derive(Debug)]
 pub struct EncodedGradient {
+    /// The cache key for the gradient.
+    pub cache_key: CacheKey<GradientCacheKey>,
     /// The underlying kind of gradient.
     pub kind: EncodedKind,
     /// A transform that needs to be applied to the position of the first processed pixel.
@@ -712,12 +720,6 @@ pub struct EncodedGradient {
     pub y_advance: Vec2,
     /// The color ranges of the gradient.
     pub ranges: Vec<GradientRange>,
-    /// The original color stops.
-    pub stops: ColorStops,
-    /// The color space used for interpolation.
-    pub interpolation_cs: ColorSpaceTag,
-    /// The hue direction for cylindrical color space interpolation.
-    pub hue_direction: HueDirection,
     /// Whether the gradient should be padded.
     pub pad: bool,
     /// Whether the gradient requires `source_over` compositing.
@@ -737,6 +739,33 @@ impl EncodedGradient {
     pub fn f32_lut<S: Simd>(&self, simd: S) -> &GradientLut<f32> {
         self.f32_lut
             .get_or_init(|| GradientLut::new(simd, &self.ranges))
+    }
+}
+
+/// Cache key for gradient color ramps based on color-affecting properties.
+#[derive(Debug, Clone)]
+pub struct GradientCacheKey {
+    /// The color stops (offsets + colors).
+    pub stops: ColorStops,
+    /// Color space used for interpolation.
+    pub interpolation_cs: ColorSpaceTag,
+    /// Hue direction used for interpolation.
+    pub hue_direction: HueDirection,
+}
+
+impl BitHash for GradientCacheKey {
+    fn bit_hash<H: Hasher>(&self, state: &mut H) {
+        self.stops.bit_hash(state);
+        core::mem::discriminant(&self.interpolation_cs).hash(state);
+        core::mem::discriminant(&self.hue_direction).hash(state);
+    }
+}
+
+impl BitEq for GradientCacheKey {
+    fn bit_eq(&self, other: &Self) -> bool {
+        self.stops.bit_eq(&other.stops)
+            && self.interpolation_cs == other.interpolation_cs
+            && self.hue_direction == other.hue_direction
     }
 }
 
@@ -1012,12 +1041,13 @@ impl<T: FromF32Color> GradientLut<T> {
     }
 }
 
-fn determine_lut_size(ranges: &[GradientRange]) -> usize {
-    // Of course in theory we could still have a stop at 0.0001 in which case this resolution
-    // wouldn't be enough, but for all intents and purposes this should be more than sufficient
-    // for most real cases.
-    const MAX_LEN: usize = 4096;
+/// The maximum size of the gradient LUT.
+// Of course in theory we could still have a stop at 0.0001 in which case this resolution
+// wouldn't be enough, but for all intents and purposes this should be more than sufficient
+// for most real cases.
+pub const MAX_GRADIENT_LUT_SIZE: usize = 4096;
 
+fn determine_lut_size(ranges: &[GradientRange]) -> usize {
     // Inspired by Blend2D.
     // By default:
     // 256 for 2 stops.
@@ -1038,7 +1068,7 @@ fn determine_lut_size(ranges: &[GradientRange]) -> usize {
         // For example, if the first stop is at 0.001, then we need a resolution of at least 1000
         // so that we can still safely capture the first stop.
         let res = ((1.0 / (x1 - last_x1)).ceil() as usize)
-            .min(MAX_LEN)
+            .min(MAX_GRADIENT_LUT_SIZE)
             .next_power_of_two();
         min_size = min_size.max(res);
         last_x1 = x1;
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -9,7 +9,7 @@ use crate::color::{ColorSpaceTag, HueDirection, Srgb, gradient};
 use crate::kurbo::{Affine, Point, Vec2};
 use crate::math::{FloatExt, compute_erf7};
 use crate::paint::{Image, ImageSource, IndexedPaint, Paint, PremulColor};
-use crate::peniko::{ColorStop, Extend, Gradient, GradientKind, ImageQuality};
+use crate::peniko::{ColorStop, ColorStops, Extend, Gradient, GradientKind, ImageQuality};
 use alloc::borrow::Cow;
 use alloc::fmt::Debug;
 use alloc::vec;
@@ -235,6 +235,9 @@ impl EncodeExt for Gradient {
             x_advance,
             y_advance,
             ranges,
+            stops: ColorStops(stops.into_owned()),
+            interpolation_cs: self.interpolation_cs,
+            hue_direction: self.hue_direction,
             pad,
             has_opacities,
             u8_lut: OnceCell::new(),
@@ -709,6 +712,12 @@ pub struct EncodedGradient {
     pub y_advance: Vec2,
     /// The color ranges of the gradient.
     pub ranges: Vec<GradientRange>,
+    /// The original color stops.
+    pub stops: ColorStops,
+    /// The color space used for interpolation.
+    pub interpolation_cs: ColorSpaceTag,
+    /// The hue direction for cylindrical color space interpolation.
+    pub hue_direction: HueDirection,
     /// Whether the gradient should be padded.
     pub pad: bool,
     /// Whether the gradient requires `source_over` compositing.
--- a/sparse_strips/vello_hybrid/Cargo.toml
+++ b/sparse_strips/vello_hybrid/Cargo.toml
@@ -22,6 +22,7 @@ wgpu = { workspace = true, optional = true }
 vello_sparse_shaders = { workspace = true, optional = true }
 log = { workspace = true }
 guillotiere = "0.6.2"
+hashbrown = "0.15"
 
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
@@ -51,5 +52,6 @@ roxmltree = "0.20.0"
 
 [features]
 default = ["wgpu"]
+std = ["wgpu", "vello_common/std"]
 wgpu = ["dep:wgpu", "dep:vello_sparse_shaders"]
 webgl = ["dep:js-sys", "dep:web-sys", "dep:vello_sparse_shaders", "vello_sparse_shaders/glsl"]
--- a/sparse_strips/vello_hybrid/examples/scenes/src/gradient.rs
+++ b/sparse_strips/vello_hybrid/examples/scenes/src/gradient.rs
@@ -1,0 +1,337 @@
+// Copyright 2025 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Gradient rendering example scenes.
+//! Scenes demonstrating gradient rendering with different extend modes. Taken from Vello Classic
+//! test scenes:
+//! - `GradientExtendScene`:
+//!     - `gradient_extend` method from `https://github.com/linebender/vello/blob/0f3ef03a823eb10b0d7a60164e286cde77ffa222/examples/scenes/src/test_scenes.rs#L815`
+//! - `RadialScene`:
+//!     - `two_point_radial` method from `https://github.com/linebender/vello/blob/0f3ef03a823eb10b0d7a60164e286cde77ffa222/examples/scenes/src/test_scenes.rs#L882`
+
+use crate::ExampleScene;
+use smallvec::smallvec;
+use vello_common::color::palette::css::{BLACK, BLUE, LIME, RED, WHITE, YELLOW};
+use vello_common::kurbo::{Affine, Ellipse, Point, Rect, Shape, Stroke};
+use vello_common::peniko::{
+    Color, ColorStop, ColorStops, Extend, Gradient, GradientKind, color::DynamicColor,
+};
+use vello_hybrid::Scene;
+
+/// Gradient scene state
+#[derive(Debug, Default)]
+pub struct GradientExtendScene {}
+
+impl GradientExtendScene {
+    /// Create a new gradient extend scene
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl ExampleScene for GradientExtendScene {
+    fn render(&mut self, scene: &mut Scene, root_transform: Affine) {
+        enum Kind {
+            Linear,
+            Radial,
+            Sweep,
+        }
+
+        /// Helper function to create color stops
+        fn create_color_stops(colors: &[Color]) -> ColorStops {
+            ColorStops(smallvec![
+                ColorStop {
+                    offset: 0.0,
+                    color: DynamicColor::from_alpha_color(colors[0]),
+                },
+                ColorStop {
+                    offset: 0.5,
+                    color: DynamicColor::from_alpha_color(colors[1]),
+                },
+                ColorStop {
+                    offset: 1.0,
+                    color: DynamicColor::from_alpha_color(colors[2]),
+                },
+            ])
+        }
+
+        /// Helper function to create a square with a specific gradient type and extend mode
+        fn square(scene: &mut Scene, kind: Kind, transform: Affine, extend: Extend) {
+            let colors = [RED, LIME, BLUE];
+            let width = 300.0;
+            let height = 300.0;
+
+            let gradient = match kind {
+                Kind::Linear => {
+                    let start_x = width * 0.35;
+                    let start_y = height * 0.5;
+                    let end_x = width * 0.65;
+                    let end_y = height * 0.5;
+
+                    Gradient {
+                        kind: GradientKind::Linear {
+                            start: Point::new(start_x, start_y),
+                            end: Point::new(end_x, end_y),
+                        },
+                        stops: create_color_stops(&colors),
+                        extend,
+                        ..Default::default()
+                    }
+                }
+                Kind::Radial => {
+                    let center_x = width * 0.5;
+                    let center_y = height * 0.5;
+                    #[allow(
+                        clippy::cast_possible_truncation,
+                        reason = "Width is always positive and bounded"
+                    )]
+                    let radius = (width * 0.25) as f32;
+
+                    Gradient {
+                        kind: GradientKind::Radial {
+                            start_center: Point::new(center_x, center_y),
+                            start_radius: radius * 0.25,
+                            end_center: Point::new(center_x, center_y),
+                            end_radius: radius,
+                        },
+                        stops: create_color_stops(&colors),
+                        extend,
+                        ..Default::default()
+                    }
+                }
+                Kind::Sweep => {
+                    let center_x = width * 0.5;
+                    let center_y = height * 0.5;
+
+                    Gradient {
+                        kind: GradientKind::Sweep {
+                            center: Point::new(center_x, center_y),
+                            start_angle: 30.0,
+                            end_angle: 150.0,
+                        },
+                        stops: create_color_stops(&colors),
+                        extend,
+                        ..Default::default()
+                    }
+                }
+            };
+
+            scene.set_transform(transform);
+            scene.set_paint(gradient);
+            scene.fill_rect(&Rect::new(0.0, 0.0, width, height));
+        }
+
+        let extend_modes = [Extend::Pad, Extend::Repeat, Extend::Reflect];
+        for (x, extend) in extend_modes.iter().enumerate() {
+            for (y, kind) in [Kind::Linear, Kind::Radial, Kind::Sweep]
+                .into_iter()
+                .enumerate()
+            {
+                let transform = root_transform
+                    * Affine::translate((x as f64 * 350.0 + 50.0, y as f64 * 350.0 + 100.0));
+                square(scene, kind, transform, *extend);
+            }
+        }
+    }
+}
+
+/// Two-point radial gradient scene
+#[derive(Debug, Default)]
+pub struct RadialScene;
+
+impl RadialScene {
+    /// Create a new radial scene
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl ExampleScene for RadialScene {
+    fn render(&mut self, scene: &mut Scene, root_transform: Affine) {
+        /// Helper function to create color stops
+        fn create_color_stops(colors: &[Color]) -> ColorStops {
+            ColorStops(smallvec![
+                ColorStop {
+                    offset: 0.0,
+                    color: DynamicColor::from_alpha_color(colors[0]),
+                },
+                ColorStop {
+                    offset: 0.5,
+                    color: DynamicColor::from_alpha_color(colors[1]),
+                },
+                ColorStop {
+                    offset: 1.0,
+                    color: DynamicColor::from_alpha_color(colors[2]),
+                },
+            ])
+        }
+
+        /// Helper function to create a two-point radial gradient rectangle
+        fn make(
+            scene: &mut Scene,
+            x0: f64,
+            y0: f64,
+            r0: f32,
+            x1: f64,
+            y1: f64,
+            r1: f32,
+            transform: Affine,
+            extend: Extend,
+        ) {
+            let colors = [RED, YELLOW, Color::from_rgb8(6, 85, 186)];
+            let width = 400.0;
+            let height = 200.0;
+
+            scene.set_transform(transform);
+            scene.set_paint(WHITE);
+            scene.fill_rect(&Rect::new(0.0, 0.0, width, height));
+
+            let gradient = Gradient {
+                kind: GradientKind::Radial {
+                    start_center: Point::new(x0, y0),
+                    start_radius: r0,
+                    end_center: Point::new(x1, y1),
+                    end_radius: r1,
+                },
+                stops: create_color_stops(&colors),
+                extend,
+                ..Default::default()
+            };
+
+            scene.set_paint(gradient);
+            scene.fill_rect(&Rect::new(0.0, 0.0, width, height));
+
+            // Draw stroke circles showing the gradient extents
+            let r0 = r0 as f64 - 1.0;
+            let r1 = r1 as f64 - 1.0;
+            scene.set_paint(BLACK);
+            scene.set_stroke(Stroke::new(1.0));
+            scene.stroke_path(&Ellipse::new((x0, y0), (r0, r0), 0.0).to_path(0.1));
+            scene.stroke_path(&Ellipse::new((x1, y1), (r1, r1), 0.0).to_path(0.1));
+        }
+
+        // These demonstrate radial gradient patterns similar to the examples shown
+        // at <https://learn.microsoft.com/en-us/typography/opentype/spec/colr#radial-gradients>
+
+        // Row 1: Basic two-point radial gradient
+        for (i, mode) in [Extend::Pad, Extend::Repeat, Extend::Reflect]
+            .iter()
+            .enumerate()
+        {
+            let y = 100.0;
+            let x0 = 140.0;
+            let x1 = x0 + 140.0;
+            let r0 = 20.0;
+            let r1 = 50.0;
+            make(
+                scene,
+                x0,
+                y,
+                r0,
+                x1,
+                y,
+                r1,
+                root_transform * Affine::translate((i as f64 * 420.0 + 20.0, 20.0)),
+                *mode,
+            );
+        }
+
+        // Row 2: Reversed two-point radial gradient
+        for (i, mode) in [Extend::Pad, Extend::Repeat, Extend::Reflect]
+            .iter()
+            .enumerate()
+        {
+            let y = 100.0;
+            let x0: f64 = 140.0;
+            let x1 = x0 + 140.0;
+            let r0 = 20.0;
+            let r1 = 50.0;
+            make(
+                scene,
+                x1,
+                y,
+                r1,
+                x0,
+                y,
+                r0,
+                root_transform * Affine::translate((i as f64 * 420.0 + 20.0, 240.0)),
+                *mode,
+            );
+        }
+
+        // Row 3: Equal radii gradient
+        for (i, mode) in [Extend::Pad, Extend::Repeat, Extend::Reflect]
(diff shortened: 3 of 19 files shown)


sparse_strips/vello_hybrid/examples/scenes/src/gradient.rs
grebmeg, line 1: Hey @LaurenzV, while working on gradient rendering I noticed a couple of differences between Vello Classic and Vello CPU/Hybrid. Do you happen to know more about these?
- I haven’t dug deeply into this yet, but I think one of the renderers might need to swap the Y-direction?

Vello Classic|Vello CPU/Hybrid
---|---
image|image


- I also found an issue with how we calculate the reflect extend — if I recall correctly, `r1` [here](https://github.com/linebender/vello/blob/0f3ef03a823eb10b0d7a60164e286cde77ffa222/sparse_strips/vello_common/src/encode.rs#L124-L128) ends up being -10 in this case:

Vello Classic|Vello CPU/Hybrid
---|---
image|image

LaurenzV: > I haven’t dug deeply into this yet, but I think one of the renderers might need to swap the Y-direction?

Yeah that's right, we measure the angle in counter-clockwise direction while Vello Classic seems to measure it in the clockwise direction. I don't think that either renderer is right or wrong, it's simply a matter of convention and you can convert between them by applying an approopriate transform.

> I also found an issue with how we calculate the reflect extend — if I recall correctly, r1 [here](https://github.com/linebender/vello/blob/0f3ef03a823eb10b0d7a60164e286cde77ffa222/sparse_strips/vello_common/src/encode.rs#L124-L128) ends up being -10 in this case:

Hmmm, I see... if I recall correctly negative radii are not properly supported, so it makes sense that this might cause issues. I haven't thought about it too deeply yet, but if this approach does cause problems we might have to change the approach and to the extend calculation during fine rasterization instead.

DJMcNab: > Yeah that's right, we measure the angle in counter-clockwise direction while Vello Classic seems to measure it in the clockwise direction.

Presumably this is based on https://docs.rs/peniko/latest/peniko/enum.GradientKind.html#variant.Sweep?
So it *looks* like Vello Classic is probably wrong here.

LaurenzV: Ah, I didn't notice that peniko defines this!

sparse_strips/vello_common/src/encode.rs
```diff
@@ -709,6 +712,12 @@ pub struct EncodedGradient {
     pub y_advance: Vec2,
     /// The color ranges of the gradient.
     pub ranges: Vec<GradientRange>,
+    /// The original color stops.
+    pub stops: ColorStops,
+    /// The color space used for interpolation.
+    pub interpolation_cs: ColorSpaceTag,
+    /// The hue direction for cylindrical color space interpolation.
+    pub hue_direction: HueDirection,
```
LaurenzV: Do we really need to store those for the cache key? The `ranges` field already contains the ranges that have been approximated in case a different color space was used, so shouldn't this be enough?

sparse_strips/vello_hybrid/src/gradient_cache.rs
LaurenzV, line 1: I imagine this would be useful in `vello_common`? But from my side fine to keep it here for now.

sparse_strips/vello_sparse_tests/tests/mix.rs
```diff
@@ -65,12 +69,12 @@ fn mix(ctx: &mut impl Renderer, blend_mode: BlendMode) {
     ctx.pop_layer();
 }
 
-#[vello_test]
```
LaurenzV, line 68: I imagine those are necessary because vello_hybrid uses the u8_lut instead of f32_lut by default? But the speed boost you get from u8 is probably worth more than a bit more accuracy, not sure. 😄

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    epoch: u64,
+    /// Cache mapping gradient signature to cached ramps and last access time.
+    cache: HashMap<CacheKey<GradientRampKey>, (CachedRamp, u64)>,
+    /// Packed gradient luts.
+    luts: Vec<u8>,
+    /// Whether the packed luts needs to be re-uploaded.
+    has_changed: bool,
+}
+
+// TODO: Rebuild the packed luts or allow to rewrite evicted gradient ramps after eviction.
```
ajakubowicz-canva: > allow to rewrite evicted gradient ramps after eviction

Nit: I'm not sure I understand this TODO?

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub(crate) fn maintain(&mut self) {
+        self.epoch += 1;
+
+        if self.cache.len() > RETAINED_COUNT {
+            self.evict_old_entries();
+        }
+    }
+
+    /// Evict old cache entries using LRU policy.
+    fn evict_old_entries(&mut self) {
```
ajakubowicz-canva: I think there's an issue here. This isn't exclusively a LRU cache. This cache also evicts anything not used in the last 3 epochs. But, we increment epoch on any access or creation of gradients. Thus, this essentially caps the LRU to only ever containing 3 gradients when `evict_old_entries` is called.

If I understand the code correctly, the following would happen. Assume that `RETAINED_COUNT` is `5`.

1. get_or_create gradient A. (epoch = 1)
2. get_or_create gradient B. (epoch = 2)
3. get_or_create gradient C. (epoch = 3)
4. get_or_create gradient D. (epoch = 4)
5. get_or_create gradient E. (epoch = 5) // self.cache.len() >= RETAINED_COUNT triggers evict_old_entries

Gradient A and B are evicted leaving `C`, `D`, and `E`. This seems excessive. With `RETAINED_COUNT` at 64 then the cache will grow to 64 entries and then be flushed down to only 3 entries.

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Cache mapping gradient signature to cached ramps and last access time.
+    cache: HashMap<CacheKey<GradientRampKey>, (CachedRamp, u64)>,
+    /// Packed gradient luts.
+    luts: Vec<u8>,
+    /// Whether the packed luts needs to be re-uploaded.
+    has_changed: bool,
+}
+
+// TODO: Rebuild the packed luts or allow to rewrite evicted gradient ramps after eviction.
+impl GradientRampCache {
```
ajakubowicz-canva, line 54: Is it valuable to add a module scoped test for `GradientRampCache` to enforce the LRU behavior?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -211,41 +246,145 @@ impl Renderer {
...
 
-    fn prepare_gpu_encoded_paints(&self, encoded_paints: &[EncodedPaint]) -> Vec<GpuEncodedImage> {
-        let mut bytes: Vec<GpuEncodedImage> = Vec::new();
-        for paint in encoded_paints {
+    fn prepare_gpu_encoded_paints(&mut self, encoded_paints: &[EncodedPaint]) {
+        self.encoded_paints
+            .resize_with(encoded_paints.len(), || GPU_PAINT_PLACEHOLDER);
+        self.paint_idxs.resize(encoded_paints.len() + 1, 0);
+
+        self.gradient_cache.maintain();
```
ajakubowicz-canva: Is the `maintain` method valuable if it only does work if that cache is greater than the `RETAINED_COUNT`?

Removing the `maintain` method would still allow LRU behavior to be maintained on the insertion of new luts which itself is also gated on `RETAINED_COUNT >= current cache size`.

This has the additional benefit of only incurring gradient maintenance LRU cache cost on gradient insertion.

sparse_strips/vello_hybrid/src/gradient_cache.rs
ajakubowicz-canva: Ah - I see. Currently the `luts` vector grows unbounded, effectively a memory leak.

I think this needs to be resolved before this PR can land? Maybe by maintaining a free_slots list for recycling blocks, or by occasionally rebuilding luts entirely.

sparse_strips/vello_common/src/encode.rs
grebmeg: They’re all essentially used as cache keys for gradient ramp cache, and the most optimal data to store appeared to be the `stops` plus two discriminants, rather than the combination of `range x1: 32`, `bias: [f32; 4]`, and `[f32; 4]` without any color information. Since we already have a hash implementation for `ColorStops` in `peniko`, I thought it would be more compact and efficient to use those three instead.

sparse_strips/vello_hybrid/examples/scenes/src/gradient.rs
grebmeg: > I don't think that either renderer is right or wrong, it's simply a matter of convention and you can convert between them by applying an approopriate transform.

I agree, but from a consumer’s perspective it would be very helpful to minimize cases where conversion is required when switching to another renderer. Since peniko already defines this, do we just need to update our angle direction then?

> Hmmm, I see... if I recall correctly negative radii are not properly supported, so it makes sense that this might cause issues. I haven't thought about it too deeply yet, but if this approach does cause problems we might have to change the approach and to the extend calculation during fine rasterization instead.

I tried updating the calculation for Reflect, but it got a bit messy and led to more complex formulas 😅, though I still feel it should be possible. Could you remind me why we initially chose this approach instead of using extend in fine?

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: Yeah, that’s possible, and we can always move it if needed.

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: Oh, sorry, the comment wasn’t very clear. What I meant is that we currently extend luts in one direction, with a hard limit of 64. The TODO is supposed to indicate that we should be reusing/recycling blocks. After reading your comment, I also realized it’s currently quite easy to reach a state where all luts are occupied and no new ones can be added. I’ll update this accordingly to address both your comment and my TODO.

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[allow(
+        clippy::cast_possible_truncation,
+        reason = "Conversion from usize to u32 is safe, used for texture coordinates"
+    )]
+    pub(crate) fn get_or_create_ramp(&mut self, gradient: &EncodedGradient) -> (u32, u32) {
+        self.epoch += 1;
+
+        // Create cache key from gradient color properties.
+        let cache_key = CacheKey(GradientRampKey {
+            stops: gradient.stops.clone(),
```
taj-p: This is an allocation that is only useful if `cache_key` isn't already in the cache.

In Parley, we have a similar problem where most of the time, the `cache_key` is merely used to "get" the value rather than for its allocation to become owned as a new key in `self.cache`.

When "getting or creating" a value, we pass a borrowed key `K` that implements `Equivalent<ID>` and `Into<ID>` where `ID` is the owned type stored within the cache. This means that we can use borrowed data for lookup and, only when required, `into` it into an owned type (potentially paying the allocation). 

https://github.com/linebender/parley/blob/b250bb19e1d5d93d611792d732d0cb1512472a26/parley/src/lru_cache.rs#L37-L40

https://github.com/linebender/parley/blob/b250bb19e1d5d93d611792d732d0cb1512472a26/parley/src/shape/cache.rs#L39-L93

I suspect that's a pattern we'd like to extend to caches in Vello too.

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        // Check if we already have this gradient cached.
+        if let Some((cached_ramp, last_used)) = self.cache.get_mut(&cache_key) {
+            *last_used = self.epoch;
+            return (cached_ramp.lut_start, cached_ramp.width);
+        }
+
+        // Generate new gradient LUT.
+        let lut_start = (self.luts.len() / 4) as u32;
+        generate_gradient_lut_dispatch(Level::new(), gradient, &mut self.luts);
```
taj-p: Vello CPU defines a Level that is used throughout all its dependencies [here](https://github.com/taj-p/vello/blob/58b80d660e2fc5aef3bd32b24af3f95e973aab95/sparse_strips/vello_cpu/src/render.rs#L65). I think we want to do the same. Although, in hybrid we do not yet allow user-defined `Level`, I think we still want to pass down the one [defined on creation](https://github.com/taj-p/vello/blob/58b80d660e2fc5aef3bd32b24af3f95e973aab95/sparse_strips/vello_hybrid/src/scene.rs#L76) to make later migration easier.

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// Generate the gradient LUT.
+#[inline(always)]
+fn generate_gradient_lut_impl<S: vello_common::fearless_simd::Simd>(
+    simd: S,
+    gradient: &vello_common::encode::EncodedGradient,
+    output: &mut Vec<u8>,
+) {
+    let lut = gradient.u8_lut(simd);
+    for color in lut.lut() {
```
taj-p: Could add a TODO here that `lut()` is allocating the entire gradient LUT within the `OnceCell` only for us to copy it to `output` and immediately free. If `lut()` returned an iterator, we could avoid the duplicate allocation in `lut` and `output`

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1020,6 +1165,173 @@ impl Programs {
...
+
+    /// Upload gradient data to the texture.
+    fn upload_gradient_texture(&mut self, queue: &Queue, gradient_cache: &GradientRampCache) {
+        let gradient_texture = &self.resources.gradient_texture;
+        let gradient_texture_width = gradient_texture.width();
+        let gradient_texture_height = gradient_texture.height();
+
+        // Copy the flat gradient LUT data
+        if !gradient_cache.is_empty() {
+            let total_capacity = (gradient_texture_width * gradient_texture_height * 4) as usize;
```
taj-p, line 1313: Do we need `gradient_data`? It seems we're allocating the additional memory and paying the copy cost only to allow for resizing the staging buffer. Could we instead do something like:


```rs
            let total_capacity = (gradient_texture_width * gradient_texture_height * 4) as usize;

            let mut luts = gradient_cache.take_luts();
            let old_luts_len = luts.len();
            luts.resize(total_capacity, 0);

            queue.write_texture(
                wgpu::TexelCopyTextureInfo {
                    texture: gradient_texture,
                    mip_level: 0,
                    origin: wgpu::Origin3d::ZERO,
                    aspect: wgpu::TextureAspect::All,
                },
                &luts,
                wgpu::TexelCopyBufferLayout {
                    offset: 0,
                    // 4 bytes per RGBA8 pixel
                    bytes_per_row: Some(gradient_texture_width << 2),
                    rows_per_image: Some(gradient_texture_height),
                },
                wgpu::Extent3d {
                    width: gradient_texture_width,
                    height: gradient_texture_height,
                    depth_or_array_layers: 1,
                },
            );

            luts.truncate(old_luts_len);
            gradient_cache.restore_luts(luts);
```

sparse_strips/vello_hybrid/examples/scenes/src/gradient.rs
LaurenzV: > Since peniko already defines this, do we just need to update our angle direction then?

My understanding is that Vello classic is the one that needs updating, not ours.

> I tried updating the calculation for Reflect, but it got a bit messy and led to more complex formulas 😅, though I still feel it should be possible. Could you remind me why we initially chose this approach instead of using extend in fine?

No particular reason, it's just that the formula for extending with reflect is more complex, so I felt like it would be easier to instead only calculate this once during encoding to save calculations on the pixel-level. Maybe we can just open an issue for it? I don't believe that this is blocking anything, right? Seems like an edge case to me.

sparse_strips/vello_common/src/encode.rs
LaurenzV: I'll leave it up to you, I'm just a bit concerned about storing more information than strictly necessary in the already pretty big struct, as the paints will be stored long-term in a vector.

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+        output.extend_from_slice(&[color[0], color[1], color[2], color[3]]);
+    }
```
taj-p: instead of copying one slice at a time to `output` (with the potential of repeated grows on `output`) we can perform 1 large `memmove` via:

```suggestion
    let lut = gradient.u8_lut(simd);
    let bytes: &[u8] = bytemuck::cast_slice(lut.lut());
    output.reserve(bytes.len());
    output.extend_from_slice(bytes);
```

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+        });
+
+        // Check if we already have this gradient cached.
+        if let Some((cached_ramp, last_used)) = self.cache.get_mut(&cache_key) {
+            *last_used = self.epoch;
+            return (cached_ramp.lut_start, cached_ramp.width);
+        }
+
+        // Generate new gradient LUT.
+        let lut_start = (self.luts.len() / 4) as u32;
```
taj-p: Could you please add a comment here on why it's div by 4

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -681,16 +898,52 @@ impl WebGlPrograms {
...
+    ) {
+        if gradient_cache.is_empty() {
+            return;
         }
+
+        let gradient_texture_width = self.resources.max_texture_dimension_2d;
+        let gradient_texture_height = self.resources.gradient_texture_height;
+        let gradient_luts = gradient_cache.luts();
+
+        self.gradient_data[0..gradient_luts.len()].copy_from_slice(gradient_luts);
```
taj-p: Per the WGPU comment, could we reuse the allocation within `gradient_luts` somehow?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -52,26 +67,179 @@ pub struct GpuStrip {
...
+    pub(crate) fn serialize_to_buffer(paints: &[Self], buffer: &mut [u8]) -> usize {
+        let mut offset = 0;
+        for paint in paints {
+            let paint_bytes = paint.as_bytes();
+            let end_offset = offset + paint_bytes.len();
+            buffer[offset..end_offset].copy_from_slice(paint_bytes);
+            offset = end_offset;
+        }
+        offset
+    }
```
taj-p:
```suggestion
    /// Serialize paint enums directly into the provided buffer.
    pub(crate) fn serialize_to_buffer(paints: &[Self], buffer: &mut Vec<u8>) {
        buffer.clear();
        for paint in paints {
            buffer.extend_from_slice(paint.as_bytes());
        }
    }
```

I don't think any consumer reads the number of bytes written. WDYT about taking a vec and simplifying the loop block?

taj-p (commented):
Changes look really good. I've still got `render/common.rs` and the shader to review - will do that tomorrow morning

sparse_strips/vello_hybrid/src/gradient_cache.rs
taj-p: edit: On second look, it's backed by a `SmallVec`, so it should be really cheap to clone most of the time. At the minimum, would you be able to leave a TODO?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -33,14 +33,36 @@ const COLOR_SOURCE_BLEND: u32 = 2u;
...
+// Radial gradient types.
+const RADIAL_GRADIENT_TYPE_STANDARD: u32 = 0u;
+const RADIAL_GRADIENT_TYPE_STRIP: u32 = 1u;
+const RADIAL_GRADIENT_TYPE_FOCAL: u32 = 2u;
+
+// Mathematical constants.
+const PI: f32 = 3.1415926535897932384626433832795028;
+const TWO_PI: f32 = 2.0 * PI;
+// Tolerance for nearly zero comparisons (matching vello_cpu implementation).
+const NEARLY_ZERO_TOLERANCE: f32 = 1.0 / 4096.0;
```
taj-p, line 65: Are you able to add a comment [here](https://github.com/taj-p/vello/blob/58b80d660e2fc5aef3bd32b24af3f95e973aab95/sparse_strips/vello_common/src/math.rs#L21) stating that changing that value should also see a change in `render_strips.wgsl`?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1020,6 +1165,173 @@ impl Programs {
...
+        let texture_width = self.resources.alphas_texture.width();
+        let texture_height = self.resources.alphas_texture.height();
+        debug_assert!(
+            alphas.len() <= (texture_width * texture_height * 16) as usize,
+            "Alpha texture dimensions are too small to fit the alpha data"
+        );
+
+        // After this copy to `self.alpha_data`, there may be stale trailing alpha values. These
+        // are not sampled, so can be left as-is.
+        self.alpha_data[0..alphas.len()].copy_from_slice(alphas);
```
taj-p, line 1252: Happy to leave this one as a TODO since it's not directly attributable to this PR, but we also don't necessarily need to copy all the alphas into a staging buffer. We can use `alphas` directly (similar to https://github.com/linebender/vello/pull/1196/files#r2335625698) by taking the alphas from the strip generator as a vec, resizing it appropriately, truncating, and restoring it back to the generator.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -600,20 +810,23 @@ impl WebGlPrograms {
...
             .unwrap();
 
-            self.render_size = render_size.clone();
+            self.render_size = new_render_size.clone();
         }
+    }
 
-        // Process alpha data for texture
+    /// Upload alpha data to the texture.
+    fn upload_alpha_texture(&mut self, gl: &WebGl2RenderingContext, alphas: &[u8]) {
```
taj-p, line 826: Similar comment to the WGPU comment: We don't necessarily need the `self.alpha_data` staging buffer + redundant copy (but happy to address this comment by adding a `TODO` to the code).

taj-p (commented):
Looks great! I think there is some rework required in the cache to allow for more than 3 values? Happy to approve after that and other comments addressed

sparse_strips/vello_common/src/encode.rs
grebmeg: Hey @LaurenzV, what do you think about calculating the gradient cache key directly during encoding? That way we can avoid passing the data around and eliminate redundant calculations in each `get_or_create_ramp` call in the cache.

LaurenzV: Sounds fine to me!

sparse_strips/vello_hybrid/examples/scenes/src/gradient.rs
grebmeg: Yeah, that makes sense. And it’s not blocking. I’ve opened this issue #1212 to track the reflect extend problem.

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: 100%, I'll add tests.

Fixed

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: I moved `cache_key` calculation into gradient encoding step, which avoids passing extra data in the `EncodedGradient` struct, I think that makes sense ([this](https://github.com/linebender/vello/pull/1196/files#r2331537454) is the discussion about that). Now we only need to clone when inserting. How does that sound now?

sparse_strips/vello_common/src/encode.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: Yeah, that makes sense!

Fixed

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: Totally agree. Unfortunately, after my first draft, it slipped through and I lost track of it. I’ve now adjusted it to behave more like an LRU cache, though I avoided adding an extra queue to keep things simpler and not over-optimize. I also included a calculation for the worst-case scenario (4096 colors in a ramp) to get how much we can fit into a gradient texture and which is the gradients ramp cache size. Once the cache reaches that limit, the LRU entry is evicted, which comes with some overhead costs, draining and updating the LUT start offsets, both of which are worst O(N).

Fixed

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: I agree, that would be better.

Added TODO.

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: This's awesome! Thanks Taj!

Fixed.

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Hmm, I don’t think we can do it, since in our case the buffer is assumed to be `encoded_paints_data`, which should already be preallocated with the necessary size for writing to the texture. We could follow your suggestion, but then we’d need to extend it to the required size, so for simplicity, I’d stick with the current code (I’ll also remove the return type since it isn’t used anyway).

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: I agree with you. This idea originally came from `ramp_cache` in Vello Classic, and I had the same question as you. I’ve now removed it, so eviction is handled only on insertion.

Fixed.

sparse_strips/vello_hybrid/src/gradient_cache.rs
taj-p: edit: On third look (so sorry), looks like you could simply [implement `Equivalent`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.get) for the borrowed type for the same effect it seems. But, I've read elsewhere you might move key construction upstream, so this might be OBE

sparse_strips/vello_sparse_tests/tests/mix.rs
grebmeg: Actually, the tests in the mix test suite actually require setting `hybrid_tolerance` because of the cowboy image used there. 😅 If I remove that image and rerun the tests, everything passes. So the issue lies in image rendering in hybrid, which i'd like to try addressing a bit later.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Sure thing!

Fixed

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Thank you! This is a very good optimization suggestion. It addresses a real inefficiency in the current implementation.

Fixed.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Yep, good idea. I’ll leave it as a TODO since it’s an overall improvement compared to the staging pattern we used before, and we can address it later in follow-up PRs.

Added TODO.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Similar to `upload_alpha_texture` in wgpu — left as a TODO.

Added TODO.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Addressed the same way as in `upload_gradient_texture` in wgpu.

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,418 @@
...
+        // Generate new gradient LUT.
+        let lut_start = self.luts.len() as u32 / BYTES_PER_TEXEL;
+        generate_gradient_lut_dispatch(self.level, gradient, &mut self.luts);
+        let lut_end = self.luts.len() as u32 / BYTES_PER_TEXEL;
+        let width = lut_end - lut_start;
+        let cached_ramp = CachedRamp { width, lut_start };
+        self.has_changed = true;
+        self.cache
+            .insert(gradient.cache_key.clone(), (cached_ramp, self.epoch));
+        self.maybe_evict_lru_entry();
```
taj-p: Couldn't this evict a gradient from the current frame? Also, evicting an entry and mutating `lut_start`s for gradients that were stored after the evicted entry would invalidate previously returned gradient `lut_start`s, I think? 

I think we might need to use a `maintain` approach to allow accumulating gradients within a single render frame like we were previously

taj-p: In other words, a test like this should pass IIUC. In the second frame, the `lut_start_2` is being mutated due to the eviction of `lut_start_1` when getting `lut_start_3`

```rs
    #[test]
    fn test_lut_start_invalidation() {
        let mut cache = GradientRampCache::new(2, Level::fallback());
        
        let gradient_1 = create_encoded_gradient(create_gradient(0.1));
        let gradient_2 = create_encoded_gradient(create_gradient(0.2));
        let gradient_3 = create_encoded_gradient(create_gradient(0.3));

        // Frame 1 start
        let lut_start_1 = cache.get_or_create_ramp(&gradient_1).0;
        let _lut_start_2 = cache.get_or_create_ramp(&gradient_2).0;
        // Frame 1 end

        // Frame 2 start
        let lut_start_2 = cache.get_or_create_ramp(&gradient_2).0;
        let lut_start_3 = cache.get_or_create_ramp(&gradient_3).0;
        // Frame 2 end

        // Gradient lut shouldn't mutate within a frame
        assert_eq!(lut_start_2, cache.cache.get(&gradient_2.cache_key).unwrap().0.lut_start);
        assert_eq!(lut_start_3, cache.cache.get(&gradient_3.cache_key).unwrap().0.lut_start);
    }
```

sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -22,6 +22,7 @@ wgpu = { workspace = true, optional = true }
 vello_sparse_shaders = { workspace = true, optional = true }
 log = { workspace = true }
 guillotiere = "0.6.2"
+hashbrown = "0.15"
```
taj-p, line 25: Should we be using "0.16"? Or does using "0.15" allow for dependency de-duplication somehow?

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: Oh yeah, it’s actually a major issue. I reverted to maintain with some changes to `gradient_cache`.

Fixed.

sparse_strips/vello_hybrid/Cargo.toml
grebmeg: I thought it did, since many other crates have dependency on "0.15.”

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -211,41 +259,145 @@ impl Renderer {
...
 
-    fn prepare_gpu_encoded_paints(&self, encoded_paints: &[EncodedPaint]) -> Vec<GpuEncodedImage> {
-        let mut bytes: Vec<GpuEncodedImage> = Vec::new();
-        for paint in encoded_paints {
+    fn prepare_gpu_encoded_paints(&mut self, encoded_paints: &[EncodedPaint]) {
+        self.encoded_paints
+            .resize_with(encoded_paints.len(), || GPU_PAINT_PLACEHOLDER);
+        self.paint_idxs.resize(encoded_paints.len() + 1, 0);
+
+        self.gradient_cache.maintain();
```
taj-p: It's likely better to run eviction after `do_scene` within `render` so that cache invalidation processing doesn't block the issuing and, thus, running of GPU commands (which can run in parallel to cache eviction if we re-order the operations).

sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -0,0 +1,515 @@
...
+        let cached_ramp = CachedRamp { width, lut_start };
+        self.has_changed = true;
+        self.cache
+            .insert(gradient.cache_key.clone(), (cached_ramp, self.epoch));
+
+        (lut_start, width)
+    }
+
+    /// Maintain the gradient cache by evicting old entries.
+    pub(crate) fn maintain(&mut self) {
```
taj-p, line 96: I think we can endeavor to reduce the allocations and improve this function by collecting less, using iterators, and using scratch space. I suspect something [like this](https://github.com/linebender/vello/compare/gemberg/gradients-rendering...taj-p:vello:tajp/gradient-cache-suggestions?expand=1) will reduce the latency cost of this function significantly

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: That make sense!

Fixed

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg: That approach with scratch space is really useful! At first, I was trying to make the hot path is fast as possible with less allocations a possible, in `maintain` I assumed that worst scenario we'd run quite rare, which is actually not true if we want to use smaller size cache. Anyway, thank you for the suggestion; It’s definitely worth incorporating!

Another note: since we compact LUTs and write data into the texture, a large cache can actually increase texture write latency, especially when the cache holds a lot of entries that are no longer needed.

PR #1201, opened 2025-09-08
Add skipping the GPU tests to some tests

DJMcNab:
This should fix CI on Windows.

See [#vello > Windows-Latest CI failures](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Windows-Latest.20CI.20failures/with/538194315)

### DIFF
--- a/vello_tests/tests/known_issues.rs
+++ b/vello_tests/tests/known_issues.rs
@@ -113,7 +113,7 @@ const DATA_IMAGE_PNG: &[u8] = include_bytes!("../snapshots/smoke/data_image_roun
 
 /// Test for <https://github.com/linebender/vello/issues/972>
 #[test]
-#[ignore = "CI runs these tests on a CPU, leading to them having unrealistic precision"]
+#[ignore = "CI runs these tests on a CPU, leading to them having unrealistic precision"] // Uncomment below line when removing this.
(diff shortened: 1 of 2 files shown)


vello_tests/tests/known_issues.rs
```diff
@@ -113,6 +114,7 @@ const DATA_IMAGE_PNG: &[u8] = include_bytes!("../snapshots/smoke/data_image_roun
 /// Test for <https://github.com/linebender/vello/issues/972>
 #[test]
 #[ignore = "CI runs these tests on a CPU, leading to them having unrealistic precision"]
+// #[cfg_attr(skip_gpu_tests, ignore)]
```
sagudev, line 117: Why is this commented out?

sagudev (approved):
Approving, because it unblocks CI.Sorry for all trouble caused.

vello_tests/tests/known_issues.rs
DJMcNab: Because it's an unused attribute; the `ignore` above always applies, so we'd get a warning if we uncommented this.

As it happens, I don't think that warning would block CI, but that's not particularly intentional.

sagudev: I see.

DJMcNab: I've added some guidance

PR #1202, opened 2025-09-08
Try to fix clip optimization

dfrg:


vello_shaders/shader/fine.wgsl
```diff
@@ -987,18 +988,23 @@ fn main(
...
+                        rgba[i] *= rgba_mul;
+                    }                        
                 } else {
                     let blend_in_scratch = clip_depth - BLEND_STACK_SPLIT;
                     let local_tile_ix = local_id.x * PIXELS_PER_THREAD + local_id.y * TILE_WIDTH;
                     let local_blend_start = blend_offset + blend_in_scratch * TILE_WIDTH * TILE_HEIGHT + local_tile_ix;
                     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
                         blend_spill[local_blend_start + i] = pack4x8unorm(rgba[i]);
-                        rgba[i] = vec4(0.0);
+                        rgba[i] *= rgba_mul;
```
DJMcNab, line 1007: Obviously this is assuming that doing a floating point multiply is extremely cheap, right?
At least in the non-isolated case, not doing anything at all is of course more ideal. Although that would of course involve hoisting the condition, leading to code duplication.

vello_shaders/shader/fine.wgsl
```diff
@@ -987,18 +988,23 @@ fn main(
                 cmd_ix += 2u;
             }
             case CMD_BEGIN_CLIP: {
```
DJMcNab, line 990: My vague plan for implementing this was to turn it into a whole new command, as the operations are at least a little different. I presume you didn't want to do that to avoid the code duplication?

vello_shaders/shader/fine.wgsl
DJMcNab, line 1007: I presume that the multiply by one (or indeed by zero) can't cause rendering differences. Given that these are only colours, I can't see that it can.

DJMcNab (commented):
It's a shame this doesn't build on #1199, so it doesn't have a test.

I think this will also need to apply to the CPU shaders?

vello_shaders/shader/fine.wgsl
```diff
@@ -987,18 +988,23 @@ fn main(
...
+                var rgba_mul = vec4(0.0);
+                if is_non_isolated {
+                    rgba_mul = vec4(1.0);
+                } 
                 if clip_depth < BLEND_STACK_SPLIT {
                     for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
                         blend_stack[clip_depth][i] = pack4x8unorm(rgba[i]);
-                        rgba[i] = vec4(0.0);
-                    }
+                        rgba[i] *= rgba_mul;
```
DJMcNab, line 999: You're sure the backdrop isn't now double counted, even if it has alpha? I think that `a over a` if a is partially transparent isn't equal to a.

vello_shaders/shader/coarse.wgsl
```diff
@@ -413,7 +412,12 @@ fn main(
                         if tile.segment_count_or_ix == 0u && tile.backdrop == 0 {
                             clip_zero_depth = clip_depth + 1u;
                         } else {
-                            write_begin_clip();
+                            let blend = scene[dd];
```
DJMcNab, line 415: Handy that this was available. Seeing what the blend mode was at "write begin clip" time was my greatest worry in implementing this (as in I was worried it would be annoying)

PR #1203, opened 2025-09-09
Introduce a new clipping algorithm for non-layer based clipping

LaurenzV:
Bear with me as this is still not done, lots of cleanup, more tests and documentation needed, just opening this for tracking. Mainly wanted to get a quick dirty version working first! :)

# Motivation
Currently, the only way to do efficient clipping in Vello Sparse is by pushing a new isolated layer and associating a clip path with that layer. This might be sufficient for cases like for example SVG, but the problem is that doing so forces the user to conflate layers and clipping, which is not what should be done in some imaging models. For example, clip paths in PDF have nothing to do with isolation, and it is therefore desirable to be able to separate the two. I believe the same is the case for HTML canvas, where the clip stack is not really related to layers, and the fact they are causes problems (see https://github.com/linebender/vello/issues/1117). In practice, this mostly doesn't cause issues as (to my understanding) source-over compositing with premultiplied alpha is associative and it therefore mostly doesn't matter whether you clip with isolation or not. However, problems start to occur once you add masks or blending to the mix, as the behavior with and without isolation is not identical.

It should be noted that above can be simulated by creating a new mask for each clip path, but this is obviously expensive, especially for deeply-nested clips.

# Solution
Because of this, I propose a second way to support clipping in Vello CPU (could be added to hybrid as well) that is based on an independent clip stack that can be pushed and popped to without actually having to create new layers. In particular, you do not need to clear the clip stack before rasterizing, which should adddress the issue in #1117.

I think how this works is very neat and demonstrates how useful sparse strips are! Fundamentally, this is based on a new algorithm that, given two paths in sparse strips representation, computes a new path in sparse strips representation that represents the intersection of the two. While the actual implementation is a bit finicky, conceptually it's very simple: We iterate over each stripped and fill region of the two paths in lock step and determine all of the overlaps between the two. In case we have two fill regions, the overlap region will also be filled. In case we have one strip and one fill region, the overlap will copy the alpha mask of the strip region. Finally, if we have two strip regions, we combine the alpha masks of both. That's it, this is enough to compute a new sparse strips path that when filled is equivalent to intersecting the two strips. The beauty of this is that this is incredibly fast because of the usual sparseness of paths and the result will also still be as sparse as possible as we preserve sparse fill regions whenever possible.

Given this, we introduce two new `push_clip_path` and `pop_clip_path` methods that operate on their own clip stack and completely independently from the push/pop layer commands. In case a clip path is currently active, once we draw a path we simply calculate the intersection after converting it into sparse strips, and then use that new path as the basis in coarse rasterization instead. In case we call `push_clip_path` multiple times, we can simply intersect the new clip path with the one in the previous stack position, to end up with one single new clip path that represents the intersection of all previous ones. Therefore, even for deeply nested clips we only need to perform one intersection per path drawn.

The disadvantage is that in multi-threading, it requires us to process new clip paths on the main thread, but the advantage is that (unlike the current clip algorithm) causes no additional overhead during rasterization (which is currently the only serial part), and the clipping calculation is done by each worker separately for the path they are currently processing, which means it scales very well!

# Testing
I will add more tests, but I confirmed the current version to cause no regressions in my PDF test suite.

### DIFF
--- a/sparse_strips/vello_common/src/strip_generator.rs
+++ b/sparse_strips/vello_common/src/strip_generator.rs
@@ -7,7 +7,7 @@ use crate::fearless_simd::Level;
 use crate::flatten::{FlattenCtx, Line};
 use crate::kurbo::{Affine, PathEl, Stroke};
 use crate::peniko::Fill;
-use crate::strip::Strip;
+use crate::strip::{intersect, IntersectInputOwned, IntersectInputRef, IntersectOutput, Strip};
 use crate::tile::Tiles;
 use crate::{flatten, strip};
 use alloc::vec::Vec;
@@ -16,11 +16,13 @@ use alloc::vec::Vec;
 #[derive(Debug)]
 pub struct StripGenerator {
     level: Level,
-    alphas: Vec<u8>,
+    global_alphas: Vec<u8>,
+    temp_alphas: Vec<u8>,
     line_buf: Vec<Line>,
     flatten_ctx: FlattenCtx,
     tiles: Tiles,
-    strip_buf: Vec<Strip>,
+    strip_buf1: Vec<Strip>,
+    strip_buf2: Vec<Strip>,
     width: u16,
     height: u16,
 }
@@ -29,11 +31,13 @@ impl StripGenerator {
     /// Create a new strip generator.
     pub fn new(width: u16, height: u16, level: Level) -> Self {
         Self {
-            alphas: Vec::new(),
+            global_alphas: Vec::new(),
+            temp_alphas: Vec::new(),
             level,
             line_buf: Vec::new(),
             tiles: Tiles::new(level),
-            strip_buf: Vec::new(),
+            strip_buf1: Vec::new(),
+            strip_buf2: Vec::new(),
             flatten_ctx: FlattenCtx::default(),
             width,
             height,
@@ -43,11 +47,12 @@ impl StripGenerator {
     /// Generate the strips for a filled path.
     pub fn generate_filled_path<'a>(
         &'a mut self,
-        path: impl IntoIterator<Item = PathEl>,
-        fill_rule: Fill,
+        path: impl IntoIterator<Item=PathEl>,
+        mut fill_rule: Fill,
         transform: Affine,
         aliasing_threshold: Option<u8>,
-        func: impl FnOnce(&'a [Strip]),
+        clip_path: Option<IntersectInputRef<'_>>,
+        func: impl FnOnce(&'a [Strip], &'a [u8]),
     ) {
         flatten::fill(
             self.level,
@@ -56,8 +61,33 @@ impl StripGenerator {
             &mut self.line_buf,
             &mut self.flatten_ctx,
         );
-        self.make_strips(fill_rule, aliasing_threshold);
-        func(&mut self.strip_buf);
+        
+        if let Some(clip_path) = clip_path {
+            self.make_strips(fill_rule, aliasing_threshold, true);
+            
+            let input_path = IntersectInputRef {
+                strips: &self.strip_buf2,
+                alphas: &self.temp_alphas,
+                fill: fill_rule,
+            };
+            
+            let target = IntersectOutput {
+                strips: &mut self.strip_buf1,
+                alphas: &mut self.global_alphas,
+                fill: &mut fill_rule,
+            };
+
+            intersect(
+                self.level,
+                clip_path,
+                input_path,
+                target
+            );
+        }   else {
+            self.make_strips(fill_rule, aliasing_threshold, false);
+        }
+
+        func(&self.strip_buf1, &self.global_alphas);
     }
 
     /// Generate the strips for a stroked path.
@@ -67,8 +97,11 @@ impl StripGenerator {
         stroke: &Stroke,
         transform: Affine,
         aliasing_threshold: Option<u8>,
+        clip_path: Option<IntersectInputRef<'_>>,
         func: impl FnOnce(&'a [Strip]),
     ) {
+        let mut fill_rule = Fill::NonZero;
+        
         flatten::stroke(
             self.level,
             path,
@@ -77,52 +110,87 @@ impl StripGenerator {
             &mut self.line_buf,
             &mut self.flatten_ctx,
         );
-        self.make_strips(Fill::NonZero, aliasing_threshold);
-        func(&mut self.strip_buf);
+
+        if let Some(clip_path) = clip_path {
+            self.make_strips(fill_rule, aliasing_threshold, true);
+
+            let input_path = IntersectInputRef {
+                strips: &self.strip_buf2,
+                alphas: &self.temp_alphas,
+                fill: fill_rule,
+            };
+
+            let target = IntersectOutput {
+                strips: &mut self.strip_buf1,
+                alphas: &mut self.global_alphas,
+                fill: &mut fill_rule,
+            };
+
+            intersect(
+                self.level,
+                clip_path,
+                input_path,
+                target
+            );
+        }   else {
+            self.make_strips(fill_rule, aliasing_threshold, false);
+        }
+
+        func(&self.strip_buf1);
     }
 
     /// Return a reference to the current alpha buffer of the strip generator.
     pub fn alpha_buf(&self) -> &[u8] {
-        &self.alphas
+        &self.global_alphas
     }
 
     /// Extend the alpha buffer with the given alphas.
     pub fn extend_alpha_buf(&mut self, alphas: &[u8]) {
-        self.alphas.extend_from_slice(alphas);
+        self.global_alphas.extend_from_slice(alphas);
     }
 
     /// Set the alpha buffer.
     pub fn set_alpha_buf(&mut self, alpha_buf: Vec<u8>) {
-        self.alphas = alpha_buf;
+        self.global_alphas = alpha_buf;
     }
 
     /// Take the alpha buffer and set it to an empty one.
     pub fn take_alpha_buf(&mut self) -> Vec<u8> {
-        core::mem::take(&mut self.alphas)
+        core::mem::take(&mut self.global_alphas)
     }
 
     /// Swap the alpha buffer with the given one.
     pub fn replace_alpha_buf(&mut self, alphas: Vec<u8>) -> Vec<u8> {
-        core::mem::replace(&mut self.alphas, alphas)
+        core::mem::replace(&mut self.global_alphas, alphas)
     }
 
     /// Reset the strip generator.
     pub fn reset(&mut self) {
         self.line_buf.clear();
         self.tiles.reset();
-        self.alphas.clear();
-        self.strip_buf.clear();
+        self.global_alphas.clear();
+        self.temp_alphas.clear();
+        self.strip_buf1.clear();
+        self.strip_buf2.clear();
     }
 
-    fn make_strips(&mut self, fill_rule: Fill, aliasing_threshold: Option<u8>) {
+    fn make_strips(&mut self, fill_rule: Fill, aliasing_threshold: Option<u8>, temp: bool) {
         self.tiles
             .make_tiles(&self.line_buf, self.width, self.height);
         self.tiles.sort_tiles();
+        
+        let (alphas, strip_buf) = if temp {
+            self.temp_alphas.clear();
+            (&mut self.temp_alphas, &mut self.strip_buf2)
+        }   else {
+            (&mut self.global_alphas, &mut self.strip_buf1)
+        };
+        
         strip::render(
             self.level,
             &self.tiles,
-            &mut self.strip_buf,
-            &mut self.alphas,
+            strip_buf,
+            alphas,
             fill_rule,
             aliasing_threshold,
             &self.line_buf,
@@ -147,17 +215,18 @@ mod tests {
             Fill::NonZero,
             Affine::IDENTITY,
             None,
-            |_| {},
+            None,
+            |_, _| {},
         );
 
         assert!(!generator.line_buf.is_empty());
-        assert!(!generator.strip_buf.is_empty());
-        assert!(!generator.alphas.is_empty());
+        assert!(!generator.strip_buf1.is_empty());
+        assert!(!generator.global_alphas.is_empty());
 
         generator.reset();
 
         assert!(generator.line_buf.is_empty());
-        assert!(generator.strip_buf.is_empty());
-        assert!(generator.alphas.is_empty());
+        assert!(generator.strip_buf1.is_empty());
+        assert!(generator.global_alphas.is_empty());
     }
 }
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -11,7 +11,6 @@ use crate::dispatch::multi_threaded::MultiThreadedDispatcher;
 use crate::dispatch::single_threaded::SingleThreadedDispatcher;
 use crate::kurbo::{PathEl, Point};
 use alloc::boxed::Box;
-#[cfg(feature = "text")]
 use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -27,7 +27,7 @@ use vello_common::peniko::color::palette::css::BLACK;
 use vello_common::peniko::{BlendMode, Compose, Fill, Mix};
 use vello_common::pixmap::Pixmap;
 use vello_common::recording::{PushLayerCommand, Recordable, Recording, RenderCommand};
-use vello_common::strip::Strip;
+use vello_common::strip::{intersect, IntersectInputOwned, Strip};
 use vello_common::strip_generator::StripGenerator;
 #[cfg(feature = "text")]
 use vello_common::{
@@ -48,6 +48,8 @@ pub struct RenderContext {
     pub(crate) transform: Affine,
     pub(crate) fill_rule: Fill,
     pub(crate) temp_path: BezPath,
+    pub(crate) clip_stack: Vec<Arc<IntersectInputOwned>>,
+    pub(crate) strip_generator: StripGenerator,
     pub(crate) aliasing_threshold: Option<u8>,
     pub(crate) encoded_paints: Vec<EncodedPaint>,
     #[cfg_attr(
@@ -140,6 +142,8 @@ impl RenderContext {
             aliasing_threshold,
             paint,
             render_settings: settings,
+            clip_stack: vec![],
+            strip_generator: StripGenerator::new(width, height, settings.level),
             paint_transform,
             fill_rule,
             stroke,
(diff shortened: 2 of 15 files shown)


LaurenzV:
cc @sagudev maybe interesting for you as well :)

LaurenzV:
I think this should be ready for review now.

LaurenzV:
@taj-p Lemme know if that would be useful for you in vello_hybrid, but since this hasn't really been talked about before I presume that the current clipping algo is good enough for you, so I haven't implemented it there.

taj-p:
> @taj-p Lemme know if that would be useful for you in vello_hybrid, but since this hasn't really been talked about before I presume that the current clipping algo is good enough for you, so I haven't implemented it there.

We actually make quite a lot of use of clip paths, so I'm wondering whether this would be faster than the layer implementation. For significantly clipped images, I imagine this would be faster (because the layer implementation requires drawing the entire image to a layer for the next to sample from). 

But, for milder clips, I'm not so sure since it's a question of whether it's faster to either: 
1. rasterise the strips using the GPU and suffer a texture copy to sample those pixels; or
2. run `intersect` in the CPU and rasterise the strips directly to the target in the GPU without sampling

> The beauty of this is that this is incredibly fast

My feeling is that `intersect` seems very quick, so this might be faster. I created a quick way to benchmark this in hybrid (https://github.com/linebender/vello/pull/1204). Essentially, I just add more of the deeply nested circles to the scene and measure FPS.

I saw the `clip_path` variant to be roughly 1/3 slower than the `clip_layer` approach, which I thought was mildly surprising. I suspect this is also owing to running single threaded on a Mac with a powerful GPU 🤔 . I would suspect users on less powerful GPUs will see a significant boost with this PR.

Are you able to confirm whether I properly migrated this to work as you'd expect in hybrid?

image
 
Cc @grebmeg for your thoughts

LaurenzV:
Thanks! The result also surprises me a bit, though it's possible that pushing/popping on the GPU is so fast that it doesn't really matter, but I think it would matter for the CPU. I will try your experiment in vello_cpu tomorrow and see how it performs.

To be fair, I didn't mean to imply that this version is faster than the current layer-based clipping! Just that this version of clipping should be much faster than using masks for simulating non-isolated clips. 😄 Based on intuition, on the CPU I would at least expect that this version is faster when dealing with many nested clip paths where only few paths are drawn, while the layer-based approach is faster when few clips are used but many paths are drawn with those clips active. But I haven't benchmarked it and it's possible that things are different on the GPU.

grebmeg:
It’s really nice that you can use sparse strips of two paths and get their intersection, and it’s SIMD-accelerated!

That said, I’d like to understand the main goal of this PR more clearly: is it primarily about performance, about enabling clipping of intermediate results (as in issue #1117), or about separating two concepts that don’t quite fit together, or all together?

From a performance perspective, `push_clip_path` form this PR seems to replace our existing GPU clipping algorithm, which is based on two textures ping-ponging. On the CPU side, we only compute where and when those clips should be applied, which is why it’s fairly fast. In contrast, this change adds some per-path intersection overhead. Roughly speaking, our current GPU clipping looks equivalent to `intersect` in this PR. My impression is that stronger GPUs would benefit more here, but I’m curious about where the crossover point lies where a CPU-only approach actually becomes faster. Would it be possible to benchmark the both clip approaches (similar to what Taj did) across various patterns?

That said, your change brings some nice improvements on CPU, especially avoiding extra clip buffer allocations per tile, and making it possible to immediately see clipping result when a clip is active.

I’m also wondering whether your intersection algorithm could be integrated into current layer clipping in `pop_clip`. Right now, that code works with clip bounding boxes and calculates all clip commands on pop. Would it be possible to use your `intersect` function to compute the intersection between a rectangular clip box and the path strips, and then run `wide.generate` on the resulting strips, assuming they are now all either clip alpha or clip fill?

I also think that for `vello_hybrid` it might be worth exploring whether we can avoid using `intersect` and instead emulate the push/clip/pop commands ourselves, allowing the shader code to perform the actual clipping for each path and this way, we could also see the clipping result immediately for each path.

LaurenzV:
> That said, I’d like to understand the main goal of this PR more clearly: is it primarily about performance, about enabling clipping of intermediate results (as in issue https://github.com/linebender/vello/issues/1117), or about separating two concepts that don’t quite fit together, or all together?

It's basically as outlined in the `Motivation` section in this PR: The main motivation here is not performance, but simply that the current clipping algorithm assumes that we apply layer isolation each time we use a clip path, which is not what we want to do in some imaging models. My intention isn't to replace the existing layer clipping, but simply provide an alternative so that we can achieve it without layer isolation, so you can still choose between the two if you want. With that said, Taj has successfully nerd-sniped me in wanting to understand the performance differences better now. :)

> Would it be possible to use your intersect function to compute the intersection between a rectangular clip box and the path strips, and then run wide.generate on the resulting strips, assuming they are now all either clip alpha or clip fill?

Looking at it a bit more closely, I believe this _should_ be possible and it could simplify the code logic a bit, but the core difficulty in the current algo seems to be in other places so I'm not sure how much simpler it would make the current code.

LaurenzV:
I can reproduce the slowdown in `vello_cpu` and the reason for it is that there is a lot of allocating/deallocating for the clip path data going as the number of circles increases. I probably know how to fix it, but it requires some small architectural changes which I will submit as separate PRs before revisiting this one.

dominikh:
This way of clipping does have the perhaps surprising effect that clips will not affect layer composition at all. When we pop a layer, the active non-layer clip will not restrict where the layer gets composited.

LaurenzV:
Yeah, the intention is that you pass a clip path to `clip_layer` instead if you want that (although it's still broken right now). But I agree that we need to make lots of improvements to the documentation, not only for this, but in general.

dominikh:
> Yeah, the intention is that you pass a clip path to clip_layer instead if you want that

It might make sense to expose the active non-layer clip path in that case, so that it can easily be set on a layer, too.

I'm imagining a hypothetical UI building use case, where each child widget gets clipped to its dimensions before it is allowed to render itself. That's a prime example of nested clips that might benefit from using this new algorithm. However, every once in a while, a widget may have need for an isolated layer and fancy compositing, in which case it should be able to inherit the active non-layer clip.

(There might also be some value in renaming layer clips to something else to avoid confusion. Maybe stencils or shapes or masks?)

Interestingly, while Skia separates clips and layers, they share a save/restore stack. You can either save just the current clip (and transformation matrix), or save the current clip and create a new layer. In either case you use `restore` to pop the state and in the case of a saved layer, blend it. Under that model, it would be possible to apply the current clip to the new layer at layer creation time. I don't think Skia has a separate layer-specific clipping model. For vello_cpu, that'd mean intersecting the current non-layer clip with the layer's optional clip.

That model should also allow for the following optimization: when the current layer's clip is identical to the current non-layer clip, then we don't have to intersect paths with the non-layer clip, as the layer already takes care of it. Realistically, "identical" would be implemented as "neither clip has changed since the layer's creation". Of course at that point we're moving further away from the PDF model, where clips and layers (transparency groups) don't share the same stack, I think (the full PDF model has a lot of fun corner cases and full support for it should probably not be the primary goal...)

LaurenzV:
@taj-p Feel free to retry now, when I try it now the new clip version is faster! Though we do need to consider that it's kind of an edge case because we are only drawing one path per clip path, I assume the difference is going to amortize and turn around as the number of paths per clip path increases (haven't tested it carefully, though).

taj-p:
> @taj-p Feel free to retry now, when I try it now the new clip version is faster! Though we do need to consider that it's kind of an edge case because we are only drawing one path per clip path, I assume the difference is going to amortize and turn around as the number of paths per clip path increases (haven't tested it carefully, though).

I'm seeing a roughly 20-25% improvement in performance 😮 ! It's likely to be more for users with worse GPUs too. I wouldn't consider this too much an edge case. It's common to, for example, use a clipping region across a single image.

cc @grebmeg / @ajakubowicz-canva

sparse_strips/vello_example_scenes/src/clip.rs
```diff
@@ -68,26 +100,57 @@ pub fn render(ctx: &mut Scene, root_transform: Affine) {
...
+
+        for _ in 0..outer_count {
+            for color in COLORS.iter() {
+                let clip_circle = Circle::new(center, radius).to_path(0.1);
+                draw_clipping_outline(ctx, &clip_circle);
+                if use_clip_path {
+                    ctx.push_clip_path(&clip_circle);
+                } else {
+                    ctx.push_clip_layer(&clip_circle);
+                }
```
grebmeg, line 141: When toggling, the results seem to differ in antialiasing?

push_clip_path|push_clip_layer
---|---
Image|Image

sparse_strips/vello_sparse_tests/tests/clip.rs
grebmeg, line 1: Maybe it would help to add a test showing a more complex option that uses both clipping APIs? Roughly, something like this:
```rust
#[vello_test]
fn clip_non_isolated_rectangle_with_star_evenodd(ctx: &mut impl Renderer) {
    let rect = Rect::new(0.0, 0.0, 100.0, 100.0);
    let rect2 = Rect::new(10.0, 10.0, 80.0, 80.0);
    let rect3 = Rect::new(40.0, 40.0, 90.0, 90.0);
    let rect4 = Rect::new(70.0, 70.0, 75.0, 75.0);

    ctx.push_clip_path(&rect2.to_path(0.1));
    ctx.push_layer(Some(&rect3.to_path(0.1)), None, None, None);
    ctx.push_clip_path(&rect4.to_path(0.1));
    ctx.set_paint(REBECCA_PURPLE);
    ctx.fill_rect(&rect);
    ctx.pop_clip_path();
    ctx.pop_layer();
    ctx.pop_clip_path();
}
```

grebmeg: Just to confirm, is it expected behavior that pushed clip paths (`push_clip_path`) still affect internal paths rendering within the layer (in `push_layer`)?

grebmeg (commented):
> I’m also wondering whether your intersection algorithm could be integrated into current layer clipping in pop_clip. Right now, that code works with clip bounding boxes and calculates all clip commands on pop. Would it be possible to use your intersect function to compute the intersection between a rectangular clip box and the path strips, and then run wide.generate on the resulting strips, assuming they are now all either clip alpha or clip fill?

> Looking at it a bit more closely, I believe this should be possible and it could simplify the code logic a bit, but the core difficulty in the current algo seems to be in other places so I'm not sure how much simpler it would make the current code.

I suspect your intersection algorithm could also be applied to `push_clip`/`pop_clip` in coarse, which might remove the need for the current non-SIMD code while still replicating the existing push/pop clip buffer behavior and generating clip commands (I’m curious whether this approach would lead to better performance). It might also make sense to maintain a single clip stack.

sparse_strips/vello_example_scenes/src/clip.rs
LaurenzV: Hmm that does look weird, will try to take a look when I find the time!

sparse_strips/vello_sparse_tests/tests/clip.rs
LaurenzV: Yes, that is expected. All clip paths pushed via `push_clip_path` apply to all path drawing operations, regardless of which layer they are drawn on.

sparse_strips/vello_example_scenes/src/clip.rs
LaurenzV: I cannot reproduce this on my machine, both cases look like the `push_clip_layer` one for me...

taj-p:
I will review this over the next few days 🙏

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -428,6 +428,21 @@ impl RenderContext {
         self.glyph_caches.as_mut().unwrap().maintain();
     }
 
+    /// Push a new clip path to the clip stack.
+    pub fn push_clip_path(&mut self, path: &BezPath) {
```
taj-p, line 451: Would it be possible to add documentation here specifying why you would choose `push_clip_path` over `push_layer` or `push_clip_layer` (and also update those methods with reference to when to use `push_clip_path`? (both here and in hybrid)

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -428,6 +428,21 @@ impl RenderContext {
...
+        self.dispatcher.push_clip_path(
+            path,
+            self.fill_rule,
+            self.transform,
+            self.aliasing_threshold,
+        );
+    }
+
+    /// Pop a clip path from the clip stack.
```
taj-p, line 460: Should we specify that consumers actually needn't pop the clip paths stack back to empty like they need to for layers? (both here and in hybrid)

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -114,34 +126,64 @@ impl StripGenerator {
...
+            let path_data = PathDataRef {
+                strips: &self.temp_storage.strips,
+                alphas: &self.temp_storage.alphas,
+            };
+
+            intersect(self.level, clip_path, path_data, strip_storage);
+        } else {
+            self.make_strips(strip_storage, fill_rule, aliasing_threshold, false);
+        }
     }
```
taj-p, line 176:
```suggestion
    fn generate_with_clip(
        &mut self,
        aliasing_threshold: Option<u8>,
        strip_storage: &mut StripStorage,
        fill_rule: Fill,
        clip_path: Option<PathDataRef<'_>>,
    ) {
        if strip_storage.generation_mode == GenerationMode::Replace {
            strip_storage.strips.clear();
        }

        self.tiles
            .make_tiles(&self.line_buf, self.width, self.height);
        self.tiles.sort_tiles();

        if let Some(clip_path) = clip_path {
            strip::render(
                self.level,
                &self.tiles,
                &mut self.temp_storage.strips,
                &mut self.temp_storage.alphas,
                fill_rule,
                aliasing_threshold,
                &self.line_buf,
            );
            let path_data = PathDataRef {
                strips: &self.temp_storage.strips,
                alphas: &self.temp_storage.alphas,
            };

            intersect(self.level, clip_path, path_data, strip_storage);
        } else {
            strip::render(
                self.level,
                &self.tiles,
                &mut strip_storage.strips,
                &mut strip_storage.alphas,
                fill_rule,
                aliasing_threshold,
                &self.line_buf,
            );
        }
    }
```

We suffer an unnecessary branch in `make_strips` via the `temp` flag - one `if`/`else` in `generate_with_clip` on `clip_path` and again in `make_strips` via `temp`.
 
We can skip the redundant branching by inlining the function (and removing `make_strips`)

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+
+        Self {
+            input,
+            cur_idx,
+            strip_y,
+            on_strip: true,
+        }
+    }
+
+    fn cur_strip(&self) -> &Strip {
```
taj-p, line 413:
```suggestion
    #[inline(always)]
    fn cur_strip(&self) -> &Strip {
```

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        &self.input.strips[*self.cur_idx]
+    }
+
+    fn next_strip(&self) -> &Strip {
```
taj-p, line 418:
```suggestion
    #[inline(always)]
    fn next_strip(&self) -> &Strip {
```

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        &self.input.strips[*self.cur_idx + 1]
+    }
+
+    fn cur_strip_width(&self) -> u16 {
```
taj-p, line 423:
```suggestion
    #[inline(always)]
    fn cur_strip_width(&self) -> u16 {
```

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        let cur = self.cur_strip();
+        let next = self.next_strip();
+        ((next.alpha_idx() - cur.alpha_idx()) / Tile::HEIGHT as u32) as u16
+    }
+
+    fn cur_strip_alphas(&self) -> &'a [u8] {
```
taj-p, line 430:
```suggestion
    #[inline(always)]
    fn cur_strip_alphas(&self) -> &'a [u8] {
```

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        // zero winding so we don't need to special case this.
+        if next.fill_gap() {
+            let x = cur.x + self.cur_strip_width();
+            let width = next.x - x;
+
+            Some(FillRegion { start: x, width })
+        } else {
+            None
+        }
+    }
```
taj-p:
```suggestion
        let next = self.next_strip();

        // Note that if the next strip happens to be on the next line, it will always have
        // zero winding so we don't need to special case this.
        if next.fill_gap() {
            let cur = self.cur_strip();
            let x = cur.x + self.cur_strip_width();
            let width = next.x - x;

            Some(FillRegion { start: x, width })
        } else {
            None
        }
```

`cur` isn't required unless `next.fill_gap()` is true

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+enum Region<'a> {
+    Fill(FillRegion),
+    Strip(StripRegion<'a>),
+}
+
+impl Region<'_> {
+    fn start(&self) -> u16 {
```
taj-p, line 338:
```suggestion
    #[inline(always)]
    fn start(&self) -> u16 {
```

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        match self {
+            Region::Fill(fill) => fill.start,
+            Region::Strip(strip) => strip.start,
+        }
+    }
+
+    fn width(&self) -> u16 {
```
taj-p, line 346:
```suggestion
    #[inline(always)]
    fn width(&self) -> u16 {
```

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        match self {
+            Region::Fill(fill) => fill.width,
+            Region::Strip(strip) => strip.width,
+        }
+    }
+
+    fn end(&self) -> u16 {
```
taj-p, line 354:
```suggestion
    #[inline(always)]
    fn end(&self) -> u16 {
```

I think we want all these functions inlined so that LLVM can freely reuse the loaded values (so we don't end up re-running any meaningful work by not storing some of these calls as locals)

sparse_strips/vello_common/src/clip.rs
```diff
@@ -0,0 +1,717 @@
...
+        strips.push(Strip::new(
+            state.x,
+            cur_y * Tile::HEIGHT,
+            state.alpha_idx,
+            state.fill_gap,
+        ));
+    }
+}
+
+fn start_strip(strip_data: &mut Option<StripState>, alphas: &[u8], x: u16, fill_gap: bool) {
```
taj-p, line 516:
```suggestion
#[inline(always)]
fn start_strip(strip_data: &mut Option<StripState>, alphas: &[u8], x: u16, fill_gap: bool) {
```

taj-p (approved):
LGTM 🚀 🎉

sparse_strips/vello_cpu/src/render.rs
LaurenzV: If it's okay, I will put this as a TODO becuse I'm planning on creating some more in-depth examples that explain this once everything is merged.

PR #1206, opened 2025-09-10
Store data about filled areas in a strip

LaurenzV:
Currently, in a strip we store the coarse winding number that was determined while generating it. The thing is that in later part of the stages, we don't care about the actual winding number but just whether the area before a strip is filled. Because just the winding number doesn't give us that information, we need to carry the fill rule around into every part of the code.

Therefore, I propose changing the representation of a strip from storing its coarse winding number to just storing a boolean flag that indicates  whether the gap on the left of the strip should be filled. In doing so, `Vec<Strip>` contains all the information to unambiguously represent the rendered shape, saving us a lot of work later on.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -1073,7 +1073,7 @@ mod tests {
     use crate::color::AlphaColor;
     use crate::color::palette::css::TRANSPARENT;
     use crate::paint::{Paint, PremulColor};
-    use crate::peniko::{BlendMode, Compose, Fill, Mix};
+    use crate::peniko::{BlendMode, Compose, Mix};
     use crate::strip::Strip;
     use alloc::{boxed::Box, vec};
 


sparse_strips/vello_common/src/strip.rs
```diff
@@ -19,8 +19,9 @@ pub struct Strip {
     pub y: u16,
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
-    /// The winding number at the start of the strip.
-    pub winding: i32,
+    /// Whether the gap that lies between this strip and the previous in the _same_
+    /// row should be filled.
+    pub fill_gap: bool,
```
tomcur, line 24: This comment is just an observation and not meant to be blocking.

One downside is that we lose the ability to reuse the same strips (e.g., from a cache) under a different fill rule. Granted, that's not something the current API allows, either.

Even keeping that option open, storing the winding could be reduced to two bits representing three distinct states: non-zero-ness and evenness (or oddness, mapping 0 winding to `00`).

LaurenzV: That's true, we can easily add that in case we need it. 😄

sparse_strips/vello_common/src/strip.rs
```diff
@@ -19,8 +19,9 @@ pub struct Strip {
     pub y: u16,
     /// The index into the alpha buffer.
     pub alpha_idx: u32,
```
taj-p, line 21: I think there might be some value in using the last bit of `alpha_idx` as `fill_gap` instead of creating an additional bool. Since `Strip` has an alignment of 4 bytes (due to `alpha_idx`), the contribution of `fill_gap` is an additional 4 bytes. This means `Strip` becomes 12 bytes or 5.3 strips per cache line.

If we instead use the last bit of `alpha_idx` as `fill_gap`, `Strip` becomes 8 bytes or 8 Strips per cache line and much less to allocate. 

You can see this [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e6546efa7a37309828d39a475ad0ce7d).

I'm wondering whether it's possible to evaluate that change ([very rough example branch](https://github.com/linebender/vello/compare/laurenz/strip_winding...taj-p:vello:tajp/strip_winding_review?expand=1)) using your benchmark apparatus? When I tried this on Vello Hybrid, I saw no change in performance for direct rendering, but a minor improvement in rendering from recordings (likely due to `Strip` being cheaper to allocate and iterate).

taj-p: I assume that we're fine using 30 bits for `alpha_idx` if we do end up needing what Tom suggests [here](https://github.com/linebender/vello/pull/1206#discussion_r2336263447)?

LaurenzV: I do think this makes sense! I'll open an issue for it though, so that it doesn't block this change (which should be an improvement to what we had before, anyway 😄).

PR #1208, opened 2025-09-10
Separate storage of strips and alphas from strip generator

LaurenzV:
I based this on top of #1206 to avoid merge conflicts.

Currently, the strip generator also stores the strip and alpha buffers. This is fine if we can assume that there only is one single buffer for those globally, but this has already been proven to be a bit awkard once recordings were introduced. In addition to that, it also makes things more difficult for the new clipping PR as well as an optimization I want to implement for multi-threaded rendering. 

Therefore, this PR separates them into two different objects by introducing a new `StripStorage` and forcing the user to explicitly pass a `StripStorage` that should be used as the basis for the generation when interacting with the strip generator.

### DIFF
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -41,9 +41,7 @@ impl CachedStrips {
 
     /// Check if this cached strips is empty.
     pub fn is_empty(&self) -> bool {
-        self.strip_storage.strips.is_empty()
-            && self.strip_storage.alphas.is_empty()
-            && self.strip_start_indices.is_empty()
+        self.strip_storage.is_empty() && self.strip_start_indices.is_empty()
     }
 
     /// Get the number of strips.
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -12,59 +12,58 @@ use crate::paint::PaintType;
 use crate::peniko::Font;
 use crate::peniko::{BlendMode, Fill};
 use crate::strip::Strip;
+use crate::strip_generator::StripStorage;
 use alloc::vec::Vec;
 
 /// Cached sparse strip data.
 #[derive(Debug, Default)]
 pub struct CachedStrips {
-    /// The cached sparse strips.
-    strips: Vec<Strip>,
-    /// The alpha buffer data.
-    alphas: Vec<u8>,
+    /// The strip storage.
+    strip_storage: StripStorage,
     /// Strip start indices for each geometry command.
     strip_start_indices: Vec<usize>,
 }
 
 impl CachedStrips {
     /// Create a new cached strips instance.
-    pub fn new(strips: Vec<Strip>, alphas: Vec<u8>, strip_start_indices: Vec<usize>) -> Self {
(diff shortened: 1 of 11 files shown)


sparse_strips/vello_common/src/recording.rs
```diff
@@ -12,59 +12,58 @@ use crate::paint::PaintType;
...
-        self.alphas.clear();
+        self.strip_storage.reset();
         self.strip_start_indices.clear();
     }
 
     /// Check if this cached strips is empty.
     pub fn is_empty(&self) -> bool {
-        self.strips.is_empty() && self.alphas.is_empty() && self.strip_start_indices.is_empty()
+        self.strip_storage.strips.is_empty()
+            && self.strip_storage.alphas.is_empty()
```
grebmeg: I noticed that in some other places, including tests, these two are often used together. Could we also add `is_empty` on `strip_storage`? It seems useful.

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -12,15 +12,30 @@ use crate::tile::Tiles;
...
+pub struct StripStorage {
+    /// The strips in the storage.
+    pub strips: Vec<Strip>,
+    /// The alphas in the storage.
+    pub alphas: Vec<u8>,
+}
+
+impl StripStorage {
+    /// Reset the storage.
+    pub fn reset(&mut self) {
```
grebmeg: Semantically, this method feels closer to `clear` than `reset`: it empties the contents while keeping the storage valid and preserving allocation, rather than restoring it to an initial state.

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -77,52 +91,40 @@ impl StripGenerator {
...
-        self.strip_buf.clear();
     }
 
-    fn make_strips(&mut self, fill_rule: Fill, aliasing_threshold: Option<u8>) {
+    fn make_strips(
+        &mut self,
+        strip_storage: &mut StripStorage,
+        fill_rule: Fill,
+        aliasing_threshold: Option<u8>,
+        clear_strips: bool,
```
grebmeg: I think `strip_storage` should already provide the context on whether strips need to be cleared or not, so it would be nice to remove `clear_strips` boolean, maybe replacing it with something like the following?
```rust
#[derive(Debug, Default)]
pub struct StripStorage {
    pub strips: Vec<Strip>,
    pub alphas: Vec<u8>,
    mode: GenerationMode,
}

#[derive(Debug, Default)]
enum GenerationMode {
    #[default]
    // Clear before generating
    Replace,
    // Don't clear, just append  
    Append,
}

impl StripStorage {
    pub fn set_replace_mode(&mut self) -> &mut Self {
        self.mode = GenerationMode::Replace;
        self
    }
    
    pub fn set_append_mode(&mut self) -> &mut Self {
        self.mode = GenerationMode::Append;
        self
    }
    
    pub fn prepare_for_generation(&mut self) {
        match self.mode {
            GenerationMode::Replace => {
                self.strips.clear();
            }
            GenerationMode::Append => {
                // Do nothing - just append
            }
        }
    }
}
```

grebmeg (approved):
I’ve left a few minor suggestions, but overall — LGTM!

PR #1210, opened 2025-09-10
Optimize multi-threading when drawing small simple paths

LaurenzV:
This is based on #1206 as it makes use of some functionality introduced there.

Previously, each render task and coarse task in a batch would store its own allocation of path elements and strips. Instead, we now only store one such buffer per batch, and the commands simply store an index pointing to the exact location of where the necessary components are stored. This makes running times more consistent in situations where large batch sizes are created since the paths themselves are very cheap to rasterize:

image

image

image

We also can remove the ugly `SmallPath` abstraction we introduced previously, yay.

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -254,6 +254,9 @@ impl MultiThreadedDispatcher {
         let task_sender = self.task_sender.as_mut().unwrap();
         let task = RenderTask {
             idx: task_idx,
+            // TODO: Explore whether the main thread can hold a store of previous allocations so
+            // we can reuse them. Same for the strips that are returned from
+            // a child thread.
             path: path.into(),
             tasks: tasks.into(),
         };


LaurenzV:
Sorry, made a mistake here! This one's dependent on #1208, so that would be needed first.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -243,10 +247,17 @@ impl MultiThreadedDispatcher {
...
+    fn send_pending_tasks(&mut self) {
         let task_idx = self.bump_task_idx();
+        let tasks = self.task_batch.as_slice();
+        let path = self.batch_path.elements();
         let task_sender = self.task_sender.as_mut().unwrap();
-        task_sender.send((task_idx, tasks)).unwrap();
+        let task = RenderTask {
+            idx: task_idx,
+            path: path.into(),
+            tasks: tasks.into(),
```
taj-p, line 261: Optional comment: These `.into` are allocating onto the heap IIUC. There might be an opportunity to pass this memory directly to the worker and for the worker to return the allocations to a free list maintained on the main thread.

I did a quick POC [here](https://github.com/linebender/vello/compare/laurenz/better_mt...taj-p:vello:tajp/mtSuggestion?expand=1). I'm not sure how to reproduce your benchmarks to confirm whether it has much of an effect, but leaving it here in case you find it useful.

sparse_strips/vello_cpu/src/dispatch/multi_threaded/worker.rs
```diff
@@ -177,7 +165,14 @@ impl Worker {
             }
         }
 
-        result_sender.send(task_idx as usize, task_buf).unwrap();
+        let strips_slice = self.strip_storage.strips.as_slice();
+
+        let task = CoarseTask {
+            strips: strips_slice.into(),
```
taj-p, line 171: There might be a similar way to preserve `self.strip_storage.strips` allocations too 🤔

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
taj-p: I didn't actually see much benefit with it from doing a quick runs of `vello_toy` over ghosttiger with 8 threads. But, I'll leave it with you.

taj-p (approved):
LGTM

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Yeah you are right, this creates a new allocation. I definitely like the idea and think it's worth exploring, but I think I'll leave this for a future PR so it's easier to review. I'll add a TODO!

LaurenzV: Also, I think the main reason you are not seeing much benefit is that coarse rasterization really seems to be the main bottleneck now. Unfortunately parallelizing this will be trickier, because of the way our current clipping algorithm works. 😄

LaurenzV:
Thanks for the reviews!!

PR #1211, opened 2025-09-10
Add special case path for lines that only intersect one tile in `make_tiles`

LaurenzV:
Has nice perf impact for shapes with many short lines:

image

image


LaurenzV:
(This PR isn't relevant anymore.)

PR #1214, opened 2025-09-14
Don't manually flatten curves for which the control points are strictly within the tolerance threshold

LaurenzV:
A revival from an older PR. Confirmed to give a decent speed boost for 8x8 and 16x16 shapes.

### DIFF
--- a/sparse_strips/vello_common/src/flatten_simd.rs
+++ b/sparse_strips/vello_common/src/flatten_simd.rs
@@ -49,10 +49,25 @@ pub(crate) fn flatten<S: Simd>(
             }
             PathEl::QuadTo(p1, p2) => {
                 if let Some(p0) = last_pt {
-                    // If all control points are within the tolerance distance from the
-                    // start and end points, we can just draw a straight line and it's guaranteed
-                    // to be within the given tolerance.
-                    if (p0 - p1).hypot2() <= TOL_2 && (p2 - p1).hypot2() <= TOL_2 {
+                    // An upper bound on the shortest distance of any point on the quadratic Bezier
+                    // curve to the line segment [p0, p2] is 1/2 of the maximum of the
+                    // endpoint-to-control-point distances.
+                    //
+                    // The derivation is similar to that for the cubic Bezier (see below). In
+                    // short:
+                    //
+                    // q(t) = B0(t) p0 + B1(t) p1 + B2(t) p2
+                    // dist(q(t), [p0, p1]) <= B1(t) dist(p1, [p0, p1])
+                    //                       = 2 (1-t)t dist(p1, [p0, p1]).
+                    //
+                    // The maximum occurs at t=1/2, hence
+                    // max(dist(q(t), [p0, p1] <= 1/2 dist(p1, [p0, p1])).
+                    //
+                    // A cheap upper bound for dist(p1, [p0, p1]) is max(dist(p1, p0), dist(p1, p2)).
+                    //
+                    // The following takes the square to elide the square root of the Euclidean
+                    // distance.
+                    if f64::max((p1 - p0).hypot2(), (p1 - p2).hypot2()) <= 4. * TOL_2 {
                         callback.callback(PathEl::LineTo(p2));
                     } else {
                         let q = QuadBez::new(p0, p1, p2);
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -49,38 +49,56 @@ pub(crate) fn flatten<S: Simd>(
...
+                    // If all control points are within the tolerance distance from the
+                    // start and end points, we can just draw a straight line and it's guaranteed
+                    // to be within the given tolerance.
+                    if (p0 - p1).hypot2() <= TOL_2
+                        && (p0 - p2).hypot2() <= TOL_2
+                        && (p3 - p1).hypot2() <= TOL_2
+                        && (p3 - p2).hypot2() <= TOL_2
+                    {
+                        callback.callback(PathEl::LineTo(p3));
+                    } else {
```
tomcur: Yes... in fact, I believe the following tighter and cheaper-to-compute bound is right.

```suggestion
                    // An upper bound on the shortest distance of any point on the cubic Bezier
                    // curve to the line segment [p0, p3] is 3/4 of the maximum of the
                    // endpoint-to-control-point distances.
                    //
                    // With Bernstein weights Bi(t), we have
                    // c(t) = B0(t) p0 + B1(t) p1 + B2(t) p2 + B3(t) p3
                    // with t from 0 to 1 (inclusive).
                    //
                    // Through convexivity of the Euclidean distance function and the line segment,
                    // we have
                    // dist(c(t), [p0, p3]) <= B1(t) dist(p1, [p0, p3]) + B2(t) dist(p2, [p0, p3])
                    //                      <= B1(t) ||p1-p0|| + B2(t) ||p2-p3||
                    //                      <= (B1(t) + B2(t)) max(||p1-p0||, ||p2-p3|||)
                    //                       = 3 ((1-t)t^2 + (1-t)^2t) max(||p1-p0||, ||p2-p3||).
                    //
                    // The inner polynomial has its maximum of 1/4 at t=1/2, hence
                    // max(dist(c(t), [p0, p3])) <= 3/4 max(||p1-p0||, ||p2-p3||).
                    //
                    // The following takes the square to elide the square root of the Euclidean
                    // distance.
                    if f64::max((p0 - p1).hypot2(), (p3 - p2).hypot2()) <= 16. / 9. * TOL_2 {
                        callback.callback(PathEl::LineTo(p3));
                    } else {
```

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -49,38 +49,56 @@ pub(crate) fn flatten<S: Simd>(
...
-                    let step = 1.0 / (n as f64);
-                    for i in 1..n {
-                        let u = (i as f64) * step;
-                        let t = q.determine_subdiv_t(&params, u);
-                        let p = q.eval(t);
-                        callback.callback(PathEl::LineTo(p));
+                    // If all control points are within the tolerance distance from the
+                    // start and end points, we can just draw a straight line and it's guaranteed
+                    // to be within the given tolerance.
+                    if (p0 - p1).hypot2() <= TOL_2 && (p2 - p1).hypot2() <= TOL_2 {
```
tomcur: Another tighter bound, similar to the cubic Bezier one.

```suggestion
                    // An upper bound on the shortest distance of any point on the quadratic Bezier
                    // curve to the line segment [p0, p2] is 1/2 of the maximum of the
                    // endpoint-to-control-point distances.
                    //
                    // The derivation is similar to that for the cubic Bezier (see below). In
                    // short:
                    //
                    // q(t) = B0(t) p0 + B1(t) p1 + B2(t) p2
                    // dist(q(t), [p0, p1]) <= B1(t) dist(p1, [p0, p1])
                    //                       = 2 (1-t)t dist(p1, [p0, p1]).
                    //
                    // The maximum occurs at t=1/2, hence
                    // max(dist(q(t), [p0, p1] <= 1/2 dist(p1, [p0, p1])).
                    //
                    // A cheap upper bound for dist(p1, [p0, p1]) is max(dist(p1, p0), dist(p1, p2)).
                    //
                    // The following takes the square to elide the square root of the Euclidean
                    // distance.
                    if f64::max((p1 - p0).hypot2(), (p1 - p2).hypot2()) <= 4. * TOL_2 {
```

tomcur:
I've suggested two tighter (and cheaper to compute) bounds that I believe to be correct.

LaurenzV:
Nice! Yes I also figured that there probably is a way to generate an even tighter bbox, but lacked the maths skills to figure it out. 😄

LaurenzV:
Really impressive, this gives yet another pretty significant speed boost on top of what was initially introduced in this PR!

LaurenzV:
Here's the comparison to main now:

image
image
image

tomcur:
Not bad at all!

tomcur:
I've opened https://github.com/linebender/kurbo/pull/505, which may allow us to introduce tighter bounds.

The idea would be to test whether the control points are very close to their projection onto the chord/line segment defined by the Bezier end points, rather than just close to the end points.

Edit: see https://github.com/linebender/vello/pull/1216.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -49,38 +49,84 @@ pub(crate) fn flatten<S: Simd>(
...
+                    //                       = 2 (1-t)t dist(p1, [p0, p1]).
+                    //
+                    // The maximum occurs at t=1/2, hence
+                    // max(dist(q(t), [p0, p1] <= 1/2 dist(p1, [p0, p1])).
+                    //
+                    // A cheap upper bound for dist(p1, [p0, p1]) is max(dist(p1, p0), dist(p1, p2)).
+                    //
+                    // The following takes the square to elide the square root of the Euclidean
+                    // distance.
+                    if f64::max((p1 - p0).hypot2(), (p1 - p2).hypot2()) <= 4. * TOL_2 {
```
dominikh, line 70: @LaurenzV The function takes a `tolerance` argument, but hardcodes TOL_2 here. Is that an oversight?

LaurenzV: Uhmm good question, I think you are right, we probably should remove the parameter and always use the constant

PR #1215, opened 2025-09-14
Glyph Caches: Hinting Instances + Outline Paths

taj-p:
We found that ~80% of the time in preparing a glyph to render is in building hinting instances and extracting glyph outlines as shown by the below profile.

image

This PR adds LRU caches for hinting instances and getting outlines. The strategy mimics the one used by [Vello Classic](https://github.com/linebender/vello/blob/main/vello_encoding/src/glyph_cache.rs) for simplicity.

The outline cache grows to accommodate all the glyphs that can be rendered in a given frame and runs its eviction process within the renderer's `flush` function. Since outlines are relatively cheap on memory, I think that this approach is better than checking cache bounds on each access.

Note: This implies that `flush` must always be called once per frame for hybrid like vello cpu. But, I think this is fine. It's likely we'll want to run some things in a multi threaded environment in hybrid at some point.

Depending on the text being rendered, we've observed significant boosts to performance using this PR. In Vello Hybrid testing, for a super simple text field, we've observed 100+ FPS boost. It's hard to tell at this stage what real world performance will be seen.

## Future Work

- Only pay for `font_ref.outline_glyphs()` when we actually need it or use an acceleration data structure to generate it.
- Instead of storing outline paths in the outline cache, store translatable strips (to totally bypass strip rendering).


LaurenzV:
I haven't looked at this in detail yet, but wouldn't it be possible to do the eviction when calling `reset`?

taj-p:
> I haven't looked at this in detail yet, but wouldn't it be possible to do the eviction when calling `reset`?

Oooo. Much better, yes. Let's do that! Thank you @LaurenzV 

edit: Moved in https://github.com/linebender/vello/pull/1215/commits/5e7dc730171aa883f47f93cdb41cfa0fb74f49ac

nicoburns:
This definitely seems to help for Blitz rendering it's own readme. Kinda hard to tell how much by with my very unscientific method of eyeballing frame times, but I'd say it's maybe 20% faster (end-to-end time for rendering the entire scene) for Blitz rendering it's own README (scrolled to no images in view) with `vello_cpu`. Absolutely numbers are something like 3.9ms rather than 4.9ms.

Looking forward to having an actual full glyph cache!

Scene being tested:

Screenshot 2025-09-15 at 17 07 14

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -623,3 +656,331 @@ mod tests {
...
+        size: Size,
+        var_key: VarLookupKey<'_>,
+        outline_glyph: &skrifa::outline::OutlineGlyph<'_>,
+        hinting_instance: Option<&HintingInstance>,
+    ) -> &OutlinePath {
+        let key = OutlineKey {
+            glyph_id,
+            font_id,
+            font_index,
+            size_bits: size.ppem().unwrap().to_bits(),
```
grebmeg, line 836: Could tiny floating-point precision differences when using `to_bits` cause producing cache misses?

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -623,3 +656,331 @@ mod tests {
...
+    // Split caches for glyf/cff because the instance type can reuse
+    // internal memory when reconfigured for the same format.
+    glyf_entries: Vec<HintEntry>,
+    cff_entries: Vec<HintEntry>,
+    serial: u64,
+}
+
+impl Debug for HintCache {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "HintCache")
```
grebmeg: Would it make sense to add more debugging information here?

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -255,29 +272,37 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
 
             render_glyph(self.renderer, prepared_glyph);
         }
+
+        self.renderer.restore_glyph_caches(GlyphCaches {
+            outline_cache,
+            hinting_cache,
+        });
```
grebmeg, line 281: Would it make sense to restore the cache on `drop`, since a panic could occur between take/restore, leaving the cache permanently lost?

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -623,3 +656,331 @@ mod tests {
...
+            DrawSettings::unhinted(size, var_key.0)
+        };
+
+        path.0.truncate(0);
+        outline_glyph.draw(draw_settings, &mut path).unwrap();
+
+        let entry = self
+            .map
+            .entry(key)
+            .or_insert(OutlineEntry::new(path, self.serial));
```
grebmeg: Should it just be this? 
```rust 
.insert(key, OutlineEntry::new(path, self.serial));
```

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -623,3 +656,331 @@ mod tests {
...
+        if self.map.contains_key(&key) {
+            return self
+                .map
+                .get_mut(&key)
+                .map(|entry| {
+                    entry.serial = self.serial;
+                    &entry.path
+                })
+                .unwrap();
+        }
```
grebmeg: It’s totally up to you, but would it make sense to use pattern matching here?

```rust
match self.map.entry(key) {
  Entry::Occupied(mut entry) => {
    entry.get_mut().serial = self.serial;
    &entry.into_mut().path
  }
  Entry::Vacant(vacant_entry) => {
    // entry insertion logic
   }
}
```

sparse_strips/vello_bench/src/glyph.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut builder = layout_cx.ranged_builder(&mut font_cx, text, scale, true);
+        builder.push_default(FontFamily::parse("Roboto").unwrap());
+        let mut layout: Layout<Brush> = builder.build(text);
+        let max_advance = Some(WIDTH as f32);
+        layout.break_all_lines(max_advance);
+        layout.align(max_advance, Alignment::Start, AlignmentOptions::default());
+        layout
+    };
+
+    g.bench_function("cached", |b| {
```
grebmeg: I think it would also be useful to track cache performance metrics like hit rate, miss rate, etc. That way, when benchmarking different layout configurations, you can see how they affect performance, also possibly by tweaking some cache parameters as well.

sparse_strips/vello_bench/src/glyph.rs
```diff
@@ -0,0 +1,174 @@
...
+            x: glyph_x,
+            y: glyph_y,
+        }
+    });
+
+    let run = glyph_run.run();
+    renderer
+        .glyph_run(run.font())
+        .font_size(run.font_size())
+        .hint(true)
```
grebmeg: Maybe it’s worth splitting the benchmarks into runs with hinting enabled and disabled.

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -623,3 +656,331 @@ mod tests {
...
+    }
+    if entries.len() < MAX_CACHED_HINT_INSTANCES {
+        let instance = key.instance()?;
+        let ix = entries.len();
+        entries.push(HintEntry {
+            font_id: key.font_id,
+            font_index: key.font_index,
+            instance,
+            serial: 0,
```
grebmeg, line 984: Looks like this should be the current serial (updated on line 931). Maybe worth adding a short comment to avoid confusion?

sparse_strips/vello_common/src/glyph.rs
```diff
@@ -623,3 +656,331 @@ mod tests {
...
+            // 3. use unsafe
+            // Pick 1 bad option :(
+            if outline_cache.variable_map.contains_key(&var_key) {
+                outline_cache.variable_map.get_mut(&var_key).unwrap()
+            } else {
+                outline_cache
+                    .variable_map
+                    .entry(var_key.into())
+                    .or_default()
+            }
```
grebmeg: Would this approach work?

```rust
match outline_cache
  variable_map
  .raw_entry_mut()
  .from_key(&var_key)
{
    RawEntryMut::Occupied(entry) => entry.into_mut(),
    RawEntryMut::Vacant(entry) => entry.insert(var_key.into(), HashMap::new()).1,
}
```

grebmeg (approved):
LGTM! 🎉 

Consider my suggestions as advisory, I didn’t find any major issues with the current approach!

sparse_strips/vello_common/src/glyph.rs
taj-p: If two runs had their sizes calculated by different derivations this could happen, yes!

sparse_strips/vello_common/src/glyph.rs
taj-p: I'd like to use `entry` so that the inserted entry is returned in order to return `&entry.path` without suffering another hashmap lookup. `.insert` returns an optional value of what was maybe replaced by the new value

sparse_strips/vello_bench/src/glyph.rs
taj-p: Added TODO for this 🙏

sparse_strips/vello_common/src/glyph.rs
taj-p: Wow! Excellent. Thank you

sparse_strips/vello_common/src/glyph.rs
taj-p: You're a genius @grebmeg . Thank you. This is a huge improvement. Implemented in https://github.com/linebender/vello/pull/1215/commits/8a7b5150475e4c679f1bf7002ec2dbe28bb0961d

sparse_strips/vello_bench/src/glyph.rs
taj-p: Good idea. Added in https://github.com/linebender/vello/pull/1215/commits/7fd20b36c60e6b0b774b30743ff1d4e38bb84022

image

sparse_strips/vello_common/src/glyph.rs
taj-p: Added in https://github.com/linebender/vello/pull/1215/commits/2d4e137cedf427d6b83d0a41af869997a0421fc1

sparse_strips/vello_common/src/glyph.rs
taj-p: Added in https://github.com/linebender/vello/pull/1215/commits/6558c7725ef1463aab0c32043a3176b83aea4b7d

sparse_strips/vello_common/src/glyph.rs
taj-p: Added TODO to consider this

PR #1216, opened 2025-09-14
Tighten bounds for direct curve-to-line-segment flattening

tomcur:
Draft, as this ideally uses https://github.com/linebender/kurbo/pull/505 and requires the comments to be updated. Wanted to test this on CI in the meantime.

This tightens the bounds of the early curve-to-line-segment flattening we've introduced in https://github.com/linebender/vello/pull/1214, by projecting the control points onto the line segment, rather than just checking the distance to the end points.

### DIFF
--- a/sparse_strips/vello_common/src/flatten_simd.rs
+++ b/sparse_strips/vello_common/src/flatten_simd.rs
@@ -8,7 +8,7 @@
 use crate::flatten::TOL_2;
 #[cfg(not(feature = "std"))]
 use crate::kurbo::common::FloatFuncs as _;
-use crate::kurbo::{CubicBez, ParamCurve, PathEl, Point, QuadBez};
+use crate::kurbo::{CubicBez, Line, ParamCurve, PathEl, Point, QuadBez};
 use alloc::vec;
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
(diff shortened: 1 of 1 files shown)


LaurenzV:
image
image
image
image
image

tomcur:
Nice, thanks for the benchmarks! As I suspected, this helps especially bring down the timings on some of the larger geometry, too (relative to https://github.com/linebender/vello/pull/1214, which impacted smaller geometry especially), as all Beziers that are nearly lines with control points near the interior should now be detected; not just those with small control point lengths.

I've run the tests locally (CPU-only, GPU crashes for me), and there are 63 failures, all on `glyph::` or `recording::`. Will look into the failures more closely later, but it may well be that these are brushing up against the tolerance of `0.25` pixels, causing some coverage differences.

LaurenzV:
This seems to be the failure for the "glyphs_filled" test. Previously, we generated two segments, now it's just one:
`QuadBez { p0: (50.26953125, 46.0), p1: (52.66015625, 46.0), p2: (54.31640625, 45.140625) }`. At least intuitively, it does seem like it should be within the threshold.

LaurenzV:
I've run this through the PDF test suite and there are pixel differences in nearly all images (highlighting how common such small curves are!) but nothing that stands out as a regression. So, assuming this is mathematically valid, seems good to me.

tomcur:
> This seems to be the failure for the "glyphs_filled" test. Previously, we generated two segments, now it's just one: `QuadBez { p0: (50.26953125, 46.0), p1: (52.66015625, 46.0), p2: (54.31640625, 45.140625) }`. At least intuitively, it does seem like it should be within the threshold.

The max distance between that quadratic and the line segment `[p0,p2]` is approximately 0.2483 and the current tolerance is 0.25, so that's right around the threshold. Makes sense that it got subdivided before, as even a slight overestimation in error would bump it over.

LaurenzV:
Not sure if someone with a better maths background should double-check it, but I would be fine re-running my tests on them tomorrow and then approving.

LaurenzV (approved):
Gonna have to trust you on the math, but otherwise LGTM, especially the speed bump. Also reran it against my PDF tests and not noticing anything off.

PR #1217, opened 2025-09-15
fix transform handling in sparse_strips recording API

ajakubowicz-canva:
### Context

I was investigating using the recording API ran into a rough edge where the transform was not correctly captured by the recording.

This PR enforces that transforms are handled in the recording, and adds more tests to make the recorder API usage clearer. 

## Change

https://github.com/linebender/vello/blob/418803ed8420f569b0687b72d5d33d363b90c3d5/sparse_strips/vello_common/src/recording.rs#L426-L430

https://github.com/linebender/vello/blob/418803ed8420f569b0687b72d5d33d363b90c3d5/sparse_strips/vello_common/src/recording.rs#L364-L368

 - `Recorder::glyph_run` used `self.recording.transform` eagerly, without the `recording.transform` ever being set. This resulted in the glyph_run unconditionally using `Affine::IDENTITY`. This is fixed in this PR by setting the transform eagerly whenever `set_transform` is called on the recorder.

However, glyph_run's transform could still be wrong if a `set_transform` was not within the recording. For example the following code would set a transform on the rendering context, but then the glyph_run would be built at `Affine::IDENTITY`:

```rs
    ctx.set_transform(Affine::translate((0., f64::from(font_size))));

    let mut recording = Recording::new();
    ctx.record(&mut recording, |ctx| {
        ctx.set_paint(REBECCA_PURPLE.with_alpha(0.5));
        ctx.glyph_run(&font)
            .font_size(font_size)
            .hint(true)
            .fill_glyphs(glyphs.into_iter());
    });
```

This was fixed by modifying the `record` trait such that it is no longer built in. Now it eagerly pushes a `set_transform` such that the transform of the recording is tracked. This correctly builds the glyphs.

Fix: Recording is now correctly absolutely positioned when it is replayed because it _always_ pushes a set_transform on recording start.

### Test plan

Added multiple unique test cases and also checked the examples.

Results from `cargo run -p vello_hybrid_winit --release`:

```sh
# GhostTiger
SVG  | Fresh     : 12.040ms | Strips: 6850 | Alphas: 308048
SVG  | Identical : 0.957ms | Strips: 6850 | Alphas: 308048
SVG  | Identical : 0.483ms | Strips: 6850 | Alphas: 308048
# Text
Text | Fresh     : 0.693ms | Strips: 89 | Alphas: 2896
Text | Fresh     : 0.215ms | Strips: 89 | Alphas: 2896
Text | Fresh     : 0.212ms | Strips: 91 | Alphas: 3248
Text | Identical : 0.004ms | Strips: 91 | Alphas: 3248
Text | Identical : 0.015ms | Strips: 91 | Alphas: 3248
```

Results from `cargo run_wasm -p native_webgl --release`:
 - Checked performance tab and you can't even see the `request_animation_frame` that contains the cached render 😆 .

### Caveats

Note – recording API still crashes if the sparse strips go off the top left of the screen. That behavior remains consistent with `main`.

This PR fixes the transform handling of the recording – maintaining the expected behavior that the recording will be replayed at the transforms that it was recorded at. In the future we can extend this API to add a translation when we support it. Or we can add a new API that executes a recording at a translation.

### DIFF
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -158,6 +158,11 @@ impl Recording {
         }
     }
 
+    /// Return the initial transform of the recording.
+    pub fn relative_transform(&self) -> Option<&Affine> {
+        self.relative_transform.as_ref()
+    }
+
     /// Set the transform.
     pub(crate) fn set_transform(&mut self, transform: Affine) {
         if self.relative_transform.is_none() {
@@ -171,7 +176,15 @@ impl Recording {
     /// can be relatively transformed.
     pub fn enforce_matching_transform(&self, transform: &Affine) {
         let relative_transform = self.relative_transform.unwrap();
-        assert!(relative_transform.eq(transform), "renderer must set_transform to match recording before executing recording")
+        let a_coeffs = relative_transform.as_coeffs();
+        let b_coeffs = transform.as_coeffs();
+        let tolerance = 1e-6;
+
+        for i in 0..6 {
+            if (a_coeffs[i] - b_coeffs[i]).abs() > tolerance {
+                panic!("renderer must set_transform to match recording before executing recording")
+            }
+        }
     }
 
     /// Get commands as a slice.
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -174,11 +174,14 @@ impl Recording {
     /// Until recording can be relatively positioned, enforce that renderer set_transform matches
     /// the recording's initial conditions. This will make forward migration easier once recording
     /// can be relatively transformed.
-    /// 
+    ///
     /// TODO: This can be removed when relative transform of recording is implemented.
     pub fn enforce_matching_transform(&self, transform: &Affine) {
         let relative_transform = self.relative_transform.unwrap();
-        assert_eq!(relative_transform, *transform, "renderer must set_transform to match recording before executing recording");
+        assert_eq!(
+            relative_transform, *transform,
+            "renderer must set_transform to match recording before executing recording"
+        );
     }
 
     /// Get commands as a slice.
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -167,7 +167,7 @@ impl Recording {
         self.transform = transform;
     }
 
-    /// Until recording can be relatively positioned, enforce that renderer set_transform matches
+    /// Until recording can be relatively positioned, enforce that renderer `set_transform` matches
     /// the recording's initial conditions. This will make forward migration easier once recording
     /// can be relatively transformed.
     ///
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -174,10 +174,16 @@ impl Recording {
     /// TODO: This can be removed when relative transform of recording is implemented.
     pub fn enforce_matching_transform(&self, transform: &Affine) {
         let relative_transform = self.relative_transform.unwrap();
-        assert_eq!(
-            relative_transform, *transform,
-            "renderer must set_transform to match recording before executing recording"
-        );
+        // Although Affine implements `eq`, WASM tests require this more flexible check.
+        let a_coeffs = relative_transform.as_coeffs();
+        let b_coeffs = transform.as_coeffs();
+        let tolerance = 1e-6;
+
+        for i in 0..6 {
+            if (a_coeffs[i] - b_coeffs[i]).abs() > tolerance {
+                panic!("renderer must set_transform to match recording before executing recording")
+            }
+        }
     }
 
     /// Get commands as a slice.
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -174,17 +174,11 @@ impl Recording {
     /// Until recording can be relatively positioned, enforce that renderer set_transform matches
     /// the recording's initial conditions. This will make forward migration easier once recording
     /// can be relatively transformed.
+    /// 
+    /// TODO: This can be removed when relative transform of recording is implemented.
     pub fn enforce_matching_transform(&self, transform: &Affine) {
         let relative_transform = self.relative_transform.unwrap();
-        let a_coeffs = relative_transform.as_coeffs();
-        let b_coeffs = transform.as_coeffs();
-        let tolerance = 1e-6;
-
-        for i in 0..6 {
-            if (a_coeffs[i] - b_coeffs[i]).abs() > tolerance {
-                panic!("renderer must set_transform to match recording before executing recording")
-            }
-        }
+        assert_eq!(relative_transform, *transform, "renderer must set_transform to match recording before executing recording");
     }
 
     /// Get commands as a slice.
--- a/sparse_strips/vello_common/src/recording.rs
+++ b/sparse_strips/vello_common/src/recording.rs
@@ -88,8 +88,15 @@ pub struct Recording {
     commands: Vec<RenderCommand>,
     /// Cached sparse strips.
     cached_strips: CachedStrips,
-    /// Last recorded transform.
+    /// Track the transform of the underlying context.
     transform: Affine,
+    /// The transform which the recording is relative to.
+    /// This is the initial transform of the renderer when the recording was initialized.
+    ///
+    /// TODO: Changing the transform on the renderer and executing the recording should execute the
+    /// recording relative to the new transform. Allowing you to "stamp" the recording multiple
+    /// times at different transforms.
+    relative_transform: Option<Affine>,
 }
 
 /// Command for pushing a new layer.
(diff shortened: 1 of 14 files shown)


LaurenzV:
> Crash on non matching transforms

I think in the long term, this will be fixed my making the API not stateful and then simply not allowing applying a transform to a recording directly.

ajakubowicz-canva:
@LaurenzV I'm interested in further elaboration because my (very light understanding) from @grebmeg was that in the future we want the Recording's transform to be relative the recorded transform. The idea (as I understand) is so a translation can reuse the cached sparse strips when executing a recording.

I am not sure I understand how the API can be made not stateful as the recording is capturing the renderers command state? Very keen to understand that better. Thank you!

LaurenzV:
Sorry, I realize now that my explanation was a bit too brief. 😅  Perhaps I'm also out of the loop or misunderstanding, but AFAIK, in the future `set_transform` will be removed so that `RenderContext` doesn't store any transform. Instead, methods like `filll_path` and `stroke_path` will take an additional `transform` attribute that will apply to just that path. Therefore, there is no transform the recording would need to be relative to, and instead there probably would need to be a custom method like `draw_recording(recording, x_shift, y_shift)` that allows replaying a recording with a specific shift. But perhaps I'm wrong!

sparse_strips/vello_common/src/recording.rs
```diff
@@ -147,15 +147,10 @@ impl Recording {
...
-    pub fn set_transform(&mut self, transform: Affine) {
+    pub(crate) fn set_transform(&mut self, transform: Affine) {
         self.transform = transform;
     }
 
-    /// Get the current transform.
-    pub fn transform(&self) -> Affine {
-        self.transform
-    }
-
```
ajakubowicz-canva, line 158: These were internal details and were removed or scoped.

sparse_strips/vello_common/src/recording.rs
```diff
@@ -276,11 +271,7 @@ pub trait Recordable {
     /// ```
     fn record<F>(&mut self, recording: &mut Recording, f: F)
     where
-        F: FnOnce(&mut Recorder<'_>),
-    {
-        let mut recorder = Recorder::new(recording);
-        f(&mut recorder);
```
ajakubowicz-canva, line 282: Default method implementation was removed as it now depends on the private `transform` state provided by vello_cpu or vello_hybrid. This ensures that recordings have their initial transform state captured.

sparse_strips/vello_sparse_tests/tests/recording.rs
```diff
@@ -95,3 +95,115 @@ fn recording_glyphs(ctx: &mut impl Renderer) {
...
+            .hint(true)
+            .fill_glyphs(glyphs.into_iter());
+    });
+
+    ctx.prepare_recording(&mut recording);
+
+    // Clear the canvas.
+    ctx.set_transform(Affine::IDENTITY);
+    ctx.set_paint(GREEN);
+    ctx.fill_rect(&Rect::new(0.0, 0.0, 300.0, 100.0));
```
grebmeg: Maybe clear half of the canvas just to show that `LIGHT_SALMON` is still visible?

sparse_strips/vello_sparse_tests/tests/recording.rs
```diff
@@ -95,3 +95,115 @@ fn recording_glyphs(ctx: &mut impl Renderer) {
...
+        ctx.fill_rect(&Rect::new(10.0, 10.0, 90.0, 90.0));
+    });
+    ctx.prepare_recording(&mut recording);
+
+    recording.clear();
+
+    ctx.set_transform(Affine::IDENTITY);
+    ctx.record(&mut recording, |ctx| {
+        ctx.set_paint(GREEN);
+        ctx.fill_rect(&Rect::new(10.0, 10.0, 90.0, 90.0));
```
grebmeg: To make it clear that nothing is behind the rect, maybe use a triangle instead?

```rust
ctx.record(&mut recording, |ctx| {
  ctx.set_paint(GREEN);
  let mut triangle_path = BezPath::new();
  triangle_path.move_to((50.0, 10.0));
  triangle_path.line_to((10.0, 90.0));
  triangle_path.line_to((90.0, 90.0));
  triangle_path.close_path();
  ctx.fill_path(&triangle_path);
});
```

sparse_strips/vello_common/src/recording.rs
taj-p: The default impl was also removed in https://github.com/linebender/vello/pull/1215

sparse_strips/vello_sparse_tests/tests/recording.rs
```diff
@@ -95,3 +95,115 @@ fn recording_glyphs(ctx: &mut impl Renderer) {
...
+
+    ctx.set_transform(Affine::IDENTITY);
+    ctx.execute_recording(&recording);
+}
+
+#[vello_test(width = 300, height = 100)]
+fn recording_mixed_with_direct_drawing(ctx: &mut impl Renderer) {
+    let mut recording = Recording::new();
+    // Record a rectangle on the left.
+    ctx.set_transform(Affine::translate((20.0, 30.0)));
```
grebmeg, line 142: Can we add another test with multiple `set_transform` calls, both outside and inside record contexts? Something like this:
```rust
#[vello_test(width = 50, height = 50)]
fn recording_is_executed_with_multiple_transforms(ctx: &mut impl Renderer) {
    ctx.set_transform(Affine::translate((15., 15.)));

    let font_size: f32 = 10_f32;
    let (font, glyphs) = layout_glyphs_roboto("A", font_size);
    let mut recording = Recording::new();
    ctx.record(&mut recording, |ctx| {
        ctx.set_paint(GOLD);
        ctx.fill_rect(&Rect::new(0.0, 0.0, 5.0, 5.0));
        ctx.set_paint(FUCHSIA);
        ctx.set_transform(Affine::translate((20., 20.)));
        ctx.fill_rect(&Rect::new(0.0, 0.0, 5.0, 5.0));
        ctx.set_paint(GREEN);
        ctx.set_transform(Affine::translate((25., 25.)));
        ctx.fill_rect(&Rect::new(0.0, 0.0, 5.0, 5.0));

        ctx.set_transform(Affine::translate((5.0, 15.0)));
        ctx.set_paint(ORANGE);
        ctx.glyph_run(&font)
            .font_size(font_size)
            .hint(true)
            .fill_glyphs(glyphs.into_iter());
    });

    ctx.set_transform(Affine::translate((30., 30.)));
    ctx.set_paint(PALE_VIOLET_RED);
    ctx.fill_rect(&Rect::new(0.0, 0.0, 5.0, 5.0));

    ctx.set_transform(Affine::translate((35., 35.)));
    ctx.set_paint(PURPLE);
    ctx.fill_rect(&Rect::new(0.0, 0.0, 5.0, 5.0));
}
```

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -395,6 +395,13 @@ impl GlyphRenderer for Scene {
 }
 
 impl Recordable for Scene {
+    fn record<F>(&mut self, recording: &mut Recording, f: F)
```
grebmeg, line 398: If you’re thinking of adding a `transform` here to make it less implicit, I actually think the implicitness is helpful in some cases, the consumer doesn’t need to worry about which transform to pass. Maybe we could consider adding a `record_with_transform` method to support the explicit case instead?

sparse_strips/vello_common/src/recording.rs
```diff
@@ -335,8 +326,11 @@ pub struct Recorder<'a> {
 
 impl<'a> Recorder<'a> {
     /// Create a new recorder for the given recording.
-    pub fn new(recording: &'a mut Recording) -> Self {
-        Self { recording }
+    pub fn new(recording: &'a mut Recording, transform: Affine) -> Self {
+        let mut s = Self { recording };
+        // Ensure that the initial transform is saved on the recording.
+        s.set_transform(transform);
```
grebmeg, line 332: Would it make sense to move `set_transform` into `record`? Something like this:
```rust
fn record<F>(&mut self, recording: &mut Recording, f: F)
where
  F: FnOnce(&mut Recorder<'_>),
{
  let mut recorder = Recorder::new(recording, self.transform);
  recorder.set_transform(self.transform);
  f(&mut recorder);
}
```

sparse_strips/vello_hybrid/src/scene.rs
ajakubowicz-canva: This is another good call-out and strengthens my thought that the API discussion should sit as a followup to this correctness PR.

sparse_strips/vello_sparse_tests/tests/recording.rs
ajakubowicz-canva: Great suggestion – done in https://github.com/linebender/vello/pull/1217/commits/35a2e957b9e3e176f187380417d148dd6565cd02#diff-d069040caec58466f6c1d0bbee83acd98f8eff7280f71040f09f2f590f51b5caR207-R212

sparse_strips/vello_sparse_tests/tests/recording.rs
ajakubowicz-canva: Great point – done in https://github.com/linebender/vello/pull/1217/commits/35a2e957b9e3e176f187380417d148dd6565cd02#diff-d069040caec58466f6c1d0bbee83acd98f8eff7280f71040f09f2f590f51b5caR167-R170

sparse_strips/vello_sparse_tests/tests/recording.rs
ajakubowicz-canva: Additional test added: https://github.com/linebender/vello/pull/1217/commits/35a2e957b9e3e176f187380417d148dd6565cd02#diff-d069040caec58466f6c1d0bbee83acd98f8eff7280f71040f09f2f590f51b5caR218-R252

sparse_strips/vello_common/src/recording.rs
ajakubowicz-canva: I prefer as written currently because it allows the `set_transform` to have a more constrained visibility. Of course I think this can be changed when we revisit the API.

Great comment!

PR #1220, opened 2025-09-15
Fix wrong calculation in alpha compositing

LaurenzV:
Turns out none of our tests actually covered the fast path for simple alpha compositing...


ajakubowicz-canva:
Awesome! Thanks Laurenz!

PR #1221, opened 2025-09-15
Change the direction of the angle in sweep gradients

LaurenzV:
In preparation for the next peniko release.

### DIFF
--- a/sparse_strips/vello_common/src/colr.rs
+++ b/sparse_strips/vello_common/src/colr.rs
@@ -325,7 +325,7 @@ impl ColorPainter for ColrPainter<'_> {
                     extend,
                     ..Default::default()
                 };
-                
+
                 let paint_transform = self.cur_transform() * Affine::scale_non_uniform(1.0, -1.0);
 
                 self.painter.set_paint_transform(paint_transform);
--- a/sparse_strips/vello_common/src/colr.rs
+++ b/sparse_strips/vello_common/src/colr.rs
@@ -313,9 +313,11 @@ impl ColorPainter for ColrPainter<'_> {
                     }
                 }
 
+                // We need to invert the direction of the gradient to bridge the gap between
+                // peniko and COLR.
                 let grad = Gradient {
                     kind: GradientKind::Sweep {
-                        center: p0,
+                        center: Point::new(p0.x, -p0.y),
                         start_angle,
                         end_angle,
                     },
@@ -323,8 +325,10 @@ impl ColorPainter for ColrPainter<'_> {
                     extend,
                     ..Default::default()
                 };
+                
+                let paint_transform = self.cur_transform() * Affine::scale_non_uniform(1.0, -1.0);
 
-                self.painter.set_paint_transform(self.cur_transform());
+                self.painter.set_paint_transform(paint_transform);
                 self.painter.fill_gradient(grad);
             }
         };
--- a/sparse_strips/vello_cpu/src/fine/common/gradient/sweep.rs
+++ b/sparse_strips/vello_cpu/src/fine/common/gradient/sweep.rs
@@ -26,8 +26,7 @@ impl<S: Simd> SimdSweepKind<S> {
 impl<S: Simd> SimdGradientKind<S> for SimdSweepKind<S> {
     #[inline(always)]
     fn cur_pos(&self, x_pos: f32x8<S>, y_pos: f32x8<S>) -> f32x8<S> {
-        let angle = x_y_to_unit_angle(self.simd, x_pos, y_pos * f32x8::splat(self.simd, -1.0))
-            * f32x8::splat(self.simd, 2.0 * PI);
+        let angle = x_y_to_unit_angle(self.simd, x_pos, y_pos) * f32x8::splat(self.simd, 2.0 * PI);
 
         (angle - self.start_angle) * self.inv_angle_delta
     }
@@ -58,5 +57,5 @@ fn x_y_to_unit_angle<S: Simd>(simd: S, x: f32x8<S>, y: f32x8<S>) -> f32x8<S> {
     // Clears all NaNs, using the property that NaN != NaN.
     phi = simd.select_f32x8(phi.simd_eq(phi), phi, c0);
 
-    1.0 - phi
+    phi
 }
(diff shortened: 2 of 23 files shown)


sparse_strips/vello_common/src/colr.rs
```diff
@@ -313,9 +313,11 @@ impl ColorPainter for ColrPainter<'_> {
...
                 }
 
+                // We need to invert the direction of the gradient to bridge the gap between
+                // peniko and COLR.
                 let grad = Gradient {
                     kind: GradientKind::Sweep {
-                        center: p0,
+                        center: Point::new(p0.x, -p0.y),
                         start_angle,
                         end_angle,
```
DJMcNab, line 322: So, to clarify: The COLR spec specifies counter-clockwise in y-up. Skrifa converts this to clockwise in y-up, and then we need to convert it *again* into clockwise in y-down (which is the same as the first of these). cc @dfrg, the conversion to clockwise in y-up happens at https://github.com/googlefonts/fontations/blob/1b5175d4ddb33f10a3ddfb64a825ae7ccd74b0d6/skrifa/src/color/traversal.rs#L439-L446; it's not clear to me what "for the shader" means here.

And so the way that you've converted it into clockwise in y-down is to treat it as clockwise in y-up, then invert it. But ideally, we'd just skip that first conversion here.

sparse_strips/vello_cpu/src/fine/common/gradient/sweep.rs
```diff
@@ -58,5 +58,5 @@ fn x_y_to_unit_angle<S: Simd>(simd: S, x: f32x8<S>, y: f32x8<S>) -> f32x8<S> {
     // Clears all NaNs, using the property that NaN != NaN.
     phi = simd.select_f32x8(phi.simd_eq(phi), phi, c0);
 
-    phi
+    1.0 - phi
```
DJMcNab: This *feels* suspect to me; I think the correct fix might be to remove the `-1 * y` at the call site of this function.

LaurenzV: Oops yeah, you are right!

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -412,7 +412,7 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
             
             // For sweep gradient, calculate angle from center using fast polynomial approximation
             // Flip Y coordinate for screen coordinate system (Y increases downward)
```
DJMcNab: Is this what we've just removed?

tomcur: I believe so, though I'm not entirely sure what the original wording means. I think it's comparing the screen coordinate system to an implicit Y-up coordinate system for the gradients (in which case we've now dropped that).

sparse_strips/vello_common/src/colr.rs
```diff
@@ -324,7 +326,9 @@ impl ColorPainter for ColrPainter<'_> {
                     ..Default::default()
                 };
 
-                self.painter.set_paint_transform(self.cur_transform());
+                let paint_transform = self.cur_transform() * Affine::scale_non_uniform(1.0, -1.0);
```
tomcur, line 329: This inverse affine is somewhat unfortunate, but I don't think there's a great way around it, as I believe we require gradients to have positive angle delta (`end_angle > start_angle`). Otherwise we'd be able to do the following to flip the gradient direction.

```
                let grad = Gradient {
                    kind: GradientKind::Sweep {
                        center: p0,
                        start_angle: -start_angle,
                        end_angle: -end_angle,
                    },
                    stops,
                    extend,
                    ..Default::default()
                };
```

LaurenzV: Yes, we require positive deltas unfortunately.

sparse_strips/vello_common/src/colr.rs
tomcur: I believe if we drop all intermediate flipping, only the following final transform in the renderer/shader would be required.

https://github.com/linebender/vello/blob/f39593e8a42868a789630eb4ff99ecc5c50e17a9/sparse_strips/vello_common/src/glyph.rs#L424-L437

tomcur (approved):
The changes to `sweep.rs` look good to me.

For COLR, I believe the intermediate transform flipping the Y axis for the gradients is the right way to go about it for now, as Skrifa flips the gradient direction (see these comments: https://github.com/linebender/vello/pull/1221#discussion_r2348897527). As we require positive angle deltas in the sweep gradient, doing the inverse of Skrifa's transform would require flipping color stops (as Skrifa also conditionally does), which would be more expensive.

sparse_strips/vello_common/src/colr.rs
DJMcNab: > I believe if we drop all intermediate flipping, only the following final transform in the renderer/shader would be required.

Just to confirm, do you mean, if Skrifa dropped their intermediate flipping, or something else?

tomcur: Yes, my message wasn't very clear. I think if Skrifa drops the gradient direction flipping, then we need to drop the gradient direction flipping, and I think that would only leave the final (required) glyph transform from Y-up to Y-down. (That final transform should probably remain in the renderer, as it is now).

It would be good to try this out with local changes to Skrifa removing those `360 - angle` transforms (still doing the angle and stop reversal in case of  `start_angle > end_angle`, to keep positive angle delta).

tomcur: (I'll try that suggestion out ^)

tomcur: Alright, I've wired it up.

Vello sparse: https://github.com/tomcur/vello/commit/0eb5a083aa9de87ba13e4a734f5cdee8510b92c5
on this commit of Skrifa: https://github.com/tomcur/fontations/commit/50a4ca5a003abd6516fdef54d91dbce6845e7526

It appears to work mostly as expected, though there are some pixel differences for the `colr_test_glyphs` tests (see the following image).

The most surprising thing is that there appears to be some funkiness going on here with the "remove multiple final stops" stuff of COLR, I had to make the following change to get the same results https://github.com/tomcur/vello/commit/0eb5a083aa9de87ba13e4a734f5cdee8510b92c5. I don't have time this evening to figure out whether the behavior we're testing for is correct or incorrect.

glyphs_colr_test_glyphs_cpu_f32_avx2

PR #1222, opened 2025-09-16
[sparse_strips] fix out of bounds recording crash

ajakubowicz-canva:
### Context

Previously, recording crashes when executing strips that are out of bounds.

https://github.com/user-attachments/assets/f3409cb4-b1b2-45a9-8703-b157ac3ad38d

After this PR – out of bounds strips are handled:

https://github.com/user-attachments/assets/6fe68444-157c-4fca-b84f-7f76952348b5

### Why?

`execute_recording`:
 - Iterates over the recording commands, unconditionally calling `process_geometry_command` on the strips.
 - Except – for commands which are fully offscreen there are no strips to process.

Previously this would cause a crash. The new behavior is we handle an empty count of strips – meaning there is nothing to render for this command.

### Test plan

Added a test which renders completely outside the viewport. It used to crash and no longer crashes. I also tested manually.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -859,6 +859,10 @@ impl RenderContext {
             .copied()
             .unwrap_or(adjusted_strips.len());
         let count = end - start;
+        if count == 0 {
+            // There are no strips to generate.
+            return;
+        }
         assert!(
             start < adjusted_strips.len() && count > 0,
             "Invalid strip range"
--- a/sparse_strips/vello_sparse_tests/tests/recording.rs
+++ b/sparse_strips/vello_sparse_tests/tests/recording.rs
@@ -207,3 +207,20 @@ fn recording_can_be_cleared(ctx: &mut impl Renderer) {
     ctx.prepare_recording(&mut recording);
     ctx.execute_recording(&recording);
 }
+
+#[vello_test(width = 20, height = 20, no_ref)]
+fn recording_handles_completely_offscreen_content(ctx: &mut impl Renderer) {
+    let mut recording = Recording::new();
+    ctx.record(&mut recording, |ctx| {
+        ctx.set_paint(ORCHID);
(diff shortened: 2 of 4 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -859,6 +859,10 @@ impl RenderContext {
             .copied()
             .unwrap_or(adjusted_strips.len());
         let count = end - start;
+        if count == 0 {
+            // There are no strips to generate.
+            return;
+        }
```
grebmeg, line 865: I have a feeling this could be fixed in a slightly different way by not pushing indices to `strip_start_indices` in `generate_strips_from_commands` when no strips are generated, so we don’t waste memory. But then we’d need a way to skip commands in `execute_recording` that generate empty strips. Though, not sure it’s worth the extra complexity right now, unless it’s easy to implement this way. I’m okay with these changes being merged.

sparse_strips/vello_sparse_tests/tests/recording.rs
```diff
@@ -207,3 +207,20 @@ fn recording_can_be_cleared(ctx: &mut impl Renderer) {
...
+        ctx.set_transform(Affine::translate((-200., -200.)));
+        ctx.fill_rect(&Rect::new(0.0, 0.0, 10.0, 10.0));
+        // A rectangle completely outside the viewport in then positive x and y direction.
+        ctx.set_transform(Affine::translate((200., 200.)));
+        ctx.fill_rect(&Rect::new(0.0, 0.0, 10.0, 10.0));
+    });
+
+    ctx.prepare_recording(&mut recording);
+    ctx.execute_recording(&recording);
+}
```
grebmeg, line 226: No newline at end of file

sparse_strips/vello_cpu/src/render.rs
AndrewJakubowicz: Thank you! I'll go ahead and leave it this way for now. This unblocks usage for us and is simple. 
I agree that there's a better solution.

Issue #1225, opened 2025-09-16
Disabling `default-features` for `wgpu` dependency in `Cargo.toml`?


StT191:
Why does `vello` not disable `default-features` for it's `wgpu` dependency?

As a consumer of the `vello` library I would rather decide on what features of `wgpu` I want to exclude in my application rather than being forced to include all the default features by `vello`. More specifically `wgpu` allows me to select which graphics backends I would like to support via features.

Please consider setting `default-features = false` for the `wgpu` dependency in `Cargo.toml` of the workspace!

DJMcNab:
Having wgpu's default features enabled by default is somewhat forced by #591; if we require you to enable wgpu's default features yourself, then there's no point providing that re-export.
We also should not create features for each of wgpu's features; it's impractical to keep up with that over time.

However, I agree that we should support not setting wgpu's default features, for users who require that. As such, I propose that our features look like:
```toml
[features]
default = [..., "wgpu_default"]
wgpu_default = ["wgpu/default", "wgpu"]
wgpu = ["dep:wgpu"]
```

A PR to implement this would be welcome!

StT191:
I agree, putting the wgpu-defaults behind a feature that is enabled in vello's default feature would be the best compromise as it would keep current defaults.

I'll prepare a [PR](https://github.com/linebender/vello/pull/1229) for this.

PR #1229, opened 2025-09-17
Put wgpu's default features behind a 'wgpu_default' feature flag

StT191:
This PR enables users of `vello` to disable `wgpu`'s default features (see [#1225](https://github.com/linebender/vello/issues/1225)). This is especially useful as `wgpu` allows enabling and disabling of all backends via feature flags since version `25.0.0`.

The previous behavior is behind an new feature flag `wgpu_default` that is included in `vellos`'s default features so current users of `vello` who rely on them don't get surprised.

### DIFF
--- a/sparse_strips/vello_hybrid/Cargo.toml
+++ b/sparse_strips/vello_hybrid/Cargo.toml
@@ -51,7 +51,7 @@ vello_common = { workspace = true, features = ["pico_svg"] }
 roxmltree = "0.20.0"
 
 [features]
-default = ["wgpu", "wgpu/default"]
+default = ["wgpu", "wgpu_default"]
 std = ["wgpu", "vello_common/std"]
 wgpu = ["dep:wgpu", "dep:vello_sparse_shaders"]
 wgpu_default = ["wgpu", "wgpu/default"]
--- a/sparse_strips/vello_hybrid/Cargo.toml
+++ b/sparse_strips/vello_hybrid/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 bytemuck = { workspace = true, features = ["derive"] }
 thiserror = { workspace = true }
 vello_common = { workspace = true, features = ["std", "text"] }
-wgpu = { workspace = true, optional = true }
+wgpu = { workspace = true, default-features = false, optional = true }
 vello_sparse_shaders = { workspace = true, optional = true }
 log = { workspace = true }
 guillotiere = "0.6.2"
@@ -51,7 +51,8 @@ vello_common = { workspace = true, features = ["pico_svg"] }
 roxmltree = "0.20.0"
 
 [features]
-default = ["wgpu"]
+default = ["wgpu", "wgpu/default"]
 std = ["wgpu", "vello_common/std"]
 wgpu = ["dep:wgpu", "dep:vello_sparse_shaders"]
+wgpu_default = ["wgpu", "wgpu/default"]
 webgl = ["dep:js-sys", "dep:web-sys", "dep:vello_sparse_shaders", "vello_sparse_shaders/glsl"]
--- a/sparse_strips/vello_hybrid/Cargo.toml
+++ b/sparse_strips/vello_hybrid/Cargo.toml
@@ -54,5 +54,8 @@ roxmltree = "0.20.0"
 default = ["wgpu", "wgpu_default"]
 std = ["wgpu", "vello_common/std"]
 wgpu = ["dep:wgpu", "dep:vello_sparse_shaders"]
+# Enable wgpu with its default features. If you need to customise the set of enabled wgpu features,
+# please disable this crate's default features, enable its "wgpu" feature, then depend on wgpu directly
+# with the features which you need enabled.
 wgpu_default = ["wgpu", "wgpu/default"]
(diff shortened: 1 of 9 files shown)


sparse_strips/vello_hybrid/Cargo.toml
```diff
@@ -51,7 +51,8 @@ vello_common = { workspace = true, features = ["pico_svg"] }
 roxmltree = "0.20.0"
 
 [features]
-default = ["wgpu"]
+default = ["wgpu", "wgpu/default"]
```
DJMcNab: These forms are equivalent (because we want to never actually use `cfg(feature="wgpu_default")`), but it's better to be consistent across crates.
```suggestion
default = ["wgpu", "wgpu_default"]
```

DJMcNab: cc @ajakubowicz-canva, does this make sense for you? This is potentially a breaking change (moreso for Vello Hybrid than Vello Classic, which it wasn't really meaningful to disable the default features on).

vello/Cargo.toml
```diff
@@ -16,13 +16,14 @@ default-target = "x86_64-unknown-linux-gnu"
...
 [features]
-default = ["wgpu"]
+default = ["wgpu", "wgpu_default"]
 # Enables GPU memory usage estimation. This performs additional computations
 # in order to estimate the minimum required allocations for buffers backing
 # bump-allocated GPU memory.
 # TODO: Turn this into a runtime option used at resolve time and remove the feature.
 bump_estimate = ["vello_encoding/bump_estimate"]
 wgpu = ["dep:wgpu", "dep:vello_shaders", "dep:futures-intrusive"]
+wgpu_default = ["wgpu", "wgpu/default"]
```
DJMcNab, line 29: Please can we add a short comment as docs here, e.g.:

```suggestion
# Enable wgpu with its default features. If you need to customise the set of enabled wgpu features,
# please disable this crate's default features, enable its "wgpu" feature, then depend on wgpu directly
# with the features which you need enabled.
wgpu_default = ["wgpu", "wgpu/default"]
```

vello/src/wgpu_engine.rs
```diff
@@ -769,7 +769,15 @@ impl WgpuEngine {
...
+                #[allow(
+                    clippy::drop_non_drop,
+                    reason = "suppress warning when no wgpu backend is enabled"
+                )]
                 drop(texture);
+                #[allow(
+                    clippy::drop_non_drop,
+                    reason = "suppress warning when no wgpu backend is enabled"
+                )]
                 drop(view);
```
DJMcNab: Can we just replace this whole thing with:
```rust
if let Some((_texture, _view)) = self.bind_map.image_map.remove(&id) {
    // TODO: have a pool to avoid needless re-allocation
}
```
I think the explicit drops were good before, but since we hit https://github.com/rust-lang/rust-clippy/issues/9332, it's better to just leave it explicit. `.remove` is already pretty clear.

sparse_strips/vello_hybrid/Cargo.toml
StT191: Sure, I'll change it in the next commit.

vello/Cargo.toml
StT191: I'll add it to the docs.

vello/src/wgpu_engine.rs
StT191: Agreed, removing the explicit drops looks much nicer than adding clippy attributes!

sparse_strips/vello_hybrid/Cargo.toml
StT191: I removed the feature inconsistency in an upcoming commit.

I'm not familiar with `vello_hybrid` and it's use cases myself. Should there be stricter feature requirements for wgpu here?

ajakubowicz-canva: I had a play with this locally and tested out what happens if you use default-features = false on vello-hybrid and _I think_ this is an improvement.

It does make it slightly harder to use `vello_hybrid` with `default-features = false` as you become responsible for activating your backend. But, as far as I understand this is desirable. The new `wgpu_default` also provides a workaround.

I'd be curious to better understand what sort of breaking change this could cause that is concerning (@DJMcNab ). Otherwise, this LGTM.

As a side-note, I think since vello_hybrid is experimental the breaking change is also alright.

DJMcNab: > I'd be curious to better understand what sort of breaking change this could cause that is concerning. Otherwise, this LGTM.

So this would be if someone were dependent on Vello Hybrid with `default-features=false` and the "wgpu" feature enabled. I just wanted to be sure that Canva was aware of this, so that we don't break you inadvertently.

ajakubowicz-canva: Thank you for the call out! LGTM from my end!

DJMcNab (approved):
Thank you! I've pushed a few final tweaks.

Let me know if you are ready for this to land (i.e. if you have any comment on my tweaks). 
I'll merge this soon if I don't hear from you.

StT191:
> Thank you! I've pushed a few final tweaks.
> 
> Let me know if you are ready for this to land (i.e. if you have any comment on my tweaks). I'll merge this soon if I don't hear from you.

I'm absolutely onboard with this. Great experience to get involved. 
Love to see it get merged soon.

PR #1233, opened 2025-09-18
Reuse allocations across multiple runs in multi-threading

LaurenzV:
@taj-p Thanks for the suggestion, has a nice performance impact!!

image
image
image
image

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -606,7 +606,7 @@ impl<T> AllocationManager<T> {
     ///
     /// The vector is guaranteed to have been cleared before.
     fn get(&mut self) -> Vec<T> {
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -600,11 +596,96 @@ impl Debug for MultiThreadedDispatcher {
...
+
+        AllocationGroup {
+            path,
+            render_tasks,
+            coarse_tasks,
+            strips,
+        }
+    }
+
+    fn put(&mut self, allocation: AllocationGroup) {
```
taj-p, line 660: nit: We could assert that we haven't allocated a huge number of stored allocations (to help detect leaks due to regression) or test that 1 send causes a `get` and 1 `run_coarse` causes a `put`

taj-p (approved):
Looks great to me! 🚀

PR #1236, opened 2025-09-19
refact: create `vello_example_scenes` crate with common examples

grebmeg:
I was adding some path examples from Vello Classic to validate visual correctness in Vello Sparse Strips and noticed we already have almost the same examples in `vello_cpu`. Along the way, I decided to consolidate them into a single set of example implementations. In the end, `RenderingContext` abstraction looks very similar to what’s used in `vello_sparse_tests`, so it might also make sense to unify those and rely on just one.


sparse_strips/vello_example_scenes/src/path.rs
```diff
@@ -0,0 +1,548 @@
...
+pub struct FunkyPathsScene {}
+
+impl FunkyPathsScene {
+    /// Create a new funky paths scene
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl ExampleScene for FunkyPathsScene {
```
grebmeg, line 195: I noticed that `FunkyPathsScene` example is rendering incorrectly. cc @LaurenzV


image

sparse_strips/vello_cpu/examples/wasm_cpu/Cargo.toml
```diff
@@ -36,10 +37,11 @@ web-sys = { version = "0.3.77", features = [
     "CanvasRenderingContext2d",
 ] }
 wgpu = { workspace = true, features = ["webgl"] }
+smallvec = { workspace = true }
+image = { workspace = true, features = ["jpeg"] }
```
T-256: did these expected add to here?

sparse_strips/vello_example_scenes/Cargo.toml
```diff
@@ -9,12 +9,20 @@ publish = false
...
+
 [dependencies]
 bytemuck = { workspace = true, features = [] }
 smallvec = { workspace = true }
 vello_hybrid = { workspace = true }
 vello_common = { workspace = true, features = ["pico_svg", "png"] }
 image = { workspace = true, features = ["jpeg"] }
+vello_cpu = { workspace = true, optional = true }
+console_error_panic_hook = "0.1.7"
+console_log = "1.0"
```
T-256: IIUC, the convention here is to use crate's dependencies from workspace specified dependencies, to avoid have multiple versions of dependencies.
```suggestion
console_error_panic_hook = { workspace = true }
console_log = { workspace = true }
```

sparse_strips/vello_cpu/examples/wasm_cpu/Cargo.toml
grebmeg: Thank you! It wasn’t expected to be here!

Removed.

sparse_strips/vello_example_scenes/Cargo.toml
grebmeg: Yes, that makes sense. It wasn’t added to the workspace Cargo.toml, and I overlooked adding it there. Thanks for pointing it out!

Fixed.

sparse_strips/vello_example_scenes/src/path.rs
grebmeg: Filed issue #1240

sparse_strips/vello_example_scenes/README.md
```diff
@@ -0,0 +1,13 @@
+# vello_example_scenes
+
+A collection of scenes used in Vello Hybrid examples.
```
ajakubowicz-canva:
```suggestion
A collection of scenes used for Vello examples.
```

Nit: Do not make this specific to hybrid, as examples are also used for CPU.

sparse_strips/vello_example_scenes/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+use vello_common::paint::ImageSource;
+pub use vello_common::paint::{Paint, PaintType};
+pub use vello_common::peniko::{BlendMode, Fill, FontData};
+use vello_common::recording::{Recordable, Recorder, Recording};
+#[cfg(feature = "cpu")]
+use vello_cpu::RenderContext;
+use vello_hybrid::Scene;
+
+/// A generic rendering context.
+pub trait RenderingContext: Sized + GlyphRenderer {
```
ajakubowicz-canva:
```suggestion
pub trait RenderingContext: Sized {
```

sparse_strips/vello_example_scenes/src/path.rs
ajakubowicz-canva: Can you link to the issue as a comment?

sparse_strips/vello_example_scenes/README.md
grebmeg: Ah, I missed that — thanks!

Fixed.

sparse_strips/vello_example_scenes/src/lib.rs
grebmeg: Thanks!

Fixed.

PR #1237, opened 2025-09-19
Add support for non-isolated masks

LaurenzV:
I also measured this using the Blend2D benchmark suite and it doesn't result in in any noticeable slowdown.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -235,7 +235,13 @@ impl<const MODE: u8> Wide<MODE> {
     ///    - Generate alpha fill commands for the intersected wide tiles
     /// 2. For active fill regions (determined by fill rule):
     ///    - Generate solid fill commands for the regions between strips
-    pub fn generate(&mut self, strip_buf: &[Strip], paint: Paint, thread_idx: u8) {
+    pub fn generate(
+        &mut self,
+        strip_buf: &[Strip],
+        paint: Paint,
+        thread_idx: u8,
+        mask: Option<Mask>,
+    ) {
         if strip_buf.is_empty() {
             return;
         }
@@ -307,6 +313,7 @@ impl<const MODE: u8> Wide<MODE> {
                     thread_idx,
                     paint: paint.clone(),
                     blend_mode: None,
+                    mask: mask.clone(),
                 };
                 x += width;
                 col += u32::from(width);
@@ -332,8 +339,12 @@ impl<const MODE: u8> Wide<MODE> {
                     let x_wtile_rel = x % WideTile::WIDTH;
                     let width = x2.min((wtile_x + 1) * WideTile::WIDTH) - x;
                     x += width;
-                    self.get_mut(wtile_x, strip_y)
-                        .fill(x_wtile_rel, width, paint.clone());
+                    self.get_mut(wtile_x, strip_y).fill(
+                        x_wtile_rel,
+                        width,
+                        paint.clone(),
+                        mask.clone(),
+                    );
                 }
             }
         }
@@ -817,7 +828,7 @@ impl<const MODE: u8> WideTile<MODE> {
         }
     }
 
-    pub(crate) fn fill(&mut self, x: u16, width: u16, paint: Paint) {
+    pub(crate) fn fill(&mut self, x: u16, width: u16, paint: Paint, mask: Option<Mask>) {
         if !self.is_zero_clip() {
             match MODE {
                 MODE_CPU => {
@@ -852,6 +863,7 @@ impl<const MODE: u8> WideTile<MODE> {
                             width,
                             paint,
                             blend_mode: None,
+                            mask,
                         }));
                     }
                 }
@@ -861,6 +873,7 @@ impl<const MODE: u8> WideTile<MODE> {
                         width,
                         paint,
                         blend_mode: None,
+                        mask,
                     }));
                 }
                 _ => unreachable!(),
@@ -1003,6 +1016,8 @@ pub struct CmdFill {
     pub paint: Paint,
     /// The blend mode to apply before drawing the contents.
     pub blend_mode: Option<BlendMode>,
+    /// A mask to apply to the command.
+    pub mask: Option<Mask>,
 }
 
 /// Fill a consecutive region of a wide tile with an alpha mask.
@@ -1022,6 +1037,8 @@ pub struct CmdAlphaFill {
     pub paint: Paint,
     /// A blend mode to apply before drawing the contents.
     pub blend_mode: Option<BlendMode>,
+    /// A mask to apply to the command.
+    pub mask: Option<Mask>,
 }
 
 /// Same as fill, but copies top of clip stack to next on stack.
@@ -1096,11 +1113,13 @@ mod tests {
             0,
             10,
             Paint::Solid(PremulColor::from_alpha_color(TRANSPARENT)),
+            None,
         );
         wide.fill(
             10,
             10,
             Paint::Solid(PremulColor::from_alpha_color(TRANSPARENT)),
+            None,
         );
         wide.pop_buf();
 
@@ -1116,8 +1135,8 @@ mod tests {
 
         let mut wide = WideTile::<MODE_CPU>::new(0, 0);
         wide.push_buf();
-        wide.fill(0, 10, paint.clone());
-        wide.fill(10, 10, paint.clone());
+        wide.fill(0, 10, paint.clone(), None);
+        wide.fill(10, 10, paint.clone(), None);
         wide.blend(blend_mode);
         wide.pop_buf();
 
@@ -1133,7 +1152,7 @@ mod tests {
 
         let mut wide = WideTile::<MODE_CPU>::new(0, 0);
         wide.push_buf();
-        wide.fill(0, 10, paint.clone());
+        wide.fill(0, 10, paint.clone(), None);
         wide.blend(blend_mode);
         wide.pop_buf();
 
--- a/sparse_strips/vello_common/src/mask.rs
+++ b/sparse_strips/vello_common/src/mask.rs
@@ -5,14 +5,18 @@
 
 use crate::pixmap::Pixmap;
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 
-/// A mask.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Mask {
-    data: Arc<[u8]>,
+#[derive(Debug, PartialEq, Eq)]
+struct MaskRepr {
+    data: Vec<u8>,
     width: u16,
     height: u16,
 }
+
+/// A mask.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Mask(Arc<MaskRepr>);
 
 impl Mask {
     /// Create a new alpha mask from the pixmap.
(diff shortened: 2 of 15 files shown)


LaurenzV:
I should note it's a bit unfortunate the size of a command increases, but I think if we really want to optimize for that there are other avanues we could explore first (for example the `paint` field is currently very big since it currently stores the f32 version and u8 version of a color inline.  I could also put it behind a feature flag somehow s that at least `vello_hybrid` is not affected.

LaurenzV:
I've rebased this onto main now. I also reran the Blend2D benchmark harness, and there does seem to be a slight slowdown in some cases, but nothing worrying (around 1-2%, in extreme cases 3%). It is a bit unfortunate that every command stores an Option<Mask> now, so maybe there is something better we can do here, but it's probably also not a huge deal either?

taj-p:
Sorry for the delay. I'll review this over the next couple days

LaurenzV:
All good, it can wait until after your vacation! Enjoy it!

sparse_strips/vello_common/src/mask.rs
```diff
@@ -25,55 +29,85 @@ impl Mask {
...
+    /// The pixels are in row-major order.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `data` vector is not of length `width * height`.
+    pub fn from_parts(data: Vec<u8>, width: u16, height: u16) -> Self {
+        assert_eq!(
+            data.len(),
+            usize::from(width) * usize::from(height),
+            "Expected `data` to have length of exactly `width * height`"
```
taj-p:
```suggestion
            "`data` should have `width * height` length"
```

Saves a few bytes in the binary

sparse_strips/vello_common/src/mask.rs
```diff
@@ -25,55 +29,85 @@ impl Mask {
...
+        self.0.height
     }
 
     /// Sample the value at a specific location.
     ///
     /// This function might panic or yield a wrong result if the location
     /// is out-of-bounds.
+    #[inline(always)]
     pub fn sample(&self, x: u16, y: u16) -> u8 {
         debug_assert!(
```
taj-p, line 109:
```suggestion
         let repr = &*self.0;
         debug_assert!(
```

Could try dereference the arc once instead of once per access (if we haven't checked that the compiler is optimising this away).

This form might make it easier for the compiler to optimise it away completely for successive calls to `sample`

sparse_strips/vello_common/src/mask.rs
```diff
@@ -5,15 +5,19 @@
...
 
-/// A mask.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Mask {
-    data: Arc<[u8]>,
+#[derive(Debug, PartialEq, Eq)]
+struct MaskRepr {
+    data: Vec<u8>,
     width: u16,
     height: u16,
```
taj-p, line 14: By putting `width` and `height` in the Arc, we're prioritising the performance of the clone (saving 4 bytes) over accessing these fields (since they now suffer an additional indirection). 

I'm wondering why we want to Arc these fields alongside `data`?

Edit: ahhh, I see. I think we're doing this to keep the size of a command smaller? WDYT about adding a small comment specifying that?

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -99,7 +100,12 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
         alphas: Option<&[u8]>,
     ) {
         if let Some(alphas) = alphas {
-            alpha_fill::alpha_composite_solid(simd, dest, src, alphas);
+            alpha_fill::alpha_composite_solid(
+                simd,
+                dest,
+                src,
+                alphas.chunks_exact(4).map(|c| [c[0], c[1], c[2], c[3]]),
```
taj-p: Is it possible to use bytemuck to zero cost reinterpret `alphas` (here and in the below instances)?

Oh 🤦‍♂️- I realise these are iterators. I'll leave it up to you - it might not be worth doing this

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -116,7 +122,7 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
                 simd,
                 dest,
                 src.chunks_exact(16).map(|el| f32x16::from_slice(simd, el)),
-                alphas,
+                alphas.chunks_exact(4).map(|c| [c[0], c[1], c[2], c[3]]),
```
taj-p: As above

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -130,14 +136,56 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
...
         src: impl Iterator<Item = Self::Composite>,
         blend_mode: BlendMode,
         alphas: Option<&[u8]>,
+        mask: Option<&Mask>,
     ) {
-        if let Some(alphas) = alphas {
-            alpha_fill::blend(simd, dest, src, alphas, blend_mode);
-        } else {
-            fill::blend(simd, dest, src, blend_mode);
+        let alpha_iter = alphas.map(|a| a.chunks_exact(4).map(|d| [d[0], d[1], d[2], d[3]]));
```
taj-p: As above

sparse_strips/vello_common/src/mask.rs
```diff
@@ -25,55 +29,85 @@ impl Mask {
...
+        Self(Arc::new(MaskRepr {
             data,
             width: pixmap.width(),
             height: pixmap.height(),
-        }
+        }))
     }
 
     /// Return the width of the mask.
+    #[inline]
```
taj-p: Should these be inline always?

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -130,14 +136,56 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
...
+
+        let mask_iter = mask.map(|m| {
+            core::iter::from_fn(|| {
+                let sample = |x: u16, y: u16| {
+                    if x < m.width() && y < m.height() {
```
taj-p: We can guarantee that the width and height calls are optimised away by implementing them ourselves via closure capture

```rs
let width = m.width();   // Load once, explicitly
let height = m.height(); // Load once, explicitly
let sample = |x: u16, y: u16| {
    if x < width && y < height {  // Just use the captured values
        m.sample(x, y)
    } else {
        255
    }
};
```

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -130,14 +136,56 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
...
+                        255
+                    }
+                };
+
+                let samples = [
+                    sample(start_x, start_y),
+                    sample(start_x, start_y + 1),
+                    sample(start_x, start_y + 2),
+                    sample(start_x, start_y + 3),
+                ];
```
taj-p: This performs a bounds check on a per sample basis. You know how masks are best used for your use case, but there might be a later optimisation to check whether something like the below is worth to do 1 check if the whole mask is in bounds:

```rs
let width = m.width();
let height = m.height();

let samples = if start_x < width && start_y + 3 < height {
    // All in bounds, sample directly
    [
        m.sample(start_x, start_y),
        m.sample(start_x, start_y + 1),
        m.sample(start_x, start_y + 2),
        m.sample(start_x, start_y + 3),
    ]
} else {
    // Fallback: check each individually
    [
        if start_x < width && start_y < height { m.sample(start_x, start_y) } else { 255 },
        if start_x < width && start_y + 1 < height { m.sample(start_x, start_y + 1) } else { 255 },
        if start_x < width && start_y + 2 < height { m.sample(start_x, start_y + 2) } else { 255 },
        if start_x < width && start_y + 3 < height { m.sample(start_x, start_y + 3) } else { 255 },
    ]
};

File: sparse_strips/vello_cpu/src/fine/lowp/mod.rs [hunk 10]
```diff
@@ -121,7 +124,14 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
         alphas: Option<&[u8]>,
     ) {
         if let Some(alphas) = alphas {
-            alpha_fill::alpha_composite_solid(simd, dest, src, alphas);
+            alpha_fill::alpha_composite_solid(
+                simd,
+                dest,
+                src,
+                alphas
```
taj-p: Should we use bytemuck to reinterpret here?

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
```diff
@@ -136,7 +146,14 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
         let src_iter = src.chunks_exact(32).map(|el| u8x32::from_slice(simd, el));
 
         if let Some(alphas) = alphas {
-            alpha_fill::alpha_composite(simd, dest, src_iter, alphas);
+            alpha_fill::alpha_composite(
+                simd,
+                dest,
+                src_iter,
+                alphas
+                    .chunks_exact(8)
```
taj-p: As above

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
```diff
@@ -145,14 +162,67 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
...
         blend_mode: BlendMode,
         alphas: Option<&[u8]>,
+        mask: Option<&Mask>,
     ) {
-        if let Some(alphas) = alphas {
-            alpha_fill::blend(simd, dest, src, blend_mode, alphas);
-        } else {
-            fill::blend(simd, dest, src, blend_mode);
+        let alpha_iter = alphas.map(|a| {
+            a.chunks_exact(8)
```
taj-p: As above

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -415,6 +422,12 @@ impl RenderContext {
         self.fill_rule = fill_rule;
     }
 
+    // TODO: Add explanation on how this differs to layer masks.
+    /// Set the mask to use for path-painting operations.
```
taj-p, line 459: We might need to share how masks are positioned (i.e. at (0,0)) and how sampling outside its bounds works

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -415,6 +422,12 @@ impl RenderContext {
...
+    pub fn set_mask(&mut self, mask: Option<Mask>) {
```
taj-p: Our other `set_*` APIs take a defined value and have a corresponding `reset_*` method. It might make sense to be consistent here and migrate all API to take an option (if that's what we want).

That said, I may not understand the use case thoroughly enough to comment on this aspect. What are your thoughts?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1029,6 +1042,8 @@ pub struct CmdFill {
     pub paint: Paint,
     /// The blend mode to apply before drawing the contents.
     pub blend_mode: BlendMode,
+    /// A mask to apply to the command.
+    pub mask: Option<Mask>,
```
taj-p, line 1502: What do you think of asserting on the size of a `Cmd` so that we force ourselves to more closely scrutinise changes that increase its size? I think I might have missed this without your comment above 


```rs
fn test_cmd_size() { 
    #[cfg(target_pointer_width = "64")] 
    assert_eq!(std::mem::size_of::<Cmd>(), XXX);
    #[cfg(target_pointer_width = "32")] 
    assert_eq!(std::mem::size_of::<Cmd>(), XXX); 
}
```

taj-p (approved):
LGTM!! The code looks good! I wasn't able to pull out an IDE and step through the code, but it looks good from just reading the code!

Sorry for the delay 🙏.

taj-p:
> I've rebased this onto main now. I also reran the Blend2D benchmark harness, and there does seem to be a slight slowdown in some cases, but nothing worrying (around 1-2%, in extreme cases 3%). It is a bit unfortunate that every command stores an Option<Mask> now, so maybe there is something better we can do here, but it's probably also not a huge deal either?

A quick win might be to stop using `usize` in `Cmd` since we really only need 32 bits for alpha indices. Happy to optimise this later.

sparse_strips/vello_cpu/src/fine/highp/mod.rs
LaurenzV: Once it's available on stable, we can use [`array_chunks`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.array_chunks)

LaurenzV: But will do the casts as suggested.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: Yep, I plan on adding those to the docs. 😄

sparse_strips/vello_common/src/mask.rs
LaurenzV: Done.

sparse_strips/vello_common/src/mask.rs
LaurenzV: Done.

sparse_strips/vello_common/src/mask.rs
LaurenzV: Done.

sparse_strips/vello_cpu/src/fine/highp/mod.rs
LaurenzV: Done.

sparse_strips/vello_common/src/mask.rs
LaurenzV: Done.

sparse_strips/vello_cpu/src/fine/highp/mod.rs
LaurenzV: Done.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Good idea!

sparse_strips/vello_cpu/src/render.rs
LaurenzV: Makes sense, done!

LaurenzV:
Thank you!! Now I can finally link hayro to main vello_cpu, which should also come in handy in case we want to test bigger changes. 😄

Issue #1213, opened 2025-09-12
Evaluate whether to merge `alpha_idx` and `fill_gap` in `Strip` into a single packed field


LaurenzV:
See: https://github.com/linebender/vello/pull/1206#discussion_r2345540559

PR #1238, opened 2025-09-21
Pack `fill_gap` into `alpha_idx`

taj-p:
Fixes https://github.com/linebender/vello/issues/1213

See https://github.com/linebender/vello/pull/1206#discussion_r2345540559

> Since Strip has an alignment of 4 bytes (due to alpha_idx), the contribution of fill_gap is an additional 4 bytes. This means Strip becomes 12 bytes or 5.3 strips per cache line.

> If we instead use the last bit of alpha_idx as fill_gap, Strip becomes 8 bytes or 8 Strips per cache line and much less to allocate.

The main motivations for the packing are to:
- Reduce the cost of `Recording`s and `Strip` allocation
- Improve cache line utilization

> When I tried this on Vello Hybrid, I saw no change in performance for direct rendering, but a minor improvement (3-5% improvement) in rendering from recordings (likely due to Strip being cheaper to allocate and iterate).

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -17,17 +17,62 @@ pub struct Strip {
     pub x: u16,
     /// The y coordinate of the strip, in user coordinates.
     pub y: u16,
-    /// The index into the alpha buffer.
-    pub alpha_idx: u32,
-    /// Whether the gap that lies between this strip and the previous in the _same_
-    /// row should be filled.
-    pub fill_gap: bool,
+    /// Packed alpha index and fill gap flag.
+    ///
+    /// Bit layout (u32):
+    /// - bit 31: (fill_gap) See `Strip::fill_gap()`.
+    /// - bits 0..=30: (alpha_idx) See `Strip::alpha_idx()`.
+    packed_alpha_idx_fill_gap: u32,
 }
 
 impl Strip {
+    /// The bit mask for `fill_gap` packed into `packed_alpha_idx_fill_gap`.
+    const FILL_GAP_MASK: u32 = 1 << 31;
+
+    /// Create a new strip.
+    pub fn new(x: u16, y: u16, alpha_idx: u32, fill_gap: bool) -> Self {
+        // Ensure `alpha_idx` does not collide with the fill flag bit.
+        assert!(alpha_idx & Self::FILL_GAP_MASK == 0);
+        let fill_gap = u32::from(fill_gap) << 31;
+        Self {
+            x,
+            y,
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_common/src/strip.rs
```diff
@@ -17,18 +17,72 @@ pub struct Strip {
...
+
+    /// Returns the alpha index.
+    #[inline(always)]
+    pub fn alpha_idx(&self) -> u32 {
+        self.packed_alpha_idx_fill_gap & !Self::FILL_GAP_MASK
+    }
+
+    /// Sets the alpha index.
+    ///
+    /// Note that the largest value that can be stored in the alpha index is `u32::MAX - 1`, as the
```
LaurenzV: Isn't it `u32::MAX / 2` (or `u32::MAX << 1`)?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -915,7 +915,7 @@ impl RenderContext {
             .iter()
             .map(move |strip| {
                 let mut adjusted_strip = *strip;
-                adjusted_strip.alpha_idx += alpha_offset;
+                adjusted_strip.set_alpha_idx(adjusted_strip.alpha_idx().wrapping_add(alpha_offset));
```
LaurenzV: Do we really want a wrapping add here? Wouldn't it be better to just use normal add so that at least in debug mode it panics?

LaurenzV (approved):
It is a bit unfortunate that this reduces the number of alpha values that can be stored by a lot, but I doubt that we will exceed 2GB under normal circumstances.

sparse_strips/vello_common/src/strip.rs
taj-p:
```suggestion
    /// Note that the largest value that can be stored in the alpha index is `u32::MAX << 1`, as the
```

taj-p: Whoops - fixed in https://github.com/linebender/vello/pull/1238/commits/bd97a68e2e575fbb5d2986369755e5be3062e07c

sparse_strips/vello_cpu/src/render.rs
taj-p: Yes, you're right. Thank you - nice catch! Fixed in https://github.com/linebender/vello/pull/1238/commits/a217c0cd72a7af210679497d6e6e1864adfe8514

taj-p: And fixed in hybrid in https://github.com/linebender/vello/pull/1238/commits/c0947d3830b4bdb02d7952975cdd0bbd903c27a5

PR #1252, opened 2025-09-29
[`vello_hybrid`]: Multiple image atlases support

grebmeg:
This PR adds support for multiple image atlases to `vello_hybrid`, replacing the previous single-atlas limitation. The change introduces a new multi-atlas management system that can dynamically allocate additional texture atlases when needed.

The implementation includes updates to `ImageCache`, both webgl and wgpu renderers, and the underlying shaders to handle multiple atlas textures. This enhancement provides better scalability for image-heavy applications and resolves memory constraints that could occur when a single atlas became full.

### DIFF
--- a/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
@@ -17,9 +17,7 @@ use vello_common::{
     paint::{ImageId, ImageSource},
 };
 use vello_example_scenes::{AnyScene, image::ImageScene};
-use vello_hybrid::{
-    AllocationStrategy, AtlasConfig, Pixmap, RenderSettings, RenderTargetConfig, Renderer, Scene,
-};
+use vello_hybrid::{AtlasConfig, Pixmap, RenderSettings, RenderTargetConfig, Renderer, Scene};
 use wasm_bindgen::prelude::*;
 use web_sys::{Event, HtmlCanvasElement, KeyboardEvent, MouseEvent, WheelEvent};
 
@@ -90,16 +88,7 @@ impl RendererWrapper {
                 level: Level::try_detect().unwrap_or(Level::fallback()),
                 atlas_config: AtlasConfig {
                     atlas_size: (max_texture_dimension_2d, max_texture_dimension_2d),
-                    // In WGPU’s GLES backend, heuristics are used to decide whether a texture
-                    // should be treated as D2 or D2Array. However, this can cause a mismatch:
-                    // - when depth_or_array_layers == 1, the backend assumes the texture is D2,
-                    // even if it was actually created as a D2Array. This issue only occurs with the GLES backend.
-                    //
-                    // @see https://github.com/gfx-rs/wgpu/blob/61e5124eb9530d3b3865556a7da4fd320d03ddc5/wgpu-hal/src/gles/mod.rs#L470-L517
-                    initial_atlas_count: 2,
-                    max_atlases: 10,
-                    auto_grow: true,
-                    allocation_strategy: AllocationStrategy::FirstFit,
+                    ..AtlasConfig::default()
                 },
             },
         );
--- a/sparse_strips/vello_hybrid/src/image_cache.rs
+++ b/sparse_strips/vello_hybrid/src/image_cache.rs
@@ -1,8 +1,6 @@
 // Copyright 2025 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-#![allow(dead_code, reason = "Clippy fails when --no-default-features")]
-
 use crate::multi_atlas::{AtlasConfig, AtlasError, AtlasId, MultiAtlasManager};
 use alloc::vec::Vec;
 use guillotiere::AllocId;
@@ -49,15 +47,6 @@ impl core::fmt::Debug for ImageCache {
 }
 
 impl ImageCache {
-    /// Create a new image cache with default atlas configuration.
-    pub(crate) fn new(width: u32, height: u32) -> Self {
-        Self {
-            atlas_manager: MultiAtlasManager::new_with_initial_atlas(width, height),
-            slots: Vec::new(),
-            free_idxs: Vec::new(),
-        }
-    }
-
     /// Create a new image cache with custom atlas configuration.
     pub(crate) fn new_with_config(config: AtlasConfig) -> Self {
         Self {
@@ -89,7 +78,7 @@ impl ImageCache {
         });
 
         let image_id = ImageId::new(slot_idx as u32);
-        self.slots[slot_idx] = Some(ImageResource {
+        let image_resource = ImageResource {
             id: image_id,
             width: width as u16,
             height: height as u16,
@@ -99,8 +88,9 @@ impl ImageCache {
                 atlas_alloc.allocation.rectangle.min.y as u16,
             ],
             atlas_alloc_id: atlas_alloc.allocation.id,
-        });
-        Ok(image_id)
+        };
+        self.slots[slot_idx] = Some(image_resource);
+        Ok(self.slots[slot_idx].as_ref().unwrap().id)
     }
 
     /// Deallocate an image from the cache, returning the image resource if it existed.
@@ -108,12 +98,14 @@ impl ImageCache {
         let index = id.as_u32() as usize;
         if let Some(image_resource) = self.slots.get_mut(index).and_then(Option::take) {
             // Deallocate from the appropriate atlas
-            let _ = self.atlas_manager.deallocate(
-                image_resource.atlas_id,
-                image_resource.atlas_alloc_id,
-                image_resource.width as u32,
-                image_resource.height as u32,
-            );
+            self.atlas_manager
+                .deallocate(
+                    image_resource.atlas_id,
+                    image_resource.atlas_alloc_id,
+                    image_resource.width as u32,
+                    image_resource.height as u32,
+                )
+                .unwrap();
             self.free_idxs.push(index);
             Some(image_resource)
         } else {
@@ -126,21 +118,9 @@ impl ImageCache {
         &self.atlas_manager
     }
 
-    /// Get atlas stats.
-    pub(crate) fn atlas_stats(&self) -> Vec<(AtlasId, &crate::multi_atlas::AtlasUsageStats)> {
-        self.atlas_manager.atlas_stats()
-    }
-
     /// Get the number of atlases.
     pub(crate) fn atlas_count(&self) -> usize {
         self.atlas_manager.atlas_count()
-    }
-
-    /// Clear all images from the cache.
-    pub(crate) fn clear(&mut self) {
-        self.slots.clear();
-        self.free_idxs.clear();
-        self.atlas_manager.clear();
     }
 }
 
@@ -152,7 +132,10 @@ mod tests {
 
     #[test]
     fn test_insert_single_image() {
-        let mut cache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let mut cache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         let id = cache.allocate(100, 100).unwrap();
 
@@ -161,12 +144,16 @@ mod tests {
         assert_eq!(resource.id, id);
         assert_eq!(resource.width, 100);
         assert_eq!(resource.height, 100);
-        assert_eq!(resource.offset, [0, 0]); // First image should be at origin
+        // First image should be at origin
+        assert_eq!(resource.offset, [0, 0]);
     }
 
     #[test]
     fn test_insert_multiple_images() {
-        let mut cache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let mut cache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         let id1 = cache.allocate(50, 50).unwrap();
         let id2 = cache.allocate(75, 75).unwrap();
@@ -186,7 +173,10 @@ mod tests {
 
     #[test]
     fn test_get_nonexistent_image() {
-        let cache: ImageCache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let cache: ImageCache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         assert!(cache.get(ImageId::new(0)).is_none());
         assert!(cache.get(ImageId::new(999)).is_none());
@@ -194,7 +184,10 @@ mod tests {
 
     #[test]
     fn test_remove_image() {
-        let mut cache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let mut cache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         let id = cache.allocate(100, 100).unwrap();
         assert!(cache.get(id).is_some());
@@ -205,7 +198,10 @@ mod tests {
 
     #[test]
     fn test_remove_nonexistent_image() {
-        let mut cache: ImageCache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let mut cache: ImageCache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         // Should not panic when unregistering non-existent image
         cache.deallocate(ImageId::new(0));
@@ -214,7 +210,10 @@ mod tests {
 
     #[test]
     fn test_slot_reuse_after_remove() {
-        let mut cache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let mut cache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         // Register three images
         let id1 = cache.allocate(50, 50).unwrap();
@@ -243,7 +242,10 @@ mod tests {
 
     #[test]
     fn test_multiple_remove_and_reuse() {
-        let mut cache = ImageCache::new(ATLAS_SIZE, ATLAS_SIZE);
+        let mut cache = ImageCache::new_with_config(AtlasConfig {
+            atlas_size: (ATLAS_SIZE, ATLAS_SIZE),
+            ..Default::default()
+        });
 
         // Register several images
         let ids: Vec<_> = (0..5)
--- a/sparse_strips/vello_hybrid/src/lib.rs
+++ b/sparse_strips/vello_hybrid/src/lib.rs
@@ -41,7 +41,7 @@ mod scene;
 #[cfg(any(all(target_arch = "wasm32", feature = "webgl"), feature = "wgpu"))]
 mod schedule;
 pub mod util;
-pub use multi_atlas::{AllocationStrategy, AtlasConfig, AtlasError, AtlasId};
+pub use multi_atlas::{AllocationStrategy, AtlasConfig};
 #[cfg(feature = "wgpu")]
 pub use render::{AtlasWriter, RenderTargetConfig, Renderer};
 pub use render::{Config, GpuStrip, RenderSize};
--- a/sparse_strips/vello_hybrid/src/multi_atlas.rs
+++ b/sparse_strips/vello_hybrid/src/multi_atlas.rs
@@ -8,6 +8,7 @@
 
 use alloc::vec::Vec;
 use guillotiere::{AllocId, Allocation, AtlasAllocator, size2};
+use thiserror::Error;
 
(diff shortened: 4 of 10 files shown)


sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -75,13 +78,30 @@ impl RendererWrapper {
...
+            RenderSettings {
+                level: Level::try_detect().unwrap_or(Level::fallback()),
+                atlas_config: AtlasConfig {
+                    atlas_size: (max_texture_dimension_2d, max_texture_dimension_2d),
+                    // In WGPU’s GLES backend, heuristics are used to decide whether a texture
+                    // should be treated as D2 or D2Array. However, this can cause a mismatch:
+                    // - when depth_or_array_layers == 1, the backend assumes the texture is D2,
+                    // even if it was actually created as a D2Array. This issue only occurs with the GLES backend.
+                    //
+                    // @see https://github.com/gfx-rs/wgpu/blob/61e5124eb9530d3b3865556a7da4fd320d03ddc5/wgpu-hal/src/gles/mod.rs#L470-L517
```
taj-p: Wow! That's unexpected by WGPU. Thanks for the detailed comment.

sparse_strips/vello_hybrid/src/multi_atlas.rs
```diff
@@ -0,0 +1,539 @@
...
+        width: u32,
+        height: u32,
+    ) -> Result<(), AtlasError> {
+        for atlas in &mut self.atlases {
+            if atlas.id == atlas_id {
+                atlas.deallocate(alloc_id, width, height);
+                return Ok(());
+            }
+        }
+        Err(AtlasError::AtlasNotFound(atlas_id))
```
taj-p:
```suggestion
        let atlas = self
            .atlases
            .get_mut(atlas_id.0 as usize)
            .ok_or_else(|| AtlasError::AtlasNotFound(atlas_id))?;
        atlas.deallocate(alloc_id, width, height);
        Ok(())
```

Since atlases only grow (never deallocate) and id is the index into the atlases vec, we can do a lookup instead of a linear search

sparse_strips/vello_hybrid/src/multi_atlas.rs
```diff
@@ -0,0 +1,539 @@
...
+            .map(|atlas| (atlas.id, atlas.stats()))
+            .collect()
+    }
+
+    /// Get the number of atlases.
+    pub(crate) fn atlas_count(&self) -> usize {
+        self.atlases.len()
+    }
+
+    /// Clear all atlases.
```
taj-p:
```suggestion
    /// Clear all atlases. Note that atlases are _not_ removed, simply emptied of contents.
```

sparse_strips/vello_hybrid/src/multi_atlas.rs
taj-p: Extra question: Did we not want to remove atlases ever (even if they're empty?)? Or should we leave that to a later optimization (since it has implications on a lot of things)?

sparse_strips/vello_hybrid/src/multi_atlas.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// Choose the atlas with the smallest remaining space that can fit the image.
+    BestFit,
+    /// Prefer the atlas with the lowest usage percentage.
+    LeastUsed,
+    /// Cycle through atlases in round-robin fashion.
+    RoundRobin,
+}
+
+#[cfg(test)]
+mod tests {
```
taj-p, line 455: What do you think about adding a handful of unit tests for the different `allocation_strategy`s?

sparse_strips/vello_hybrid/src/multi_atlas.rs
```diff
@@ -0,0 +1,539 @@
...
+        // Check if the image is too large for any atlas
+        if width > self.config.atlas_size.0 || height > self.config.atlas_size.1 {
+            return Err(AtlasError::TextureTooLarge { width, height });
+        }
+
+        // If no atlases exist and auto-grow is enabled, create one
+        if self.atlases.is_empty() && self.config.auto_grow {
+            self.create_atlas()?;
+        }
+
```
taj-p:
```suggestion
```

Don't each allocation strategy fallback to growing?

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -35,11 +35,13 @@ extern crate alloc;
 
 mod gradient_cache;
 mod image_cache;
+mod multi_atlas;
 mod render;
 mod scene;
 #[cfg(any(all(target_arch = "wasm32", feature = "webgl"), feature = "wgpu"))]
 mod schedule;
 pub mod util;
+pub use multi_atlas::{AllocationStrategy, AtlasConfig, AtlasError, AtlasId};
```
taj-p:
```suggestion
pub use multi_atlas::{AllocationStrategy, AtlasConfig};
```

AFAICT, these fields don't need to be exposed to public consumers of hybrid

sparse_strips/vello_hybrid/src/multi_atlas.rs
```diff
@@ -0,0 +1,539 @@
...
+            .field("id", &self.id)
+            .field("stats", &self.stats)
+            .field("allocation_counter", &self.allocation_counter)
+            .finish_non_exhaustive()
+    }
+}
+
+/// Errors that can occur during atlas operations.
+#[derive(Debug)]
+pub enum AtlasError {
```
taj-p: Could consider using `thiserror` like we do for [`RenderError`](https://github.com/taj-p/vello/blob/4c7dd4f717efbbbddc05ab701dc0ae37778f3aee/sparse_strips/vello_hybrid/src/lib.rs#L54-L68)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -27,12 +29,15 @@ pub(crate) const DEFAULT_TOLERANCE: f64 = 0.1;
 pub struct RenderSettings {
     /// The SIMD level that should be used for rendering operations.
     pub level: Level,
+    /// The configuration for the atlas.
```
```suggestion
    /// The configuration for the texture atlas.
```

Nit: Could be worth expanding what this setting means and how the Atlas is used so that consumer can get a better understanding of how this config may affect their application.

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
taj-p: I'm wondering whether we should use a `cfg` gate with a `WGPU` backend to ensure `initial_atlas_count` is at least 2?

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -104,42 +93,68 @@ impl ImageCache {
...
 
-    /// Deallocate an image from the cache, returning true if it existed
+    /// Deallocate an image from the cache, returning the image resource if it existed.
     pub(crate) fn deallocate(&mut self, id: ImageId) -> Option<ImageResource> {
         let index = id.as_u32() as usize;
         if let Some(image_resource) = self.slots.get_mut(index).and_then(Option::take) {
-            // Deallocate from the atlas using the stored allocation ID
-            self.atlas.deallocate(image_resource.atlas_alloc_id);
+            // Deallocate from the appropriate atlas
+            let _ = self.atlas_manager.deallocate(
```
taj-p: I was wondering how `deallocate` can return an error. It seems it can only return an error if it can't find the `atlas_id` which shouldn't be possible right? Is it worth `.unwrap`ing here so that we can at least ensure that is upheld

sparse_strips/vello_hybrid/src/image_cache.rs
```diff
@@ -104,42 +93,68 @@ impl ImageCache {
...
+        self.atlas_manager.atlas_stats()
+    }
+
+    /// Get the number of atlases.
+    pub(crate) fn atlas_count(&self) -> usize {
+        self.atlas_manager.atlas_count()
+    }
+
+    /// Clear all images from the cache.
     pub(crate) fn clear(&mut self) {
```
taj-p: I couldn't see where this method is used. Should we remove it?

sparse_strips/vello_hybrid/src/render/common.rs
taj-p: Is it possible to pack atlas_index here? (I think there's 24 bits available here for use)

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -301,25 +308,33 @@ impl WebGlRenderer {
...
+    /// Clear a specific region of the atlas texture array with uninitialized data.
+    fn clear_atlas_region(
+        &mut self,
+        atlas_id: crate::AtlasId,
+        offset: [u32; 2],
+        width: u32,
+        height: u32,
+    ) {
         // Rgba8Unorm is 4 bytes per pixel
         let uninitialized_data = vec![0_u8; (width * height * 4) as usize];
```
taj-p: Could we reuse a scratch buffer (see `WebGlPrograms::[alpha_data|encoded_paints_data]`) instead of allocating a new vector that will be freed at the end of this function?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -655,6 +670,112 @@ impl WebGlPrograms {
...
+            self.copy_atlas_texture_data(gl, &new_atlas_texture_array, current_atlas_count);
+
+            // Replace the old resources
+            self.resources.atlas_texture_array = new_atlas_texture_array;
+        }
+    }
+
+    /// Copy texture data from the old atlas texture array to a new one.
+    /// This is necessary when resizing the texture array to preserve existing atlas data.
+    fn copy_atlas_texture_data(
```
taj-p, line 706: Could something like this work to avoid a `draw_framebuffer` and a blit by using `copy_tex_sub_image_3d`


```rs
    fn copy_atlas_texture_data(
        &self,
        gl: &WebGl2RenderingContext,
        new_atlas_texture_array: &WebGlTextureArray,
        layer_count_to_copy: u32,
    ) {
        let WebGlTextureSize { width, height, .. } = self.resources.atlas_texture_array.size();

        let read_framebuffer = gl.create_framebuffer().unwrap();

        // Save original bindings we will modify (READ_FRAMEBUFFER and TEXTURE_2D_ARRAY)
        let original_read_fb = gl
            .get_parameter(WebGl2RenderingContext::READ_FRAMEBUFFER_BINDING)
            .ok()
            .and_then(|v| v.dyn_into::<web_sys::WebGlFramebuffer>().ok());
        let original_tex_array_binding = gl
            .get_parameter(WebGl2RenderingContext::TEXTURE_BINDING_2D_ARRAY)
            .ok()
            .and_then(|v| v.dyn_into::<web_sys::WebGlTexture>().ok());

        // Bind destination atlas texture array once; subsequent copies write into this binding
        gl.active_texture(WebGl2RenderingContext::TEXTURE0);
        gl.bind_texture(
            WebGl2RenderingContext::TEXTURE_2D_ARRAY,
            Some(&new_atlas_texture_array.texture),
        );

        // Bind the READ framebuffer once
        gl.bind_framebuffer(
            WebGl2RenderingContext::READ_FRAMEBUFFER,
            Some(&read_framebuffer),
        );

        // Copy each layer using copyTexSubImage3D from the READ framebuffer to the bound array
        for layer in 0..layer_count_to_copy {
            // Attach source layer to READ framebuffer
            gl.framebuffer_texture_layer(
                WebGl2RenderingContext::READ_FRAMEBUFFER,
                WebGl2RenderingContext::COLOR_ATTACHMENT0,
                Some(&self.resources.atlas_texture_array.texture),
                0,
                layer as i32,
            );

            // Copy into destination array layer
            gl.copy_tex_sub_image_3d(
                WebGl2RenderingContext::TEXTURE_2D_ARRAY,
                0, // mip level
                0,
                0,
                layer as i32, // xoffset, yoffset, zoffset (dest layer)
                0,
                0, // x, y (src)
                width as i32,
                height as i32,
            );
        }

        // Restore original bindings
        gl.bind_framebuffer(
            WebGl2RenderingContext::READ_FRAMEBUFFER,
            original_read_fb.as_ref(),
        );
        gl.bind_texture(
            WebGl2RenderingContext::TEXTURE_2D_ARRAY,
            original_tex_array_binding.as_ref(),
        );

        // Clean up
        gl.delete_framebuffer(Some(&read_framebuffer));
    }
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1756,56 +1908,86 @@ impl WebGlAtlasWriter for WebGlTexture {
         unreachable!("WebGlTexture height must be provided by caller")
     }
 
-    fn write_to_atlas(
+    fn write_to_atlas_layer(
```
taj-p, line 2039: I think this can undergo the same treatment as the copy command (maybe use a single read framebuffer and avoid a blit):

```rs
        // Create a framebuffer for read operations only
        let read_framebuffer = gl.create_framebuffer().unwrap();

        // Save original bindings we will modify (READ_FRAMEBUFFER and TEXTURE_2D_ARRAY)
        let original_read_fb = gl
            .get_parameter(WebGl2RenderingContext::READ_FRAMEBUFFER_BINDING)
            .ok()
            .and_then(|v| v.dyn_into::<web_sys::WebGlFramebuffer>().ok());
        let original_tex_array_binding = gl
            .get_parameter(WebGl2RenderingContext::TEXTURE_BINDING_2D_ARRAY)
            .ok()
            .and_then(|v| v.dyn_into::<web_sys::WebGlTexture>().ok());

        // Bind the destination atlas texture array once
        gl.active_texture(WebGl2RenderingContext::TEXTURE0);
        gl.bind_texture(
            WebGl2RenderingContext::TEXTURE_2D_ARRAY,
            Some(atlas_texture_array),
        );

        // Bind READ framebuffer and attach the source 2D texture
        gl.bind_framebuffer(
            WebGl2RenderingContext::READ_FRAMEBUFFER,
            Some(&read_framebuffer),
        );
        gl.framebuffer_texture_2d(
            WebGl2RenderingContext::READ_FRAMEBUFFER,
            WebGl2RenderingContext::COLOR_ATTACHMENT0,
            Some(self),
            0,
        );

        // Copy from the READ framebuffer into the destination array layer at the given offset
        gl.copy_tex_sub_image_3d(
            WebGl2RenderingContext::TEXTURE_2D_ARRAY,
            0,                // mip level
            offset[0] as i32, // xoffset
            offset[1] as i32, // yoffset
            layer as i32,     // zoffset (dest layer)
            0,                // src x
            0,                // src y
            width as i32,
            height as i32,
        );

        // Restore original bindings
        gl.bind_framebuffer(
            WebGl2RenderingContext::READ_FRAMEBUFFER,
            original_read_fb.as_ref(),
        );
        gl.bind_texture(
            WebGl2RenderingContext::TEXTURE_2D_ARRAY,
            original_tex_array_binding.as_ref(),
        );

        // Clean up
        gl.delete_framebuffer(Some(&read_framebuffer));
```

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1237,6 +1287,98 @@ impl Programs {
         }
     }
 
+    /// Resize the texture array to accommodate more atlases.
+    fn maybe_resize_atlas_texture_array(
```
taj-p, line 1349: I have a couple questions about this method that I'm unsure about:

- Is it incorrect to use the `CommandEncoder` passed to `upload_image` rather than to create a new encoder and queue submission in `copy_atlas_texture_data`?

- Is it possible to use a single `copy_texture_to_texture` in `copy_atlas_texture_data`?

Something like:

```rs
    fn copy_atlas_texture_data(
        encoder: &mut CommandEncoder,
        old_atlas_texture_array: &Texture,
        new_atlas_texture_array: &Texture,
        layer_count_to_copy: u32,
        width: u32,
        height: u32,
    ) {
        // Copy all existing layers in a single command. 
        encoder.copy_texture_to_texture(
            wgpu::TexelCopyTextureInfo {
                texture: old_atlas_texture_array,
                mip_level: 0,
                origin: wgpu::Origin3d { x: 0, y: 0, z: 0 },
                aspect: wgpu::TextureAspect::All,
            },
            wgpu::TexelCopyTextureInfo {
                texture: new_atlas_texture_array,
                mip_level: 0,
                origin: wgpu::Origin3d { x: 0, y: 0, z: 0 },
                aspect: wgpu::TextureAspect::All,
            },
            wgpu::Extent3d {
                width,
                height,
                depth_or_array_layers: layer_count_to_copy, // <--- copy this count, rather than 1 by 1
            },
        );
    }
```

taj-p (approved):
Strategy LGTM! Nice work 🚀 . I left a few comments, let me know if anything isn't clear

edit: I tested the atlas by creating 100 images (uploaded in different textures) in `winit` and `native_webgl` and checked they rendered correctly

image

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
grebmeg: Yeah, I was surprised too, but I couldn’t get past that check. I think we can use a `cfg` gate here, and it’d actually be better to place the comment in the `AtlasConfig` instead of in this example. Thanks!

Fixed.

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Ah, I must have missed that. Thanks for pointing it out!

Fixed.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: I realized that clearing can be performed more efficiently using gpu-based clearing, which eliminates cpu memory allocations and keeps the operation entirely on gpu! Thanks for your comment!

Fixed.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Yes, definitely, that makes sense in this context. I’m only really curious how their performance compares to each other.

Fixed.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Yes, that’s right! I created `copy_to_texture_array_layer` helper to unify both implementations into a single one.

Fixed.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: > Is it incorrect to use the CommandEncoder passed to upload_image rather than to create a new encoder and queue submission in copy_atlas_texture_data?

It’s perfectly valid! I just initially assumed it might be faster, but without actual data it’s hard to say. A sync barrier would still be applied between copying the atlas texture array and the immediate write into that texture, so the benefit of splitting into multiple encoders may even diminish potential perf win. Howerver, I’m not sure; I’d need to measure real performance. For now, I’d stick with using a single encoder.

> Is it possible to use a single copy_texture_to_texture in copy_atlas_texture_data?

Yes! I somehow missed that awesome enhancement over webgl texture array copying. Thank you! 

Fixed.

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: Yes, that’s right, I agree!

Fixed.

sparse_strips/vello_hybrid/src/image_cache.rs
grebmeg: Yes, I had `#![allow(dead_code)]` set for the file, so the CI skipped failing on those unused methods and I’ve realized there are a few others too. I’ll clean them all up. Thanks!

Fixed.

sparse_strips/vello_hybrid/src/lib.rs
grebmeg: Fixed.

sparse_strips/vello_hybrid/src/multi_atlas.rs
grebmeg: Yes, that's correct.

Fixed.

sparse_strips/vello_hybrid/src/multi_atlas.rs
grebmeg: > Since atlases only grow (never deallocate) and id is the index into the atlases vec, we can do a lookup instead of a linear search

Fixed.

> Extra question: Did we not want to remove atlases ever (even if they're empty?)? Or should we leave that to a later optimization (since it has implications on a lot of things)?

It definitely complicates things in a few places, and I was thinking we could defer addressing it until later, if you don’t mind.

sparse_strips/vello_hybrid/src/multi_atlas.rs
grebmeg: Since it’s unused for now, I’ll remove that method and re-add it whenever we need it.

Fixed.

sparse_strips/vello_hybrid/src/multi_atlas.rs
grebmeg: Absolutely, thanks for pointing that out! I really prefer `thiserror`!

Fixed.

sparse_strips/vello_hybrid/src/multi_atlas.rs
grebmeg: Certainly, I’ll include some basic tests.

Added.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Actually, I think it’s definitely worth adding! Thanks!

Done.

PR #1253, opened 2025-09-29
sparse_strips: Use peniko::Brush

sagudev:
demo for https://github.com/linebender/peniko/pull/134

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -20,7 +20,8 @@ use core::hash::{Hash, Hasher};
 use fearless_simd::{Simd, SimdBase, SimdFloat, f32x4, f32x16};
 use peniko::color::cache_key::{BitEq, BitHash, CacheKey};
 use peniko::{
-    InterpolationAlphaSpace, LinearGradientPosition, RadialGradientPosition, SweepGradientPosition,
+    ImageSampler, InterpolationAlphaSpace, LinearGradientPosition, RadialGradientPosition,
+    SweepGradientPosition,
 };
 use smallvec::ToSmallVec;
 // So we can just use `OnceCell` regardless of which feature is activated.
@@ -430,7 +431,7 @@ impl EncodeExt for Image {
     fn encode_into(&self, paints: &mut Vec<EncodedPaint>, transform: Affine) -> Paint {
         let idx = paints.len();
 
-        let mut quality = self.quality;
+        let mut sampler = self.sampler;
 
         let c = transform.as_coeffs();
 
@@ -441,9 +442,9 @@ impl EncodeExt for Image {
             && (c[3] as f32 - 1.0).is_nearly_zero()
             && ((c[4] - c[4].floor()) as f32).is_nearly_zero()
             && ((c[5] - c[5].floor()) as f32).is_nearly_zero()
-            && quality == ImageQuality::Medium
+            && sampler.quality == ImageQuality::Medium
         {
-            quality = ImageQuality::Low;
+            sampler.quality = ImageQuality::Low;
         }
 
         // Similarly to gradients, apply a 0.5 offset so we sample at the center of
@@ -452,12 +453,11 @@ impl EncodeExt for Image {
 
         let (x_advance, y_advance) = x_y_advances(&transform);
 
-        let encoded = match &self.source {
+        let encoded = match &self.image {
             ImageSource::Pixmap(pixmap) => {
                 EncodedImage {
                     source: ImageSource::Pixmap(pixmap.clone()),
-                    extends: (self.x_extend, self.y_extend),
-                    quality,
+                    sampler,
                     // While we could optimize RGB8 images, it's probably not worth the trouble.
                     has_opacities: true,
                     transform,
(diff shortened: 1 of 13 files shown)


sparse_strips/vello_common/src/paint.rs
```diff
@@ -133,9 +121,18 @@ impl Image {
             .data
             .data()
             .chunks_exact(4)
-            .map(|rgba| {
+            .map(|pixel| {
+                let rgba: [u8; 4] = match brush.image.format {
+                    peniko::ImageFormat::Rgba8 => pixel.try_into().unwrap(),
+                    peniko::ImageFormat::Bgra8 => [pixel[2], pixel[1], pixel[0], pixel[3]],
```
sagudev, line 111: this stuff is still not handled

LaurenzV:
Thanks a lot, that looks reasonable to me. Assuming that we can get CI to pass, I would be fine merging the peniko change.

sparse_strips/vello_common/src/paint.rs
sagudev: I think we should just port unimplemented for now in renderer and then implement it in follow up rather than in this PR.

sagudev: Actually we only need to deal with image alpha (on sample I suppose). BGRA and premultiply are props of ImageData that vello_cpu does not use.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -430,7 +431,12 @@ impl EncodeExt for Image {
     fn encode_into(&self, paints: &mut Vec<EncodedPaint>, transform: Affine) -> Paint {
         let idx = paints.len();
 
-        let mut quality = self.quality;
+        let mut sampler = self.sampler;
+
+        if sampler.alpha != 1.0 {
+            // If the sampler alpha is not 1.0, we need to force alpha compositing.
+            unimplemented!("Applying opacity to image commands");
+        }
```
LaurenzV, line 439: Doesn't have to be addressed in this PR, but I think it would be smarter to apply the opacity during encoding time instead of at sampling time (using `Arc::make_mut(pixmap)`)? I'm concerned that adding an if branch for that for sampling will be pretty bad for performance, especially when doing bilinear/bicubic filtering.

I also thought we already support image alphas, but seems like I remember wrong...

LaurenzV: And I think using `unimplented` is fine for now, but before the first minor release we should probably replace all panics with a warning instead.

sagudev: > Doesn't have to be addressed in this PR, but I think it would be smarter to apply the opacity during encoding time instead of at sampling time (using `Arc::make_mut(pixmap)`)? I'm concerned that adding an if branch for that for sampling will be pretty bad for performance, especially when doing bilinear/bicubic filtering.

I am not sure this would work, for cases where one wants to put same image with different alpha (this is currently possible in vello classic and is expressed as alpha in ImageSampler).

> And I think using unimplented is fine for now, but before the first minor release we should probably replace all panics with a warning instead.

Yes, that's how we already have other stuff, so I just continued the pattern.

LaurenzV: > I am not sure this would work, for cases where one wants to put same image with different alpha (this is currently possible in vello classic and is expressed as alpha in ImageSampler).

Well, in this case can just do the encoding twice? I think that applying an opacity to an image is already pretty rare, placing the same image twice with different alphas is even more rare. I don't think it's worth sacrificing rendering performance for _all_ images just to account for those special cases.

sagudev:
@LaurenzV feel free to land if everything is ok.

sparse_strips/vello_common/src/encode.rs
sagudev: > > I am not sure this would work, for cases where one wants to put same image with different alpha (this is currently possible in vello classic and is expressed as alpha in ImageSampler).
> 
> Well, in this case can just do the encoding twice? I think that applying an opacity to an image is already pretty rare, placing the same image twice with different alphas is even more rare. I don't think it's worth sacrificing rendering performance for _all_ images just to account for those special cases.

I wonder if we can do this in creation of image sampler (`ImagePainterData`) using generics rather than doing in sampling (which is hot). This way only do "dispatch" once.

LaurenzV: In theory yes, but the main problem is that we already have lots of generics in `vello-cpu` which are already impacting compile times (and probably build size) pretty badly. :/ So I would be careful about adding even more, especially since each additional generic serves as a multiplying factor.

sparse_strips/vello_common/src/paint.rs
```diff
@@ -77,84 +77,68 @@ pub enum ImageSource {
...
+                        b: multiply(rgba[2]),
+                        a: rgba[3],
+                    }
+                } else {
+                    PremulRgba8 {
+                        r: rgba[0],
+                        g: rgba[1],
+                        b: rgba[2],
+                        a: rgba[3],
+                    }
```
grebmeg, line 129: cc @ajakubowicz-canva re: image premultiply/unpremultiply

Issue #1130, opened 2025-07-28
[vello_cpu] How to draw pixmap on a pixmap?


T-256:
I'm trying to add `vello_cpu` renderer to iced. I started with same structure as `tiny_skia` has been implemented in iced. Currently I'm blocked by translating `tiny_skia::painter::PixmapMut` to its alternative via `vello_cpu::RenderContext`. the signature of function which allows draw new pixmap from specific point on top of another pixmap:
```rs
    pub fn draw_pixmap(
        &mut self,
        x: i32,
        y: i32,
        pixmap: PixmapRef,
        paint: &PixmapPaint,
        transform: Transform,
        mask: Option<&Mask>,
    ) 
```

My work is in https://github.com/T-256/iced/tree/vell_cpu_ar (currently only drawing quads works which is still buggy 🤷‍♂️ )

LaurenzV:
We definitely need to add an API to make that easier. 

Right now, the way to do it is to set an image paint and then draw a rectangle covering the part of the image you want to see visible. Perhaps the following examples can be helpful: https://github.com/linebender/vello/blob/main/sparse_strips/vello_sparse_tests/tests/image.rs

LaurenzV:
> (currently only drawing quads works which is still buggy 🤷‍♂️ )

Buggy due to bugs in `vello_cpu`?

T-256:
> Perhaps the following examples can be helpful: https://github.com/linebender/vello/blob/main/sparse_strips/vello_sparse_tests/tests/image.rs

I checked `image_with_multiple_clip_layers`, it may work but it'd be expensive and, imo, it's not correct way to implement.


> Buggy due to bugs in `vello_cpu`?

I'm just figuring out, I'm not sure, but as visually as I can see it's slower than tiny_skia. if I found any specific bug about `vello-cpu`, I'll fill new issue with details.

LaurenzV:
> it may work but it'd be expensive 

I'm not sure what you mean, what part is expensive? You don't need a clip layer for images, that's just one specific test.

> I'm not sure, but as visually as I can see it's slower than tiny_skia

If you are on an x86 machine, it's possible that `vello_cpu` is currently slower than `tiny-skia`, because we haven't implemented x86 SIMD yet (it's being worked on, though). However, if you are on ARM, then `vello_cpu` should be much faster than `tiny-skia`.

T-256:
[...]

PR #1256, opened 2025-10-01
[vello_cpu] Add a `draw_pixmap` method

LaurenzV:

1) Currently, the only way to draw an image is by drawing a rectangle with a pattern fill. If you know how to do it, you know it, but this is probably not the most intuitive way (as can be seen in #1130). Because of this, I think it's a good idea to add an explicit `draw_pixmap` method that can be called by the user, which should be much more clear.

2) Currently, there is the problem that we _have_ to define an extend for images, which makes sense for patterns but not necessarily for just drawing a single image. For example, I encountered the situation where when drawing a very small bitmap glyph using bilinear/biubic interpolation, it will basically look like a black block, because when sampling pixels outside we will apply an extend instead of sampling a transparent pixel, which would make more sense:
image

Because of this, when drawing an image we add a new mode that applies no extend at all and samples transparent pixels when going oob. This makes the letter a look nicer (the V looks a bit funky for different reasons):
image

This does have some impact on performance when rendering normal patterns (left is before, right after), but nothing that I would consider very too serious. The other option would be to introduce generic parameters, but that would make our already very generics-heavy code even more heavy, so I'd like to avoid that.

image

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -12,7 +12,6 @@ use crate::dispatch::single_threaded::SingleThreadedDispatcher;
 use crate::kurbo::{PathEl, Point};
 use crate::peniko::{ImageQuality, ImageSampler};
 use alloc::boxed::Box;
-#[cfg(feature = "text")]
 use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
@@ -21,7 +20,6 @@ use vello_common::encode::{EncodeExt, EncodedPaint};
 use vello_common::fearless_simd::Level;
 use vello_common::kurbo::{Affine, BezPath, Cap, Join, Rect, Stroke};
 use vello_common::mask::Mask;
-#[cfg(feature = "text")]
 use vello_common::paint::ImageSource;
 use vello_common::paint::{Image, Paint, PaintType};
 use vello_common::peniko::color::palette::css::BLACK;


LaurenzV:
Hmmm, I'm not actually sure now whether this is the intended way.... I will mark this as draft for now.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -524,6 +471,30 @@ pub struct EncodedImage {
     pub x_advance: Vec2,
     /// The advance in image coordinates for one step in the y direction.
     pub y_advance: Vec2,
+    /// Whether the extends in `ImageSampler` should be ignored.
+    pub ignore_extend: bool,
```
grebmeg: Actually, I have a similar task in my backlog 😅 In some cases like the one below, we’d prefer to render transparent pixels instead of using `Extend::Pad`. I think the more appropriate solution would be to add another extend `Transparent` [here](https://github.com/linebender/peniko/blob/15466cf836e2dd57c92ff90fe8a485b8d91b2929/src/brush.rs#L159-L167). What do you think?

Image

LaurenzV: Makes sense, but unfortunately we just cut a new peniko release yesterday. :/

grebmeg: No problem! We can include it in one of the next releases.

LaurenzV:
WIll split this PR into two.

LaurenzV:
I've removed the other change now, so hopefully it's easier to review!

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -169,6 +168,35 @@ impl RenderContext {
...
+    pub fn draw_pixmap(&mut self, pixmap: Arc<Pixmap>, image_quality: ImageQuality, alpha: f32) {
+        let (width, height) = (pixmap.width(), pixmap.height());
+        let image = Image {
+            image: ImageSource::Pixmap(pixmap),
+            sampler: ImageSampler {
+                x_extend: Default::default(),
+                y_extend: Default::default(),
+                quality: image_quality,
+                alpha,
+            },
```
grebmeg, line 186: Would it be better to pass `ImageSampler` as a parameter instead of the current `image_quality` and `alpha` values? As a consumer, I’d prefer having the flexibility to pass extends as well.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -169,6 +168,35 @@ impl RenderContext {
         }
     }
 
+    /// Draw an image into the pixmap. This is a convenience method that is functionally
+    /// equivalent to drawing a rectangle with the same size as the image and applying a pattern fill.
+    ///
+    /// By default, the image will be positioned with the top-left corner at the point (0, 0)
+    /// and the width and height set to the natural dimensions of the image. The positioning and
+    /// scaling of the image can be changed by applying a transform to the render context.
+    pub fn draw_pixmap(&mut self, pixmap: Arc<Pixmap>, image_quality: ImageQuality, alpha: f32) {
```
grebmeg, line 177: Currently, `vello_hybrid` can’t automatically upload pixmaps (it'd actually panic), but it does work with an `OpaqueId`. From the Unified API perspective, it might be better to use `ImageSource` here instead of a `pixmap`.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: I don't think that any other extend mode than pad makes sense, since the rectangle is always going to be the same size as the image. So the extend would at most apply to the anti-aliased pixels.

LaurenzV:
Hmm yeah, I'm not sure how necessary this method is, so I'll close it.

PR #1257, opened 2025-10-01
Update to Peniko v0.5.0

DJMcNab:
Release notes: https://github.com/linebender/peniko/releases/tag/v0.5.0

### DIFF
--- a/vello/src/scene.rs
+++ b/vello/src/scene.rs
@@ -89,9 +89,7 @@ impl Scene {
     ///
     /// **However, the transforms are *not* saved or modified by the layer stack.**
     /// That is, the `transform` argument to this function only applies a transform to the `clip` shape.
-    ///
-    /// Clip layers (`blend` = [`Mix::Clip`]) should have an alpha value of 1.0.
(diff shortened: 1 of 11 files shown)


vello/src/scene.rs
```diff
@@ -102,7 +100,10 @@ impl Scene {
     ) {
         let blend = blend.into();
         if blend.mix == Mix::Clip && alpha != 1.0 {
-            log::warn!("Clip mix mode used with semitransparent alpha");
+            log::warn!(
+                "Clip mix mode used with semitransparent alpha.\
+                This is wrong, and you should be using `push_clip_layer` instead."
```
nicoburns: If semitransparent alpha is passed, then they probably ought to be using `Mix::Normal`?

nicoburns (approved):
Looks good. An optional nit on one of the warnings, but I'd consider it non-blocking.

vello/src/scene.rs
DJMcNab: Yeah, that makes sense to me, I'll change it.

Realistically, people will have fixed the `deprecated` first, so probably will never see this now.

PR #1262, opened 2025-10-03
Support premultiplied images in vello classic

sagudev:


vello_encoding/src/encoding.rs
```diff
@@ -454,7 +451,8 @@ impl Encoding {
             .extend_from_slice(bytemuck::cast_slice(bytemuck::bytes_of(&DrawImage {
                 xy: 0,
                 width_height: (brush.image.width << 16) | (brush.image.height & 0xFFFF),
-                sample_alpha: ((brush.image.format as u32) << 14
+                sample_alpha: ((brush.image.alpha_type as u32) << 15
```
DJMcNab: This seems like a strange ordering; why encode the format in a constrained area? We know that the `AlphaType` is always just going to be a single bit, whereas it seems reasonable that we'd gain more image formats.

DJMcNab:
We also need to update the changelog to remove "Only `ImageAlphaType::Alpha` is currently supported in Vello."

vello_encoding/src/encoding.rs
sagudev: Indeed, I will fix this.

sagudev:
> Again, this should probably have a test.

That's why it's draft.

vello_encoding/src/encoding.rs
```diff
@@ -431,10 +431,7 @@ impl Encoding {
 
     /// Encodes an image brush.
     pub fn encode_image<'b>(&mut self, brush: impl Into<ImageBrushRef<'b>>, alpha: f32) {
-        let brush = brush.into();
-        if brush.image.alpha_type != peniko::ImageAlphaType::Alpha {
-            unimplemented!("Unsupported image alpha type: {:?}", brush.image.alpha_type);
-        }
+        let brush: ImageBrushRef<'b> = brush.into();
```
sagudev, line 434: I wrote type explicitly for RA, or else it cannot infer type and give good suggestions.

nicoburns: This is fine, but another option in these circumstances if `let brush = ImageBrushRef::from(brush);`

PR #1263, opened 2025-10-03
Port to latest Fearless SIMD commit

DJMcNab:
Note:
- I still need to port the uses of `simd_dispatch` to `dispatch`, but pushing so the other changes can be reviewed.
- The negation on `msub` *should* be optimised out on neon (see [#simd > Porting Vello to latest Fearless SIMD](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Porting.20Vello.20to.20latest.20Fearless.20SIMD/with/542926308))

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -6,7 +6,7 @@
 use crate::flatten_simd::Callback;
 use crate::kurbo::{self, Affine, PathEl, Stroke, StrokeCtx, StrokeOpts};
 use alloc::vec::Vec;
-use fearless_simd::{Level, Simd, simd_dispatch};
+use fearless_simd::{Level, Simd, dispatch};
 use log::warn;
 
 pub use crate::flatten_simd::FlattenCtx;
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
LaurenzV, line 1: Are the changes here really relevant to this PR? Maybe they would make sense as a spearate PR?

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -82,17 +82,9 @@ pub fn fill(
     line_buf: &mut Vec<Line>,
     ctx: &mut FlattenCtx,
 ) {
-    fill_dispatch(level, path, affine, line_buf, ctx);
+    dispatch!(level, simd => fill_impl(simd, path, affine, line_buf, ctx));
```
LaurenzV, line 85: Much nicer! 😄

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -302,9 +302,10 @@ mod alpha_fill {
     ) {
         let bg_c = f32x16::from_slice(s, dest);
         let mask_a = extract_masks(s, masks);
-        let inv_src_a_mask_a = one.msub(src_a, mask_a);
+        // 1 - src_a * mask_a
+        let inv_src_a_mask_a = -(src_a.msub(mask_a, one));
```
LaurenzV: Instead of using negation to be able to use `msub` here, can't we just replace all msub's we've had so far with `a.madd(-b, c)`, or am I missing something?

DJMcNab: Yes, we can! That optimises to a `vfmsq_f32` as expected as well. Good catch

CHANGELOG.md
DJMcNab: Hmm, I obviously was intending to start from main, but did so badly. The commit seems to be from #1261.

LaurenzV (approved):

PR #1266, opened 2025-10-03
Release Vello CPU v0.0.3

nicoburns:
Companion release that matches dependency versions with Vello 0.6.0 and Parley 0.6.0


DJMcNab (approved):
Happy to do this, but I want to get #1264 done first.

PR #1268, opened 2025-10-06
[classic] Support gradient interpolation in unpremultiplied alpha space

### LINKED ISSUE #1657: vello Classic ramp_cache interpolates gradients in premultiplied alpha; diverges from tiny_skia / SVG renderers for stops at differing alpha (split from #1056)
## Summary

`vello` Classic (via `vello_encoding::ramp_cache::make_ramp`) interpolates gradient stops in **premultiplied alpha space**, whereas `tiny_skia` (and per #1056, eventually `vello_hybrid`) interpolates in **straight alpha space**. For any gradient whose stops have non-uniform alpha, this produces a visibly different mid-gradient color — by tens of intensity levels per channel, not a precision artifact.

Splitting this off from #1056, which was opened against the original premultiplied behavior and closed when `vello_hybrid` switched. The Classic codepath was not updated; the last comment on #1056 said "working on fixing this in vello classic" but there is no tracking issue. This is that issue.

## Repro (no SVG)

A 256×32 rect filled with a linear gradient: cream (`#ffecbb` opaque) → 50%-alpha black, rendered via `vello::Renderer::render_to_texture` and `tiny_skia::PixmapMut::fill_rect` at the same size. Standalone, no `vello_svg` / `usvg` in the path.

```rust
let stops = vec![
    ColorStop { offset: 0.0, color: DynamicColor::from_alpha_color(
        Color::from_rgba8(0xff, 0xec, 0xbb, 0xff)) },
    ColorStop { offset: 1.0, color: DynamicColor::from_alpha_color(
        Color::from_rgba8(0x00, 0x00, 0x00, 0x80)) },
];
let gradient = Gradient::new_linear((0.0, 0.0), (256.0, 0.0))
    .with_stops(stops.as_slice());
scene.fill(
    Fill::NonZero,
    Affine::IDENTITY,
    &Brush::Gradient(gradient),
    None,
    &Rect::new(0.0, 0.0, 256.0, 32.0),
);
// ... render_to_texture into an Rgba8Unorm target, readback ...
```

Tiny-skia equivalent uses `tiny_skia::LinearGradient::new` with the same two stops and `fill_rect`.

## Output

Mid-gradient pixel (x = 128, y = 16):

|              | RGBA (as stored) | Storage format         | Un-premultiplied RGB |
| ------------ | ---------------- | ---------------------- | -------------------- |
| `vello`      | (170, 158, 124, 191) | Rgba8Unorm (straight) | (170, 158, 124)     |
| `tiny_skia`  | ( 95,  89,  70, 191) | Rgba8 premultiplied   | (127, 118,  93)     |

Normalizing both to *unpremultiplied* RGB at the mid pixel:

- vello: (170, 158, 124) — ≈ 89 % of the cream stop, lerped towards black via **premultiplied** path. Matches `lerp(premul((255,236,187,255)), premul((0,0,0,128)), 0.5) → un_premultiply` = (170, 157, 125).
- tiny_skia: (127, 118, 93) — straight RGB lerp `(255+0)/2, (236+0)/2, (187+0)/2`, alpha lerped separately to 191.

Mean abs diff over the full 256 × 32 RGBA = **33.54** (out of 255). Not AA noise.

## Root cause

`vello_encoding/src/ramp_cache.rs` `make_ramp` calls `AlphaColor::lerp`:

```rust
let c = if du < 1e-9 {
    this_c
} else {
    last_c.lerp(this_c, (u - last_u) / du, HueDirection::default())
};
c.premultiply().to_rgba8().to_u32()
```

`AlphaColor::lerp` in `color` 0.3.3 is defined as:

```rust
pub fn lerp(self, other: Self, t: f32, direction: HueDirection) -> Self {
    self.premultiply()
        .lerp(other.premultiply(), t, direction)
        .un_premultiply()
}
```

So the stops are premultiplied, lerped in premul space, un-premultiplied, then premultiplied again for storage — equivalent to a straight premultiplied lerp.

`Gradient::with_interpolation_alpha_space(InterpolationAlphaSpace::Unpremultiplied)` does **not** change this — `make_ramp` doesn't read that field. (Verified: I set it both ways from `vello_svg`, identical pixel output.)

## Why this matters

`vello_svg` users (consumed via `floem` for instance, which is how I hit this) get visibly different gradient renders from every other SVG renderer in the ecosystem — `resvg`/`tiny_skia`, Inkscape, Firefox, Chromium. The affected case is "gradient stops at differing alpha", which is common in stock app icons (the originating case was the `qv4l2.svg` shipped with `v4l-utils`). See https://github.com/linebender/vello_svg/issues/83 for the full bisection across a 49-icon XDG corpus.

## Suggested fix

Either:

1. Have `make_ramp` do component-wise straight-RGBA lerp on `(r, g, b, a)` directly, premultiplying only at the `to_rgba8()` step. Simplest, matches `tiny_skia` and SVG renderers in practice.
2. Honour `Gradient::interpolation_alpha_space` in `make_ramp` and pick path 1 when it's `Unpremultiplied`. Slightly more work, preserves both behaviours.

Happy to do (1) as a PR if maintainers prefer that direction — but I don't know whether the original #1056 thread reached consensus on whether vello Classic should default to straight or premultiplied (CSS Color Module 4 says premultiplied; HTML Canvas / SVG-in-practice say straight). Pointer appreciated.

Repro binary lives at https://github.com/mxaddict/vello_svg/blob/fix/gradient-transform-composition/icon-corpus/src/alpha_ramp_repro.rs — `cargo run -p icon-corpus --bin alpha-ramp-repro --release` from the fork's root reproduces the table.

sagudev:
Fixes #1657 (which is #1056 for vello classic).


vello_encoding/src/ramp_cache.rs
```diff
@@ -104,7 +125,17 @@ fn make_ramp(stops: &[ColorStop]) -> impl Iterator<Item = u32> + '_ {
...
-            last_c.lerp(this_c, (u - last_u) / du, HueDirection::default())
+            match interpolation_alpha_space {
+                InterpolationAlphaSpace::Premultiplied => {
+                    last_c.lerp(this_c, (u - last_u) / du, HueDirection::default())
+                }
+                InterpolationAlphaSpace::Unpremultiplied => {
+                    AlphaColor::<Srgb>::new(std::array::from_fn(|i| {
+                        last_c.components[i]
+                            + (this_c.components[i] - last_c.components[i]) * ((u - last_u) / du)
+                    }))
```
sagudev: NOTE: lerp on AlphaColor in color converts colors to premultiplied does interpolation and then unpremultiplied (interpolation happens in premultiplied alpha space).

nicoburns:
We should publish a patch release of `peniko` with https://github.com/linebender/peniko/pull/148 before merging this to avoid a git dependency on `peniko` (#AlwaysReleasableMainBranch)

sagudev:
> We should publish a patch release of `peniko` with [linebender/peniko#148](https://github.com/linebender/peniko/pull/148) before merging this to avoid a git dependency on `peniko` (#AlwaysReleasableMainBranch)

Yes, that's why it's draft.

mxaddict:
Heads up — verified peniko 0.6.1 (published 2026-05-15) ships the `Eq`+`Hash` derives on `InterpolationAlphaSpace` from peniko#148, and both vello `main` and this PR already pin `peniko = "0.6.1"`, so the `#AlwaysReleasableMainBranch` blocker called out by @nicoburns is resolved. PR is non-draft, mergeable, all CI green. Would unblock #1657 / fix the alpha-stop divergence flagged at https://github.com/linebender/vello_svg/issues/83 — happy to test the icon corpus against it once it lands. Friendly nudge for a maintainer review when convenient.

vello_tests/tests/known_issues.rs
DJMcNab, line 1: I'm a little bit confused by this new test. Shouldn't this pr be making it work? If so, why is it a known issue?

vello_encoding/src/ramp_cache.rs
```diff
@@ -104,7 +125,14 @@ fn make_ramp(stops: &[ColorStop]) -> impl Iterator<Item = u32> + '_ {
         let c = if du < 1e-9 {
             this_c
         } else {
-            last_c.lerp(this_c, (u - last_u) / du, HueDirection::default())
+            match interpolation_alpha_space {
+                InterpolationAlphaSpace::Premultiplied => {
+                    last_c.lerp(this_c, (u - last_u) / du, HueDirection::default())
+                }
+                InterpolationAlphaSpace::Unpremultiplied => {
+                    last_c + (this_c - last_c) * ((u - last_u) / du)
```
DJMcNab: Small style nit: could we do `let t = (u - last_u) / du`? I don't particularly care if that's just in this branch or shared between both.

This isn't blocking. And it's likely only because of how this wraps on mobile...

vello_tests/tests/known_issues.rs
```diff
@@ -137,6 +138,38 @@ fn test_gradient_color_alpha() {
...
+                ColorStop {
+                    offset: 1.,
+                    color: Color::from_rgba8(0, 0, 255, 255).into(),
+                },
+            ])
+            .with_interpolation_alpha_space(InterpolationAlphaSpace::Unpremultiplied),
+        None,
+        &viewport,
+    );
+    let mut params = TestParams::new("gradient_color_alpha_unpremultiplied", 100, 50);
```
DJMcNab: A comment that the thing to observe is a yellow hue to the white would be good.

Ideally, just having the same test but for premultiplied would also do it.

DJMcNab (approved):
The only blocking concern are the small thoughts on testing.

Sorry this had stalled; I didn't see that this had left draft :(

vello_tests/tests/known_issues.rs
sagudev: we have premultiplied test above already

vello_tests/tests/known_issues.rs
sagudev: premultiplied test also works and it's here

mxaddict:
So happy to see this moving again <3

mxaddict:
Thanks for the hard work guys!

vello_encoding/src/ramp_cache.rs
```diff
@@ -21,7 +21,7 @@ pub struct Ramps<'a> {
 #[derive(Default)]
 pub(crate) struct RampCache {
     epoch: u64,
-    map: HashMap<CacheKey<ColorStops>, (u32, u64)>,
+    map: HashMap<(InterpolationAlphaSpace, CacheKey<ColorStops>), (u32, u64)>,
```
waywardmonkeys: Feels like we should have a type here for the ramp cache key rather than a tuple. Maybe.

DJMcNab: I had the same thought. I don't think that blocks this pr, and I also don't think it's super meaningfully better. However, it would make future extensions that much easier (e.g. for interpolation colour space).

DJMcNab (approved):
Thanks for moving the test. This is good to land, pending Bruce's comment.

vello_encoding/src/ramp_cache.rs
sagudev: I just shamelessly stole https://github.com/linebender/vello/blob/1245d3e1a3cbef32b6d7d5156fb30ff4a43207c0/sparse_strips/vello_common/src/encode.rs#L795 for https://github.com/linebender/vello/pull/1268/commits/90cfbba2f9d9d305caec56d53aaf2a5cb028a8dc

PR #1271, opened 2025-10-10
Prepare sparse-strips-v0.0.4 release

grebmeg:
We'll do this once [fearless_simd release](https://github.com/linebender/fearless_simd/pull/107/files) is out.

### DIFF
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -47,6 +47,7 @@ This is the initial release. No changelog was kept for this release.
 
 See also the [vello_cpu 0.0.1](../vello_cpu/CHANGELOG.md#001---2025-05-10) release.
 
+[Unreleased]: https://github.com/linebender/fearless_simd/compare/sparse-strips-v0.0.4...HEAD
 [0.0.4]: https://github.com/linebender/vello/compare/sparse-stips-v0.0.3...sparse-strips-v0.0.4
 [0.0.3]: https://github.com/linebender/vello/compare/sparse-stips-v0.0.2...sparse-strips-v0.0.3
 [0.0.2]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.1...sparse-stips-v0.0.2
--- a/sparse_strips/vello_cpu/CHANGELOG.md
+++ b/sparse_strips/vello_cpu/CHANGELOG.md
@@ -47,6 +47,7 @@ This is the initial release. No changelog was kept for this release.
 
 See also the [vello_common 0.0.1](../vello_common/CHANGELOG.md#001---2025-05-10) release.
 
+[Unreleased]: https://github.com/linebender/fearless_simd/compare/sparse-strips-v0.0.4...HEAD
 [0.0.4]: https://github.com/linebender/vello/compare/sparse-stips-v0.0.3...sparse-strips-v0.0.4
 [0.0.3]: https://github.com/linebender/vello/compare/sparse-stips-v0.0.2...sparse-strips-v0.0.3
 [0.0.2]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.1...sparse-stips-v0.0.2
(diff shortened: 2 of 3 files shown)


sparse_strips/vello_common/CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+Subheadings to categorize changes are `added, changed, deprecated, removed, fixed, security`.
+
+-->
+
+# Changelog
+
+The latest published vello_common release is [0.0.4](#004---2025-10-14) which was released on 2025-10-14.
+You can find its changes [documented below](#004---2025-10-14).
+
+## [Unreleased]
```
grebmeg, line 14: I was thinking it might be a good idea to start keeping changelogs? It doesn’t take much time to update when we make changes, especially bigger ones, but it really helps with keeping track of what’s going on and makes the project more transparent. What do you think?

sparse_strips/vello_common/CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+
+## [0.0.2][] - 2025-09-22
+
+This release has an [MSRV][] of 1.85.
+
+## [0.0.1][] - 2025-05-10
+
+This release has an [MSRV][] of 1.85.
+
+- Initial release
```
grebmeg: As for this, I think I can give it a try and have AI fill everything in accordingly. Hopefully it’ll handle that part easily!

LaurenzV: If we really want to do the changelog, I think it's fine to just do it from now on? Not sure we really need to fill in the past ones as well.

nicoburns: If we're on a synced release schedule we might want to do a combined changelog? (I'm thinking less overhead for releases).

grebmeg: > If we really want to do the changelog, I think it's fine to just do it from now on? Not sure we really need to fill in the past ones as well.

I agree, I think we can go that route, but my perfectionism is making me want to give it a try anyway 😅 I’ll timebox it to 30 minutes to add just the main points; if it takes longer than that, we’ll stick with starting from now on.

> If we're on a synced release schedule we might want to do a combined changelog? (I'm thinking less overhead for releases).

Could you elaborate on what kind of overhead you mean?
I’m happy with either approach, but personally I think managing separate changelogs per crate and then combining them only at release time makes a bit more sense to me. It gives more accurate version history per crate, lets each crate link directly to its own changelog on crates.io, and keeps things cleaner and more focused for users who only care about one crate; right now we’re on a synced release schedule, but would this still hold if we move toward an independent release cadence? And lastly, I feel like this approach scales a bit better in the long run.

nicoburns: Just things like updating the date if the date of the release changes. But if you want to keep separate changelogs, then by all means let's give it a go.

DJMcNab: This seems like it would not be a very efficient use of reviewer time, as:
- A huge amount of things have changed
- An LLM will get the summarisation wrong, it's just a question of how wrong.
- There is *no* possible user of the actual changelog contents for these historical releases.

I'd be much happier with text like "No changelog was kept for this release", as in:
- https://github.com/linebender/fearless_simd/blob/main/CHANGELOG.md#011-2018-11-05
We should link to the corresponding changelog version section for Vellos CPU and Hybrid in each, though.

Separately, keeping a seperate changelog for each of Vello CPU and Vello Hybrid does seem reasonable.
I think that keeping one for Vello Common doesn't really make sense; it's still not intended for consumption except as an implementation detail of the former.

grebmeg: Thanks everyone for the feedback! 🙏

I think for now I’ll leave all three changelogs as I find it useful to have one for `vello_common` as well, for understanding historical changes without needing to dig through commits. It helps provide context for how things have evolved over time.

Let’s live with this setup for a bit and see how useful and convenient it actually is. If it turns out to be unnecessary or cumbersome, we can always iterate and adjust to whatever works best.

PR #1272, opened 2025-10-11
Improve robustness when drawing large paths on large pixmaps

LaurenzV:


sparse_strips/vello_common/src/coarse.rs
```diff
@@ -277,14 +277,17 @@ impl<const MODE: u8> Wide<MODE> {
...
             // Calculate which wide tiles this strip intersects
             let wtile_x0 = (x0 / WideTile::WIDTH).max(bbox.x0());
             // It's possible that a strip extends into a new wide tile, but we don't actually
             // have as many wide tiles (e.g. because the pixmap width is only 512, but
             // strip ends at 513), so take the minimum between the rounded values and `width_tiles`.
-            let wtile_x1 = x1.div_ceil(WideTile::WIDTH).min(bbox.x1());
+            let wtile_x1 = x1
+                .div_ceil(WideTile::WIDTH)
+                .min(bbox.x1())
+                .min(u16::MAX / WideTile::WIDTH);
```
grebmeg: I see `u16::MAX / WideTile::WIDTH` used in a few places, maybe we could add it as a constant like `MAX_COORDINATE` (or more appropriate name) in `WideTile`?

grebmeg: If `x1` is already saturated and `bbox.x1` is saturated to `u16::MAX / WideTile::WIDTH`, do we still need line 290?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -71,7 +71,19 @@ impl Tile {
...
+            // with the tile width and height during strips generation.
+            x.min(u16::MAX / Self::WIDTH),
+            y.min(u16::MAX / Self::HEIGHT),
+            line_idx,
+            winding,
+        )
+    }
+
+    #[inline]
+    pub(crate) const fn new_const(x: u16, y: u16, line_idx: u32, winding: bool) -> Self {
```
grebmeg: Maybe the name `new_unchecked` could be a bit more descriptive? Since it’s a `const`, you’re already implying something like `new_const`. Or, if you prefer, `new_checked` and just `new`?

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Admittedly the test case still seems to pass if I remove it, but I think conceptually it should still be there, because `x1` refers to the x coordinate of a normal 4x4 tile, and since we do `div_ceil` I think it's possible that we end up with a larger `wtile_x1` then should be.

sparse_strips/vello_common/src/tile.rs
LaurenzV: How about `new_clamped`  and `new`?

grebmeg: Sounds good! Though personally, I think `checked_*` feels more idiomatic here, as in Rust, a `checked_*` function typically performs an operation that might overflow or fail but returns `Option` instead of panicking. That said, I was probably wrong in suggesting `checked`, since `saturating_*` actually fits better in our case (clamps to max). Anyway, I’m totally fine with your suggestion!

LaurenzV: I'll go with the clamped then. `checked` would arguably be the best, but I want to avoid having to deal with an `Option` everywhere, and `new_saturating` sounds a bit weird to me. 😄

PR #1273, opened 2025-10-11
Pretranslate Bitmap emoji to mirror Colr path

ArthurCose:
Found this while trying to figure out why emojis were rendering in the wrong place in [Floem](https://github.com/lapce/floem) with the vello renderer

### Testing

Clone Floem's main branch (this commit as of writing): https://github.com/lapce/floem/tree/6148cd8f915e25882052c3497aba08c188f2df46

Run the widget-gallery example with any display scale other than 100% (I use 112.5%):
image

Enter some text with a few emojis, such as `maskldmasldmsakdasldmsak🫠🫠🫠asdmsald`, into the Input example:
image

After switching to `pre_translate`  by using a local edit of `vello` and `vello_svg`:
image

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -10,6 +10,7 @@
 [default.extend-identifiers]
 thm = "thm"
 wdth = "wdth"
+ArthurCose = "ArthurCose"
 
 # Match Inside a Word - Case Insensitive
 [default.extend-words]


ArthurCose:
Just added a changelog entry, hopefully it looks right.

DJMcNab:
Thank you. The changelog entry was perfect! I've pushed a change to permit your name, and will now land this PR; that's an unfortunate failure mode...

PR #1280, opened 2025-10-28
Update to wgpu 27.0.1, wgpu-profiler 0.25

theoparis:
Bumped the MSRV to 1.88 and updated wgpu and wgpu-profiler.

Unfortunately I was unable to run the tests because of this error in libdeflate-sys:
```
cargo:warning=libdeflate/lib/x86/adler32_template.h:295:14: error: always_inline function '_mm512_dpbusd_epi32' requires target feature 'avx512vnni', but would be inlined into function 'adler32_x86_avx512_vl512_vnni' that is compiled without support for 'avx512vnni'
  cargo:warning=  295 |                                 v_s2   = VDPBUSD(v_s2,   data_a, mults);
  cargo:warning=      |                                          ^
  cargo:warning=libdeflate/lib/x86/adler32_template.h:109:28: note: expanded from macro 'VDPBUSD'
  cargo:warning=  109 | #  define VDPBUSD(a, b, c)      _mm512_dpbusd_epi32((a), (b), (c))
  ...
```
It seems like oxipng depends on libdeflater, and libdeflater hasn't had a release in a while.

However vello seems to compile succesfully and I am able to run my other project with this branch as a dependency.

### DIFF
--- a/sparse_strips/vello_hybrid/src/multi_atlas.rs
+++ b/sparse_strips/vello_hybrid/src/multi_atlas.rs
@@ -169,14 +169,14 @@ impl MultiAtlasManager {
             }
         }
 
-        if let Some(idx) = best_atlas_idx {
-            if let Some(allocation) = self.atlases[idx].allocate(width, height) {
-                let atlas_id = self.atlases[idx].id;
-                return Ok(AtlasAllocation {
-                    atlas_id,
-                    allocation,
-                });
-            }
+        if let Some(idx) = best_atlas_idx
+            && let Some(allocation) = self.atlases[idx].allocate(width, height)
+        {
+            let atlas_id = self.atlases[idx].id;
+            return Ok(AtlasAllocation {
+                atlas_id,
+                allocation,
+            });
         }
 
         // Fallback to first-fit if least-used didn't work
@@ -366,12 +366,6 @@ impl AtlasId {
     }
 }
 
-impl Default for AllocationStrategy {
-    fn default() -> Self {
-        Self::FirstFit
-    }
-}
-
 /// Usage statistics for an atlas.
 #[derive(Debug, Clone)]
 pub(crate) struct AtlasUsageStats {
@@ -439,9 +433,10 @@ impl Default for AtlasConfig {
 }
 
 /// Strategy for allocating images across multiple atlases.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub enum AllocationStrategy {
     /// Try atlases in order until one has space.
+    #[default]
     FirstFit,
     /// Choose the atlas with the smallest remaining space that can fit the image.
     BestFit,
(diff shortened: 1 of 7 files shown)


DJMcNab:
Welcome!

Bumping MSRV is fine, especially for wgpu support; there are instructions in CI.yml, and also in Cargo.toml. Obviously CI has also spotted a missed migration to the new timeout aware polling support, which needs to be fixed.

If you need to run the tests locally (they run fine on CI, so it's okay if you can't), are you able to try a different C compiler?

raphlinus:
The underlying issue seems to be https://github.com/libdeflater/libdeflater/issues/41. Not sure what the best path forward is.

theoparis:
Yeah, it seems like there is a fix in the master branch of libdeflate, but it's not in a release yet. I was using a trunk build of clang, but I think the issue may be present with GCC 16 as well 🤔

DJMcNab:
The MSRV bump is leading to suggestions from Clippy to use let chains.
You should be able to just apply those directly, or we can move those into a different PR; let us know which you'd prefer.

vello/README.md
```diff
@@ -7,7 +7,7 @@
 [![Latest published version.](https://img.shields.io/crates/v/vello.svg)](https://crates.io/crates/vello)
 [![Documentation build status.](https://img.shields.io/docsrs/vello.svg)](https://docs.rs/vello)
 [![Apache 2.0 or MIT license.](https://img.shields.io/badge/license-Apache--2.0_OR_MIT-blue.svg)](#license)
-[![Required wgpu version.](https://img.shields.io/badge/wgpu-v26.0.1-orange.svg)](https://crates.io/crates/wgpu)
+[![Required wgpu version.](https://img.shields.io/badge/wgpu-v27.0.1-orange.svg)](https://crates.io/crates/wgpu)
```
DJMcNab, line 10: Thanks for catching this - we've missed this before a few times we've done these updates.

sparse_strips/vello_hybrid/src/multi_atlas.rs
```diff
@@ -439,9 +433,10 @@ impl Default for AtlasConfig {
 }
 
 /// Strategy for allocating images across multiple atlases.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub enum AllocationStrategy {
     /// Try atlases in order until one has space.
+    #[default]
```
DJMcNab, line 439: This change being here is a little bit surprising to me. Is it a new lint?

DJMcNab (approved):
Thanks! I've just pushed a small change, and I'll land this now

Edit: Oh, you've disabled maintainers editing the PR, so I'll do it in a follow-up

sparse_strips/vello_hybrid/src/multi_atlas.rs
sagudev: Yes, I've hit it when updating servo to 1.91.

PR #1282, opened 2025-11-04
Optimize `take_unpremultiplied` for in-place conversion

JustForFun88:
The `Pixmap::take_unpremultiplied` method was reworked to perform the RGBA conversion in-place, without allocating a new buffer to copy the entire image data (which can be very large in some cases).

This change reduces memory usage for large images and improves efficiency by skipping redundant calculations for fully transparent and fully opaque pixels.

A test was added to verify the correctness of the conversion.

### DIFF
--- a/sparse_strips/vello_common/Cargo.toml
+++ b/sparse_strips/vello_common/Cargo.toml
@@ -17,7 +17,7 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = []
 
 [dependencies]
-bytemuck = { workspace = true, features = [] }
+bytemuck = { workspace = true, features = ["extern_crate_alloc"] }
 peniko = { workspace = true, features = ["bytemuck"] }
 fearless_simd = { workspace = true }
 hashbrown = { version = "0.15", optional = true }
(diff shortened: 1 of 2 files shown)


DJMcNab (commented):
Welcome!

I am however slightly surprised by this PR; I understand that this code is written to take advantage of the hardcoded optimisation in the standard library to re-use allocations in such cases. See https://doc.rust-lang.org/stable/src/alloc/vec/in_place_collect.rs.html.
Have you validated in the codegen that the intermediate allocation actually happens?

If that optimisation does apply as I suspect, a comment to that affect would be a welcome addition, and I'd suggest changing this PR to be adding that.

JustForFun88:
> Welcome!
> 
> I am however slightly surprised by this PR; I understand that this code is written to take advantage of the hardcoded optimisation in the standard library to re-use allocations in such cases. See https://doc.rust-lang.org/stable/src/alloc/vec/in_place_collect.rs.html. Have you validated in the codegen that the intermediate allocation actually happens?
> 
> If that optimisation does apply as I suspect, a comment to that affect would be a welcome addition, and I'd suggest changing this PR to be adding that.

Thanks a lot! You were right - my apologies, I should have checked the asm output first. I see now that the original code doesn't have an extra allocation, thanks to the `in_place_collect optimization`. Even worse, my version actually broke the vectorization, so it's slower. 🤦

Appreciate you catching that! Closing this PR.

PR #1293, opened 2025-11-13
Add tile intersection checking

b0nes164:
This PR adds generation of intersection data to `make_tiles.` 

For our MSAA calculations, we need the points where a line intersects with a tile. Since the end goal is to perform the MSAA in parallel, we need these intersections to be watertight between different threads.  By performing these calculations in `make_tiles`, we can create a source of ground truth, ensuring watertightness.

Although it is feasible to calculate the exact intersection points in `make_tiles`, we defer that work to the gpu. Instead we create an intersection bitmask, which unambiguously defines which edges the line intersects. We say a line *intersects* the edge of a tile if it touches that edge AND continues into another tile. A endpoint that exactly touches an edge does NOT count as intersecting, though it may still contribute to winding (in the case of a top edge touch).

The bitmask is 5 bits, and will be consumed downstream by the rasterization stage:
`P(erfect) | R(ight) | L(eft) | B(ottom) | T(op)`
The lower 4 bits correspond to the intersection edges. The "Perfect" bit is necessary to resolve ambiguity in cases where a line perfectly intersects with a corner of a tile. 

---
Cases of 3 bit ambiguity can be resolved by always calculating intersections on the opposing edges of the tile.
Consider a tile with intersections `T | L | B`:
```
o--------+
|\       |
| \      |
|  \     | 
+---o----+
```
Calculate the intersections on `T` and `B`!

---

Cases of 2 bit ambiguity require the "Perfect" bit, which is set when there is exactly one unique edge intersection.
Consider a tile with intersections `T | L`:

This is valid:
```
o--------+
|\       |
| o      |
|        | 
+--------+
```

But so is this:
 ```
+--o-----+
| /      |
o        |
|        | 
+--------+
```

With the perfect bit, the first case would be `P | T | L` and the second case would be `T | L`

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -89,14 +89,17 @@ impl Tile {
     }
 
     /// The base tile constructor
+    ///
+    /// Unlike [`Self::new_clamped`], this constructor stores `x` and `y` exactly as provided.
+    /// Callers must ensure these coordinates do not exceed the limits required by downstream
+    /// processing (typically `u16::MAX / WIDTH` and `u16::MAX / HEIGHT`).
     #[inline]
     pub const fn new(x: u16, y: u16, line_idx: u32, intersection_mask: u32) -> Self {
         #[cfg(debug_assertions)]
         if line_idx >= MAX_LINES_PER_PATH {
             panic!("Max. number of lines per path exceeded.");
         }
-        // We pack the winding bit as the MSB of the high section (Bit 31).
-        // The intersection_mask is expected to contain bits 0-4 (T, B, L, R, P).
+        // The intersection_mask is expected to contain bits 0-5 (T, B, L, R, P, W).
         Self {
             x,
             y,
@@ -377,7 +380,7 @@ impl Tiles {
                 line_bottom_y.floor() as i32
             };
 
-            // special-case out lines which are fully contained within a tile.
+            // Special-case out lines which are fully contained within a tile.
             let not_same_tile = p0_tile_y != p1_tile_y || p0_tile_x != p1_tile_x;
             if not_same_tile {
                 // For ease of logic, special-case purely vertical tiles.
@@ -399,7 +402,7 @@ impl Tiles {
                     }
 
                     // Middle
-                    // If the start was culled, the first tile inside the viewport is a middle
+                    // If the start was culled, the first tile inside the viewport is a middle.
                     let y_start = if is_start_culled {
                         y_top_tiles
                     } else {
@@ -421,7 +424,7 @@ impl Tiles {
                             self.tile_buf.push(tile);
                         }
 
-                        // Perfect touching B case
+                        // Perfect touching B case.
                         let winding = ((f32::from(y_last) >= line_top_y) as u32) << 5;
                         let intersection_mask = if GEN_INT_MASK {
                             let is_end_tile = ((y_last as i32) == p1_tile_y) as u32;
@@ -457,7 +460,7 @@ impl Tiles {
                         (x_idx as i32 == p1_tile_x) && (y_idx as i32 == p1_tile_y)
                     };
 
-                    // Line walks rows top to bottom, left to right, y-exclusive, x-inclusive
+                    // Line walks rows top to bottom, left to right, y-exclusive, x-inclusive.
                     for y_idx in y_top_tiles..y_bottom_tiles {
                         let y = f32::from(y_idx);
 
@@ -509,15 +512,15 @@ impl Tiles {
                                 mask |= vert_exit << 1;
                                 mask |= hor_exit << dx_dir << 2;
 
-                                // Check if the line passes through any of the four corners of this tile.
-                                // It passes through a corner if the x-intersection with the top or bottom
-                                // edge equals either the left (x_start) or right (x_start + 1) tile boundary
+                                // Check if the line passes through any of the four corners of this
+                                // tile. It passes through a corner if the x-intersection with the
+                                // top or bottom edge equals either the left (x_start) or right
+                                // (x_start + 1) tile boundary.
                                 let x_start_f = x_start as f32;
                                 let x_right_f = (x_start + 1) as f32;
 
                                 // TODO Verify all cases of bottom left tiebreaking. Preliminary
                                 // testing appears to be fine.
-                                // TODO SIMD here!
                                 let trc = ((line_row_top_x == x_right_f) as u32) & !start_tile;
                                 let tlc = ((line_row_top_x == x_start_f) as u32) & !start_tile;
                                 let brc = ((line_row_bottom_x == x_right_f) as u32) & !end_tile;
@@ -525,8 +528,8 @@ impl Tiles {
                                 // The top left corner must be treated specially.
                                 let tie_break = tlc & (canonical_row_start ^ 1);
 
-                                // If we touch corners, force the corresponding horizontal intersection
-                                // bits to ensure that no intersection gets destructed.
+                                // If we touch corners, force the corresponding horizontal
+                                // intersection bits to ensure that no intersection gets destructed.
                                 mask |= (tie_break | blc) << 2;
                                 mask |= (trc | brc) << 3;
 
@@ -582,7 +585,6 @@ impl Tiles {
                                 let x_end_f = x_end as f32;
                                 let x_right_f = (x_end + 1) as f32;
 
-                                // TODO SIMD here!
                                 let trc = ((line_row_top_x == x_right_f) as u32) & !start_tile;
                                 let tlc = ((line_row_top_x == x_end_f) as u32) & !start_tile;
                                 let brc = ((line_row_bottom_x == x_right_f) as u32) & !end_tile;

### CODE REVIEW

**Comment by LaurenzV** [comment]
I ran this through my PDF test suite and there only seem to be small unnoticeable single-pixel differences, so no regressions. 👍 However, I won't be able to take a closer look at this before Monday.

But just one thing I noticed, to me, it does seem like this will add some processing time for a single tile. Perhaps we could add a const generic to the method indicating whether tile intersection data should be computed, and when false it just sets it to 0? This way, we can skip all of the calculations for vello_cpu, where it's not needed.

Also, this PR would make https://github.com/linebender/vello/pull/1211 irrelevant, right?

**Comment by b0nes164** [comment]
> However, I won't be able to take a closer look at this before Monday. 

No worries! Take your time.

> But just one thing I noticed, to me, it does seem like this will add some processing time for a single tile.

Yes there is a special case for a single tile.   

>  This way, we can skip all of the calculations for vello_cpu

Yes, I was thinking the same thing. Not super familiar rust, so not sure what's the rustiest way to do this, but I think even two different `make_tiles` could work.

> Also, this PR would make https://github.com/linebender/vello/pull/1211 irrelevant, right?

I think so. I was debating whether it would be worth it to case out more fast paths, but as mentioned above, I do have a "line completely enclosed in tile" case.

**Review by tomcur — COMMENTED** [review]
Cool stuff. The packing is clever and looks good.

Your explanation of the packing is great, I think it would be worthwhile to add that as documentation to the `packed_winding_line_idx`.

> Cases of 3 bit ambiguity can be resolved by always calculating intersections on the opposing edges of the tile.

Perhaps if you add the documentation to the code, also mention the 4-bit case (where, if my understanding is correct, you just calculate for any two opposing sides).

To be able to properly review the core idea itself I probably need a bit more background on the needs of MSAA, would you have some pointers?

> This way, we can skip all of the calculations for vello_cpu
>
> Yes, I was thinking the same thing.

Conditionally running this seems like a good idea. Here's an example of the `const`-generic pattern: https://github.com/linebender/parley/blob/4f887570bac98c6693e4a3c6937eebb3423dde72/parley/src/layout/alignment.rs#L84-L97, but two versions also makes sense, especially if the two versions can be optimized differently.

**Comment by LaurenzV** [comment]
I think by default I would prefer a single version, and only if it really turns out that having two separate methods is faster do that. Otherwise, when changing stuff in the logic we need to always remember to do it in both methods.

**Comment by b0nes164** [comment]
Re:All
I will work in your comments, eta tomorrow.

If I understand it, a `const-generic` is the equivalent of a C++ non-typed templating? I.e. the templated/const-generic variable is compile-time visible, and if we use a bool as the variable we can conditionally compile what we need?

This would work, but I expect it may be a little messy, as the function cannot be split cleanly. I will include this in the version tomorrow, so we can get a picture of what it would look like.

--- 
Re:Tom

> Perhaps if you add the documentation to the code, also mention the 4-bit case (where, if my understanding is correct, you just calculate for any two opposing sides).

Yes this is correct.

> To be able to properly review the core idea itself I probably need a bit more background on the needs of MSAA, would you have some pointers?

Yes apologies!

Conceptually, the MSAA-version of sparse-strips is almost identical to the analytic version but instead of calculating the coverage-mask through the area of the trapezoid formed by the line, we instead iterate through *N* subpixel sampling locations and determine if they are "inside" or "outside" the line. You then take the bitmask of these inside/outsides, count them, and turn that into coverage.

For example (note there is a bug in the topmost-left tile):
image

Naively this can be done by maintaining a winding number per subpixel sample, but because we want to do this in parallel, communicating 8, 16, 32 windings across threads becomes prohibitively costly.

Instead, we can minimize the communication required to a single winding number per tile---the same coarse winding used in the analytic version---by applying three rules. For simplicity we'll assume even/odd fill rule, for a single non-overlapping tile:

For any given pixel in a tile, the winding is the XOR of the coarse winding number for the whole tile, with the left edge intersection, with the per-pixel calculations.
`pixel[x][y] = coarse_winding ^ left_edge_intersection ^ per_pixel_calculations`

1) The coarse winding we get from the tiling.

2) The left-edge-intersection rule is: If the line intersecting a tile intersects its left edge, then left_edge_intersection is true for all pixel rows below the pixel row the line intersected. IF the line perfectly intersects the top-left corner, this requires tie-breaking logic, which is not implemented yet. (I realized today that the current logic is insufficient). 

3) The per-pixel-calculations involve determining which pixels the line intersects, then getting the subpixel sample mask, but I am keeping this deliberately vague, as it doesn't relate to the tiling.

Example, the left-edge-intersection in action:
image

Example of the current bug with top left intersection:
image

Because the left-edge-intersection seeds an entire row of pixels, any discrepancy in the left-edge pixel intersection between tiles (due to floating point errors) is potentially catastrophic.  So the idea is to take advantage of the fact that we fully traverse the line during the tiling to create a source of ground-truth for the compute-shader to use.

---

Instead of going straight into the code, I think the best to way to review this is to go into the test cases, and see if the results match what you would expect. Or add your own test case and see what you get.

**Comment by b0nes164** [comment]
Added changes as discussed, sans the additional comments. I deferred adding comments since the top-left intersection logic needs to be reworked to address the issue mentioned above. Previously, I had planned on doing all top-left corner tie-breaking logic downstream. However, I need to reconsider this decision because of the bug...

**Comment by LaurenzV** [comment]
Sorry for the delay, I hope that I will be able to take a look in the next few days.

**Comment by b0nes164** [comment]
@tomcur 

I'm currently working on an overhaul to the intersections which will change the intersection data mask. Laurenz is working on a patch to the blender2d benchmarking suite to better integrate it with vello.

We would like to get both done, and then benchmark the performance before moving forward, so for now there is no need to review.

**Comment by b0nes164** [comment]
This change is ready for review.

**Comment by LaurenzV** [comment]
Do you think it would be possible to rebase onto main (or just merge it into this branch)? Seems like there is only one file with conflicts, so hopefully it won't be too much of a pain. It would make it a bit easier to compare the performance of this branch against main. 🙂

**Comment by LaurenzV** [comment]
Reran the benchmarks. Not sure why, but performance seems to actually have improved slightly! Maybe because of the special-casing for 1x1 tiles?

image
image
image

So from that side we should be good to go.

**Comment by LaurenzV** [comment]
PDF test suite also passes. 👍

File: sparse_strips/vello_common/src/tile.rs [hunk 1]
```diff
@@ -72,27 +77,30 @@ impl Tile {
...
             // with the tile width and height during strips generation.
             x.min(u16::MAX / Self::WIDTH),
             y.min(u16::MAX / Self::HEIGHT),
             line_idx,
-            winding,
+            intersection_mask,
         )
     }
 
+    /// The base tile constructor
```
LaurenzV, line 91:
```suggestion
    /// The base tile constructor.
```

LaurenzV, line 91: Since we are making this public maybe we should document the max x/y values.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -267,76 +324,304 @@ impl Tiles {
...
-                    // The line's x-coordinates at the line's top- and bottom-most points within the
-                    // tile row.
-                    let line_row_top_x = p0_x + (line_row_top_y - p0_y) * x_slope;
-                    let line_row_bottom_x = p0_x + (line_row_bottom_y - p0_y) * x_slope;
+            // Get tile coordinates for start/end points, use i32 to preserve negative coordinates
+            let p0_tile_x = line_top_x.floor() as i32;
+            let p0_tile_y = line_top_y.floor() as i32;
+            let p1_tile_x = line_bottom_x.floor() as i32;
+
+            // TODO This may now be unnecessary? This should not produce a visual difference.
```
LaurenzV, line 361: I tried replacing the statement with `line_bottom_y.floor() as i32` and it doesn't cause any visual differences for any PDF, so should be fine?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -267,76 +324,304 @@ impl Tiles {
...
+                line_bottom_y.floor() as i32
+            };
 
-                    // The line's x-coordinates at the line's left- and right-most points within the
-                    // tile row.
-                    let line_row_left_x =
-                        f32::min(line_row_top_x, line_row_bottom_x).max(line_left_x);
-                    let line_row_right_x =
-                        f32::max(line_row_top_x, line_row_bottom_x).min(line_right_x);
+            // special-case out lines which are fully contained within a tile.
```
LaurenzV:
```suggestion
            // Special-case out lines which are fully contained within a tile.
```

b0nes164:
Awesome! Slightly awkward, but can you do one more benchmark pass with the const generic [here](https://github.com/linebender/vello/blob/5f222a7af0fd48ae046262eaa1a43cd4660854f8/sparse_strips/vello_common/src/strip_generator.rs#L144) set to true? I suspect the performance will be worse with it on :)

sparse_strips/vello_common/src/tile.rs
```diff
@@ -267,76 +324,304 @@ impl Tiles {
...
+                let x_slope = (p1_x - p0_x) / (p1_y - p0_y);
+                let line_row_top_x = p0_x + ((line_row_top_y) - p0_y) * x_slope;
+                let line_row_bottom_x = p0_x + (line_row_bottom_y - p0_y) * x_slope;
+                let x_idx = (f32::min(line_row_top_x, line_row_bottom_x).max(line_left_x) as u16)
+                    .min(tile_columns + 1);
+                let tile = Tile::new_clamped(
+                    x_idx,
+                    y_top_tiles,
+                    line_idx,
+                    ((y >= line_top_y) as u32) << 5,
```
LaurenzV, line 624: If the line only covers 1 tile, can it even happen that the top is intersected?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -267,76 +324,304 @@ impl Tiles {
...
+                            };
+
+                            let tile = Tile::new(x_end, y_idx, line_idx, intersection_mask);
+                            self.tile_buf.push(tile);
+                        }
+                    }
                 }
+            } else {
+                // Case: Line is fully contained within a single tile.
+                // TODO: Must exactly match the general case? Or can be simplified?
```
LaurenzV, line 611: Not sure I understand this TODO? Is there a way to simplify this even further?

LaurenzV:
> Awesome! Slightly awkward, but can you do one more benchmark pass with the const generic [here](https://github.com/linebender/vello/blob/5f222a7af0fd48ae046262eaa1a43cd4660854f8/sparse_strips/vello_common/src/strip_generator.rs#L144) set to true? I suspect the performance will be worse with it on :)

It depends on the test case, but it does look like it has an impact in some cases.

image
image
image

On another note, I tried rerunning the PDF test cases with the flag enabled and in some very few cases there a small number of pixel differences, do you know if this is intended? I would've assumed it shouldn't have any impact at all since vello_cpu doesn't make use of those masks.

b0nes164:
Thanks!

> there a small number of pixel differences

There *shouldn't* be any differences. The only thing I can think of is that there is a mistake which is causing the intersections to corrupt the winding, but looking through it, I don't see anything that seems suspicious?

sparse_strips/vello_common/src/tile.rs
b0nes164: Since the line is self contained to a tile, I think you can use the raw line endpoints to get the tile's x coordinate. Notice everything else besides `x_idx` does not need these calculations. However, when I tried this, I got artifacts. My thinking is that in isolation this would be fine, but it causes watertightness issues with the tiles calculated by the general "multi-tile" case.

b0nes164: It's probably possible to nudge the x in the right direction more cheaply than doing all the calculations, but I haven't dug into it yet.

LaurenzV:
Will try to find a reproducing example

sparse_strips/vello_common/src/tile.rs
b0nes164: This is calculating the winding. Note it is valid to have the winding set for a fully self contained tile:
https://github.com/linebender/vello/blob/5f222a7af0fd48ae046262eaa1a43cd4660854f8/sparse_strips/vello_common/src/tile.rs#L1662-L1673

LaurenzV:
Figured it out! The reason is that the flags make up the most significant bits while the line index makes the tail. Therefore, if we have multiple tiles at the same location, the specific sort order might be different (since we use `sort_unstable`), and this presumably can lead to floating-point rounding differences during strip generation. We could probably fix this by moving the line index up in the field? But I don't think this needs to happen in this PR.

b0nes164:
Nice! Does this mean that the correct ordering is dependent on the winding being the top bit? As in, the updated packed u32 should be: `winding | line_idx | intersection_mask` ?

I'll add this to a followup CL. I'll also plan to case out the logic even further, as the 10% performance penalty is quite heavy, and there's still room for improvement.

sparse_strips/vello_common/src/tile.rs
b0nes164: The Analytic AA shouldn't be affected with or without this line. However, it can effect the MSAA. I think it can be removed, but I want to keep this in until I have testing setup for the MSAA so that I can be certain that removing it doesn't regress anything.

LaurenzV:
I think the best thing would just be `line_idx | winding | intersection_mask`, so that the order is always unique based on the line index (but yours would probably work as well).

PR #1294, opened 2025-11-17
CPU: feature flag u8 and f32 pipelines

nicoburns:
Adds feature flags to `vello_cpu`s `u8` and `f32` pipelines so that you can save on binary size if you're only planning on using one of them. Disabling the `f32` pipeline knocks ~60kb off the binary size for me ~250kb -> ~190kb.

### DIFF
--- a/sparse_strips/vello_sparse_tests/Cargo.toml
+++ b/sparse_strips/vello_sparse_tests/Cargo.toml
@@ -17,7 +17,7 @@ path = "tests/mod.rs"
 [dependencies]
 vello_api = { workspace = true }
 vello_common = { workspace = true, features = ["std"] }
-vello_cpu = { workspace = true, features = ["multithreading", "std"] }
+vello_cpu = { workspace = true, features = ["multithreading", "std", "f32_pipeline"] }
 vello_hybrid = { workspace = true }
 wgpu = { workspace = true, default-features = true }
 pollster = { workspace = true }


LaurenzV:
I’m not sure locking enum variants behind a feature flag is a good idea because it might break existing matching code downstream if someone activates one of the features.

nicoburns:
> I’m not sure locking enum variants behind a feature flag is a good idea because it might break existing matching code downstream if someone activates one of the features.

Yeah, good point. We could mark the enum as non-exhaustive. But perhaps it's better to just keep both variants but ignore the value passed if only one pipeline is enabled? The enum values are "phrased" to be hint-like anyway...

LaurenzV:
Yes I think ignoring is better.

nicoburns:
@DJMcNab Any ideas on how to setup CI for this:

- Two new features `u8_pipeline` and `f32_pipeline`.
- Only exist for `vello_cpu` crate
- Should ideally be a compile error if neither of them is specified

Perhaps we need to exclude `vello_cpu` from the workspace cargo-hack and test it separately?

DJMcNab:
I *think* it should be able to be treated in the same way that we treat `std` and `libm`.

nicoburns:
I have fixed CI by enabling both pipelines everywhere we're not explicitly checking that it compiles with just one of them enabled, and I've added a `compile_error!` for the case where neither pipeline is enabled.

LaurenzV:
It seems like there are still failures

nicoburns:
> It seems like there are still failures

Ah, I think the `compile_error!` when neither feature is enabled reintroduced CI failures. Fixed.

LaurenzV:
I'm wondering whether it make sense to only enable the u8 pipeline by default. IMO, that's pretty much always the one you should use. But not sure.

nicoburns:
> I'm wondering whether it make sense to only enable the u8 pipeline by default. IMO, that's pretty much always the one you should use. But not sure.

I would agree with this. I've made the `f32_pipeline` feature non-default.

.github/workflows/ci.yml
```diff
@@ -146,13 +146,19 @@ jobs:
...
 
       - name: cargo clippy
-        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features std -- -D warnings
+        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features std,u8_pipeline,f32_pipeline -- -D warnings
+
+      - name: cargo clippy (only u8)
+        run: cargo clippy -p vello_cpu --locked --no-default-features --features std,text,png,u8_pipeline
+
+      - name: cargo clippy (only f32)
+        run: cargo clippy -p vello_cpu --locked --no-default-features --features std,text,png,f32_pipeline
```
LaurenzV: Maybe the name should reflect that it's only for vello_cpu?

LaurenzV (approved):
A bit annoying to make this work with CI, but it is what it is. 😄

nicoburns:
> A bit annoying to make this work with CI, but it is what it is. 😄

There is an `--at-least-one-of` flag to cargo hack that looks like it would make it nicer. But that requires changing to `--feature-powerset` rather than using `--each-feature` which seems like it would be a much bigger change.

PR #1295, opened 2025-11-17
Update authors to include Andrew, Alex, and Taj

taj-p:
Per title. @DJMcNab mentioned awhile ago that we should add ourselves 😅 - my bad on the delay!!


ajakubowicz-canva (approved):

DJMcNab (approved):
Given that everyone added has approved, this looks good to me.

PR #1298, opened 2025-11-21
feat: support ohos

richerfu:
Render successfully. :)

PureX 6.0.0.112

e1436e1c1459d7cc4f640d4a5d0b8082

### DIFF
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -46,7 +46,7 @@ use winit::window::{Window, WindowAttributes};
 
 use vello::wgpu::{self, PipelineCache};
 
-#[cfg(not(any(target_arch = "wasm32", target_os = "android")))]
+#[cfg(not(any(target_arch = "wasm32", target_os = "android", target_env = "ohos")))]
 mod hot_reload;
 mod minimal_pipeline_cache;
 mod multi_touch;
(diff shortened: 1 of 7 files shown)


examples/with_winit/src/lib.rs
```diff
@@ -106,12 +106,12 @@ struct RenderState {
     window: Arc<Window>,
 }
 
-#[cfg(not(target_os = "android"))]
+#[cfg(not(any(target_os = "android", target_env = "ohos")))]
```
waywardmonkeys, line 109: Is this actually necessary here? IIRC, this stuff was a workaround for the multiple shaders being slow to compile ... I'd rather just not have this in here perhaps, but if we don't need to replicate this for ohos, that'd be good.

waywardmonkeys (commented):
I know this is a work in progress / draft.

Overall, this makes me feel like we should clean up and improve our example code.

Is there a timeline on getting the support in winit landed? Will there be a 0.30 release with support or will it end up going into 0.31?

examples/with_winit/src/lib.rs
richerfu: I'm not sure it. But i think it's not bad to keep same with android now. :) May we can remove it later.

richerfu:
> I know this is a work in progress / draft.
> 
> Overall, this makes me feel like we should clean up and improve our example code.
> 
> Is there a timeline on getting the support in winit landed? Will there be a 0.30 release with support or will it end up going into 0.31?

I hope 0.31 can support ohos. I'm trying to do it. :)

PR #1299, opened 2025-11-21
Implement an initial version of Vello API

DJMcNab:
This is an initial implementation of Vello API, our abstraction between Vello CPU and Vello Hybrid (and also maybe Vello Classic). 

A few notes:
 
- This is designed to be combined with something like a "Parley Draw", i.e. it's designed to allow external glyph rendering.
- The design is fully `no_std` + `alloc` compatible.
- Vello API does not try to abstract away the unabstractable, e.g. handling external textures.
- The multi-threading design is something which *will* need more thought. I think we're likely to end up with an explicit CommandEncoder like API, similarly to that seen in wgpu.

Importantly, this PR is intentionally *not* making internal changes to the extant crates, and instead is only adding new API surface for the abstraction. The plan would be to make these changes only once this has landed, to limit merge conflicts as much as possible.

I think it's valuable to get something landed here, whether or not it ends up being the final state. The main advantage there is that it allows porting tests in a piecemeal fashion, as well as making changes to the APIs of Vellos CPU and Hybrid (without creating a massive merge conflict headache). This PR is currently scoped small enough to be easily revertible if we want to go a different direction.

A previous version of this PR used the renderer abstraction for the tests. That code is now gone as it wasn't landable, but can be seen in https://github.com/linebender/vello/commit/b200d5e239d3632273ca092986667ae923782adf and earlier.

### DIFF
--- a/sparse_strips/vello_api/src/painter.rs
+++ b/sparse_strips/vello_api/src/painter.rs
@@ -20,7 +20,7 @@ use crate::scene::{OurBrush, Scene};
 /// - The brush, which describes what will be drawn by the following commands.
 ///   This can be a solid colour ([`set_solid_brush`]), a gradient, an image (both [`set_brush`]), or a blurred rounded rectangle ([`set_blurred_rounded_rect_brush`]).
 ///   The brush's coordinate system is currently relative to the drawn path's coordinate system.
-///   <!-- We might also want to allow this to be explicitly unset, in which case drawing will fail -->.
+///   <!-- We might also want to allow this to be explicitly unset, in which case drawing will fail. -->
 /// - The area over which this brush is to be drawn.
 ///   This can either be a filled path ([`fill_path`])
 /// - The layer stack, which allows for multiple 2d contexts to be blended together, with clipping.
--- a/sparse_strips/vello_api/src/paths.rs
+++ b/sparse_strips/vello_api/src/paths.rs
@@ -16,8 +16,8 @@
 
 use alloc::vec::Vec;
 use peniko::{
-    Fill, Style,
-    kurbo::{PathEl, Shape, Stroke},
+    Style,
+    kurbo::{PathEl, Shape},
 };
 
 /// The id for a single path within a given [`PathSet`].
@@ -55,6 +55,9 @@ pub struct PathSet {
 }
 
 impl PathSet {
+    /// Create a new, empty path collection.
+    ///
+    /// This method doesn't allocate.
     pub fn new() -> Self {
         Self {
             elements: Vec::new(),
@@ -62,6 +65,9 @@ impl PathSet {
         }
     }
 
+    /// Clears the path set, removing all values.
+    ///
+    /// This does not free the underlying allocations.
     pub fn clear(&mut self) {
         self.elements.clear();
         self.meta.clear();
@@ -75,7 +81,13 @@ impl Default for PathSet {
 }
 
 impl PathSet {
-    pub fn prepare_fill(&mut self, fill_rule: Fill, shape: &impl Shape) -> PathId {
+    /// Prepare an outline for drawing as a shape with the given style.
+    ///
+    /// This returns the id of this path in this `PathSet`.
+    /// See the docs on [`append`](PathSet::append) for how this changes when path sets are combined.
+    ///
+    /// This method is generally only expected to be used by [`Scene`](crate::Scene).
+    pub fn prepare_shape(&mut self, shape: &impl Shape, style: impl Into<Style>) -> PathId {
         let start_index = self.elements.len();
         // TODO: We hard-code this tolerance to be 0.1, as every other call does so.
         // We should maybe change that at some point?
@@ -85,28 +97,19 @@ impl PathSet {
         let meta_index = self.meta.len();
         self.meta.push(PathMeta {
             start_index,
-            operation: Style::Fill(fill_rule),
+            operation: style.into(),
         });
 
         // TODO: Better error handling here?
         PathId(meta_index.try_into().unwrap())
     }
 
-    // TODO: Maybe require stroke expansion happen *prior* to this stage?
-    pub fn prepare_stroke(&mut self, stroke_rule: Stroke, shape: &impl Shape) -> PathId {
-        let start_index = self.elements.len();
-        // TODO: Perform dash expansion now, even if not stroke expansion?
-        // See https://xi.zulipchat.com/#narrow/channel/260979-kurbo/topic/Removing.20dash_pattern.20from.20Stroke/with/561141820
-        self.elements.extend(shape.path_elements(0.1));
-        let meta_index = self.meta.len();
-        self.meta.push(PathMeta {
-            start_index,
-            operation: Style::Stroke(stroke_rule),
-        });
-
-        PathId(meta_index.try_into().unwrap())
-    }
-
+    /// Append the shapes in `other` to this pathset.
+    ///
+    /// The return value should be added to the field of [`PathId`]s from `other`
+    /// for use in the combined pathset (i.e. the new value of `self`).
+    ///
+    /// This method is expected to be used to implement [`PaintScene::append`](crate::PaintScene::append).
     #[must_use]
     pub fn append(&mut self, other: &Self) -> u32 {
         let external_correction_factor = self.meta.len().try_into().unwrap();
@@ -121,12 +124,14 @@ impl PathSet {
     }
 }
 
+/// Metadata about a single path in a [`PathSet`].
 #[derive(Debug, Clone)]
 pub struct PathMeta {
     // Would u32 work here?
     /// The index in [`PathSet::elements`] from which this path's elements starts.
     ///
-    /// The path ends at the start of the next `PathMeta`, and so.
+    /// The path ends at the start of the next `PathMeta`, and so for the final path
+    /// the elements are the remaining elements.
     pub start_index: usize,
     /// How the path will be rendered.
     ///
@@ -137,7 +142,7 @@ pub struct PathMeta {
     ///
     /// There are arguments for splitting again, into "paths" and "styled paths" or similar, but
     /// that piles on complexity; and realistically how many people will use that?
-    /// Alternatively, if we made PathMeta store a range instead of a single index,
+    /// Alternatively, if we made `PathMeta` store a range instead of a single index,
     /// that makes reusing segments much easier.
     ///
     /// As stroke expansion is going to be happening on the CPU anyway,
--- a/sparse_strips/vello_api/src/scene.rs
+++ b/sparse_strips/vello_api/src/scene.rs
@@ -70,7 +70,7 @@ pub struct BlurredRoundedRectBrush {
 
 /// A reusable sequence of drawing commands for renders to a specific [`Renderer`].
 ///
-///  # Hinting
+/// # Hinting
 ///
 /// TODO: Describe how a scene can be "hinted", i.e. the drawing operations know they will only be translated, and therefore can
 /// design to specific pixels.
@@ -227,7 +227,7 @@ impl PaintScene for Scene {
         fill_rule: peniko::Fill,
         path: impl peniko::kurbo::Shape,
     ) {
-        let idx = self.paths.prepare_fill(fill_rule, &path);
+        let idx = self.paths.prepare_shape(&path, fill_rule);
         self.commands.push(RenderCommand::DrawPath(transform, idx));
     }
     fn stroke_path(
@@ -236,7 +236,7 @@ impl PaintScene for Scene {
         stroke_params: &peniko::kurbo::Stroke,
         path: impl peniko::kurbo::Shape,
     ) {
-        let idx = self.paths.prepare_stroke(stroke_params.clone(), &path);
+        let idx = self.paths.prepare_shape(&path, stroke_params.clone());
         self.commands.push(RenderCommand::DrawPath(transform, idx));
     }
 
@@ -294,10 +294,10 @@ impl PaintScene for Scene {
         // mask: Option<Mask>,
     ) {
         let clip_idx = if let Some(clip_path) = clip_path {
-            Some(self.paths.prepare_fill(
-                // TODO?
-                peniko::Fill::NonZero,
+            Some(self.paths.prepare_shape(
                 &clip_path,
+                // TODO: Make this configurable for clip paths.
+                peniko::Fill::NonZero,
             ))
         } else {
             None


sparse_strips/vello_api/src/texture.rs
```diff
@@ -0,0 +1,55 @@
...
+
+//! Texture types used in Vello API.
+
+use bitflags::bitflags;
+use core::fmt::Debug;
+
+#[derive(Copy, Clone, Debug)]
+pub struct TextureDescriptor {
+    // TODO: Maybe a better type is `atomicow::CowArc<'static, str>`
+    pub label: Option<&'static str>,
```
nicoburns, line 49: https://lib.rs/crates/smol_str is another option here (Winit depends on it)

PoignardAzur: I'm wondering how much effort it would be to coordinate a UI-ecosystem-wide change to one of the other mini-string crates. IIRC some of them have more features.

Maybe at next RustWeek.

sparse_strips/vello_api/README.md
```diff
@@ -16,19 +16,77 @@
...
+- presenting rendered content to an operating system window.
+
+These functionalities are however catered for where applicable by APIs on the specific renderers.
+The renderer API supports downcasting to the specific renderer, so that these extensions can be called.
+Each supported renderer will/does have examples showing how to achieve this yourself.
+
+## Text
+
+Vello API does not handle text/glyph rendering itself.
+This allows for improved resource sharing of intermediate text layout data, for hinting and ink splitting underlines.
```
taj-p:
```suggestion
This allows for improved resource sharing of intermediate text layout data, for hinting and ink skipping underlines.
```

sparse_strips/vello_api/src/lib.rs
```diff
@@ -1,9 +1,85 @@
...
+//!
+//! This abstraction is tailored for the Vello renderers, as we believe that these have a sufficiently broad coverage of the trade-off
+//! space to be viable for any consumer.
+//! Vello API guarantees identical rendering between these renderers, barring subpixel differences due to precision/different rounding.
+//! <!-- TODO: Is ^ true? -->
+//!
+//! # Abstraction Boundaries
+//!
+//! The abstractions in this crate are focused on 2D rendering, and the resources required to perform that.
+//! In particular, this does abstract over strategies for:
```
dominikh: `s/this does/this does not/`?

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,126 @@
...
+    // This can be accessed through downcasting:
+    // (Well, kind of...; that doesn't apply to Scene, which is by far the most common type here.)
+    // fn set_aliasing_threshold(&mut self, aliasing_threshold: Option<u8>);
+    // TODO: What does this mean?
+    // fn set_blend_mode(&mut self, blend_mode: BlendMode);
+
+    // TODO: Do we want this exposed?
+    // TODO: &impl Shape?
+    // fn push_clip_path(&mut self, path: &kurbo::BezPath);
+    // fn pop_clip_path(&mut self);
```
PoignardAzur: We need to remove these TODO comments and commented-out methods, and either discard the features or have an explicit roadmap for them.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    // TODO: We need to support a fill rule here for clip paths...
```
PoignardAzur: Same here.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn push_layer(
+        &mut self,
+        clip_transform: Affine,
+        clip_path: Option<impl Shape>,
+        blend_mode: Option<BlendMode>,
+        opacity: Option<f32>,
+        // mask: Option<Mask>,
+        // filter: Option<Filter>
```
PoignardAzur: Same.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,126 @@
...
+    );
+
+    // TODO: Do we need all of these?
+    // TODO: We need to support a fill rule here...
```
PoignardAzur: Same.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn push_clip_layer(&mut self, clip_transform: Affine, path: impl Shape) {
+        self.push_layer(clip_transform, Some(path), None, None);
+    }
+
+    fn push_blend_layer(&mut self, blend_mode: BlendMode) {
+        self.push_layer(
+            Affine::IDENTITY,
+            // This needing to be turbofished shows a real footgun with the proped "push_layer" API here.
+            None::<BezPath>,
```
PoignardAzur: I don't think we'll solve this in the short-term, but this is definitely worth creating an issue about.

sparse_strips/vello_api/src/paths.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PathSet {
+    pub fn prepare_fill(&mut self, fill_rule: Fill, shape: &impl Shape) -> PathId {
+        let start_index = self.elements.len();
+        // TODO: Maybe determine the optimal tolerance from the transform.
+        // If this can be post-transformed, that doesn't actually meaningfully apply?
```
PoignardAzur: We should either create an issue or remove this comment.

sparse_strips/vello_api/src/paths.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        // TODO: Better error handling here?
+        PathId(meta_index.try_into().unwrap())
+    }
+
+    // TODO: Maybe require stroke expansion happen *prior* to this stage?
+    pub fn prepare_stroke(&mut self, stroke_rule: Stroke, shape: &impl Shape) -> PathId {
+        let start_index = self.elements.len();
+        // TODO: Determine the optimal tolerance from the transform.
+        // TODO: Perform dash expansion now, even if not stroke expansion?
```
PoignardAzur: We should create an issue, or at least have a discussion of the trade-offs involved somewhere.

sparse_strips/vello_api/src/paths.rs
```diff
@@ -0,0 +1,130 @@
...
+//! for each path).
+
+use alloc::vec::Vec;
+use peniko::{
+    Fill,
+    kurbo::{PathEl, Shape, Stroke},
+};
+
+#[derive(Debug, Clone, Copy)]
+/// The id for a single path within a given [`PathSet`].
```
PoignardAzur: We usually put the derives after the doc comments.

sparse_strips/vello_api/src/paths.rs
```diff
@@ -0,0 +1,130 @@
...
+            it.start_index += internal_correction_factor;
+            it
+        }));
+
+        external_correction_factor
+    }
+}
+
+#[derive(Debug, Clone)]
+/// How a path will be rendered.
```
PoignardAzur: Same here.

sparse_strips/vello_api/src/sync.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// See the [module level documentation](crate::sync) for more information.
+    ///
+    /// When the standard library is enabled, this is an alias for `Send + Sync`.
+    /// If not (or on Wasm), this is implemented by every type.
+    pub trait Share: Send + Sync {}
+
+    impl<T: Send + Sync + ?Sized> Share for T {}
+
+    #[derive(Debug)]
+    /// The most practical locking primitive for interior mutability in 2D renderers.
```
PoignardAzur: Same here.

sparse_strips/vello_api/src/renderer.rs
```diff
@@ -0,0 +1,121 @@
...
+    sync::Share,
+    texture::{TextureDescriptor, TextureHandle, TextureId},
+};
+
+// TODO: Maybe?
+// pub enum MaskOperation {
+//     SvgLuminance,
+//     CssLuminance,
+//     Alpha,
+// }
```
PoignardAzur: Remove commented-out code or add an issue.

sparse_strips/vello_api/src/renderer.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Create a texture for use in renders with this device.
+    fn alloc_untracked_texture(&self, descriptor: TextureDescriptor) -> TextureId;
+    // Error if the texture was already freed/not associated with this renderer.
+    fn free_untracked_texture(&self, texture: TextureId) -> Result<(), ()>;
+    // TODO: Texture resizing? Reasonable reasons to not do that include cannot resize wgpu textures
+    // Also what does that mean for existing content, etc.
+
+    // TODO: Better error kinds.
+    // TODO: Also accept TextureHandle?
```
PoignardAzur: Remove TODO comments or create issues.

sparse_strips/vello_api/src/renderer.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn upload_image(&self, to: &TextureId, data: &peniko::ImageData) -> Result<(), ()>;
+
+    fn as_arc(&self) -> Arc<Self>
+    where
+        Self: Sized;
+    fn as_dyn_arc(&self) -> Arc<dyn Renderer>;
+
+    // TODO: Reason about how we want downloads to work.
+    // fn queue_download(&self, of: &TextureId) -> DownloadId;
```
PoignardAzur: I think we've agreed that downloads are outside the scope of this MVP, so we should just get rid of this for now.

sparse_strips/vello_api/src/renderer.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    // fn create_mask(descriptor: MaskOperation) -> Mask;
+    // fn mask_from_scene(from: &Texture, to: &Scene, MaskDescriptor { subset_rect });
```
PoignardAzur: Either remove this or create an issue.

sparse_strips/vello_api/src/renderer.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn create_scene(&self, to: &TextureId, options: SceneOptions) -> Result<Self::ScenePainter, ()>
+    where
+        Self: Sized;
+    fn queue_render(&self, from: Self::ScenePainter)
+    where
+        Self: Sized;
+}
+
+// TODO: impl dyn Renderer {fn downcast}
+// TODO: impl Arc<dyn Renderer> {fn downcast}
```
PoignardAzur: Remove this.

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// Clip path.
+    pub clip_path: Option<PathId>,
+    /// Blend mode.
+    pub blend_mode: Option<BlendMode>,
+    /// Opacity.
+    pub opacity: Option<f32>,
+    // /// Mask.
+    // pub mask: Option<crate::renderer::Mask>,
+    // /// Filter.
+    // pub filter: Option<Filter>,
```
PoignardAzur: Needs to be removed.

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,297 @@
...
+        self.commands.clear();
+        self.paths.clear();
+    }
+    pub fn hinted(&self) -> bool {
+        self.hinted
+    }
+}
+
+// TODO: Change module and give a better name.
+pub type OurBrush = peniko::Brush<peniko::ImageBrush<TextureHandle>>;
```
PoignardAzur, line 121: I'd suggest "VelloBrush" for now.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,126 @@
...
+    // TODO: We need to support a fill rule here for clip paths...
+    fn push_layer(
+        &mut self,
+        clip_transform: Affine,
+        clip_path: Option<impl Shape>,
+        blend_mode: Option<BlendMode>,
+        opacity: Option<f32>,
+        // mask: Option<Mask>,
+        // filter: Option<Filter>
+    );
```
PoignardAzur, line 111: Is there any reason why the API doesn't let you add transform layers?

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,297 @@
...
+                RenderCommand::PopLayer => RenderCommand::PopLayer,
+                RenderCommand::SetPaint(affine, brush) => {
+                    // Don't update the paint_transform, as it's already path local.
+                    RenderCommand::SetPaint(*affine, brush.clone())
+                }
+                RenderCommand::BlurredRoundedRectPaint(brush) => {
+                    // Don't update the paint_transform, as it's (currently) already path local.
+                    RenderCommand::BlurredRoundedRectPaint(brush.clone())
+                }
+            }));
```
PoignardAzur, line 209: For example, seems like this block of code could be replaced with a transform layer.

sparse_strips/vello_api/src/texture.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub width: u16,
+    pub height: u16,
+    pub usages: TextureUsages,
+    // TODO: Format? Premultiplication? Hdr?
+    // TODO: Explicit atlasing?
```
PoignardAzur: If you're not sure what fields you'll add, this struct should be `#[non_exhaustive]`.

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,297 @@
...
+    // which also don't allow bringing in textures?
+    pub renderer: Arc<dyn Renderer>,
+    pub hinted: bool,
+    /// Handles for each texture stored in this scene.
+    ///
+    /// We store these to ensure that the textures won't be freed until this scene
+    /// is fully rendered (and dropped).
+    // TODO: HashSet? We'd need to bring in hashbrown, but that's probably fine
+    pub textures: Vec<TextureHandle>,
+}
```
PoignardAzur: The Scene type needs way more documentation. As it is, I don't know how it's supposed to be used, how it fields relate to each other, what assumptions it makes that mustn't be broken, etc.

It's especially not clear to me how `Scene` and `PathSet` relate. Does each scene store its own set of paths? Doesn't that kind of defeat the point of path caching?

PoignardAzur (commented):
Overall, not sure how to think about this PR. It has a lot of TODO comments and hedges in places where we'll need to make judgment calls sooner or later.

sparse_strips/vello_api/src/painter.rs
DJMcNab: It isn't an existing feature in any of the Vellos {Classic, Hybrid, CPU}.

It also doesn't provide new functionality over the user manually keeping track of a transform externally, but it makes reasoning about a number of things more difficult (primarily drawing hinted glyphs, but also in the future how filter effects and masks will be applied).

Implementation wise, it's also a bit confused because it doesn't behave in the same way as any other layer type; i.e. it would be a layer type which doesn't implement a new isolated context, unless you set any of the other fields.

sparse_strips/vello_api/src/scene.rs
DJMcNab: I don't think that's true; this code is equivalent to the implementation of [`Scene::append`](https://docs.rs/vello/latest/vello/struct.Scene.html#method.append) in Vello Classic. That code would still need to exist in a version of this proposal with transform layers; the path references still need to be updated.

The only part of this code which would change is the two calls to `correct_transform`; instead those would be handled by the transform layer.

sparse_strips/vello_api/src/scene.rs
DJMcNab: > It's especially not clear to me how `Scene` and `PathSet` relate. Does each scene store its own set of paths? Doesn't that kind of defeat the point of path caching?

Sure, this is definitely under-documented. I'm planning to spend the time to add some more docs as my near-term priority here.

Essentially, each `Scene` accumulates the paths which it's drawing. These can either be copied into any painters which the scene is used in, to be rasterised in a just-in-time manner, or moved into the renderer (i.e. cached).
This accumulation pattern is designed to avoid an allocation for each path, whilst still allowing reference-count based garbage collection for cached paths (they are reference counted as a group of "related" paths).

But whilst `PathSet` currently exists for that design, the caching is not yet implemented (as it doesn't actually exist in the underlying renderers, for one). But I feel it still makes sense as an abstraction boundary to have, albeit with better docs...

sparse_strips/vello_api/src/paths.rs
DJMcNab: I'll link to [#kurbo > Removing dash_pattern from Stroke](https://xi.zulipchat.com/#narrow/channel/260979-kurbo/topic/Removing.20dash_pattern.20from.20Stroke/with/561141820)

sparse_strips/vello_api/src/texture.rs
DJMcNab: Can you clarify how you expect this to work, if users need to create this struct?

This type mirrors the descriptors from wgpu, which aren't `non_exhaustive`.
Hopefully we get default_field_values soon enough, ideally with a nice `non_exhaustive` story.

sparse_strips/vello_api/src/paths.rs
DJMcNab: I've linked to [#vello > Determining correct &#96;Shape&#96; tolerance](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Determining.20correct.20.60Shape.60.20tolerance/with/565793318). I don't feel that an issue would be the right way to discuss this, but if you want there to be one, feel free to create it.

grebmeg:
Great job, @DJMcNab. This looks like a strong foundational implementation, and designing a unified API is genuinely hard work. I haven’t dived deeply into the details yet, but I did a brief skim.

My initial impression, which might be off, is that this PR introduces a lot of new abstractions and concepts at once. Because of that density, I worry that once this lands it may effectively become the finished API, with some parts receiving less thorough review than they deserve. At the same time, many of these new concepts could naturally spark discussion across different consumers. 

Not sure if this is feasible, but would it be possible to extract some of the core logic into multiple, smaller PRs and introduce things iteratively? That could make reviews simpler, reduce cognitive load, and ultimately improve review quality. The idea would be to break this into multiple small, logically scoped PRs, starting with traits and then layering in initial methods and implementations. I know this would be more time consuming, but in some cases it may be worth it.

Keen to hear your thoughts.

DJMcNab:
> Not sure if this is feasible, but would it be possible to extract some of the core logic into multiple, smaller PRs and introduce things iteratively? That could make reviews simpler, reduce cognitive load, and ultimately improve review quality. The idea would be to break this into multiple small, logically scoped PRs, starting with traits and then layering in initial methods and implementations. I know this would be more time consuming, but in some cases it may be worth it.

I think this is viable. The most obvious split is:
1) Start with `PaintScene`, `Scene` and `PathSet`. This is #1360
2) Follow-up with Renderer.

I don't think it's easy to have a much smaller scope. This hadn't seemed viable before designing the API, but it does now seem more so.

PR #1301, opened 2025-11-22
Optimize gradient rendering on CPU

### LINKED ISSUE #1300: `vello_cpu`'s `GradientPainter` is slower on AVX2 with the latest unreleased `fearless_simd`
I've been benchmarking vello_cpu on the unreleased latest version of fearless_simd, and the `fine/gradient/linear/opaque_f32_avx2` benchmark seems to be ~twice as slow. The relevant code is [`GradientPainter::next()` and `GradientPainter::paint_f32()`](https://github.com/linebender/vello/blob/9c38a4a8fbe4926743ce3e6cf62ecf329c47b35a/sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs#L74-L175).

There are two problems here:

- `paint_f32` does not `Simd::vectorize` its body. fearless_simd 0.3 just so happens to work, because none of the f32x8 operations used in `GradientPainter::next()` or `GradientPainter::paint_f32()` actually call any architecture-specific intrinsics. However, now that AVX2 is implemented, it *does* use intrinsics for f32x8 operations. This results in a massive slowdown since these would previously fall back to `core::arch` intrinsic function calls. With that problem fixed, though...

- `paint_f32` is actually *pessimized* by SIMD, because all it's doing is shuffling a bunch of data around. Particularly expensive are the `gather!` operations in `GradientPainter::next()`, where we interleave a bunch of RGBA pixels before deinterleaving them immediately afterwards to store them. My guess is that LLVM recognizes this pattern in the non-SIMD version and can eliminate those operations, but loses track of this in the SIMD version and falls back to a bunch of lane inserts.

I debated whether to make this a fearless_simd issue or a vello_cpu issue, but I think it's a problem with vello_cpu. We shouldn't be interleaving data only to deinterleave it immediately afterwards.

valadaptive:
Resolves #1300. This gets rid of all the interleave/deinterleave ops, and just performs a bunch of straight array copies.

There's some nuance around handling NaNs. Previously, there were two select operations done: one at the beginning, to set the LUT indices to be in bounds if the corresponding *t* value was NaN, and one for each channel to make the color value completely transparent if the corresponding non-extended *t* value was NaN.

The second select operation is a bit awkward since the channel values are now packed together, so I've added an extra transparent color value to the end of each gradient LUT. If a *t* value is NaN, we set the LUT index to that value instead. This only requires one select. It's a bit ugly that we do this for all gradients in `vello_common` even though it's only necessary for radial gradients on the CPU path; I'm very open to ideas on how to make it better.

(EDIT: I've gone ahead and applied the same trick to the GPU path as well.)

I've also wrapped all the gradient paint operations in `vectorize` calls, since they were missing before and not actually being vectorized as a result.

On my machine (Ryzen 7 7700x), this takes `fine/gradient/linear/opaque_u8_avx2` from 830ns/iter to 620ns/iter, and `fine/gradient/linear/opaque_f32_avx2` from 850ns/iter to 660ns/iter. `fine/gradient/linear/opaque_f32_scalar` regresses from 890ns/iter to 1.1µs/iter, and I'm not sure why.

The `fine/gradient/linear/opaque_u8` benches appear to use the `lowp` module's gradient painter, but the *tests* use the `common` module's gradient painter. I don't know how to benchmark the latter, so the `fine/gradient/linear/opaque_u8_avx2` gain seems to be solely from adding the `vectorize` wrapper.

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -1037,7 +1037,7 @@ impl<T: FromF32Color> GradientLut<T> {
                 // Make sure not to overwrite any extra transparent color at the end (it's not
                 // counted in `lut_size`)
                 let lut = &mut lut[idx..(idx + 4).min(lut_size)];
-                lut.copy_from_slice(&rs[..lut.len().min(4)]);
+                lut.copy_from_slice(&rs[..lut.len()]);
             });
         }
 


sparse_strips/vello_common/src/encode.rs
```diff
@@ -1019,19 +1032,27 @@ impl<T: FromF32Color> GradientLut<T> {
...
                 let (r1, r2) = simd.split_f32x8(im1);
                 let (r3, r4) = simd.split_f32x8(im2);
+                let rs = [r1, r2, r3, r4].map(T::from_f32);
 
-                let lut = &mut lut[idx..][..4];
-                lut[0] = T::from_f32(r1);
-                lut[1] = T::from_f32(r2);
-                lut[2] = T::from_f32(r3);
-                lut[3] = T::from_f32(r4);
+                let lut = &mut lut[idx..];
```
DJMcNab: Can we redefine the `lut` we are using here (i.e. the one on the rhs of this operation) to be `&mut lut[..lut_size]`?

I believe that could avoid the need for the "probably" block, I think. A comment saying "probably" is all kinds of confusing!

valadaptive: Yes! By changing it to `let lut = &mut lut[idx..lut_size];` I can get rid of that extra operation at the end.

`&mut lut[idx..lut_size]` still allows overwriting elements past the end of the current ramp range, which is fine since the next ramp range will write the proper values. I think that's what we want, since we can separate the paths into "copy four values" and "copy the remainder', and the former branch will be taken for every loop iteration but the last. That should be a bit faster than doing the remainder checks every time.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -1019,19 +1032,24 @@ impl<T: FromF32Color> GradientLut<T> {
...
-                lut[2] = T::from_f32(r3);
-                lut[3] = T::from_f32(r4);
+                // Make sure not to overwrite any extra transparent color at the end (it's not
+                // counted in `lut_size`)
+                let lut = &mut lut[idx..lut_size];
+                if lut.len() >= 4 {
+                    lut[..4].copy_from_slice(&rs);
+                } else {
+                    lut.copy_from_slice(&rs[..lut.len()]);
+                }
```
LaurenzV: Couldn't we make this branchless and do `lut.len().min(4)`?

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
```diff
@@ -72,105 +72,83 @@ impl<'a, S: Simd> GradientPainter<'a, S> {
...
 
 impl<S: Simd> Iterator for GradientPainter<'_, S> {
-    type Item = ShaderResultF32<S>;
+    type Item = u32x8<S>;
 
     #[inline(always)]
     fn next(&mut self) -> Option<Self::Item> {
         let extend = self.gradient.extend;
-        let pos = f32x8::from_slice(self.simd, self.t_vals.next()?);
+        let pos = f32x8::from_slice(self.simd, self.t_vals.next().unwrap());
```
LaurenzV: Can we still keep the ? here?

LaurenzV:
I tested this on ARM, and it seems to result in a slight slowdown, but nothing concerning: 
image

With that said, I was well aware that the interleaved loading/storing was only really fast on ARM but not necessarily on x86 (and perhaps even WASM). So, if it results in good speedups on x86, I'm definitely fine with making that change.

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
```diff
@@ -72,105 +72,83 @@ impl<'a, S: Simd> GradientPainter<'a, S> {
...
+            self.simd.select_u32x8(
+                pos.simd_eq(pos),
+                indices,
+                u32x8::splat(
+                    self.simd,
+                    self.lut.transparent_index().unwrap_or_default() as u32,
+                ),
+            )
+        } else {
+            indices
```
LaurenzV: I think we can just do if let Some(transparent_index) {} else {}..., right? Because the transparent index will only be set if it was undefined in the first place.

sparse_strips/vello_common/src/encode.rs
valadaptive: I changed the code to:
```rust
                let lut = &mut lut[idx..(idx + 4).min(lut_size)];
                lut.copy_from_slice(&rs[..lut.len()]);
```
which I think is what you meant. We can't do:
```rust
                let lut = &mut lut[idx..];
                lut.copy_from_slice(&rs[..lut.len().min(4)]);
```
because the source and destination slice lengths need to be the same.

It's not really "branchless" in the performance sense, but it does move the branches from *our* code into the `copy_from_slice` code, which will be inlined.

The reason to keep the if-else is that `lut[..4].copy_from_slice(&rs);` will probably be compiled to four unrolled writes, whereas `lut.copy_from_slice(&rs[..lut.len()]);` requires a branch before each write.

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
valadaptive: Added it back.

sparse_strips/vello_cpu/src/fine/common/gradient/mod.rs
valadaptive: I did that and removed `has_undefined` from `GradientPainter`. There was some trickiness here--I thought we could de-duplicate `gradient_painter` and `gradient_painter_with_undefined`, but `U8Kernel` overrides the former method and not the latter, and it *doesn't* handle undefined gradient values.

valadaptive:
I should also mention I'm quite surprised that the scalar and NEON versions are slower. We should be performing strictly fewer operations, especially since we only need one select now. Maybe LLVM's instruction scheduling isn't doing well here?

sparse_strips/vello_common/src/encode.rs
LaurenzV: Ah okay, I'll leave it to you then how to do it.

PR #1303, opened 2025-11-24
fix: Filter expansion logic for transforms with scale/skew and for clipped layers

grebmeg:
The previous expansion calculation extracted only x/y scales from the transform and applied them separately to each direction. This failed for rotated filters (e.g., drop shadows, blurs) where the axis-aligned bounding box needs to account for the full transform matrix including rotation and shear.

Additionally, clipped filter layers were being processed incorrectly: the filter only saw the clipped region instead of the full layer. Filters must process the entire layer first, then apply the clip, otherwise effects like blurs get cut off at clip boundaries.

Before|After
---|---
image|image

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -740,9 +740,10 @@ impl<const MODE: u8> Wide<MODE> {
             }
         }
 
+        let in_clipped_filter_layer = layer.filter.is_some() && layer.clip;
         // Decrement the depth counter after popping a filtered layer with clip
-        if layer.filter.is_some() && layer.clip {
-            self.clipped_filter_layer_depth = self.clipped_filter_layer_depth.saturating_sub(1);
+        if in_clipped_filter_layer {
+            self.clipped_filter_layer_depth -= 1;
         }
     }
 


LaurenzV:
Maybe we could add a new test for this?

grebmeg:
Sure, I’ve added another one. Here’s the diff to show exactly what was wrong with the previous code. The left side shows the current version and the right side shows the previous one. Since we were rendering only the clipped geometry, the blur was applied only to that area, causing lighter regions around the geometry edge and creating noticeable artifacts along the vertical clip edges.

image

sparse_strips/vello_common/Cargo.toml
```diff
@@ -46,6 +46,10 @@ text = ["dep:skrifa"]
 
 # Development only features
 
+# Enable all debug utilities
+debug = ["debug-cmds"]
+# Enable command listing and introspection utilities
+debug-cmds = []
```
LaurenzV: Could we maybe instead use the `log` crate and print the debugging things using `debug!`? I think it's a bit unfortunate having this feature, but maybe not the end of the world.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -77,6 +79,10 @@ pub struct Wide<const MODE: u8 = MODE_CPU> {
     /// Initialized with node 0 (the root node representing the final output).
     /// As layers with filters are pushed, their node IDs are added to this stack.
     filter_node_stack: Vec<usize>,
+    /// True when currently rendering inside a filtered layer with a clip path.
+    /// When set, command generation uses full viewport bounds instead of clip bounds
+    /// to ensure filter effects can process the full layer before applying the clip.
+    in_clipped_filter_layer: bool,
```
LaurenzV: Not for this PR, but I'm wondering whether we should switch `vello_cpu` to always use the new clipping approach. Especially with filters now it seems like there are a lot of edge-case interactions between layer clipping and filters, which would likely be completely eliminated if we always used the new algorithm. On the other hand, it would probably be problematic to lose the parity between vello_cpu and vello_hybrid...

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -237,15 +243,15 @@ impl WideTilesBbox {
...
-        right_px: u16,
-        bottom_px: u16,
-        max_x: u16,
-        max_y: u16,
-    ) -> Self {
+    pub fn expand_by_pixels(&self, expansion: Rect, max_x: u16, max_y: u16) -> Self {
+        // The expansion rect is centered at origin:
+        // - Negative coordinates (x0, y0) represent left/top expansion
+        // - Positive coordinates (x1, y1) represent right/bottom expansion
+        let left_px = (-expansion.x0).max(0.0).ceil() as u16;
```
LaurenzV, line 250: Don't we need `floor` for the left and top pixels?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -725,6 +734,11 @@ impl<const MODE: u8> Wide<MODE> {
                 }
             }
         }
+
+        // Reset the flag after popping the filtered layer with clip
+        if layer.filter.is_some() && layer.clip {
+            self.in_clipped_filter_layer = false;
+        }
```
LaurenzV: What if we have two nested filter layers with clip? Won't that reset it for the first layer as well? Don't we need a stack of booleans here again?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1392,6 +1427,40 @@ impl<const MODE: u8> WideTile<MODE> {
...
+    /// let commands = wide_tile.list_commands();
+    /// println!("{}", commands);
+    /// // Output:
+    /// // 0: PushBuf(Regular)
+    /// // 1: FillPath
+    /// // 2: PushZeroClip
+    /// // 3: FillPath
+    /// // 4: PopBuf
+    /// ```
+    pub fn list_commands(&self) -> String {
```
LaurenzV: Does this need to be pub or is pub(crate) enough? And as mentioned, I think I personally would prefer not having a feature gate for this and just using the `log` crate.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -214,6 +214,10 @@ impl SingleThreadedDispatcher {
                     // Apply the filter effect to the completed layer.
                     fine.filter_layer(&mut pixmap, filter, layer_manager, *transform);
 
+                    // Save the filtered pixmap to disk for debugging.
+                    #[cfg(feature = "debug-filters")]
+                    save_filtered_layer_debug(&pixmap, *layer_id);
```
LaurenzV: I know it's a bit annoying if you want to use this multiple times, but maybe it would be better to just store the snippet locally and paste it whenever you need it?

sparse_strips/vello_example_scenes/src/lib.rs
```diff
@@ -304,19 +310,62 @@ impl<T: RenderingContext> AnyScene<T> {
...
+
+    /// Toggle the tile grid overlay.
+    pub fn toggle_tile_grid(&mut self) {
+        self.show_widetile_columns = !self.show_widetile_columns;
+    }
+
+    /// Draw the tile grid overlay.
+    fn draw_widetile_columns(&self, ctx: &mut T) {
+        ctx.set_transform(Affine::IDENTITY);
+        ctx.set_paint(WHITE);
```
LaurenzV, line 354: Do we need to undo these in the end?

sparse_strips/vello_example_scenes/src/lib.rs
```diff
@@ -304,19 +310,62 @@ impl<T: RenderingContext> AnyScene<T> {
...
+
+        // Draw lines across the entire scene
+        let max_width = 2000.0;
+        let max_height = 2000.0;
+
+        let mut tile_x = 0.0;
+        while tile_x <= max_width {
+            // Draw a thin vertical line
+            let line_width = 1.0;
```
LaurenzV: Can be lifted out of the loop.

sparse_strips/vello_common/src/coarse.rs
grebmeg: > Not for this PR, but I'm wondering whether we should switch vello_cpu to always use the new clipping approach. 

When you say "always use", what exactly do you mean? Should we just simply state explicitly that `push_clip_path` is the default clipping method? Btw, I’m not sure what the most correct terminology is for these two clipping methods? If there isn’t an established naming convention, what about calling them Immediate clipping (clip affects each draw directly — the new approach) and Deferred (or Layer-based) clipping?

>  Especially with filters now it seems like there are a lot of edge-case interactions between layer clipping and filters, which would likely be completely eliminated if we always used the new algorithm.

Yeah, you’re right about the edge cases, but I think it’s really hard to avoid them. Even the new approach doesn’t fully work, because it clips the path immediately. With filters like `DropShadow` that shift the filtered output, this prevents the entire path from being rendered into the filter layer, which is necessary for the filter to be applied and offset correctly.

> On the other hand, it would probably be problematic to lose the parity between vello_cpu and vello_hybrid...

I assume we’d still keep that method on `vello_hybrid`, but it would just run entirely on CPU, right?

sparse_strips/vello_common/src/coarse.rs
grebmeg: `ceil` is used here because the expansion values represent distances to expand in each direction. For example, if `expansion.x0 = -2.3`, that means "expand 2.3 pixels to the left". The code computes:

```rust
left_px = (-(-2.3)).max(0.0).ceil() = ceil(2.3) = 3
```

sparse_strips/vello_common/src/coarse.rs
grebmeg: Yes, you’re absolutely right! Thanks for catching that!

Fixed.

sparse_strips/vello_common/Cargo.toml
grebmeg: I think using [debug_assertions](https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions) would be even better?

Fixed.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
grebmeg: I initially did that, but I got tired of switching branches and then having to copy it back each time. I could automate it for myself, of course, but I figured this snippet would be convenient for others when debugging. If you don’t mind, I’ll remove the feature gate, put those lines behind `debug_assertions`, and put them under the comment.

sparse_strips/vello_example_scenes/src/lib.rs
grebmeg: Thanks!

Fixed.

sparse_strips/vello_example_scenes/src/lib.rs
grebmeg: We could undo it, but I don’t think it’s necessary in this case because wide tiles columns are only rendered at the end of `render`, and nothing else should run afterward in the same render cycle.

sparse_strips/vello_common/src/coarse.rs
grebmeg: I think `pub(crate)` is sufficient. I also removed all the debugging feature gates and now rely solely on `debug_assertions`.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -728,6 +739,11 @@ impl<const MODE: u8> Wide<MODE> {
                 }
             }
         }
+
+        // Decrement the depth counter after popping a filtered layer with clip
+        if layer.filter.is_some() && layer.clip {
+            self.clipped_filter_layer_depth = self.clipped_filter_layer_depth.saturating_sub(1);
+        }
```
LaurenzV: Can't we use the same condition as above in?

```
if in_clipped_filter_layer {
            self.clipped_filter_layer_depth += 1;
        }
```

so that it's really synchronous? I also think the saturating sub shouldn't be necessary then?

LaurenzV (approved):

sparse_strips/vello_common/src/coarse.rs
grebmeg: Sure, you’re right, it’s not necessary to use saturation here.

Fixed.

PR #1307, opened 2025-11-26
Fix bitmap glyph skew origin

devunt:
Bitmap glyphs use Y-down coordinate system (top at y=0), while outline/COLR fonts use Y-up (baseline at y=0). When `glyph_transform` (skew for synthetic italic) is applied to bitmap glyphs, the skew origin is at the top-left corner instead of the baseline, causing the italic effect to lean in the wrong direction.

## Fix

- Negate the shear coefficients to account for Y-down coordinates
- Apply the skew around the bottom of the image (y=height) instead of the top (y=0)

## Screenshots
CleanShot 2025-11-26 at 22 32 26
CleanShot 2025-11-26 at 22 32 50

### DIFF
--- a/vello/src/scene.rs
+++ b/vello/src/scene.rs
@@ -753,7 +753,14 @@ impl<'a> DrawGlyphs<'a> {
                         }),
                     };
                     if let Some(glyph_transform) = self.run.glyph_transform {
-                        transform *= glyph_transform.to_kurbo();
+                        let gt = glyph_transform.to_kurbo();
+                        let [a, b, c, d, e, f] = gt.as_coeffs();
+                        let corrected = Affine::new([a, -b, -c, d, e, f]);
+                        let h = f64::from(image.image.height);
+                        transform = transform
+                            * Affine::translate((0.0, h))
+                            * corrected
+                            * Affine::translate((0.0, -h));
                     }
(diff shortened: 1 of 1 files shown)


DJMcNab:
I think it would probably be easier to review if you merged this and #1306 into a single PR, as they're touching very similar code.

Would you be willing to add some tests here? I believe that the existing test in https://github.com/linebender/vello/blob/a04070aeecf4a00ed885b1052f1dfbfe84c539b5/vello_tests/tests/emoji.rs#L160-L173 has most of the infra you'd need, and just needs to be copied to add a more interesting transform. You'll need to have git lfs installed.

My gut reaction here is that a much simpler solution is possible, possibly to both issues. But for awareness, I haven't looked too closely. Essentially, the issue in both cases is that the `glyph_transform` is applied before or after another transform in the pipeline, so the transform doesn't apply properly. I just want to double check that you have confirmed that there aren't any solutions of that form for this issue.

This is related to https://github.com/linebender/vello/pull/1273, but there's nothing actionable to be taken from that PR here.

PoignardAzur:
At this point, I don't think we should expect @devunt to follow up on this. Should we review the PRs anyway? Leave them as-is?

devunt:
Sorry for the late reply. I'll look up on this.

Issue #1302, opened 2025-11-23
Panic rendering gradients


nicoburns:
I'm seeing a few panics in the current gradient LUT code in `vello_cpu`.

May be fixed by https://github.com/linebender/vello/pull/1301

```
range end index 4 out of range for slice of length 3
Panicked at /Users/nico/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vello_common-0.0.4/src/encode.rs:1023:42
Backtrace:
   7: core::slice::index::slice_index_fail::do_panic::runtime
   8: core::slice::index::slice_index_fail
   9: vello_common::encode::GradientLut<T>::new
  10: core::cell::once::OnceCell<T>::try_init
  11: vello_cpu::fine::Fine<S,T>::fill
  12: vello_cpu::fine::Fine<S,T>::run_cmd
  13: <vello_cpu::dispatch::single_threaded::SingleThreadedDispatcher as vello_cpu::dispatch::Dispatcher>::rasterize
  14: wpt::test_runners::ref_test::render_html_to_buffer
  15: wpt::test_runners::ref_test::process_ref_test
  16: wpt::test_runners::process_test_file
  17: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &F>::call_mut
  18: rayon::iter::plumbing::Folder::consume_iter
  19: rayon::iter::plumbing::bridge_producer_consumer::helper
  20: rayon_core::job::StackJob<L,F,R>::run_inline
  21: rayon_core::join::join_context::{{closure}}
  22: rayon::iter::plumbing::bridge_producer_consumer::helper
  23: rayon_core::job::StackJob<L,F,R>::run_inline
  24: rayon_core::join::join_context::{{closure}}
  25: rayon::iter::plumbing::bridge_producer_consumer::helper
  26: rayon_core::join::join_context::{{closure}}
  27: rayon::iter::plumbing::bridge_producer_consumer::helper
  28: rayon_core::job::StackJob<L,F,R>::run_inline
  29: rayon_core::join::join_context::{{closure}}
  30: rayon::iter::plumbing::bridge_producer_consumer::helper
  31: rayon_core::join::join_context::{{closure}}
  32: rayon::iter::plumbing::bridge_producer_consumer::helper
  33: <rayon_core::job::StackJob<L,F,R> as rayon_core::job::Job>::execute
  34: rayon_core::registry::WorkerThread::wait_until_cold
  35: rayon_core::join::join_context::{{closure}}
[...]

PR #1311, opened 2025-12-01
vello_common: harden gradient LUT against invalid stop positions

waywardmonkeys:
- Sanitize gradient stop positions with `sanitize_stop_position` to clamp NaNs, infinities, and out-of- range values into [0, 1].
- Use sanitized positions in `determine_lut_size` and LUT ramp construction to ensure indices are finite, monotonic, and within bounds.
- Keep the 4-wide write in `GradientLut::new` bounded by `lut_size` so very small LUTs cannot trigger end- index panics.
- Add targeted tests to verify sanitization behavior and ensure `GradientLut` handles infinite and out-of-range stop offsets without panicking.

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -37,6 +37,20 @@ const DEGENERATE_THRESHOLD: f32 = 1.0e-6;
 const NUDGE_VAL: f32 = 1.0e-7;
 const PIXEL_CENTER_OFFSET: f64 = 0.5;
 
+/// Sanitize a stop position so that it is finite and in the `[0.0, 1.0]` range.
+///
+/// This is intended as a last line of defence for inputs that bypass CSS-style validation or
+/// encode semantics directly in terms of ranges.
+fn sanitize_stop_position(pos: f32) -> f32 {
+    if pos.is_nan() {
+        0.0
+    } else if pos.is_infinite() {
+        if pos.is_sign_negative() { 0.0 } else { 1.0 }
+    } else {
+        pos.clamp(0.0, 1.0)
+    }
+}
+
 #[cfg(feature = "std")]
 fn exp(val: f32) -> f32 {
     val.exp()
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
This should fix against the panic mentioned by @nicoburns  in a comment in #1302.

Some notes / questions:

* We should probably improve some handling of this with Peniko.
* Callers should definitely do better.
* We should probably document these invariants better?
* Should I `debug_assert` instead of clamping / sanitizing to make callers fix their code? Or debug_assert, but still sanitize for non-debug-assertions builds?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -37,6 +37,20 @@ const DEGENERATE_THRESHOLD: f32 = 1.0e-6;
...
+fn sanitize_stop_position(pos: f32) -> f32 {
+    if pos.is_nan() {
+        0.0
+    } else if pos.is_infinite() {
+        if pos.is_sign_negative() { 0.0 } else { 1.0 }
+    } else {
+        pos.clamp(0.0, 1.0)
+    }
+}
+
```
tomcur, line 53: The following is a branchless way to do this.

```suggestion
#[inline]
fn sanitize_stop_position(pos: f32) -> f32 {
    pos.max(0.).min(1.)
}
```

`f32:{max,min}` don't propagate NaNs (so the NaN gets turned in to `0.`, if `pos.max(0.)` is first).

https://godbolt.org/z/nrzG75b1K

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3cd614c3df1900afdee1a4c481e5625a

LaurenzV:
I'm actually surprised this doesn't already work, because we already are checking that each stop is between 0 and 1...

waywardmonkeys:
I will figure out why and fix it!

waywardmonkeys:
(I already know why, just have to drive home)

LaurenzV:
This should already be resolved!

Issue #1278, opened 2025-10-16
Gradients in hybrid don't work as expected within a clip layer


ajakubowicz-canva:
Minimal repro test (which I reproduced by adding to vello_sparse_tests clip.rs):

```rs
#[vello_test(width = 600, height = 200)]
fn clip_with_gradient_fill(ctx: &mut impl Renderer) {
    use crate::util::stops_green_blue;
    // Border to see crop.
    ctx.set_paint(PINK);
    ctx.fill_rect(&Rect::new(0., 0., 600., 200.));

    // Crop 10px border around image.
    let clip = Rect::new(10., 10., 590., 190.0);
    ctx.push_clip_layer(&clip.to_path(0.));
    
    // Fill a full screen rect with a gradient.
    let box_rect = Rect::new(0., 0., 600.0, 200.0);
    
    let gradient = Gradient {
        kind: LinearGradientPosition {
            start: Point::new(100.0, 100.0),
            end: Point::new(300.0, 100.0),
        }.into(),
        stops: stops_green_blue(),
        ..Default::default()
    };
    
    ctx.set_paint(gradient);
    ctx.fill_rect(&box_rect);
    
    ctx.pop_layer();
}
```

### Expected Image

<img width="600" height="200" alt="Image" src="https://github.com/user-attachments/assets/fe51ea75-c056-4364-862d-52668b6c1009" />

Note – this image is consistent with Vello CPU.

### Hybrid's incorrect result

<img width="1800" height="200" alt="Image" src="https://github.com/user-attachments/assets/188f3157-75e6-497b-a556-f1fd903ece6f" />

LaurenzV:
This was fixed in #1313.

PR #1313, opened 2025-12-03
Gradients in hybrid work as expected within a clip layer

RobertBrewitz:
I grabbed the test from https://github.com/linebender/vello/issues/1278

I had to resolve this for my own use, looks better now and the test passes.

Gradient

Screenshot from 2025-12-03 11-23-14

Before

Screenshot from 2025-12-03 11-23-33

After

Screenshot from 2025-12-03 11-23-45

### Linear

Before

clip_with_linear_gradient_fill_hybrid

After

clip_with_linear_gradient_fill

### Radial

Before

clip_with_radial_gradient_fill_hybrid

After

clip_with_radial_gradient_fill

### Sweep

Before

clip_with_sweep_gradient_fill_hybrid

After

clip_with_sweep_gradient_fill

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/gradient.rs
+++ b/sparse_strips/vello_sparse_tests/tests/gradient.rs
@@ -1096,13 +1096,13 @@ mod sweep {
         gradient_with_transform(ctx, transform, 25.0, 25.0, 75.0, 75.0);
     }
 
-    #[vello_test(hybrid_tolerance = 10)]
+    #[vello_test(diff_pixels = 4)]
     fn gradient_sweep_with_transform_skew_y_1(ctx: &mut impl Renderer) {
         let transform = Affine::translate((0.0, 50.0)) * Affine::skew(0.0, -tan_45());
         gradient_with_transform(ctx, transform, 25.0, 25.0, 75.0, 75.0);
     }
 
-    #[vello_test(hybrid_tolerance = 10)]
+    #[vello_test(diff_pixels = 3)]
     fn gradient_sweep_with_transform_skew_y_2(ctx: &mut impl Renderer) {
         let transform = Affine::translate((0.0, -50.0)) * Affine::skew(0.0, tan_45());
         gradient_with_transform(ctx, transform, 25.0, 25.0, 75.0, 75.0);


RobertBrewitz:
~I don't get why some of the snapshot images are missing in the ubuntu and macos test runners, I tried to remove and add them back because I thought it might be LFS, but that was not it.~

Edit: Ah, now I actually get real errors after adding the missing images to lfs.

RobertBrewitz:
Not sure what is going on with `gradient_sweep_with_transform_skew_y_1_hybrid` on macos-latest and `gradient_sweep_with_transform_skew_y_2_hybrid` on ubuntu-latest.

There is no visual diff here when I compare the test output with the existing snapshot.

Snapshot gradient_sweep_with_transform_skew_y_2

gradient_sweep_with_transform_skew_y_2

Output of test gradient_sweep_with_transform_skew_y_2 on hybrid

gradient_sweep_skew_y_2

The only difference is that the snapshot is RGB and the output of hybrid is RGBA.

RobertBrewitz:
Disabling two tests for vello_hybrid after some discussion here [#vello > ✔ vello_hybrid skew tests are a bit strange on ubuntu &amp; m...](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/.E2.9C.94.20vello_hybrid.20skew.20tests.20are.20a.20bit.20strange.20on.20ubuntu.20.26.20m.2E.2E.2E/with/561844619), diffs here:

macos-latest gradient_sweep_with_transform_skew_y_1 diff

gradient_sweep_with_transform_skew_y_1_hybrid

ubuntu-latest gradient_sweep_with_transform_skew_y_2 diff

gradient_sweep_with_transform_skew_y_2_hybrid

LaurenzV:
@grebmeg Maybe you could take a look?

sparse_strips/vello_sparse_tests/tests/clip.rs
```diff
@@ -394,3 +398,96 @@ fn clip_non_isolated_deeply_nested_circles(ctx: &mut impl Renderer) {
         }
     }
 }
+
+#[vello_test(width = 600, height = 200)]
+fn clip_with_linear_gradient_fill(ctx: &mut impl Renderer) {
+    use crate::util::stops_green_blue;
```
grebmeg: [nit]: Could you please move this import, along with the similar imports in the functions below, to the import section at the beginning of the file?

sparse_strips/vello_sparse_tests/tests/clip.rs
```diff
@@ -394,3 +398,96 @@ fn clip_non_isolated_deeply_nested_circles(ctx: &mut impl Renderer) {
...
+    // Border to see crop.
+    ctx.set_paint(PINK);
+    ctx.fill_rect(&Rect::new(0., 0., 600., 200.));
```
grebmeg: I think it can be avoided because our default parameter is a white background, and you can use the `transparent` parameter for the tests if you need transparency, but in this tests I think a white background   is what we need.

sparse_strips/vello_sparse_tests/tests/gradient.rs
```diff
@@ -1096,13 +1096,13 @@ mod sweep {
...
 
-    #[vello_test]
+    #[vello_test(skip_hybrid)]
     fn gradient_sweep_with_transform_skew_y_1(ctx: &mut impl Renderer) {
         let transform = Affine::translate((0.0, 50.0)) * Affine::skew(0.0, -tan_45());
         gradient_with_transform(ctx, transform, 25.0, 25.0, 75.0, 75.0);
     }
 
-    #[vello_test]
+    #[vello_test(skip_hybrid)]
```
grebmeg: Instead of fully skipping these tests, can we use `hybrid_tolerance` or `diff_pixels` parameter? Ideally, the values should differ by a small amount between channels, so you can use the first parameter to set the threshold value. If the value is quite large, I would say more than 10px, then I recommend using the second parameter.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -246,6 +246,16 @@ fn vs_main(
                 + encoded_image.transform.zw * f32(scene_strip_y)
                 + encoded_image.transform.xy * x * f32(width)
                 + encoded_image.transform.zw * y * f32(config.strip_height);
+        } else if paint_type == PAINT_TYPE_LINEAR_GRADIENT || paint_type == PAINT_TYPE_RADIAL_GRADIENT || paint_type == PAINT_TYPE_SWEEP_GRADIENT {
+            // Unpack view coordinates for gradient calculation
+            let scene_strip_x = instance.payload & 0xffffu;
+            let scene_strip_y = instance.payload >> 16u;
```
grebmeg: Maybe let’s move these two lines, along with the same code from the branch above, outside of the the branches (we can place it right after `let paint_type = ...`)?

sparse_strips/vello_sparse_tests/tests/gradient.rs
RobertBrewitz: Hybrid tolerance 10 did not help, using diff pixels instead.

PR #1315, opened 2025-12-05
Use naga from wgpu 27

waywardmonkeys:


DJMcNab (approved):
Seems reasonable, as we're already on the wgpu 27 release. We probably should move this to a workspace dependency (with the features specified per-crate?), so that it's more obvious that they should be updated in lockstep.

waywardmonkeys:
Someone (maybe even me) can do that later. I need this in so that I can land a change in Masonry so that I can stop working for the weekend.

PR #1316, opened 2025-12-05
Move `naga` dependency to workspace

tomcur:
See https://github.com/linebender/vello/pull/1315#pullrequestreview-3544948356.


tomcur:
> I could justify adding a new comment in the root Cargo.toml (i.e. make sure to keep these in sync), but I don't really think that's actually needed.

For posterity!

PR #1318, opened 2025-12-08
Reorder packed_winding_line_idx for sorting

b0nes164:
Changes as discussed to fix sorting issues caused by `intersection_mask` occupying the upper bits of the `packed_winding_line_idx`.

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -46,14 +46,18 @@ pub struct Tile {
     /// and winding data packed together.
     ///
     /// The layout is:
-    /// - **Bits 0-25 (26 bits):** The line index (`line_idx`).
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_common/src/tile.rs
```diff
@@ -608,20 +617,11 @@ impl Tiles {
                 }
             } else {
                 // Case: Line is fully contained within a single tile.
-                // TODO: Must exactly match the general case? Or can be simplified?
```
LaurenzV: This probably belongs to the other PR. 😄

b0nes164: It should be "upstream of the other PR," not familiar with how this goes on github. . . let me merge the other pr and then I'll rebase onto main.

PR #1319, opened 2025-12-09
Separate make_tiles into AA and MSAA versions. Fully cases out MSAA version

b0nes164:


sparse_strips/vello_common/src/tile.rs
```diff
@@ -301,12 +301,7 @@ impl Tiles {
     // TODO: Tiles are clamped to the left edge of the viewport, but lines fully to the left of the
     // viewport are not culled yet. These lines impact winding, and would need forwarding of
     // winding to the strip generation stage.
-    pub fn make_tiles<const GEN_INT_MASK: bool>(
-        &mut self,
-        lines: &[Line],
-        width: u16,
-        height: u16,
-    ) {
+    pub fn make_tiles_analytic_aa(&mut self, lines: &[Line], width: u16, height: u16) {
```
LaurenzV, line 306: Maybe we can add a 2-3 sentences to that method analogously to the MSAA one? (like the "Generates tile commands for MSAA (Multisample Anti-Aliasing) rasterization." one

sparse_strips/vello_common/src/tile.rs
```diff
@@ -362,27 +357,246 @@ impl Tiles {
...
-            // point will not be an intersection. Because it is not an intersection, the rasterizing
-            // algorithm will use the raw point, which can potentially lead to a watertightness
-            // difference with the point calculated by the top tile.
-            //
-            // The solution is to check whether the bottommost point in the line is perfectly
-            // axis-aligned, and if that is true, to adjust the location of the bottom tile, such
-            // that a bottom touch is not considered an intersection.
             let p1_tile_y = if line_bottom_y == line_bottom_y_ceil {
                 line_bottom_y as i32 - 1
             } else {
```
LaurenzV, line 604: Didn't we change this to just `floor` before? Seems like the analytical_aa method uses this, so maybe just a copy-paste error? Or does it have to be different for MSAA?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -423,196 +632,302 @@ impl Tiles {
...
-                                0b11 | winding
-                            } else {
-                                winding
-                            };
+                            let intersection_mask = 0b100011; // W | B | T
                             let tile = Tile::new_clamped(x, y_idx, line_idx, intersection_mask);
                             self.tile_buf.push(tile);
                         }
 
                         // Perfect touching B case.
```
LaurenzV, line 642: What does "B case" mean (I guess that comment already existed before tho)?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -423,196 +632,302 @@ impl Tiles {
...
+
+                    // Check if the line is fully within the horizontal viewport bounds. If it is,
+                    // we can skip the min/max clamping per row.
+                    let min_x = p0_x.min(p1_x);
+                    let max_x = p0_x.max(p1_x);
+                    // Note: We use >= on the right edge to ensure strictly safe integer truncation
+                    let needs_clamping = min_x < line_left_x || max_x >= line_right_x;
+
+                    // Handles the bitmask logic for start/end tiles. Invariant to clamping.
+                    macro_rules! push_edge {
```
LaurenzV, line 678: I think it's a bit unfortunate to use a macro here, is it because you need mutable access to the tile_buf? If so, you could just explicitly pass the buffer into the closure each time?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -723,17 +1066,18 @@ mod tests {
...
-                Tile::new(1, 0, 1, W | T),
-                Tile::new(2, 0, 2, W | T),
-                Tile::new(0, 0, 3, W | T),
-            ]
-        );
+        tiles.make_tiles_msaa(&lines, VIEW_DIM, VIEW_DIM);
+        assert_eq!(tiles.tile_buf, expected);
+
+        tiles.make_tiles_analytic_aa(&lines, VIEW_DIM, VIEW_DIM);
+        check_analytic_aa_matches(&tiles.tile_buf, &expected);
```
LaurenzV: Since there are a bit repetitive, could you maybe just create a impl block with a new function (inside of `mod tests`) and call those?

```rs
    impl Tiles {
        fn make_all(&mut self, lines: &[Line], width: u16, height: u16, expected: &[Tile]) {
            self.make_tiles_msaa(&lines, VIEW_DIM, VIEW_DIM);
            assert_eq!(self.tile_buf, expected);

            self.make_tiles_analytic_aa(&lines, VIEW_DIM, VIEW_DIM);
            check_analytic_aa_matches(&self.tile_buf, &expected);
        }
    }
```

LaurenzV (approved):
A bit unfortunate, but if it's necessary to split it up, it's necessary. 😄

sparse_strips/vello_common/src/tile.rs
b0nes164: My primary motivation for using a macro was to allow the compiler to perform dead code elimination on the hard coded logic paths. Do you think the compiler will inline and DCE on the closure?

sparse_strips/vello_common/src/tile.rs
b0nes164: I was unsure if we could remove it, so I left it in. Unfortunately, I think we have to keep it in. See the "B case" comment.

sparse_strips/vello_common/src/tile.rs
LaurenzV: You could mark the closures with `#[inline(always)]`, but hard to say without trying it. But unless there really is a performance benefit I think the closure version is easier to read, but I'll leave it to you!

sparse_strips/vello_common/src/tile.rs
b0nes164: A quirk of the tiling is that because our vertical loop is exclusive, an exact bottom edge touch does not produce an additional tile. (By contrast, a right edge touch *will* produce a new tile). This exclusive behavior is correct, the line is not actually crossing, and producing an empty tile would incorrectly increment winding in that scanline. However, a side effect of this is that the bottommost tile produced by this case will not count itself as end_tile without this line. 

This can have two potential effects: 
1) Watertightness: if one tile counts the endpoint as a "touch" and another counts the endpoint as an "intersection," the touch case will use the raw line endpoint, and the intersection will actually calculate the tile/line intersection. This can potentially lead to two different points between tiles.

2) Tie breaking:  In our updated tiler, we now "tie break" corners inside `make_tiles.` So we never produce a mask with more than 2 bits set. (For reasons we can discuss later, I'm still keeping the perfect corner bit, but it can be safely removed in the future). To do this however, we need to know if we are in a line endpoint tile. This is because a line endpoint may "perfectly touch" a corner, but should not be tie-broken. (Versus a non-endpoint tile, where we know a corner touch is always an intersection, and should always be tie broken).
See the logic here:
https://github.com/linebender/vello/blob/2b30a462b67f7c7fb62f3ca79e6205745159282b/sparse_strips/vello_common/src/tile.rs#L711-L718 

Case 1, I don't think will produce a visual difference. But Case 2 means that we have to keep it in.

b0nes164: Example, courtesy of gemini:

### 1. The Setup
* **Grid:** Tile 1 ends at `y=2.0`.
* **Line:** Vertical, ends **exactly** at boundary `y=2.0`.
* **Processing:** Inside the loop for **Tile 1** (`y_idx = 1`).

### 2. The Logic Failure
The bitmask logic uses `is_end_tile` to decide if the line *stops* or *passes through*.

| State | Without Fix (`p1_tile_y = 2`) | With Fix (`p1_tile_y = 1`) |
| :--- | :--- | :--- |
| **Logic** | `1 != 2` → `is_end_tile = false` | `1 == 1` → `is_end_tile = true` |
| **Mask** | `vert_exit = 1` | `vert_exit = 0` |
| **Meaning** | "I am crossing into Tile 2" | "I stopped at the boundary" |
| **Result** | **Topology Error** | **Watertight** |

### 3. Visual Representation

```text
       Tile 1 (Row 1)
      +---------------+
      |       |       |
      |       |       |
      |       V       |  <-- Real Endpoint (2.0)
      +-------|-------+
              |          <-- PHANTOM EXIT (The Bug)
              ?              The mask says "Exit", but Tile 2 is empty.
       Tile 2 (Row 2)        The renderer sees an "open" shape.
      +---------------+
      |               |
      |   (Empty)     |
      |               |
      +---------------+

sparse_strips/vello_common/src/tile.rs
b0nes164: Let me test this and get back to you. I agree closure would be cleaner, but if there is a difference I think the macro is good enough-ish. :)

b0nes164: The macro appears to be slightly faster, and on repeated runs, tends to fall into the "fast hot path" faster than the closure.

dominikh:
@b0nes164 After separating `make_tiles` into analytic AA and MSAA versions, is `make_tiles_analytic_aa` functionally different from the original `make_tiles`, before we ever tracked intersections (i.e. before #1293)? I'm asking because `make_tiles_analytic_aa` seems quite a bit more complex than the old `make_tiles` and I'm trying to understand if that complexity serves a purpose, or is a remnant of #1293.

b0nes164:
Good question;`make_tiles_analytic_aa` is functionally the same. 

I added the "start/middle/end" casing logic to it because I had intended on integrating the culling functionality directly into those loops. In the culling scenario, you really want the casing because you only need to run the complicated partial winding calculations in the start and end rows (see here on the active culling PR):
https://github.com/linebender/vello/blob/7003e5ec1ff5072b983456b2ea26d7917cf209ec/sparse_strips/vello_common/src/tile.rs#L407-L479

> I'm trying to understand if that complexity serves a purpose

Given the way the culling is turning out (where the culling is an early exit distinct block), I may remove the "s/m/e"casing from the non-culled code path. I would have to test the perf. One of the benefits of the casing is that it actually simplifies the winding checks because again, you only need to check for winding contribution in the start row, because, by definition, middle and end rows must have had a top edge crossing.

Edit:
For more context, originally, I was only planning on early culling rows where the line fully traversed the height of the tile. (i.e. captive or middle rows). This was because I wanted to keep the histogram as `i8.` So it was hard requirement to integrate the culling into the general case path, because you still needed to push clamped tiles in the start and end rows. 

However, the middle-only culling performance turned out to be underwhelming. A side effect of flattening is that there are often many small lines in a path, which typically produce only one or two tiles. So if you only cull "middle" rows, these lines don't get culled.

PR #1320, opened 2025-12-09
Preparatory fearless_simd changes

valadaptive:
The next release of fearless_simd will remove some redundant API surfaces and probably make the `val` field of vector types inaccessible (https://github.com/linebender/fearless_simd/pull/159). In preparation, this PR switches over from using the to-be-removed methods to their replacements:

- Instead of taking a reference to the `val` field, all our use cases let us use `as_slice`
- For getting the value as an array, we can use `.into()`
- Instead of the `reinterpret` methods, we can use the `Bytes` trait
- Instead of indexing on the `val` field, we can index on the type directly. I thought this wouldn't work since we haven't implemented `Index`/`IndexMut` on the vector types in fearless_simd 0.3, but I guess since the vector types all `Deref` to their respective array types, it does?
- Instead of the `.shr` method, we can use the shift right operator (`>>`)
- Instead of `cvt_u32` and `cvt_f32`, we can use the `to_int` and `to_float` methods

The only *breaking* changes that we can't make yet are changing `Level::fallback()` to `Level::baseline()` and changing `madd`/`msub` to `mul_add`/`mul_sub`.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/lowp/gradient.rs
+++ b/sparse_strips/vello_cpu/src/fine/lowp/gradient.rs
@@ -44,7 +44,7 @@ impl<S: Simd> Iterator for GradientPainter<'_, S> {
         let indices = (t_vals * self.scale_factor).to_int::<u32x16<S>>();
 
         let mut vals = [0_u8; 64];
-        for (val, idx) in vals.chunks_exact_mut(4).zip(indices.val) {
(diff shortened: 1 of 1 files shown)


LaurenzV:
> madd/msub to mul_add/mul_sub

Wasn't this already done in a previous PR?

LaurenzV:
Oh sorry! I didn't realize this was the vello repo, not fearless_simd. 😂

PR #1322, opened 2025-12-11
Update docs for `vello_cpu::RenderMode`

T-256:
- Follow-up to #1294


nicoburns:
There's possibly more we can here, but at least mentioning this seems like an improvement :)

PR #1324, opened 2025-12-15
Update to newest fearless_simd

LaurenzV:
While I know we have decided to try to mostly stick to released versions, having vello_cpu point to the current main of fearless_simd will make it easier to benchmark some upcoming changes.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/lowp/gradient.rs
+++ b/sparse_strips/vello_cpu/src/fine/lowp/gradient.rs
@@ -34,20 +34,27 @@ impl<'a, S: Simd> GradientPainter<'a, S> {
 }
 
 impl<S: Simd> crate::fine::Painter for GradientPainter<'_, S> {
+    #[inline(never)]
     fn paint_u8(&mut self, buf: &mut [u8]) {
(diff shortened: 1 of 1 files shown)


grebmeg:
Would you be able to publish `vello_cpu` now if it has a git dependency? From what I’m reading in the docs, although they aren’t very explicit, the manifest for such a publish would be rejected.

Also, why not publish `fearless_simd` and use the published version here instead?

valadaptive:
> Would you be able to publish `vello_cpu` now if it has a git dependency? From what I’m reading in the docs, although they aren’t very explicit, the manifest for such a publish would be rejected.

I believe we would not.

> Also, why not publish `fearless_simd` and use the published version here instead?

We want to benchmark some upcoming changes that should make it in before the next release.

LaurenzV:
I mean, I wouldn’t mind making another release of fearless_simd now before we make further changes. I can try to do it today.

raphlinus:
Can this be closed now that we've released sparse strips 0.0.5?

LaurenzV:
I don't think so, we still need to bump fearless_simd eventually.

PR #1325, opened 2025-12-16
Reduce the memory footprint of wide tile commands

LaurenzV:
This PR reduces the memory footprint of wide tile commands from 48 bytes to 16 bytes. This is achieved by introducing a number of changes:

- For fill/alpha fills, we store the core properties (paint, mask, etc.) in a separate buffer that is built on a per-path basis, and the commands now simply contain an index that allows us to fetch the properties once we need them without having to duplicate them in each command.
- A similar pattern is used for clip commands.
- Clip commands now use u16 instead of u32 for x/width.
- Instead of storing the alpha index as a usize, we store an absolute offset on a per-path basis, and each command only stores the relative offset to the index of the path it belongs to. By doing so, using u32 for each command to store the alpha index is sufficiently large.

This leads to some nice speedups in many cases:
image
image

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -72,7 +72,7 @@ pub struct Wide<const MODE: u8 = MODE_CPU> {
     /// The wide tiles in the container.
     pub tiles: Vec<WideTile<MODE>>,
     /// Shared command properties, referenced by index from fill and clip commands.
-    pub props: CommandAttrs,
+    pub attrs: CommandAttrs,
     /// The stack of layers.
     layer_stack: Vec<Layer>,
     /// The stack of active clip regions.
@@ -304,7 +304,7 @@ impl<const MODE: u8> Wide<MODE> {
             tiles,
             width,
             height,
-            props: CommandAttrs::default(),
+            attrs: CommandAttrs::default(),
             layer_stack: vec![],
             clip_stack: vec![],
             // Start with root node 0.
@@ -325,7 +325,7 @@ impl<const MODE: u8> Wide<MODE> {
             tile.cmds.clear();
             tile.layer_ids.truncate(1);
         }
-        self.props.clear();
+        self.attrs.clear();
         self.layer_stack.clear();
         self.clip_stack.clear();
         self.filter_node_stack.truncate(1);
@@ -413,9 +413,9 @@ impl<const MODE: u8> Wide<MODE> {
 
         let alpha_base_idx = strip_buf[0].alpha_idx();
 
-        // Create shared properties for all commands from this path
-        let props_idx = self.props.fill.len() as u32;
-        self.props.fill.push(FillProps {
+        // Create shared attributes for all commands from this path
+        let attrs_idx = self.attrs.fill.len() as u32;
+        self.attrs.fill.push(FillAttrs {
             thread_idx,
             paint,
             blend_mode,
@@ -489,7 +489,7 @@ impl<const MODE: u8> Wide<MODE> {
                     x: x_wtile_rel,
                     width,
                     alpha_offset: col * u32::from(Tile::HEIGHT) - alpha_base_idx,
-                    props_idx,
+                    attrs_idx,
                 };
                 x += width;
                 col += u32::from(width);
@@ -516,9 +516,9 @@ impl<const MODE: u8> Wide<MODE> {
                     .min(bbox.x1())
                     .min(WideTile::MAX_WIDE_TILE_COORD);
 
-                let fill_props = &self.props.fill[props_idx as usize];
-                let override_color = match &fill_props.paint {
-                    Paint::Solid(s) if s.is_opaque() && fill_props.mask.is_none() => Some(*s),
+                let fill_attrs = &self.attrs.fill[attrs_idx as usize];
+                let override_color = match &fill_attrs.paint {
+                    Paint::Solid(s) if s.is_opaque() && fill_attrs.mask.is_none() => Some(*s),
                     _ => None,
                 };
 
@@ -535,7 +535,7 @@ impl<const MODE: u8> Wide<MODE> {
                     self.get_mut(wtile_x, strip_y).fill(
                         x_wtile_rel,
                         width,
-                        props_idx,
+                        attrs_idx,
                         current_layer_id,
                         override_color,
                     );
@@ -943,10 +943,10 @@ impl<const MODE: u8> Wide<MODE> {
             return;
         }
 
-        // Compute base alpha index and create shared clip properties
+        // Compute base alpha index and create shared clip attributes
         let alpha_base_idx = strips[0].alpha_idx();
-        let clip_props_idx = self.props.clip.len() as u32;
-        self.props.clip.push(ClipProps {
+        let clip_attrs_idx = self.attrs.clip.len() as u32;
+        self.attrs.clip.push(ClipAttrs {
             thread_idx,
             alpha_base_idx,
         });
@@ -1045,7 +1045,7 @@ impl<const MODE: u8> Wide<MODE> {
                     x: x_rel,
                     width,
                     alpha_offset: col * u32::from(Tile::HEIGHT) - alpha_base_idx,
-                    props_idx: clip_props_idx,
+                    attrs_idx: clip_attrs_idx,
                 };
                 x += width;
                 col += u32::from(width);
@@ -1214,7 +1214,7 @@ impl<const MODE: u8> WideTile<MODE> {
         &mut self,
         x: u16,
         width: u16,
-        props_idx: u32,
+        attrs_idx: u32,
         current_layer_id: LayerId,
         override_color: Option<PremulColor>,
     ) {
@@ -1247,7 +1247,7 @@ impl<const MODE: u8> WideTile<MODE> {
                         self.cmds.push(Cmd::Fill(CmdFill {
                             x,
                             width,
-                            props_idx,
+                            attrs_idx,
                         }));
                     }
                 }
@@ -1256,7 +1256,7 @@ impl<const MODE: u8> WideTile<MODE> {
                     self.cmds.push(Cmd::Fill(CmdFill {
                         x,
                         width,
-                        props_idx,
+                        attrs_idx,
                     }));
                 }
                 _ => unreachable!(),
@@ -1596,9 +1596,9 @@ impl Cmd {
     }
 }
 
-/// Shared properties for alpha fill commands.
+/// Shared attributes for alpha fill commands.
 #[derive(Debug, Clone, PartialEq)]
-pub struct FillProps {
+pub struct FillAttrs {
     /// The index of the thread that owns the alpha buffer
     /// containing the mask values at `alpha_idx`.
     /// Always 0 in single-threaded mode.
@@ -1616,16 +1616,16 @@ pub struct FillProps {
     alpha_base_idx: u32,
 }
 
-impl FillProps {
+impl FillAttrs {
     /// Compute the absolute alpha buffer index from a relative offset.
     pub fn alpha_idx(&self, offset: u32) -> u32 {
         self.alpha_base_idx + offset
     }
 }
 
-/// Shared properties for clip alpha fill commands.
+/// Shared attributes for clip alpha fill commands.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ClipProps {
+pub struct ClipAttrs {
     /// The index of the thread that owns the alpha buffer
     /// containing the mask values at `alpha_idx`.
     /// Always 0 in single-threaded mode.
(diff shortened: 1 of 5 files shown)


grebmeg:
Hey @LaurenzV, awesome work! I’m adding myself as a reviewer to indicate that I plan to review the PR, but I’ll likely be able to get to it on Monday. 🙏

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -71,6 +71,8 @@ pub struct Wide<const MODE: u8 = MODE_CPU> {
     pub height: u16,
     /// The wide tiles in the container.
     pub tiles: Vec<WideTile<MODE>>,
+    /// Shared command properties, referenced by index from fill and clip commands.
+    pub props: Props,
```
grebmeg: [nit] `Props` feels a bit generic to me. Would a more specific name better reflect its purpose? Maybe something like:

```rust
pub struct CmdAttrs {
    pub fill: Vec<FillAttrs>,
    pub clip: Vec<ClipAttrs>,
}
```

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -92,7 +94,8 @@ struct Clip {
     pub clip_bbox: WideTilesBbox,
     /// The rendered path in sparse strip representation
     pub strips: Box<[Strip]>,
-    #[cfg(feature = "multithreading")]
+    /// The index of the thread that owns the alpha buffer.
+    /// Always 0 in single-threaded mode.
```
grebmeg, line 98: If I’m calculating the struct size correctly, it’s now 32 bytes instead of 24 bytes in single-threaded mode, so we’re paying an additional 8 bytes even though this field isn’t used there. While we probably won’t have thousands of clips simultaneously, would it still make sense to keep this behind a feature gate?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -474,12 +488,8 @@ impl<const MODE: u8> Wide<MODE> {
                 let cmd = CmdAlphaFill {
                     x: x_wtile_rel,
                     width,
-                    alpha_idx: (col * u32::from(Tile::HEIGHT)) as usize,
-                    #[cfg(feature = "multithreading")]
-                    thread_idx,
-                    paint: paint.clone(),
-                    blend_mode,
-                    mask: mask.clone(),
+                    alpha_offset: col * u32::from(Tile::HEIGHT) - alpha_base_idx as u32,
```
grebmeg: [nit] Since this runs inside the loop, could we avoid casting to `u32` here and instead cast `alpha_base_idx` to `usize` when creating `FillProps` above?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -506,6 +516,21 @@ impl<const MODE: u8> Wide<MODE> {
...
+                    if let Paint::Solid(s) = &self.props.fill[props_idx as usize].paint {
+                        if s.is_opaque() && self.props.fill[props_idx as usize].mask.is_none() {
+                            Some(*s)
+                        } else {
+                            None
+                        }
+                    } else {
+                        // TODO: Implement for indexed paints.
+                        None
+                    };
```
grebmeg: I think this would require updating my image optimization #1327. I’d probably need to return an enum here so I can decide which path to take when generating fill commands. I’ll sort that out in my PR, but for now maybe something like this:

```suggestion
let fill_props = &self.props.fill[props_idx as usize];
let override_color = match &fill_props.paint {
    Paint::Solid(s) if s.is_opaque() && fill_props.mask.is_none() => Some(*s),
    _ => None,
};
```

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -924,11 +944,25 @@ impl<const MODE: u8> Wide<MODE> {
...
+            self.props.clip.push(ClipProps {
+                thread_idx,
+                alpha_base_idx,
+            });
+            (props_idx, alpha_base_idx)
+        } else {
+            // In case we have 0 strips, those variables won't be accessed
+            // anyway.
+            (u32::MAX, usize::MAX)
+        };
```
grebmeg: Could we early-return when there are no strips, similar to `generate`?


```suggestion
if n_strips == 0 {
    return;
}

let alpha_base_idx = strips[0].alpha_idx() as usize;
let clip_props_idx = self.props.clip.len() as u32;
self.props.clip.push(ClipProps {
    thread_idx,
    alpha_base_idx,
});
```

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1015,16 +1049,15 @@ impl<const MODE: u8> Wide<MODE> {
...
 
                 // Create clip strip command for rendering the partial coverage
                 let cmd = CmdClipAlphaFill {
                     x: x_rel,
-                    width: u32::from(width),
-                    alpha_idx: col as usize * Tile::HEIGHT as usize,
-                    #[cfg(feature = "multithreading")]
-                    thread_idx,
+                    width,
+                    alpha_offset: col * u32::from(Tile::HEIGHT) - alpha_base_idx as u32,
```
grebmeg: The same with casting `u32` here

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1586,48 +1608,101 @@ impl Cmd {
...
+    pub thread_idx: u8,
+    /// The paint (color, gradient, etc.) to fill the region with.
+    pub paint: Paint,
+    /// The blend mode to apply before drawing the contents.
+    pub blend_mode: BlendMode,
+    /// A mask to apply to the command.
+    pub mask: Option<Mask>,
+    /// Base index into the alpha buffer for this path's commands.
+    /// Commands store a relative offset that is added to this base.
+    alpha_base_idx: usize,
```
grebmeg: Would it make sense to shrink `usize` to `u32` here?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1586,48 +1608,101 @@ impl Cmd {
...
 
+/// Shared properties for alpha fill commands.
+#[derive(Debug, Clone, PartialEq)]
+pub struct FillProps {
+    /// The index of the thread that owns the alpha buffer
+    /// containing the mask values at `alpha_idx`.
+    /// Always 0 in single-threaded mode.
+    pub thread_idx: u8,
+    /// The paint (color, gradient, etc.) to fill the region with.
+    pub paint: Paint,
```
grebmeg, line 1609: Just an idea: `Paint` is 24 bytes, so what if we stored solid `PremulColor` values as indexed paints? Could that give us any performance benefits? Maybe we can store it as `paint_idx: u32` reducing the size from 24 bytes down to 4 bytes?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -616,13 +619,14 @@ impl Scheduler {
                     let el = state.stack.last_mut().unwrap();
                     let draw = self.draw_mut(el.round, el.get_draw_texture(depth));
 
-                    let col_idx = (alpha_fill.alpha_idx / usize::from(Tile::HEIGHT))
+                    let fill_props = &props.fill[alpha_fill.props_idx as usize];
+                    let alpha_idx = fill_props.alpha_idx(alpha_fill.alpha_offset);
+                    let col_idx = (alpha_idx / usize::from(Tile::HEIGHT))
                         .try_into()
                         .expect("Sparse strips are bound to u32 range");
```
grebmeg: This looks like a repeated pattern, so if it makes sense, maybe we could add `col_idx` method next to `alpha_idx`?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -301,6 +304,7 @@ impl<const MODE: u8> Wide<MODE> {
             tiles,
             width,
             height,
+            props: Props::default(),
```
grebmeg: Would it make sense to assign reasonable starting capacities for vectors in `Props`, specifically for `FillProps`? I think we can assume it will need to grow during the first render since it’s the primary data-holding structure. It might also be worth checking whether this shows any impact in benchmarks.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Is this really a problem though? That struct is only used to keep track of the clip stack, and realistically you aren't going to have more than a few nested clip path, so the impact on memory is very negligible I think (unlike the wide tile commands, where you could end up with thousands or even tens of thousands of commands depending on what you are drawing)? I could add it back, but the main reason I removed it is that it keeps the code simpler because we don't have to slap on the cfg attribute everywhere we use it.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: I just realized that the alpha index can never exceed u32 in any case (even when considering the absolute offset), so I'll just change it to use u32 everywhere.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: True!

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Can be removed now since we store it as u32.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: We could do that, but now that we don't store the paint in every wide tile command but just once per path I don't think it'll have a huge benefit. I'll add it as a TODO comment.

LaurenzV: It's even possible that we can get rid of indexed paints as a whole and in the render context only store the currently set paint instead of a list of paints.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Any suggestions for that? I think the impact will be negligible since, as before, we aren't likely pushing a lot of elements into it, but if you have a preference for the starting capacity I can add that. I don't think it would have a performance impact because the benchmarks (or at least the Blend2D benchmarks) runs with a "warm" render context, i.e. a render context that has been previously rested.

sparse_strips/vello_common/src/coarse.rs
grebmeg: No, it’s not a problem, I just think paying for something you don’t use isn’t ideal, even if the cost is negligible and the code is slightly simpler.

sparse_strips/vello_common/src/coarse.rs
grebmeg: No worries then, choosing the right capacity size isn’t easy. Regarding the warm render context, are you saying the render context isn’t recreated from scratch, but instead reset after each render?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -71,6 +71,8 @@ pub struct Wide<const MODE: u8 = MODE_CPU> {
     pub height: u16,
     /// The wide tiles in the container.
     pub tiles: Vec<WideTile<MODE>>,
+    /// Shared command properties, referenced by index from fill and clip commands.
+    pub props: CommandAttrs,
```
grebmeg: [nit]: when I suggested using `attrs` instead of `props`, I also meant applying it in other places where props is currently used, including when it’s used as a suffix.

sparse_strips/vello_hybrid/src/schedule.rs
grebmeg: Yeah, right, but it might still make sense to add an additional `col_idx` method on `ClipProps` and `FillProps`.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1586,48 +1596,103 @@ impl Cmd {
...
+    }
+}
+
+/// Shared properties for clip alpha fill commands.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClipProps {
+    /// The index of the thread that owns the alpha buffer
+    /// containing the mask values at `alpha_idx`.
+    /// Always 0 in single-threaded mode.
+    pub thread_idx: u8,
```
grebmeg, line 1632: I guess we can consider this negligible in `ClipProps` as well? And for `FillProps`, not having `thread_idx` wouldn’t actually change the struct size, right?

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Sorry, my oversight!

LaurenzV:
> No worries then, choosing the right capacity size isn’t easy. Regarding the warm render context, are you saying the render context isn’t recreated from scratch, but instead reset after each render?

That's how it's currently done in the Blend2D benchmark suite, yep!

sparse_strips/vello_common/src/coarse.rs
LaurenzV: I think so, yes. As mentioned though, I can change it back in case you consider it important. 😄

LaurenzV:
@grebmeg Thanks a lot!! I'll try to take a look at your other PRs soon.

PR #1327, opened 2025-12-17
perf: eliminate overdraw for opaque image fills

grebmeg:
This change mirrors what we [do for solid colors](https://github.com/linebender/vello/blob/4ea47b47c3e04a68b20422c9a12b2b98c735f547/sparse_strips/vello_common/src/coarse.rs#L1201-L1222), where we clear commands if a solid color covers the entire wide tile. We now apply the same approach to fully opaque images. 

The benchmark scene below shows roughly a _30%_ performance improvement for this case, although the exact gain depends on how many images overlap across full wide tiles. 

bench_output

- Low quality:
image

- Medium quality:
image

I’ll also open a follow-up PR to update [this](https://github.com/linebender/vello/blob/4ea47b47c3e04a68b20422c9a12b2b98c735f547/sparse_strips/vello_common/src/encode.rs#L506) where `has_opacities` currently returns `true` for all images.

### DIFF
--- a/sparse_strips/vello_bench/src/integration.rs
+++ b/sparse_strips/vello_bench/src/integration.rs
@@ -1,7 +1,7 @@
 // Copyright 2025 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! Full scene rendering benchmarks.
+//! Integration benchmarks for full rendering pipelines.
 
 use std::sync::Arc;
 
@@ -12,6 +12,7 @@ use vello_common::peniko::ImageSampler;
 use vello_common::peniko::{Extend, ImageQuality};
 use vello_common::pixmap::Pixmap;
 use vello_cpu::RenderContext;
+use vello_cpu::color::AlphaColor;
 
 /// Image scene rendering benchmark.
 pub fn images(c: &mut Criterion) {
@@ -41,7 +42,6 @@ pub fn images(c: &mut Criterion) {
                 let scale = width / original_width;
                 let height = original_height * scale;
 
-                renderer.set_transform(Affine::IDENTITY);
                 renderer.set_paint_transform(Affine::scale(scale));
                 renderer.set_paint(Image {
                     image: flower_image.clone(),
@@ -79,14 +79,9 @@ fn load_flower_image() -> ImageSource {
         rgba_data
             .chunks_exact(4)
             .map(|rgba| {
-                let alpha = u16::from(rgba[3]);
-                let premultiply = |component| (alpha * u16::from(component) / 255) as u8;
-                vello_common::color::PremulRgba8 {
-                    r: premultiply(rgba[0]),
-                    g: premultiply(rgba[1]),
-                    b: premultiply(rgba[2]),
-                    a: alpha as u8,
-                }
+                AlphaColor::from_rgba8(rgba[0], rgba[1], rgba[2], rgba[3])
+                    .premultiply()
+                    .to_rgba8()
             })
             .collect(),
         width as u16,
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -63,21 +63,6 @@ pub const MODE_CPU: u8 = 0;
 /// generation specific for `vello_hybrid`.
 pub const MODE_HYBRID: u8 = 1;
 
-/// Optimization hint for fill operations, computed in `Wide::generate` and passed to `WideTile::fill`.
-///
-/// This enum communicates whether a fill operation can benefit from overdraw elimination:
-/// - For opaque solid colors: we can set the background color directly and skip the fill
-/// - For opaque images: we can clear previous commands but still need to emit the fill
-#[derive(Debug, Clone, Copy)]
-pub enum FillOptimization {
-    /// No optimization possible, emit fill command normally.
-    None,
-    /// Paint is an opaque solid color - can replace background if conditions are met.
-    OpaqueSolid(PremulColor),
-    /// Paint is an opaque image - can clear previous commands if conditions are met.
-    OpaqueImage,
-}
-
 /// A container for wide tiles.
 #[derive(Debug)]
 pub struct Wide<const MODE: u8 = MODE_CPU> {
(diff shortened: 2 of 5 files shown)


nicoburns:
Sounds like a big win for large images (which I am very happy to see, because this is a case (the main case?) where Vello renderers seem to me to be noticably slower than alternatives).

grebmeg:
> Sounds like a big win for large images 

Yeah, I think it’s a really nice win, though I should point out that it mainly applies to overlapping images.

> (which I am very happy to see, because this is a case (the main case?) where Vello renderers seem to me to be noticably slower than alternatives).

Oh, I thought we didn’t have image benchmarks in t[he Blend2 performance suite](https://laurenzv.github.io/vello_chart/), but after reading the options documentation more carefully, I realized that `Pattern_NN` and `Pattern_BI` correspond to nearest-neighbor filtering and bilinear interpolation, respectively. And yes, for large images, the simple `FillRectA` doesn’t look great compared to the others. Blend2 looks insanely fast, we need this!

image

LaurenzV:
In the Blend2D suite, it's using transparent images so it probably wouldn't make a difference there. But yes, I there probably is more that can be done to optimize the performance of images. https://github.com/linebender/fearless_simd/pull/171 might also help there.

LaurenzV:
Oh and another thing, in that particular case the images only have 1x scaling and are pixel-aligned and I believe Blend2D has a special case for that, that's why we are much slower there. If you take a look at FillRectU and FillRectRot for example, the story looks different already.

nicoburns:
> Oh and another thing, in that particular case the images only have 1x scaling and are pixel-aligned and I believe Blend2D has a special case for that, that's why we are much slower there. If you take a look at FillRectU and FillRectRot for example, the story looks different already.

I think I've also seen discussed that a lot of renderers generate mipmaps for images, which would presumably greatly help performance in cases with large downscaling factors. https://servo.org for example features several images with native dimensions of in some cases over 4000px, some of which are rendered downscaled by a factor of ~10x.

LaurenzV:
I'm not sure this would help with performance though? I actually think it would be more costly, because you need to compute the mipmaps, which you don't need currently. When rendering we only sample the affected pixels, so the size of the image doesn't really make a difference here, I think.

nicoburns:
> When rendering we only sample the affected pixels

Hmm... I had assumed that we would be doing what https://en.wikipedia.org/wiki/Image_scaling describes as "box sampling" for when downscaling with a scale factor > 2x. But perhaps we're currently just dropping pixels?

LaurenzV:
If you are using NN, then yes, the pixels will just be dropped right now. For bilinear/bicubic we do still sample neighboring pixels, but that isn't enough if the downscale factor is larger.

sparse_strips/vello_bench/src/integration.rs
LaurenzV, line 1: I'm wondering whether integration (similar as in integration tests) would be a better name? But not sure, up to you.

sparse_strips/vello_bench/src/scene.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        b.iter(|| {
+            renderer.reset();
+
+            for i in (1..=image_count).rev() {
+                let width = 256.0 * i as f64;
+                let scale = width / original_width;
+                let height = original_height * scale;
+
+                renderer.set_transform(Affine::IDENTITY);
```
LaurenzV: (I think this could be omitted since we don't ever change the transform, right?)

sparse_strips/vello_bench/src/scene.rs
```diff
@@ -0,0 +1,97 @@
...
+            .chunks_exact(4)
+            .map(|rgba| {
+                let alpha = u16::from(rgba[3]);
+                let premultiply = |component| (alpha * u16::from(component) / 255) as u8;
+                vello_common::color::PremulRgba8 {
+                    r: premultiply(rgba[0]),
+                    g: premultiply(rgba[1]),
+                    b: premultiply(rgba[2]),
+                    a: alpha as u8,
+                }
```
LaurenzV:
```suggestion
                AlphaColor::from_rgba8(rgba[0], rgba[1], rgba[2], rgba[3])
                    .premultiply()
                    .to_rgba8()
```

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1194,50 +1197,73 @@ impl<const MODE: u8> WideTile<MODE> {
...
+                                }
+                                return;
+                            }
+                            Paint::Indexed(idx) => {
+                                // TODO: Add optimization for gradients.
+                                // Check if the indexed paint is an opaque image.
+                                if let Some(EncodedPaint::Image(img)) =
+                                    encoded_paints.get(idx.index())
+                                    && !img.has_opacities
+                                    && img.sampler.alpha == 1.0
```
LaurenzV: I think once we implement `has_opacities` we should always set it to true in case `img.sampler.alpha != 1`, so that we don't need the second check `img.sampler.alpha == 1`

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1194,50 +1197,73 @@ impl<const MODE: u8> WideTile<MODE> {
...
+                                {
+                                    // Opaque image: clear previous commands but still emit the fill.
+                                    self.cmds.clear();
+                                    self.bg = PremulColor::from_alpha_color(TRANSPARENT);
+                                    if let Some(ranges) =
+                                        self.layer_cmd_ranges.get_mut(&current_layer_id)
+                                    {
+                                        ranges.clear();
+                                    }
+                                    // Fall through to emit the fill command below.
```
LaurenzV: Can you maybe add to the comment something like "as opposed to solid paints, where we have a return statement"?

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -270,7 +270,8 @@ impl MultiThreadedDispatcher {
             allocation_group,
         };
         task_sender.send(task).unwrap();
-        self.run_coarse(true);
+        // TODO: Support encoded_paints in multithreading.
```
LaurenzV: Can you clarify this comment? This means we don't support clearing commands with indexed paints, but indexed paints themselves still work, right?

LaurenzV (approved):

sparse_strips/vello_common/src/coarse.rs
grebmeg: Good point! Since #1329 implements tracking `has_opacities`, I’ll update it there. Thank you!

sparse_strips/vello_common/src/coarse.rs
grebmeg: Done.

sparse_strips/vello_bench/src/integration.rs
grebmeg: I think you made a good point, “integration” is a more descriptive name here. Thank you!

Fixed.

sparse_strips/vello_bench/src/scene.rs
grebmeg: Yep, right.

Fixed.

sparse_strips/vello_bench/src/scene.rs
grebmeg: Thanks!

Fixed.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
grebmeg: Correct. Indexed paints should render fine, the missing part is the overdraw elimination optimization for opaque indexed paints, since opacity can’t be determined without `encoded_paints`.

Todo comment updated.

PR #1328, opened 2025-12-18
tests(sparse_strips): add JSON diff reports for sparse test snapshot failures

grebmeg:
I find this report very useful! It looks like this:

```json
{
  "pixel_count": 20,
  "max_difference": [
    1,
    3,
    1,
    0
  ],
  "pixels": [
    {
      "x": 14,
      "y": 10,
      "reference": [
        94,
        139,
        198,
        255
      ],
      "actual": [
        95,
        142,
        199,
        255
      ],
      "difference": [
        1,
        3,
        1,
        0
      ]
    },
  ]
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,5 +154,5 @@ console_log = "1.0"
 proc-macro2 = "1.0.95"
 syn = { version = "2.0.101", features = ["full", "extra-traits"] }
 quote = "1.0.40"
-serde = { version = "1.0", default-features = false }
-serde_json = "1.0"
+serde = { version = "1.0.225", default-features = false }
+serde_json = "1.0.145"
--- a/sparse_strips/vello_sparse_tests/tests/util.rs
+++ b/sparse_strips/vello_sparse_tests/tests/util.rs
@@ -381,7 +381,11 @@ pub(crate) fn check_ref(
         let json_data = serde_json::to_string_pretty(&report).unwrap();
         std::fs::write(&json_path, json_data).unwrap();
 
-        panic!("test didnt match reference image");
+        panic!(
+            "test didn't match reference image\n  diff image: {}\n  diff report: {}",
+            diff_path.display(),
+            json_path.display()
+        );
     }
 }
 


Cargo.toml
```diff
@@ -154,3 +154,5 @@ console_log = "1.0"
 proc-macro2 = "1.0.95"
 syn = { version = "2.0.101", features = ["full", "extra-traits"] }
 quote = "1.0.40"
+serde = { version = "1.0", default-features = false }
+serde_json = "1.0"
```
DJMcNab: These version numbers should be fully specified. This is a blocker.

sparse_strips/vello_sparse_tests/tests/util.rs
```diff
@@ -338,6 +365,22 @@ pub(crate) fn check_ref(
...
+        });
+        let report = DiffReport {
+            pixel_count: diff_data.len(),
+            max_difference,
+            pixels: diff_data,
+        };
+        let json_data = serde_json::to_string_pretty(&report).unwrap();
+        std::fs::write(&json_path, json_data).unwrap();
+
         panic!("test didnt match reference image");
```
DJMcNab: Can we provide the path to the diff report in this error message? This is not a blocker.

I keep meaning to come back through and massively improve the UX of these, but that's not for this PR...

sparse_strips/vello_sparse_tests/tests/util.rs
```diff
@@ -21,6 +21,33 @@ use vello_cpu::{Level, RenderMode};
...
+#[derive(Debug, serde::Serialize)]
+pub(crate) struct PixelDiff {
+    /// The x coordinate of the differing pixel.
+    pub x: u32,
+    /// The y coordinate of the differing pixel.
+    pub y: u32,
+    /// The RGBA values from the reference image.
+    pub reference: [u8; 4],
+    /// The RGBA values from the actual image.
+    pub actual: [u8; 4],
```
DJMcNab, line 46: Have you considered serialising this as either `#rrggbbaa` or `rgba(rrr, ggg, bbb, aaa)` (i.e. as a string); I believe if we do that, vscode will show a colour picker as a preview. Plus, the formatting of the lines will be much shorter.

I think we can either do that with a [custom serialise function](https://serde.rs/field-attrs.html#serialize_with) (which would use https://docs.rs/serde/latest/serde/trait.Serializer.html#method.collect_str), or by storing the field as a String (which of course would be much more allocation intensive, and so wouldn't be my preferred solution).

DJMcNab (approved):
This looks interesting; having exactly what RGBA colour things worked out to is good. I think it could land as-is, but I'd like to see a few things changed. If you do change the formatting, please re-request my review.

One small question; I notice that your example data there should be 20 entries, but it only has one. Am I correct that the rest were elided for space?

Cargo.toml
grebmeg: Thanks!

Fixed.

sparse_strips/vello_sparse_tests/tests/util.rs
grebmeg: That’s really good advice, thank you! 

> I keep meaning to come back through and massively improve the UX of these, but that's not for this PR...

Honestly, I’ve already started on this and even have a branch with changes to the testing framework, including some API updates. I just haven’t had enough time to get it ready for review yet. If you’re keen, I’d be happy to share it.

Fixed.

sparse_strips/vello_sparse_tests/tests/util.rs
grebmeg: Good idea! I had been thinking about custom serialization from a slightly different angle, mainly focusing on putting the color data on single lines because it felt easier to read and compare. I hadn’t considered using `#rrggbbaa`, but that could definitely be worth adding alongside the current format. When I tried implementing it, things got a bit heavy, but I’m happy for this to be explored further, either by me or someone else as a follow-up later.

sparse_strips/vello_sparse_tests/tests/util.rs
DJMcNab: Making the UX of this kind of testing much less opaque is really helpful for contributors. I put a lot of care into that in Vello Classic's tests, which unfortunately didn't get carried into the sparse strips work.

When you have a PR, I can try and prioritise it.

PR #1329, opened 2025-12-18
perf(vello_common): track has_opacities to skip alpha blending

grebmeg:
By tracking image opacities, we can skip alpha blending for fully opaque image fills and directly override `blend_buf` (code [here](https://github.com/linebender/vello/blob/4ea47b47c3e04a68b20422c9a12b2b98c735f547/sparse_strips/vello_cpu/src/fine/mod.rs#L717)), resulting in a notable performance gain over the main branch (the change here adds about _~24%_ improvement):

- Low quality:
image

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/lowp/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/lowp/image.rs
@@ -5,10 +5,10 @@ use crate::fine::PosExt;
 use crate::fine::common::image::{ImagePainterData, extend, sample};
 use crate::fine::macros::u8x16_painter;
 use vello_common::encode::EncodedImage;
-use vello_common::fearless_simd::{Simd, SimdBase, f32x4, u8x16, u16x16};
+use vello_common::fearless_simd::{Simd, SimdBase, SimdFloat, f32x4, u8x16, u16x16};
 use vello_common::pixmap::Pixmap;
 use vello_common::simd::element_wise_splat;
-use vello_common::util::f32_to_u8;
+use vello_common::util::{Div255Ext, f32_to_u8};
 
 /// A faster bilinear image renderer for the u8 pipeline.
 #[derive(Debug)]
@@ -78,17 +78,17 @@ impl<S: Simd> Iterator for BilinearImagePainter<'_, S> {
 
         let fx = f32_to_u8(element_wise_splat(
             self.simd,
-            fract(x_positions + 0.5) * 256.0,
+            fract(x_positions + 0.5).madd(255.0, 0.5),
         ));
         let fy = f32_to_u8(element_wise_splat(
             self.simd,
-            fract(y_positions + 0.5) * 256.0,
+            fract(y_positions + 0.5).madd(255.0, 0.5),
         ));
-        // Widen to u16, then compute `256 - fx` to ensure fx + fx_inv = 256.
+
         let fx = self.simd.widen_u8x16(fx);
         let fy = self.simd.widen_u8x16(fy);
-        let fx_inv = u16x16::splat(self.simd, 256) - fx;
-        let fy_inv = u16x16::splat(self.simd, 256) - fy;
+        let fx_inv = u16x16::splat(self.simd, 255) - fx;
+        let fy_inv = u16x16::splat(self.simd, 255) - fy;
 
         let x_pos1 = extend_x(x_positions - 0.5);
         let x_pos2 = extend_x(x_positions + 0.5);
@@ -108,15 +108,9 @@ impl<S: Simd> Iterator for BilinearImagePainter<'_, S> {
             .simd
             .widen_u8x16(sample(self.simd, &self.data, x_pos2, y_pos2));
 
-        // Add rounding bias before shifting: round(x/256) = floor((x + 128) / 256).
-        // Without this, `>> 8` truncates toward zero, causing off-by-one errors.
-        // E.g., 254.996 would become 254 instead of rounding to 255.
-        let round = u16x16::splat(self.simd, 128);
-        let ip1 = (p00 * fx_inv + p10 * fx + round) >> 8;
-        let ip2 = (p01 * fx_inv + p11 * fx + round) >> 8;
-        let res = self
-            .simd
-            .narrow_u16x16((ip1 * fy_inv + ip2 * fy + round) >> 8);
+        let ip1 = (p00 * fx_inv + p10 * fx).div_255();
+        let ip2 = (p01 * fx_inv + p11 * fx).div_255();
+        let res = self.simd.narrow_u16x16((ip1 * fy_inv + ip2 * fy).div_255());
 
         self.data.cur_pos += self.data.image.x_advance;
 


sparse_strips/vello_bench/src/integration.rs
```diff
@@ -17,6 +18,7 @@ use vello_cpu::color::AlphaColor;
 /// Image scene rendering benchmark.
 pub fn images(c: &mut Criterion) {
     let mut g = c.benchmark_group("images");
+    g.measurement_time(Duration::from_secs(10));
```
LaurenzV: Isn't 10 seconds a bit long?

sparse_strips/vello_bench/src/integration.rs
```diff
@@ -71,21 +73,27 @@ fn load_flower_image() -> ImageSource {
     let height = image.height();
     let rgba_data = image.into_rgba8().into_vec();
 
+    let mut has_opacities = false;
     #[expect(
         clippy::cast_possible_truncation,
-        reason = "Image dimensions fit in u16"
+        reason = "u16 * u8 / 255 fits in u8, image dimensions fit in u16"
```
LaurenzV: Is this comment still up-to-date? I believe the old one should be correct again now.

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -21,13 +21,25 @@ pub struct Pixmap {
     height: u16,
     /// Buffer of the pixmap in RGBA8 format.
     buf: Vec<PremulRgba8>,
+    /// Whether the pixmap has any non-opaque pixels.
```
LaurenzV:
```suggestion
    /// Whether the pixmap may have non-opaque pixels.
```

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -90,6 +130,36 @@ impl Pixmap {
...
+    /// Returns whether the pixmap may have non-opaque pixels.
+    ///
+    /// This value is computed at construction time. It may become stale if pixels are
+    /// modified directly via [`data_mut()`](Self::data_mut),
+    /// [`data_as_u8_slice_mut()`](Self::data_as_u8_slice_mut), or [`set_pixel()`](Self::set_pixel).
+    ///
+    /// Use [`set_has_opacities()`](Self::set_has_opacities) to manually update the flag,
+    /// or [`recompute_has_opacities()`](Self::recompute_has_opacities) to recalculate it
+    /// by scanning all pixels.
+    pub fn has_opacities(&self) -> bool {
```
LaurenzV: Should we maybe rename to `may_have_opacities`?

LaurenzV: (If we do, we also should do it for the other methods.)

sparse_strips/vello_cpu/src/fine/lowp/image.rs
LaurenzV, line 1: If I'm understanding correctly, those changes are not directly related to the optimization, right? If so, could we maybe split them out into a separate PR?

sparse_strips/vello_bench/src/integration.rs
grebmeg: I’ll remove it for now, which will fall back to the default value of `5s`. Previously, I had set it to `10s`, as advised by criterion to increase it, since they internally calculate the preferred measurement time.

sparse_strips/vello_bench/src/integration.rs
grebmeg: Yep, I agree.

Fixed.

sparse_strips/vello_common/src/pixmap.rs
grebmeg: Thanks!

Fixed.

sparse_strips/vello_common/src/pixmap.rs
grebmeg: I agree with the renaming, since that flag can become stale, so I think the `may_` prefix better communicates this uncertainty.

Renamed.

sparse_strips/vello_cpu/src/fine/lowp/image.rs
grebmeg: I left this fix here because when the `may_have_opacities` optimisation is enabled for opaque images, that flag becomes `false` and fine rasterisation goes through a different branch in [`fill_complex_paint`](https://github.com/linebender/vello/blob/895c1cb4866d2a3c19441659196ff7623dc2966f/sparse_strips/vello_cpu/src/fine/mod.rs#L700). As a result, we end up with slightly different values during rasterisation.

I could move this fix into the first PR, but then the tests would effectively not exercise this path, because `have_opacities` would always be `true`. I thought it would be better to keep the fix here to demonstrate that enabling the optimisation takes a different path that previously produced incorrect values. If you think it would still be better to move this into a separate PR, let me know.

sparse_strips/vello_cpu/src/fine/lowp/image.rs
```diff
@@ -84,11 +84,11 @@ impl<S: Simd> Iterator for BilinearImagePainter<'_, S> {
             self.simd,
             fract(y_positions + 0.5) * 256.0,
         ));
-        let fx_inv = self.simd.widen_u8x16(u8x16::splat(self.simd, 255) - fx);
-        let fy_inv = self.simd.widen_u8x16(u8x16::splat(self.simd, 255) - fy);
-
+        // Widen to u16, then compute `256 - fx` to ensure fx + fx_inv = 256.
         let fx = self.simd.widen_u8x16(fx);
         let fy = self.simd.widen_u8x16(fy);
+        let fx_inv = u16x16::splat(self.simd, 256) - fx;
```
LaurenzV: Are you sure it's correct to use 256 here? I believe the previous 255 was correct. 

Just to explain, in the expression `fract(y_positions + 0.5) * 256.0`, the expression `fract(y_positions + 0.5)` will yield a value in the range [0.0, 1.0) (note that 1.0 is exclusive), and therefore it will be mapped to the range [0.0, 256.0), so if we end up with something like 255.5, it will be clamped to 255 after converting to u8. The inverse should then be 0, because the overall value is supposed to be at most 255 (the maximum value of u8), not 256. It's been a while since I wrote this code, but I think that should be right, or am I missing something?

sparse_strips/vello_cpu/src/fine/lowp/image.rs
```diff
@@ -108,9 +108,15 @@ impl<S: Simd> Iterator for BilinearImagePainter<'_, S> {
             .simd
             .widen_u8x16(sample(self.simd, &self.data, x_pos2, y_pos2));
 
-        let ip1 = (p00 * fx_inv + p10 * fx) >> 8;
-        let ip2 = (p01 * fx_inv + p11 * fx) >> 8;
-        let res = self.simd.narrow_u16x16((ip1 * fy_inv + ip2 * fy) >> 8);
+        // Add rounding bias before shifting: round(x/256) = floor((x + 128) / 256).
```
LaurenzV: When I wrote this code I was aware that `>> 8` is technically not correct, because we really want to divide by 255 and not 256, but I didn't consider it as critical since there is going to be a lot of imprecision anyway from the various calculations that are performed, and doing the shift is much faster.

So, if we really want to fix this, I think the correct approach would be to use the `div_255` method instead of bit-shifting. However, it is probably worth checking how much this slows down the benchmarks. What do you think?

sparse_strips/vello_cpu/src/fine/lowp/image.rs
LaurenzV: Let's say for example that we are sampling the exact center of the pixel (0.5, 0.5). In this case, `fx` ends up being 0, and using your code `fx_inv` would be 256. If we are now sampling a white pixel, we would then be calculating `256 * 255`, which would overflow the u16. So I think it should be right that `fx + fx_inv = 255`. 🤔

LaurenzV:
I think this is how it should look like, what do you think?

```rs
let fx = f32_to_u8(element_wise_splat(
    self.simd,
    fract(x_positions + 0.5).madd(255.0, 0.5),
));
let fy = f32_to_u8(element_wise_splat(
    self.simd,
    fract(y_positions + 0.5).madd(255.0, 0.5),
));

let fx = self.simd.widen_u8x16(fx);
let fy = self.simd.widen_u8x16(fy);
let fx_inv = u16x16::splat(self.simd, 255) - fx;
let fy_inv = u16x16::splat(self.simd, 255) - fy;

let x_pos1 = extend_x(x_positions - 0.5);
let x_pos2 = extend_x(x_positions + 0.5);
let y_pos1 = extend_y(y_positions - 0.5);
let y_pos2 = extend_y(y_positions + 0.5);

let p00 = self
    .simd
    .widen_u8x16(sample(self.simd, &self.data, x_pos1, y_pos1));
let p10 = self
    .simd
    .widen_u8x16(sample(self.simd, &self.data, x_pos2, y_pos1));
let p01 = self
    .simd
    .widen_u8x16(sample(self.simd, &self.data, x_pos1, y_pos2));
let p11 = self
    .simd
    .widen_u8x16(sample(self.simd, &self.data, x_pos2, y_pos2));

let ip1 = (p00 * fx_inv + p10 * fx).div_255();
let ip2 = (p01 * fx_inv + p11 * fx).div_255();
let res = self
    .simd
    .narrow_u16x16((ip1 * fy_inv + ip2 * fy).div_255());
```

I tried it locally and the tests still seem to pass.

sparse_strips/vello_cpu/src/fine/lowp/image.rs
LaurenzV: No worries, fine from my side!

sparse_strips/vello_cpu/src/fine/lowp/image.rs
grebmeg: I used `256` because it allows a simple `>> 8` shift for division, avoiding the need for `div_255()`.

There’s no overflow risk, the maximum intermediate value is `256 × 255 + 128 = 65,408`, which fits within `u16`. The `+128` rounding bias compensates for some precision loss, since `round(x / 256) = floor((x + 128) / 256)`.

You’re right that this introduces a small asymmetry at the edges, when `fract() → 1.0`, `fx` caps at `255` due to `u8` clamping, but this should be imperceptible in practice.

I benchmarked your snippet, it behaves correctly and passes the tests, but it introduces a __~2.5%__ regression. Given that, what’s your opinion on whether a more precise but slightly slower approach is preferable, or if the faster, less precise version is acceptable here?

LaurenzV: Ah true, I got the overflow part wrong, my bad!

LaurenzV (approved):
Will leave it up to you if you want to change back to 256, but I think 2.5% is small enough that it's better to just use the more accurate version!

PR #1330, opened 2025-12-21
Remove `wgpu` feature. Add `wgpu_27` and `wgpu_28` features.

nicoburns:
# Objective

Allow the same version of Vello to compile with multiple versions of WGPU (current v27 and v28):

- This should make it easier to upgrade to new versions of WGPU promptly without worrying about leaving users who may still need an older WGPU version behind.
- In many cases, it may also make upgrading WGPU a non-breaking change.

## Prior Art

The `i-slint-core` crate has similar feature flags for WGPU versions (https://docs.rs/crate/i-slint-core/latest/features)

## Changes made

- Replace the `wgpu` and `naga` feature flags in the `vello`, `vello_hybrid`, `vello_shaders` and `vello_sparse_shaders` crates with `wgpu_27`, `wgpu_28` (and `naga_27`, `naga_28`) feature flags.
- Re-export e.g. `wgpu_28 as wgpu` in the crate roots
- The `wgpu-profiler` and `hot-reload` debug features are currently configured to only work with the latest WGPU version.

## Todo

- [ ] Use crates.io version of `wgpu-profiler` once https://github.com/Wumpf/wgpu-profiler/pull/103 lands

### DIFF
--- a/sparse_strips/vello_sparse_shaders/src/types.rs
+++ b/sparse_strips/vello_sparse_shaders/src/types.rs
@@ -26,7 +26,8 @@ impl ReflectionMap {
     /// Create a new `ReflectionMap` given the [`naga`] compile info.
     pub(crate) fn new(info: ReflectionInfo, global_vars: &Arena<GlobalVariable>) -> Self {
         debug_assert_eq!(info.varying.len(), 0, "unimplemented");
-        debug_assert_eq!(info.push_constant_items.len(), 0, "unimplemented");
+        // debug_assert_eq!(info.push_constant_items.len(), 0, "unimplemented");
+        debug_assert_eq!(info.immediates_items.len(), 0, "unimplemented");
         let mut texture_mapping = BTreeMap::default();
         let mut uniforms = BTreeMap::default();
 
--- a/sparse_strips/vello_sparse_shaders/src/types.rs
+++ b/sparse_strips/vello_sparse_shaders/src/types.rs
@@ -8,6 +8,7 @@
     reason = "False positives as this module is used at build time."
 )]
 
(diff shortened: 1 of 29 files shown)


waywardmonkeys:
I'd like to see this happen and get a release out!

Do you want to update this or should someone else take it up?

cc: @xStrom @PoignardAzur

xStrom:
I'm not sure if it's worth actually supporting multiple wgpu versions. However, just as a technical note, the features should also work if both are selected. Probably the highest version should win then. So in this PR if both `wgpu_27` and `wgpu_28` are enabled, it should work the same way as if just `wgpu_28` was enabled. This keeps `--all-features` functional.

sparse_strips/vello_sparse_shaders/src/types.rs
```diff
@@ -26,7 +27,8 @@ impl ReflectionMap {
     /// Create a new `ReflectionMap` given the [`naga`] compile info.
     pub(crate) fn new(info: ReflectionInfo, global_vars: &Arena<GlobalVariable>) -> Self {
         debug_assert_eq!(info.varying.len(), 0, "unimplemented");
-        debug_assert_eq!(info.push_constant_items.len(), 0, "unimplemented");
+        // debug_assert_eq!(info.push_constant_items.len(), 0, "unimplemented");
+        debug_assert_eq!(info.immediates_items.len(), 0, "unimplemented");
```
PoignardAzur, line 31: Leftover comment?

PoignardAzur (commented):
I don't have a strong opinion on this, but it seems like a lot of churn.

PR #1331, opened 2025-12-22
Serialize JSON report references as HTML color codes

DJMcNab:
Follow-up from https://github.com/linebender/vello/pull/1328

The new formatting looks like:
```json
...
{
  "x": 1,
  "y": 18,
  "reference": "#ece5f2",
  "actual": "#bfffbf",
  "difference": [
    -45,
    26,
    -51,
    0
  ]
},
{
  "x": 2,
  "y": 1,
  "reference": "#d9cce5",
  "actual": "#7fff7f",
  "difference": [
    -90,
    51,
    -102,
    0
  ]
},
...
```

I've also tweaked the in-repo settings of vscode to show the colour picker in these diff files:
image

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/util.rs
+++ b/sparse_strips/vello_sparse_tests/tests/util.rs
@@ -41,13 +41,15 @@ pub(crate) struct PixelDiff {
     pub x: u32,
     /// The y coordinate of the differing pixel.
     pub y: u32,
-    /// The RGBA values from the reference image.
+    /// The RGBA values from the target image (i.e. the saved reference).
+    // Note that this field name is chosen to be the same length as `actual`
+    // That makes it easier to compare the results in the printed JSON.
     #[serde(serialize_with = "hex_string")]
-    pub reference: [u8; 4],
+    pub target: [u8; 4],
     /// The RGBA values from the actual image.
     #[serde(serialize_with = "hex_string")]
     pub actual: [u8; 4],
-    /// Per-channel difference (actual - reference) as signed values.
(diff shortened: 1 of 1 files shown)


DJMcNab:
On keeping the full decimal values around, personally I'm not convinced that adds any value. The motivation for this PR was to remove the verbosity that keeping the numbers fully required. I also think that an LLM is likely to be more "familiar" with hex formatted colours than lists of `u8` components. Unless you have evidence that it's needed, I'd much prefer to leave it out.

On matching the lengths, that's a very good idea. It's important to keep the old and the new distinct/make it clear which is which, so I'll workshop the names tomorrow. `reference` and `observed`, which spring to mind, are a bit too "cute". Maybe `target` and `actual`?

grebmeg:
> On keeping the full decimal values around, personally I'm not convinced that adds any value. The motivation for this PR was to remove the verbosity that keeping the numbers fully required. I also think that an LLM is likely to be more "familiar" with hex formatted colours than lists of u8 components. Unless you have evidence that it's needed, I'd much prefer to leave it out.

Yeah, I don’t have it, so let’s leave it as it is. Thank you!

> On matching the lengths, that's a very good idea. It's important to keep the old and the new distinct/make it clear which is which, so I'll workshop the names tomorrow. `reference` and `observed`, which spring to mind, are a bit too "cute". Maybe `target` and `actual`?

I like `target`/`actual` and agree that clearer naming here would help. Some other options I like: `gold`/`test`, `source`/`actual`, and `expect`/`actual`.

PR #1332, opened 2025-12-23
vello: make Scene clip/layers honor fill rule (even-odd clips)

waywardmonkeys:
Scene layer APIs now take a `peniko::StyleRef` for the clip shape (`push_layer`, `push_clip_layer`, `push_luminance_mask_layer`), fixing the long-standing hardcoded `Fill::NonZero` clip behavior.

This is required for correctness in CSS/SVG semantics (e.g. `clip-rule`, `fill-rule`, and masking/clipPath content that relies on even-odd holes).

Using `StyleRef` also enables clipping to a stroke (pass `&Stroke`), so clips can be defined by a stroked outline as well as a filled interior.

To make even-odd clipping work end-to-end, plumb the fill rule through the clip pipeline:

- Give `BEGIN_CLIP` a 1-word info slot so clip draw flags can be stored.
- Populate clip `draw_flags` for `BEGIN_CLIP` in `draw_leaf`.
- In `clip_leaf`, propagate `info_offset` from `BEGIN_CLIP` to the corresponding `END_CLIP`.
- In coarse, apply clip layers with the correct fill rule by propagating `BEGIN_CLIP` draw flags through to `END_CLIP`, and by updating the “clip-zero” fast path to respect even-odd backdrop parity (GPU + CPU paths).

Existing callsites can continue passing `Fill::NonZero` to preserve prior behavior; `Fill::EvenOdd` selects even-odd clips.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This release has an [MSRV][] of 1.88.
 ### Changed
 
 - Breaking change: wgpu has been updated to wgpu 27. ([#1280][] by [@theoparis][])  
+- Breaking change: Make `Scene` clip / layers honor fill rule (even-odd clips). ([#1332][] by [@waywardmonkeys][])
+  When pushing a layer, you should use `Fill::NonZero` as the clip fill rule to achieve the same behavior as previous versions.
 
 ### Fixed
 
@@ -383,6 +385,7 @@ This release has an [MSRV][] of 1.75.
 [#1273]: https://github.com/linebender/vello/pull/1273
 [#1280]: https://github.com/linebender/vello/pull/1280
 [#1323]: https://github.com/linebender/vello/pull/1323
+[#1332]: https://github.com/linebender/vello/pull/1332
 
 <!-- Note that this still comparing against 0.5.0, because 0.5.1 is a cherry-picked patch -->
 [Unreleased]: https://github.com/linebender/vello/compare/v0.5.0...HEAD
--- a/vello_shaders/shader/coarse.wgsl
+++ b/vello_shaders/shader/coarse.wgsl
@@ -410,7 +410,9 @@ fn main(
                         write_image(di + 1u);
                     }
                     case DRAWTAG_BEGIN_CLIP: {
-                        if tile.segment_count_or_ix == 0u && tile.backdrop == 0 {
+                        let even_odd = (draw_flags & DRAW_INFO_FLAGS_FILL_RULE_BIT) != 0u;
+                        let backdrop_clear = select(tile.backdrop, abs(tile.backdrop) & 1, even_odd) == 0;
+                        if tile.segment_count_or_ix == 0u && backdrop_clear {
                             clip_zero_depth = clip_depth + 1u;
                         } else {
                             write_begin_clip();
@@ -421,8 +423,7 @@ fn main(
                     }
                     case DRAWTAG_END_CLIP: {
                         clip_depth -= 1u;
-                        // A clip shape is always a non-zero fill (draw_flags=0).
-                        write_path(tile, tile_ix, /*draw_flags=*/0u);
+                        write_path(tile, tile_ix, draw_flags);
                         let blend = scene[dd];
                         let alpha = bitcast<f32>(scene[dd + 1u]);
                         write_end_clip(CmdEndClip(blend, alpha));
--- a/vello_shaders/shader/shared/drawtag.wgsl
+++ b/vello_shaders/shader/shared/drawtag.wgsl
@@ -23,7 +23,7 @@ const DRAWTAG_FILL_RAD_GRADIENT = 0x29cu;
 const DRAWTAG_FILL_SWEEP_GRADIENT = 0x254u;
 const DRAWTAG_FILL_IMAGE = 0x28Cu;
 const DRAWTAG_BLURRED_ROUNDED_RECT = 0x2d4u;
-const DRAWTAG_BEGIN_CLIP = 0x9u;
+const DRAWTAG_BEGIN_CLIP = 0x49u;
 const DRAWTAG_END_CLIP = 0x21u;
 
 /// The first word of each draw info stream entry contains the flags. This is not a part of the
--- a/vello_shaders/src/cpu/coarse.rs
+++ b/vello_shaders/src/cpu/coarse.rs
@@ -334,7 +334,13 @@ fn coarse_main(
                                 tile_state.write_blur_rect(config, bump, ptcl, rgba_color, di + 1);
                             }
                             DrawTag::BEGIN_CLIP => {
-                                if tile.segment_count_or_ix == 0 && tile.backdrop == 0 {
+                                let even_odd = (draw_flags & DRAW_INFO_FLAGS_FILL_RULE_BIT) != 0;
+                                let backdrop_clear = if even_odd {
+                                    tile.backdrop.abs() & 1 == 0
+                                } else {
+                                    tile.backdrop == 0
+                                };
+                                if tile.segment_count_or_ix == 0 && backdrop_clear {
                                     clip_zero_depth = clip_depth + 1;
                                 } else {
                                     tile_state.write_begin_clip(config, bump, ptcl);
@@ -347,8 +353,7 @@ fn coarse_main(
                             }
                             DrawTag::END_CLIP => {
                                 clip_depth -= 1;
-                                // A clip shape is always a non-zero fill (draw_flags=0).
-                                tile_state.write_path(config, bump, ptcl, tile, 0);
+                                tile_state.write_path(config, bump, ptcl, tile, draw_flags);
                                 let blend = scene[dd as usize];
                                 let alpha = f32::from_bits(scene[dd as usize + 1]);
                                 tile_state.write_end_clip(config, bump, ptcl, blend, alpha);
--- a/examples/scenes/src/test_scenes.rs
+++ b/examples/scenes/src/test_scenes.rs
@@ -769,7 +769,7 @@ mod impls {
             &rect,
         );
         let alpha = params.time.sin() as f32 * 0.5 + 0.5;
-        scene.push_layer(Mix::Normal, alpha, Affine::IDENTITY, &rect);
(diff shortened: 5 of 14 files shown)


waywardmonkeys:
This was done with LLM assistance as it apparently understands some of the Vello internals better than me.

vello_shaders/shader/shared/drawtag.wgsl
```diff
@@ -23,7 +23,7 @@ const DRAWTAG_FILL_RAD_GRADIENT = 0x29cu;
 const DRAWTAG_FILL_SWEEP_GRADIENT = 0x254u;
 const DRAWTAG_FILL_IMAGE = 0x28Cu;
 const DRAWTAG_BLURRED_ROUNDED_RECT = 0x2d4u;
-const DRAWTAG_BEGIN_CLIP = 0x9u;
+const DRAWTAG_BEGIN_CLIP = 0x49u;
```
tomcur, line 26: This looks right, getting an `info_offset` of `1`: `(0x40u >> 6u) && 0x0fu == 1u)`.

https://github.com/linebender/vello/blob/c15340d42933286e74f41d0ffd50fa5d51633924/vello_shaders/shader/shared/drawtag.wgsl#L47-L54

vello_shaders/shader/coarse.wgsl
```diff
@@ -375,6 +375,8 @@ fn main(
             if clip_zero_depth == 0u {
                 let tile_ix = sh_tile_base[el_ix] + sh_tile_stride[el_ix] * tile_y + tile_x;
                 let tile = tiles[tile_ix];
+                let even_odd = (draw_flags & DRAW_INFO_FLAGS_FILL_RULE_BIT) != 0u;
+                let backdrop_clear = select(tile.backdrop, abs(tile.backdrop) & 1, even_odd) == 0;
```
tomcur: The rule looks correct. Nit: I believe it's possible to move this into the `DRAWTAG_BEGIN_CLIP` branch (edit: seems you've already done that in `coarse.rs`).

tomcur (approved):
This looks right to me, but I've not looked much at Vello's shaders before, so it probably makes sense to have at least one more approval.

I was first confused at how this was able to work at all with the following `include_tile` logic, but have since convinced myself that that was a case of vestigial documentation of old logic (which I propose a fix for in https://github.com/linebender/vello/pull/1333). The actual checks there and what the non-zero-filled clipping does seem to match.

https://github.com/linebender/vello/blob/c15340d42933286e74f41d0ffd50fa5d51633924/vello_shaders/shader/coarse.wgsl#L335-L344

vello_shaders/src/cpu/coarse.rs
```diff
@@ -334,7 +334,13 @@ fn coarse_main(
...
                             }
                             DrawTag::BEGIN_CLIP => {
-                                if tile.segment_count_or_ix == 0 && tile.backdrop == 0 {
+                                let even_odd = (draw_flags & DRAW_INFO_FLAGS_FILL_RULE_BIT) != 0;
+                                let backdrop_clear = if even_odd {
+                                    tile.backdrop.abs() & 1 == 0
+                                } else {
+                                    tile.backdrop == 0
+                                };
+                                if tile.segment_count_or_ix == 0 && backdrop_clear {
```
DJMcNab, line 343: As Tom points out, it isn't clear to me why this code should be in a different place in the logic as compared to the GPU implementation. In general, I think it's probably most useful to keep the "shared" logic as similar as possible.

vello_tests/tests/property.rs
```diff
@@ -195,3 +195,59 @@ fn premultiplied_image() {
...
+    let mut scene = Scene::new();
+    let viewport = Rect::new(0.0, 0.0, width as f64, height as f64);
+
+    scene.fill(
+        Fill::NonZero,
+        Affine::IDENTITY,
+        palette::css::BLUE,
+        None,
+        &viewport,
+    );
```
DJMcNab: It would probably be more idiomatic to just set the base color.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This release has an [MSRV][] of 1.88.
 ### Changed
 
 - Breaking change: wgpu has been updated to wgpu 27. ([#1280][] by [@theoparis][])  
+- Breaking change: make `Scene` clip / layers honor fill rule (even-odd clips). ([#1332][] by [@waywardmonkeys][])
```
DJMcNab:
```suggestion
- Breaking change: Make `Scene` clip / layers honor fill rule (even-odd clips). ([#1332][] by [@waywardmonkeys][])
```
The reason this doesn't apply to the above is that the wgpu "wordmark" is stylised as lowercase - as seen on https://wgpu.rs/.

DJMcNab: I'd also like to see a slightly expanded version of this. Even just the really simple migration guide of "use `add_layer(...)` -> `add_layer(Fill::NonZero, ...)` to match the old behaviour" would be helpful to people porting to the new versions.

vello_tests/tests/property.rs
DJMcNab, line 1: I'd also like to see a screenshot test for this as well; it's something that's easier to parse visually.

I believe that we can do so fairly simply by adding the `clip_test` test scene to https://github.com/linebender/vello/blob/main/vello_tests/tests/snapshot_test_scenes.rs (I don't have any recollection that there's a reason for it to not already be in there).

tomcur:
> Obviously, I'm not a massive fan of the extra API complexity for this fairly niche use case. But I don't see how to avoid that in an ergonomic way, unless we introduce helpers which ignore this, which is also pretty nasty.

I'm somewhat less worried about API complexity here, it's an important feature for correctness/completeness, but there's a good general question here about function parameter growth in these kinds of APIs versus, say, the `wgpu`-style of introducing something like `LayerDescriptor` structs.

One very nitty detail (which doesn't really solve the issue in your comment) is that it may be nice to name the parameter `clip_fill_rule` (as it only applies to the clip), and, while the current order of arguments agrees with e.g. `Scene::fill`, it may then be nice to order the `clip_fill_rule` parameter to be just in front of `clip`.

vello_shaders/src/cpu/coarse.rs
waywardmonkeys: There are so many differences between WGSL and Rust. :(

CHANGELOG.md
waywardmonkeys: Done.

waywardmonkeys: Done.

vello_shaders/src/cpu/coarse.rs
waywardmonkeys: Moved.

vello_shaders/shader/coarse.wgsl
waywardmonkeys: Moved.

waywardmonkeys:
> One very nitty detail (which doesn't really solve the issue in your comment) is that it may be nice to name the parameter `clip_fill_rule` (as it only applies to the clip), and, while the current order of arguments agrees with e.g. `Scene::fill`, it may then be nice to order the `clip_fill_rule` parameter to be just in front of `clip`.

I've renamed it. I did go with this order to match `Scene::fill` ... but I guess we have until we do a release to bikeshed the ordering. :)

vello_tests/tests/property.rs
waywardmonkeys: I removed this test in favor of the clip_test snapshot test.

vello_tests/tests/property.rs
waywardmonkeys: Done. (Hopefully it passes CI.)

tomcur (approved):
Looks good!

dfrg:
> I would also quite like one of Raph or Chad to give it a once over

The logic LGTM.

If we're adding a new parameter, I wonder if we should go all the way and take a [`peniko::StyleRef`](https://docs.rs/peniko/latest/peniko/enum.StyleRef.html) instead? This would allow clipping to a stroke as well.

waywardmonkeys:
> > I would also quite like one of Raph or Chad to give it a once over
> 
> The logic LGTM.
> 
> If we're adding a new parameter, I wonder if we should go all the way and take a [`peniko::StyleRef`](https://docs.rs/peniko/latest/peniko/enum.StyleRef.html) instead? This would allow clipping to a stroke as well.

At first, I didn't really understand how this would work ... but with some LLM assistance, I think I have it working.

This deserves a bit of review, especially the bit about handling an empty stroke.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ This release has an [MSRV][] of 1.88.
 ### Changed
 
 - Breaking change: wgpu has been updated to wgpu 27. ([#1280][] by [@theoparis][])  
+- Breaking change: Make `Scene` clip / layers honor fill rule (even-odd clips). ([#1332][] by [@waywardmonkeys][])
+  When pushing a layer, you should use `Fill::NonZero` as the clip fill rule to achieve the same behavior as previous versions.
```
tomcur, line 22: This should now be updated, something like the following.

```suggestion
- Breaking change: Allow setting `Scene` layer clipping style, adding even-odd filled path clipping and stroked path clipping. ([#1332][] by [@waywardmonkeys][])
  When pushing a layer, you should use `StyleRef::Fill(Fill::NonZero)` as the clip fill rule to achieve the same behavior as previous versions.
```

waywardmonkeys: It takes `Into<StyleRef>` so `Fill::NonZero` works and is more concise.

tomcur: Good point, didn't realize we had the `From`, but that makes sense.

DJMcNab (approved):
My only hesitance is that Tom's comment suggests that dashes are silently ignored. If we have a plan to immediately fix that, there's no reason to treat that as a blocker, though.

Otherwise, this seems good to land. The testing looks good.

tomcur (commented):
> My only hesitance is that Tom's comment suggests that dashes are silently ignored. If we have a plan to immediately fix that, there's no reason to treat that as a blocker, though.

I'm going to open a PR adding that once this PR lands.

PR #1335, opened 2025-12-29
perf(vello_cpu): optimize image rendering for axis-aligned images

grebmeg:
For bilinear sampling of axis-aligned images with no skew, this PR adds `PlainBilinearImagePainter`, which pre-computes constant values such as y positions and interpolation weights that remain constant across a strip. For skewed images, it falls back to the generic `BilinearImagePainter`. This change alone reduces execution by _~33%_ for bilinear sampling (Medium quality).

image

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/lowp/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/lowp/image.rs
@@ -129,8 +129,6 @@ pub(crate) struct PlainBilinearImagePainter<'a, S: Simd> {
     fy: u16x16<S>,
     /// Pre-computed inverse y interpolation weight
     fy_inv: u16x16<S>,
-    /// Pre-computed rounding bias for integer division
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -802,7 +818,26 @@ impl<S: Simd, T: FineKernel<S>> Fine<S, T> {
...
+                                    fill_complex_paint!(
+                                        i.has_opacities,
+                                        T::plain_medium_quality_image_painter(
+                                            self.simd, i, pixmap, start_x, start_y
+                                        )
+                                    );
+                                } else {
+                                    fill_complex_paint!(
+                                        i.has_opacities,
+                                        T::high_quality_image_painter(
```
tomcur: Would a `plain_high_quality_image_painter` also be in scope?

PR #1336, opened 2025-12-31
perf(vello_common): use SIMD dispatch in flattening codegen

tomcur:
This is a ~10% reduction in flattening time on x86, I haven't measured AArch64.

Flattening was already dispatched to have access to the SIMD witness, but it did not yet unambiguously make use of target features for codegen as the functions weren't forced to be inlined.

```
flatten/Ghostscript_Tiger
                        time:   [208.76 µs 209.06 µs 209.42 µs]
                        change: [-12.177% -11.979% -11.768%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  4 (4.00%) high severe
flatten/paris-30k       time:   [13.157 ms 13.202 ms 13.253 ms]
                        change: [-10.728% -10.307% -9.8772%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -86,6 +86,7 @@ pub fn fill(
 }
 
 /// Flatten a filled bezier path into line segments.
+#[inline(always)]
 pub fn fill_impl<S: Simd>(
     simd: S,
     path: impl IntoIterator<Item = PathEl>,
@@ -94,7 +95,10 @@ pub fn fill_impl<S: Simd>(
     flatten_ctx: &mut FlattenCtx,
 ) {
     line_buf.clear();
-    let iter = path.into_iter().map(|el| affine * el);
+    let iter = path.into_iter().map(
+        #[inline(always)]
+        |el| affine * el,
+    );
 
     let mut lb = FlattenerCallback {
         line_buf,
--- a/sparse_strips/vello_common/src/flatten_simd.rs
+++ b/sparse_strips/vello_common/src/flatten_simd.rs
@@ -25,6 +25,7 @@ pub(crate) trait Callback {
 /// <https://docs.rs/kurbo/latest/kurbo/fn.flatten.html>
 ///
 /// This version works using a similar approach but using f32x4/f32x8 SIMD instead.
+#[inline(always)]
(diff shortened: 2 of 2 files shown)


sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -111,17 +112,12 @@ pub(crate) fn flatten<S: Simd>(
...
-                                    tolerance as f32,
-                                    &mut flattened_cubics,
-                                )
-                            },
+                        let max = flatten_cubic_simd(
+                            simd,
+                            c,
+                            flatten_ctx,
+                            tolerance as f32,
+                            &mut flattened_cubics,
```
tomcur, line 120: This `vectorize` is no longer necessary, as the `flatten_cubic_simd` call gets inlined into `flatten` which itself gets vectorized.

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -94,7 +95,10 @@ pub fn fill_impl<S: Simd>(
     flatten_ctx: &mut FlattenCtx,
 ) {
     line_buf.clear();
-    let iter = path.into_iter().map(|el| affine * el);
+    let iter = path.into_iter().map(
+        #[inline(always)]
+        |el| affine * el,
+    );
```
Philipp-M, line 101: I'm curious whether this has any effect? I would expect this to be inlined basically always given the small closure?

tomcur: > I'm curious whether this has any effect? 

This one more than likely has no effect as this indeed very likely gets inlined without the attribute, too, but this makes it as unambiguous as Rust allows. This follows the suggestion in https://docs.rs/fearless_simd/0.3.0/fearless_simd/#inlining.

tomcur:
> No change on ARM in my benchmarks.

Thanks for checking, the compiler made better inlining decisions on ARM then!

DJMcNab:
> Thanks for checking, the compiler made better inlining decisions on ARM then!

Unfortunately, it's even stupider than that :)

Essentially, all of the relevant `aarch64` targets already unconditionally enable the `neon` feature, so the inlining actually isn't at all necessary on those targets...

PR #1337, opened 2025-12-31
Add DHAT to vello cpu example

taj-p:
DHAT provides allocation statistics with full stack traces which is useful for identifying allocation hot spots.

For example, see how 20% of all allocations from this example trace occur from pushing commands into a wide tile:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,7 +157,7 @@ quote = "1.0.40"
 serde = { version = "1.0.225", default-features = false }
 serde_json = "1.0.145"
 
-# Profiling profile: release optimizations with debug info for better backtraces.
+# Profiling profile: release optimizations with debug info for demangled backtraces.
 [profile.profiling]
 inherits = "release"
 debug = 1


Cargo.toml
```diff
@@ -156,3 +156,8 @@ syn = { version = "2.0.101", features = ["full", "extra-traits"] }
 quote = "1.0.40"
 serde = { version = "1.0.225", default-features = false }
 serde_json = "1.0.145"
+
+# Profiling profile: release optimizations with debug info for better backtraces.
```
taj-p:
```suggestion
# Profiling profile: release optimizations with debug info for demangled backtraces.
```

LaurenzV:
Not opposing having this, but in case you didn't know, you can also trace allocations using Apple Instruments! Here is an example how it looks like: 
image

grebmeg:
> Not opposing having this, but in case you didn't know, you can also trace allocations using Apple Instruments! Here is an example how it looks like:

Thank you for sharing, this is nice tooling! Especially on M4 or later devices, where you can [analyze CPU usage using Processor Trace](https://developer.apple.com/documentation/xcode/analyzing-cpu-usage-with-processor-trace). I think there’s also the idea of writing tests that verify a certain piece of code performs a specific amount of heap allocation when it runs. [`dhat` supports this](https://docs.rs/dhat/latest/dhat/#heap-usage-testing), but I’m not sure whether `dhat` is planned to be used for that.

PR #1338, opened 2025-12-31
perf(sparse_strips): Reuse flattened_cubics allocation

taj-p:
By reusing this allocation, we see 5 to 6.6% improvement to flattening performance.

image

### DIFF
--- a/sparse_strips/vello_common/src/flatten_simd.rs
+++ b/sparse_strips/vello_common/src/flatten_simd.rs
@@ -485,12 +485,7 @@ fn output_lines_simd<S: Simd>(
 }
 
 #[inline(always)]
-fn flatten_cubic_simd<S: Simd>(
-    simd: S,
-    c: CubicBez,
-    ctx: &mut FlattenCtx,
-    accuracy: f32,
-) -> usize {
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -133,6 +134,9 @@ pub(crate) fn flatten<S: Simd>(
             }
         }
     }
+
+    // Put the allocation back into the context for reuse in subsequent calls.
```
taj-p:
```suggestion
    // Return `flattened_cubics` ownership.
```

LaurenzV:
Is it not possible to pass the context all the way down and then access the buffer from the context directly instead of passing it separately? This way we might be able to avoid the mem take and putting it back later on.

taj-p:
> Is it not possible to pass the context all the way down and then access the buffer from the context directly instead of passing it separately? This way we might be able to avoid the mem take and putting it back later on.

Good idea! Thank you - implemented in https://github.com/linebender/vello/pull/1338/commits/401e9fb38b59312817374c53bc6b934935a80f72

PR #1340, opened 2026-01-02
vello_common: move subpath closing logic into `flatten`

tomcur:
This is just a logic change and slight simplification of the lines representation. By moving the burden of closing input subpaths to flattening itself, this will allow, e.g., culling geometry at the level of Béziers. Before, the closing of subpaths lived as a post-processing step after flattening.

The actual change that starts using this for something more exciting is in https://github.com/linebender/vello/pull/1341.

This by itself does not really bring down timings (plus noise makes it hard to measure).

```
flatten/Ghostscript_Tiger
                        time:   [203.93 µs 204.28 µs 204.68 µs]
                        change: [-2.1022% -1.8481% -1.5684%] (p = 0.00 < 0.05)
                        Performance has improved.
  10 (10.00%) high mild
  2 (2.00%) high severe
flatten/paris-30k       time:   [12.222 ms 12.298 ms 12.384 ms]
                        change: [+0.7747% +1.3966% +2.2290%] (p = 0.00 < 0.05)
Found 13 outliers among 100 measurements (13.00%)
  13 (13.00%) high severe
```

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -3,7 +3,7 @@
 
 //! Flattening filled and stroked paths.
 
-use crate::flatten_simd::Callback;
+use crate::flatten_simd::{Callback, LinePathEl};
 use crate::kurbo::{self, Affine, PathEl, Stroke, StrokeCtx, StrokeOpts};
 use alloc::vec::Vec;
 use fearless_simd::{Level, Simd, dispatch};
@@ -102,17 +102,12 @@ pub fn fill_impl<S: Simd>(
 
     let mut lb = FlattenerCallback {
         line_buf,
-        start: kurbo::Point::default(),
-        p0: kurbo::Point::default(),
+        start: Point::ZERO,
+        p0: Point::ZERO,
         is_nan: false,
-        closed: false,
     };
 
     crate::flatten_simd::flatten(simd, iter, TOL, &mut lb, flatten_ctx);
-
-    if !lb.closed {
-        close_path(lb.start, lb.p0, lb.line_buf);
-    }
 
     // A path that contains NaN is ill-defined, so ignore it.
     if lb.is_nan {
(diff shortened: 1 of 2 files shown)


LaurenzV (approved):
Not seeing anything that speaks against this, but it's not entirely clear to me why it's necessary. If culling of cubic's happens before flattening, why does it matter whether we close unclosed sub-paths during flattening or afterwards? 🤔

tomcur:
> why does it matter whether we close unclosed sub-paths during flattening or afterwards? 🤔

We need to ensure open input subpaths get closed at some point (as we've been doing before), but then want to allow open subpaths later in rendering for culling reasons.

Say we we have a path like the following, with the box describing the viewport and `*` marking the path.

```
                  * * * * *
                *         *
   ---------- * -----     *
   |        *       |     *
   |      *         |     *
   |    *           |     *
   |  *             |     *
   |*               |     *
  *|                |     *
*  |                |     *
*  |                |     *
*  ------------------     *
*                         *
*                         *
*                         *
*                         *
*                         *
*                         *
*                         *
* * * * * * * * * * * * * *
```

The two elements of that path that matter for rendering are the diagonal line and the left-of-viewport vertical line, and we'd want to render as follows.

```
                  *
                * 
   ---------- * -----
   |        *░░░░░░░|
   |      *░░░░░░░░░|
   |    *░░░░░░░░░░░|
   |  *░░░░░░░░░░░░░|
   |*░░░░░░░░░░░░░░░|
  *|░░░░░░░░░░░░░░░░|
*  |░░░░░░░░░░░░░░░░|
*  |░░░░░░░░░░░░░░░░|
*  ------------------
*
*
* 
*
*
*
*
```

If we cull the three elements that do not matter, and as a post-processing step enforce paths to be closed, we instead get

```
                  *
                **
   ---------- * *----
   |        *░░*    |
   |      *░░░*     |
   |    *░░░░*      |
   |  *░░░░░*       |
   |*░░░░░░*        |
  *|░░░░░░*         |
*  |░░░░░*          |
*  |░░░░*           |
*  --- * ------------
*     *
*    *
*   *
*  *
* *
**
*
```

tomcur:
I've added the above as docstrings to https://github.com/linebender/vello/pull/1341.

PR #1341, opened 2026-01-02
perf(vello_common): Cull Bézier path elements during flattening

tomcur:
On top of https://github.com/linebender/vello/pull/1340.

tldr: If only some of Tiger's whiskers are visible, this results in -90% and -60% timings on flattening and tiling.

This conservatively checks whether Bézier path elements we're about to flatten are outside the viewport. If they are fully to the right, top, or bottom of the viewport, the Bézier does not impact pixel coverage or coarse winding at all, and can be ignored.

If it is fully to the left, it does impact pixel coverage and coarse winding, but only the element's start and endpoint y-values matter, not the exact shape, meaning we can just yield a line rather than finely flattening.

The following two Ghostscript Tigers have their viewboxes reduced to `50 50 100 100` and `90 90 20 20`, down from `0 0 200 200`. Their flattening time is reduced by 52% and 90% respectively, and their tiling time by 22% and 60%.

<details>
<summary>Expand for the Tiger files</summary>

`Ghostscript_Tiger-viewboxed.svg`

Ghostscript_Tiger-viewboxed

`Ghostscript_Tiger-viewboxed-extreme.svg`

Ghostscript_Tiger-viewboxed-extreme

</details>

If more or less everything ends up being in the viewport, the additional calculation is wasted and increases flattening time by ~3%. This is of course workload-dependent.

Flattening timings:

```
flatten/Ghostscript_Tiger
                        time:   [209.94 µs 210.21 µs 210.51 µs]
                        change: [+2.6850% +3.1753% +3.6309%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (4.00%) high mild
  1 (1.00%) high severe
flatten/Ghostscript_Tiger-viewboxed
                        time:   [97.189 µs 97.287 µs 97.399 µs]
                        change: [-52.787% -52.650% -52.514%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) high mild
  2 (2.00%) high severe
flatten/Ghostscript_Tiger-viewboxed-extreme
                        time:   [19.722 µs 19.741 µs 19.761 µs]
                        change: [-90.311% -90.280% -90.255%] (p = 0.00 < 0.05)
                        Performance has improved.
  7 (7.00%) high mild
  1 (1.00%) high severe
flatten/paris-30k       time:   [12.740 ms 12.764 ms 12.788 ms]
                        change: [+2.6014% +3.3631% +4.0837%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
```

Tiling timings:

```
tile/Ghostscript_Tiger  time:   [175.39 µs 175.79 µs 176.28 µs]
                        change: [-0.4403% -0.0016% +0.4400%] (p = 1.00 > 0.05)
  1 (2.00%) high mild
tile/Ghostscript_Tiger-viewboxed
                        time:   [78.932 µs 79.147 µs 79.409 µs]
                        change: [-23.209% -22.803% -22.369%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 50 measurements (10.00%)
  1 (2.00%) high mild
  4 (8.00%) high severe
tile/Ghostscript_Tiger-viewboxed-extreme
                        time:   [13.378 µs 13.390 µs 13.405 µs]
                        change: [-60.417% -60.306% -60.199%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 50 measurements (12.00%)
  2 (4.00%) high mild
  4 (8.00%) high severe
tile/paris-30k          time:   [20.970 ms 21.001 ms 21.034 ms]
                        change: [-0.4881% -0.2397% +0.0108%] (p = 0.07 > 0.05)
  1 (2.00%) high mild
```

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -200,6 +200,10 @@ pub fn fill_impl<S: Simd>(
     height: u16,
 ) {
     line_buf.clear();
+    let iter = path.into_iter().map(
+        #[inline(always)]
+        |el| affine * el,
+    );
 
     let mut lb = FlattenerCallback {
         line_buf,
@@ -208,7 +212,7 @@ pub fn fill_impl<S: Simd>(
         is_nan: false,
     };
 
-    crate::flatten_simd::flatten(simd, affine, path, &mut lb, flatten_ctx, width, height);
+    crate::flatten_simd::flatten(simd, iter, &mut lb, flatten_ctx, width, height);
 
     // A path that contains NaN is ill-defined, so ignore it.
     if lb.is_nan {
--- a/sparse_strips/vello_common/src/flatten_simd.rs
+++ b/sparse_strips/vello_common/src/flatten_simd.rs
@@ -8,7 +8,7 @@
 use crate::flatten::{SQRT_TOL, TOL, TOL_2};
 #[cfg(not(feature = "std"))]
 use crate::kurbo::common::FloatFuncs as _;
-use crate::kurbo::{Affine, CubicBez, Line, ParamCurve, ParamCurveNearest, PathEl, Point, QuadBez};
+use crate::kurbo::{CubicBez, Line, ParamCurve, ParamCurveNearest, PathEl, Point, QuadBez};
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
 use fearless_simd::*;
@@ -43,7 +43,6 @@ pub(crate) trait Callback {
 #[inline(always)]
 pub(crate) fn flatten<S: Simd>(
     simd: S,
-    affine: Affine,
     path: impl IntoIterator<Item = PathEl>,
     callback: &mut impl Callback,
     flatten_ctx: &mut FlattenCtx,
@@ -60,8 +59,6 @@ pub(crate) fn flatten<S: Simd>(
     let mut last_pt = Point::ZERO;
 
     for el in path {
-        let el = affine * el;
-
         match el {
             PathEl::MoveTo(p) => {
                 if !closed && last_pt != start_pt {


tomcur (commented):
I wonder whether we should track the axis-aligned bounding box of the active clip (and separately, we may want to provide a fast-path for rectangular clips). If we have that bounding box, we can generalize this viewport culling to clip culling during tiling as well as here. For some workloads that may provide quite a few more culling opportunities.

Second, a small point: the culling we do here and already did in tiling, means a paths' cached strips cannot be translated if any culling occurred. I believe some discussion was already held that we may want to move tiles and strips to `i16` coordinates to support translation, and additionally, if paths are to be cached, we may not want to do any culling. Otherwise work will have to done to rerender if a path is translated.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -27,105 +43,159 @@ pub(crate) trait Callback {
...
+                //
+                // A cheap upper bound for dist(p1, [p0, p1]) is max(dist(p1, p0), dist(p1, p2)).
+                //
+                // The following takes the square to elide the square root of the Euclidean
+                // distance.
+                else if [p0, p1, p2].into_iter().all(|p| p.x < 0.)
+                    || f64::max((p1 - p0).hypot2(), (p1 - p2).hypot2()) <= 4. * TOL_2
+                {
+                    callback.callback(LinePathEl::LineTo(p2));
+                } else {
```
tomcur: A small observation: for the left-of-viewport case, this could in fact just return a vertical line, as it does not have to be watertight at all, but in terms of performance that isn't likely to buy us much. It's an additional branch here for a slight performance improvement during tiling.

LaurenzV:
> I wonder whether we should track the axis-aligned bounding box of the active clip (and separately, we may want to provide a fast-path for rectangular clips). If we have that bounding box, we can generalize this viewport culling to clip culling during tiling as well as here. For some workloads that may provide quite a few more culling opportunities.

yup, this would probably be possible!

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -76,31 +86,120 @@ impl Line {
...
     line_buf: &mut Vec<Line>,
     flatten_ctx: &mut FlattenCtx,
+    width: u16,
+    height: u16,
 ) {
     line_buf.clear();
-    let iter = path.into_iter().map(
-        #[inline(always)]
-        |el| affine * el,
-    );
```
LaurenzV, line 103: I'm wondering whether there is a specific reason this had to be moved into the method?

tomcur: It's been a while so I don't remember, probably just something left over from an early iteration on this. I've reverted that part. (As a sanity check, benches show no difference, as expected.)

tomcur: (Well-spotted btw!)

PR #1342, opened 2026-01-05
classic: implement dashed stroke clipping

tomcur:
This follows up on https://github.com/linebender/vello/pull/1332, implementing dashed stroke clipping as well.

The first commit here does just that (plus rejigging some of the control flow), the second factors out stroking logic shared with `Scene::stroke` to a new private method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ This release has an [MSRV][] of 1.88.
 ### Changed
 
 - Breaking change: wgpu has been updated to wgpu 27. ([#1280][] by [@theoparis][])  
-- Breaking change: Allow setting `Scene` layer clip shape drawing style, adding even-odd filled path clipping and stroked path clipping to the various scene layer methods (`Scene::{push_layer, push_luminance_mask_layer, push_clip_layer}`). ([#1332][] by [@waywardmonkeys][], [#1342][] by [@tomcur][])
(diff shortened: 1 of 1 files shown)


vello/src/scene.rs
```diff
@@ -432,6 +400,50 @@ impl Scene {
...
+        debug_assert!(encoded_stroke, "Stroke width is non-zero");
+
+        // We currently don't support dashing on the GPU. If the style has a dash pattern, then
+        // we convert it into stroked paths on the CPU and encode those as individual draw
+        // objects.
+        let encode_result = if style.dash_pattern.is_empty() {
+            #[cfg(feature = "bump_estimate")]
+            self.estimator
+                .count_path(shape.path_elements(SHAPE_TOLERANCE), &t, Some(style));
+            self.encoding.encode_shape(shape, false)
```
tomcur, line 425: I didn't want to tackle (or think about) tolerances here, and this is just lifted directly from the code in `main`, but I'm not entirely sure this is correct... The `Estimator::count_path` call gets `SHAPE_TOLERANCE` passed in (with a value of 0.01), but the `Encoding::encode_shape` call has a hard-coded `0.1` tolerance.

Additionally, as the note about tolerances in `Scene::stroke` says, this does nothing for handling tolerance under scaling.

DJMcNab:
We should probably also update the changelog to better reflect the support for stroked clips too. https://github.com/linebender/vello/pull/1332's changelog seems to reflect the version before https://github.com/linebender/vello/pull/1332#issuecomment-3687341472

tomcur:
> We should probably also update the changelog to better reflect the support for stroked clips too. https://github.com/linebender/vello/pull/1332 changelog seems to reflect the version before https://github.com/linebender/vello/pull/1332#issuecomment-3687341472

I'll probably get to this tonight!

PR #1344, opened 2026-01-07
Update the changelog for sparse strips renderers

LaurenzV:
I've set the release date for tomorrow.

### DIFF
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -22,6 +22,7 @@ This release has an [MSRV][] of 1.88.
 - An `extend` method has been added to `StripStorage` to extends
   its alphas/strips from another `StripStorage` ([#1203][] by [@LaurenzV])
 - Added a `from_parts` method for masks ([#1237][] by [@LaurenzV])
+- Add initial support for image filters ([#1286][] by [@grebmeg])
 
 ### Changed
 - `WideTile::generate` now takes an additional `BlendMode` as a parameter ([#1159][] by [@LaurenzV])
@@ -35,6 +36,11 @@ This release has an [MSRV][] of 1.88.
 - The `generate` method of `Wide` now takes an optional mask as an
   additional argument ([#1237][] by [@LaurenzV])
 - `CmdFill` and `CmdAlphaFill` now store an optional mask ([#1237][] by [@LaurenzV])
+- Performance improvements for gradient rendering ([#1301][] by [@valadaptive])
+- Various changes to the logic for computing tile intersections and 
+  representation of tiles ([#1293][], [#1317][], [#1318][] by [@b0nes164])
+- Support for computing data necessary to implement multi-sampled anti-aliasing ([#1319][], by [@b0nes164])
+- Numerous performance and memory-efficiency improvements ([#1325][] by [@LaurenzV], [#1327][] by [@grebmeg], [#1336][] by [@tomcur], [#1338][] by [@taj-p])
 
 ## [0.0.4][] - 2025-10-17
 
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_common/src/recording.rs
```diff
@@ -445,6 +445,8 @@ impl<'a> Recorder<'a> {
     }
 
     /// Push a new filter layer.
+    ///
+    /// WARNING: Note that filters are currently incomplete and experimental.
```
nicoburns, line 449: Can we be a little more specific here. I believe:
- Completely ignored in `vello_hybrid`
- Likely to crash in `vello_cpu` with multithreading enabled
- Limited in `vello_cpu` (single-threaded) (perhaps link to github issue?)

sparse_strips/vello_common/CHANGELOG.md
```diff
@@ -69,12 +79,28 @@ This is the initial release. No changelog was kept for this release.
...
+[#1301]: https://github.com/linebender/vello/pull/1301
+[#1293]: https://github.com/linebender/vello/pull/1293
+[#1286]: https://github.com/linebender/vello/pull/1286
 [#1237]: https://github.com/linebender/vello/pull/1237
 [#1203]: https://github.com/linebender/vello/pull/1203
 [#1159]: https://github.com/linebender/vello/pull/1159
 
-[Unreleased]: https://github.com/linebender/fearless_simd/compare/sparse-strips-v0.0.4...HEAD
+[Unreleased]: https://github.com/linebender/fearless_simd/compare/sparse-strips-v0.0.5...HEAD
+[0.0.4]: https://github.com/linebender/vello/compare/sparse-stips-v0.0.4...sparse-strips-v0.0.5
```
DJMcNab:
```suggestion
[0.0.5]: https://github.com/linebender/vello/compare/sparse-stips-v0.0.4...sparse-strips-v0.0.5
```

sparse_strips/vello_common/src/recording.rs
nicoburns: (and/or perhaps on the `vello_cpu` method - comment on `vello_hybrid` seems fine)

nicoburns (changes requested):

sparse_strips/vello_common/src/recording.rs
LaurenzV: I believe I added a comment it in vello_cpu, no? I'll clarify the comment in vello_hybrid that they will actually be ignored.

nicoburns: Oh, I missed that. Nevermind.

nicoburns (approved):

sparse_strips/vello_common/src/recording.rs
LaurenzV: Or do you think i should duplicate those into vello_common? But since vello_common is kind of an internal crate I think it should be fine, no?

LaurenzV:
I'll wait until tomorrow in case someone else wants to take a look. Perhaps #1329 and #1335 will be merged by then, but probably not a big deal if not.

DJMcNab (commented):
I haven't reviewed the content, just given a first pass and found this issue.

As a small process note, I see that you've not updated the versions in the Cargo.toml files. That is correct, because this is the CHANGELOG update PR, but note that we also shouldn't be updating the version numbers in the CHANGELOG files either, until the actual release PR. That ensures that the changelog file on main accurately reflects the actual released/unreleased state of the commit. We should also probably change the title of this PR.
I'd also like to see a PR to make sure our dependencies are up-to-date before the release.

Remember that we have https://linebender.org/wiki/process/release/ documented. If anything in there is unclear, feel free to reach out to ask!

LaurenzV:
I reverted the version bump in the changelog, as mentioned in the Zulip, I'm not sure what to do about the dependencies though. Should I just blindly update all of them, even if some aren't used by sparse strips?

PR #1351, opened 2026-01-10
vello_cpu(filters): Add offset filter

waywardmonkeys:
* Implement `FilterPrimitive::Offset` bounds expansion in `vello_common`
* Add `vello_cpu` `Offset` filter using shared in-place pixel shift helper
* Refactor `DropShadow` to reuse shared shift code
* Add `vello_sparse_tests` coverage + reference snapshot for offset

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/filter.rs
+++ b/sparse_strips/vello_sparse_tests/tests/filter.rs
@@ -907,14 +907,14 @@ fn filter_offset(ctx: &mut impl Renderer) {
     ctx.set_paint(SEA_GREEN);
     ctx.fill_rect(&marker);
 
-    // Draw the filtered (shifted) star as a stroke and fill, then draw the marker through the
+    // Draw the filtered (shifted) star as a fill and stroke, then draw the marker through the
     // filter layer as well.
     ctx.push_filter_layer(filter);
+    ctx.set_paint(TOMATO);
+    ctx.fill_path(&star_path);
     ctx.set_paint(BLACK);
     ctx.set_stroke(Stroke::new(1.5));
     ctx.stroke_path(&star_path);
-    ctx.set_paint(TOMATO);
-    ctx.fill_path(&star_path);
     // With (dx, dy) = (18, -12) this should land at (67, 15).
     ctx.set_paint(VIOLET);
     ctx.fill_rect(&marker);
--- a/sparse_strips/vello_sparse_tests/snapshots/filter_offset.png
+++ b/sparse_strips/vello_sparse_tests/snapshots/filter_offset.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b40c31850d2d961cfccada9e5afa759cd303dc14e92f2aaf671ea44f0ab10a2d
-size 2313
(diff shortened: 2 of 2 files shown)


sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -884,6 +884,29 @@ fn filter_drop_shadow_zero_offset(ctx: &mut impl Renderer) {
...
 
+/// Test offset filter primitive.
+///
+/// This shifts content within a filter layer and should not clip content to the original bounds.
+#[vello_test(skip_hybrid, skip_multithreaded)]
+fn filter_offset(ctx: &mut impl Renderer) {
+    let filter = Filter::from_primitive(FilterPrimitive::Offset {
+        dx: 18.0,
+        dy: -12.0,
+    });
```
grebmeg, line 895: I think the offset might be working incorrectly. Starting from (49, 27) with (dx, dy) = (18, -12) should shift the image to (67, 15), right? That said, I don’t think this is caused by your PR, it mostly moves code out and reuses existing logic, and I don’t think any update logic changed, correct? If this is a bug, maybe we should create an issue for it?

image

waywardmonkeys: Codex has a response for you:

> That screenshot is comparing two different things:
> 
> * The blue pixel at (49, 27) is from the unfiltered stroked outline (stroke_path(&star_path)).
> * The red pixels are from the filtered filled star drawn inside push_filter_layer(filter) (fill_path(&star_path)).
> 
> A stroke is not the same geometry as a fill (it’s expanded by ~½ stroke width and then anti-aliased), so a particular stroke-edge pixel coordinate is not expected to map to a particular fill pixel after offset. On top of that, the pixel the arrow points at is an AA fringe pixel, so even “same geometry” comparisons are often off by 1.
> 
> So I don’t think (49,27) + (18,-12) “should” imply anything about where the red fill lands in sparse_strips/vello_sparse_tests/tests/filter.rs’s filter_offset test — it’s just not tracking the same feature.
> 
> If you want a crisp correctness check for feOffset, the test should render a fill-to-fill (or marker-to-marker) comparison, e.g.:
> 
> * draw a 1×1 or 2×2 filled rect/pixel inside the filter layer and assert it appears at exactly (x+dx, y+dy), or
> * draw the same filled shape twice (unfiltered outline/fill in one color + filtered fill in another) and compare interior pixels, not stroke edges.

I pointed out to Codex that it was the one that chose to use a stroke + fill.

We're amending the test to be less prone to this.

tomcur: > A stroke is not the same geometry as a fill (it’s expanded by ~½ stroke width and then anti-aliased), so a particular stroke-edge pixel coordinate is not expected to map to a particular fill pixel after offset.

That's correct. The stroke width here is  1.5, so you'd expect a ~0.75 pixel offset. Modulo offset and ignoring joins, the edge of the filled shape should land roughly in the middle of the stroked shape.

grebmeg: Nice! Thank you, somehow when I noticed the difference, I didn’t pay attention to the fact that fill and stroke are being used. 🙏

PR #1353, opened 2026-01-13
Update to skrifa 0.40

waywardmonkeys:
This matches harfrust 0.5.0 and what should be in the next Parley release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1054,6 +1054,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "font-types"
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Do you want this to be in v0.7.0? We can always do a patch release to bump this now, at least.

waywardmonkeys:
I would like it in 0.7 ...

waywardmonkeys:
Then I can update Parley too and remove some duplicate stuff for a while.

PR #1356, opened 2026-01-13
Document features in Sparse Strips more correctly

DJMcNab:
Also bring in some better consistency with Xilem/Masonry readmes.

I also audited the features of Vello Common, and found that there were two available feature flags which were unmentioned; i.e. removing them didn't actually require any other changes. These are the `bytemuck` and `simd` features. As such, I suggest that we should remove them, and have done so in this PR.

### DIFF
--- a/sparse_strips/vello_cpu/README.md
+++ b/sparse_strips/vello_cpu/README.md
@@ -103,7 +103,7 @@ to better understand how to interact with Vello CPU's API,
   The `u8` pipeline will be used for [`OptimizeSpeed`][RenderMode::OptimizeSpeed], if both pipelines are enabled.
   If you're using Vello CPU for application rendering, you should prefer this pipeline.
 - `f32_pipeline`: Enable the `f32` pipeline, which is slower but has more accurate
-  and results. This is espectially useful for rendering test snapshots.
+  results. This is espectially useful for rendering test snapshots.
   The `f32` pipeline will be used for [`OptimizeQuality`][RenderMode::OptimizeQuality], if both pipelines are enabled.
 
 At least one of `std` and `libm` is required; `std` overrides `libm`.
--- a/sparse_strips/vello_cpu/src/lib.rs
+++ b/sparse_strips/vello_cpu/src/lib.rs
@@ -73,7 +73,7 @@
 //!   The `u8` pipeline will be used for [`OptimizeSpeed`][RenderMode::OptimizeSpeed], if both pipelines are enabled.
 //!   If you're using Vello CPU for application rendering, you should prefer this pipeline.
 //! - `f32_pipeline`: Enable the `f32` pipeline, which is slower but has more accurate
-//!   and results. This is espectially useful for rendering test snapshots.
+//!   results. This is espectially useful for rendering test snapshots.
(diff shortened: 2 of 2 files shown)


sparse_strips/vello_common/Cargo.toml
```diff
@@ -29,12 +29,10 @@ libm = { version = "0.2.15", optional = true }
 log = { workspace = true }
 
 [features]
+# If adding new features, also document in `src/lib.rs`
 default = ["std", "png", "text"]
-# Enable using SIMD instructions for rendering
-simd = []
 # Get floating point functions from the standard library (likely using your target's libc).
 std = ["peniko/std", "skrifa?/std", "fearless_simd/std"]
-bytemuck = ["bytemuck/bytemuck_derive"]
```
nicoburns, line 37: `flatten_simd.rs` is using this (but it's also using it unconditionally, not feature flagged)

sparse_strips/vello_common/Cargo.toml
```diff
@@ -29,12 +29,10 @@ libm = { version = "0.2.15", optional = true }
 log = { workspace = true }
 
 [features]
+# If adding new features, also document in `src/lib.rs`
 default = ["std", "png", "text"]
-# Enable using SIMD instructions for rendering
-simd = []
```
nicoburns, line 34: Note that there is SIMD code is `vello_common`, it's just not currently feature flagged.

sparse_strips/vello_common/Cargo.toml
DJMcNab: Yeah. The fact that this code didn't use the feature flag meant I was safe to remove it.

I will check whether we need to add the derive feature; it's likely that we do

sparse_strips/vello_common/Cargo.toml
DJMcNab: ... exactly, that's why I removed the feature flag. 

You're right that I should have called it out in the PR description though!

sparse_strips/vello_cpu/README.md
```diff
@@ -93,12 +94,22 @@ to better understand how to interact with Vello CPU's API,
...
-  Also required for rendering glyphs with an embedded PNG.
-- `multithreading`: Enable multi-threaded rendering.
+  Also required for rendering glyphs with an embedded PNG. Implies `std`.
+- `multithreading`: Enable multi-threaded rendering. Implies `std`.
+- `text` (enabled by default): Enables glyph rendering ([`glyph_run`][RenderContext::glyph_run]).
+- `u8_pipeline` (enabled by default): Enable the u8 pipeline, for speed focused rendering using u8 math.
+  The `u8` pipeline will be used for [`OptimizeSpeed`][RenderMode::OptimizeSpeed], if both pipelines are enabled.
+  If you're using Vello CPU for application rendering, you should prefer this pipeline.
+- `f32_pipeline`: Enable the `f32` pipeline, which is slower but has more accurate
+  and consistent results. This is espectially usedful for rendering text snapshots.
```
LaurenzV: I would maybe cross the consistent? I think it would actually be less consistent in general due to floating point differences across systems, while integer integer arihmetic should be the same everywhere.

DJMcNab: Changed.

PR #1359, opened 2026-01-13
docs: update vello hybrid readme and lib.rs to include feature-flags

### LINKED ISSUE #1355: Document Vello Hybrid's features
Vello Hybrid's feature flags need to be documented in its lib.rs/README.md, and aren't currently.
See https://github.com/linebender/vello/blob/184d6416521d348380563c933e245766a2123647/sparse_strips/vello_hybrid/README.md#features and
https://github.com/linebender/vello/blob/184d6416521d348380563c933e245766a2123647/sparse_strips/vello_hybrid/Cargo.toml#L49-L57

Compare to https://github.com/linebender/vello/blob/184d6416521d348380563c933e245766a2123647/sparse_strips/vello_cpu/README.md#features

ZeleOeO:
Resolves #1355 

I added the feature flags for Vello Hybrid in the README and also in its `lib.rs`, which weren't there

I had to add a new features flag section since it already has a features section, but I didn't want to mix it up

I also added the note from the `Cargo.toml`  about the default vs wgpu configuration, which seemed like a useful detail.

This is my first time contributing here, so let me know if the formatting looks off or if you'd rather I merge the two feature sections

### DIFF
--- a/sparse_strips/vello_hybrid/src/lib.rs
+++ b/sparse_strips/vello_hybrid/src/lib.rs
@@ -26,7 +26,9 @@
 //! - `wgpu_default` (enabled by default): Enables wgpu with its default hardware backends (such as Vulkan, Metal, and DX12).
 //! - `webgl`: Enables the WebGL rendering backend for browser support, using GLSL shaders for compatibility.
 //!
-//! If you need to customize the set of enabled wgpu features, disable this crate's default features then enable its wgpu feature, and then use the wgpu flag directly with the specific features you require.
+//! If you need to customize the set of enabled wgpu features, disable this crate's default features then enable its `wgpu` feature.
+//! You can then depend on wgpu directly, setting the specific features you require.
+//! Don't forget to also disable wgpu's default features.
 //!
 //! ## Architecture
 //!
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -19,6 +19,16 @@
 //! - GPU-accelerated compositing and blending
 //! - Support for both windowed and headless rendering
 //!
+//!
+//! ## Feature Flags
+//!
+//! - `wgpu` (enabled by default): Enables the GPU rendering backend via wgpu and includes the required sparse shaders.
+//! - `wgpu_default` (enabled by default): Enables wgpu with its default hardware backends (such as Vulkan, Metal, and DX12).
+//! - `std`: Enables the standard library for integration with vello_common and standard error traits.
```
DJMcNab: Can you clarify why you think this feature enables the standard error traits? I've not been able to find the code which does so. The reason I checked is that the `Error` trait also moved to `core` earlier than our MSRV, so it would have been a bug if we did so.

DJMcNab (commented):
Welcome! Thank you for addressing this issue; this implementation is good (barring the one confusing comment about the standard error traits). 

However, it's uncovered a different issue that the `std` feature flag on this crate doesn't make sense (`std` is not default, it enables wgpu, and we also unconditionally enable Vello Common's `std` feature).
To let us land this, can you please just remove the mention of the `std` feature entirely from these docs, and I will remove it entirely as a follow-up.

sparse_strips/vello_hybrid/src/lib.rs
ZeleOeO: Okay, I've updated the PR and removed the std feature flag mentions from the readme and the lib.rs file.

My apologies for the confusion regarding the error trait part in the documentation. Since the flag linked to vello_common/std, I made the assumption that it was a bridge for no_std compatibility, specifically for the error trait, which is a common pattern. I've cleared that out now.

Thanks for pointing that out to me. I shouldn't have made that assumption, though.

PR #1360, opened 2026-01-14
Implement single-`Scene` painting for Vello API

DJMcNab:
This is a core part of #1299, split out to be re-viewable separately.

One observation which I hadn't properly engaged with until today is that the form proposed in this PR is *independently* useful. That is, useful apps could be made renderer-agnostic with just this API, with none of the "renderer abstraction" pain of #1299. I do think that the renderer API in #1299 does have valuable aspects (having textures be reference counted is still something I really want us to have), but it isn't needed to get something off-the-ground.

The core of this PR is two things:
1) The `PaintScene` trait. This is your standard 2d imaging model trait (comparable e.g. with web canvas, [`vello::Scene`](https://docs.rs/vello/latest/vello/struct.Scene.html), [`anyrender::PaintScene`](https://docs.rs/anyrender/latest/anyrender/trait.PaintScene.html)).
2) A renderer-erased implementation of that trait, called `vello_api::Scene`. I call it renderer-erased, because it is (or... will be) invalid to combine renderers within a Scene graph. That's because of image resources and the like.

This PR does not have glyph support, and indeed I still believe that Vello API should continue to not have that.

### DIFF
--- a/sparse_strips/vello_cpu/Cargo.toml
+++ b/sparse_strips/vello_cpu/Cargo.toml
@@ -32,7 +32,7 @@ default = ["std", "png", "text", "u8_pipeline"]
 # Get floating point functions from the standard library (likely using your target’s libc).
 std = ["vello_common/std", "vello_api/std"]
 # Use floating point implementations from libm.
-libm = ["vello_common/libm", "vello_api/std"]
+libm = ["vello_common/libm", "vello_api/libm"]
 # Allow loading Pixmap from PNG, and drawing png glyphs.
 png = ["vello_common/png"]
 # Enable multi-threaded rendering.


sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Pop the current layer.
+    PopLayer,
+    /// Set the current paint.
+    ///
+    /// The affine is currently path local for future drawing operations.
+    /// That is something I expect *could* change in the future/want to change.
+    ///
+    /// This doesn't use [`OurBrush`] because we want to limit the
+    /// number of textures stored.
+    SetPaint(Affine, peniko::Brush<peniko::ImageBrush<TextureId>>),
```
nicoburns: This type appears to be the same as `OurBrush`. Possibly an outdated comment?

sparse_strips/vello_api/src/paths.rs
```diff
@@ -0,0 +1,153 @@
...
+/// This separation is designed for a future path caching mechanism, where the rasterised geometry
+/// of a path can be computed once, then re-used with multiple brushed, to increase efficiency.
+/// Note however that this plan is not proven in the current version of Vello API.
+#[derive(Debug)]
+// The same "reason about visibility" comment applies as in `Scene`
+pub struct PathSet {
+    // There are arguments for a "dynamic length" encoding here, as PathEl is sized for 6 f64s (plus a disciminant)
+    // It depends somewhat on what proportion of the elements are a CurveTo
+    /// The elements of the contained paths.
+    pub elements: Vec<PathEl>,
```
nicoburns, line 52: We should definitely defer this until later, but it occurs to me that this kind of optimisation might also work well for gradient color stops

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,154 @@
...
+        let shape: Rect = rect.inflate(kernel_size, kernel_size);
+        self.fill_path(transform, Fill::EvenOdd, shape);
+    }
+
+    fn push_layer(
+        &mut self,
+        clip_transform: Affine,
+        clip_path: Option<impl Shape>,
+        blend_mode: Option<BlendMode>,
+        opacity: Option<f32>,
```
nicoburns, line 204: Q: What is the rationale for these being `Option<T>` rather than requiring the user to set `opacity: 1.0` or `BlendMode: BlendMode::NORMAL`?

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,154 @@
...
+    // Error if associated with different renderer.
+    // TODO: This also "clobbers" the brush; we need to document that.
+    fn append(&mut self, transform: Affine, scene: &Scene) -> Result<(), ()>;
+
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape);
+    fn stroke_path(&mut self, transform: Affine, stroke_params: &Stroke, path: impl Shape);
+
+    fn set_brush(
+        &mut self,
+        brush: impl Into<OurBrush>,
```
nicoburns: FWIW, I'm not entirely convinced that having `impl Into` in the API makes the ergnomics better. It's nice if you're passing a literal (no need to call `.into()`). But it's annoying if you're converting from some other type, because it breaks type inference and means that an `Into` implementation from that type to e.g. `OurBrush` won't work (and you also can't turbofish `Into::into`, so you generally have to break out the conversion into it's own let binding).

sparse_strips/vello_api/src/lib.rs
```diff
@@ -1,9 +1,140 @@
...
+//!   It is one of the fastest CPU-only 2D renderers in Rust.
+//! - Vello Hybrid, which runs the most compute intensive portions of rendering on the GPU, improving performance over Vello CPU.
+//!   It has wide compatibility with most devices, so long as they have a GPU, and it runs well on the web.
+//! <!-- We might also have, to be determined:
+//! - Vello Classic, which performs almost all rendering on the GPU, which gives great performance on devices with decent GPUs.
+//!   However, it cannot run well on devices with weak GPUs, or in contexts without support for compute shaders, such as the web.
+//!   It also has unavoidably high memory usage, and can silently fail to render if the scene gets too big.
+//! -->
+//!
+//! As a general guide for consumers, you should prefer Vello Hybrid for applications, and Vello CPU for headless use cases
```
nicoburns: While this *will* (hopefully) be true, I would currently unconditionally recommend Vello CPU over Vello Hybrid: Hybrid is still too crashy to be recommended, and is also slower than CPU in many cases. Classic is still the only option is you want performance competitive with other GPU renderers.

sparse_strips/vello_api/src/lib.rs
```diff
@@ -1,9 +1,140 @@
...
+//!
+//! # Text
+//!
+//! Vello API does not handle text/glyph rendering itself.
+//! This allows for improved resource sharing of intermediate text layout data, for hinting and ink splitting underlines.
+//!
+//! Text can be rendered to Vello API scenes using the "Parley Draw" crate.
+//! Note that this crate is not currently implemented; design work is ongoing.
+//! We also support rendering using using traditional glyph atlases, which may be preferred by some consumers.
+//! This is especially useful to achieve subpixel rendering, such as ClearType, which Vello doesn't currently support directly.
```
nicoburns, line 69: How soon do you we anticipate Parley Draw being available? In my mind it could still be several month away. As such, I wonder if it would be worth temporarily adding a method to the `PaintScene` trait that mirrors the glyph rendering API that `vello_cpu` and `vello_hybrid` currently have. The idea being that we would remove it once we have an implementation available for the Parley Draw approach.

(we could also optionally consider having a transitional period where both APIs are available)
(this is essentially the approach that AnyRender has taken)

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,154 @@
...
+/// As such, the inferred clipping from the scene viewports is limited.
+///
+/// [`Renderer::create_scene`]: crate::Renderer::create_scene
+/// [`set_blurred_rounded_rect_brush`]: PaintScene::set_blurred_rounded_rect_brush
+/// [`set_brush`]: PaintScene::set_brush
+/// [`set_solid_brush`]: PaintScene::set_solid_brush
+/// [`fill_path`]: PaintScene::fill_path
+/// [`append`]: PaintScene::append
+/// [`Renderer`]: crate::Renderer
+pub trait PaintScene: Any {
```
nicoburns, line 59: I definitely get the appeal of the `Any` bound here. But I would note that this would prevent `PaintScene` from being implemented for any type that contains a reference. We may want to consider some other kind of reference-centric downcasting...

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,154 @@
...
+        paint_transform: Affine,
+        color: Color,
+        rect: &Rect,
+        radius: f32,
+        std_dev: f32,
+    );
+    fn set_solid_brush(&mut self, color: Color) {
+        // The transform doesn't matter for a solid color brush.
+        self.set_brush(Brush::Solid(color), Affine::IDENTITY);
+    }
```
nicoburns, line 148: Q: What's the rationale behind the API being stateful?

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,290 @@
...
+///
+/// The [`PathId`]s contained within are the index into the pathset associated with this `Scene`.
+/// As such, when moving these commands between scenes, the path id must be updated.
+/// (N.B. this will be less true when we get path caching, if it follows the expected design).
+///
+/// The abstract renderer these commands operate on has the state described in the [`PaintScene`] trait; that is, the current brush and a layer stack.
+#[derive(Debug)]
+pub enum RenderCommand {
+    /// Draw a path with the current brush.
+    DrawPath(Affine, PathId),
```
nicoburns, line 36: API Style: could we use structs rather than tuple/struct enum variants? The latter can be difficult to work with (can't take reference or pass int a function, etc).

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,290 @@
...
+/// design to specific pixels.
+/// Usefulness for drawing glyphs (i.e. characters of text) and, but more invalidation required.
+#[derive(Debug)]
+// TODO: Reason about visibility; do we want a more limited "expose" operation
+// which lets you read all the fields?
+// This also applies to `PathSet`
+pub struct Scene {
+    pub paths: PathSet,
+    pub commands: Vec<RenderCommand>,
+    pub hinted: bool,
```
nicoburns, line 110: Q: Is "hintedness" actually a global property of a `Scene`. Or might it apply to some items in the scene but not others?

sparse_strips/vello_api/src/lib.rs
```diff
@@ -1,9 +1,140 @@
 // Copyright 2025 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! A stub crate.
-//! Planned to provide an abstraction between different renderers in the future.
-//! This abstraction is not yet designed.
+// After you edit the crate's doc comment, run this command, then check README.md for any missing links
+// cargo rdme --workspace-project=vello_api
+
+// These docs are written for a more complete version of Vello API.
```
nicoburns, line 7: A lot of this documentation feels like it might be better suited to being in either:

- The PR description
- A dedicated .md file design document

with rustdoc reserved for user-facing documentation.

(I certainly don't think we could ship it in this state)

sparse_strips/vello_api/src/lib.rs
```diff
@@ -1,9 +1,140 @@
...
+//!
+//! There are some features which are only implemented in one of our target renderers, so cannot yet be included in the generalised API.
+//! As an interim solution, we intend to expose these features as renderer specific extensions (see the "excluded for expedience section").
+//!
+//! For Vello CPU, these are (i.e. Vello Hybrid does not implement these):
+//!
+//! - Masks
+//! - Filter effects
+//! - Non-isolated blending (this is "supported" by Vello Hybrid, but currently silently ignored)
+//! - Blurred rounded rectangles (note that currently this is actually included in the abstraction, despite this status)
```
nicoburns, line 115: IMO all of these ought to be in the abstraction, with a feature matrix for which backends implement which features in the docs. And ideally also with some kind of runtime feature detection mechanism that allows for consuming code to branch on features implemented (features would be hardcoded into backends, but code consuming the abstraction could read them as runtime flags).

(a lot of my reasoning here is that I would expect all backends to eventually support all of these features)

sparse_strips/vello_api/src/texture.rs
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Types and utilities for textures, which are 2d images owned by a renderer.
+
+/// A renderer-specific ID for a texture.
+///
+/// Note that we don't currently expect these to be constructed,
+/// because the.
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TextureId(u64);
```
nicoburns, line 11: FYI, I much prefer this being a `u64` newtype rather than reference counted. When previously discussing reference counting I thought it would be an internal implementation detail of a rendering backend to prevent a resource that is part of an in-flight render from being deallocated (even if the user explicitly calls "free").

(some kind of generational scheme may also work for this)

sparse_strips/vello_api/src/scene.rs
DJMcNab: It's sort of the opposite, a future-dated comment; essentially, this is looking forward to a future where we have proper handles for textures. I'll change it

sparse_strips/vello_api/src/painter.rs
DJMcNab: This is to match the existing Vello CPU/Vello Hybrid API. I agree that the semantics are unclear here, but I find it hard to argue that the time to solve that is now...

sparse_strips/vello_api/src/painter.rs
DJMcNab: This is largely copying Vello Classic's existing API. I'm also not aware of having run into any cases where such a conversion has been necessary with Vello Classic's API. We can always remove it as a follow-up.

> and you also can't turbofish `Into::into`

For the sake of pedantry: this isn't true - you can use `Into::<Whatever>::into`. But I agree that it isn't very ergonomic.

sparse_strips/vello_api/src/lib.rs
DJMcNab: I've removed the recommendation, whilst keeping the spirit of this section.

sparse_strips/vello_api/src/painter.rs
nicoburns: Ah I see. AnyRender copied Vello Classic which doesn't use `Option`. Either way is probably fine for now in any case.

sparse_strips/vello_api/src/painter.rs
DJMcNab: 1) Something like `impl<P: PaintScene> PaintScene for &mut P` is probably a trap; almost every API should be accepting an `&mut impl PaintScene` anyway. That's not a reason in itself to disallow inner references, however.
2) Because of `Scene`, I agree that downcasting this type directly as a user doesn't actually make any sense.
3) You do however need to have some way to downcast the "final" scene. My vision for this would be custom brush/command types which the "final" implementation of `PaintScene` (i.e. `CPUScenePainter` or w/e) calls with a type-erased version of itself. However, for that to work, those end-types pretty much do still need to be downcastable. I presume you're imaging some near-equivalent to that, but using GATs?

As I believe we will need these to be downcastable at some stage, I'm inclined to leave this for this PR

sparse_strips/vello_api/src/painter.rs
DJMcNab: The docs in
https://github.com/DJMcNab/vello/blob/3f6179d41fd8c4724a9557ab81feb6b6962d4573/sparse_strips/vello_api/src/painter.rs#L28-L37

already give my reasoning; this is something we know needs to be redesigned, but it is what I already had implemented, and we just need to get something landed. I don't imagine this aspect survives to even the "v0.0.1" equivalent of Vello API.

sparse_strips/vello_api/src/scene.rs
DJMcNab: We could, and there are reasons to think it could be nicer. Either way, I'm not going to change that in this PR.

(Fwiw, these structures are taken pretty much as-is from Vello Common's recording types)

sparse_strips/vello_api/src/scene.rs
DJMcNab: I believe this is a property of an entire Scene, because:
1) A scene being hinted is an (enforced) guide that you can't rescale/rotate/etc. it.
2) There's no way to subdivide a scene (e.g. to split out the parts which are hinted for separate processing)
3) I believe that implicitly making a Scene be hinted by trying to include hinted content into it would be a really error-prone design to use, as the failure wouldn't be localised to the code which was "wrong".

sparse_strips/vello_api/src/lib.rs
DJMcNab: The second sentence of this section states:

> As an interim solution

That is, this is explicitly planned to be made part of the core abstraction. But for expedience of getting something which is actually landable, they're excluded. I find it quite hard to convince myself that blocking this PR on designing a workable abstraction for masks, filter effects and non-isolated blending is a good use of time.

Fwiw, I did previously have these as a commented out part of the abstraction with notes to say "TODO: Design", but Olivier insisted those comments be removed, and so adding them here was my way of ensuring they weren't lost.

nicoburns: Totally agree that this PR shouldn't block on those

sparse_strips/vello_api/src/lib.rs
DJMcNab: The story here isn't at all clear from my perspective, and is changing as we get image atlasing and path caching actually working. To start off with, I'm not going to add text in this PR, again mostly for ease of review.
A few notes:
1) I know that we do have a Parley Draw crate already in the Parley repo. I haven't looked into what form that has taken due to a lack of time, but I presume you have more context to know that isn't what we have here.
2) The code to build text support on top of the API in this crate is not an awful lot of code (if we ignore bitmap glyphs). As such, my short-term proposal is to implement a naïve, cacheless "Parley Draw" in this repo as something like "Vello Interim Text".
3) Adding glyph support in the form you propose to this version of Vello API is *not* the path I'd recommend us taking. Essentially, I don't think it's actually any quicker than getting "Vello Interim Text" working.

Essentially, all of this needs more conversation to actually work out what problem Parley Draw is solving/will be solving. That isn't for this PR. I realise that "the long-term story should become clearer in a couple of months" isn't a useful message for consumers. The timeline being so long is an unfortunate side effect of the timing of my changing role; I'm having to focus my remaining time on this contract carefully, and Parley Draw is one of the things falling off as a result.

sparse_strips/vello_api/src/lib.rs
DJMcNab: I don't have the mental bandwidth to rewrite these docs at the moment. That's especially true when I believe this is likely to further change again, around textures being added. I agree that it's important, but I really would like to avoid blocking this PR on it. See also below.

> I certainly don't think we could ship it in this state

From my perspective, this PR is being submitted as a base for further collaboration and scoped review, rather than as something to be shipped as-is. It definitely needs a lot more follow-up before release.

sparse_strips/vello_api/src/texture.rs
DJMcNab: To clarify, I still think that for applications (i.e. not web browsers and similar which are already managing image lifetimes), the place in the trade-off space which maximises usability is to use reference counted handles for images. And that doesn't mean that you can't have manually managed resources (indeed, #1299 did allow that); it just means that you have to be careful with their lifetimes if you do. I do find it hard to imagine wanting to make a second tier of reference counting for in-flight renders, if we'll already have one for "not in-flight" renders.

But that doesn't need to be tied into the initial PR for getting this landed.

sparse_strips/vello_api/src/painter.rs
nicoburns: > You do however need to have some way to downcast the "final" scene.

The trick I've been using with AnyRender is for calling code to create the underlying type (`e.g. `vello::Scene` or `vello_cpu::RenderContext`) and wrap it in the `impl PaintScene`. Then, as the `impl PaintScene` only borrows the underlying type, there is nothing to be done to get the underyling type back. Simply drop the `impl PaintScene` and you can use the wrapped object again.

sparse_strips/vello_api/src/texture.rs
nicoburns: My concern here is that it's going to be hard to type-erase anything which can't be coerced into a primitive type. Which will make it difficult to to wrap this API in a more general purpose one that covers renderers other than Vello (and just generally leaves you dealing with an associated type rather than a concrete one).

sparse_strips/vello_api/src/scene.rs
nicoburns: Ah ok, so "hinted: true" really means "maybe hinted". And `hinted: false` means "definitely not hinted.

(including non-hinted content into a hinted scene must presumably be valid? Certainly I can imagine wanting to render an animated SVG (unhinted) along with non-animated text (hinted) into the same overall scene)

sparse_strips/vello_api/src/lib.rs
nicoburns: Looks like we currently have a copy-paste of the Vello code in Parley Draw. I don't think anyone has actually used this code in anger.

---
I'm happy to go down the "get Parley Draw working route". We should just bear in mind that this won't be very useful until we have that.

---
It occurs to me that taking "render glyphs" out of the abstraction will make it much easier to implement rendering backends (e.g. `iced_wgpu`, `tiny-skia` implement everything except glyph rasterization), so that's a nice benefit of this approach.

nicoburns (approved):
I have lots of nits, but the overall direction seems about right to me. So happy for this to be merged and then iterated on.

nicoburns:
I also think it may be worth reconsidering whether to merge with AnyRender at some point, given the now very similar designs. I certainly plan to import the ideas that seem good to me (notably the `Scene` type) into AnyRender.

sparse_strips/vello_api/src/lib.rs
taj-p: For context, we're investing heavily in Parley Draw. @grebmeg is adding glyph caching to Parley Draw and @valadaptive has been implementing a bunch of fixes, performance improvements, refactoring Parley tests to use Parley Draw for rendering, and underline support.

The API _can_ be used today for Vello CPU. Canva will be migrating our text rendering to use Parley Draw in the coming 1 to 3 weeks. 

I see Parley Draw's API evolving a number of ways. As it pertains to Vello API, it is likely possible to implement a Vello API backend for Parley Draw. But, we also need to consult with other owners on these topics (you two included).

But, there are some design and implementation decisions we need to discuss. Alex's ongoing and impactful work into glyph caching has surfaced some questions we will want to answer - for example, supporting sampling from alpha textures and performantly applying paint to those alphas.  

I am also in no doubt that Valadaptive's ongoing work will surface even more questions that need to be answered.

All said, the API is expected to undergo significant iteration. So, consumers beware! 😆 ⚠️

sparse_strips/vello_cpu/src/api.rs
```diff
@@ -0,0 +1,224 @@
...
+                RenderCommand::DrawPath(affine, path_id) => {
+                    self.render_context.set_transform(scene_transform * *affine);
+                    let path = &input_paths.meta[usize::try_from(path_id.0).unwrap()];
+                    let path_end = &input_paths
+                        .meta
+                        .get(usize::try_from(path_id.0).unwrap() + 1)
+                        .map_or(input_paths.elements.len(), |it| it.start_index);
+                    let segments = &input_paths.elements[path.start_index..*path_end];
+                    // Obviously, ideally we'd not be allocating here. This is forced by the current public API of Vello CPU.
+                    let bezpath = BezPath::from_iter(segments.iter().cloned());
```
taj-p, line 65: Is it not possible to store a re-useable `BezPath` on `CPUScenePainter` so we don't need to pay this allocation (ditto below)?

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -0,0 +1,201 @@
...
+                RenderCommand::DrawPath(affine, path_id) => {
+                    self.scene.set_transform(scene_transform * *affine);
+                    let path = &input_paths.meta[usize::try_from(path_id.0).unwrap()];
+                    let path_end = &input_paths
+                        .meta
+                        .get(usize::try_from(path_id.0).unwrap() + 1)
+                        .map_or(input_paths.elements.len(), |it| it.start_index);
+                    let segments = &input_paths.elements[path.start_index..*path_end];
+                    // Obviously, ideally we'd not be allocating here. This is forced by the current public API of Vello CPU.
+                    let bezpath = BezPath::from_iter(segments.iter().cloned());
```
taj-p, line 64: As above

sparse_strips/vello_cpu/src/api.rs
DJMcNab: So we certainly could do that. The reason I don't in this PR is that I believe that would be a trap, because there's no semantic difference between a reference into a `BezPath` owned by the `CPUScenePainter` and a reference into the `PathSet`.

That is to say, the correct way to fix this is to change the internal APIs to work with a slice of `PathEl`, rather than papering over it with amortised allocations. That's one of the things on the task list before any "serious" release of this code.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,251 @@
...
+use core::any::Any;
+
+use peniko::kurbo::{Affine, BezPath, Rect, Shape, Stroke};
+use peniko::{BlendMode, Brush, Color, Fill, ImageBrush};
+
+use crate::scene::Scene;
+use crate::texture::TextureId;
+
+/// The brush type used for most painting operations in Vello API.
+pub type StandardBrush = Brush<ImageBrush<TextureId>>;
```
grebmeg, line 13: [nit]: I’m not sure I understand why it’s called **Standard**. I actually liked the previous **Paint** and had gotten quite used to it, but I guess “brush” is the vocabulary we generally use?

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// For solid colors, this has no effect, and so you may prefer [`set_solid_brush`](PaintScene::set_solid_brush).
+    fn set_brush(
+        &mut self,
+        brush: impl Into<StandardBrush>,
+        // We'd like to support both brushes which are "object-local" and brushes which are "scene-local".
+        // Image for example you want a gradient to be applied over a whole paragraph of text.
+        // The naive solution would need to apply the gradient with a paint transform which undoes the
+        // object's transform. That's clearly pretty poor, and we can do better.
+        // However, this isn't exposed in the current Vellos, so we're choosing to defer this.
+        // transform: Affine,
```
grebmeg, line 112: Should we use `transform` instead of `paint_transform`? ditto below where `paint_transform` used.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,251 @@
...
+        paint_transform: Affine,
+    );
+
+    /// Set the current brush to represent a rounded rectangle blurred with an approximate gaussian filter.
+    ///
+    /// **This method is currently unimplemented on all backends, so should not be used.**
```
grebmeg, line 118: Would it be better to exclude this method entirely, since it isn’t implemented in both renderers?

sparse_strips/vello_api/src/scene.rs
grebmeg: > (including non-hinted content into a hinted scene must presumably be valid? Certainly I can imagine wanting to render an animated SVG (unhinted) along with non-animated text (hinted) into the same overall scene)

I agree with this, and I can also imagine such a case. Would it be simpler to omit this for now to simplify the unified API?

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,298 @@
...
+    fn append(
+        &mut self,
+        mut scene_transform: Affine,
+        Self {
+            // Make sure we consider all the fields of Scene by destructuring
+            paths: other_paths,
+            commands: other_commands,
+            hinted: other_hinted,
+        }: &Scene,
+    ) -> Result<(), ()> {
```
grebmeg, line 180: I might be mistaken, but I was a little surprised to see this, as I understood it to be a new feature related to merging scenes. Could you please clarify the use cases?

sparse_strips/vello_hybrid/src/api.rs
grebmeg, line 1: I think about 90% of the API implementations (vello_cpu & vallo_hybrid are the same). It seems currently unavoidable because of how `PaintScene` is implemented, right? I’m just curious whether there is a way around this, and whether it would make sense to make it generic. If the implementations are so similar, could both renderers share a single default implementation?

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,298 @@
...
+/// This method assumes that the translations will be at a reasonable scale for 2d rendering.
+/// Whilst we don't validate this precisely, values less than ~10 billion pixels should be fine.
+pub fn extract_integer_translation(transform: Affine) -> Option<(f64, f64)> {
+    fn is_nearly(a: f64, b: f64) -> bool {
+        // TODO: This is a very arbitrary threshold.
+        (a - b).abs() < 0.01
+    }
+    let [a, b, c, d, dx, dy] = transform.as_coeffs();
+    // If there's a skew, rotation or scale, then the transform is not compatible with hinting.
+    if !(is_nearly(a, 1.0) && is_nearly(b, 0.0) && is_nearly(c, 1.0) && is_nearly(d, 0.0)) {
```
grebmeg, line 156: Should it be?

```suggestion
    if !(is_nearly(a, 1.0) && is_nearly(b, 0.0) && is_nearly(c, 0.0) && is_nearly(d, 1.0)) {
```

sparse_strips/vello_cpu/src/api.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+        self.render_context.set_transform(transform);
+        self.render_context.set_fill_rule(fill_rule);
+        // TODO: Tweak inner `fill_path` API to either take a `Shape` or an &[PathEl]
+        self.render_context.fill_path(&path.to_path(0.1));
```
grebmeg, line 133: Would it be possible to apply the tolerance when preparing the shape, or is the tolerance supposed to change depending on the transform? Of course, this would require changing the API to use slice.

sparse_strips/vello_cpu/src/api.rs
```diff
@@ -0,0 +1,224 @@
...
+                        "Vello CPU doesn't expose drawing a blurred rounded rectangle in custom shapes yet."
+                    )
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
```
grebmeg, line 129: `impl Shape` is used in multiple places, would this increase binary size due to monomorphized functions? Would it be better to stay with `BezPath` instead?

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    fn push_layer(
+        &mut self,
+        clip_transform: Affine,
+        clip_path: Option<impl Shape>,
+        blend_mode: Option<BlendMode>,
+        opacity: Option<f32>,
+        // mask: Option<Mask>,
+    ) {
+        self.scene.set_transform(clip_transform);
```
grebmeg, line 180: In the CPU code, we use next before this line, should it appear here as well?

```rust
self.render_context.set_fill_rule(Fill::NonZero);
```

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// Layers with an opacity of zero may be optimised out, but this is not currently guaranteed.
+    ///
+    /// **However, the transforms are *not* saved or modified by the layer stack.**
+    /// That is, the `transform` argument to this function only applies a transform to the `clip` shape.
+    fn push_layer(
+        &mut self,
+        clip_transform: Affine,
+        clip_path: Option<impl Shape>,
+        blend_mode: Option<BlendMode>,
+        opacity: Option<f32>,
```
grebmeg, line 204: I guess the filters are deliberately omitted here?

sparse_strips/vello_cpu/src/api.rs
laurenz-canva: This function is very short, so I think this should be very negligible.

nicoburns: I'm not a huge fan of the `impl Shape` either. Having said that, there are potentially algorithmic-level improvements available from `impl Shape` (e.g. downcasting to Rect and sending Rect down a fast path).

DJMcNab:
> Additionally, it might be worth adding exp or experimental prefixes to the `api.rs` files and possibly to the `vello_api` crate itself, just to make it completely clear that this is experimental.

I've added some docs for this in #1372. It's plausible that an `experimental_api` name for the modules would also make sense, but I'm not planning to make that PR. It's a question we need to ask when we're looking at doing another release, probably.

> Right now, we seem to be missing an example of a "real" application that demonstrates how the API is used in practice.

On a small-scale, #1369 is trying to provide that.

> Such an example would also help clarify how we envision apps being built and how `vello_api` is meant to be consumed.

I agree that we need this, although my thinking is that we need to get it to a more stable position first (e.g. changing the brush model).

sparse_strips/vello_api/src/painter.rs
DJMcNab: "Standard" in this case is referring to the fact that it's a non-specialised brush; that is, it's a brush from the standard imaging model (so solid colour, gradient or image). The alternative is the specialised brushes, which are currently only blurred rounded rectangles.

The name of Brush/Paint is from my experience having come from the Vello Classic side. I'm happy for us to change it to paint. I think perhaps I missed something; if the "paint" is what is drawn, but the brush describes the paint. So the metaphor is coherent that you call `set_paint` with a brush.

sparse_strips/vello_api/src/painter.rs
DJMcNab: Unfortunately, I'm not sure what alternative you're proposing here. Would you mind expanding this question?

Ultimately, this block going to have to be part of my upcoming work to reconsider how brushes/paint styles work.

sparse_strips/vello_api/src/painter.rs
DJMcNab: There are two arguments you could be making, one for which I'm loosely in favour, and one which I'm against.
- I'm against removing *just* this method, because I don't think that `fill_blurred_rounded_rect` is the right API for blurred rounded rectangles. It is important to be able to separate what the paint type is from what area is painted; the currently supported `fill_blurred_rounded_rect` can only fill the full rectangle, with no cutouts.
- However, the alternative reading is for removing the blurred rounded rectangle feature from Vello API on an interim basis, until we can support `set_blurred_rounded_rect_brush` (and also `fill_blurred_rounded_rect` in Vello Hybrid, see #1373). I'm loosely in favour of this, with a few caveats. The main reason I'd like to keep this in the API, is to have an example of the "specialised paint" I mentioned in my other comment, which is useful for design work.

I also expect blurred rounded rectangles to be part of the final imaging model for this crate, which is why they're in here to start with.

sparse_strips/vello_api/src/scene.rs
DJMcNab: As far as I'm aware, that case is already valid. It is correct that this flag that a `Scene` is hinted was added without much discussion (although it's also not much code), and probably should have been a follow-up. I still think there is value in it, but that tearing it out might be best.

sparse_strips/vello_api/src/scene.rs
DJMcNab: Once you have the `Scene` type, this is a pretty natural extension, I think? Use cases would include having e.g. a widget in a GUI toolkit, which can contain a translatable/scalable child SVG. The widget would be provided with a `Scene` to render into which is owned by the toolkit, and it itself would own the `Scene` of the SVG, which is a more efficient representation for painting (especially with cached paths).

sparse_strips/vello_hybrid/src/api.rs
DJMcNab: I don't know of a way to achieve what you want. Ultimately, this similarity comes from adapting the existing public API to this new (more) stateless API. My vague plan would be for us to transition to a state where `PaintScene` was the actually implemented API for each kind of scene, but that's a much more invasive change, and definitely wouldn't have been right to try and do in this PR.

Edit: In the current `PaintScene::append`, I believe this is possible, but I think it might fall down for a future scenario where we have custom paint commands. I'll think about it a bit more.

sparse_strips/vello_api/src/scene.rs
DJMcNab: Yes! I'll fix this tomorrow. Edit: https://github.com/linebender/vello/pull/1375

sparse_strips/vello_cpu/src/api.rs
DJMcNab: This is nearly the same thing as [#vello > Determining correct &#96;Shape&#96; tolerance](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Determining.20correct.20.60Shape.60.20tolerance/with/567789803), except that we do actually know the final transform here.

sparse_strips/vello_cpu/src/api.rs
taj-p: > one of the things on the task list

Is this task list captured somewhere or captured in a `TODO`? I worry that without documentation we'll forget this very important optimisation.

sparse_strips/vello_api/README.md
```diff
@@ -22,11 +22,129 @@ Full documentation at https://github.com/orium/cargo-rdme -->
...
+
+- Rendering directly to a surface.
+- Importing "external" textures (e.g. from a `wgpu::Texture`)
+
+## Excluded for expedience
+
+- Renderer specific painting commands (i.e. using downcasting).
+  This is intended to be an immediate follow-up to the MVP landing.
+- Pushing/popping clip paths (i.e. non-isolated clipping).
+  This feature should be easy to restore, although it isn't clear how it will work with "Vello GPU", i.e. Hybrid with GPU rasterisation
```
taj-p, line 113:
```suggestion
  This feature should be easy to restore, although it isn't clear how it will work with "Vello GPU", i.e. Hybrid with GPU sparse strip rendering
```

sparse_strips/vello_api/README.md
taj-p, line 1: Nice!

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+        self.scene.set_transform(transform);
+        self.scene.set_fill_rule(fill_rule);
+        // TODO: Tweak inner `fill_path` API to either take a `Shape` or an &[PathEl]
+        self.scene.fill_path(&path.to_path(0.1));
```
taj-p, line 131: Could we consume the path via `into_path` to save an allocation for BezPath?

IIUC, whatever we do, we're cloning here for borrowed data. But, if you pass an owned `BezPath`, we can at least save one allocation.

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+        self.scene.set_transform(transform);
+        self.scene.set_fill_rule(fill_rule);
+        // TODO: Tweak inner `fill_path` API to either take a `Shape` or an &[PathEl]
```
taj-p, line 130:
```suggestion
        // TODO: Tweak inner `fill_path` API to take a `PathEl` iterator
```

I think we can pass an iterator through instead. Single threaded and hybrid can operate over the iterator directly (`flatten::fill` takes an iterator) and multi threaded CPU could collect into its allocation group directly (with some tweaks to allocation group)

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// Both `path` and the current brush will be transformed using the 2d affine `transform`.
+    /// For self-intersecting or nested paths, the areas which are treated as the interior
+    /// will be determined using the given `fill_rule`.
+    /// See the documentation on [`Fill`]'s variants for details on when you might choose a given
+    /// fill rule.
+    ///
+    /// If `path` is not a closed path, a fallback straight-line closing segment may be appended.
+    /// (TODO: Should we give a warning of some kind as well?)
+    // It would be really nice to have images of nested paths here, to explain winding numbers.
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape);
```
taj-p, line 90: Do we need the "_path" suffix on `fill_path` and `stroke_path`?

If we're taking a `impl Shape`, the closer analog is `fillShape` / `strokeShape`. 

---

When I did a bit of research:
- Skia does `drawFoo` where `Foo` is like Rect | Path | etc.
- Canvas 2D is `fill()` and `stroke()`
- Cairo is `stroke` and `fill` (albeit with their `cairo_` prefix)

So, `fill` and `stroke` might better reflect Canvas 2D and Cairo. I'm also unsure whether "_path" is doing much here in assisting understanding of what is going on.

taj-p: To be honest, I also see the appeal of Skia's approach to have dedicated APIs for different shapes which may offer varying fast paths - e.g. skipping flattening for axis aligned rects or similar.

Maybe an additive change we can discuss another time.

sparse_strips/vello_api/src/painter.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// - `scene` is hinted, but `self` cannot guarantee that the hinting properties would be maintained.
+    ///   Hinted scenes can be appended to the final, renderer-specific, implementations of `PaintScene`
+    ///   or to other hinted `Scene`s.
+    /// - `scene` is hinted, but the transform would remove the hinting property.
+    ///   For hinted scenes, the only valid transforms are integer translations.
+    /// - `scene` does not apply to the same renderer as `self`.
+    /// - `scene` has unbalanced layers (TODO: This isn't implemented yet).
+    fn append(&mut self, transform: Affine, scene: &Scene) -> Result<(), ()>;
+
+    /// Fill the interior of the closed path described by `path` with the current brush.
```
taj-p, line 79: The documentation for `fill_path` seems inconsistent:

- Line 79 says "the closed path" (singular), but line 82 mentions "nested paths" (plural)
- The parameter is impl Shape, but the docs consistently say "path"
- Line 87 says "If path is not a closed path..." but with multiple subpaths, each would need its own implicit closing

Something like this might be clearer:

```
/// Fill the interior of `shape` with the current brush.
///
/// The shape may contain multiple subpaths (e.g., for shapes with holes like a donut).
/// Each subpath is implicitly closed if needed with a straight line.
/// For self-intersecting or nested subpaths, the `fill_rule` determines which areas are considered interior.
///
/// Both `shape` and the current brush will be transformed using the 2d affine `transform`.
/// See the documentation on [`Fill`]'s variants for details on when you might choose a given fill rule.
````

File: sparse_strips/vello_api/src/painter.rs [hunk 30]
```diff
@@ -0,0 +1,251 @@
...
+/// [`fill_path`]: PaintScene::fill_path
+/// [`append`]: PaintScene::append
+/// [`Renderer`]: todo
+pub trait PaintScene: Any {
+    // TODO: This also "clobbers" the brush; we need to document that.
+    /// Insert the contents of `Scene` into the drawing sequence at this point, with the 2d affine
+    /// `transform` applied to its contents.
+    ///
+    /// This will change the currently active brush, so you must reset the brush after calling this method.
+    /// (TODO: Change the brush design, because that's stupid!)
```
taj-p, line 65: 😆

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,298 @@
...
+/// This hinting property allows improved performance through caching
+/// of intermediate rendering stages.
+///
+/// However, this does mean that more advanced transformations are not possible on hinted scenes.
+///
+/// # Usage
+///
+/// This type has public fields as an interim basis.
+/// To use this type, you should use the methods on its implementation of [`PaintScene`] instead.
+#[derive(Debug)]
```
taj-p, line 100: Should `Scene` be `Clone`?

sparse_strips/vello_api/src/scene.rs
```diff
@@ -0,0 +1,298 @@
...
+            }));
+
+        Ok(())
+    }
+
+    fn fill_path(&mut self, transform: Affine, fill_rule: peniko::Fill, path: impl kurbo::Shape) {
+        let idx = self.paths.prepare_shape(&path, fill_rule);
+        self.commands.push(RenderCommand::DrawPath(transform, idx));
+    }
+    fn stroke_path(
```
taj-p, line 235:
```suggestion

    fn stroke_path(
```

sparse_strips/vello_api/src/paths.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// See the docs on [`append`](PathSet::append) for how this changes when path sets are combined.
+    ///
+    /// This method is generally only expected to be used by [`Scene`](crate::Scene).
+    pub fn prepare_shape(&mut self, shape: &impl Shape, style: impl Into<Style>) -> PathId {
+        let start_index = self.elements.len();
+        // TODO: We hard-code this tolerance to be 0.1, as every other call does so.
+        // We should maybe change that at some point?
+        // https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Determining.20correct.20.60Shape.60.20tolerance/with/565793178
+        // If you need a different tolerance, you should pass in a `BezPath` (i.e. using Shape::to_path with the tolerance you require)
+        self.elements.extend(shape.path_elements(0.1));
```
taj-p, line 96: But, what if you need to support much more than 10x zoom? For example, IIUC, won't the 23,000% zoom in graphite below produce a ~23 pixel error using this tolerance?

Consider:
- [Affinity's renderer](https://www.youtube.com/watch?v=QzxG-ebXx54&t=775s); or
- Zooming 23,000% in Graphite (below video); or
- Building any form of whiteboard / design product; or
- Building some form of 2D game with dynamically scaling scenes

https://github.com/user-attachments/assets/43502349-3037-424a-8844-df5984f61558

The solution in the comment to convert to `BezPath` first with a lower tolerance seems surprising and confusing for a consumer - it's a footgun. They could use a very low tolerance, but this creates many segments and suffers a performance penalty even if the user never scales the scene. 

So, to summarise:

`path_elements()` is called at scene construction time, but the final transform isn't known until render time (or later, when the scene is appended). But, if you ever need to later apply a transform to the scene greater than 10x, you should know ahead of time what tolerance you require, and pay the performance cost of lower tolerance. 

So, there is a tension here. Scene wants to be resolution-independent and reuseable at any scale for best ergonomics. But, `path_elements()` and the like bakes in resolution-dependent implementation. This seems contradictory. 

For SVG, this means that `<circle>` and `<ellipse>` are subject to this tension (assuming the parser doesn't convert them to paths before reaching Vello). 

I think we should accept `BezPath` (or better an iterator of `PathEl`) into the API until we have an answer to this problem because otherwise our internal implementation can produce incorrect renderings for expected and common use cases. 

By passing in `BezPath` (or `PathEl` iterator) instead of `impl Shape`, we force the consumer to think about the issue and push the problem out until we are ready to properly tackle it. Yes, this is less ergonomic but is more correct.

This approach may also better reflect our current state where path caching isn't yet a known dimension.

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -0,0 +1,201 @@
...
+        &mut self,
+        clip_transform: Affine,
+        clip_path: Option<impl Shape>,
+        blend_mode: Option<BlendMode>,
+        opacity: Option<f32>,
+        // mask: Option<Mask>,
+    ) {
+        self.scene.set_transform(clip_transform);
+        self.scene.push_layer(
+            clip_path.map(|it| it.to_path(0.1)).as_ref(),
```
taj-p, line 182:
```suggestion
            clip_path.map(|it| it.into_path(0.1)).as_ref(),
```

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -0,0 +1,201 @@
...
+            None,
+            None,
+        );
+    }
+
+    fn push_clip_layer(&mut self, clip_transform: Affine, path: impl Shape) {
+        self.scene.set_transform(clip_transform);
+        self.scene.push_clip_layer(
+            // TODO: Not allocate
+            &path.to_path(0.1),
```
taj-p, line 194:
```suggestion
            &path.into_path(0.1),
```

sparse_strips/vello_api/src/paths.rs
laurenz-canva: Maybe I'm missing something, but couldn't we change vello_cpu/vello_hybrid to take an `impl Shape` as well, and we pass it all the way down to flattening, where the tolerance is multiplied by the scale/skew factor in each direction, as it's currently done for the stroke tolerance?

sparse_strips/vello_cpu/src/api.rs
DJMcNab: > `impl Shape` is used in multiple places, would this increase binary size due to monomorphized functions? Would it be better to stay with `BezPath` instead?

The problem with `&BezPath` is that the ergonomics are bad (and you also end up doing unwanted allocations). Taj has suggested taking in a pathsegment iterator, which is a much better fit. Follow the discussion in other threads for more.

sparse_strips/vello_hybrid/src/api.rs
DJMcNab: Yes! Thanks. Edit: https://github.com/linebender/vello/pull/1375

sparse_strips/vello_api/README.md
DJMcNab: Do we have dictionary somewhere of what these terms mean in the pipeline? I was treating "rasterisation" as being the "convert flattened lines to alphas". However, I see that this terminology is more consistent with Raph's RustWeek slides, so I'll change it in my follow-up (edit: https://github.com/linebender/vello/pull/1375)

sparse_strips/vello_api/src/painter.rs
DJMcNab: This naming is all open to change. Part of the reasoning was wanting to be flexible with future explicit path caching, but I do think I agree that plain `fill` and `stroke` are probably nicer for now

sparse_strips/vello_api/src/scene.rs
DJMcNab: So, yes and no. In the current state, I believe it could be, and I didn't add it only because I didn't think about this. That being said, the reason I want to avoid it is for future "custom operations".

So for example, in Vello CPU, we can easily set a per-path blend mode, which isn't possible (in general at least) for Vello Hybrid. We want some way to represent that (and similar features), and that will end up looking like `Box<dyn SomeTrait>`. That type obviously isn't cloneable in itself.
That being said, it is possible to make it cloneable, and there are strong arguments for doing so.

sparse_strips/vello_api/src/painter.rs
DJMcNab: I found it very difficult to get a firm handle on a consistent terminology here. It's the distinction between a shape and its interior which I was lacking the proper word for.

Your proposed doc is so much better, thanks! I'm applying this in https://github.com/linebender/vello/pull/1375 - the "various easy fixes" follow-up.

sparse_strips/vello_hybrid/src/api.rs
DJMcNab: Good catch. Applied in https://github.com/linebender/vello/pull/1375

(noting of course that the API will change a bit later)

sparse_strips/vello_api/src/scene.rs
DJMcNab: Done in https://github.com/linebender/vello/pull/1375

sparse_strips/vello_cpu/src/api.rs
DJMcNab: I've made these an explicit TODO in https://github.com/linebender/vello/pull/1375

sparse_strips/vello_hybrid/src/api.rs
DJMcNab: See https://github.com/linebender/vello/pull/1360#discussion_r2711576115

sparse_strips/vello_api/src/painter.rs
DJMcNab: Yeah, that's correct. I did originally have a TODO here, but Olivier asked for it to be moved. This is now captured as https://github.com/linebender/vello/blob/0718844d2c76426edc5667fbbd3d4cbf8a1deeec/sparse_strips/vello_api/src/lib.rs#L116

sparse_strips/vello_api/src/paths.rs
nicoburns: Are there concerns with making Shape -> BezPath conversion single-threaded instead of multi-threaded here? I have similar (potential) concerns around some of the Parley Draw API design...

sparse_strips/vello_hybrid/src/api.rs
DJMcNab: I agree, but I've been avoiding making breaking changes to the existing APIs in both Vellos CPU and Hybrid, especially in this PR.

sparse_strips/vello_api/src/paths.rs
DJMcNab: I did discuss this before making this decision, in [#vello > Determining correct &#96;Shape&#96; tolerance](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Determining.20correct.20.60Shape.60.20tolerance/with/567789803), although clearly that thread didn't get enough attention. I've actioned the proposal we discussed in that thread in https://github.com/linebender/vello/pull/1376. That doesn't allow arbitrary path element iterators (probably we should have a suitable wrapper type), but it does maintain ergonomics in the common case.

> Maybe I'm missing something, but couldn't we change `vello_cpu/vello_hybrid` to take an `impl Shape` as well, and we pass it all the way down to flattening, where the tolerance is multiplied by the scale/skew factor in each direction, as it's currently done for the stroke tolerance?

I believe this would be possible, although it would unfortunately require.

> Are there concerns with making Shape -> BezPath conversion single-threaded instead of multi-threaded here? I have similar (potential) concerns around some of the Parley Draw API design...

If you can make measurements showing it's an issue, we can explore different designs. But I personally find it hard to think it would be a problem.

DJMcNab: That affinity demo is really impressive!

DJMcNab:
I think that the follow-ups of #1375 and #1376 address all of the review concerns raised, and so conversation on their respective parts should continue in those PRs. Let me know if I've missed any comment; it's not intentional, it would just be because of GitHub's UI not making it easy to keep track of these things.

sparse_strips/vello_api/src/paths.rs
taj-p: > Maybe I'm missing something, but couldn't we change vello_cpu/vello_hybrid to take an `impl Shape` as well, and we pass it all the way down to flattening, where the tolerance is multiplied by the scale/skew factor in each direction, as it's currently done for the stroke tolerance?

Would that work for scenes though? At construction time, they do not know their ultimate transform; so we would need to store the traits on the scene until render time, which, I think, would be hard to do elegantly

DJMcNab: Yeah, that is probably what I was going to say in my sentence fragment (if I ever got around to finishing it!); that it would require some way to store `Shape`s generically, which doesn't currently exist, and would likely require boxing or unsafe shenanigans.

PR #1364, opened 2026-01-15
vello_cpu: Fix `clippy::cast_possible_wrap` in Gaussian blur

tomcur:
Fixed by adding a reasonable limitation on `MAX_KERNEL_SIZE`, and storing the kernel size as a `u8` instead of `usize`.

The lint itself checks for possible wrapping when going from an unsigned to signed integer, and is not currently in our lint set (hence not triggered on CI). It triggers when manually requested.

It probably should be included in the lint set, but that's for another time (https://github.com/linebender/linebender.github.io/pull/135).

### DIFF
--- a/sparse_strips/vello_cpu/src/filter/gaussian_blur.rs
+++ b/sparse_strips/vello_cpu/src/filter/gaussian_blur.rs
@@ -668,18 +668,18 @@ mod tests {
 
         // Kernel should be symmetric
         for i in 0..size / 2 {
-            assert!((kernel[i] - kernel[size - 1 - i]).abs() < 1e-6);
+            assert!((kernel[usize::from(i)] - kernel[usize::from(size - 1 - i)]).abs() < 1e-6);
         }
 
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/filter/gaussian_blur.rs
```diff
@@ -28,12 +28,23 @@ use vello_common::peniko::color::PremulRgba8;
...
+#[cfg(test)]
+const _: () = const {
+    if MAX_KERNEL_SIZE.is_multiple_of(2) {
+        panic!("`MAX_KERNEL_SIZE` must be odd");
+    }
+    if MAX_KERNEL_SIZE > u8::MAX as usize {
+        panic!("`MAX_KERNEL_SIZE` must be less than or equal to `u8::MAX`");
+    }
+};
+
```
tomcur, line 47: To ensure we don't accidentally break our own promise later!

PR #1368, opened 2026-01-20
Add early culling for lines fully left of the viewport.

b0nes164:
Apologies this took an embarrassingly long time due to a funny bug. Adds changes as discussed. 

UPDATED:
Lines left of the viewport are not visible but still contribute winding, creating a left to right dependency for rendering. This patch simplifies rendering in this scenario by culling the left-of-viewport portion of lines, and recording their equivalent winding contribution inside a histogram. If a line is *entirely* left of the viewport, it is completely culled and produces no tiles.

Correspondingly `strip.rs` has been updated to consume this histogram. Notably, `strip.rs` now handles cases where no tiles are present, but culling occured, and cases where winding is "captive" either between left edge of the viewport and the first tile in the row, or between two non-consecutive rows.

Removes the previous handling of left of edge lines in `strip.rs`.

No culling, Ghostscript_Tiger shifted 50% off screen:
```
tile_aaa_shift50/Ghostscript_Tiger
                        time:   [51.574 µs 51.650 µs 51.732 µs]
```

With culling:
```
tile_aaa_shift50_cull/Ghostscript_Tiger
                        time:   [46.319 µs 46.380 µs 46.436 µs]
```


tomcur (commented):
Very cool, and I think this is roughly what it should look like. I do have a few observations.

> Early culling is conditionally enabled based on whether a culling opportunity was detected in the previous strip generation pass.

I'm not sure `culling_opportunity` are the best semantics. A path having had a culling opportunity may not be a good predictor of culling being possible the next time `generate_with_clip` is called, and same if the path didn't have a culling opportunity, the next path may in actuality well have one. These semantics will of course mean that we're always going to miss the performance benefit of culling at least during the first path in a sequence of cullable paths.

Perhaps, instead we should be conditional only on whether we want culling at all -- for example, if we're caching strips, perhaps all types of culling should be disabled (including culling where we don't account for winding, i.e., above, to the right, and below the viewport). In that case it may still make sense sense to have the dispatch be const-generic over the culling, as you have now, but perhaps the difference in performance can be so minimal that we don't have to bother.

E.g., for tiling, with the current implementation it doesn't seem to matter if we always cull or not:

```
tile/Ghostscript_Tiger  time:       [184.92 µs 185.14 µs 185.41 µs]
tile/Ghostscript_Tiger-cull time:   [181.96 µs 182.34 µs 182.84 µs]
```

That of course doesn't preclude dispatching to a strip rendering implementation that does or does not use culling, depending on whether we've culled something. That may well be important for performance.

Should also benchmark strip rendering, but that requires some bigger changes to the benchmark code...

It does seem both tiling and strip rendering are about 6% slower with these changes, but there is a little bit of noise so that has to be measured a bit more carefully (and this can perhaps be optimized more).

> Currently this does not use the bitvec, and instead traverses the histogram.

We have to carefully check what the performance impact is of this where rendering a bunch of small geometry (like glyphs) which will usually just occupy a few rows.

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -69,11 +69,15 @@ pub struct StripGenerator {
     tiles: Tiles,
     width: u16,
     height: u16,
+    use_early_culling: bool,
+    partial_windings: Vec<[f32; Tile::WIDTH as usize]>,
```
tomcur:
```suggestion
    partial_windings: Vec<[f32; Tile::HEIGHT as usize]>,
```

sparse_strips/vello_common/src/tile.rs
```diff
@@ -299,15 +322,39 @@ impl Tiles {
...
-    // viewport are not culled yet. These lines impact winding, and would need forwarding of
-    // winding to the strip generation stage.
-    pub fn make_tiles_analytic_aa(&mut self, lines: &[Line], width: u16, height: u16) {
+    pub fn make_tiles_analytic_aa<const USE_EARLY_CULL: bool>(
+        &mut self,
+        level: Level,
+        lines: &[Line],
+        width: u16,
+        height: u16,
+        partial_windings: &mut [[f32; Tile::WIDTH as usize]],
```
tomcur:
```suggestion
        partial_windings: &mut [[f32; Tile::HEIGHT as usize]],
```

b0nes164:
>  A path having had a culling opportunity may not be a good predictor of culling being possible the next time generate_with_clip is called, and same if the path didn't have a culling opportunity, the next path may in actuality well have one. 

These are good points! For more context, we made the decision to make the culling conditional on the previous `make_tiles` call because we were concerned about the cost of clearing the histograms. However, now that we have the data, the histogram cost seems to be negligible.  But, I suspect the cost of traversing the rows inside strip.rs is not negligible.

So here's a thought: WDYT about culling always enabled in `make_tile` and then conditionally dispatching the culling logic in strips. I think that gives us the best of both? (and also only clearing the histograms if the previous `make_tile` culled?).

Bitvec to come!

tomcur:
> So here's a thought: WDYT about culling always enabled in `make_tile` and then conditionally dispatching the culling logic in strips. I think that gives us the best of both? (and also only clearing the histograms if the previous `make_tile` culled?).

Yes! In fact, I edited my message (probably as you were typing) to say this:

> That of course doesn't preclude dispatching to a strip rendering implementation that does or does not use culling, depending on whether we've culled something. That may well be important for performance.

b0nes164:
@LaurenzV do you know if it's safe to early cull when the storage generation mode is not `GenerationMode::Replace` ?

For safety, I make the assumption no, but after thinking about it, I think it's fine? Once the strips and alpha are generated (and cached), there's no need to hold onto the culled winding contributions (similarly to how tiles are also cleared regardless of the storage mode)?

b0nes164:
Otherwise, ready for review.

LaurenzV:
I don't believe it should make a difference for the generation mode, as long as the strips are there it should be fine!

b0nes164:
Early culling is now always enabled. I've retained the `USE_EARLY_CULL` generic, as it's useful to manually enable/disable culling for tests and benchmarking.

tomcur:
I have not forgotten about this, but haven't had a chance for a proper look yet!

b0nes164:
PTAL!

Note, the behavior is different than what I had before. Now, left edge crossing lines are culled (if possible) inside `tile.rs`, and the corresponding logic in `strip.rs` has been removed.

Re moving culling earlier into `flatten.rs`: while possible, it would be extremely awkward, as the code has no concept of the tile grid at that point. From both a maintainability and a performance perspective, I think (potentially) pushing a couple extra lines is better than dragging a ton of logic to handle the histogramming in `flatten.rs`.

Re using clip instead of viewport rect: I will add that as a follow on CL.

b0nes164:
Added more testing, some benchmarking, and fixed the issue identified with hayro.

LaurenzV:
PDF tests seem to pass now!

sparse_strips/vello_common/src/tile.rs
```diff
@@ -218,6 +221,14 @@ impl Tile {
         // the in-memory representation.
         ((self.y as u64) << 48) | ((self.x as u64) << 32) | self.packed_winding_line_idx as u64
     }
+
+    /// Whether the tile is the "SENTINEL"
+    ///
+    /// An organic tile cannot have this coordinate because of the division by tile size on creation.
```
LaurenzV, line 227:
```suggestion
    /// Whether the tile is the sentinel tile
    //
    // An organic tile cannot have this coordinate because of the division by tile size on creation.
```

I'd put the last sentenc as a normal comment because it's more of an implementation note, not something users of this functions need to be aware of.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -218,6 +221,14 @@ impl Tile {
...
+    #[inline(always)]
+    pub const fn is_sentinel(&self) -> bool {
+        self.x == u16::MAX
```
LaurenzV, line 230: Should we also check `y` here?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -310,19 +321,88 @@ impl Tiles {
...
-    pub fn make_tiles_analytic_aa(&mut self, lines: &[Line], width: u16, height: u16) {
+    pub fn make_tiles_analytic_aa<const USE_EARLY_CULL: bool>(
+        &mut self,
+        level: Level,
+        lines: &[Line],
+        width: u16,
+        height: u16,
+        partial_windings: &mut [[f32; Tile::HEIGHT as usize]],
+        coarse_windings: &mut [i8],
+        active_rows: &mut [u32],
```
LaurenzV, line 378: Can we put these 3, along with `culled_tiles`, into a single struct maybe and store it like that in `StripGenerator` so that we don't have to add three new parameters everywhere?

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -87,11 +88,17 @@ pub struct StripGenerator {
     tiles: Tiles,
     width: u16,
     height: u16,
+    culled_tiles: bool,
+    partial_windings: Vec<[f32; Tile::HEIGHT as usize]>,
+    coarse_windings: Vec<i8>,
+    active_rows: Vec<u32>,
```
LaurenzV, line 94: As mentioned elsewhere, those probably should be encapsulated into a single struct.

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -87,11 +88,17 @@ pub struct StripGenerator {
...
 }
 
 impl StripGenerator {
     /// Create a new strip generator.
     pub fn new(width: u16, height: u16, level: Level) -> Self {
+        let num_rows = (height.div_ceil(Tile::HEIGHT) + 1) as usize;
```
LaurenzV, line 100: Why +1? Can you add a comment?

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -87,11 +88,17 @@ pub struct StripGenerator {
...
+        let num_bits = num_rows.div_ceil(u32::MAX.count_ones() as usize);
```
LaurenzV, line 101: Could we just use `size_of<u32> * 8` here?

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -101,6 +108,10 @@ impl StripGenerator {
             temp_storage: StripStorage::default(),
             width,
             height,
+            culled_tiles: true,
```
LaurenzV, line 111: Should this really be true by default? The first time we build the strips we shouldn't need to fill everything with zeroes, no?

sparse_strips/vello_common/src/strip_generator.rs
LaurenzV: That new struct should probably live in `tile.rs` then.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -310,19 +321,88 @@ impl Tiles {
...
 
+    /// Marks if a row was culled early for faster traversal in strip generation.
+    #[inline(always)]
+    pub fn mark_row_active(active_rows: &mut [u32], row_idx: usize) {
+        active_rows[row_idx >> 5] |= 1 << (row_idx & 31);
+    }
+
+    /// Bulk marks a range of rows as active [`start_row`, `end_row`).
+    #[inline(always)]
+    pub fn mark_row_range_active(active_rows: &mut [u32], start_row: usize, end_row: usize) {
```
LaurenzV, line 332: These methods should probably become a method on the new struct then. Can we maybe add two unit tests, one for the `start_word` == `end_word` case and one for the other?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -310,19 +321,88 @@ impl Tiles {
         self.tile_buf.iter()
     }
 
+    /// Marks if a row was culled early for faster traversal in strip generation.
+    #[inline(always)]
+    pub fn mark_row_active(active_rows: &mut [u32], row_idx: usize) {
+        active_rows[row_idx >> 5] |= 1 << (row_idx & 31);
```
LaurenzV, line 327: It probably makes sense to factor the 5 and 31 out into two constants and document that they need to change if the type of `active_rows` changes.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -335,6 +415,11 @@ impl Tiles {
         let tile_columns = width.div_ceil(Tile::WIDTH);
         let tile_rows = height.div_ceil(Tile::HEIGHT);
 
+        let px_top: f32x4<_> = [0.0, 1.0, 2.0, 3.0].simd_into(s);
```
LaurenzV, line 418: Would maybe just use `f32x4::from_slice` instead, then you don't need the type annotation.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
             // the viewport OR it is perfectly horizontal and aligned to the tile grid, contributing
             // no winding. In either case, it should be culled.
             if y_top_tiles >= y_bottom_tiles {
-                // Technically, the `>` part of the `>=` is unnecessary due to clamping, but this
-                // gives stronger signal
```
LaurenzV, line 373: Is this comment really outdated now?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
...
+                    let start_v = f32x4::splat(s, local_y_start);
+                    let end_v = f32x4::splat(s, local_y_end);
+
+                    (px_bottom.min(end_v) - px_top.max(start_v)).max(simd_zero)
+                }};
+            }
+
+            // Lines fully to the left of the viewport are not visible but still produce winding
+            // which we record here and forward to the rendering stage.
+            if USE_EARLY_CULL && line_right_x < 0.0 {
```
LaurenzV, line 479: Doesn't meed to be addressed now and we can just remove it in a follow-up if needed, but did we decide somewhere that we want this to be a const generic? Why not just always cull since we already do that for the top, right and bottom side? Is it much more expensive?

At least in the long run, it seems to me like all sides should either always be culled or never, but not just the left side.

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
...
-                // Technically, the `>` part of the `>=` is unnecessary due to clamping, but this
-                // gives stronger signal
+                continue;
+            }
+
+            let dir = if p0_y >= p1_y { 1 } else { -1 };
+            let f_dir = dir as f32;
+            let f_dir_v = f32x4::splat(s, f_dir);
+
+            macro_rules! calc_fractional_coverage {
```
LaurenzV, line 464: Can this not just be a closure?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
...
+                ($y_idx:expr, $segment_top_y:expr, $segment_bottom_y:expr) => {{
+                    let y_idx_f32 = f32::from($y_idx);
+                    let local_y_start = ($segment_top_y - y_idx_f32) * tile_height_f32;
+                    let local_y_end = ($segment_bottom_y - y_idx_f32) * tile_height_f32;
+
+                    let start_v = f32x4::splat(s, local_y_start);
+                    let end_v = f32x4::splat(s, local_y_end);
+
+                    (px_bottom.min(end_v) - px_top.max(start_v)).max(simd_zero)
```
LaurenzV, line 473: Can this ever be less than zero?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
...
+                let is_start_culled = line_top_y < 0.0;
+                if !is_start_culled {
+                    Self::mark_row_active(active_rows, y_top_tiles as usize);
+
+                    let crosses_top = line_top_y <= f32::from(y_top_tiles);
+                    if crosses_top {
+                        coarse_windings[y_top_tiles as usize] += dir;
+                    }
+
+                    let h = calc_fractional_coverage!(y_top_tiles, line_top_y, line_bottom_y);
```
LaurenzV, line 489: Maybe we can name this `fractional_coverages` or something?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
...
+                    let target_row = &mut partial_windings[y_top_tiles as usize];
+                    let current = f32x4::from_slice(s, target_row);
+
+                    let next = if crosses_top {
+                        h.mul_add(f_dir_v, current - f_dir_v)
+                    } else {
+                        h.mul_add(f_dir_v, current)
+                    };
```
LaurenzV, line 497: Can you add a comment explaining what happens here? I assume i's to avoid double adding if we already added a coarse winding?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -369,8 +454,82 @@ impl Tiles {
...
+                }
+                Self::mark_row_range_active(
+                    active_rows,
+                    y_start_middle as usize,
+                    y_end_middle as usize,
+                );
+
+                if line_bottom_y != line_bottom_floor
+                    && y_end_middle < tile_rows
+                    && (is_start_culled || y_end_middle != y_top_tiles)
```
LaurenzV, line 520: Could you add a comment for this last part of the condition?

sparse_strips/vello_common/src/tile.rs
LaurenzV: Ah I see there is a comment further below already, so maybe just a `See the comment further below`?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -403,22 +563,13 @@ impl Tiles {
...
                         self.tile_buf.push(tile);
                     }
-
-                    // Row End, handle the final tile (y_end_idx), but *only* if the line does
-                    // not perfectly end on the top edge of the tile. In the case that it does,
-                    // it gets handled by the middle logic above.
-                    if line_bottom_y != line_bottom_floor && y_end_idx < tile_rows {
-                        let tile = Tile::new_clamped(x, y_end_idx, line_idx, W);
-                        self.tile_buf.push(tile);
-                    }
```
LaurenzV, line 420: Am I understanding correctly that this would not have been necessary before either?

sparse_strips/vello_sparse_tests/tests/basic.rs
```diff
@@ -579,3 +580,171 @@ fn composite_to_pixmap_at_offset() {
...
+    let rect = Rect::new(-50.0, -50.0, 80.0, 80.0);
+
+    ctx.set_transform(Affine::rotate_about(
+        7.0 * PI / 180.0,
+        Point::new(15.0, 15.0),
+    ));
+    ctx.set_paint(REBECCA_PURPLE);
+    ctx.fill_rect(&rect);
+}
+
```
LaurenzV, line 640: Is there any benefit for using masks here as opposed to just rendering directly into the render context directly? If not, i think it would be better to just use the render context directly.

sparse_strips/vello_bench/src/tile.rs
```diff
@@ -7,15 +7,42 @@ use vello_common::flatten::Line;
 use vello_common::tile::Tiles;
 use vello_cpu::Level;
 
-fn run_tile_benchmark<F>(c: &mut Criterion, group_name: &str, op: F)
+fn shift_lines_50_percent(lines: &[Line]) -> Vec<Line> {
```
LaurenzV, line 10: I think there should be no need to duplicate this method?

sparse_strips/vello_common/src/tile.rs
```diff
@@ -440,7 +591,59 @@ impl Tiles {
...
+                                    );
+                                    let target_row = &mut partial_windings[y_idx as usize];
+                                    let current = f32x4::from_slice(s, target_row);
+                                    let next = h.mul_add(f_dir_v, current);
+                                    target_row.copy_from_slice(next.as_slice());
+                                }
+                            }
+                        }
+
+                        let x_start = (row_left_x.max(0.0)) as u16;
```
LaurenzV, line 646: I don't think that change is needed, if it's < 0.0 then converting to u16 will automatically clamp to 0.

LaurenzV (commented):
Still need to review `strip.rs`, but here some first comments!

sparse_strips/vello_common/src/strip.rs
```diff
@@ -99,21 +198,53 @@ pub fn render(
...
+                                                         tiles,
+                                                         strip_buf,
+                                                         alpha_buf,
+                                                         fill_rule,
+                                                         aliasing_threshold,
+                                                         lines,
+                                                         partial_windings,
+                                                         coarse_windings,
+                                                         active_rows));
+    }
```
LaurenzV, line 228: Could you please benchmark the impact of just passing the boolean as a normal parameter instead of a constant generic? I suspect there should be any notable slowdown because the boolean check doesn't appear in any hot loop, and I'm a bit concerned about code size since the method is already parametrized on the SIMD level. If there's no big impact, I would prefer just using a normal argument for this.

sparse_strips/vello_sparse_tests/tests/basic.rs
b0nes164: Is it possible to exercise the `clip.rs` code path with a single render context? This is based on the hayro snippet you sent me. The issue was with how `clip.rs` advances through strips.

sparse_strips/vello_common/src/strip_generator.rs
b0nes164: Yea, forgot these are already getting cleared by the initialization

sparse_strips/vello_common/src/tile.rs
b0nes164: Checking the x is sufficient, as mentioned above, it is impossible to produce a tile with this x-value from `make_tiles`.

sparse_strips/vello_sparse_tests/tests/basic.rs
LaurenzV: Oh, yeah sure you can just call push clip path on the main render context!

sparse_strips/vello_common/src/strip.rs
b0nes164: K, yea, I left this in as a convenient way to enable/disable the culling. I'm happy to change it, but given that we only pass in `true` in production---we only disable for testing---it shouldn't be compiling multiple variants

b0nes164: Edit: was responding on mobile I thought this was referring to `tile.rs` and not `strip.rs`. And, yes I'll benchmark this one, and most likely remove it.

For the `tile.rs`, the previous comment still holds: since we only compile a single variant outside of testing, it shouldn't increase the binary size, but we retain the ability to statically enable/disable culling.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -88,6 +88,105 @@ impl Strip {
...
+    {
+        if start >= end {
+            return;
+        }
+
+        let mut push_strip = |y_pos: u16| {
+            strips.push(Self::new(0, y_pos, alphas.len() as u32, false));
+            alphas.extend([255_u8; Tile::HEIGHT as usize * Tile::WIDTH as usize]);
+            strips.push(Self::new(u16::MAX, y_pos, alphas.len() as u32, true));
+        };
```
LaurenzV: Can we add a TODO here to explore clamping the position of the second strip to the scene width instead of u16::MAX? I think right now all renderers clamp this themselves, but would be good to track this.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -88,6 +88,105 @@ impl Strip {
...
+
+        let start_word = (start >> 5) as usize;
```
LaurenzV: Similarly to how I mentioned somewhere else (I think), would be better to use constants here I think.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -88,6 +88,105 @@ impl Strip {
...
+    ///
+    /// We reconstruct this "background" fill using `row_windings` (the winding at x=0) to emit solid
+    /// strips for:
+    ///      1. All rows vertically above the first visible tile.
+    ///      2. 'Captive' rows between two tile-containing rows.
+    ///      3. All rows vertically below the last visible tile.
+    #[inline(always)]
+    fn emit_culled_background<F>(
+        start: u16,
+        end: u16,
```
LaurenzV, line 104: Can you add a comment that `end` is exclusive?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -88,6 +88,105 @@ impl Strip {
...
+                while word != 0 {
+                    let bit = word.trailing_zeros();
+                    word &= !(1 << bit);
+
+                    let row = (word_idx << 5) + bit as usize;
+                    if should_fill(row_windings[row] as i32) {
+                        let y_pos = row as u16 * Tile::HEIGHT;
+                        push_strip(y_pos);
+                    }
+                }
```
LaurenzV: Seems like this can be deduplicated with above code? Maybe with another closure? Also further below.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -138,18 +274,49 @@ fn render_impl<S: Simd>(
...
+        Strip::emit_culled_background(
+            0,
+            row_max,
+            strip_buf,
+            alpha_buf,
+            active_rows,
+            row_windings,
+            should_fill,
+        );
+        if !tiles.is_empty() {
```
LaurenzV: I think it would be better to have
```rs
if tiles.is_empty() {
return;
}
```

And then just have the bigger part not in an else block.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -138,18 +274,49 @@ fn render_impl<S: Simd>(
...
+            winding_delta = row_windings[prev_tile.y as usize] as i32;
+            if should_fill(winding_delta) && !left_viewport {
+                strip_buf.push(Strip::new(
+                    0,
+                    prev_tile.y * Tile::HEIGHT,
+                    alpha_buf.len() as u32,
+                    false,
+                ));
+                alpha_buf.extend([255_u8; Tile::HEIGHT as usize * Tile::WIDTH as usize]);
```
LaurenzV: I'm wondering if those are necessary (same for above), but not for you to figure out. For now, that's definitely the better choice, unsure if there is code that assumes strips have at least a width of 4.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -138,18 +274,49 @@ fn render_impl<S: Simd>(
...
+            }
+            accumulated_winding = f32x4::splat(s, winding_delta as f32);
+            if left_viewport {
+                let fine_winding: f32x4<_> = partial_windings[prev_tile.y as usize].simd_into(s);
```
LaurenzV: Same here, `from_slice` would probably be better.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -240,13 +408,47 @@ fn render_impl<S: Simd>(
                     ));
                 }
 
-                winding_delta = 0;
-                accumulated_winding = f32x4::splat(s, 0.0);
+                // Logic identical to the start (see above): fill any vertical gaps (empty rows)
+                // between the previous and current tile using the row windings.
```
LaurenzV, line 332: No way this could be deduplicated then somehow?

sparse_strips/vello_common/src/tile.rs
b0nes164: Up to you if you want to change it; on my pc it's 43us with the macro vs 45us with the closure

sparse_strips/vello_common/src/tile.rs
b0nes164: Yes, and I've also changed back to the simplified loop in the general case too. For the AA path, there's no need to distinguish between when the final row is fully crossed versus partially crossed.

sparse_strips/vello_common/src/tile.rs
b0nes164: Yes, if the line does not cross the entire vertical height of the tile, the "missed pixels" will be less than zero and require zeroing.

sparse_strips/vello_common/src/strip_generator.rs
b0nes164: I originally included this because I did not have this check in the predicate for the row walking:
```
y_end_middle < tile_rows
```

Which meant that, if you had a line which ended exactly at the bottom of the viewport, this logic:
```
let y_end_middle = (line_bottom_floor as u16).min(tile_rows);
```
Would make `y_end_middle = tile_rows` and OOB.

That being said, it should be safe to remove now.

sparse_strips/vello_common/src/strip.rs
b0nes164: For posterity here are the test results for `render_strips`:
With const generic:
Tiger 50% culled: ~38us
Tiger normal: ~96 us

Without const generic:
Tiger 50% culled: ~40us 
Tiger normal: ~102 us

I've changed to the non-const generic version

sparse_strips/vello_common/src/strip.rs
b0nes164: Not sure about other downstream affects, but they are necessary for at least clipping. Without injecting in the alphas, this:
```
#[inline(always)]
fn cur_strip_width(&self) -> u16 {
    let cur = self.cur_strip();
    let next = self.next_strip();
    ((next.alpha_idx() - cur.alpha_idx()) / Tile::HEIGHT as u32) as u16
}
```

returns 0, resulting in a 0 width fill region, so downstream in:
```
fn overlap_relationship(&self, other: &Region<'_>) -> OverlapRelationship {
        if self.end() <= other.start() {
            OverlapRelationship::Advance(Advance::Left)
        } else if self.start() >= other.end() {
            OverlapRelationship::Advance(Advance::Right)
        } else {
           . . .
        }
    }
```

if you have a clip that also starts at the left edge (also possibly culled?), you get:
```
if self.end() <= other.start() {  // 0 <= 0 (Evaluates to TRUE)
    OverlapRelationship::Advance(Advance::Left)
}
```
Resulting in no overlap

b0nes164:
PTAL!

LaurenzV (approved):
LGTM, thanks for your patience with this! Happy to see this finally being landed. If you are happy with the two commits I added, feel free to merge.

Did some benchmarking, testing in interactive mode and also another test run against my PDF test suite, and didn't observe any regressions.

b0nes164:
Yup LGTM! Thanks for getting this over the finish line!

PR #1370, opened 2026-01-21
Fix double-scaling bug with non-hinted COLR glyphs in vello_common

conor-93:
**Overview**

Spotted when using `vello_cpu`. This fixes a bug where scaling is applied twice. The fix works by simply removing the extra scale component from the glyph run transform (by applying an extra, interverted transform) before it is used to compose the per-glyph transform. We also separate the width and height components of the bounding box scaling now, so that the extra inverted transform can be non-uniform.

`vello_common`'s COLR glyph rendering is duplicated in `parley_draw`, see also: https://github.com/linebender/parley/pull/521

**Examples** 

LHS: Correct output (this PR)
RHS: double-scaling output (existing behaviour)

2x scale being applied twice (snapshot test):
image

Half scale being applied twice (snapshot test):
image

### DIFF
--- a/sparse_strips/vello_common/src/glyph.rs
+++ b/sparse_strips/vello_common/src/glyph.rs
@@ -435,7 +435,7 @@ fn prepare_colr_glyph<'a>(
     // exactly one pixel (or more) per device pixel, to ensure that no quality is lost. Therefore,
     // we simply use the scaling/skewing factor to calculate how much to scale each axis by.
     let (scale_factor_x, scale_factor_y) = {
-        let (x_vec, y_vec) = x_y_advances(&transform.pre_scale(f64::from(font_size_scale)));
+        let (x_vec, y_vec) = x_y_advances(&transform.pre_scale(font_size_scale));
         (x_vec.length(), y_vec.length())
     };
 


taj-p:
Sorry about the extra work - Could we please also apply the fix to: https://github.com/linebender/parley/blob/main/parley_draw/src/glyph.rs before we sort out the duplication between Vello and Parley Draw? 🙏

laurenz-canva:
I've tweaked your code to make the logic a bit simpler, and I think I've also fixed the non-uniform scaling issue, see here: https://github.com/linebender/vello/compare/main...laurenz-canva:vello:laurenz/colr-fix-part-2

Does this look good to you? If so, feel free to cherry-pick and then we can merge. :)

sparse_strips/vello_sparse_tests/tests/glyph.rs
```diff
@@ -265,6 +265,116 @@ fn glyphs_colr_noto(ctx: &mut impl Renderer) {
         .fill_glyphs(glyphs.into_iter());
 }
 
+/// Test COLR emoji with 2x scene scale.
+/// This tests the fix for the double-scaling bug where scene transforms
+/// were applied twice to COLR glyphs (once in area dimensions, once in `glyph_transform`).
+/// Hinting is disabled to preserve the scale in the transform passed to `prepare_colr_glyph`.
```
laurenz-canva: Aside from that, maybe we can create a reusable method that takes a scale factor/transform, and then make each test just a wrapper around it? So that we don't have to repeat everything.

conor-93:
> Does this look good to you? If so, feel free to cherry-pick and then we can merge. :)

Fantastic, this is a huge improvement. Thank you Laurenz! As you suggested, I've cherrypicked your changes to this PR.

sparse_strips/vello_sparse_tests/tests/glyph.rs
conor-93: Done! Great idea.

LaurenzV:
Thanks!!

PR #1371, opened 2026-01-22
Simplify the implementation of the `Flood` filter

laurenz-canva:


laurenz-canva:
I don't think we need to specialize the premultiplication step for lowp, since it's only performed once anyway, so it's very cheap.

PR #1376, opened 2026-01-23
Avoid hardcoding a tolerance in Vello API, instead require the user to provide it

DJMcNab:
This was one of the key points raised for follow-up in #1360, to make this API harder to misuse.

The implementation strategy is to have a new `ExactPathElements` trait, which is implemented for `Shape`s which don't require approximation.
To allow shapes which do require a tolerance to be used without allocation, the `within` function can be used, which creates an exact shape from an inexact shape *within* the provided tolerance.

This ensures that the common-case of rectangles and similar is ergonomic, whilst the case where tolerance is needed is handled by the user. This still completely avoids per-path allocations.

The `ExactPathElements` is based on code by @tomcur from [#vello > Determining correct &#96;Shape&#96; tolerance](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Determining.20correct.20.60Shape.60.20tolerance/with/567789803). There are reasonable arguments that `ExactPathElements` actually belongs in Kurbo, but I also think that, if this is the direction we decide to go, migrating that code into Kurbo would (relatively) straightforward.

### DIFF
--- a/sparse_strips/vello_api/src/exact.rs
+++ b/sparse_strips/vello_api/src/exact.rs
@@ -75,9 +75,9 @@ impl<T: ExactPathElements> ExactPathElements for &T {
 /// approximation is only valid for a fixed range of transforms.
 ///
 /// As the user of this function, you are responsible for determining the correct tolerance for your use case.
-/// A reasonable approach might be to select a tolerance which allows scaling up ("zooming in") by 4x (for example, 0.00001) to be
-/// within your intended tolerance bound, then recomputing the [`Scene`](crate::Scene) from your base data
-/// representation once that is exceeded.
+/// A reasonable approach might be to select a tolerance which allows scaling up ("zooming in") by 4x (for example;
+/// you should evaluate the correct value yourself) and remaining within your intended tolerance bound.
+/// If the user zoomed past that limit, you would then recomputing the [`Scene`](crate::Scene) from your base data representation once that is exceeded.
 /// If you know that the shape will not be scaled, you can use [`UNSCALED_TOLERANCE`].
 /// The resulting path will be within 1/10th of a pixel of the actual shape, which is a negligible
 /// difference for rendering.


sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+/// A generic trait for shapes that can be mapped exactly to Bézier path elements (i.e., without
+/// approximation).
+///
+/// The methods on [`PaintScene`](crate::PaintScene) use this trait ensure that consistent behaviour
+/// is maintained when Vello API is used to render content which might be rescaled.
+///
+/// This is implemented for several [`Shape`]s from Kurbo.
+/// To convert a [`Shape`] which requires approximation (such as [`Circle`](kurbo::Circle) or
+/// [`RoundedRect`](kurbo::RoundedRect)), you can use [`within`].
+/// Note that the returned shape will use.
```
taj-p: This sentence is incomplete

sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+use kurbo::{BezPath, CubicBez, Line, QuadBez, Rect, Segments, Shape, Triangle, segments};
+use peniko::kurbo;
+
+/// A generic trait for shapes that can be mapped exactly to Bézier path elements (i.e., without
+/// approximation).
+///
+/// The methods on [`PaintScene`](crate::PaintScene) use this trait ensure that consistent behaviour
+/// is maintained when Vello API is used to render content which might be rescaled.
+///
+/// This is implemented for several [`Shape`]s from Kurbo.
```
taj-p:
```suggestion
/// This is implemented for [`Shape`]s from Kurbo that can be exactly mapped to Bézier path elements.
```

sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Use an approximated shape where an [`ExactPathElements`] is required, by approximating it to
+/// within the given tolerance.
+///
+/// This is useful for drawing shapes such as [`Circle`](kurbo::Circle)s and [`RoundedRect`](kurbo::RoundedRect)s
+/// to renderers which use Vello API.
+///
+/// As the user of this function, you are responsible for determining the correct tolerance for your use case.
+/// A reasonable approach might be to select a tolerance which allows scaling up ("zooming in") by 4x to be
+/// within your intended tolernace bound, then recomputing the [`Scene`](crate::Scene) from your base data
```
taj-p:
```suggestion
/// within your intended tolerance bound, then recomputing the [`Scene`](crate::Scene) from your base data
```

sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+        impl ExactPathElements for $ty {
+            fn exact_path_elements(&self) -> Self::PathElementsIter<'_> {
+                // We use a tolerance of zero here because we know this to be exact.
+                self.path_elements(0.)
+            }
+        }
+    };
+}
+passthrough!(BezPath);
+
```
taj-p:
```suggestion

passthrough!(BezPath);
```

sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn winding(&self, pt: kurbo::Point) -> i32 {
+        // TODO: Is this correct?
+        self.shape.winding(pt)
+    }
+
+    fn bounding_box(&self) -> Rect {
+        // TODO: Is this correct?
+        self.shape.bounding_box()
+    }
```
taj-p:
```suggestion
    fn area(&self) -> f64 {
        self.path_segments(self.tolerance).area()
    }

    fn perimeter(&self, accuracy: f64) -> f64 {
        self.path_segments(self.tolerance).perimeter(accuracy)
    }

    fn winding(&self, pt: kurbo::Point) -> i32 {
        self.path_segments(self.tolerance).winding(pt)
    }

    fn bounding_box(&self) -> Rect {
        self.path_segments(self.tolerance).bounding_box()
    }
```

I believe these should be derived after applying the tolerance to be correct - but these path segment methods are private to kurbo 🤔

sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+/// This is useful for drawing shapes such as [`Circle`](kurbo::Circle)s and [`RoundedRect`](kurbo::RoundedRect)s
+/// to renderers which use Vello API.
+///
+/// As the user of this function, you are responsible for determining the correct tolerance for your use case.
+/// A reasonable approach might be to select a tolerance which allows scaling up ("zooming in") by 4x to be
+/// within your intended tolernace bound, then recomputing the [`Scene`](crate::Scene) from your base data
+/// representation once that is exceeded.
+/// If you know that the shape will not be scaled, you can use [`UNSCALED_TOLERANCE`].
+/// The resulting path will be within 1/10th of a pixel of the actual shape, which is a negligible
+/// difference for rendering.
```
taj-p, line 83: Edited by @DJMcNab to make suggestion take up less space.

<details><summary>Already applied suggestion</summary>
<p>

```suggestion
/// Use an approximated shape where an [`ExactPathElements`] is required, by approximating it to
/// within the given tolerance.
///
/// *WARNING*: Unlike [`ExactPathElements`], which will produce correct renderings for any scale, this
/// approximation is only valid for a fixed range of transforms.
///
/// As the user of this function, you are responsible for determining the correct tolerance for your use case.
/// A reasonable approach might be to select a tolerance which allows scaling up ("zooming in") by 4x (for example, 0.00001) to be
/// within your intended tolerance bound, then recomputing the [`Scene`](crate::Scene) from your base data
/// representation once that is exceeded.
/// If you know that the shape will not be scaled, you can use [`UNSCALED_TOLERANCE`].
/// The resulting path will be within 1/10th of a pixel of the actual shape, which is a negligible
/// difference for rendering.
///
/// This is useful for drawing shapes such as [`Circle`](kurbo::Circle)s and [`RoundedRect`](kurbo::RoundedRect)s.
```

</p>
</details>

sparse_strips/vello_api/src/exact.rs
```diff
@@ -0,0 +1,148 @@
...
+// TODO: Provide an `fn ideal_tolerance(transform: Affine) -> f64`
+
+/// Implement `ExactPathElements` for an existing [`Shape`], which we know
+// In theory, the impl is the wrong way around; instead the `Shape` impl should be in terms of `ExactPathElements`.
+macro_rules! passthrough {
+    ($ty: ty) => {
+        impl ExactPathElements for $ty {
+            fn exact_path_elements(&self) -> Self::PathElementsIter<'_> {
+                // We use a tolerance of zero here because we know this to be exact.
+                self.path_elements(0.)
```
taj-p, line 130: If I understand this PR correctly, we're patching over Kurbo's `Shape` trait using a marker trait to fix a leaky abstraction in `Shape`: that `path_elements` and `Shape` hide a quality that callers need to care about, tolerance. In this case, we're using a marker trait (`ExactPathElements`) to identify shapes who do not use tolerance. I believe in this case the `Shape` abstraction is making things more complex than necessary.

I can't help but feel this is technical debt we're adding to Vello API as a means to bolt on `Shape` argument support. But, we don't actually make use of any `Shape` methods besides getting an iterator of path elements. So, from my perspective, we're adding technical debt for little to no gain. 

I think until we do see a benefit to the implementation for supporting `Shape` it shouldn't be added to the API. Do we have anything in the works that would benefit from its other methods? 

Separately, if we were to make use of different shape fast paths, I don't believe we would want to use the generic `fill_path(..., shape)` method. We would prefer a `fill_rect` type of API like Skia so we don't need to check for type via `as_rect()`, `as_circle()`, etc. In any case, that has not been investigated and we don't know where it might land (inclusive of `.bounding_box()` optimisation).

If we're thinking about ergonomics, the difference between these two calls:

```rs
scene.fill_path(transform, Fill::NonZero, rect);
scene.fill_path(transform, Fill::NonZero, rect.path_elements(0.0));
```

is very small. If we really want to pass a shape without `path_elements`, then I recommend removing the `Shape` supertrait from `ExactPathElements` for the same effect. In that way, we also aren't incorrectly implementing its other methods (see above comment).

We want to keep technical debt outside of Vello and push it to consumers until it can be properly addressed. Taking `shape` brings in conceptual baggage (and technical debt as this PR is introducing) that isn't made use of.

So, could we instead accept `impl IntoIterator<Item = PathEl>` directly to match what the implementation needs or drop the `Shape` supertrait? NOTE: Dropping the `Shape` super trait in no way impacts API consumption - users can still pass `Shape`s that implement `ExactPathElements`.

I am curious to understand the context for supporting `Shape` methods - I note its use in Vello Classic (even though its other methods like `area`, `perimeter`, etc are unused) - if it's only for ergonomics, I think dropping `Shape` supertrait from `ExactPathElements` is preferable with non-exact shapes being passed in as path el iterators or via `within` function or similar. 

This approach keeps the current API lean, while leaving room for later `Shape` super trait extension - if that's ever needed.

sparse_strips/vello_cpu/src/api.rs
```diff
@@ -133,16 +134,25 @@ impl PaintScene for CPUScenePainter {
         Ok(())
     }
 
-    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl ExactPathElements) {
```
taj-p: It is a little confusing that we take an owned `ExactPathElements` instead of a borrowed type. For hybrid and single threaded CPU, we only require borrowed data. For multithreading, we still only require borrowed data IIUC since [we copy the path into the allocation group](https://github.com/taj-p/vello/blob/e0206994f6067fe3c1e3d2c6a389750de40964c4/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L441). 

Should this be `&impl ExactPathElements` to better reflect that only borrowed data is required from the caller? I believe Vello Classic took this approach - is there context for why `impl ExactPathElements` is better?

sparse_strips/vello_api/src/exact.rs
DJMcNab: I have one concern here, which is the 0.00001 bound. I believe that would support a $$10^4$$ times zoom, rather than a 4x zoom, right?

sparse_strips/vello_api/src/exact.rs
DJMcNab: As you note, removing the `Shape` supertrait is probably correct.

sparse_strips/vello_api/src/exact.rs
DJMcNab: I'll remove the `Shape` supertrait. To clarify, the reason I kept it was for porting Vello CPU and Vello Hybrid, which currently use the `into_path` methods (well, as of #1375), to avoid some allocations. Those methods don't exist on `ExactPathElements`, and.

However, the consequence that `WithTolerance: Shape` is, as you rightly point out, unworkable. ~~I'll think about adding an `as_bezpath` method or similar.~~ Edit: Scratch that, these allocations will be alleviated when we change the underlying API to not require allocations, which we know is necessary.

sparse_strips/vello_cpu/src/api.rs
DJMcNab: I thought I was copying Vello Classic, but a revisit of the docs for `vello::Scene` shows that my memory was mistaken. I'll change this; I think it makes as much sense to do it in this PR.

DJMcNab: In applying this change, it raises a small ergonomics question; for the vast majority of non-`BezPath` shapes, the extra `&` isn't doing anything meaningful here. A conceivable compromise here would be `impl ExactPathElements + Copy`, which would not require the `&` for shapes where it isn't consuming, but would force it for e.g. `BezPath`.

In this PR, I've changed it to be `&` for simplicity/to avoid blocking the rest of this work on that question. But it's something I'll consider for a follow-up.

sparse_strips/vello_cpu/src/api.rs
```diff
@@ -133,17 +135,28 @@ impl PaintScene for CPUScenePainter {
...
 
-    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: &impl ExactPathElements) {
         self.render_context.set_transform(transform);
         self.render_context.set_fill_rule(fill_rule);
-        // TODO: Tweak inner `fill_path` API to either take a `Shape` or an &[PathEl]
-        self.render_context.fill_path(&path.into_path(0.1));
+        // This tolerance parameter is meaningless, because this is an `ExactPathElements`
+        // However, using `to_path` avoids allocation in some cases.
+        // TODO: Tweak inner API to accept an `ExactPathElements` (or at least, the resultant iterator)
```
taj-p, line 142: We probably want to just pass the iterator so we don't monomorphise across every trait impl

sparse_strips/vello_cpu/src/api.rs
```diff
@@ -133,17 +135,28 @@ impl PaintScene for CPUScenePainter {
         Ok(())
     }
 
-    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: &impl ExactPathElements) {
         self.render_context.set_transform(transform);
         self.render_context.set_fill_rule(fill_rule);
-        // TODO: Tweak inner `fill_path` API to either take a `Shape` or an &[PathEl]
-        self.render_context.fill_path(&path.into_path(0.1));
+        // This tolerance parameter is meaningless, because this is an `ExactPathElements`
```
taj-p:
```suggestion
```
I think this is an outdated comment

sparse_strips/vello_hybrid/src/api.rs
```diff
@@ -131,17 +133,27 @@ impl PaintScene for HybridScenePainter {
         Ok(())
     }
 
-    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: impl Shape) {
+    fn fill_path(&mut self, transform: Affine, fill_rule: Fill, path: &impl ExactPathElements) {
         self.scene.set_transform(transform);
         self.scene.set_fill_rule(fill_rule);
-        // TODO: Tweak inner `fill_path` API to either take a `Shape` or an &[PathEl]
-        self.scene.fill_path(&path.into_path(0.1));
+        // This tolerance parameter is meaningless, because this is an `ExactPathElements`
```
taj-p:
```suggestion
```

nicoburns:
Which shapes cannot implement `ExactPathElements`?

Is it primarily those based on circular geometry? (Arc, Circle, Ellipse, etc?). If so, might it make sense to create an extended version of `PathEl` that is able to encode those losslessly (i.e. by adding an `ArcTo` variant to the enum) and then use that in `Scene` (we could add an `extended_path_elements` method or similar directly to the `Shape` trait). Then the conversion to `BezPath` could be deferred into the renderers that would have context about absolute scale, and would be able to determine a sensible value for the tolerance.

Or are (infinitely many) other shape types that we may need to support such that this wouldn't work?

tomcur:
> Which shapes cannot implement `ExactPathElements`?
>
> Is it primarily those based on circular geometry? (Arc, Circle, Ellipse, etc?).

That's right. More specifically, only shapes that can be represented by cubic polynomial segments (including line and quadratic segments) can be `ExactPathElements` (given the current path element kinds).

> If so, might it make sense to create an extended version of PathEl that is able to encode those losslessly (i.e. by adding an ArcTo variant to the enum) 

I don't think so. The issue with something like a circular `ArcTo` is that it cannot be preserved under affine transforms. Béziers (including lines) are nice, because the affine transform of a Bézier path is just again a regular Bézier with simply the original control points transformed.

Edit: with your edits I now understand you meant keeping those extended path elements around until such time the actual transforming is done. Something like that is possible, but at that point may be quite similar to just keeping an enumeration of shapes or a `dyn Shape` around. For extending path elements to allow arcs while still doing lossless transformation anywhere in the pipeline, an elliptic arc path element could work, as it's also closed under affine transformation.

nicoburns:
> Edit: with your edits I now understand you meant keeping those extended path elements around until such time the actual transforming is done. Something like that is possible, but at that point may be quite similar to just keeping an enumeration of shapes or a dyn Shape around. For extending path elements to allow arcs while still doing lossless transformation anywhere in the pipeline, an elliptic arc path element could work, as it's also closed under affine transformation.

Yeah, that's the idea.

Ellitical arcs (and/or perhaps even hyper-elliptical arcs - people love their squircles) would make sense to me.

It would be similar to keeping around Shape, but a more practical way of implementing it I think. Storing `dyn Shape` would require allocating a `Box<dyn Shape>` for every single shape. And I guess the idea is that it does still somewhat simplify an arbitrary Shape down into simpler primitives (e.g. this would support `RoundedRectangle` without having to have a special case for it.

tomcur:
Skia has e.g. [`SkPath::conicTo`](https://api.skia.org/classSkPath.html#a9edc41978765cfe9a0b16e9ecf4d276e).

sparse_strips/vello_cpu/src/api.rs
DJMcNab: I believe it would still in most cases be a separate impl for each `Shape`, as each Shape has their own iterator type (unless I guess we used something like `&mut dyn Iterator`?)

sparse_strips/vello_api/src/exact.rs
DJMcNab: I've removed the `0.00001` for now; we can always restore it in a follow-up.

PR #1377, opened 2026-01-24
[vello_common] Add CPU MSAA

b0nes164:
Uploading this as a potential feature, and if we do want to land it, I can polish it up from here. Currently, the performance is pretty bad, about 3x Analytic anti-aliasing (~4.3ms vs ~1.5ms on tiger). However, the pixel intersection logic is currently written GPU-ish style, and I can imagine switching to more serial friendly DDA approach could yield some savings.

### DIFF
--- a/sparse_strips/vello_common/src/strip.rs
+++ b/sparse_strips/vello_common/src/strip.rs
@@ -93,6 +93,61 @@ impl Strip {
     }
 }
 
+
+
+/// This struct acts as a cache for expensive per-line calculations (square roots,
+/// divisions) that would otherwise happen per-tile.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct PreprocessedLine {
+    /// 1.0 / (dx^2 + dy^2).sqrt()
+    pub recip_len: f32,
+    /// p1.x - p0.x
+    pub dx: f32,
+    /// p1.y - p0.y
+    pub dy: f32,
+    /// 1.0 / dx (or 0.0 if vertical)
+    pub idx: f32,
+    /// 1.0 / dy (or 0.0 if horizontal)
+    pub idy: f32,
+    /// dx / dy (slope relative to y)
+    pub dxdy: f32,
+}
+
+/// Calculate per line information
+pub fn preprocess_lines(lines: &[Line]) -> Vec<PreprocessedLine> {
+    lines
+        .iter()
+        .map(|line| {
+            let dx = line.p1.x - line.p0.x;
+            let dy = line.p1.y - line.p0.y;
+
+            let delta_sq = dx * dx + dy * dy;
+            let recip_len = if delta_sq < 1e-12 {
+                0.0
+            } else {
+                1.0 / delta_sq.sqrt()
+            };
+
+            let is_vertical = dx.abs() <= f32::EPSILON;
+            let is_horizontal = dy.abs() <= f32::EPSILON;
+
+            let idx = if is_vertical { 0.0 } else { 1.0 / dx };
+            let idy = if is_horizontal { 0.0 } else { 1.0 / dy };
+
+            let dxdy = dx * idy;
+
+            PreprocessedLine {
+                recip_len,
+                dx,
+                dy,
+                idx,
+                idy,
+                dxdy,
+            }
+        })
+        .collect()
+}
+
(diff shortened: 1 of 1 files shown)


b0nes164:
4.3ms -> 3.6ms on tiger

b0nes164:
~3.6ms -> ~2.8ms on tiger.

Things that I tried but were either the same perf or worse:

- morton encoding the LUT
- u8x32 SIMD instead of the current u32x8
- dda stepping the rows

Last thing that could be tried is switching from processing by row to processing by column. This would at the very least save the transpose necessary before writing to the alpha buffer.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -417,3 +494,601 @@ fn render_impl<S: Simd>(
...
+/// Clips a line segment to a tile.
+///
+/// Returns the start and end points of the clipped line segment relative to the tile origin.
+#[inline]
+pub fn clip_to_tile(
+    line: &Line,
+    bounds: &[f32; 4],
+    derivatives: &[f32; 4],
+    intersection_data: u32,
+    cannonical_x_dir: bool,
```
dominikh, line 578: canonical, not cannonical

sparse_strips/vello_common/src/strip.rs
```diff
@@ -417,3 +494,601 @@ fn render_impl<S: Simd>(
...
+                        let zero_markers = subbed.and(diff.not()).and(bias);
+                        let active_markers = zero_markers.xor(bias);
+                        let counts = active_markers.shr(7).mul(ones).shr(24);
+                        let evens = counts.unzip_low(counts);
+                        let odds = counts.unzip_high(counts);
+                        let pixel_counts = evens.add(odds);
+                        computed_rows[y] = pixel_counts.mul(scale_mul).add(scale_add).shr(3);
+                    }
+
+                    for x in 0..4 {
```
dominikh, line 789: I'm probably missing something, but why is computed_rows an array of u32x8 when we only use the first four values per row? Shouldn't it be u32x4 instead?

b0nes164: Good question, also, for whatever reason I didn't get a notification for these? 

First, yes the code here is for 4x4 tiles. But it is u32x8 because each pixel in the tile requires MSAA count x byte amount of space to record the winding for each sample point.

So at MSAA 8, 8 bytes per pixel, 4 pixels per row, 32 bytes total, u32x8!

dominikh: But am I wrong in seeing that for each u32x8 row in computed_rows, we never observe the elements 4..8? We compute some values for them, but they're never stored anywhere.

For each row, we do:

- initialize to all zeros
- set to the result of `pixel_counts.mul(scale_mul).add(scale_add).shr(3)`
- Store elements 0..4 in alpha_buf

And unless I'm missing something, the whole data flow graph for that computation is insensitive to the elements 4..8? That is, I didn't spot any lane crossing operations.

dominikh: If you're just using u32x8 because it meshes better with the types of other variables involved here, that'd be understandable (but probably deserving of a comment?). But all the types involved seem to derive from a single `v = mask[y]`, and it'd probably be easy enough to take the low half of mask[y] to clarify the types involved here?

b0nes164: I misread your original question, I thought you were asking about the use of u32x8 in general, using `compute_rows` as an example.

Yes, `compute_rows` should be u32x4. I initially wrote this to push the rows directly into the alpha buffer, but then I remembered that Vello expects collumn-major so I haphazardly added this for the transpose!

If you're trying to run it locally, iirc the last commit passes the lines in incorrectly, so watch out for that too.

b0nes164: That being said, this version is a little outdated from what I have on Skia right now, the biggest differing being that I'm using Raph's [slope and intercept based LUT](https://github.com/linebender/vello/issues/270) as opposed to the [angle and distance LUT](https://github.com/Mochimazui/gpu-scanline-path-rendering) from the  Li et al paper.

Furthermore, if you've kept up with the zulip, I've been tinkering with a RLE tile representation, and there actually might be a new technique it enables, so there may be more changes coming in the future!

PR #1384, opened 2026-01-28
Remove Process Row Incremental Macro and Perfect Bit

b0nes164:
Removes the "Perfect" bit---the bit used to indicate when a line perfectly intersects a tile's corner. As Raph suspected, tracking this scenario was unnecessary. Commensurately, the max line count has been bumped back up to 2^27.

Originally, I was intending to remove the entire macro setup, but replacing it would require a double closure, which was performance loss (possibly due to lack of inlining or compiler difficulty in handling the mutable references necessary for double closure). So instead, this patch only removes the `!process_row_incremental` macro, which after more benchmarking, appears to be a performance loss.

Refactors the `vello_common/tile.rs` to use `const`s instead of magic numbers for processing and packing the intersection mask.

Refactors the `vello_bench/tile.rs` to allow benchmarking the msaa or aaa tilers.

### DIFF
--- a/sparse_strips/vello_bench/src/tile.rs
+++ b/sparse_strips/vello_bench/src/tile.rs
@@ -27,13 +27,6 @@ where
 }
 
 pub fn tile(c: &mut Criterion) {
-    #[expect(clippy::type_complexity, reason = "tiler variants")]
-    let methods: &[(&str, fn(&mut Tiles, &[Line], u16, u16))] = &[
-        ("tile_aaa", |t, l, w, h| t.make_tiles_analytic_aa(l, w, h)),
-        ("tile_msaa", |t, l, w, h| t.make_tiles_msaa(l, w, h)),
-    ];
-
-    for (name, op) in methods {
-        run_tile_benchmark(c, name, *op);
-    }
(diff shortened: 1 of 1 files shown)


b0nes164:
Apologies, I just wanted to run the CI on the change, it wasn't ready for review yet! And I agree, there should definitely be some `const`s.

sparse_strips/vello_bench/src/tile.rs
```diff
@@ -2,28 +2,38 @@
...
-        tile_single!(item);
+pub fn tile(c: &mut Criterion) {
+    #[expect(clippy::type_complexity, reason = "tiler variants")]
+    let methods: &[(&str, fn(&mut Tiles, &[Line], u16, u16))] = &[
+        ("tile_aaa", |t, l, w, h| t.make_tiles_analytic_aa(l, w, h)),
+        ("tile_msaa", |t, l, w, h| t.make_tiles_msaa(l, w, h)),
+    ];
+
+    for (name, op) in methods {
+        run_tile_benchmark(c, name, *op);
```
LaurenzV: Would it not be easier to just call the methods twice, without the array? 😄

PR #1385, opened 2026-01-28
vello_sparse_tests: Use `u32` when calculating pixel differences

laurenz-canva:
Otherwise, we might get an overflow if the diff image has more than 65k different pixels.


laurenz-canva:
I've changed it to use u32 throughout, but I'll leave the renaming to a future PR.

PR #1387, opened 2026-01-28
vello_hybrid: Add a clarifying comment

laurenz-canva:
Is my understanding correct here?


laurenz-canva:
Nevermind, this is already documented in other parts of the source code (perhaps we can turn it into a method for better clarity, though).

tomcur:
As a stylistic point, these could be written as `x * 16` directly. The compiler does the optimization.

PR #1388, opened 2026-01-29
vello_hybrid: Add two clarifying comments and tweak an if condition

laurenz-canva:
That part of the shader was confusing me a bit, maybe this way it's a bit clearer what's going on?

### DIFF
--- a/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
@@ -169,7 +169,7 @@ struct StripInstance {
     // width = dense_width. If the StripInstance represents a sparse fill region, then dense_width = 0.
     // TODO: In the future, this could be optimized such that `width` always represents the width and a simple
     // 1-bit flag is used to distinguish between sparse fill region and strip. This frees up 15 other bits.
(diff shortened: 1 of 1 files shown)


LaurenzV:
Good point! That would indeed make a lot of sense.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -275,9 +280,11 @@ var clip_input_texture: texture_2d<f32>;
 fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
     let x = u32(floor(in.tex_coord.x));
     var alpha = 1.0;
-    // Determine if the current fragment is within the dense (alpha) region
-    // If so, sample the alpha value from the texture; otherwise, alpha remains fully opaque (1.0)
-    if x < in.dense_end {
```
grebmeg, line 280: I think this is actually more aligned with the approach @tomcur mentioned [here](https://github.com/linebender/vello/pull/1388#pullrequestreview-3722560825), but if we get to that optimization, we can revisit this.

LaurenzV: Yeah, that definitely makes sense. It just wasn’t clear to me that this was the intention since it isn’t actually implemented yet. 😄

PR #1393, opened 2026-01-29
vello_hybrid: Remove alpha staging buffer

laurenz-canva:
# Context

When uploading our alpha values to the texture, we need to ensure it has the same size as the texture itself (which is always padded to the maximum width of a texture). Therefore, up until now we kept a separate buffer that is always aligned to the width, and copied the alpha values into the staging buffer before uploading it. 

# Solution

As was noted in the code, this can be changed such that we upload the values directly from the original alpha storage by resizing it temporarily, uploading it and then truncating it back. I chose the route of wrapping the strip storage into a `RefCell` (since it's only rarely accessed, this doesn't have any noticeable impact at runtime). An alternative would be to force users to provide a `&mut Scene` when rendering, but in my opinion this is the less favorable solution.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -1412,6 +1412,10 @@ impl Programs {
 
     /// Upload alpha data to the texture.
     fn upload_alpha_texture(&mut self, queue: &Queue, alphas: &mut Vec<u8>) {
+        if alphas.is_empty() {
+            return;
+        }
+
         let texture_width = self.resources.alphas_texture.width();
         let texture_height = self.resources.alphas_texture.height();
         let total_size = texture_width as usize * texture_height as usize * 16;


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -902,45 +894,45 @@ impl WebGlPrograms {
...
 
     /// Upload alpha data to the texture.
-    fn upload_alpha_texture(&mut self, gl: &WebGl2RenderingContext, alphas: &[u8]) {
-        if !alphas.is_empty() {
-            let alpha_texture_width = self.resources.max_texture_dimension_2d;
-            let alpha_texture_height = self.resources.alpha_texture_height;
+    fn upload_alpha_texture(&mut self, gl: &WebGl2RenderingContext, alphas: &mut Vec<u8>) {
+        if alphas.is_empty() {
+            return;
+        }
```
grebmeg, line 900: Should we add a similar check in the equivalent wgpu method?

grebmeg (approved):
LGTM

sparse_strips/vello_hybrid/src/render/webgl.rs
laurenz-canva: Yeah I noticed this as well, can add it, I just didn't because it wasn't there originally.

PR #1395, opened 2026-01-29
vello_hybrid: Create a reusable state for tiles and buffers

laurenz-canva:
Before (1110ms):
image

After (863ms):
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -1097,20 +1097,18 @@ fn prepare_cmds(cmds: &[Cmd], state: &mut SchedulerState) {
         match cmd {
             Cmd::PushBuf(_layer_id) => {
                 // TODO: Handle layer_id for filter effects when implemented.
-                state
-                    .pointer_to_push_buf_stack
-                    .push(annotated_commands.len());
+                pointer_to_push_buf_stack.push(annotated_commands.len());
             }
             Cmd::PopBuf => {
-                state.pointer_to_push_buf_stack.pop();
+                pointer_to_push_buf_stack.pop();
             }
             Cmd::Blend(_) => {
                 // For blending of two layers to work in vello_hybrid, the two slots being blended
                 // must be on the same texture. Hence, annotate the next-on-stack (nos) tile such
                 // that it uses a temporary slot on the same texture as this blend.
-                if state.pointer_to_push_buf_stack.len() >= 2 {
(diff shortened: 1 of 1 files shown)


laurenz-canva:
I've refactored this now to make it a less invasive change and also include the temporary buffers that are used for annotated commands!

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -247,12 +247,38 @@ struct Round {
...
+}
+
 /// State for a single wide tile.
 #[derive(Debug, Default)]
 struct TileState {
     stack: Vec<TileEl>,
 }
 
+impl TileState {
+    fn reset(&mut self) {
```
grebmeg: [nit]: Maybe also `clear`?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1052,37 +1081,43 @@ fn has_non_zero_alpha(rgba: u32) -> bool {
...
                 // must be on the same texture. Hence, annotate the next-on-stack (nos) tile such
                 // that it uses a temporary slot on the same texture as this blend.
-                if pointer_to_push_buf_stack.len() >= 2 {
+                if state.pointer_to_push_buf_stack.len() >= 2 {
                     let push_buf_idx =
-                        pointer_to_push_buf_stack[pointer_to_push_buf_stack.len() - 2];
+                        state.pointer_to_push_buf_stack[state.pointer_to_push_buf_stack.len() - 2];
                     annotated_commands[push_buf_idx] = AnnotatedCmd::PushBufWithTemporarySlot;
                 }
             }
```
grebmeg, line 1114:
```suggestion
            Cmd::PushBuf(_layer_id) => {
                // TODO: Handle layer_id for filter effects when implemented.
                pointer_to_push_buf_stack.push(annotated_commands.len());
            }
            Cmd::PopBuf => {
                pointer_to_push_buf_stack.pop();
            }
            Cmd::Blend(_) => {
                // For blending of two layers to work in vello_hybrid, the two slots being blended
                // must be on the same texture. Hence, annotate the next-on-stack (nos) tile such
                // that it uses a temporary slot on the same texture as this blend.
                if pointer_to_push_buf_stack.len() >= 2 {
                    let push_buf_idx =
                        pointer_to_push_buf_stack[pointer_to_push_buf_stack.len() - 2];
                    annotated_commands[push_buf_idx] = AnnotatedCmd::PushBufWithTemporarySlot;
                }
            }
```

PR #1399, opened 2026-02-02
vello_common: Remove the `tolerance` parameter

laurenz-canva:
We were just passing `TOL` anyway, and I don't think we'll make this configurable, so let's just use the constants directly.

### DIFF
--- a/sparse_strips/vello_common/src/flatten.rs
+++ b/sparse_strips/vello_common/src/flatten.rs
@@ -12,7 +12,7 @@ use log::warn;
 pub use crate::flatten_simd::FlattenCtx;
 
 // The current tolerance is set to 0.25. Since `sqrt` doesn't work in const contexts, we instead
-// hardcode the squared tolerance and derive the other's from that.
+// hardcode the squared tolerance and derive the others from that.
 pub(crate) const SQRT_TOL: f64 = 0.5;
 pub(crate) const TOL: f64 = SQRT_TOL * SQRT_TOL;
 pub(crate) const TOL_2: f64 = TOL * TOL;


sparse_strips/vello_common/src/flatten.rs
```diff
@@ -12,7 +12,8 @@ use log::warn;
 pub use crate::flatten_simd::FlattenCtx;
 
 /// The flattening tolerance.
-const TOL: f64 = 0.25;
+pub(crate) const TOL: f64 = 0.25;
+pub(crate) const SQRT_TOL: f64 = 0.5;
 pub(crate) const TOL_2: f64 = TOL * TOL;
```
tomcur, line 18: Take it or leave it, but to ensure this remains in sync this could be written as follows (without requiring a `f64::sqrt`, which isn't `const`). Downside is it's harder to read the value of `TOL`.

```rust
pub(crate) const SQRT_TOL: f64 = 0.5;
pub(crate) const TOL: f64 = SQRT_TOL * SQRT_TOL;
pub(crate) const TOL_2: f64 = TOL * TOL;
```

sparse_strips/vello_common/src/flatten.rs
```diff
@@ -11,9 +11,10 @@ use log::warn;
 
 pub use crate::flatten_simd::FlattenCtx;
 
-/// The flattening tolerance.
-pub(crate) const TOL: f64 = 0.25;
+// The current tolerance is set to 0.25. Since `sqrt` doesn't work in const contexts, we instead
+// hardcode the squared tolerance and derive the other's from that.
```
tomcur: Typo:
```suggestion
// hardcode the squared tolerance and derive the others from that.
```

PR #1400, opened 2026-02-02
Canvas composite stateful

waywardmonkeys:
This is for @raphlinus to see.


raphlinus:
@sagudev You probably want to take a look at this, I'd love to hear your comments. I'll be taking a closer look myself soon. A major advantage of this approach is that it doesn't consume any memory for blend stack buffers.

waywardmonkeys:
Winding down my work in this area. Someone else can pick it up if they like.

PR #1401, opened 2026-02-02
vello_hybrid: Add some comments to scheduling logic

laurenz-canva:
This where some of the things missing for me when trying to understand the (clipping-side) of the code, so hopefully this can be helpful to others. It would be good to have someone with more knowledge double-check what I wrote, though.

I haven't looked at blending yet.

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -420,11 +420,17 @@ impl Scheduler {
 
         let slot_ix = self.free[texture].pop().unwrap();
 
-        match texture {
-            0 => Ok(ClaimedSlot::Texture0(slot_ix)),
-            1 => Ok(ClaimedSlot::Texture1(slot_ix)),
+        let slot = match texture {
+            0 => ClaimedSlot::Texture0(slot_ix),
+            1 => ClaimedSlot::Texture1(slot_ix),
             _ => panic!("invalid slot texture"),
-        }
+        };
+
+        // Since the slot was claimed, it needs to be cleared in the given round.
+        let round = self.get_round(self.round);
+        round.clear[slot.get_texture()].push(slot.get_idx() as u32);
+
+        Ok(slot)
     }
 
     pub(crate) fn do_scene<R: RendererBackend>(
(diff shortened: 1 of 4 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -522,6 +522,8 @@ impl Scheduler {
             };
 
             if draw.0.is_empty() {
+                // TODO: Isn't this only necessary if round.clear[i].len() was > 0? i.e. if all
```
laurenz-canva: @ajakubowicz-canva Do you know if my thinking is correct here?

ajakubowicz-canva: My memory is fuzzy here, but there are two cases where `round.clear[i].len() was > 0`.

`load` is `LoadOp::Clear` when `round.clear[i].len() + self.free[i].len() == self.total_slots` is fulfilled. In this case we are saying, clear the whole slot texture before we raster to it. However – this will only clear the slot texture on the condition there is drawing to do. Thus, this block is enforcing that we clear the slots even if no drawing is to take place.

If "round.clear[i].len() is > 0" but the above condition is not fulfilled, then [line 519](https://github.com/linebender/vello/pull/1401/changes#diff-0030050b806d0dfbe6da8852b72f5d26b5a4d8ea3276a4d9159919bd8cae419fR519) has already had the side-effect of clearing those slots.

In short, we need this code to force the full slot texture to clear in the absence of a `render_strips` render pass.

laurenz-canva: Yes, but what I meant is:

As, you mentioned, `load` is `LoadOp::Clear` when `round.clear[i].len()` + `self.free[i].len()` == `self.total_slots`. Indeed, we need this condition for `round.clear[i].len() is > 0`, but it can also happen that `round.clear[i].len()` is 0 and `self.free[i].len()` == `self.total_slots` (in the very first round, when we haven't drawn anything yet and thus there is nothing to clear), and in this case we will still clear the slots. Because we just check `load == LoadOp::Clear` instead of `load == LoadOp::Clear && round.clear[i].len() > 0`, even though it shouldn't be necessary, if I'm understanding correctly.

But anyway, that's just a small edge case, so it's probably not worth worrying about. I'll remove the comment, thanks! :)

ajakubowicz-canva: Oh!!!! Yes – you are totally right!

grebmeg (approved):

Issue #1383, opened 2026-01-28
Lazy tile push_buf for blending


raphlinus:
In the current code, whenever there's a `Wide::push_layer()` that requires a buffer (mostly blending), there's an eager evaluation of the tile `push_buf()` method for each of the wide tiles in the viewport, then a corresponding pop. That's often wasteful, and is the cause of poor performance in scenes that have multiple blend layers (see #1218 and #1380 in particular). It's easy to observe poor performance on the blend example scene, especially when the viewport becomes large, even when the content is not zoomed in.

We should be lazy, and only push the buffer when draw operations are actually placed in the layer. Fortunately, the logic for that is not all that complicated; here's a sketch.

Add an additional field to the `WideTile` struct, which represents the actual blend depth of that tile (number of PushBuf commands minus number of PopBuf commands). Also add a blend depth to `Wide`, which is incremented on each `push_layer` with a blend and decremented on pop. An invariant is that the blend depth of each tile is less than or equal to the global blend depth.

I think we also need to add a `layer_kind` field to the `Layer` struct, as we will need that to perform the lazy pushing. The layer also gains a list of pushed tiles.

The `push_layer()` simply increments the global blend depth and doesn't do any per-tile operations.

All drawing operations into tiles do a lazy blend check; if the tile's blend depth is less than the global blend depth, then it reaches into the layer stack and does a `push_buf()` for each layer in the difference, adding the tile to the list of pushed tiles in the layer, and updating the blend depth in the tile.

The `Wide::pop_layer()` method does a blend and `pop_buf()` for each tile in the pushed tiles list; this is very similar to the current logic which does it for all tiles, is just selective.

The above is a bit simplified; I haven't worked out all the interactions with filters and masks, but I'm not seeing any conceptual problems. It also might be overly aggressive when the compositing mode is one where a transparent pixel in the source changes the destination (as opposed to leaving it unchanged as in the most common source-over operation). I believe these are rarely used, but also likely that they can be optimised further, as they seem to clear the destination in that case.

The goal of the above logic is to eliminate empty push/pop pairs with no drawing in between. Thinking about this a bit more, blends can also be omitted when the under-layer (the previous state of "dest" in Porter-Duff terminology) is known to be all transparent. I think it's likely that would further reduce the amount of blending.

nicoburns:
Does the logic above account for clipping? (it mentions the viewport, but not the current clip). I think it would be common to have a layer that only takes up a small portion of the screen.

[...]

Issue #1380, opened 2026-01-26
[Hybrid] Slow rendering of https://old.reddit.com/r/rust


nicoburns:
### Problem

When rendering https://old.reddit.com/r/rust with Blitz, I'm getting frame times of:

- ~110ms with Vello Hybrid
- ~36ms with Vello CPU
- ~18ms with Vello Classic
- ~6ms with Skia (although there is one 215ms frame right at the start)

These frame times are only for the "paint pass" (constructing the scene, submitting it, and waiting on presentation). They do not include style, layout or any other non-rendering work.

### To reproduce

- Clone https://github.com/DioxusLabs/blitz
- Checkout commit `bdbe4d3f2a1d8bd3e7b77cd163bb4de280738550` (Vello Hybrid is broken for commits beyond that point for some reason that seems to be related to the WInit 0.31 upgrade,)
- Run `cargo run -rp browser --no-default-features --features floats,incremental,cookies,cache,log_frame_times,hybrid`
- Navigate to https://old.reddit.com/r/rust using the URL bar
- Resize the window to be approx 1000x1000 (on my macbook laptop I have it full height, and approximately square)
- Remain scrolled to the top of the webpage and move mouse over the page to generate new frames
- The `hybrid` on the end of that command can be replaced with `vello`, `cpu`, or `skia` to run with those renderers for comparison

### Screenshot

<img width="1512" height="945" alt="Image" src="https://github.com/user-attachments/assets/3991f7ef-cee9-45de-9bb9-1c1d7cc6b588" />

raphlinus:
I'm trying to repro under debug observation, and running into challenges. When I run it as "debug executable" in Xcode, the content doesn't render (though the URL bar does).

When I try vello_hybrid_winit, and it does display content and I am able to see meaningful GPU traces.

I'm also trying to do CPU profiling in Instruments, and not seeing any debug symbols even though I'm doing `--profile profile`, but this is probably operator error of some kind. I've made sure that strip=false and debug=true in the profile, and objdump is showing symbols.

I am able to run blitz using "cargo instruments" but it has a strange hybrid of the two behaviors: I *am* seeing symbols just fine, but the content is black, and the reported render times (and profile) suggest it isn't rendering any challenging content.

<img width="1381" height="634" alt="Image" src="https://github.com/user-attachments/assets/5467e5a0-905e-4fa7-b2ab-baf2d8d0b794" />

[...]

PR #1403, opened 2026-02-02
sparse strips: Stop generating layer operations for empty buffers

tomcur:
Currently, when a layer requires its own buffer but that buffer ends up with no content, we still emit opacity/blend/mask and a `PopBuf`. That means we're effectively generating `O(viewport_width x viewport_height)` commands for every blend layer.

This suggests checking when popping a layer whether there is any actual content in the buffer, and if there isn't, we don't emit `mask` or `opacity` commands and simply pop the buffer without emitting layer operations. The similar check that was already done for `blend` commands has been merged into this. A more involved change that also stops the viewport iteration during command generation (comparatively not that expensive) will likely be based on https://github.com/linebender/vello/issues/1383.

This brings down the timings of rendering `https://old.reddit.com/r/rust` with Blitz mentioned in https://github.com/linebender/vello/issues/1380 from 54ms to 16ms on my desktop, and 195ms to 50ms on my laptop.

Desktop (AMD Radeon RX 7800 XT)
|             | total | cmd | render | present | wait |
| ---       | --- | --- | --- | --- | --- |
| Before | 54ms | 21ms | 19ms | 0.029ms | 15ms |
| After  | 16ms | 15ms | 0.78ms | 0.018ms | 0.84ms |

Laptop (AMD Radeon 860M)
|             | total | cmd | render | present | wait |
| ---       | --- | --- | --- | --- | --- |
| Before | 195ms | 43ms | 63ms | 0.41ms | 88ms |
| After  | 50ms | 40ms | 2.4ms | 0.10ms | 7.5ms |

```
Before:
vello_hybrid: 54ms (cmd: 21ms, render: 19ms, present: 29us, wait: 15ms)

After:
vello_hybrid: 16ms (cmd: 15ms, render: 776us, present: 18us, wait: 841us)
```

There's some noise in `cmd`, I believe we should mostly expect `render` and `wait` to change.

### DIFF
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -18,6 +18,7 @@ This release has an [MSRV][] of 1.88.
 ### Changed
 
 - Improved Bézier flattening performance by catching more Béziers whose chords are immediately within rendering tolerance. ([#1216][] by [@tomcur][])
+- Significantly improved rendering performance of scenes including blend layers by ensuring no commands are generated for wide tiles without layer content. ([#1399][] by [@tomcur][])
 
 ## [0.0.6][] - 2026-01-15
 
@@ -119,6 +120,7 @@ See also the [vello_cpu 0.0.1](../vello_cpu/CHANGELOG.md#001---2025-05-10) relea
 [#1349]: https://github.com/linebender/vello/pull/1349
 [#1353]: https://github.com/linebender/vello/pull/1353
 [#1354]: https://github.com/linebender/vello/pull/1354
+[#1399]: https://github.com/linebender/vello/pull/1399
 
 [Unreleased]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.6...HEAD
 [0.0.6]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.5...sparse-strips-v0.0.6


LaurenzV:
Is my understanding correctly that, thanks to us now not pushing blend/opacity commands if nothing has been drawn yet, the following optimization now triggers properly:

https://github.com/linebender/vello/blob/aed93cba17c2ac76dce21d3992be9c895d27e02f/sparse_strips/vello_common/src/coarse.rs#L1437-L1440

And that's where the speed boost comes from?

tomcur:
> Is my understanding correctly that, thanks to us now not pushing blend/opacity commands if nothing has been drawn yet, the following optimization now triggers properly:
> 
> https://github.com/linebender/vello/blob/aed93cba17c2ac76dce21d3992be9c895d27e02f/sparse_strips/vello_common/src/coarse.rs#L1437-L1440
> 
> And that's where the speed boost comes from?

Yes. We *may* be able to do something simpler... Perhaps just optimize opacity handling with something like:

```diff
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -1464,9 +1464,11 @@

     /// Apply an opacity to the whole buffer.
     pub fn opacity(&mut self, opacity: f32) {
-        if opacity != 1.0 {
-            self.cmds.push(Cmd::Opacity(opacity));
+        if opacity == 1.0 || matches!(self.cmds.last(), Some(&Cmd::PushBuf(_))) {
+            return;
         }
+
+        self.cmds.push(Cmd::Opacity(opacity));
     }
```

I don't have my head fully wrapped around the layer handling yet.

dominikh:
Is the cost of pushing buffers for all tiles negligible? I would have thought we'd want to lazily push buffers (and then skip generating layer operations if the buffer was never pushed.) Though the difference is only really noticeable for large--i.e. mostly unclipped--layers.

nicoburns:
Damn. This is a big improvement!

And I want to highlight it's very uneven in terms of which phase of the rendering is affected. The "cmd" phase has improved (21ms -> 15ms), but the "render" and "wait" phases have both collapsed from  19ms/15ms to <1ms!

Your numbers above formatted into a table:

|             | cmd | render | present | wait |
| ---       | --- | --- | --- | --- |
| Before | 21ms | 19ms | 0.029ms | 15ms |
| After  | 15ms | 0.78ms | 0.018ms | 0.84ms |

tomcur:
> Is the cost of pushing buffers for all tiles negligible? I would have thought we'd want to lazily push buffers (and then skip generating layer operations if the buffer was never pushed.) Though the difference is only really noticeable for large--i.e. mostly unclipped--layers.

It's not a huge cost at least, I have a rough local change doing that and didn't notice much effect on command generation time; that said, it is perhaps more principled to do lazy pushing and merge that logic with deciding to generate the pop commands (rather than explicitly tracking having content).

LaurenzV:
> Yes. We may be able to do something simpler... Perhaps just optimize opacity handling with something like:

Yes, I just had the same thought and it seems to work: https://github.com/linebender/vello/compare/main...laurenz/opt

I think this would be the preferable variant since it doesn't require us to keep track of whether the buffer has content manually. As long as we don't touch filter layers (for now), this _should_ be safe.

tomcur:
> I think this would be the preferable variant since it doesn't require us to keep track of whether the buffer has content manually. As long as we don't touch filter layers (for now), this should be safe.

That sounds about right, also see @dominikh's comment and my response above about lazy pushing, as that is and would still be doing `O(viewport size)` of (admittedly nearly-trivial) work.

LaurenzV:
Sure, but I think we can merge that as a first step and then refine it later to do actual lazy pushing, right? Given the huge performance improvement this brings. 😄

tomcur:
> Sure, but I think we can merge that as a first step and then refine it later to do actual lazy pushing, right? Given the huge performance improvement this brings. 😄

Agreed. I also experimented a bit more with lazy pushing and it needs more careful thought, so shouldn't block this.


> Your numbers above formatted into a table: [...]


Thanks! I've updated the PR text above to include the table, and included numbers generated on my laptop as well.

tomcur:
I've simplified this by removing the tracking in favor of a check during layer popping that any commands have been emitted. That technically might be able to catch less if we add stuff in the future that emits commands without adding content, but that should probably be caught once lazy buffer pushing is done anyway.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1475,17 +1484,13 @@ impl<const MODE: u8> WideTile<MODE> {
...
     /// Blend the current buffer into the previous buffer in the stack.
-    pub fn blend(&mut self, blend_mode: BlendMode) {
-        // Optimization: If no drawing happened since the last `PushBuf` and the blend mode
-        // is not destructive, we do not need to do any blending at all.
-        if !matches!(self.cmds.last(), Some(&Cmd::PushBuf(_))) || blend_mode.is_destructive() {
-            self.cmds.push(Cmd::Blend(blend_mode));
-        }
+    fn blend(&mut self, blend_mode: BlendMode) {
+        self.cmds.push(Cmd::Blend(blend_mode));
     }
```
tomcur, line 1494: I've made some methods private, as I believe these should really only be called internally by methods like `pop_layer`. As-written, the old code would be buggy if someone would call `blend` manually as the first command inside a new layer.

taj-p:
This is amazing!!!

tomcur:
Ran out of time for today, will do a last check and merge this PR in my morning. I have an experimental version of lazy pushing here: https://github.com/tomcur/vello/tree/lazy-push-buf.

nicoburns:
It makes sense seeing as the fix is in `vello_common`, but this is also a significant speedup 33ms -> 19ms for `vello_cpu` on the Old Reddit scene.

|             | total | cmd | flush | render | present |
| ---       | --- | --- | --- | --- | --- |
| Before | 33ms | 12ms | 0ms | 18ms | 2.5ms |
| After  | 19ms | 11ms | 0ms | 6ms | 2.4ms |

There is noise in `cmds` so that is effectively unchanged. And `present` is external to `vello_cpu` (measuring the time for the pixels crate to present the pixel buffer to screen). But 18ms -> 6ms for the "render" stage is a huge improvement.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -765,11 +765,20 @@ impl<const MODE: u8> Wide<MODE> {
                 for y in 0..self.height_tiles() {
```
grebmeg, line 765: It might be a good next step to explore whether `Layer::wtile_bbox` can be used here, something like the following:

```rust
let bbox = layer.wtile_bbox;
    
// Only iterate over tiles that actually have content
for x in bbox.x_min..=bbox.x_max {
    for y in bbox.y_min..=bbox.y_max {
        let t = self.get_mut(x, y);
        // ... mask/opacity/blend logic ...
        t.pop_buf();
    }
}
```

LaurenzV: Yeah, now that I think about it, the same is probably also applicable to filter layers (since the bbox already contains the necessary expansion for filters, if I'm reading correctly). But probably best to do this in a follow-up.

PR #1409, opened 2026-02-02
vello_hybrid: Use a pool for storing `Round` objects

laurenz-canva:
A round object stores a number of vectors for tracking the state of the round: 

https://github.com/linebender/vello/blob/6a45f75661129aa3dcbfeeaed5bbdae2451b880e/sparse_strips/vello_hybrid/src/schedule.rs#L236-L248

In particular, for each texture slot it holds a `Draw` object, which itself stores all of the strips that are needed for that round. Since it's quite a common case that all draws are scheduled for the first round, this vector can get pretty big.

Currently, each time we want to create a new round, we do so by initializing a new one via `Round::default`:
https://github.com/linebender/vello/blob/6a45f75661129aa3dcbfeeaed5bbdae2451b880e/sparse_strips/vello_hybrid/src/schedule.rs#L552-L554

Once the round is flushed, the round object is popped from the queue and then simply discarded:
https://github.com/linebender/vello/blob/6a45f75661129aa3dcbfeeaed5bbdae2451b880e/sparse_strips/vello_hybrid/src/schedule.rs#L492-L493

Therefore, in each frame we have to recreate the rounds (and their allocations) from new.

This PR proposes to store a small allocator pool for `Round` objects in `Scheduler`. Each time we need a new round, we first try to fetch an existing object from the pool, and once the round is popped we submit it back to the pool. This way, the objects can be reused.

The reduction of allocations, while not a huge win in terms of performance, is also visible in the profile:

Before:
image

After:
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -246,7 +246,7 @@ impl RoundPool {
 
         // Avoid caching too many objects in adversarial scenarios.
         if self.0.len() < MAX_ELEMENTS {
-            // Make sure the round is resetted if we reuse it in the future.
+            // Make sure the round is reset if we reuse it in the future.
             round.clear();
 
             self.0.push(round);


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -231,6 +231,23 @@ pub(crate) struct Scheduler {
     free: [Vec<usize>; 2],
     /// Rounds are enqueued on push clip commands and dequeued on flush.
     rounds_queue: VecDeque<Round>,
+    /// A pool of `Round` objects that can be reused, so that we can reduce
+    /// the number of allocations.
+    round_pool: Vec<Round>,
+}
+
+impl Scheduler {
+    fn submit_round(&mut self, round: Round) {
```
taj-p: Could these methods be inline(always)?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -231,6 +231,23 @@ pub(crate) struct Scheduler {
...
+        self.round_pool.push(round);
+    }
+
+    fn fetch_round(&mut self) -> Round {
+        let mut round = self.round_pool.pop().unwrap_or_default();
+        // Reset it in case it's a reused one.
+        round.clear();
```
taj-p: Should we clear when we submit the round instead?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -231,6 +231,23 @@ pub(crate) struct Scheduler {
...
     /// Rounds are enqueued on push clip commands and dequeued on flush.
     rounds_queue: VecDeque<Round>,
+    /// A pool of `Round` objects that can be reused, so that we can reduce
+    /// the number of allocations.
+    round_pool: Vec<Round>,
+}
+
+impl Scheduler {
+    fn submit_round(&mut self, round: Round) {
+        self.round_pool.push(round);
```
taj-p: Did we want to put a maximum ceiling of rounds stored so this isn't unbounded?

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Submit makes it sound like round is being passed to something that will use it. Maybe it's worth using a tiny `RoundPool` struct to also compartmentalize the maximum number of rounds stored? 

Otherwise we can simply rename these methods to include "pool" like "return_round_to_pool" and "take_round_from_pool"

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Sure, any suggestion?

taj-p: So, rounds are returned to the pool when we flush. We flush when we have no more slots for rendering. For clip depth >= 3, I believe it's something like every even depth, there's an additional round. So 32 would give us ~60 depth, but considering these are relatively cheap, we could upper bound on 64 (for generous headroom) and only very complex scenes will see their pools grow that large.

LaurenzV: I think such large clip depths are very uncommon, no? 🤔 Maybe 10 or something is enough?

taj-p: That works for me! 

edit: The pool should grow to accommodate the scene (and stay small for simple scenes), so, I'm not too concerned with a large value - my only concern is it being unbounded. That said, I'm happy with 10. For my use case, I'd be surprised if it hits 2 😅 !

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -231,6 +231,23 @@ pub(crate) struct Scheduler {
...
+    }
+
+    fn fetch_round(&mut self) -> Round {
+        let mut round = self.round_pool.pop().unwrap_or_default();
+        // Reset it in case it's a reused one.
+        round.clear();
+
+        round
+    }
```
grebmeg: [nit] Maybe like below, since calling `clear()` is unnecessary, `Default` already creates empty vectors?

```rust
fn fetch_round(&mut self) -> Round {
    match self.round_pool.pop() {
        Some(mut round) => {
            round.clear();
            round
        }
        None => Round::default(),
    }
}
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -247,6 +264,20 @@ struct Round {
...
+        self.draws[0].clear();
+        self.draws[1].clear();
+        self.draws[2].clear();
+
+        self.free[0].clear();
+        self.free[1].clear();
+
+        self.clear[0].clear();
+        self.clear[1].clear();
+    }
```
grebmeg, line 290: [nit] What do you think about this?

```suggestion
        for draw in &mut self.draws {
            draw.clear();
        }
        for free in &mut self.free {
            free.clear();
        }
        for clear in &mut self.clear {
            clear.clear();
        }
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -247,6 +264,20 @@ struct Round {
     clear: [Vec<u32>; 2],
 }
 
+impl Round {
+    fn clear(&mut self) {
```
grebmeg, line 278: If we inline this as well would it give any benefits?

laurenz-canva: Probably not, but I can do it anyway.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: I've applied Taj's suggetsion of clearing upon submitting instead.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: Done!

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: I've done just `inline`, I think this should be enough.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: Done!

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: Done.

PR #1414, opened 2026-02-03
sparse strips: Lazily push wide tile layer buffers

tomcur:
This continues on https://github.com/linebender/vello/pull/1403, and is based on/closes https://github.com/linebender/vello/issues/1383.

In `coarse.rs` we are doing `O(viewport_width x viewport_size)` work for pushing and popping buffers for each layer needing buffers. PR #1403 already ensures we aren't actually sending commands for buffer blending through to rendering anymore for empty buffers. This PR makes command generation lazy so that we also don't do the `push_buf` work for the entire viewport anymore, but only for the wide tiles that actually get drawn on. This can significantly cut down on `coarse` timings (up to -80% in benchmarks below, testing some reasonable patterns).

Some operations still fall back to pushing buffers for the full viewport, like filter layers and destructive blends. We may be able to do away with some of that in the future, but for now, those operations require buffers even if they don't have content (doing away or tightening the condition causes various tests to fail or asserts to be hit).

### DIFF
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -19,6 +19,7 @@ This release has an [MSRV][] of 1.88.
 
 - Improved Bézier flattening performance by catching more Béziers whose chords are immediately within rendering tolerance. ([#1216][] by [@tomcur][])
 - Significantly improved rendering performance of scenes including blend layers by ensuring no commands are generated for wide tiles without layer content. ([#1399][] by [@tomcur][])
+- Further improved performance of scenes including blend layers by only allocating layer scratch buffers for a wide tile upon draw commands being performed in that wide tile. ([#1414][] by [@tomcur][])
 - Improved flattening and tiling performance by culling out-of-viewport Béziers before flattening. ([#1341][] by [@tomcur][])  
   This drops all out-of-viewport geometry that does not impact winding or pixel coverage (i.e., geometry above, to the right, or below the viewport).
   For geometry that is to the left of the viewport (impacting winding), this skips flattening by directly yielding the Bézier's chord.
@@ -129,6 +130,7 @@ See also the [vello_cpu 0.0.1](../vello_cpu/CHANGELOG.md#001---2025-05-10) relea
 [#1353]: https://github.com/linebender/vello/pull/1353
 [#1354]: https://github.com/linebender/vello/pull/1354
 [#1399]: https://github.com/linebender/vello/pull/1399
+[#1414]: https://github.com/linebender/vello/pull/1414
 [#1422]: https://github.com/linebender/vello/pull/1422
 
 [Unreleased]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.6...HEAD


tomcur:
(Getting rid of the subtraction (`n_bufs - n_clip`) makes no substantial difference.)

LaurenzV:
Hmm, I guess some of those gains do look pretty nice. If we decide to merge some variant of https://github.com/linebender/vello/pull/1454, the "no layer" case would also become irrelevant since we would completely skip coarse rasterization (at least for vello_hybrid).

However, would like to hear what @taj-p or @grebmeg think before reviewing this PR more deeply, as it does add some more complexity.

laurenz-canva:
Wow, I just benchmarked this more carefully and this is a huge win!! On a scene with 750 rectangles, each wrapped in an opacity layer, rendering for 150 frames gives me the following number:

Before:
image

After:
image

Since the no-layer case will be optimized to completely circumvent coarse rasterization anyway, I think that's a trade-off worth making!

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -63,6 +63,64 @@ pub const MODE_CPU: u8 = 0;
 /// generation specific for `vello_hybrid`.
 pub const MODE_HYBRID: u8 = 1;
 
+#[derive(Debug)]
+struct LayerKindAndOccupiedTiles {
+    /// The
```
LaurenzV: ? 😄

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -63,6 +63,64 @@ pub const MODE_CPU: u8 = 0;
...
+        } else {
+            self.stack[self.len].kind = kind;
+        }
+        self.len += 1;
+    }
+
+    fn pop(&mut self) {
+        debug_assert!(self.len > 0, "Called `pop` at the root");
+        self.len -= 1;
+        self.stack[self.len].occupied_tiles.clear();
```
LaurenzV: Would it make sense to move `clear` to `push` instead of `pop`? Not sure.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -311,9 +374,44 @@ impl<const MODE: u8> Wide<MODE> {
...
+            layer_bufs <= self.layers_needing_buf_stack.len,
+            "tile `layer_buf_depth` exceeds active layer stack"
+        );
+        // It may be quite likely that no buffers need to be pushed: e.g. a tile that has content
+        // may well have more than one call to `ensure_layer_stack_bufs`, and layers needing
+        // buffers are not necessarily the most common case. As such, we keep this check inlined
+        // (because the function itself is inlined), and if buffers are needed do an explicit
+        // function call that is not inlined. That keeps the generated code size small at our call
+        // sites.
+        (layer_bufs < self.layers_needing_buf_stack.len).then(
```
LaurenzV: I feel like this would be better expressed as a simple if statement instead of the closure.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -63,6 +63,64 @@ pub const MODE_CPU: u8 = 0;
...
+///
+/// The wide tile push buffer command [`WideTile::push_buf`] is performed lazily upon actually
+/// being drawn into.
+///
+/// See [`LayerKindAndOccupiedTiles`].
+///
+/// This is a small wrapper around [`Vec`] so we can keep the
+/// [`LayerKindAndOccupiedTiles::occupied_tiles`] allocations around.
+#[derive(Debug, Default)]
+struct NeedsBufLayerStack {
```
LaurenzV, line 91: Not for this PR, but just a drive-by comment, I realized something like this would also be useful for the scratch buffers that are allocated in vello_cpu. It might be worth turning this into a small vector abstraction so that it can be reused in multiple places.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -311,9 +374,44 @@ impl<const MODE: u8> Wide<MODE> {
...
+    /// Push all layer buffers that have not yet been pushed for the tile at index `tile_idx`.
+    ///
+    /// We push wide tile layer buffers lazily. Calling this before performing other commands on a
+    /// wide tile ensures all layer stack buffers are pushed for the wide tile. Calling this
+    /// multiple times won't push additional buffers beyond the first call (until the layer stack
+    /// changes).
+    #[inline(always)]
+    fn ensure_layer_stack_bufs(&mut self, tile_idx: usize) {
+        let tile = &mut self.tiles[tile_idx];
+        let layer_bufs = tile.n_bufs - tile.n_clip;
```
LaurenzV: I think a comment would be good we why need to do `- tile.n_clip` here

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -502,7 +605,9 @@ impl<const MODE: u8> Wide<MODE> {
                 };
                 x += width;
                 col += u32::from(width);
-                self.get_mut(wtile_x, strip_y).strip(cmd, current_layer_id);
+                let idx = self.get_idx(wtile_x, strip_y);
+                self.ensure_layer_stack_bufs(idx);
+                self.tiles[idx].strip(cmd, current_layer_id);
```
LaurenzV: I think this (as well as other places) is pretty unfortunate, because it's pretty easy to miss that you need to call `ensure_layer_stack_bufs` before calling any method on the wide tile. Would it not be possible to move that method into `WideTile`, and then change each method on the wide tile to take a &mut NeedsBufLayerStack such that it becomes impossible to miss this?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -764,36 +881,43 @@ impl<const MODE: u8> Wide<MODE> {
...
+                .unwrap()
+                .occupied_tiles
+                .iter()
+            {
+                let tile = &mut self.tiles[tile_idx];
+
+                // Optimization: If no drawing happened since the last `PushBuf`, then we don't
+                // need to do any masking or buffer-wide opacity work. The same holds for
+                // blending, unless it is destructive blending.
+                let has_draw_commands = !matches!(tile.cmds.last().unwrap(), &Cmd::PushBuf(_));
```
LaurenzV: Can this even happen now that layers are pushed lazily?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -764,36 +881,43 @@ impl<const MODE: u8> Wide<MODE> {
...
 
         let in_clipped_filter_layer = layer.filter.is_some() && layer.clip;
         // Decrement the depth counter after popping a filtered layer with clip
         if in_clipped_filter_layer {
             self.clipped_filter_layer_depth -= 1;
+            if self.clipped_filter_layer_depth == 0 {
+                for tile in &mut self.tiles {
+                    tile.in_clipped_filter_layer = false;
+                }
+            }
```
LaurenzV: Am I correct that this isn't related to this PR specifically, but a bug you found as part of your work?

LaurenzV (commented):
.

sparse_strips/vello_common/src/coarse.rs
tomcur: We could do an if statement, but the `#[inline(never)]` (one line below) is the important part here, as it reduces the cost for the common case where no buffers need to be pushed (because the code size of `ensure_layer_stack_bufs` becomes smaller at the call sites, trading off with the cost of calling into the inner function if a buffer *does* need to be pushed).

It had a few percentage points of impact in my benchmarks; perhaps you could try running yours with an `#[inline(never)]` and an `#[inline(always)]` here.

LaurenzV: Sorry about that, I completely missed the comment directly above it. 🤦

sparse_strips/vello_common/src/coarse.rs
tomcur: Yes, but it seems this isn't necessary anymore for tests to pass, but I believe it was in an earlier version of this. It may be best to leave it out for now. Do you know if the bool is intended to be `true` precisely when the layer itself is a clipped filter layer, or also when it is any other kind of layer nested in a clipped filter layer?

sparse_strips/vello_common/src/coarse.rs
tomcur: Yes, e.g. destructive blends still need to eagerly push buffers (e.g. copy blending has to copy `(0,0,0,0)` from empty source tiles to the destination).

I've clarified the comment a bit

sparse_strips/vello_common/src/coarse.rs
tomcur: I've done this refactor, and this should actually be slightly more performant in some cases, because the wide tile itself has a bit more information on whether buffers need pushing (e.g., if it's in a zero clip, a fill doesn't need buffers). One unfortunate ergonomic aspect remains in that knowing the wide tile's index is required, so we still need to do a little `let idx = ...; self.tiles[idx].fill(idx, ...)` dance, instead of being able to do `self.get_mut(...).fill(...)`. (A wrapper could work, but perhaps that's just even more mental overhead.)

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -67,6 +67,65 @@ pub const MODE_HYBRID: u8 = 1;
...
+
+    fn pop(&mut self) {
+        debug_assert!(self.len > 0, "Called `pop` at the root");
+        self.len -= 1;
+    }
+
+    fn last(&self) -> Option<&LayerKindAndOccupiedTiles> {
+        self.len.checked_sub(1).map(|idx| &self.stack[idx])
+    }
+}
```
LaurenzV, line 128: Should these be inlined just to be safe?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1261,6 +1359,40 @@ impl<const MODE: u8> WideTile<MODE> {
...
+            "tile `layer_buf_depth` exceeds active layer stack"
+        );
+        // It may be quite likely that no buffers need to be pushed: e.g. a tile that has content
+        // may well have more than one call to `ensure_layer_stack_bufs`, and layers needing
+        // buffers are not necessarily the most common case. As such, we keep this check inlined
+        // (because the function itself is inlined), and if buffers are needed do an explicit
+        // function call that is not inlined. That keeps the generated code size small at our call
+        // sites.
+        (layer_bufs < layers.len).then(
+            #[inline(never)]
```
LaurenzV, line 1387: This is what I get when I turn it into a normal if statement:

```
coarse_with_layer/Ghostscript_Tiger
                        time:   [3.1904 µs 3.2009 µs 3.2124 µs]
                        change: [-6.1956% -5.3240% -4.3456%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 50 measurements (20.00%)
  7 (14.00%) high mild
  3 (6.00%) high severe
```

I think we should just do it? It's just a single for loop after all, so i don't think we need to be too worried about code bloat. :D

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1198,7 +1294,9 @@ pub struct WideTile<const MODE: u8 = MODE_CPU> {
     pub n_zero_clip: usize,
     /// The number of non-zero-winding clips.
     pub n_clip: usize,
-    /// The number of pushed buffers.
+    /// The number of pushed buffers, including buffers for clips.
+    ///
+    /// Note not all layers require their own buffers.
```
LaurenzV: That sentence confuses me a bit. 😄 Why do not all layers require their own buffers?

LaurenzV: Oh right, I see now, basically if `needs_buf` returns `None` then we don't have an explicit layer, right?

sparse_strips/vello_common/src/coarse.rs
tomcur: Interesting... I get these results (on x86) if I turn it into a normal if:

```
coarse_with_layer/Ghostscript_Tiger
                        time:   [5.8312 µs 5.8537 µs 5.8745 µs]
                        change: [+1.8618% +2.2918% +2.7258%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (6.00%) high mild
coarse_with_layer/paris-30k
                        time:   [661.21 µs 666.10 µs 673.03 µs]
                        change: [+1.9323% +3.6281% +5.2749%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (4.00%) high mild
  2 (4.00%) high severe

coarse_with_layer_4k/Ghostscript_Tiger
                        time:   [130.49 µs 133.27 µs 137.10 µs]
                        change: [+8.5701% +10.965% +13.581%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 50 measurements (16.00%)
  1 (2.00%) high mild
  7 (14.00%) high severe
coarse_with_layer_4k/paris-30k
                        time:   [359.92 µs 365.88 µs 373.64 µs]
                        change: [+0.4684% +2.1311% +3.9630%] (p = 0.02 < 0.05)
Found 7 outliers among 50 measurements (14.00%)
  2 (4.00%) high mild
  5 (10.00%) high severe
```

sparse_strips/vello_common/src/coarse.rs
tomcur: > basically if needs_buf returns None then we don't have an explicit layer, right?

Right! Added a documentation link in the comment to that method.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Oh damn... Okay, I suppose let's leave it how it is then.

tomcur: Maybe this is something where [`core::hint::cold_path`](https://doc.rust-lang.org/stable/core/hint/fn.cold_path.html) would be helpful, but it's nightly-only.

tomcur: (Or automatic profile-guided optimization.)

sparse_strips/vello_common/src/coarse.rs
tomcur: Doclink doesn't like that the `WideTile` field is public and the method the docs refers to is private :) maybe all `WideTile` fields should just become private. Only the module itself uses them.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -67,6 +67,65 @@ pub const MODE_HYBRID: u8 = 1;
...
+    /// The actual length of [`Self::stack`], as we never actually shrink it (see the note above
+    /// about keeping allocations around).
+    len: usize,
+}
+
+impl NeedsBufLayerStack {
+    fn clear(&mut self) {
+        for idx in 0..self.len {
+            self.stack[idx].occupied_tiles.clear();
+        }
```
LaurenzV: I'm wondering, do we even need this if we clear upon every push? It's not like this will free the memory, right? And it will also only clear the elements of the current length, not previous entries that we might have popped in the meanwhile.

tomcur: You're right. It was only necessary when we still did the clearing on `pop`.

Issue #1382, opened 2026-01-27
Support `PixmapMut<'_>` and stride in `vello_cpu`?


madsmtm:
Hey! I'm working on [improving Softbuffer](https://github.com/rust-windowing/softbuffer/milestone/2), which you use in `vello_cpu`'s examples to render to a window. I believe that this combination of `winit`, `softbuffer` and either `tiny-skia` or `vello_cpu` is a good match for simple demos and such, and I'll be recommending using your crates in our documentation in https://github.com/rust-windowing/softbuffer/pull/325.

Softbuffer works by creating special shared memory that can be efficiently sent to the compositor, so for it to be zero-copy, it has to be in control of the buffer (it cannot just create its buffers from a `Vec<u8>`). Additionally, for it to be zero-copy on Android and macOS, the buffer needs to have a stride, see https://github.com/rust-windowing/softbuffer/pull/315. Both of these are a bit unfortunate, because it means that integration with `vello_cpu` requires rendering into a separate buffer first.

Would it be possible (and desirable) for `vello_cpu` to support something like `tiny-skia`'s `PixmapMut<'_>`, and to support passing a `byte stride != width * 4` in its creation method?

This is the complementary issue to https://github.com/linebender/tiny-skia/issues/168, see that for a code example of how I imagine an integration would look like in the future.

LaurenzV:
You can already use the [`render_to_buffer` method](https://doc.servo.org/vello_cpu/struct.RenderContext.html#method.render_to_buffer), which should have the same functonality as a `PixmapMut`.

I'm not sure what you mean by `byte stride != width * 4`, you mean that only every xth pixel is used for storing the result? So for example if you have a stride of 4 and you want to store the RGBA value (255, 128, 128, 128), that it should be stored as [255, 0, 0, 0, 128, 0, 0, 0, 128, 0, 0, 0, 128, 0, 0, 0]?

madsmtm:
> You can already use the [`render_to_buffer` method](https://doc.servo.org/vello_cpu/struct.RenderContext.html#method.render_to_buffer), which should have the same functonality as a `PixmapMut`.

Ah, nice, I should've investigated that more thoroughly beforehand!

[...]

PR #1416, opened 2026-02-04
feat: Add render_to_pixmap_at_offset for spritesheet support

grebmeg:
Add the ability to render a `RenderContext`'s content directly into a sub-region of a larger pixmap at a specified (x, y) offset. The method composites on top of existing content rather than clearing, allowing multiple renders to accumulate.

This is useful for building spritesheets where individual elements (like glyphs) are rendered to specific coordinates without intermediate buffer allocation.

An example test result from this PR:
render_to_pixmap_at_offset

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
@@ -399,9 +399,9 @@ impl SingleThreadedDispatcher {
         self.render_graph.has_filters()
     }
 
-    /// Rasterizes at an offset using f32 precision (high quality).
+    /// Composites at an offset using f32 precision (high quality).
     #[cfg(feature = "f32_pipeline")]
-    fn rasterize_at_offset_f32(
+    fn composite_at_offset_f32(
         &self,
         buffer: &mut [u8],
         width: u16,
@@ -414,14 +414,14 @@ impl SingleThreadedDispatcher {
     ) {
         use crate::fine::F32Kernel;
         use vello_common::fearless_simd::dispatch;
-        dispatch!(self.level, simd => self.rasterize_at_offset_with::<_, F32Kernel>(
+        dispatch!(self.level, simd => self.composite_at_offset_with::<_, F32Kernel>(
             simd, buffer, width, height, dst_x, dst_y, dst_buffer_width, dst_buffer_height, encoded_paints
         ));
     }
 
-    /// Rasterizes at an offset using u8 precision (fast).
+    /// Composites at an offset using u8 precision (fast).
     #[cfg(feature = "u8_pipeline")]
-    fn rasterize_at_offset_u8(
+    fn composite_at_offset_u8(
         &self,
         buffer: &mut [u8],
         width: u16,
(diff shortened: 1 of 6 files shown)


nicoburns:
I think this would also solve https://github.com/linebender/vello/issues/1382

grebmeg:
> I think this would also solve https://github.com/linebender/vello/issues/1382

Yeah, that sounds very applicable to that case. I also think it could be useful for #1130 as well. In my case, it’s purely about populating the atlas glyph pixmap so it can be used later for rendering purposes. The next PR I’m preparing after this one will add support for such spritesheets in vello_cpu.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -594,6 +594,33 @@ impl RenderContext {
...
+    /// The context's content (sized `self.width × self.height`) is rendered
+    /// directly to the destination pixmap starting at `(dst_x, dst_y)`.
+    /// If the region extends beyond the pixmap bounds, it is clipped.
+    ///
+    /// This method composites on top of existing pixmap content rather than
+    /// clearing it first, allowing multiple renders to accumulate.
+    ///
+    /// This is useful for rendering individual elements (like glyphs) into
+    /// a spritesheet at specific coordinates.
+    pub fn render_to_pixmap_at_offset(&self, pixmap: &mut Pixmap, dst_x: u16, dst_y: u16) {
```
LaurenzV: Two things:

1. We need to add a big fat warning that this will panic in multi-threaded mode. 😄 
2. I think it's not ideal that this method has the same name as `render_to_pixmap`, even though they do different things (one will replace the pixmap, anther one will composite it on top of existing content). Maybe we should rename it to `composite_onto_pixmap_at_offset` or something?

LaurenzV: This also applies to all other usages as well as some of the related methods (i.e. the internal `rasterize_at_offset` methods).

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -398,6 +398,93 @@ impl SingleThreadedDispatcher {
...
+        let mut fine = Fine::<S, F>::new(simd);
+
+        regions.update_regions(|region| {
+            let x = region.x;
+            let y = region.y;
+
+            let wtile = self.wide.get(x, y);
+            fine.set_coords(x, y);
+
+            fine.unpack(region);
```
LaurenzV, line 478: Maybe add a small comment that this achieves the effect of compositing onto the existing contents of a pixmap, instead of replacing it.

grebmeg: Done.

sparse_strips/vello_cpu/src/render.rs
grebmeg: I agree with both comments.

Fixed.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -594,6 +594,39 @@ impl RenderContext {
...
+    /// renders to accumulate.
+    ///
+    /// This is useful for rendering individual elements (like glyphs) into
+    /// a spritesheet at specific coordinates.
+    ///
+    /// # Panics
+    ///
+    /// This method is only supported with the single-threaded dispatcher and will
+    /// **panic** if called on a `RenderContext` using the multi-threaded dispatcher.
+    pub fn composite_to_pixmap_at_offset(&self, pixmap: &mut Pixmap, dst_x: u16, dst_y: u16) {
```
T-256, line 614: In future, Same approach for working with direct buffer would be like:
```rust
pub fn composite_to_buffer_at_offset(
    &self,
    buffer: &mut [u8],
    width: u16,
    height: u16,
    dst_x: u16,
    dst_y: u16,
) { ... }
```

Issue #1421, opened 2026-02-05
vello_cpu: Unreachable code reached


laurenz-canva:
While trying to find a minimal example for #1417, I stumbled upon a different panic for the following test:

```rs
#[vello_test(width = 512, height = 4)]
fn debug_test(ctx: &mut impl Renderer) {
    let filter_gaussian_blur = Filter::from_primitive(FilterPrimitive::GaussianBlur {
        std_deviation: 10.0,
        edge_mode: EdgeMode::None,
    });

    let path_2 = Rect::new(1.0, 1.0, 500.0, 3.0).to_path(0.1);
    let clip_path_1 = Rect::new(0.0, -5.0, 100.0, -3.0).to_path(0.1);

    ctx.push_layer(Some(&clip_path_1), None, None, None, Some(filter_gaussian_blur));
    ctx.fill_path(&path_2);
    ctx.pop_layer();
}
```

```
thread 'filter::debug_test_cpu_u8_scalar' (8668100) panicked at sparse_strips/vello_cpu/src/fine/mod.rs:614:17:
internal error: entered unreachable code
```

This one seems to trace back to: https://github.com/linebender/vello/commit/7816a37c

CC @grebmeg

PR #1429, opened 2026-02-09
vello_sparse_tests: Tweak some filter-related tests

laurenz-canva:
This PR:
- Adds some tests for the edge mode behavior in Gaussian blur.
- Renames the `filter_flood` test to `filter_flood_star`, to make space for a simpler test for the base case.
- Changes the star test to use a clipping path instead of a mask, so that we can later use that test for `vello_hybrid` as well. As far as I can tell, while the previous version uses masks, it just draws on it using a fully opaque color, so it should be functionally equivalent?
- Adds a test for #1421.

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/filter.rs
+++ b/sparse_strips/vello_sparse_tests/tests/filter.rs
@@ -3,7 +3,7 @@
 
 //! Tests demonstrating the filter effects API usage.
 
-use crate::util::{circular_star, stops_blue_green_red_yellow};
+use crate::util::circular_star;
 use crate::{renderer::Renderer, util::layout_glyphs_roboto};
 use vello_common::color::AlphaColor;
 use vello_common::color::palette::css::{
@@ -11,10 +11,9 @@ use vello_common::color::palette::css::{
 };
 use vello_common::filter_effects::{EdgeMode, Filter, FilterPrimitive};
 use vello_common::kurbo::{Affine, BezPath, Circle, Point, Rect, Shape, Stroke};
-use vello_common::peniko::{BlendMode, Compose, Gradient, Mix};
+use vello_common::peniko::{BlendMode, Compose, Mix};
 use vello_cpu::color::palette::css::{BLUE, GREEN, RED};
 use vello_cpu::kurbo::Dashes;
-use vello_cpu::peniko::LinearGradientPosition;
 use vello_dev_macros::vello_test;
 
 // TODO: We are purposefully using multiple of WideTile width/height here, because the implementation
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1091,3 +1086,63 @@ fn issue_filter_canvas_boundaries(ctx: &mut impl Renderer) {
...
+fn filter_gaussian_blur_edge_mode_wrap(ctx: &mut impl Renderer) {
+    blur_with_edge_mode(ctx, EdgeMode::Wrap);
+}
+
+#[vello_test(skip_hybrid, skip_multithreaded, width = 256, height = 100)]
+fn filter_gaussian_blur_edge_mode_mirror(ctx: &mut impl Renderer) {
+    blur_with_edge_mode(ctx, EdgeMode::Mirror);
+}
+
+// TODO: See https://github.com/linebender/vello/issues/1421, currently causes a crash.
```
grebmeg: Would it be better to put all such tests into a separate file, something named like `unresolved_issues.rs`?

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1091,3 +1086,63 @@ fn issue_filter_canvas_boundaries(ctx: &mut impl Renderer) {
...
+    ctx.pop_layer();
+}
+
+// TODO: Currently, these tests have a width/height that is a multiple of a wide tile,
+// because edge modes currently don't handle other widths/heights correctly. Once that is
+// fixed, we should change the tests back to 100x100 to exercise that path as well.
+
+#[vello_test(skip_hybrid, skip_multithreaded, width = 256, height = 100)]
+fn filter_gaussian_blur_edge_mode_none(ctx: &mut impl Renderer) {
+    // TODO: The bottom part looks wrong compared to the other edges.
```
grebmeg: I think you can move that to the issue-related tests as well, with a rename.

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1091,3 +1086,63 @@ fn issue_filter_canvas_boundaries(ctx: &mut impl Renderer) {
...
 }
+
+fn blur_with_edge_mode(ctx: &mut impl Renderer, edge_mode: EdgeMode) {
+    let filter = Filter::from_primitive(FilterPrimitive::GaussianBlur {
+        std_deviation: 6.0,
+        edge_mode,
+    });
+
+    let rect = Rect::new(0.0, 0.0, 256.0, 100.0);
+    let gradient = Gradient {
```
grebmeg: It is a bit hard to distinguish whether the filters are actually being applied to gradients. I think we should definitely keep in mind that gradients need to be filtered as well, but for now we could use a simpler layer, something like three rectangles with different colors in a full row?

grebmeg (approved):

sparse_strips/vello_sparse_tests/tests/filter.rs
LaurenzV: Makes sense.

PR #1432, opened 2026-02-11
vello shaders: Refactor alpha inversion to handle faint alpha values

tgfrerer:
Tweak alpha inversion calculation to avoid artifacts with faint alpha values.

Contrary to MSAA16, Analytic area anti-aliasing does sometime return very faint alpha values. 

This can create artifacts that look like ghosting when looking at just the RGB channels in isolation, because the RGB values get boosted out of proportion by a very low alpha value. 

Usually, this is invisible, as it gets masked when drawing the final image using alpha blending, but it may be useful to correct for this at source.

Before the change: (Notice the artifacts in the RGB-only channel view (right))
<table>
<tr>
<td>image</td>
<td>
image
</td>
</tr>
</table>

After the change: (Notice artifacts are corrected now in RGB-only channel view, while keeping same visual for RGBA)
<table>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

### DIFF
--- a/vello_shaders/shader/fine.wgsl
+++ b/vello_shaders/shader/fine.wgsl
@@ -1273,12 +1273,16 @@ fn main(
         if coords.x < config.target_width && coords.y < config.target_height {
             let fg = rgba[i];
             // let fg = base_color * (1.0 - foreground.a) + foreground;
-            // Max with a small epsilon to avoid NaNs
-            let a_inv = 1.0 / max(fg.a, 1e-6);
-            let rgba_sep = vec4(fg.rgb * a_inv, fg.a);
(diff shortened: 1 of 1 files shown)


tgfrerer:
@raphlinus thank you for looking into this :) I like the form you suggest better -- it more clearly shows the intent of the code -- and i've updated the PR accordingly.

I think it is probably neutral in terms of performance (i've seen `step()` perform oddly less well than a plain `if` in the past, even though `step` should give a driver more of an opportunity to minimize branching; but i doubt the difference would be measurable, and readability is ;)

I've also tested the newest version for good measure and it fixes the rgb artifacts just as well. 
image

PR #1434, opened 2026-02-11
Apply two adjustments for image handling

### LINKED ISSUE #1433: vello_hybrid: Images with `Extend::Pad` don't work correctly
# Problem

The following test case renders differently on vello_cpu and vello_hybrid:

```rs
#[vello_test(width = 4, height = 4)]
fn issue_demonstration(ctx: &mut impl Renderer) {
    let r = PremulRgba8::from_u8_array([255, 0, 0, 255]);
    let b = PremulRgba8::from_u8_array([0, 0, 0, 0]);

    // Three red rows, one transparent row.
    let image = vec![
        r, r, r, r,
        r, r, r, r,
        r, r, r, r,
        b, b, b, b
    ];

    let pixmap = Pixmap::from_parts(image, 4, 4);
    let source = ctx.get_image_source(Arc::new(pixmap));
    let image = Image {
        image: source,
        sampler: ImageSampler {
            x_extend: vello_common::peniko::Extend::Pad,
            y_extend: vello_common::peniko::Extend::Pad,
            quality: ImageQuality::Low,
            alpha: 1.0,
        },
    };

    ctx.set_paint(image);
    ctx.fill_rect(&Rect::new(0.0, 0.0, 4.0, 4.0));
}
```

On vello_cpu, we get 3 red rows and 1 transparent row, as expected. On vello_hybrid, we get an image with 4 red rows.

The reason is as follows, In the shader, the pixel is correctly assigned the location y = 3.5:

Image

The rows above get assigned 0.5, 1.5, 2.5, as expected. 

However, upon applying the extend, we apply a small delta to the image_size:
```
            let offset = 0.00001;
            let extended_xy = vec2<f32>(
                extend_mode(local_xy.x, encoded_image.extend_modes.x, image_size.x - offset),
                extend_mode(local_xy.y, encoded_image.extend_modes.y, image_size.y - offset)
            );
```

This means that `image_size.y` is now 3.99999 instead of 4.0. In the extend calculation, we then _again_ subtract 1 from that value, which means that we end up with 2.9999 as the maximum value:

```
fn extend_mode(t: f32, mode: u32, max: f32) -> f32 {
    switch mode {
        case EXTEND_PAD: {
            return clamp(t, 0.0, max - 1.0);
        }
        ...
    }
}
```

Therefore, our index will be clamped to the value 2.9999, and thus to the index 2, even though it should have the index 3 (the last row of the image):

Image

# Solution
The fix here is to remove the offset of 0.00001 that we are applying. However, as far as I can tell, this offset was introduced to circumvent floating point inaccuracies in the `skew_image` test case, because this is the only ones that start failing when I remove the offset:

Image

I tried it via Skia and they don't seem to produce aliasing artifacts here, so not sure what the best approach is here.

laurenz-canva:
This PR contains two changes:

1.  It resolves #1433 by changing how the small offsets are applied. See the issue for more information.
1. It adjusts the bias that is added in vello_cpu by simply always subtracting 1 instead of a small bias. Since our image dimensions don't exceed u16::MAX, this is enough. This also aligns it to the current behavior of vello_hybrid:

https://github.com/linebender/vello/blob/fe813d305396630433e0d830f56834ad7bb18def/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl#L827-L831

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/issues.rs
+++ b/sparse_strips/vello_sparse_tests/tests/issues.rs
@@ -14,7 +14,7 @@ use vello_common::peniko::{
 };
 use vello_common::pixmap::Pixmap;
 use vello_cpu::color::palette::css::{BLACK, RED};
-use vello_cpu::peniko::Compose;
+use vello_cpu::peniko::{Compose, Extend};
 use vello_cpu::{Level, RenderContext, RenderMode, RenderSettings};
 use vello_dev_macros::vello_test;
 
@@ -469,15 +469,21 @@ fn issue_1433(ctx: &mut impl Renderer) {
     let b = PremulRgba8::from_u8_array([0, 0, 0, 0]);
 
     // Three red rows, one transparent row.
-    let image = vec![r, r, r, r, r, r, r, r, r, r, r, r, b, b, b, b];
+    #[rustfmt::skip]
+    let image = vec![
+        r, r, r, r,
+        r, r, r, r,
+        r, r, r, r,
+        b, b, b, b
+    ];
 
     let pixmap = Pixmap::from_parts(image, 4, 4);
     let source = ctx.get_image_source(Arc::new(pixmap));
     let image = Image {
         image: source,
         sampler: ImageSampler {
-            x_extend: vello_common::peniko::Extend::Pad,
-            y_extend: vello_common::peniko::Extend::Pad,
+            x_extend: Extend::Pad,
+            y_extend: Extend::Pad,
             quality: ImageQuality::Low,
             alpha: 1.0,
         },


sparse_strips/vello_sparse_tests/snapshots/issue_1433.png
laurenz-canva, line 1: I know it's a bit unfortunate that the image is so small, but it's the best way to demonstrate the issue. 😅

sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -457,3 +462,27 @@ fn basic_alpha_compositing(ctx: &mut impl Renderer) {
...
+    // Three red rows, one transparent row.
+    let image = vec![r, r, r, r, r, r, r, r, r, r, r, r, b, b, b, b];
+
+    let pixmap = Pixmap::from_parts(image, 4, 4);
+    let source = ctx.get_image_source(Arc::new(pixmap));
+    let image = Image {
+        image: source,
+        sampler: ImageSampler {
+            x_extend: vello_common::peniko::Extend::Pad,
+            y_extend: vello_common::peniko::Extend::Pad,
```
grebmeg: [nit]:
```suggestion
            x_extend: Extend::Pad,
            y_extend: Extend::Pad,
```

sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -457,3 +462,27 @@ fn basic_alpha_compositing(ctx: &mut impl Renderer) {
...
     ctx.fill_rect(&Rect::new(0.0, 0.0, u16::MAX as f64 + 10.0, 8.0));
 }
+
+#[vello_test(width = 4, height = 4)]
+fn issue_1433(ctx: &mut impl Renderer) {
+    let r = PremulRgba8::from_u8_array([255, 0, 0, 255]);
+    let b = PremulRgba8::from_u8_array([0, 0, 0, 0]);
+
+    // Three red rows, one transparent row.
+    let image = vec![r, r, r, r, r, r, r, r, r, r, r, r, b, b, b, b];
```
grebmeg: [nit]
```suggestion
    #[rustfmt::skip]
    let image = vec![
        r, r, r, r,
        r, r, r, r,
        r, r, r, r,
        b, b, b, b,
    ];
```

grebmeg (approved):
Nice one! Thanks for correction!

PR #1436, opened 2026-02-13
vello_hybrid: Optimize layer blending for src-over compositing

laurenz-canva:
As was explained in #1155, we need a special mechanism that makes used of temporary slots in vello_hybrid to enable correct blending. However, in its current state, this will also be used by layers which don't have any special blending (for example clip or opacity layers), even though this shouldn't be needed: Because for normal src-over compositing, the GPU will implicitly do the "reading of the destination" for us (by defining `BlendState::PREMULTIPLIED_ALPHA_BLENDING` in the strip pipeline). This means that we only need this specialized mechanism if the layer isn't using default src-over compositing.

Therefore, this PR adds an optimization to only use temporary slots for non-default blend modes. The code path that was previously only used for copying data back to the surface has been repurposed as a general "do src-over compositing" operation.

Since we have some pretty crazy tests with deep blend layers, I hope this gives us enough confidence that this should be safe to do.

I imagine this might also give us a speed boost for opacity/clip layers, I haven't done any benchmarking, though. But it definitely shouldn't cause any performance regressions.

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -958,24 +958,25 @@ impl Scheduler {
                     let next_round: bool = depth.is_multiple_of(2) && depth > 2;
                     let round = nos.round.max(tos.round + usize::from(next_round));
 
+                    let draw = self.draw_mut(
+                        round,
+                        if depth <= 2 {
+                            2
+                        } else {
+                            nos.dest_slot.get_texture()
+                        },
+                    );
+
+                    let gpu_strip_builder = if depth <= 2 {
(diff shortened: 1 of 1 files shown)


laurenz-canva:
I did some testing, and it helps _tremendously_ for performance!  Here's the before/after with a simple scene that draws one rectangle in each layer (sorry for the bad quality, had to make it smaller so it's less than 10MB):

https://github.com/user-attachments/assets/31e2bfdb-798a-451a-9ee0-f0c3e3911ed7

Before, I was getting around 25FPS for 28 rectangles, with that commit that stays at 150FPS+!

PR #1437, opened 2026-02-13
vello_cpu: Tweak behavior for filter layers with zero clips

### LINKED ISSUE #1421: vello_cpu: Unreachable code reached

laurenz-canva:
Closes #1421. The problematic command trace looks as follows:

image

Using the current logic, when we hit a filter layer followed by `PushZeroClip`, we simply increase the command index by one. The render range is 4..4, so afterwards we would simply skip to the `Filter` command that follows `PushZeroClip`. However, as you can see there still is a `PopZeroClip` command.

I think the correct behavior is to completely skip all commands of the layer and jump to the very last `PopBuf` command, which is then executed to ensure the previously pushed (empty) buffer is freed again.

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
@@ -319,10 +319,9 @@ impl SingleThreadedDispatcher {
                     Some(Cmd::PushZeroClip(id)) if *id == *child_layer_id => {
                         // If we have a zero-clip, it means that the whole layer should not be drawn.
                         // Therefore, we want to skip to the very end so that only `PopBuf` will
-                        // be run. Therefore, we would have to do `filtered_ranges.full_range.end - 1`.
-                        // However, after the match statement we use `cmd_idx + 1` to determine the
-                        // command index, so we actually need to subtract 2 here.
-                        cmd_idx = filtered_ranges.full_range.end - 2;
+                        // be run. Therefore, we jump to `filtered_ranges.full_range.end - 1`.
+                        cmd_idx = filtered_ranges.full_range.end - 1;
+                        continue;
                     }
 
                     // Partial clip: push the clip buffer, then composite the filtered layer


sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -317,7 +317,12 @@ impl SingleThreadedDispatcher {
                     // since this tile is entirely clipped out.
                     // (PushZeroClip only appears for clipped filter layers)
                     Some(Cmd::PushZeroClip(id)) if *id == *child_layer_id => {
-                        cmd_idx += 1; // Skip the PushZeroClip command
+                        // If we have a zero-clip, it means that the whole layer should not be drawn.
+                        // Therefore, we want to skip to the very end so that only `PopBuf` will
+                        // be run. Therefore, we would have to do `filtered_ranges.full_range.end - 1`.
+                        // However, after the match statement we use `cmd_idx + 1` to determine the
+                        // command index, so we actually need to subtract 2 here.
+                        cmd_idx = filtered_ranges.full_range.end - 2;
```
grebmeg: What about the approach below? With `continue`, we skip line 357 entirely. So `full_range.end - 1` is just the `PopBuf` index — direct, no dependency on downstream code.

```suggestion
                        // Jump to PopBuf to pop the blend buffer pushed above,
                        // skipping all commands in between.
                        cmd_idx = filtered_ranges.full_range.end - 1;
                        continue;
```

LaurenzV: Works for me, did it this way because I thought it might be better if branches use the same path.

PR #1438, opened 2026-02-13
vello_hybrid: Remove command annotation logic

laurenz-canva:
~This PR is stacked on top of #1436, so that should land first.~

It removes the command annotation logic that is needed to handle blending correctly, and instead moves the bookkeeping into vello_common during coarse rasterization. Basically, we add a new `Start` command that tells us whether the original surface is a target of blending operations, and each `PushBuf` command has a similar flag. In vello_hybrid, we simply need to query this flag and add some additional processing if the root layer is the target of blending. I've considered putting the logic behind the const generic of `Wide`, but I think it's probably not worth as it makes the code less straightforward, and I think the performance overhead is negligible.

While not as big of a performance improvement as I had hoped, it's still something:
Before:
image

After:
image

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -1535,12 +1535,12 @@ impl<const MODE: u8> WideTile<MODE> {
 
     /// Blend the current buffer into the previous buffer in the stack.
     fn blend(&mut self, blend_mode: BlendMode) {
-        // Whether we use a non-default blend mode to blend into the destination.
-        let blends_into_dest =
-            blend_mode.mix != Mix::Normal || blend_mode.compose != Compose::SrcOver;
-
         #[allow(clippy::collapsible_if, reason = "better expresses intent")]
         if MODE == MODE_HYBRID {
+            // Whether we use a non-default blend mode to blend into the destination.
+            let blends_into_dest =
+                blend_mode.mix != Mix::Normal || blend_mode.compose != Compose::SrcOver;
+
             if blends_into_dest && self.push_buf_indices.len() >= 2 {
                 let nos_idx = self.push_buf_indices[self.push_buf_indices.len() - 2];
 


sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva, line 1: The diffs should become more readable once #1436 gets merged. Unfortunately, I can't set it as the target branch for this PR since it lives in my own fork.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -666,14 +625,42 @@ impl Scheduler {
         paint_idxs: &[u32],
         attrs: &CommandAttrs,
     ) -> Result<(), RenderError> {
-        for annotated_cmd in &state.annotated_commands {
+        // What is going on with the `surface_needs_wrap` and `is_blend_target` variables in
```
ajakubowicz-canva: Is this documentation potentially stale? I don't see a `surface_needs_wrap` variable in the code? I assume this is referring to `surface_is_blend_target`.
I think these docs can be simplified to only document `surface_is_blend_target`.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -952,68 +836,203 @@ impl Scheduler {
...
+            // Simple source-over compositing into the final render target.
+            self.do_blend(state, wide_tile_x, wide_tile_y, &BlendMode::default());
+            self.do_pop_buf(state);
+        }
+        Ok(())
+    }
 
-                    let next_round: bool = depth.is_multiple_of(2) && depth > 2;
-                    let round = nos.round.max(tos.round + usize::from(next_round));
+    fn do_push_buf<R: RendererBackend>(
```
ajakubowicz-canva, line 852: For these abstracted code blocks, `do_<cmd>`, should they be `#[inline(always)]`?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1166,71 +1185,3 @@ impl GpuStripBuilder {
...
     rgba >= 0x1_00_00_00
 }
-
-/// Does a single linear scan over the wide tile commands to prepare them for `do_tile`. Notably
-/// this function:
-///  - Precomputes the layers that require temporary slots due to blending.
-///
-/// TODO: Can be triggered via a const generic on coarse draw cmd generation which will avoid
-/// a linear scan.
-fn prepare_cmds(cmds: &[Cmd], state: &mut SchedulerState) {
```
taj-p, line 1176: I frequently see this function show up in browser performance profiles, so I wouldn't be surprised if we see non-trivial performance improvements in Wasm 🎉

taj-p: E.g., this frame took 300 microseconds in graphics. 100 of which are in this function!

Image

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Ah, yes the comments should be referring to `surface_is_blend_target`, thanks!

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Sounds reasonable!

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -952,68 +836,206 @@ impl Scheduler {
                     state.tile_state.stack.last_mut().unwrap().opacity = *opacity;
                 }
                 Cmd::Blend(mode) => {
-                    let tos = state.tile_state.stack.last().unwrap();
-                    let nos = &state.tile_state.stack[state.tile_state.stack.len() - 2];
+                    self.do_blend(state, wide_tile_x, wide_tile_y, mode);
+                }
+                _ => unimplemented!(),
```
RobertBrewitz: Might this cause panics for consumers of vello_hybrid at this time?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -952,68 +836,206 @@ impl Scheduler {
...
+            // Simple source-over compositing into the final render target.
+            self.do_blend(state, wide_tile_x, wide_tile_y, &BlendMode::default());
+            self.do_pop_buf(state);
+        }
+        Ok(())
+    }
 
-                    let next_round: bool = depth.is_multiple_of(2) && depth > 2;
-                    let round = nos.round.max(tos.round + usize::from(next_round));
+    #[inline(always)]
```
RobertBrewitz: does it make sense to do inline always compared to just inline?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1922,7 +1957,7 @@ mod tests {
         wide.fill(10, 10, 0, 0, FillHint::None);
         wide.pop_buf();
 
-        assert_eq!(wide.cmds.len(), 4);
+        assert_eq!(wide.cmds.len(), 5);
```
RobertBrewitz: should we test that the Start cmd is actually the first entry?

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: I think it shouldn't. Will change it to `unreachable`.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: I think it makes sense if we really want to be sure, but both would probably work in this case. Just did it because of Andrew's suggestion.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Sounds reasonable 👍

sparse_strips/vello_hybrid/src/schedule.rs
RobertBrewitz: ok, I was just thinking if there are cases were a compiler might not want to inline it because it might be faster not to?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1589,14 +1609,24 @@ impl LayerKind {
 /// (`Filter`, `Blend`, `Opacity`, `Mask`).
 #[derive(Debug, PartialEq)]
 pub enum Cmd {
+    /// Marks the start of a tile's command list.
+    ///
+    /// The first argument indicates whether the current layer (i.e. the final target surface)
+    /// is used as a destination for a blending operation with a non-default blend mode. This
+    /// information is only needed by `vello_hybrid` for scheduling purposes.
+    Start(bool),
```
grebmeg: I wonder if `Cmd::Start(bool)` may not be the best fit conceptually: the `bool` seems to be purely tile-level metadata (“surface is a blend target”) that’s set once during coarse and then only read once by the hybrid scheduler. Since it doesn’t really participate in the command stream, would it make sense to move this to `WideTile` as something like `surface_is_blend_target: bool`? Coarse could set it in `blend()`, and the scheduler could read the flag before iterating commands. That would let you drop the `Cmd::Start` variant.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1500,6 +1508,18 @@ impl<const MODE: u8> WideTile<MODE> {
...
+            blend_mode.mix != Mix::Normal || blend_mode.compose != Compose::SrcOver;
+
+        if blends_into_dest && self.push_buf_indices.len() >= 2 {
+            let nos_idx = self.push_buf_indices[self.push_buf_indices.len() - 2];
+            match &mut self.cmds[nos_idx] {
+                Cmd::PushBuf(_, is_blend_target) => *is_blend_target = true,
+                Cmd::Start(is_blend_target) => *is_blend_target = true,
+                _ => unreachable!(),
+            }
+        }
```
grebmeg: Since `WideTile` already has `const MODE: u8`, could we gate the hybrid-only code behind `MODE == MODE_HYBRID`, so the CPU path doesn’t carry extra fields/logic?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -666,14 +625,42 @@ impl Scheduler {
...
+        // Note however that all of this is not necessary when we have layers (for example for
+        // clip paths or opacity layers) with normal source-over blending, since the GPU automatically
+        // takes care of doing blending in this case. Determining which layers are the target of
+        // a blend operation has already happened during coarse rasterization, so here we just
+        // need to pass on the boolean flags.
+        //
+        // `surface_is_blend_target` is needed because the main target surface can obviously also
+        // end up being the destination of a blending operation. Since that surface is provided
+        // by the user, there is no way we can read from it, which is required for blending. Therefore,
+        // in this case we need to "wrap" _everything_ into a push/pop layer operation.
```
grebmeg, line 650: If I’m understanding this refactor correctly, the main win is moving the same “blend-target” annotation into coarse rasterization so the scheduler no longer needs an extra O(N) pass. I’m curious if we can simplify further, though:

- Rather than having the child reach backward during `blend()` to annotate the parent, could we annotate the parent when the child is pushed? At `push_layer` time we already know the child’s blend mode. Something along these lines (just a rough idea):

```rust
pub fn push_layer(&mut self, ..., blend_mode: BlendMode, ...) {
    // ...
    let blends_into_parent =
        blend_mode.mix != Mix::Normal || blend_mode.compose != Compose::SrcOver;

    if layer.needs_buf() {
        for x in 0..self.width_tiles() {
            for y in 0..self.height_tiles() {
                let tile = self.get_mut(x, y);
                if blends_into_parent {
                    tile.mark_top_as_blend_target();
                }
                tile.push_buf(layer_kind);
            }
        }
    }
    // ...
}
```

- Alternatively, could we avoid indexed mutation altogether by annotating the command we’re emitting (not a previously-emitted one)? E.g. instead of tagging the parent’s `PushBuf`, tag the child’s `PushBuf` with “I will blend non-trivially into my parent”.

```rust
PushBuf(LayerKind, BlendIntent)

enum BlendIntent {
    SrcOver,
    CustomBlend,
}

fn push_buf(&mut self, layer_kind: LayerKind, blend_intent: BlendIntent) {
    self.cmds.push(Cmd::PushBuf(layer_kind, blend_intent));
    // ...
}

Cmd::PushBuf(_, blend_intent) => {
    // If this child will blend non-trivially, ensure the parent has a temp slot
    if blend_intent == BlendIntent::CustomBlend {
        let nos = state.tile_state.stack.last_mut().unwrap();
        if nos.temporary_slot == TemporarySlot::None {
            let temp = self.claim_free_slot(...)?;
            nos.temporary_slot = TemporarySlot::Valid(temp);
        }
    }
    // Then push the child's slot
    self.do_push_buf(state, renderer)?;
}
```

I don’t think this needs to be in this PR—this is already a solid refactor, but it might be worth considering a few follow-up simplifications in the future.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Good idea, will change that.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: The first one should definitely work, the second one I need to think about. But yes, let's do this in a follow-up and keep this PR so it mirrors the existing logic as well as possible. 😄

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Changed it to just inline.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1500,6 +1535,27 @@ impl<const MODE: u8> WideTile<MODE> {
 
     /// Blend the current buffer into the previous buffer in the stack.
     fn blend(&mut self, blend_mode: BlendMode) {
+        // Whether we use a non-default blend mode to blend into the destination.
+        let blends_into_dest =
+            blend_mode.mix != Mix::Normal || blend_mode.compose != Compose::SrcOver;
```
grebmeg: Should this be also included under `if MODE == MODE_HYBRID` below?

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Thinking about it, I'm not sure if the second approach would work. It would only work if the child's `PushBuf` is the first item in the parent's command list. If not, we would start by drawing elements in the normal slot of the parent layer, before hitting the `PushBuf` command and creating the temporary slot. But at that point, the previous draw commands are already lost. Or am I missing something?

LaurenzV: For the first point, I'm realizing now that even if we do it in `PushBuf`, we still need to keep track of where the `PushBuf` command for the parent layer was in a stack, no? So I think in the end there wouldn't really be a benefit in moving it? Or am I missing something?

PR #1440, opened 2026-02-13
Enable `std` feature on fearless_simd for runtime feature detection

Shnatsel:
Without the `std` feature, [runtime feature detection is documented as being unavailable](https://github.com/linebender/fearless_simd/blob/05f68a43dd3a815d4e7247f6cfacfec739d3fb50/fearless_simd/Cargo.toml#L23-L25).

Runtime feature detection still works in benchmarks, not sure why. Possibly due to workspace feature unification, but that may very well break for anyone depending on vello via crates.io


Shnatsel:
needs more work to gel with no_std mode, shelving for now

PR #1443, opened 2026-02-16
Fix handling of `pop_clip` commands and properly reset wide tiles

laurenz-canva:
This requires #1437 to be merged first, otherwise the test doesn't pass properly. But I have verified that this fixes the issue. There were basically two problems:

1) We were erroneously exiting eagerly in `pop_clip` in case the clip path has zero strips, causing `push_clip` and `pop_clip` to not be symmetrical. This meant that if the first scene was rendered improperly, this inconsistent state could leak into future iterations.
2) We were not properly resetting `n_zero_clips` in Wide.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -324,14 +324,14 @@ impl<const MODE: u8> Wide<MODE> {
         for tile in &mut self.tiles {
             tile.bg = PremulColor::from_alpha_color(TRANSPARENT);
             tile.cmds.clear();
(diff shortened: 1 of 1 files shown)


tomcur:
> We were not properly resetting n_zero_clips in Wide.

Indeed, and I see you also corrected resetting by adding the other fields as well, as in https://github.com/linebender/vello/pull/1414/changes#diff-5e16077335b3a3c3ed1845456ec5716c0368870c15a5cea692c03ad428d207a5R410. As a very nitty nit, perhaps move these to the same position as in that PR, as that's closer to the field order of `WideTile`.

PR #1446, opened 2026-02-16
sparse_strips: Move filter handling logic from vello_cpu to vello_common

laurenz-canva:
Since this is a relatively isolated change, I decided to open a PR for this now so that it reduces the diff of follow-up PRs. All this PR does is simply:

- Moving all of the filter definitions that can be reused across both renderers into vello_common, while leaving vello_cpu specific logic where it currently is.
- Introduce an `InstantiatedFilter` abstraction that allows instantiating a filter for a specific transform, since we need the same in vello_hybrid.

There's a bit of a question whether we can deduplicate some of the predefined SVG filters and their instantiated versions, but for now this is the least invasive change, I think.

### DIFF
--- a/sparse_strips/vello_common/src/filter/drop_shadow.rs
+++ b/sparse_strips/vello_common/src/filter/drop_shadow.rs
@@ -7,7 +7,7 @@ use crate::color::{AlphaColor, Srgb};
 use crate::filter::gaussian_blur::{MAX_KERNEL_SIZE, plan_decimated_blur, transform_blur_params};
 use crate::filter::transform_offset_params;
 use crate::filter_effects::EdgeMode;
-use crate::kurbo::{Affine, Vec2};
+use crate::kurbo::Affine;
 
 /// A drop shadow filter.
 #[derive(Debug)]


sparse_strips/vello_common/src/filter/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::kurbo::{Affine, Vec2};
+
+pub mod drop_shadow;
+pub mod flood;
+pub mod gaussian_blur;
+pub mod offset;
+
+/// A filter that has been instantiated for a specific affine transformation.
+#[derive(Debug)]
+pub enum InstantiatedFilter {
```
laurenz-canva: Perhaps `PreparedFilter` would be a better name?

sparse_strips/vello_common/src/filter/drop_shadow.rs
```diff
@@ -0,0 +1,92 @@
...
+    transform: &Affine,
+) -> (f32, f32, f32) {
+    // Transform the offset vector by the full transformation matrix
+    // to correctly handle rotation, scale, and shear.
+    // We use the linear part only (no translation) since this is a vector offset.
+    let offset = Vec2::new(dx as f64, dy as f64);
+    let [a, b, c, d, _, _] = transform.as_coeffs();
+    let transformed_offset = Vec2::new(a * offset.x + c * offset.y, b * offset.x + d * offset.y);
+    let scaled_dx = transformed_offset.x as f32;
+    let scaled_dy = transformed_offset.y as f32;
```
grebmeg: I think you can reuse `transform_offset_params` from `vello_common/src/filter/mod.rs` here.

sparse_strips/vello_common/src/filter_effects.rs
```diff
@@ -54,6 +54,8 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use smallvec::SmallVec;
 
+// TODO: Move/integrate this into the `filter` module?
```
grebmeg: I actually think I like the current split: `filter_effects` defines a pure data model (and could potentially move to `peniko`(?)), while `filter/` contains the rendering-side representation/implementation.

sparse_strips/vello_common/src/filter/mod.rs
grebmeg: Yeah, agree, maybe let's name it `PreparedFilter`?

grebmeg (approved):
LGTM

sparse_strips/vello_common/src/filter/drop_shadow.rs
laurenz-canva: True!

PR #1448, opened 2026-02-16
vello_hybrid: Unify texture creation code in WebGL backend

laurenz-canva:
Perhaps I am missing something obvious here, but I'm not sure I get why we were using different clamp/filtering modes for some textures? 🤔 If I'm not mistaken, we are only ever using direct texture loading, so it seems to me like it should be fine to always use the same mode? At least all WebGL tests still seem to pass, but please let me know if I'm wrong.

Other than that, this PR just puts all of that logic into a method to reduce code duplication. 🙂

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1394,7 +1394,7 @@ fn create_texture_inner(gl: &WebGl2RenderingContext, target: u32) -> WebGlTextur
         WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
     );
     // Also only to be defensive, in theory this shouldn't be necessary since we use
-    // `NEAREST` for the filters/
+    // `NEAREST` for both filters.
     gl.tex_parameteri(
         WebGl2RenderingContext::TEXTURE_2D,
         WebGl2RenderingContext::TEXTURE_MAX_LEVEL,


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1582,29 +1497,7 @@ fn create_atlas_texture_array(
...
-    );
-    gl.tex_parameteri(
-        WebGl2RenderingContext::TEXTURE_2D,
-        WebGl2RenderingContext::TEXTURE_WRAP_T,
-        WebGl2RenderingContext::REPEAT as i32,
-    );
+    let texture = create_texture(gl);
     gl.tex_parameteri(
         WebGl2RenderingContext::TEXTURE_2D,
         WebGl2RenderingContext::TEXTURE_MAX_LEVEL,
```
laurenz-canva: Would this still be needed?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1355,6 +1355,46 @@ fn get_clear_uniforms(gl: &WebGl2RenderingContext, program: &WebGlProgram) -> Cl
...
+fn create_texture_array(gl: &WebGl2RenderingContext) -> WebGlTexture {
+    create_texture_inner(gl, WebGl2RenderingContext::TEXTURE_2D_ARRAY)
+}
+
+fn create_texture_inner(gl: &WebGl2RenderingContext, target: u32) -> WebGlTexture {
+    let texture = gl.create_texture().unwrap();
+    gl.active_texture(WebGl2RenderingContext::TEXTURE0);
+    gl.bind_texture(target, Some(&texture));
+    // The actual sampling/wrapping behavior doesn't matter since we always
+    // sample in-bounds and apply the logic for that behavior ourselves.
```
grebmeg: Maybe add a bit more detail here, just to clarify?

```suggestion
    // The filter and wrap modes are irrelevant because the shader
    // (`render_strips.wgsl`) exclusively uses `textureLoad`, which bypasses
    // the sampler entirely.
```

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: I think you’re right this is no longer needed. The old code used `NEAREST_MIPMAP_LINEAR` as the min filter, which requires mipmaps for texture completeness. Since only mip level 0 was allocated, `MAX_LEVEL = 0` was essential to prevent the texture from being incomplete (which would make all reads return black). Now that `create_texture` sets `NEAREST`, which doesn’t require mipmaps, the texture is complete with just level 0. Removing it would be consistent with the other textures.

Another option is to keep setting `MAX_LEVEL = 0` unconditionally in `create_texture` anyway as it’s harmless and defensive, and it avoids the slot texture silently breaking if someone later changes the min filter (or switches `create_texture` to something like `LINEAR_MIPMAP_LINEAR`).

PR #1451, opened 2026-02-18
vello_cpu: Add `ImageResolver` for resolving opaque IDs at rasterization time

grebmeg:
This PR introduces `ImageResolver` trait and an image registry on `RenderContext`, enabling deferred resolution of `ImageSource::OpaqueId` images during `vello_cpu` rasterization. This decouples render command encoding from image data ownership, supporting patterns like dynamic sprite atlases and glyph caches where the backing pixmap may be updated between encoding and rendering.

image

### DIFF
--- a/sparse_strips/vello_common/src/paint.rs
+++ b/sparse_strips/vello_common/src/paint.rs
@@ -148,8 +148,8 @@ pub type Image = peniko::ImageBrush<ImageSource>;
 pub trait ImageResolver: Send + Sync {
     /// Resolve an `ImageId` to its pixmap data.
     ///
-    /// In `vello_cpu`, this is called during fine rasterization for each wide tile
-    /// command that references an `ImageSource::OpaqueId`.
+    /// This method may be called repeatedly (dozens or even hundreds of times
+    /// per frame) and should therefore be very fast.
     ///
     /// Returns `None` if the image ID is not found in the registry.
     fn resolve(&self, id: ImageId) -> Option<Arc<Pixmap>>;
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -1055,7 +1055,7 @@ impl ImageResolver for ImageRegistry {
     }
 }
 
-/// Saved state for recording operations.¬
+/// Saved state for recording operations.
 #[derive(Debug)]
 pub struct RenderState {
     transform: Affine,


waywardmonkeys:
What's the story for the `vello_hybrid` version of this?

This seems like it is sitting on the edge of the discussions about resource management.

(That said, this looks like a decent direction for `vello_cpu` ... but I think we have to be thinking about the family of renderers as well.)

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -379,7 +381,12 @@ impl MultiThreadedDispatcher {
         width: u16,
         height: u16,
         encoded_paints: &[EncodedPaint],
+        _image_resolver: &dyn ImageResolver,
     ) {
+        // Note: Multi-threaded dispatcher does not support ImageSource::OpaqueId.
+        // Images with OpaqueId will panic at rasterization time.
+        let noop_resolver = NoOpImageResolver;
```
LaurenzV: If we use the `NoOpImageResolver` anyway, why does it panic? Or does it already panic _in its current state_ (i.e. before this PR)?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -75,6 +76,13 @@ pub struct RenderContext {
     dispatcher: Box<dyn Dispatcher>,
     #[cfg(feature = "text")]
     pub(crate) glyph_caches: Option<GlyphCaches>,
+    /// Image registry for resolving `ImageSource::OpaqueId` to pixmap data.
+    ///
+    /// This allows decoupling render commands from image data, enabling
+    /// patterns like spritesheet rendering.
+    image_registry: HashMap<u32, Arc<Pixmap>>,
+    /// Counter for generating unique image IDs.
+    next_image_id: u32,
```
LaurenzV: nit, but it might be worth encapsulating the two into a custom struct?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -424,6 +434,10 @@ impl RenderContext {
     }
 
     /// Set the current paint.
+    ///
+    /// If the paint is an image with `ImageSource::OpaqueId`, it will be
+    /// resolved to the corresponding pixmap at rasterization time.
+    /// Make sure to register images with [`register_image`](Self::register_image) first.
```
LaurenzV, line 434: This needs a warning that this doesn't work with multi-threading enabled.

sparse_strips/vello_common/src/paint.rs
```diff
@@ -139,6 +139,29 @@ impl ImageSource {
...
+///
+/// This allows delaying the resolution of `ImageSource::OpaqueId` until the
+/// image is actually needed during rasterization, enabling patterns like
+/// dynamic sprite atlases where the image data may be updated between
+/// encoding and rendering.
+pub trait ImageResolver {
+    /// Resolve an `ImageId` to its pixmap data.
+    ///
+    /// Returns `None` if the image ID is not found in the registry.
+    fn resolve(&self, id: ImageId) -> Option<Arc<Pixmap>>;
```
LaurenzV, line 155: Is my intuition correct that, at least in `vello_cpu`, this is called _once_ at rasterization time and then cached, instead of being called individually in each wide tile? If so, might be worth documenting this (i.e. that this will be called once at rasterization time).

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -655,6 +677,68 @@ impl RenderContext {
...
+    /// Register a pixmap in the image registry and return its [`ImageId`].
+    pub fn register_image(&mut self, pixmap: Arc<Pixmap>) -> ImageId {
+        let id = self.next_image_id;
+        self.next_image_id += 1;
+        self.image_registry.insert(id, pixmap);
+        ImageId::new(id)
+    }
+
+    /// Update an existing image in the registry with new pixmap data.
+    pub fn update_image(&mut self, id: ImageId, pixmap: Arc<Pixmap>) {
```
LaurenzV: What's the use case for this? Wouldn't it be better to just create a new image?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -954,7 +1038,7 @@ impl Recordable for RenderContext {
 
 /// Saved state for recording operations.
 #[derive(Debug)]
-struct RenderState {
+pub struct RenderState {
```
LaurenzV, line 1060: Why does this need to be public now?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -655,6 +677,68 @@ impl RenderContext {
...
+    }
+
+    /// Restore the saved rendering state.
+    pub fn restore_state(&mut self, state: RenderState) {
+        self.transform = state.transform;
+        self.fill_rule = state.fill_rule;
+        self.stroke = state.stroke;
+        self.paint = state.paint;
+        self.paint_transform = state.paint_transform;
+    }
```
LaurenzV, line 693: Same here, do those need to be public? If I'm not mistaken they are only used inside `vello_cpu`.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
grebmeg: I think I accidentally left that comment here, previously I had panic here from an `unimplemented!`. Moreover, I just verified this in multithreaded mode with the `image_spritesheet` test, and it seems to be working.

Corrected.

sparse_strips/vello_cpu/src/render.rs
grebmeg: They’re currently used in `vello_cpu`, but the main goal of almost all my PRs is to enable glyph caching in `parley_draw`, where I use both methods.

sparse_strips/vello_cpu/src/render.rs
grebmeg: ditto — I'm using it in `parley_draw` for glyph caching.

sparse_strips/vello_cpu/src/render.rs
grebmeg: Yes, I think we can do that.

Fixed.

sparse_strips/vello_cpu/src/render.rs
grebmeg: Actually, it seems to be working (I enabled this in the `image_spritesheet` test).

sparse_strips/vello_cpu/src/render.rs
grebmeg: Yeah, I think that makse sense.

Fixed.

sparse_strips/vello_common/src/paint.rs
grebmeg: Actually, the `ImageResolver` is passed all the way down to `fine.run_cmd()`, which is called per command per wide tile. When a `Cmd` references an `ImageSource::OpaqueId`, the resolver's `resolve()` is called at that point to get the pixmap. I left a short explanation here.

sparse_strips/vello_common/src/paint.rs
```diff
@@ -139,6 +139,32 @@ impl ImageSource {
...
+/// Trait for resolving opaque image IDs to pixmaps at rasterization time.
+///
+/// This allows delaying the resolution of `ImageSource::OpaqueId` until the
+/// image is actually needed during rasterization, enabling patterns like
+/// dynamic sprite atlases where the image data may be updated between
+/// encoding and rendering.
+pub trait ImageResolver: Send + Sync {
+    /// Resolve an `ImageId` to its pixmap data.
+    ///
+    /// In `vello_cpu`, this is called during fine rasterization for each wide tile
```
LaurenzV: The average user probably doesn't know what wide tile commands are. Maybe the comment should just be changed so that it says that this method might be called repeatedly dozens or even hundreds of time, and therefor should be very fast?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -952,9 +1011,53 @@ impl Recordable for RenderContext {
...
+}
+
+impl ImageResolver for ImageRegistry {
+    fn resolve(&self, id: ImageId) -> Option<Arc<Pixmap>> {
+        self.images.get(&id.as_u32()).cloned()
+    }
+}
+
+/// Saved state for recording operations.¬
```
LaurenzV:
```suggestion
/// Saved state for recording operations.
```

grebmeg: Oops 😊 Thanks!

PR #1453, opened 2026-02-18
vello_sparse_strips: Optimize pixel-aligned rect filling by bypassing path pipeline

grebmeg:
This PR adds a fast path for filling pixel-aligned rectangles that skips the full path processing pipeline (flatten → tiles → strips) and instead generates strip coverage data directly from the rectangle geometry. When `fill_rect` is called and both the transform and rect coordinates are pixel-aligned (integer translation + integer rect bounds), the renderer calls `fill_rect_fast` which produces strips directly, avoiding the overhead of path flattening and tiling. When these conditions aren't met, the renderer falls back to the standard pipeline, ensuring pixel-identical output in all cases.

```shell
cargo bench -p vello_bench -- render_rect
    Finished `bench` profile [optimized] target(s) in 0.12s
     Running unittests src/lib.rs (target/release/deps/vello_bench-71be2072ddfcacb3)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/main.rs (target/release/deps/main-706649511d898baa)
render_rect/14x14_via_path
                        time:   [278.46 ns 279.32 ns 280.08 ns]
                        change: [-2.5266% +0.5821% +2.7794%] (p = 0.75 > 0.05)
  1 (2.00%) high mild
render_rect/14x14_via_rect
                        time:   [33.874 ns 33.977 ns 34.075 ns]
                        change: [-0.4290% -0.0494% +0.2754%] (p = 0.78 > 0.05)
  2 (4.00%) high mild

```


sparse_strips/vello_common/src/strip.rs
```diff
@@ -431,3 +434,184 @@ fn render_impl<S: Simd>(
         accumulated_winding += acc;
     }
 }
+
```
LaurenzV: Do we maybe want to move all of this into a separate `rect.rs`? Up to you!

sparse_strips/vello_common/src/util.rs
```diff
@@ -61,6 +61,33 @@ pub fn normalized_mul_u8x16<S: Simd>(a: u8x16<S>, b: u8x16<S>) -> u16x16<S> {
...
 }
 
+/// Check if an affine transform is a pure integer translation.
+///
+/// Returns true if the transform only contains integer translation (no rotation,
+/// skew, or scaling), meaning rectangles will remain pixel-aligned after transformation.
+#[inline]
+pub fn is_integer_translation(transform: &Affine) -> bool {
+    let [a, b, c, d, e, f] = transform.as_coeffs();
+    (a - 1.0).abs() < 1e-9
```
LaurenzV: Can we reuse `SCALAR_NEARLY_ZERO` here?

LaurenzV: There's also a `is_nearly_zero` method, although it's only defined for f32. Maybe we can define it for f64 as well.

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -476,6 +476,31 @@ impl Dispatcher for MultiThreadedDispatcher {
...
+        blend_mode: BlendMode,
+        mask: Option<Mask>,
+        _encoded_paints: &[EncodedPaint],
+    ) {
+        // For multi-threaded, fall back to path-based rendering.
+        // TODO: Implement optimized rect strip generation in worker threads.
+        let path = rect.to_path(0.0);
+        let start = self.allocation_group.path.len() as u32;
+        self.allocation_group.path.extend(&path);
+        let end = self.allocation_group.path.len() as u32;
```
LaurenzV: Is there any difficulty to implementing this? If so, could we at least render the rectangle directly into `allocation_group.path` instead of allocating a new path using `to_path`?

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -566,6 +566,34 @@ impl Dispatcher for SingleThreadedDispatcher {
...
+        let wide = &mut self.wide;
+
+        // Generate strips directly for the rectangle (bypasses path processing).
+        self.strip_generator.generate_filled_rect_fast(
+            rect,
+            &mut self.strip_storage,
+            self.clip_context.get(),
+        );
+
+        // Generate coarse-level commands from strips (layer_id 0 = root layer).
```
LaurenzV: Isn't 0 the thread index? Or why we can just assume the root layer here?

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -175,6 +175,56 @@ impl StripGenerator {
...
+            intersect(self.level, clip_data, rect_data, strip_storage);
+        } else {
+            strip::render_rect_fast(
+                self.level,
+                clamped,
+                &mut strip_storage.strips,
+                &mut strip_storage.alphas,
+            );
+        }
+    }
```
LaurenzV: Is it possible to deduplicate this part with normal case? I think it should be if you create a new function that takes a closure for the strip rendering part.

sparse_strips/vello_common/src/strip_generator.rs
```diff
@@ -214,4 +264,186 @@ mod tests {
         assert!(generator.line_buf.is_empty());
         assert!(storage.is_empty());
     }
+
```
LaurenzV, line 264: It seems to me like all these method could also be made shorter by creating a new function (or just repurposing `assert_strips_equal` that takes a `Rect` and test name as input and does the whole logic.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -431,3 +434,184 @@ fn render_impl<S: Simd>(
...
+    let y0 = (rect_y0 / Tile::HEIGHT) * Tile::HEIGHT;
+    let y1 = (rect_y1.saturating_add(Tile::HEIGHT - 1) / Tile::HEIGHT) * Tile::HEIGHT;
+    // Include one tile past the right edge so the right-edge tile column is
+    // covered by the edge-row wide-strip loop.
+    let x_end = right_tile_x.saturating_add(Tile::WIDTH);
+
+    if x_end <= left_tile_x || y1 <= y0 {
+        return;
+    }
```
LaurenzV: This means negative rectangles aren't handled correctly, right? We can probably ignore that case for now, but might be good to fix in the future.

sparse_strips/vello_common/src/strip.rs
```diff
@@ -431,3 +434,184 @@ fn render_impl<S: Simd>(
...
+
+    let tile_start_y = y0 / Tile::HEIGHT;
+    let tile_end_y = y1 / Tile::HEIGHT;
+
+    let (max_strips, max_alphas) =
+        calc_rect_reserve_capacity(x_end - left_tile_x, tile_end_y - tile_start_y);
+    strip_buf.reserve(max_strips);
+    alpha_buf.reserve(max_alphas);
```
LaurenzV: I'm wondering whether this is worth the additional "complexity" given that the strip and alpha buffers are reused across each frame for each path, so I think it won't have any effect in 99%+ of the cases. But up to you!

sparse_strips/vello_common/src/strip.rs
LaurenzV: (by that I mean maybe adding a TODO)

sparse_strips/vello_common/src/strip.rs
grebmeg: Fixed

sparse_strips/vello_common/src/util.rs
grebmeg: Fixed

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
grebmeg: There’s no major difficulty involved, but it does require some additional time, which I’d prefer to focus on later.

Fixed.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
grebmeg: Yes, you're right.

Updated all comments.

sparse_strips/vello_common/src/strip_generator.rs
grebmeg: Yep, that's possible.

Fixed.

sparse_strips/vello_common/src/strip_generator.rs
grebmeg: Nice idea! Thanks!

Fixed.

sparse_strips/vello_common/src/strip.rs
grebmeg: Yes, you're right. I've added TODO here.

sparse_strips/vello_common/src/strip.rs
grebmeg: Actually, I agree! I don’t think this adds enough value to justify keeping it here.

Removed.

PR #1454, opened 2026-02-18
Add fast-path for skipping coarse rasterization and scheduling for scenes without layers

laurenz-canva:
This adds a fast path to vello_hybrid, where we avoid coarse rasterization and scheduling for all strips as long as no `push_layer` command is encountered.

As Alex rightly highlighted, this does have the disadvantage of not allowing the "if there is an opaque fill, clear all previous fill" optimization. However, it seems to me like this should be overshadowed by the improvements that come from not doing scheduling and coarse rasterization. Here are the timings for rendering 1000 frames of the GhotScript tiger:

Before (note in particular `Wide::generate` and `Scheduler::do_scene`:
image

After:
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -196,7 +196,7 @@ impl WebGlRenderer {
             self.fast_path_gpu_strips.clear();
 
             generate_gpu_strips_for_fast_path(
-                &scene.fast_strips_buffer,
+                &scene.fast_strips_buffer.paths,
                 scene,
                 &self.paint_idxs,
                 &mut self.fast_path_gpu_strips,
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -41,10 +41,11 @@ pub(crate) struct FastStripsPath {
 }
 
 /// A buffer that collects strips from paths that bypass coarse rasterization and scheduling.
+///
+/// Strip data itself lives in `strip_storage`. Each `FastStripsPath` records the range of strips
+/// for one path within that storage.
 #[derive(Debug, Default)]
 pub(crate) struct FastStripsBuffer {
-    /// The strips for all paths in the buffer.
-    pub(crate) strips: Vec<Strip>,
     /// All paths in the buffer.
     pub(crate) paths: Vec<FastStripsPath>,
 }
@@ -52,7 +53,6 @@ pub(crate) struct FastStripsBuffer {
 impl FastStripsBuffer {
     #[inline(always)]
     fn clear(&mut self) {
-        self.strips.clear();
         self.paths.clear();
     }
 }
@@ -157,19 +157,22 @@ pub struct Scene {
 }
 
 // We use this macro instead of a method to avoid borrowing issues in the corresponding methods.
+//
+// When the fast path is active it is in `Append` mode, so `$strip_start` (captured before generation)
+// and the current length delimit the range for this path.
+//
+// When the fast path is inactive, `strip_storage` is in `Replace` mode where each generation starts
+// with a clear, so the whole buffer is the current path's strips.
 macro_rules! submit_strips {
-    ($self:ident, $strips:expr, $paint:expr) => {
+    ($self:ident, $strip_storage:expr, $strip_start:expr, $paint:expr) => {
         if $self.strips_fast_path_active {
-            let start = $self.fast_strips_buffer.strips.len();
-            $self.fast_strips_buffer.strips.extend_from_slice($strips);
-            let end = $self.fast_strips_buffer.strips.len();
             $self.fast_strips_buffer.paths.push(FastStripsPath {
-                strips: start..end,
+                strips: $strip_start..$strip_storage.strips.len(),
                 paint: $paint,
             });
         } else {
             $self.wide.generate(
-                $strips,
+                &$strip_storage.strips,
                 $paint,
                 $self.blend_mode,
                 0,
@@ -202,7 +205,8 @@ impl Scene {
             paint_visible: true,
             stroke: render_state.stroke,
             strip_generator: StripGenerator::new(width, height, settings.level),
-            strip_storage: RefCell::new(StripStorage::default()),
+            // Start strip storage in `Append` mode since we enable the fast path by default.
+            strip_storage: RefCell::new(StripStorage::new(GenerationMode::Append)),
             transform: render_state.transform,
             fill_rule: render_state.fill_rule,
             blend_mode: render_state.blend_mode,
@@ -289,6 +293,7 @@ impl Scene {
         aliasing_threshold: Option<u8>,
     ) {
         let strip_storage = &mut self.strip_storage.borrow_mut();
+        let strip_start = strip_storage.strips.len();
         self.strip_generator.generate_filled_path(
             path,
             fill_rule,
@@ -298,7 +303,7 @@ impl Scene {
             self.clip_context.get(),
         );
 
-        submit_strips!(self, &strip_storage.strips, paint);
+        submit_strips!(self, strip_storage, strip_start, paint);
     }
 
     /// Push a new clip path to the clip stack.
@@ -347,6 +352,7 @@ impl Scene {
         aliasing_threshold: Option<u8>,
     ) {
         let strip_storage = &mut self.strip_storage.borrow_mut();
+        let strip_start = strip_storage.strips.len();
         self.strip_generator.generate_stroked_path(
             path,
             &self.stroke,
@@ -356,7 +362,7 @@ impl Scene {
             self.clip_context.get(),
         );
 
-        submit_strips!(self, &strip_storage.strips, paint);
+        submit_strips!(self, strip_storage, strip_start, paint);
     }
 
     /// Set the aliasing threshold.
@@ -406,13 +412,14 @@ impl Scene {
         let paint = self.encode_current_paint();
         let transformed_rect = self.transform.transform_rect_bbox(*rect);
         let strip_storage = &mut self.strip_storage.borrow_mut();
+        let strip_start = strip_storage.strips.len();
         self.strip_generator.generate_filled_rect_fast(
             &transformed_rect,
             strip_storage,
             self.clip_context.get(),
         );
 
-        submit_strips!(self, &strip_storage.strips, paint);
+        submit_strips!(self, strip_storage, strip_start, paint);
     }
 
     /// Stroke a rectangle with the current paint and stroke settings.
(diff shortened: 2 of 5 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1167,6 +1168,68 @@ fn has_non_zero_alpha(rgba: u32) -> bool {
     rgba >= 0x1_00_00_00
 }
 
+/// Build `GpuStrip`s directly from the fast path buffer, bypassing coarse rasterization
+/// and scheduling.
+///
+/// This replicates the strip→GpuStrip conversion that normally happens across
+/// `Wide::generate` + `Scheduler::do_tile`, but for the simple case where all draws
+/// happen at depth=1 directly to the surface with no layers or blending.
+pub(crate) fn build_gpu_strips_direct(
```
laurenz-canva: We can probably reuse existing code here, as mentioned I haven't cleaned this up yet, just a PoC.

LaurenzV:
Sounds good, looking forward to the PR! Yes you could probably optimize this even further, but this is the bare minimum that should already be a good improvement in many cases. 😄

taj-p:
One thing I didn't state but that I hope is implied: AMAZING to have a POC so quickly to validate the approach. Very cool!!! 🎉

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: I've looked into it, but doesn't seem to be straightforward. Since the logic is relatively simple, I'll leave it as it is right now.

LaurenzV:
@taj-p Should be ready, if you want to you can apply any changes you want tomorrow and then merge this PR yourself. Then you can get started on the follow ups if you want. I see three (more or less independent) potential follow ups, two of which we already discussed:

1) Change it so the user can still pass a "no blending" flag. In this case, we can extend the logic so that we do an interleaving of direct strips rendering + coarse rasterization for specific ranges, as is done in the blit rect pipeline. Each switch will invoke at least one new render pass, but this should still be pretty efficient.

2) As discussed, add a variant "RectStrip" to GpuStrip, so that rectangles can be handled directly in the fragment shader without requiring any additional CPU work.

3) It might be worth it to add a flag "is_dirty" to `Wide` to track whether anything has actually been pushed there. If not, when calling `reset` on `Wide` it can simply be a no-op. I think that could be worth it, because I'm seeing `Wide::reset` show up in the performance profile.

LaurenzV:
I'm not too worried about that, basically any test that invokes layers (which includes clip layers, blending, etc.) will use the old path. Aside from that, some parts of it are also indirectly still tested by the CPU tests. Or is there anything in particular you think that isn't getting enough test coverage anymore?

taj-p:
> I'm not too worried about that, basically any test that invokes layers (which includes clip layers, blending, etc.) will use the old path. Aside from that, some parts of it are also indirectly still tested by the CPU tests. Or is there anything in particular you think that isn't getting enough test coverage anymore?

My concern is with long term maintenance. I agree it may not seem immediately required today, but as the implementation evolves, there is risk of drift. `generate_gpu_strips_for_fast_path` acknowledges some duplication with coarse, for example. My tendency is towards "better safe than sorry" - let's leave this for now and I can investigate it when adding the "no blends" flag to assess impact on test run times.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -187,12 +192,35 @@ impl WebGlRenderer {
...
+                &mut self.fast_path_gpu_strips,
+            );
+
+            if !self.fast_path_gpu_strips.is_empty() {
+                let mut ctx = WebGlRendererContext {
+                    programs: &mut self.programs,
+                    gl: &self.gl,
+                };
+                ctx.render_strips(&self.fast_path_gpu_strips, 2, LoadOp::Clear);
+            }
```
```suggestion
            let mut ctx = WebGlRendererContext {
                programs: &mut self.programs,
                gl: &self.gl,
            };
            ctx.render_strips(&self.fast_path_gpu_strips, 2, LoadOp::Clear);
```

We already check for empty strips in `render_strips`:

https://github.com/linebender/vello/blob/d9786db93830a01efc22460b7853256ef1f3989e/sparse_strips/vello_hybrid/src/render/webgl.rs#L1612-L1614

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -30,6 +30,33 @@ use vello_common::util::{is_integer_rect, is_integer_translation};
 /// Default tolerance for curve flattening
 pub(crate) const DEFAULT_TOLERANCE: f64 = 0.1;
 
+/// Metadata for a single path stored in the fast strips buffer.
+#[derive(Debug)]
+pub(crate) struct FastStripsPath {
+    /// The index of the first strip of the path in the `strips` buffer.
+    pub(crate) strip_start: usize,
+    /// The (exclusive) last index of the path in the `strips` buffer.
+    pub(crate) strip_end: usize,
```
taj-p: Why not store a `strips: Range<usize>`?

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: And it makes us consistent with the WGPU impl

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -30,6 +30,33 @@ use vello_common::util::{is_integer_rect, is_integer_translation};
...
+    /// The paint of the path.
+    pub(crate) paint: Paint,
+}
+
+/// A buffer that collects strips from path that bypass coarse rasterization and scheduling.
```
```suggestion
/// A buffer that collects strips from paths that bypass coarse rasterization and scheduling.
```

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -30,6 +30,33 @@ use vello_common::util::{is_integer_rect, is_integer_translation};
...
+#[derive(Debug, Default)]
+pub(crate) struct FastStripsBuffer {
+    /// The strips for all paths in the buffer.
+    pub(crate) strips: Vec<Strip>,
+    /// All paths in the buffer.
+    pub(crate) paths: Vec<FastStripsPath>,
+}
+
+impl FastStripsBuffer {
+    fn clear(&mut self) {
```
taj-p, line 55:
```suggestion
    #[inline(always)]
    fn clear(&mut self) {
```

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -396,6 +464,8 @@ impl Scene {
         mask: Option<Mask>,
         filter: Option<Filter>,
     ) {
+        self.flush_fast_path();
```
taj-p: We should note that fast paths like these mean that blend modes are strictly layer level concepts. We won't support per-path blend modes, so we should probably remove the `set_blend_mode` API on scene at some point. We don't even read it when creating a layer.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -116,6 +143,42 @@ pub struct Scene {
...
+    /// don't need to do coarse rasterization or scheduling and therefore save a lot of overhead.
+    pub(crate) strips_fast_path_active: bool,
+}
+
+// We use this macro instead of a method to avoid borrowing issues in the corresponding methods.
+macro_rules! submit_strips {
+    ($self:ident, $strips:expr, $paint:expr) => {
+        if $self.strips_fast_path_active {
+            let strip_start = $self.fast_strips_buffer.strips.len();
+            $self.fast_strips_buffer.strips.extend_from_slice($strips);
```
taj-p: This copy seems unnecessary. Can we set the strip_storage to `append` mode and use that allocation instead of repeatedly copying the strips into dedicated storage?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -384,6 +427,31 @@ impl Scene {
...
+            let paint = self.fast_strips_buffer.paths[i].paint.clone();
+            self.wide.generate(
+                &self.fast_strips_buffer.strips[start..end],
+                paint,
+                BlendMode::default(),
+                0,
+                None,
+                &self.encoded_paints,
+            );
+        }
```
```suggestion
        for path in &self.fast_strips_buffer.paths {
            self.wide.generate(
                &self.fast_strips_buffer.strips[path.strips.clone()],
                path.paint.clone(),
                BlendMode::default(),
                0,
                None,
                &self.encoded_paints,
            );
        }
```

taj-p: https://github.com/linebender/vello/pull/1454/commits/c9c21a946a3e55babd33999f6a61126d2ac2021f

sparse_strips/vello_hybrid/src/scene.rs
taj-p: Implemented in https://github.com/linebender/vello/pull/1454/commits/c9c21a946a3e55babd33999f6a61126d2ac2021f

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -384,6 +426,27 @@ impl Scene {
...
+    /// This retroactively generates wide tile commands for all strips that have been generated
+    /// using the fast path.
+    fn flush_fast_path(&mut self) {
+        if !self.strips_fast_path_active {
+            return;
+        }
+        for path in &self.fast_strips_buffer.paths {
+            self.wide.generate(
+                &self.fast_strips_buffer.strips[path.strips.clone()],
+                path.paint.clone(),
```
taj-p: We can drain here to avoid the clone

taj-p: Implemented in https://github.com/linebender/vello/pull/1454/commits/7fedef920f0bc360a65cc4c0c8d896d5faa89018

sparse_strips/vello_hybrid/src/scene.rs
taj-p: Fixed in https://github.com/linebender/vello/pull/1454/commits/7fedef920f0bc360a65cc4c0c8d896d5faa89018.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -30,6 +31,32 @@ use vello_common::util::{is_integer_rect, is_integer_translation};
...
+    /// The paint of the path.
+    pub(crate) paint: Paint,
+}
+
+/// A buffer that collects strips from paths that bypass coarse rasterization and scheduling.
+///
+/// Strip data itself lives in `strip_storage`. Each `FastStripsPath` records the range of strips
+/// for one path within that storage.
+#[derive(Debug, Default)]
+pub(crate) struct FastStripsBuffer {
```
taj-p, line 48: NOTE: We kept this wrapper struct around because I expect it to be used to support batching in the next PRs.

LaurenzV:
@grebmeg Will merge nw, feel free to take another look tomorrow!

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -116,6 +143,44 @@ pub struct Scene {
...
+    pub(crate) fast_strips_buffer: FastStripsBuffer,
+    /// Whether the fast path for rendering strips directly instead of going through coarse
+    /// rasterization is active.
+    ///
+    /// This is the case if we have not performed any `push_layer` command. In such a case, we
+    /// don't need to do coarse rasterization or scheduling and therefore save a lot of overhead.
+    pub(crate) strips_fast_path_active: bool,
+}
+
+// We use this macro instead of a method to avoid borrowing issues in the corresponding methods.
```
grebmeg, line 159: I guess that’s hard to avoid, but I’d be curious to explore the options here.

PR #1456, opened 2026-02-18
Blit Rect Pipeline

taj-p:
## Intent

This PR supercharges Vello Hybrid's `fill_rect` method with a fast path for texture copies. Instead of running through the sparse strip pipeline, we upload the quad to the GPU directly. The cases for which we can do this are listed below:

https://github.com/taj-p/vello/blob/386e4bbb3d8b33d68fa8496837de9d0a343f71bd/sparse_strips/vello_hybrid/src/scene.rs#L490-L577

## Performance

The performance is very good for scenes that consist of many images. Once we have glyph caching, I believe this will be the fastest way to render from the glyph atlas. This is likely the "speed of light" for a scene that consists of non-blended / non-clipped images and text.

image

<details>
<summary>More perf benches</summary>

I was a bit worried because in that prior benchmark we use `gl.finish()` as the "stop time" between runs. I also compared against a pixel read back to "force" a full GPU flush and saw the following results. These are still very promising.

image

</details>

## Design

The design is pretty straightforward. 

- If we have a valid `fill_rect` that **doesn't** overlap any pending `[fill|stroke]_path`, then batch it into a blit rect pipeline pass. 
- If the `fill_rect` **does** overlap, then flush the strips and start a new blit rect batch.

The "overlap" regions of the strips are tracked via a SIMD-enabled `DirtyRects` struct. Note that `pop_layer` dirties the wide tiles that it covered in its layer.

We saw significant performance improvement by calculating path bounding boxes by iterating over the `line_buf` instead of performing `path.bounding_box()`.

## Degenerate cases

The blit rect pipeline currently uses a naive batching mechanism (see ## Design). It is possible to build a degenerate scene with this batching mechanism that causes flip-flop pipeline switching between blit and strips. 

There are some scenes that are not optimisable. Consider the below scene:

image

This scene sees rows of overlapping bordered rects. You _must_ draw the rect before the next stroke (i.e. border) before the next rect and so on. See below ASCII visualisation:

```
Card 0:  [===== blit_0 =====]
         [===== stroke_0 ===]
Card 1:       [===== blit_1 =====]     ← overlaps stroke_0 (pipeline switch)
              [===== stroke_1 ===]
Card 2:            [===== blit_2 =====] ← overlaps stroke_1 (pipeline switch)
``` 

To get a feel for this impact, see the below regressions of this type of scene:

image

We could consider adding runtime "opt out" of batching when the batch only consists of 1 blit or similar. But, at least for my use case, I would be surprised to find scenes with 100s of overlapping bordered images with strokes in between. (Note that overlapping images are fine, but 100s contiguously overlapping image/path are not).

If we're concerned about this, I think I propose (to be approved by reviewers):
1. Merge as-is into Vello (and remove once sparse strip pipeline is good enough)
2. Merge after introducing runtime opt-out mechanism for small blit batches
3. Review, but merge into fork
4. Add additional documentation to `expect_only_default_blending`
5. ?? Maybe you can think of an even better idea! 🙏 

Note that I dabbled in other batching mechanisms, but I think, for now, this might be good enough to at least "test the waters" of blit rect.

## Testing

- We run some handcrafted tests with and without blit batching.
- Every snapshot has been modified to run on both Vello Hybrid with and without the blit rect pipeline.

### DIFF
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -602,8 +602,6 @@ impl Scene {
             return false;
         }
 
-        // All conditions are met! We are rendering this blit in the fast path!
-
         // Pre-transform rect dimensions in geometry space.
         let rect_wh = [
             pixel_snap(rect.x1 - rect.x0).max(0.0) as u16,
@@ -613,6 +611,8 @@ impl Scene {
         if rect_wh[0] == 0 || rect_wh[1] == 0 {
             return true; // Zero-size rect, nothing to draw.
         }
+
+        // All conditions are met! We are rendering this blit in the fast path!
 
         // Compute the screen-space quad via center + column vectors.
         //
@@ -782,13 +782,7 @@ impl Scene {
     /// Pop the last pushed layer.
     pub fn pop_layer(&mut self) {
         self.enter_strip_mode();
-        let layer_bbox = self.wide.pop_layer(&mut self.render_graph);
-        if self.render_hints.blit_rect_pipeline_enabled() && !layer_bbox.is_inverted() {
-            // Push the dirty rect for the layer to the dirty rects list.
-            let [x0, y0, x1, y1] = layer_bbox.pixel_bounds();
-            self.strips_dirty_rects
-                .push(x0 as u16, y0 as u16, x1 as u16, y1 as u16);
-        }
+        self.wide.pop_layer(&mut self.render_graph);
     }
 
     /// Set the blend mode for subsequent rendering operations.
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -714,9 +714,7 @@ impl<const MODE: u8> Wide<MODE> {
     /// - Generating filter commands for each tile
     /// - Popping any associated clip
     /// - Applying mask, opacity, and blend mode operations if needed
-    ///
-    /// It returns the bounding box of the layer in wide tile coordinates.
-    pub fn pop_layer(&mut self, render_graph: &mut RenderGraph) -> WideTilesBbox {
+    pub fn pop_layer(&mut self, render_graph: &mut RenderGraph) {
         // This method basically unwinds everything we did in `push_layer`.
         let mut layer = self.layer_stack.pop().unwrap();
 
@@ -761,13 +759,11 @@ impl<const MODE: u8> Wide<MODE> {
             }
         }
 
-        let layer_bbox = layer.wtile_bbox;
-
         // Union this layer's bbox into the parent layer's bbox.
         // This ensures the parent knows about all tiles used by this child layer,
         // which is important for filter effects that may expand beyond the original content bounds.
         if let Some(parent_layer) = self.layer_stack.last_mut() {
-            parent_layer.wtile_bbox.union(layer_bbox);
+            parent_layer.wtile_bbox.union(layer.wtile_bbox);
         }
 
         if layer.clip {
@@ -805,8 +801,6 @@ impl<const MODE: u8> Wide<MODE> {
         if in_clipped_filter_layer {
             self.clipped_filter_layer_depth -= 1;
         }
-
-        layer_bbox
     }
 
     /// Adds a clipping region defined by the provided strips.


sparse_strips/vello_sparse_tests/tests/mix.rs
```diff
@@ -78,27 +78,27 @@ fn mix_normal(ctx: &mut impl Renderer) {
     mix(ctx, BlendMode::new(Mix::Normal, Compose::SrcOver));
 }
 
-#[vello_test(cpu_u8_tolerance = 1, hybrid_tolerance = 1)]
+#[vello_test(cpu_u8_tolerance = 1, hybrid_tolerance = 1, uses_blends)]
```
LaurenzV, line 81: Do you think it might make sense to automatically assume any test with "mix" or "compose" in the name uses blends? Then we dont have to annotate those ourselves.

sparse_strips/vello_sparse_tests/tests/batching.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn make_test_image() -> Pixmap {
+        let mut pixmap = Pixmap::new(10, 10);
+        for pixel in pixmap.data_mut() {
+            *pixel = vello_common::peniko::color::PremulRgba8 {
+                r: 0,
+                g: 128,
+                b: 255,
+                a: 255,
+            };
+        }
```
LaurenzV, line 37: Could probably shorten this to `pixmap.data_mut().fill(pixel)`?

sparse_strips/vello_sparse_tests/tests/batching.rs
```diff
@@ -0,0 +1,366 @@
...
+                "Rendering with blit batching enabled differs from rendering with blit batching \
+                 disabled: {diff_count} pixels differ out of {}\nFirst diffs:\n{}",
+                data_on.len() / 4,
+                first_diffs.join("\n")
+            );
+        }
+    }
+
+    /// Serialise concurrent GPU tests to avoid wgpu segfaults.
+    static GPU_MUTEX: Mutex<()> = Mutex::new(());
```
LaurenzV, line 94: Since this is a separate mutex used by the other vello_hybrid test, could it not still happen that one blit test and a normal test run concurrently?

LaurenzV: Ah, I guess tests from different modules don't run at the same time, right?

sparse_strips/vello_sparse_shaders/shaders/blit_rects.wgsl
```diff
@@ -0,0 +1,93 @@
...
+    var out: VertexOutput;
+    out.position = vec4<f32>(ndc_x, ndc_y, 0.0, 1.0);
+    out.uv = uv;
+    out.atlas_index = instance.atlas_index;
+    return out;
+}
+
+@fragment
+fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
+    return textureSample(atlas_texture_array, atlas_sampler, in.uv, in.atlas_index);
```
LaurenzV, line 92: So if bilinear filtering is enabled, this will use the GPU-native image sampler? Won't this cause issues at the border if a different image is in the same image atlas?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -155,23 +158,85 @@ impl Renderer {
...
+                    batch_idx == 0,
+                )?;
+            }
+
+            // Render blits for this batch.
+            {
+                let blits = &scene.all_blits[blit_batch.clone()];
+                if blits.is_empty() {
+                    continue;
+                }
```
LaurenzV, line 199: Haven't looked into it yet, but is there no way of early-optimizing this? i.e. not creating this batch in the first place (or merging it with the next one) if there are no blits.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -521,6 +617,18 @@ struct ClearSlotsConfig {
     pub _padding: u32,
 }
 
+/// Config uniform for the blit rects pipeline.
+#[repr(C)]
```
LaurenzV, line 621: maybe add align(16) as well for safe measure?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1600,6 +1896,47 @@ impl RendererContext<'_> {
         render_pass.draw(0..4, 0..u32::try_from(strips.len()).unwrap());
     }
 
+    /// Render blit rects to the final view in a single instanced draw call.
+    fn do_blit_render_pass(&mut self, blit_rects: &[GpuBlitRect]) {
+        if blit_rects.is_empty() {
+            return;
+        }
+
+        // Upload blit rect instance data, reusing the buffer if it's large enough.
```
LaurenzV, line 1905: Is it really reusing the buffer here? Seems like it's always writing it anew?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -65,6 +67,139 @@ pub struct GpuStrip {
...
+///   vertex(x, y) = center + col0 * (x - 0.5) + col1 * (y - 0.5)
+/// ```
+///
+/// where `(x, y)` can be `(0, 0)`, `(1, 0)`, `(0, 1)`, or `(1, 1)` (i.e. each
+/// corner of a quad).
+///
+/// You can think of `col0` and `col1` as a pre-evaluated affine so that the shader
+/// need only do a weighted sum instead of matrix multiplication (and sending those
+/// additional bytes to the GPU).
+#[repr(C)]
```
LaurenzV, line 88: Let's also enfore alignment here?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -65,6 +67,139 @@ pub struct GpuStrip {
...
+    //
+    // Note that we don't account for the image offset in the atlas until calculating `src_x` and `src_y` below.
+
+    let src_x0 = blit.img_origin[0].max(0.0);
+    let src_y0 = blit.img_origin[1].max(0.0);
+    let src_x1 = (blit.img_origin[0] + blit.rect_wh[0] as f32).min(resource.width as f32);
+    let src_y1 = (blit.img_origin[1] + blit.rect_wh[1] as f32).min(resource.height as f32);
+
+    let src_w = (src_x1 - src_x0).round() as u16;
+    let src_h = (src_y1 - src_y0).round() as u16;
```
LaurenzV, line 147: Hmm, I mean this makes sense I guess... But it does mean that the behavior here is different than to the normal strip pipeline, right? Here, if the width is for example 3.4, the last pixel wouldn't be painted at all, while in the strip pipeline it would be painted with an additional opacity applied, is that right?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -65,6 +67,139 @@ pub struct GpuStrip {
...
+
+    if src_w == 0 || src_h == 0 {
+        return None;
+    }
+
+    // Not all images in the Atlas are in the top right. We need to apply their offset to
```
LaurenzV, line 153: Should this read "Not all images in the Atlas are in the top left"?

sparse_strips/vello_hybrid/src/render/common.rs
LaurenzV: Could we maybe add a test for this just to be sure?

sparse_strips/vello_sparse_tests/tests/mod.rs
LaurenzV, line 1: By the way, I checked and it seems like none of the test cases in `image.rs` actually trigger the `blit_rect` fast path. We probably should add some more test cases (with extend pad) to make sure that path is actually triggered.

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -65,6 +67,139 @@ pub struct GpuStrip {
...
+    // the screen-space quad for the visible sub-region.
+    //
+    // How to think about `col0` and `col1`?
+    //
+    // `col0` defines the vector to go from the center to the right edge of the rect.
+    // `col1` defines the vector to go from the center to the bottom edge of the rect.
+    //
+    // Thus, we can put them together with the below formula to get each vertex of a quad:
+    //
+    //   vertex(x, y) = center + col0 * (x - 0.5) + col1 * (y - 0.5)
```
LaurenzV, line 168: Why -0.5 and not +0.5? Could you maybe add more documentation for that?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -65,6 +67,139 @@ pub struct GpuStrip {
...
+    // Calculate `col0`, `col1`, and `center`, which, when put together, can describe
+    // the screen-space quad for the visible sub-region.
+    //
+    // How to think about `col0` and `col1`?
+    //
+    // `col0` defines the vector to go from the center to the right edge of the rect.
+    // `col1` defines the vector to go from the center to the bottom edge of the rect.
+    //
+    // Thus, we can put them together with the below formula to get each vertex of a quad:
+    //
```
LaurenzV, line 167: Also, I haven't tried to fully understand the math here, but it seems like most of these is just applying transformations to vectors here. Would it make sense to reuse some of the methods from kurbo here? The only downside I see is that they use f64... But perhaps it's not worth it, up to you.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -481,20 +481,41 @@ impl Scheduler {
...
             for wide_tile_col in 0..wide_tiles_per_row {
-                let wide_tile = scene.wide.get(wide_tile_col, wide_tile_row);
+                let wide_tile = wide.get(wide_tile_col, wide_tile_row);
+
+                let tile_idx = usize::from(wide_tile_row) * usize::from(wide_tiles_per_row)
+                    + usize::from(wide_tile_col);
+                let cmd_start = cmd_starts.map_or(0, |starts| starts[tile_idx]);
+                let cmd_end = cmd_ends
+                    .map(|ends| ends[tile_idx])
+                    .unwrap_or(wide_tile.cmds.len());
```
LaurenzV, line 513: Probably can use map_or here as well for consistency? 😄

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -481,20 +481,41 @@ impl Scheduler {
...
+
+                if cmd_start == cmd_end && !is_first_segment {
```
LaurenzV, line 515: Why only if `!is_first_segment`?

LaurenzV: Oh, it's because of the background... Maybe add a comment for that?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -605,6 +709,18 @@ struct ClearSlotsConfig {
     pub _padding: u32,
 }
 
+/// Config uniform for the blit rects pipeline.
+#[repr(C)]
```
LaurenzV, line 713: Same as for the other ones, can we define alignment explicitly?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1700,6 +1895,49 @@ fn initialize_clear_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources)
...
+/// Initialize blit rect VAO.
+fn initialize_blit_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources) {
+    gl.bind_vertex_array(Some(&resources.blit_vao));
+    gl.bind_buffer(
+        WebGl2RenderingContext::ARRAY_BUFFER,
+        Some(&resources.blit_buffer),
+    );
+
+    let stride = size_of::<GpuBlitRect>() as i32;
+    debug_assert_eq!(stride, 36, "expected stride of 36");
```
LaurenzV, line 1907: could we not make this a constant assertion?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -328,9 +504,179 @@ impl Scene {
...
+        let half_col0_x = col0[0] * 0.5;
+        let half_col0_y = col0[1] * 0.5;
+        let half_col1_x = col1[0] * 0.5;
+        let half_col1_y = col1[1] * 0.5;
+        let extent_x = half_col0_x.abs() + half_col1_x.abs();
+        let extent_y = half_col0_y.abs() + half_col1_y.abs();
+        let aabb_x0 = center_xy[0] - extent_x;
+        let aabb_y0 = center_xy[1] - extent_y;
+        let aabb_x1 = center_xy[0] + extent_x;
+        let aabb_y1 = center_xy[1] + extent_y;
```
LaurenzV, line 642: Could we not use kurbo stuff here and convert the end result to f32?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -865,3 +1255,377 @@ impl Scene {
...
+            let bits = i16x8 {
+                val: mask.val,
+                simd: s,
+            };
+
+            // Compute a pair and ([a0&a1, a2&a3, b0&b1, b2&b3, …])
+            let paired = bits.unzip_low(bits).and(bits.unzip_high(bits));
+
+            // Check if any pair is non-zero.
+            if (paired.val[0] & paired.val[1]) != 0 || (paired.val[2] & paired.val[3]) != 0 {
```
LaurenzV, line 1398: Let's add a TODO that we can use https://github.com/linebender/fearless_simd/pull/141 in the future.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -210,6 +306,73 @@ impl Scene {
...
+
+    /// Check whether a blit rect can be batched into the previous blit batch
+    /// without creating a new pipeline switch.
+    ///
+    /// Returns `true` when either:
+    /// - We are already in blit mode
+    /// - A previous blit batch exists and the blit rect does not overlap any
+    ///   strip operations recorded since that blit batch
+    #[inline(always)]
+    fn can_batch_blit_aabb(&self, aabb_x0: f32, aabb_y0: f32, aabb_x1: f32, aabb_y1: f32) -> bool {
```
LaurenzV, line 332: What's `aabb`? 😄

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -210,6 +306,73 @@ impl Scene {
...
+    ///
+    /// In other words, if a blit wants to render over `bbox`, we must first flush the current strips.
+    #[inline(always)]
+    fn push_dirty_rect(&mut self, bbox: [f32; 4]) {
+        let w = u32::from(self.width);
+        let h = u32::from(self.height);
+        let x0 = (bbox[0].floor().max(0.0) as u32).min(w) as u16;
+        let y0 = (bbox[1].floor().max(0.0) as u32).min(h) as u16;
+        let x1 = (bbox[2].ceil().max(0.0) as u32).min(w) as u16;
+        let y1 = (bbox[3].ceil().max(0.0) as u32).min(h) as u16;
```
LaurenzV, line 365: That looks very simiklar to the code in `can_batch_blit_aabb`, maybe we can deduplicate somehow?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -328,9 +504,179 @@ impl Scene {
...
+                match &img.image {
+                    ImageSource::OpaqueId(id) => *id,
+                    _ => return false,
+                }
+            }
+            _ => return false,
+        };
+
+        // Condition 4: Blend mode must be default SrcOver.
+        const DEFAULT_BLEND: BlendMode = BlendMode::new(Mix::Normal, Compose::SrcOver);
```
LaurenzV, line 555: I think we already defined a constant for this above?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -328,9 +504,179 @@ impl Scene {
...
+        if self.blend_mode != DEFAULT_BLEND {
+            return false;
+        }
+
+        // Condition 5: Geometry transform must not be zero-area or mirrored.
+        let geo_coeffs = self.transform.as_coeffs();
+        if geo_coeffs[0] * geo_coeffs[3] - geo_coeffs[1] * geo_coeffs[2] <= 0.0 {
```
LaurenzV, line 562: I think `transform.determinant` does the same.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -354,6 +700,14 @@ impl Scene {
         mask: Option<Mask>,
         filter: Option<Filter>,
     ) {
+        let blend_mode = blend_mode.unwrap_or(Self::DEFAULT_BLEND_MODE);
```
LaurenzV, line 703: COuld use `unwrap_or_default` here

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -354,6 +700,14 @@ impl Scene {
...
+        if self.render_hints.blit_rect_pipeline_enabled() {
+            assert!(
+                blend_mode == Self::DEFAULT_BLEND_MODE,
+                "blit rect pipeline only supports default blending"
+            );
+        }
```
LaurenzV, line 709: I think we have that same code above, maybe add a method like `self.assert_blend_mode`?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -427,11 +781,24 @@ impl Scene {
 
     /// Pop the last pushed layer.
     pub fn pop_layer(&mut self) {
-        self.wide.pop_layer(&mut self.render_graph);
+        self.enter_strip_mode();
+        let layer_bbox = self.wide.pop_layer(&mut self.render_graph);
+        if self.render_hints.blit_rect_pipeline_enabled() && !layer_bbox.is_inverted() {
+            // Push the dirty rect for the layer to the dirty rects list.
+            let [x0, y0, x1, y1] = layer_bbox.pixel_bounds();
```
LaurenzV: You probably are aware of this, but this won't represent the real layer bbox, but instead the one snapped to the wide tile coordinates.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -786,6 +1172,10 @@ impl Scene {
         range_index: usize,
         adjusted_strips: &[Strip],
     ) {
+        self.enter_strip_mode();
+        if self.render_hints.blit_rect_pipeline_enabled() {
+            self.push_dirty_viewport();
```
LaurenzV, line 1171: Is this because we don't care about recordings for now?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -865,3 +1255,377 @@ impl Scene {
...
+                blit_x1,
+                blit_y1,
+                u16::MAX - blit_x0,
+                u16::MAX - blit_y0,
+            ],
+            s,
+        );
+
+        // Process 2 rects (of 4 u16 values each) per iteration.
+        for chunk in data.chunks_exact(8) {
```
LaurenzV, line 1377: This does mean that each time we draw a new shape, the amount of time needed to perform that check for each new rectangle increases linearly, right? :( Maybe we should impose some limit and give up if it's too many?

LaurenzV (commented):
Left some comments. Overall, this does unfortunately add a lot of complexity, but if it gives us the performance improvements we need, it's probably worth it. One request though: While the microbenchmarks you made show some clear wins, I would still like to see how this fares when actually being used with our workloads. Have you already tried integrating this, and if so do you maybe have a screencast that shows the performance difference before/after with our workloads? Would be interesting to see, I think. 🙂

sparse_strips/vello_sparse_shaders/shaders/blit_rects.wgsl
taj-p: This is a good callout - thank you! I forgot about this. My opinion is that we want to leverage native sampling capabilities. I think that optimisation could also be extended to the strips shader. 

I think the "right" approach is to amend the atlas to draw 1px gutters around the images. Although we can't use the native sampler for bicubic sampling (WGPU/WebGL samplers don't support it) or images with EXTEND_REPEAT or EXTEND_REFLECT, I think we would see performance benefits from using the native bilinear sampler since it's hardware optimised.

I'm happy to amend the atlas to add this in this PR (while implementing the fast path in `render_strips.wgsl`). But, I think it might make sense to do it separately since this PR is already quite large. 

<details>

<summary>Proposed changes</summary>

- In `ImageCache::allocate`, allocate with `width+2` and `height+2`
- Store offset as `alloc.x + 1` and `alloc.y + 1`
- <Existing consumers work as is, without any modification>
- In `ImageCache::deallocate`, pass `width+2` and `height+2` to the stats tracking. Guillotere just takes the same `AtlasID` to deallocate.

---

For upload, we copy the texture into the atlas and then additional copies to populate the gutter with copies of the edges and corners.

</details>

LaurenzV: Yup, I also thought about this later and think this would be the right way! But yes, in a different PR.

taj-p:
superseded by some brilliant work by Laurenz

PR #1458, opened 2026-02-19
vello_hybrid: Add `render_to_atlas` and `write_to_atlas` APIs for glyph caching

grebmeg:
Adds public APIs to both the wgpu and webgl renderers for rendering `Scene` content directly into atlas layers and writing pixel data to pre-allocated atlas regions. This enables a glyph caching workflow where vector glyphs are rasterized into the atlas once and reused as cached images in subsequent passes.

The key additions are `render_to_atlas`, which runs the scheduler targeting a specific atlas layer (using a placeholder dummy texture to avoid read-write hazards), and `write_to_atlas`, extracted from `upload_image`, which writes pixel data to an already-allocated region.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -365,7 +365,8 @@ impl WebGlRenderer {
                 programs: &mut self.programs,
                 gl: &self.gl,
             };
-            ctx.render_strips(&self.fast_path_gpu_strips, 2, LoadOp::Clear);
+            let load_op = if clear { LoadOp::Clear } else { LoadOp::Load };
+            ctx.render_strips(&self.fast_path_gpu_strips, 2, load_op);
             Ok(())
         } else {
             let mut ctx = WebGlRendererContext {
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -145,7 +145,7 @@ impl Renderer {
         render_size: &RenderSize,
         view: &TextureView,
     ) -> Result<(), RenderError> {
-        self.render_scene(scene, device, queue, encoder, render_size, view)
+        self.render_scene(scene, device, queue, encoder, render_size, view, true)
     }
 
     /// Render a `scene` directly into an atlas layer.
@@ -226,6 +226,7 @@ impl Renderer {
             &mut encoder,
             &atlas_render_size,
             &layer_view,
+            false,
         );
 
         // Restore the real atlas bind group.
@@ -243,6 +244,9 @@ impl Renderer {
 
     /// Shared render pipeline: prepares GPU resources, runs the scheduler against
     /// the provided `view` at `render_size`, and maintains caches.
+    ///
+    /// When `clear` is true the render target is cleared to transparent black
+    /// before drawing (normal frame rendering).
     fn render_scene(
         &mut self,
         scene: &Scene,
@@ -251,6 +255,7 @@ impl Renderer {
         encoder: &mut CommandEncoder,
         render_size: &RenderSize,
         view: &TextureView,
+        clear: bool,
     ) -> Result<(), RenderError> {
         self.prepare_gpu_encoded_paints(&scene.encoded_paints);
         // TODO: For the time being, we upload the entire alpha buffer as one big chunk. As a future
@@ -282,7 +287,8 @@ impl Renderer {
                 encoder,
                 view,
             };
-            ctx.render_strips(&self.fast_path_gpu_strips, 2, LoadOp::Clear);
+            let load_op = if clear { LoadOp::Clear } else { LoadOp::Load };
+            ctx.render_strips(&self.fast_path_gpu_strips, 2, load_op);
 
             Ok(())
         } else {


LaurenzV:
Hmm... What I did is to change add an `OutputTarget` that either refers to the final view or an intermediate texture (https://github.com/linebender/vello/blob/b23b6341fdabc71041f1552448576e31178da050/sparse_strips/vello_hybrid/src/schedule.rs#L211-L230), and then the selection can happen at runtime instead of having to duplicate the whole `render` method, would that work for you? Otherwise, seems fine at a glance, but it might be worth extracting all of the setup stuff (e.g. for encoded paints) into a wrapper function that does setup/teardown and takes a closure for the "main action". What do you think?

grebmeg:
> Hmm... What I did is to change add an `OutputTarget` that either refers to the final view or an intermediate texture (
> 
> https://github.com/linebender/vello/blob/b23b6341fdabc71041f1552448576e31178da050/sparse_strips/vello_hybrid/src/schedule.rs#L211-L230
> 
> ), and then the selection can happen at runtime instead of having to duplicate the whole `render` method, would that work for you? Otherwise, seems fine at a glance, but it might be worth extracting all of the setup stuff (e.g. for encoded paints) into a wrapper function that does setup/teardown and takes a closure for the "main action". What do you think?

@laurenz-canva   `render_to_atlas` doesn't need to redirect strips inside the scheduler, the scheduler already sends strips to the same `idx == 2` target. The difference is what GPU object backs that target: in render it's the caller's `TextureView`, in `render_to_atlas` it's an atlas layer view you create. That choice happens outside the scheduler, in the `RendererContext` that holds the view reference. I don't think `OutputTarget` can help here because it operates one level too deep — it switches targets inside `flush()`, but those two methods already diverge before the scheduler even runs (atlas resize, bind group swap, encoder creation, immediate submit).

That’s at least what I discovered after looking into `OutputTarget`. I suggest taking another look at those methods once your PR is ready, I’ll have a better understanding of how they’re applied. For now, I’ve done a bit of deduplication you recommended.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -139,6 +139,108 @@ impl Renderer {
...
+                    usage: None,
+                });
+
+        // Swap in the stub atlas bind group to avoid the read-write conflict:
+        // the real atlas texture is used as the render target (COLOR_TARGET), so it
+        // cannot also be bound as a shader resource (TEXTURE_BINDING) in the same pass.
+        core::mem::swap(
+            &mut self.programs.resources.atlas_bind_group,
+            &mut self.programs.resources.stub_atlas_bind_group,
+        );
```
LaurenzV, line 216: Just wondering, how is it going to work for glyphs that use bitmaps? How will those be handled if you can't use the existing image atlas for uploading the glyphs themselves?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -139,6 +139,108 @@ impl Renderer {
...
+    ///
+    /// The scene should be sized to the atlas layer dimensions
+    /// ([`AtlasConfig::atlas_size`]), with content positioned at the allocated offset
+    /// coordinates from `ImageCache::allocate`.
+    ///
+    /// This method creates its own command encoder and submits immediately,
+    /// ensuring atlas content is committed before any subsequent
+    /// [`render`](Self::render) call (the two methods share GPU resources that
+    /// are staged by `queue.write_*` and only applied on the next `queue.submit`).
+    pub fn render_to_atlas(
```
LaurenzV, line 166: To me, it feels like this method might be easy to misuse for others, since it exposes a somewhat internal detail (atlas ID), which IIRC is usually not exposed to the user (most users will probably only ever use the `upload_image` method, which just returns an ImageId. Are users who want to have glyph caching expected to use this, or will there be some abstraction?

My point being, if there will be an abstraction, I would propose making this method `doc(hidden)` to avoid confusing users who just browse the documentation.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -139,6 +139,108 @@ impl Renderer {
...
+        );
+
+        let result = self.render_scene(
+            scene,
+            device,
+            queue,
+            &mut encoder,
+            &atlas_render_size,
+            &layer_view,
+        );
```
LaurenzV, line 230: I think there might be a problem, which I faced when implementing filters: Our current render strips pipeline expects the color attachment to be in the native pixel format, but the atlas texture array is always RGBA8. Therefore, if the system uses a different pixel format (for example BGRA on MacOS), this will crash, right?

The solution I chose is to create two render strips pipelines, one for the native pixel format and one for always RGBA8, and use the latter when rendering into the image atlas array.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -190,6 +289,28 @@ impl Renderer {
...
+    /// allocator-assigned position. Pass `None` to use the default atlas offset.
+    pub fn write_to_atlas<T: AtlasWriter>(
+        &mut self,
+        device: &Device,
+        queue: &Queue,
+        encoder: &mut CommandEncoder,
+        image_id: vello_common::paint::ImageId,
+        writer: &T,
+        offset_override: Option<[u32; 2]>,
+    ) {
```
LaurenzV, line 350: Same here, can we make this hidden? I wouldn't expect a user to know how to correctly pass an image offset override here.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV, line 1: Same two comments as for the wgpu backend, regarding visibility.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Bitmaps would be uploaded into a single atlas alongside outlines and COLR glyphs. To support that, all glyph content should be baked into the atlas first, and the glyph cache needs to know which atlas (and region) to use when rendering a given glyph. That mapping comes from `image_cache`, which is essentially a rect allocator. For glyph rendering, we use `OpaqueId`s and render glyphs the same way we render images. This is how it works.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: I agree — there’s room for a better abstraction here, but I haven’t landed on the right one yet. I moslty focused on getting `parley_draw` implemented and working as expected. After we discuss glyph cache PR (planning to create tomorrow), I expect we’ll have clearer ideas and can revisit this to improve the approach.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Yes, I think so, and that’s why we pass `Rgba8Unorm` into the render config explicitly. But I don’t think this issue is new or introduced by the changes here; it seems like something we already had. The better approach is probably to adopt your fix once your PR lands, rather than implementing a parallel solution now.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Made it hidden

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Fixed

sparse_strips/vello_hybrid/src/render/wgpu.rs
LaurenzV: > it seems like something we already had. 

Is it? I'm not sure, because this is the first PR that introduces the ability to draw onto the image atlas texture array, no? Up until now, we've only drawn into the final surface or the slot textures, which always use the native format IIRC. I guess it's fine to ignore for now, but maybe add a TODO?

sparse_strips/vello_hybrid/src/render/wgpu.rs
LaurenzV: Makes sense, it doesn't have to be perfect, but in this case it still might be worth to hide it from the documentation until we've figured out something? So that if we make a new release in the meantime, other people won't be confused by this. But anyway, it's just a small suggestion, feel free to ignore if you don't think it's necessary!

LaurenzV: Ah nevermind, saw you already did that!

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -255,6 +237,122 @@ impl WebGlRenderer {
...
+        self.programs
+            .maybe_resize_atlas_texture_array(&self.gl, self.image_cache.atlas_count() as u32);
+
+        let atlas_size = self.programs.resources.atlas_texture_array.size();
+        let atlas_render_size = RenderSize {
+            width: atlas_size.width,
+            height: atlas_size.height,
+        };
+
+        let atlas_framebuffer = self.gl.create_framebuffer().unwrap();
```
taj-p: I'm not sure if we should be re-creating the FBO for each call to `render_to_atlas`. IIUC, an FBO is just a container of attachments, but repeated creation and destruction can see unnecessary CPU-side driver overhead. What do you think of caching the atlas frame buffer object?

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: I think you’re right. I’ll leave the TODO for now and come back to it later. It might be automatically covered once you introduce the filter effects functionality, but if there are any integration issues, I’ll take care of fixing them if you want.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Fixed now

taj-p (approved):
LGTM! Just skimmed the code

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -175,7 +287,8 @@ impl Renderer {
                 encoder,
                 view,
             };
-            ctx.render_strips(&self.fast_path_gpu_strips, 2, LoadOp::Clear);
+            let load_op = if clear { LoadOp::Clear } else { LoadOp::Load };
+            ctx.render_strips(&self.fast_path_gpu_strips, 2, load_op);
```
taj-p, line 291: cc @grebmeg - I'm not sure the `clear` member is honoured in the else branch of this block below on line 302

PR #1459, opened 2026-02-19
vello_hybrid: Add inline Pixmap rendering

grebmeg:
This PR adds support for rendering `Arc<Pixmap>` images directly via `ImageSource::Pixmap` in `vello_hybrid`, without requiring callers to manually pre-upload them. The renderer automatically deduplicates pixmaps by pointer identity across frames and lazily uploads them to the GPU atlas on demand, with periodic eviction of stale entries.

### DIFF
--- a/sparse_strips/vello_common/src/image_cache.rs
+++ b/sparse_strips/vello_common/src/image_cache.rs
@@ -8,7 +8,10 @@
 
 use crate::multi_atlas::{AllocId, AtlasConfig, AtlasError, AtlasId, MultiAtlasManager};
 use crate::paint::ImageId;
+use alloc::sync::Arc;
 use alloc::vec::Vec;
+use hashbrown::HashMap;
+use crate::pixmap::Pixmap;
 
 /// Represents an image resource for rendering.
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -480,13 +494,34 @@ impl WebGlRenderer {
...
+                                        self.pending_uploads.push(PendingImageUpload {
+                                            image_id: id,
+                                            pixmap: pixmap.clone(),
+                                        });
+                                        Some(id)
+                                    }
+                                    Err(e) => {
+                                        log::warn!("Failed to allocate pixmap in atlas: {e:?}");
+                                        None
+                                    }
```
taj-p: Should this be a `RenderError`?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -480,13 +494,34 @@ impl WebGlRenderer {
...
-                            let gpu_image = self.encode_image_paint(img, image_resource);
-                            self.encoded_paints[encoded_paint_idx] = gpu_image;
-                            current_idx += GPU_ENCODED_IMAGE_SIZE_TEXELS;
-                        }
+                    let image_id = match &img.source {
+                        ImageSource::OpaqueId(id) => Some(*id),
+                        ImageSource::Pixmap(pixmap) => match self.pixmap_register.get(pixmap) {
+                            Some(id) => Some(id),
+                            None => {
+                                match self.image_cache.allocate(pixmap.width(), pixmap.height()) {
```
taj-p: How are these sorts of pixmaps de-allocated from the texture?

taj-p: Ohhh, I see via `maintain()`.

sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -562,3 +562,36 @@ fn image_with_multiple_clip_layers(ctx: &mut impl Renderer) {
     ctx.pop_layer();
     ctx.pop_layer();
 }
+
+/// Test that `ImageSource::Pixmap` renders directly without Pixmap pre-uploading.
+///
+/// This exercises the hybrid renderer's `PixmapRegister` + `PendingImageUpload` path,
+/// where inline pixmaps are automatically allocated in the image cache and uploaded
+/// to the GPU atlas on demand. The same `Arc<Pixmap>` is used for two fills in
+/// different locations, verifying that deduplication works (only one atlas upload).
```
taj-p: "verifying that deduplication works" - is that tested here?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -442,13 +456,34 @@ impl Renderer {
...
+                                        Some(id)
+                                    }
+                                    Err(e) => {
+                                        log::warn!("Failed to allocate pixmap in atlas: {e:?}");
+                                        None
+                                    }
+                                }
+                            }
+                        },
+                    };
```
taj-p: Is there a way to do this without doing 2 lookups on the hash map?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -514,6 +549,17 @@ impl WebGlRenderer {
         self.paint_idxs[encoded_paints.len()] = current_idx;
     }
 
+    /// Upload any pixmaps that were allocated during [`Self::prepare_gpu_encoded_paints`]
+    /// but haven't been written to the GPU atlas yet.
+    fn upload_pending_images(&mut self) {
+        let uploads: Vec<_> = self.pending_uploads.drain(..).collect();
```
taj-p: Do we need to collect here or can we drain without collecting?
I.e, does this work?
```
        let mut uploads = core::mem::take(&mut self.pending_uploads);
        for upload in uploads.drain(..) {
            self.write_to_atlas(
                device,
                queue,
                encoder,
                upload.image_id,
                &upload.pixmap,
                None,
            );
        }
        self.pending_uploads = uploads;
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -514,6 +549,17 @@ impl WebGlRenderer {
...
+        for upload in uploads {
+            self.programs
+                .maybe_resize_atlas_texture_array(&self.gl, self.image_cache.atlas_count() as u32);
```
```suggestion
```

I think this is called in `write_to_atlas`

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -476,6 +511,34 @@ impl Renderer {
...
+    ) {
+        let uploads: Vec<_> = self.pending_uploads.drain(..).collect();
+        for upload in uploads {
+            Programs::maybe_resize_atlas_texture_array(
+                device,
+                encoder,
+                &mut self.programs.resources,
+                &self.programs.atlas_bind_group_layout,
+                self.image_cache.atlas_count() as u32,
+            );
```
taj-p:
```suggestion
```

I think this is called in `write_to_atlas`

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -476,6 +511,34 @@ impl Renderer {
...
+            self.write_to_atlas(
+                device,
+                queue,
+                encoder,
+                upload.image_id,
+                &upload.pixmap,
+                None,
+            );
+        }
```
taj-p:
```suggestion
        let mut uploads = core::mem::take(&mut self.pending_uploads);
        for upload in uploads.drain(..) {
            self.write_to_atlas(
                device,
                queue,
                encoder,
                upload.image_id,
                &upload.pixmap,
                None,
            );
        }
        self.pending_uploads = uploads;
```

sparse_strips/vello_common/src/image_cache.rs
taj-p, line 1: For what types of use cases would we use this over asking the caller to manage the lifetime of pixmaps? 

I'm a little wary of using this API because it means the caller needs to keep the pixmap in memory - is that right? The caller can't deallocate the pixmap if they pass it as an `ImageSource::Pixmap` because the next time they use it, they need to pass the `Arc<Pixmap>`? Or is there a certain use case for glyph caching that this is used?

taj-p (commented):
Code looks good! Just asking some questions for my understanding

LaurenzV:
Echoing Taj's comment, I would like to understand the motivation further before looking at the code more deeply.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Good call! Changed `prepare_gpu_encoded_paints` to return `Result<(), RenderError>` and propagate the `AtlasError` via `?`. The existing `#[from] AtlasError` on `RenderError` makes this clean.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Correct, `PixmapRegister::maintain()` is called at the end of each `render()` / `render_to_atlas()` call. It evicts entries that haven't been referenced for `max_entry_age` frames (default 64) and deallocates their atlas space.

sparse_strips/vello_sparse_tests/tests/image.rs
grebmeg: Good catch! The test exercises the dedup code path (same `Arc<Pixmap>` used twice) but doesn't assert on upload count. Added a dedicated unit test for `PixmapRegister` in `image_cache.rs` that verifies: first lookup misses, insert + second lookup hits, `Arc::clone` hits (same pointer), and a different `Arc` with identical content misses (pointer identity).

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Good call! Added `PixmapRegister::get_or_insert_with()` using the `entry()` API so we do a single lookup.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Switched to `mem::take` to preserve the allocation across frames. Thank you!

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Yep, thank you!

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Yep, thank you!

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Fixed

sparse_strips/vello_common/src/image_cache.rs
grebmeg: > Echoing Taj's comment, I would like to understand the motivation further before looking at the code more deeply.

I added this code with the bitmap glyphs case in mind, specifically when caching is disabled, so that instead of panicking now we can at least render something. Manually uploading all glyphs is also a tedious step, and ideally that should be hidden if possible especially for glyphs.

That said, it now seems like we’ve reached a complete model here. I do agree with your concern though, the caller would still need to keep pixmaps in memory, similar to the CPU case when the glyph cache is disabled. But this current implementation can be considered the least recommended option and a starting point, so that we can evolve it further if needed.

taj-p (changes requested):
In office hours, we discussed the direction of Vello Hybrid and textures. It seems like the direction is towards consumers owning textures, atlasing, and lifetimes of images in those atlases (moving ownership out of hybrid and to the caller).

At the same time, the direction for `glifo` seems to be further decoupling from renderers.

When you put these 2 together, I think this PR, which evolves the image cache and hybrid to own more of texture atlasing and lifetimes, conflicts with that direction.

LaurenzV:
Does that mean the user would be responsible for manually uploading and destroying the bitmaps for bitmap glyphs? I think that would be a bit unfortunate.

taj-p:
> Does that mean the user would be responsible for manually uploading and destroying the bitmaps for bitmap glyphs? I think that would be a bit unfortunate.

No. It would be glifo's responsibility

PR #1460, opened 2026-02-20
spare_strips: Add image tinting support

grebmeg:
This PR introduces the ability to tint image paints at rasterization time. A new `Tint` type carries a color and a `TintMode` — either `AlphaMask`, which treats the source image's alpha as a coverage mask and fills with the tint color (useful for glyph rendering and monochrome icons), or `Multiply`, which does a per-channel multiply of the source pixels by the tint color (useful for colorizing full-color images).

The tint is threaded through the encoding pipeline so that `EncodeExt::encode_into` accepts an optional `Tint`, and the `Scene` / `Recorder` APIs expose `set_tint` for controlling it. On the CPU side, tinting is applied in both the nearest-neighbor and filtered image painters, working in the appropriate precision for each path. On the GPU side, the WGSL shader unpacks a packed `RGBA8` tint color from the encoded paint data and applies the selected mode before blending.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/highp/mod.rs
+++ b/sparse_strips/vello_cpu/src/fine/highp/mod.rs
@@ -170,22 +170,20 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
 
         simd.vectorize(
             #[inline(always)]
-            || {
-                match tint.mode {
-                    TintMode::AlphaMask => {
-                        for chunk in dest.chunks_exact_mut(16) {
-                            let pixel = f32x16::from_slice(simd, chunk);
-                            let alphas = pixel.splat_4th();
-                            let tinted = tint_v * alphas;
-                            chunk.copy_from_slice(tinted.as_slice());
-                        }
+            || match tint.mode {
+                TintMode::AlphaMask => {
+                    for chunk in dest.chunks_exact_mut(16) {
+                        let pixel = f32x16::from_slice(simd, chunk);
+                        let alphas = pixel.splat_4th();
+                        let tinted = tint_v * alphas;
+                        chunk.copy_from_slice(tinted.as_slice());
                     }
-                    TintMode::Multiply => {
-                        for chunk in dest.chunks_exact_mut(16) {
-                            let pixel = f32x16::from_slice(simd, chunk);
-                            let tinted = pixel * tint_v;
-                            chunk.copy_from_slice(tinted.as_slice());
-                        }
+                }
+                TintMode::Multiply => {
+                    for chunk in dest.chunks_exact_mut(16) {
+                        let pixel = f32x16::from_slice(simd, chunk);
+                        let tinted = pixel * tint_v;
+                        chunk.copy_from_slice(tinted.as_slice());
                     }
                 }
             },


LaurenzV:
This does unsurprisingly cause regressions for normal images. :( But I also wouldn't know how to avoid it, so if we really need that then it is what it is.

```
fine/image/transform/none_u8_neon
                        time:   [279.76 ns 282.04 ns 284.63 ns]
                        change: [+6.0489% +7.3013% +8.6111%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (4.00%) high mild
  1 (1.00%) high severe

fine/image/transform/scale_u8_neon
                        time:   [280.17 ns 282.00 ns 284.08 ns]
                        change: [+5.6126% +6.9340% +8.5438%] (p = 0.00 < 0.05)
                        Performance has regressed.
  7 (7.00%) high mild
  2 (2.00%) high severe

fine/image/transform/rotate_u8_neon
                        time:   [428.67 ns 430.09 ns 431.49 ns]
                        change: [+20.111% +20.727% +21.340%] (p = 0.00 < 0.05)
                        Performance has regressed.
  6 (6.00%) high mild
  1 (1.00%) high severe

fine/image/quality/low_u8_neon
                        time:   [281.17 ns 283.51 ns 286.10 ns]
                        change: [+3.5725% +4.7283% +5.9705%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

                        time:   [1.3702 µs 1.3747 µs 1.3794 µs]
                        change: [+2.2158% +2.8353% +3.3934%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild

fine/image/quality/high_u8_neon
                        time:   [8.1643 µs 8.1825 µs 8.2008 µs]
                        change: [+2.4438% +2.8503% +3.2690%] (p = 0.00 < 0.05)
                        Performance has regressed.
  6 (6.00%) high mild
  3 (3.00%) high severe

fine/image/extend/pad_u8_neon
                        time:   [289.64 ns 290.53 ns 291.49 ns]
                        change: [+4.6107% +6.1515% +7.8261%] (p = 0.00 < 0.05)
                        Performance has regressed.
  6 (6.00%) high mild
  1 (1.00%) high severe

fine/image/extend/repeat_u8_neon
                        time:   [340.92 ns 342.20 ns 343.56 ns]
                        change: [+18.855% +19.406% +19.898%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

fine/image/extend/reflect_u8_neon
                        time:   [562.56 ns 563.28 ns 564.19 ns]
                        change: [+11.354% +11.655% +11.935%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) high mild
  3 (3.00%) high severe

images/overlapping      time:   [2.5914 ms 2.5936 ms 2.5961 ms]
                        change: [+1.0254% +1.2325% +1.4240%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe
```

LaurenzV:
Could you rebase onto main? Because I'm still seeing the changes from the image resolver.

grebmeg:
> Could you rebase onto main? Because I'm still seeing the changes from the image resolver.

@LaurenzV Sure, just give me some time, I need to double check my fix with performance.

grebmeg:
@LaurenzV I updated everything and refactored the tinting so that it now runs as a post pass. It should be equivalent to the previous solution, but without affecting the main hot image path. It seems to have fixed the regression.

However, I noticed that bench now enables `sse4_2`, and that is the only case that regressed. I have not looked into the details yet to understand why. Maybe you already know what could be causing it?

LaurenzV:
I don't have an x86 machine so I can't test. How much is it regressing? Seems weird though that there are regressions only for that specific level. 🤔

sparse_strips/vello_cpu/src/fine/highp/mod.rs
```diff
@@ -162,6 +162,30 @@ impl<S: Simd> FineKernel<S> for F32Kernel {
...
+                    let pixel = f32x16::from_slice(simd, chunk);
+                    let tinted = match tint.mode {
+                        TintMode::AlphaMask => {
+                            let alphas = pixel.splat_4th();
+                            tint_v * alphas
+                        }
+                        TintMode::Multiply => pixel * tint_v,
+                    };
+                    chunk.copy_from_slice(tinted.as_slice());
+                }
```
LaurenzV: I haven't tested it but it might be worth creating two separate for loops for each tint mode, to avoid the branching in the loop.

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
```diff
@@ -195,6 +195,32 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
...
+                for chunk in dest.chunks_exact_mut(16) {
+                    let pixel = u8x16::from_slice(simd, chunk);
+                    let pixel_f = u8_to_f32(pixel);
+                    let tinted = match tint.mode {
+                        TintMode::AlphaMask => {
+                            let alphas = pixel_f.splat_4th();
+                            tint_v * alphas
+                        }
+                        TintMode::Multiply => pixel_f * tint_v,
+                    };
```
LaurenzV: Same comment as for `highp`.

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
```diff
@@ -195,6 +195,32 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
         painter.paint_u8(dest);
     }
 
+    #[inline(always)]
+    fn apply_tint(simd: S, dest: &mut [Self::Numeric], tint: &Tint) {
+        let premul = tint.color.premultiply();
+        let [r, g, b, a] = premul.components;
+        let tint_v = f32x16::block_splat(f32x4::from_slice(simd, &[r, g, b, a]));
```
LaurenzV: We could probably use `u8x32` for `lowp` and use the normalized multiplication, no? Then we don't need to convert to `f32`.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -335,38 +335,43 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
                 );
-                final_color = alpha * sample_color;
             }
+
+            let is_multiply = f32(encoded_image.tint_mode);
+            final_color = alpha * mix(
+                encoded_image.tint * sample_color.a,
+                sample_color * encoded_image.tint,
+                is_multiply
+            );
```
LaurenzV: Can we not use `select` here? Would make the intent clearer, I think.

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
grebmeg: Fixed.

sparse_strips/vello_cpu/src/fine/highp/mod.rs
grebmeg: I think this should be predicted very well, since if `tint.mode` is false, it remains false for all other iterations over the image. That said, I don’t mind creating two separate loops. No regression detected if we use that approach.

Fixed.

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
grebmeg: Nice! That makes sense!

Fixed.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Sure, let's use `select` here

grebmeg:
> I don't have an x86 machine so I can't test. How much is it regressing? Seems weird though that there are regressions only for that specific level. 🤔

The regression is approximately 14%.

PR #1461, opened 2026-02-20
vello_hybrid: Optimize rendering of opaque full-tile images

grebmeg:
Skip prior draw commands when a wide tile is fully covered by an opaque fill (opaque image). This avoids unnecessary compositing work for tiles that would be entirely overwritten.

I also added additional examples that allocate more and more images when pressing “a”, so I could run a very rough benchmark. It shows roughly a 1.65x improvement in the image randomized positioning scenario.

I am not entirely happy with the approach of carrying `may_have_opacities` inside `ImageSource`, but I have not come up with a better solution yet. I would really welcome any ideas if you have suggestions.

before|after
---|---
<video src=https://github.com/user-attachments/assets/6207df28-032f-4fd0-8223-d47b6e4eeb38/>|<video src=https://github.com/user-attachments/assets/85c47dda-e493-4a59-92f0-28ccf3212302/>

### DIFF
--- a/sparse_strips/vello_common/src/paint.rs
+++ b/sparse_strips/vello_common/src/paint.rs
@@ -75,7 +75,7 @@ pub enum ImageSource {
     Pixmap(Arc<Pixmap>),
     /// Pixmap pixels were registered earlier; this is just a handle.
     ///
-    /// The `bool` indicates whether the image may contain non-opaque pixels.
+    /// The `bool` indicates whether the image may contain non-opaque pixels. If unsure, set this to `true`.
     OpaqueId(ImageId, bool),
 }
 
--- a/sparse_strips/vello_common/src/paint.rs
+++ b/sparse_strips/vello_common/src/paint.rs
@@ -74,7 +74,9 @@ pub enum ImageSource {
     /// Pixmap pixels travel with the scene packet.
     Pixmap(Arc<Pixmap>),
     /// Pixmap pixels were registered earlier; this is just a handle.
-    OpaqueId(ImageId),
+    ///
+    /// The `bool` indicates whether the image may contain non-opaque pixels.
+    OpaqueId(ImageId, bool),
 }
 
 impl ImageSource {
--- a/sparse_strips/vello_common/src/paint.rs
+++ b/sparse_strips/vello_common/src/paint.rs
@@ -74,12 +74,45 @@ pub enum ImageSource {
     /// Pixmap pixels travel with the scene packet.
     Pixmap(Arc<Pixmap>),
     /// Pixmap pixels were registered earlier; this is just a handle.
-    ///
-    /// The `bool` indicates whether the image may contain non-opaque pixels. If unsure, set this to `true`.
-    OpaqueId(ImageId, bool),
+    OpaqueId {
+        /// The image handle.
+        id: ImageId,
+        /// Whether the image may contain non-opaque pixels.
+        may_have_opacities: bool,
+    },
 }
 
 impl ImageSource {
+    /// Create an [`ImageSource`] from a pre-registered image handle.
+    ///
+    /// Conservatively assumes the image may have non-opaque pixels.
+    /// Use [`Self::opaque_id_with_opacity_hint`] when you know the image is fully opaque.
+    pub fn opaque_id(id: ImageId) -> Self {
+        Self::OpaqueId {
+            id,
+            may_have_opacities: true,
+        }
+    }
+
+    /// Create an [`ImageSource`] from a pre-registered image handle,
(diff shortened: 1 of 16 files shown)


sparse_strips/vello_common/src/paint.rs
LaurenzV, line 1: Maybe we rename this to `Custom` instead of `OpaqueId`? Maybe it should also be a typed struct?

sparse_strips/vello_common/src/paint.rs
```diff
@@ -74,7 +74,9 @@ pub enum ImageSource {
     /// Pixmap pixels travel with the scene packet.
     Pixmap(Arc<Pixmap>),
     /// Pixmap pixels were registered earlier; this is just a handle.
-    OpaqueId(ImageId),
+    ///
+    /// The `bool` indicates whether the image may contain non-opaque pixels.
```
LaurenzV:
```suggestion
    /// The `bool` indicates whether the image may contain non-opaque pixels. If unsure, set this to `true`.
```

sparse_strips/vello_example_scenes/src/multi_image.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+}
+
+impl ExampleScene for MultiImageScene {
+    fn render(&mut self, ctx: &mut impl RenderingContext, root_transform: Affine) {
+        let vw = ctx.width() as f64;
+        let vh = ctx.height() as f64;
+        let [a, b, c, d, e, f] = root_transform.as_coeffs();
+        let snapped_root = Affine::new([a, b, c, d, e.round(), f.round()]);
```
LaurenzV: Can be replaced with `transform.with_translation()`

LaurenzV (approved):
.

sparse_strips/vello_common/src/paint.rs
LaurenzV: (second one is up to you though, but the first one would be nice I think)

grebmeg: Sorry, I’m not exactly sure what you’re referring to in this comment on the file. Are you suggesting renaming `OpaqueId` to `Custom`? If so, I’m not sure what the benefit would be. Could you please clarify?

sparse_strips/vello_common/src/paint.rs
grebmeg: Fixed

sparse_strips/vello_common/src/paint.rs
LaurenzV: Yeah, that would have been my suggestion. To me, calling it OpaqueId makes sense when we only store the actual ID, but now that we also store a boolean there it seems a bit less clear to me. Hence, why I thought the enum variant should be called `Custom` (as in a custom image, as opposed to a Pixmap) so that it's not just tied to the image ID (since we now store additional metadata). But perhaps this is not necessary.

LaurenzV (approved):
Will leave it up to you whether you want to rename or not.

sparse_strips/vello_common/src/paint.rs
grebmeg: Thanks for clarifying, I understand your thinking now. I agree that the current approach is a bit ugly, but I think it’s most pragmatic.

I have a feeling we may need to introduce another abstraction that holds some `ImageSource` metadata, since both `Pixmap`s and external images may or may not have opacity. However, it’s not that simple because there’s an inherent asymmetry, one variant has pixels and the other doesn’t.

So maybe we should leave this for a future refactoring step, when a better idea comes along.

LaurenzV: Sure!

sparse_strips/vello_common/src/paint.rs
```diff
@@ -74,10 +74,45 @@ pub enum ImageSource {
     /// Pixmap pixels travel with the scene packet.
     Pixmap(Arc<Pixmap>),
     /// Pixmap pixels were registered earlier; this is just a handle.
-    OpaqueId(ImageId),
+    OpaqueId {
+        /// The image handle.
+        id: ImageId,
+        /// Whether the image may contain non-opaque pixels.
+        may_have_opacities: bool,
```
upsuper, line 81: I don't quite agree with the naming here. Anything that isn't fully transparent would "have opacity", and fully opaque content has the full opacity, as the definition of opacity is:
> The quality or state of being opaque.

If the purpose here is to check whether there is any non-fully-opaque pixel, "may have transparency" should be semantically more accurate.

LaurenzV: Makes sense!

PR #1462, opened 2026-02-20
sparse strips: Bump to `fearless_simd` v0.4.0

tomcur:
This brings a nice little boost to flattening, but badly regresses analytic AA in strip rendering.

I've traced the regression back to `max`/`max_precise` semantics introduced here: https://github.com/linebender/fearless_simd/pull/136.

I'll open a PR based on top of this one to fix the issue.

Benchmarks:

Flattening

```
flatten/Ghostscript_Tiger
                        time:   [180.70 µs 181.08 µs 181.46 µs]
                        change: [-6.4035% -6.0571% -5.6921%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  1 (1.00%) high severe
flatten/paris-30k       time:   [9.2019 ms 9.2259 ms 9.2511 ms]
                        change: [-2.0521% -1.5976% -1.1616%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
```

Strip generation

```
                        time:   [244.27 µs 244.83 µs 245.40 µs]
                        change: [+25.354% +25.929% +26.391%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (2.00%) high severe
render_strips/paris-30k_simd
                        time:   [26.936 ms 27.010 ms 27.093 ms]
                        change: [+14.664% +15.809% +16.826%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (6.00%) high mild
```

### DIFF
--- a/sparse_strips/vello_sparse_tests/Cargo.toml
+++ b/sparse_strips/vello_sparse_tests/Cargo.toml
@@ -19,7 +19,9 @@ vello_api = { workspace = true }
 vello_common = { workspace = true, features = ["std"] }
 vello_cpu = { workspace = true, features = ["multithreading", "std", "f32_pipeline"] }
 vello_hybrid = { workspace = true }
-fearless_simd = { workspace = true, features = ["force_support_fallback"] } # We require the `force_support_callback` feature to test the fallback path.
+fearless_simd = { workspace = true, features = [
+    "force_support_fallback", # We require the `force_support_callback` feature to test the fallback path.
+] }
 wgpu = { workspace = true, default-features = true }
 pollster = { workspace = true }
 vello_dev_macros = { workspace = true }
--- a/sparse_strips/vello_sparse_tests/tests/util.rs
+++ b/sparse_strips/vello_sparse_tests/tests/util.rs
@@ -123,7 +123,7 @@ pub(crate) fn get_ctx<T: Renderer>(
         #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
         "sse42" => {
             if std::arch::is_x86_feature_detected!("sse4.2") {
-                Level::Sse4_2(unsafe { vello_common::fearless_simd::Sse4_2::new_unchecked() })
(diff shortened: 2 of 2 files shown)


LaurenzV:
Have you already tested whether that fix works?

tomcur:
> Have you already tested whether that fix works?

Yep! Some more info here: [#simd > fearless_simd v0.4 @ 💬](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/fearless_simd.20v0.2E4/near/575031364).

sparse_strips/vello_dev_macros/src/test.rs
LaurenzV, line 1: For testing, since we want to explicitly test the fallback, I think the safest option would be to enable the `force_support_fallback` feature and then use fallback, like before. This way, we can be sure that it will always use that.

LaurenzV: (This is also why you are getting the CI failures in WASM)

tomcur: `vello_sparse_tests` now requires that feature, and I've reverted the tests that explicitly requested `fallback` before to do that again.

tomcur: For `vello_sparse_tests` that is, maybe we also want that for tests in `vello_cpu` and `vello_hybrid`? Not sure. They currently request `baseline`.

LaurenzV: I think it's fine for to leave as is for those. The visreg tests are the ones where it actually matters.

tomcur: `vello_cpu` seems to mostly have been constructing `Level::try_detect().unwrap_or(Level::fallback())` in its tests. `vello_hybrid` had a lot of explicit `Level::fallback()`.

PR #1464, opened 2026-02-20
 vello_common: Part Two of the strip rendering regression fix (make use of `_mm_max_ps` semantics)

tomcur:
On top of Part One: https://github.com/linebender/vello/pull/1463.

Copying from the code comment:

> We know `ymin` and `ymax` are finite. We require the `max` operation to pick `ymin` if its first operand is NaN. On x86, that's precisely the semantics of `_mm_max_ps`, which `f32x4::max` emits. For AArch64, we do require the `f32x4::max_precise` semantics (as `max` returns NaN if either operand is NaN); however, the precise version should be comparatively less expensive than on x86. For `min`, we then know both operands are finite, so we can unambigously use the relaxed version.

This may be somewhat controversial, as it `cfg`s behavior based on the target architecture, and makes use of undocumented `fearless_simd` behavior. The good part is that if this ever breaks, the tests will fail loudly (because NaNs happen a lot here!).

Benches as follows relative to `main` before bumping `fearless_simd` on my x86 machine (i7-13700k).

```
                        time:   [188.69 µs 189.12 µs 189.63 µs]
                        change: [-3.0455% -2.5888% -2.2385%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 50 measurements (14.00%)
  7 (14.00%) high mild
render_strips/paris-30k_simd
                        time:   [23.308 ms 23.373 ms 23.449 ms]
                        change: [-0.7675% +0.2120% +1.0973%] (p = 0.67 > 0.05)
  2 (4.00%) high mild
  1 (2.

### DIFF
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -26,6 +26,7 @@ This release has an [MSRV][] of 1.88.
 ### Fixed
 
 - Fixed rendering of blurred rounded rectangles with zero or very small blur standard deviations. ([#1422][] by [@tomcur][])
+- Fixed analytic AA performance regressions from `fearless_simd`'s tightened `max_precise` and `min_precise` semantics in v0.4.0. ([#1463][] and [#1464][] by [@tomcur][])
 
 ## [0.0.6][] - 2026-01-15
 
@@ -130,6 +131,8 @@ See also the [vello_cpu 0.0.1](../vello_cpu/CHANGELOG.md#001---2025-05-10) relea
 [#1354]: https://github.com/linebender/vello/pull/1354
 [#1399]: https://github.com/linebender/vello/pull/1399
 [#1422]: https://github.com/linebender/vello/pull/1422
+[#1463]: https://github.com/linebender/vello/pull/1463
+[#1464]: https://github.com/linebender/vello/pull/1464
 
 [Unreleased]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.6...HEAD
 [0.0.6]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.5...sparse-strips-v0.0.6

### CODE REVIEW

File: sparse_strips/vello_common/src/strip.rs [hunk 1]
```diff
@@ -398,17 +398,36 @@ fn render_impl<S: Simd>(
...
-            // non-NaN operand (regardless of whether it's the first or second). For `min`, we then
-            // know both operands are finite, so we can use the relaxed version.
+            // if its first operand is NaN. On x86, that's precisely the semantics of `_mm_max_ps`,
+            // which `f32x4::max` emits. For AArch64, we do require the `f32x4::max_precise`
+            // semantics (as `max` returns NaN if either operand is NaN); however, the precise
+            // version should be comparatively less expensive than on x86. For `min`, we then know
+            // both operands are finite, so we can unambigously use the relaxed version. If this
+            // ever breaks, tests should fail loudly, because NaNs happen a lot here!
+            trait F32x4MaxExt {
+                fn max_if_nan_take_second(self, rhs: Self) -> Self;
```
LaurenzV: Should it maybe be `max_if_one_nan_take_the_other` to make the intention more clear? Up to you 😄

LaurenzV (approved):
I think this is reasonable, especially because we control fearless_simd (so if it ever changes there, we will be aware), and the tests do cover this. :)

sparse_strips/vello_common/src/strip.rs
tomcur: Agree! I've named it `max_if_first_nan_take_second` as those are the semantics we need. The name `max_if_one_nan_take_the_other` doesn't quite work, because those are not the semantics that `_mm_max_ps` has: its semantics could be named `max_if_either_nan_take_second`.

PR #1467, opened 2026-02-23
Render more than ~5300 images 

taj-p:
Fixes https://github.com/linebender/vello/issues/1468

There is a bug in the shader code - we only check the first row of the encoded paints texture. On my machine, that gives me ~5400 images before I start drawing empty rectangles.

 This PR modifies our `textureLoad` calls to consider more than the first row of texels.

On my machine, I used the following test to verify the fix.

```rs
#[vello_test(width = 10, height = 10)]
fn image_many_unique_paints_wrap_encoded_texture(ctx: &mut impl Renderer) {
    const NUM_IMAGES: usize = 6000;
    let sampler = ImageSampler {
        x_extend: Extend::Pad,
        y_extend: Extend::Pad,
        quality: ImageQuality::Low,
        alpha: 1.0,
    };

    let dummy_rect = Rect::new(0.0, 0.0, 1.0, 1.0);
    for i in 0..NUM_IMAGES {
        let mut pix = Pixmap::new(1, 1);
        let val = (i % 255 + 1) as u8;
        pix.set_pixel(0, 0, PremulRgba8::from_u32(u32::from_be_bytes([val, val, val, 255])));
        let source = ctx.get_image_source(Arc::new(pix));
        ctx.set_paint(Image {
            image: source,
            sampler,
        });
        ctx.fill_rect(&dummy_rect);
    }

    let mut final_pix = Pixmap::new(1, 1);
    final_pix.set_pixel(0, 0, PremulRgba8::from_u32(u32::from_be_bytes([255, 0, 0, 255])));
    let final_source = ctx.get_image_source(Arc::new(final_pix));
    ctx.set_paint(Image {
        image: final_source,
        sampler,
    });
    ctx.fill_rect(&Rect::new(0.0, 0.0, 10.0, 10.0));
}
```

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -43,6 +43,8 @@ pub struct Config {
     /// Number of trailing zeros in the encoded paints texture width (log2 of width).
     /// Pre-calculated on CPU since downlevel targets do not support `firstTrailingBit`.
     pub encoded_paints_tex_width_bits: u32,
+    /// Padding to satisfy WebGL's 16-byte alignment requirement for uniform buffers.
+    pub _padding: [u32; 3],
 }
 
 /// Represents a GPU strip for rendering.
--- a/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
@@ -113,6 +113,10 @@ struct Config {
     // Number of trailing zeros in encoded_paints_tex_width (log2 of width).
     // Pre-calculated on CPU since WebGL2 doesn't support `firstTrailingBit`.
     encoded_paints_tex_width_bits: u32,
+    // Padding to satisfy WebGL's 16-byte alignment requirement for uniform buffers.
+    _padding0: u32,
+    _padding1: u32,
+    _padding2: u32,
 }
 
 // `paint` bit layout:
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -818,6 +818,7 @@ impl WebGlPrograms {
                     strip_height: u32::from(Tile::HEIGHT),
                     alphas_tex_width_bits: max_texture_dimension_2d.trailing_zeros(),
                     encoded_paints_tex_width_bits: max_texture_dimension_2d.trailing_zeros(),
+                    _padding: [0; 3],
                 };
 
                 gl.bind_buffer(
@@ -841,6 +842,7 @@ impl WebGlPrograms {
                     strip_height: u32::from(Tile::HEIGHT),
                     alphas_tex_width_bits: max_texture_dimension_2d.trailing_zeros(),
                     encoded_paints_tex_width_bits: max_texture_dimension_2d.trailing_zeros(),
+                    _padding: [0; 3],
                 };
 
                 gl.bind_buffer(
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -966,6 +966,7 @@ impl Programs {
                 strip_height: Tile::HEIGHT.into(),
                 alphas_tex_width_bits: alpha_texture_width.trailing_zeros(),
                 encoded_paints_tex_width_bits: alpha_texture_width.trailing_zeros(),
+                _padding: [0; 3],
             }),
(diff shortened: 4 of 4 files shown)


sparse_strips/vello_sparse_tests/tests/issues.rs
```diff
@@ -553,3 +553,37 @@ fn issue_1433(ctx: &mut impl Renderer) {
...
+
+    let mut final_pix = Pixmap::new(1, 1);
+    final_pix.set_pixel(0, 0, PremulRgba8::from_u32(u32::from_be_bytes([255, 0, 0, 255])));
+    let final_source = ctx.get_image_source(Arc::new(final_pix));
+    ctx.set_paint(Image {
+        image: final_source,
+        sampler,
+    });
+    ctx.fill_rect(&Rect::new(0.0, 0.0, 10.0, 10.0));
+}
```
grebmeg: Missing new line

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -782,11 +782,17 @@ struct EncodedImage {
     translate: vec2<f32>,
 }
 
+// Convert a flat texel index to 2D texture coordinates for the encoded paints texture.
+fn encoded_paint_coord(flat_idx: u32) -> vec2<u32> {
+    let w = textureDimensions(encoded_paints_texture).x;
```
grebmeg: I’d be interested in exploring passing the texture dimensions as a uniform later. I’m curious how it performs across different machines.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -782,11 +782,17 @@ struct EncodedImage {
...
+    return vec2<u32>(flat_idx % w, flat_idx / w);
```
grebmeg: If we can rely on `max_texture_dimension_2d` always being a power of two, we can use bitwise operations. We use the same pattern for `alphas_texture`.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
taj-p: Ohhh great idea. This would mimic what we do for `alphas_tex_width_bits`. Implemented in https://github.com/linebender/vello/pull/1467/commits/050add132d79b7ea113f968fdaa7b1b2f67aadef

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
taj-p: Great idea! Thank you! Implemented in https://github.com/linebender/vello/pull/1467/commits/050add132d79b7ea113f968fdaa7b1b2f67aadef

sparse_strips/vello_hybrid/src/render/common.rs
LaurenzV, line 31: I think if we set this to align(16), bytemuck will automatically complain in case there is not enough padding, which could be useful in case we change it again in the future

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -110,6 +110,13 @@ struct Config {
     // Number of trailing zeros in alphas_tex_width (log2 of width).
     // Pre-calculated on CPU since WebGL2 doesn't support `firstTrailingBit`.
     alphas_tex_width_bits: u32,
+    // Number of trailing zeros in encoded_paints_tex_width (log2 of width).
+    // Pre-calculated on CPU since WebGL2 doesn't support `firstTrailingBit`.
+    encoded_paints_tex_width_bits: u32,
+    // Padding to satisfy WebGL's 16-byte alignment requirement for uniform buffers.
+    _padding0: u32,
+    _padding1: u32,
+    _padding2: u32,
```
LaurenzV, line 119: Can we not use an array<3> here?

sparse_strips/vello_hybrid/src/render/common.rs
taj-p: Addressed in https://github.com/linebender/vello/pull/1474

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
taj-p: That's not possible. Apparently for downleveling uniforms to WebGL2, arrays require a stride of 16 bytes. See the table in [this section ](https://www.w3.org/TR/WGSL/#address-space-layout-constraints).

`array<S>` requires alignment `roundUp(16, AlignOf(S))`. 

edit: How wild that it requires so much memory. A big pain point for uniforms!

PR #1472, opened 2026-02-23
vello_common: Always inline `push_buf` and `pop_buf`

laurenz-canva:
Gives a decent speed bump on a scene with 750 push/pop layer commands:

Before:
image

After:
image


LaurenzV:
Note that #1414 does the same (for `push_buf`), but since it's such a small change I think it's worth landing this separately.

PR #1473, opened 2026-02-23
[Hybrid]: Add `SceneConstraints`

taj-p:
For our fast path work, we require the ability to constrain a scene to use default blend modes only. This PR introduces the ability to constrain a scene as such.


laurenz-canva:
Oh no, I didn't know this would get closed automatically 😅

PR #1478, opened 2026-02-24
vello_common: Update the ranges of all layers in case a wide tile is reset with a background color

### LINKED ISSUE #1477: Filtered layers that are overdrawn by a `bg` fill cause a panic
See the following test:

```rs
#[vello_test(width = 768, height = 4, skip_multithreaded, skip_hybrid)]
fn overdraw_after_filter_layer(ctx: &mut impl Renderer) {
    let filter = Filter::from_primitive(FilterPrimitive::Offset { dx: 0.0, dy: 0.0 });
    let rect = Rect::new(0.0, 0.0, 768.0, 4.0);

    // Push a filter layer and draw inside it.
    ctx.push_layer(None, None, None, None, Some(filter));
    ctx.set_paint(RED);
    ctx.fill_rect(&rect);
    ctx.pop_layer();

    // Full-width opaque fill at root level triggers overdraw elimination,
    // which clears cmds and layer_cmd_ranges for this tile.
    ctx.set_paint(BLACK);
    ctx.fill_rect(&rect);
}
```

The issue is that if we reset all commands in a wide tile due to there being a sparse fill that replaces the background, we only reset the layer ranges of the currently active layer. In the above test case, this would reset the layer ranges of the root layer. However, the ranges of the filter layer are never updated, and therefore are stale. As a result, if we later on try to rasterize the filter layer, we try to access a command index that is stale and doesn't even exist, causing a panic:

```
thread 'issues::overdraw_after_filter_layer_cpu_u8_neon' (653707) panicked at sparse_strips/vello_cpu/src/dispatch/single_threaded.rs:313:40:
index out of bounds: the len is 0 but the index is 1
```

The solution is to clear all layer command ranges in that wide tile, and in vello_cpu accept that some layers might not have any ranges associated with them.

laurenz-canva:
Closes #1477. See that issue for more information.

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
@@ -306,7 +306,8 @@ impl SingleThreadedDispatcher {
 
         // Process all commands in this layer's render range.
         // It can happen that the layer has no associated ranges in this wide tile in
-        // case they have been cleared by setting a new wide tile background.
+        // case they have been cleared by setting a new wide tile background, for example
+        // when filling a full-tile opaque solid color.
         let Some(ranges) = wtile.layer_cmd_ranges.get(&layer_id) else {
             return;
         };


sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -305,8 +305,11 @@ impl SingleThreadedDispatcher {
         fine.clear(clear_color);
 
         // Process all commands in this layer's render range.
-        // Invariant: tiles within a layer's bbox must have commands for that layer.
-        let ranges = wtile.layer_cmd_ranges.get(&layer_id).unwrap();
+        // It can happen that the layer has no associated ranges in this wide tile in
+        // case they have been cleared by setting a new wide tile background.
```
grebmeg:
```suggestion
        // It can happen that the layer has no associated ranges in this wide tile in
        // case they have been cleared by setting a new wide tile background, for example
        // when filling a full-tile opaque solid color.
```

PR #1479, opened 2026-02-24
vello_hybrid: Better interleaving of raw strips and layers

laurenz-canva:
# Context

With #1454 merged, we added a mechanism that allows for skipping coarse rasterization and scheduling work for path rendering operations, as long as they don't render to an intermediate layer. However, in that current version, if there is only one single `push_layer` command anywhere in the scene, all of these advantages are basically for nothing, as we retroactively run coarse rasterization for all commands then go through the normal scheduling process. Since it's not uncommon to have at least a few layers interspersed (for example for group opacities), this is not ideal.

#1475 represents Taj's attempt to resolve this by splitting the whole rendering process into interleaved pipelines of "render strips directly" and "do layer rendering". However, one major problem with the PR is that each siwtch between the two modes will incur additional render passes, which is not ideal if we have multiple layers interspersed in the scene.

# Solution

This PR attempts to solve this, by integrating the "raw strips pipeline" into the normal scheduling process, allowing us to make full use of the slot textures and keep the number of render passes to a minimum, even if there are multiple subsequent push layer calls. As long as we have enough slot textures (and the layers aren't nested), we only need to do one additional render pass to render the whole scene. In addition to that, this PR also shows that it should be possible to unify the two paths that we introduced previously in the scheduler, without incurring any additional cost for the common case of having no layers.

The key realization that makes this possible is that the main reason why scheduling was so expensive previously is that normal paths also went through coarse rasterization, and scheduling coarse rasterized paths is expensive, because we need to schedule each command in all wide tiles separately. However, now that we have a way of representing and rendering paths just with their strips, as long as we are rendering to the main surface we can just use our existing infrastructure of scheduling draw calls by directly generating the GPU strips and storing them in the round that we are currently scheduling. I think this approach is sound, which is also supported by the fact that all tests seem to pass.

# Performance
These are the numbers for a test scene of many rectangles I get, compared to **current main** (not Taj's branch! I haven't measured that, but I'm pretty sure this PR should be faster in all benchmarks)

Current main:
- 3250 rects, no layers -> 22ms per frame
- 750 rects, layer every rect -> 73ms per frame
- 750 rects, layer every second rect -> 46 ms per frame
**- 1750 rects, layer every 50th rect -> 25ms per frame**

New:
- 3250 rects, no layers -> 22ms per frame
- 750 rects, layer every rect -> 140ms
- 750 rects, layer every second rect -> 70ms
**- 1750, layer every 50th rect -> 18ms per frame**

For scenes with no layers at all, performance stays about the same. Unfortunately performance drops if we are pushing new layers very frequently. The reason for that is simply that processing a single layer is much more expensive now, since we need to over and over iterate over all wide tiles to process the commands only for that specific layer. On current main, we just need to iterate over all wide tiles once. However, if we only have very few layers (in this case every 50th draw call), this PR is a lot faster, because the more expensive layer processing cost is offset by the fact that the other 49 paths can be rendered directly instead of having to go through coarse rasterization.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -840,6 +840,7 @@ impl<const MODE: u8> Wide<MODE> {
     pub fn pop_layer(&mut self, render_graph: &mut RenderGraph) {
         // This method basically unwinds everything we did in `push_layer`.
         let mut layer = self.layer_stack.pop().unwrap();
+        let batch_count = self.batch_count;
 
         if let Some(filter) = &layer.filter {
             let mut final_bbox = WideTilesBbox::inverted();
@@ -880,8 +881,11 @@ impl<const MODE: u8> Wide<MODE> {
             for x in final_bbox.x0()..final_bbox.x1() {
                 for y in final_bbox.y0()..final_bbox.y1() {
                     let idx = self.get_idx(x, y);
-                    self.tiles[idx]
-                        .ensure_layer_stack_bufs(idx, &mut self.layers_needing_buf_stack);
+                    self.tiles[idx].ensure_layer_stack_bufs(
+                        idx,
+                        &mut self.layers_needing_buf_stack,
+                        batch_count,
+                    );
 
                     // Note that unlike commands like "mask" or "opacity" which only need to be applied
                     // to wide tiles with drawing commands, filter commands always need to be applied


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1844,7 +1826,7 @@ impl RendererBackend for WebGlRendererContext<'_> {
         self.do_clear_slots_render_pass(texture_index, slots);
     }
 
-    /// Execute a render pass for strips.
+    /// Execute a render pass for rendering strips.
```
```suggestion
    /// Execute a render pass for strips.
```

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -158,37 +153,18 @@ impl Renderer {
...
-            self.scheduler
-                .do_scene(&mut self.scheduler_state, &mut ctx, scene, &self.paint_idxs)
+        let mut ctx = RendererContext {
+            programs: &mut self.programs,
+            device,
+            queue,
+            encoder,
+            view,
         };
+        // Clear the surface before the unified scheduling pass.
```
taj-p:
```suggestion
```

FYI: This may conflict with Alex's render to atlas PR

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -209,25 +225,39 @@ pub struct Scene {
     /// This is the case if we have not performed any `push_layer` command. In such a case, we
     /// don't need to do coarse rasterization or scheduling and therefore save a lot of overhead.
     pub(crate) strips_fast_path_active: bool,
+    /// True when the scene contains fast path and scheduled content. This can happen when
```
```suggestion
    /// True when the scene contains fast path and scheduled strips. This can happen when
```

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -209,25 +225,39 @@ pub struct Scene {
...
-// When the fast path is active it is in `Append` mode, so `$strip_start` (captured before generation)
+// When the fast path is active AND we're at the top level (no layers pushed),
+// strip_storage is in `Append` mode, so `$strip_start` (captured before generation)
 // and the current length delimit the range for this path.
 //
-// When the fast path is inactive, `strip_storage` is in `Replace` mode where each generation starts
-// with a clear, so the whole buffer is the current path's strips.
+// When the fast path is inactive or we're inside a layer, `strip_storage` is in `Replace`
+// or `ReplaceAfter` mode where each generation starts with a clear/truncate, so the
+// relevant portion of the buffer is the current path's strips.
```
taj-p, line 252: Tidy up

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -611,6 +662,13 @@ impl Scene {
     /// Pop the last pushed layer.
     pub fn pop_layer(&mut self) {
         self.wide.pop_layer(&mut self.render_graph);
+        if self.fast_path_interleaved && !self.wide.has_layers() {
+            self.wide.emit_segment_end();
```
taj-p: This shows up in profiles. I think we can add these in lazily

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -442,47 +444,69 @@ impl Scheduler {
         Ok(slot)
     }
 
+    /// Schedule and render the scene.
+    ///
+    /// Iterates over the scene's command sequence, which interleaves fast-path
+    /// strips (rendered directly to the surface) and coarse-rasterized layer
+    /// content (scheduled through the round system). Fast-path strips are
+    /// appended directly to the current round's surface draw array, avoiding
+    /// separate GPU render calls.
```
taj-p: I think we can improve this somewhat

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -40,6 +40,19 @@ pub(crate) struct FastStripsPath {
...
+///
+/// This enum describes the interleaving of fast-path strips (rendered directly
+/// to the surface) and coarse-rasterized content (scheduled via the round system).
+#[derive(Debug)]
+pub(crate) enum SceneCommand {
+    /// Fast-path strips to render directly to the surface.
+    /// The range indexes into `FastStripsBuffer.paths`.
+    DirectStrips(Range<usize>),
+    /// Coarse-rasterized layer content. Process all remaining wide tile commands.
+    Scheduled,
```
taj-p: TODO: Think of better names. Maybe `Direct` and `Coarse` or `Direct` and `WideTile`

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -442,47 +444,69 @@ impl Scheduler {
...
-                    &scene.wide.attrs,
-                )?;
+        let wide = &scene.wide;
+        let rows = wide.height_tiles();
+        let cols = wide.width_tiles();
+        let num_tiles = (rows * cols) as usize;
+
+        // Per-tile offset tracking for Scheduled segments.
+        // Each Scheduled command processes cmds from offsets[tile] to the end.
+        let mut offsets = alloc::vec![0usize; num_tiles];
```
taj-p: Don't pay the allocation cost of this each frame

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -501,6 +525,99 @@ impl Scheduler {
         Ok(())
     }
 
+    /// Generate GpuStrips for a range of fast-path paths and append them
+    /// directly into the current round's surface draw array.
+    fn emit_direct_strips(
```
```suggestion
    fn push_direct_strips(
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -501,6 +525,99 @@ impl Scheduler {
...
+        paint_idxs: &[u32],
+    ) {
+        let strip_storage = scene.strip_storage.borrow();
+        let draw = self.draw_mut(self.round, 2);
+        for path in &scene.fast_strips_buffer.paths[range] {
+            generate_gpu_strips_for_path(path, &strip_storage, scene, paint_idxs, &mut draw.0);
+        }
+    }
+
+    /// Process one segment of coarse-rasterized wide tile commands.
```
taj-p: Need to consolidate naming. Sometimes its segment, sometimes its batch

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -611,6 +660,13 @@ impl Scene {
     /// Pop the last pushed layer.
     pub fn pop_layer(&mut self) {
         self.wide.pop_layer(&mut self.render_graph);
+        if self.fast_path_interleaved && !self.wide.has_layers() {
+            self.wide.end_batch();
+            self.scene_commands.push(SceneCommand::Coarse);
```
taj-p: cc @LaurenzV  - I think we want to _not_ do this because then we're creating unnecessary batching for this case:

```
push_layer(...)
   ...
pop_layer()
push_layer(...)
   ...
pop_layer()
```

This will cause us to call `process_coarse_batch` in `schedule.rs` twice, but we only need to run it once.

I think a simpler approach is to record the fast path batches and issue a `process_coarse_batch` between each fast path batch. Does that make sense to you?

taj-p: Addressed in https://github.com/linebender/vello/pull/1479/commits/6a42e8cd56f5339f1b2f236c0f6b7e616ca8e351

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1293,8 +1339,10 @@ impl<const MODE: u8> WideTile<MODE> {
             //
             // However, the extra cost of tracking such optimizations may outweigh the
             // benefit, especially in hybrid mode with GPU painting.
-            let can_override =
-                x == 0 && width == WideTile::WIDTH && self.n_clip == 0 && self.n_bufs == 0;
+
+            // TODO: Change this back once https://github.com/linebender/vello/pull/1478 has been
+            // merged.
+            let can_override = false;
```
taj-p: TODO: Can't forget this part

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -255,7 +250,8 @@ impl Renderer {
         encoder: &mut CommandEncoder,
         render_size: &RenderSize,
         view: &TextureView,
-        clear: bool,
+        // See https://github.com/linebender/vello/pull/1458/changes#r2851077556
+        _clear: bool,
```
taj-p, line 255: I think we need to honour this `clear` value

taj-p: (Unsure whether we do that in this PR or another)

LaurenzV: I think it shouldn't be too hard to fix, but doing it in a follow-up seems better to me, especially since it was seemingly already broken before. But we need to make sure @grebmeg is aware.

LaurenzV:
@taj-p I think this is ready for review now, I'm pretty happy with the current state! Depending on when Tom merges #1414, I might have to do another rebase though, there will be conflicts but conceptually the PRs should be compatible with each other.

LaurenzV:
Should be all done now!

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -31,6 +31,37 @@ use vello_common::util::{is_integer_rect, is_integer_translation};
 /// Default tolerance for curve flattening
 pub(crate) const DEFAULT_TOLERANCE: f64 = 0.1;
 
+/// The pipeline mode for strip rendering.
+///
+/// Determines whether strips are sent directly to the GPU (fast path),
+/// go through coarse rasterization, or a mix of both.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum StripPathMode {
```
taj-p, line 39: Love this!

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -501,6 +546,99 @@ impl Scheduler {
...
+        cmd_offsets: &mut [usize],
+        paint_idxs: &[u32],
+    ) -> Result<(), RenderError> {
+        for row in 0..rows {
+            for col in 0..cols {
+                let idx = (row * cols + col) as usize;
+                let tile = wide.get(col, row);
+                let start_offset = cmd_offsets[idx];
+
+                // Note that we are explicitly checking >= instead of >.
```
taj-p: But we are using `>` not `>=`

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -501,6 +546,99 @@ impl Scheduler {
...
+                // We only must paint the background if we are processing the wide tile for the
+                // first time (i.e. the start offset is 0).
+                let paint_bg = start_offset == 0;
+
+                // Find the end of this batch: scan for the next `BatchEnd` marker.
+                let end = tile.cmds[start_offset..]
+                    .iter()
+                    .position(|c| matches!(c, Cmd::BatchEnd))
+                    .map(|p| start_offset + p)
+                    .unwrap_or(tile.cmds.len());
```
taj-p: Do we need to do this scan or can `do_tile` return when it encounters a `Cmd::BatchEnd`?

The `end > start_offset` check could just check whether there is a command after `start_offset` right? (That the length is greater than `start_offset`)

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -836,6 +989,7 @@ impl Scheduler {
                 Cmd::Blend(mode) => {
                     self.do_blend(state, wide_tile_x, wide_tile_y, mode);
                 }
+                Cmd::BatchEnd => {}
```
taj-p: In the current implementation, we should expect this to be `unreachable!()` right? Otherwise it's a bug. This mightn't be needed after https://github.com/linebender/vello/pull/1479/changes#r2855280583

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1189,65 +1343,61 @@ fn has_non_zero_alpha(rgba: u32) -> bool {
...
+    let strips = &strip_storage.strips[path.strips.clone()];
 
-    for path in paths {
-        let strips = &strip_storage.strips[path.strips.clone()];
-
-        if strips.is_empty() {
-            continue;
-        }
+    if strips.is_empty() {
+        return;
```
taj-p, line 1356: @reviewers - I recommend hiding whitespace for this diff:

Image

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -442,47 +447,87 @@ impl Scheduler {
         Ok(slot)
     }
 
+    /// Schedule and render the scene.
+    ///
+    /// Iterates over the commands encoded in the scene.
+    ///
+    /// We interleave fast path strips (rendered directly to the surface) and coarse-rasterized strips.
```
taj-p: Does this comment need to be updated per the changes of `StripPathMode`? We don't always interleave.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Ah yes, the comment is wrong here. We are checking > so that if start_offset is 0 but there are also 0 commands, we don't skip.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Yes I also had the same thought, I just did this because that was in the existing code. I think it should be possible to let the function return the index.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Mmm yeah I _think_ it shouldn't be reachable.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Yes, that probably needs updating, good catch!

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -535,7 +579,21 @@ impl Scene {
...
         self.constraints.assert_blend_mode(blend_mode_val);
 
-        self.flush_fast_path();
+        let strip_offset;
+        if self.constraints.use_default_blending_only() {
+            // With default blending only we can keep fast path strips alive. Record a
+            // split point so the scheduler knows to process one coarse batch after
+            // processing fast path strips up to this point.
+            let split = self.fast_strips_buffer.paths.len();
+            self.coarse_batch_splits.push(split);
```
grebmeg: This pushes a split on every `push_layer`, but `end_batch()` in `pop_layer` only fires when the outermost layer is popped (`!self.wide.has_layers()`). So nested layers produce duplicate split entries (e.g., [1, 1]) that each trigger a full `process_coarse_batch` pass over every tile for no useful work, right?
Maybe consider guarding this with the same condition?

```rust
if !self.wide.has_layers() {
    self.coarse_batch_splits.push(split);
}
```
That way splits and `BatchEnd` markers stay 1:1.

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1832,6 +1899,10 @@ pub enum Cmd {
     ///
     /// Modulates the alpha channel of the buffer using the provided mask.
     Mask(Mask),
+    /// Marks a boundary between rendering fast path strips and coarse rasterized strips.
+    ///
+    /// Only meaningful in `MODE_HYBRID`.
+    BatchEnd,
```
grebmeg, line 1918: I'm not sure `BatchEnd` belongs in `Cmd`: it's a scheduling marker rather than a rendering command, and it leaks a `MODE_HYBRID`-only concern into the shared command enum. This also meant adding `Clone` to `Cmd` (and two sub-structs), and the CPU fine rasterizer now has a dead `=> {}` arm for it.

Would it make sense to store batch boundaries in a separate field on `WideTile` instead (e.g., `batch_ends: Vec<usize>`)? The lazy watermark approach would work the same way, just pushing indices rather than commands. What do you think?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -442,47 +447,96 @@ impl Scheduler {
...
+        let wide = &scene.wide;
+        let rows = wide.height_tiles();
+        let cols = wide.width_tiles();
+        let num_tiles = (rows * cols) as usize;
+
+        // A bit hacky, but we need this since we still need mutable access to
+        // self when processing everything.
+        let mut cmd_offsets = core::mem::take(&mut self.cmd_offsets);
+        cmd_offsets.clear();
+        cmd_offsets.resize(num_tiles, 0);
```
grebmeg, line 483: What do you think about moving the `cmd_offsets` clear+resize inside the `CoarseOnly` and `Interleaved` arms as it's not used  in `FastOnly`?

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: I'll fix that

sparse_strips/vello_hybrid/src/scene.rs
laurenz-canva: Good catch!

sparse_strips/vello_common/src/coarse.rs
laurenz-canva: Could be doable. Will add a TODO for now, since I don't want to change the logic too much now.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: We could do it but I think it should be very negligible in performance and it forces us to duplicate the code, so it seems to me it would be better to keep it this way, if you don't mind.

PR #1480, opened 2026-02-25
vello_hybrid: Expose glyph cache rendering state APIs publicly

grebmeg:
Expose `RenderState`, `Scene::take_current_state`, `Scene::restore_state`, and `Renderer::image_cache` as public APIs so that downstream consumers (e.g. glyph cache implementations) can save/restore scene rendering state and access the image cache directly.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -88,7 +88,7 @@ pub struct WebGlRenderer {
     /// WebGL context.
     gl: WebGl2RenderingContext,
     /// Image cache for storing images atlas allocations.
-    pub(crate) image_cache: ImageCache,
+    pub image_cache: ImageCache,
     /// Encoded paints for storing encoded paints.
     encoded_paints: Vec<GpuEncodedPaint>,
     /// Stores the index (offset) of the encoded paints in the encoded paints texture.


sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -726,6 +726,28 @@ impl Scene {
     pub fn height(&self) -> u16 {
         self.height
     }
+
+    /// Save current rendering state.
+    pub fn take_current_state(&mut self) -> RenderState {
```
LaurenzV, line 731: nit, but I think it's a bit inconsistent that
1) This method is called `take_state` but the description says "save the state"
2) We use `mem::take` for the stroke (essentially setting it back to its default value) but copy for all other fields.

Maybe we can rename to `save_current_state` and just use `clone` for the stroke for consistency?

LaurenzV: Nevermind, I realized that code already existed before already. 😅 But since we are making it public now maybe it's worth adjusting.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -86,7 +86,7 @@ pub struct Renderer {
     /// The state used by the scheduler.
     scheduler_state: SchedulerState,
     /// Image cache for storing images atlas allocations.
-    pub(crate) image_cache: ImageCache,
+    pub image_cache: ImageCache,
```
LaurenzV, line 89: Why only in wgpu backend but not in webgl?

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: I think the inclination here is to make it behave fully like `take`, since it’s used in many places where we want to avoid any copying. So I’m inclined to leave it as it is for now, but I’d definitely like to come back to it later.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Oh, yeah, thanks!

PR #1481, opened 2026-02-25
vello_common: Replace external `guillotiere` with built-in no_std-compatible port

grebmeg:
The glyph rendering path in `parley_draw` I've been working on depends on `multi_atlas` and `image_cache` for atlas allocation. Since [`guillotiere`](https://github.com/nical/guillotiere/blob/main/Cargo.toml) requires `std` (via `euclid/num-traits`), these modules were gated behind `cfg(feature = "std")`, which would force the entire glyph rendering pipeline to be `std`-gated as well. Porting the allocator to `no_std` removes this constraint and keeps the glyph rendering path available in `no_std` builds.

This PR imports and modifies `guillotiere`'s tree-based guillotine algorithm directly into `allocator.rs`, replacing `euclid` types with local `Size/Point/Rect` equivalents and swapping `std` imports for `core`/`alloc`.

We discussed in one of the recent office hours that it would be fine to add the simpler shelf allocator under `vello_common/`. However, since my glyph caching is already based on the ability to deallocate, I thought there was no harm in continuing to use the current algorithm. I have not added this to upstream because I am not aware of the plan there or the current review capacity, but I am happy to do so if we decide that is the right approach.

### DIFF
--- a/sparse_strips/vello_common/src/allocator.rs
+++ b/sparse_strips/vello_common/src/allocator.rs
@@ -6,6 +6,9 @@
 //! Ported from [guillotiere](https://github.com/nical/guillotiere) (Apache-2.0 OR MIT)
 //! by Nicolas Silva. Adapted for `no_std` by replacing `euclid` types with local
 //! equivalents and removing `std`-only functionality (SVG dump, serde).
+//!
+//! TODO: Once [servo/euclid#543](https://github.com/servo/euclid/pull/543) is merged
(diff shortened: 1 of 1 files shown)


grebmeg:
Hey @LaurenzV, I think I’d prefer to merge this for now, and then once servo/euclid#543 is merged and `guillotiere` is updated, we can switch the current usage to the official repo.

PR #1482, opened 2026-02-25
vello_hybrid: Faster rectangle rendering

laurenz-canva:
This PR attempts to provide an alternative to #1470 (well, it's not intended to be a replacement and we can still try to implement an approach similar to #1470, but this PR achieves a similar effect with much less additional code complexity). It works for translated and scaled rectangles, but not skewed ones. 

It basically allows us to skip strip generation for rectangles and just treat it as a special type of strip, which can then utilize the existing infrastructure to support rectangles with arbitrary colors/gradients/images (unlike #1470, which just supports images). However, the flipside is that unlike #1470, we don't use GPU-native capabilities for bilinear filtering, which I believe was another motivation for #1470.

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -583,9 +583,7 @@ impl Scheduler {
                     );
                 }
                 FastStripCommand::Rect(r) => {
-                    let strip = pack_rectangle_into_gpu(
-                        r, scene, paint_idxs
-                    );
+                    let strip = pack_rectangle_into_gpu(r, scene, paint_idxs);
 
                     draw.0.push(strip);
                 }
@@ -1440,8 +1438,7 @@ fn pack_rectangle_into_gpu(rect: &FastPathRect, scene: &Scene, paint_idxs: &[u32
     let width = (sx1 - sx0) as u16;
     let height = (sy1 - sy0) as u16;
 
-    let (payload, paint_packed) =
-        Scheduler::process_paint(&rect.paint, scene, (x, y), paint_idxs);
+    let (payload, paint_packed) = Scheduler::process_paint(&rect.paint, scene, (x, y), paint_idxs);
 
     // Determine the fractional offsets for anti-aliasing and quantize so it
     // fits into u8.


nicoburns:
This makes a lot of sense to me. My understanding is that pixel-snapping is *usually* wanted (including for glyphs), but there are exceptions (often around animated content where you may prefer a smoother animation over a sharp rendering (which won't look sharp in motion anyway)).

> It basically allows us to skip strip generation for rectangles and just treat it as a special type of strip

I suppose this would be a "dense strip"?

> However, the flipside is that unlike https://github.com/linebender/vello/pull/1470, we don't use GPU-native capabilities for bilinear filtering, which I believe was another motivation for https://github.com/linebender/vello/pull/1470.

This is for rendering images at a non-native size?

LaurenzV:
> This is for rendering images at a non-native size?

Basically whenever you set the filtering mode to bilinear (= image quality medium)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -507,6 +536,59 @@ impl Scene {
...
+            return false;
+        }
+
+        // TODO: Reject mirrored rectangles
+
+        let paint = self.encode_current_paint();
+        let transformed_rect = self.transform.transform_rect_bbox(*rect);
+
+        // Snap to integer coordinates.
+        let snapped = transformed_rect.expand();
```
taj-p: Consider this tiling of rectangles with about a 1px gap between them. The gap collapses under expand.

https://github.com/user-attachments/assets/36b99827-9417-4de0-b154-12ea987416ed

It also shimmers under pan 🤔 (first half: pixel snap on. last half: pixel snap off).

https://github.com/user-attachments/assets/3a5324ef-ac74-4929-9239-830705b20917




<details>
<summary>example scene</summary>
```rs
pub fn render(ctx: &mut impl RenderingContext, root_transform: Affine) {
    let colors: &[Color] = &[
        palette::css::CORNFLOWER_BLUE,
        palette::css::CORAL,
        palette::css::MEDIUM_SEA_GREEN,
        palette::css::GOLD,
        palette::css::MEDIUM_ORCHID,
        palette::css::TOMATO,
    ];

    // Row 1: rectangles at sub-pixel offsets to show snapping behaviour.
    for (i, &color) in colors.iter().enumerate() {
        let frac = i as f64 * 0.15;
        let x = 20.0 + i as f64 * 110.0 + frac;
        let y = 20.0 + frac;
        ctx.set_transform(root_transform);
        ctx.set_paint(color);
        ctx.fill_rect(&Rect::new(x, y, x + 90.0, y + 60.0));
    }

    // Row 2: small rectangles in a tight grid (1px gaps that collapse under expand()).
    let grid_y = 120.0;
    let cell = 16.0;
    let gap = 1.0;
    for row in 0..4 {
        for col in 0..20 {
            let x = 20.0 + col as f64 * (cell + gap) + 0.3;
            let y = grid_y + row as f64 * (cell + gap) + 0.7;
            let ci = (row * 20 + col) % colors.len();
            ctx.set_transform(root_transform);
            ctx.set_paint(colors[ci]);
            ctx.fill_rect(&Rect::new(x, y, x + cell, y + cell));
        }
    }

    // Row 3: overlapping semi-transparent rectangles to show expand() overlap.
    let alpha_colors: &[Color] = &[
        palette::css::CORNFLOWER_BLUE.with_alpha(0.5),
        palette::css::CORAL.with_alpha(0.5),
        palette::css::MEDIUM_SEA_GREEN.with_alpha(0.5),
    ];
    for (i, &color) in alpha_colors.iter().enumerate() {
        let x = 20.0 + i as f64 * 80.0 + 0.4;
        let y = 210.0 + 0.6;
        ctx.set_transform(root_transform);
        ctx.set_paint(color);
        ctx.fill_rect(&Rect::new(x, y, x + 120.0, y + 60.0));
    }

    // Row 4: scaled rectangles at fractional positions.
    for (i, &color) in colors.iter().enumerate() {
        let scale = 1.0 + i as f64 * 0.5;
        let x = 20.0 + i as f64 * 60.0;
        let y = 300.0;
        ctx.set_transform(root_transform * Affine::translate((x + 0.3, y + 0.7)) * Affine::scale(scale));
        ctx.set_paint(color);
        ctx.fill_rect(&Rect::new(0.0, 0.0, 30.0, 20.0));
    }
}
```

</details>

LaurenzV: Hmm yeah this is certainly not nice... Something else we could try is to store a quantized version of the fractional value of x/y/width/height (8 bits should be enough, so that it fits into u32) and then calculate the alpha value for border pixels inside of the fragment shader? This might allow us to support a basic form of anti-aliasing

LaurenzV: Can you send the code for the scene so I can try it later? 😄

taj-p:
```rs
pub fn render(ctx: &mut impl RenderingContext, root_transform: Affine) {
    let colors: &[Color] = &[
        palette::css::CORNFLOWER_BLUE,
        palette::css::CORAL,
        palette::css::MEDIUM_SEA_GREEN,
        palette::css::GOLD,
        palette::css::MEDIUM_ORCHID,
        palette::css::TOMATO,
    ];

    // Row 1: rectangles at sub-pixel offsets to show snapping behaviour.
    for (i, &color) in colors.iter().enumerate() {
        let frac = i as f64 * 0.15;
        let x = 20.0 + i as f64 * 110.0 + frac;
        let y = 20.0 + frac;
        ctx.set_transform(root_transform);
        ctx.set_paint(color);
        ctx.fill_rect(&Rect::new(x, y, x + 90.0, y + 60.0));
    }

    // Row 2: small rectangles in a tight grid (1px gaps that collapse under expand()).
    let grid_y = 120.0;
    let cell = 16.0;
    let gap = 1.0;
    for row in 0..4 {
        for col in 0..20 {
            let x = 20.0 + col as f64 * (cell + gap) + 0.3;
            let y = grid_y + row as f64 * (cell + gap) + 0.7;
            let ci = (row * 20 + col) % colors.len();
            ctx.set_transform(root_transform);
            ctx.set_paint(colors[ci]);
            ctx.fill_rect(&Rect::new(x, y, x + cell, y + cell));
        }
    }

    // Row 3: overlapping semi-transparent rectangles to show expand() overlap.
    let alpha_colors: &[Color] = &[
        palette::css::CORNFLOWER_BLUE.with_alpha(0.5),
        palette::css::CORAL.with_alpha(0.5),
        palette::css::MEDIUM_SEA_GREEN.with_alpha(0.5),
    ];
    for (i, &color) in alpha_colors.iter().enumerate() {
        let x = 20.0 + i as f64 * 80.0 + 0.4;
        let y = 210.0 + 0.6;
        ctx.set_transform(root_transform);
        ctx.set_paint(color);
        ctx.fill_rect(&Rect::new(x, y, x + 120.0, y + 60.0));
    }

    // Row 4: scaled rectangles at fractional positions.
    for (i, &color) in colors.iter().enumerate() {
        let scale = 1.0 + i as f64 * 0.5;
        let x = 20.0 + i as f64 * 60.0;
        let y = 300.0;
        ctx.set_transform(root_transform * Affine::translate((x + 0.3, y + 0.7)) * Affine::scale(scale));
        ctx.set_paint(color);
        ctx.fill_rect(&Rect::new(0.0, 0.0, 30.0, 20.0));
    }
}
```

taj-p: > Something else we could try is to store a quantized version of the fractional value of x/y/width/height (8 bits should be enough, so that it fits into u32) and then calculate the alpha value for border pixels inside of the fragment shader? This might allow us to support a basic form of anti-aliasing

I'm thinking that https://github.com/linebender/vello/pull/1482#pullrequestreview-3856496494 might be even simpler

taj-p:
vibe coded implementation of allowing non-integer rects in the fast path

```diff
diff --git a/sparse_strips/vello_common/src/rect.rs b/sparse_strips/vello_common/src/rect.rs
index ed0a1b96..af40720d 100644
--- a/sparse_strips/vello_common/src/rect.rs
+++ b/sparse_strips/vello_common/src/rect.rs
@@ -8,6 +8,7 @@ use crate::kurbo::Rect;
 use crate::kurbo::common::FloatFuncs as _;
 use crate::strip::Strip;
 use crate::tile::Tile;
+use crate::util::normalized_mul_u8x16;
 use alloc::vec::Vec;
 use fearless_simd::*;
 
@@ -21,7 +22,10 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
     dispatch!(level, simd => render_impl(simd, rect, strip_buf, alpha_buf));
 }
 
-/// Generates strip data for a pixel-aligned rectangle.
+/// Generates strip data for an axis-aligned rectangle, including fractional coordinates.
+///
+/// Edge pixels that are only partially covered by the rectangle receive proportional
+/// alpha values (0–255) based on the area of overlap, rather than binary 0/255.
 ///
 /// # Strip layout strategy
 ///
@@ -29,8 +33,8 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
 ///
 /// - **Edge rows** (top/bottom of rect): the rect boundary crosses partway
 ///   through the tile vertically, so individual pixels need per-cell alpha.
-///   We emit a *single wide strip* spanning all tile columns, with alpha =
-///   `x_mask & y_mask` (each is 0x00 or 0xFF, so AND gives the intersection).
+///   We emit a *single wide strip* spanning all tile columns, with per-pixel
+///   alpha = `(x_coverage * y_coverage + 128) / 255`.
 ///
 /// - **Interior rows**: every pixel in the tile has full vertical coverage,
 ///   so we only need to handle the left and right partial-column edges.
@@ -40,23 +44,27 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
 ///
 /// The x-alpha masks for the left/right edge tiles are y-independent, so they
 /// are precomputed once and reused across all interior rows.
-// TODO: Consider extending this to handle arbitrary axis-aligned rectangles (with fractional
-// coordinates) by computing partial coverage alpha values for edge pixels instead of
-// binary 0/255.
 fn render_impl<S: Simd>(s: S, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &mut Vec<u8>) {
-    // TODO: Negative rect coordinates are not handled correctly — casting negative
-    // f64 to u16 saturates to 0. The caller currently clamps to the viewport, but if
-    // that changes, this will need signed-integer math or explicit clamping.
-    let rect_x0 = rect.x0.floor() as u16;
-    let rect_y0 = rect.y0.floor() as u16;
-    let rect_x1 = rect.x1.ceil() as u16;
-    let rect_y1 = rect.y1.ceil() as u16;
-
-    let left_tile_x = (rect_x0 / Tile::WIDTH) * Tile::WIDTH;
-    let right_tile_x = (rect_x1 / Tile::WIDTH) * Tile::WIDTH;
-
-    let y0 = (rect_y0 / Tile::HEIGHT) * Tile::HEIGHT;
-    let y1 = (rect_y1.saturating_add(Tile::HEIGHT - 1) / Tile::HEIGHT) * Tile::HEIGHT;
+    if rect.x0 >= rect.x1 || rect.y0 >= rect.y1 {
+        return;
+    }
+
+    // Pixel-snapped bounds for tile iteration; the original f64 coordinates
+    // drive the per-pixel coverage computation.
+    // NOTE: Negative rect coordinates are not handled correctly — casting
+    // negative f64 to u16 saturates to 0. The caller currently clamps to the
+    // viewport, but if that changes, this will need signed-integer math or
+    // explicit clamping.
+    let px_x0 = rect.x0.floor() as u16;
+    let px_y0 = rect.y0.floor() as u16;
+    let px_x1 = rect.x1.ceil() as u16;
+    let px_y1 = rect.y1.ceil() as u16;
+
+    let left_tile_x = (px_x0 / Tile::WIDTH) * Tile::WIDTH;
+    let right_tile_x = (px_x1 / Tile::WIDTH) * Tile::WIDTH;
+
+    let y0 = (px_y0 / Tile::HEIGHT) * Tile::HEIGHT;
+    let y1 = (px_y1.saturating_add(Tile::HEIGHT - 1) / Tile::HEIGHT) * Tile::HEIGHT;
     // Include one tile past the right edge so the right-edge tile column is
     // covered by the edge-row wide-strip loop.
     let x_end = right_tile_x.saturating_add(Tile::WIDTH);
@@ -71,26 +79,29 @@ fn render_impl<S: Simd>(s: S, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf:
     // A right strip is only needed when the rect spans more than one tile column.
     let needs_right_strip = right_tile_x > left_tile_x;
 
-    let left_x_mask = x_alpha_tile(s, left_tile_x, rect_x0, rect_x1);
-    let right_x_mask = x_alpha_tile(s, right_tile_x, rect_x0, rect_x1);
+    let left_x_mask = x_alpha_tile(s, left_tile_x, rect.x0, rect.x1);
+    let right_x_mask = x_alpha_tile(s, right_tile_x, rect.x0, rect.x1);
 
     for tile_y in tile_start_y..tile_end_y {
         let strip_y = tile_y * Tile::HEIGHT;
+        let strip_y_f = f64::from(strip_y);
+        let strip_y_end = f64::from(strip_y + Tile::HEIGHT);
 
         // A row is an "edge" if the rect's top or bottom boundary falls
         // *inside* it (i.e. partial vertical coverage).
-        let is_top_edge = strip_y < rect_y0 && rect_y0 < strip_y + Tile::HEIGHT;
-        let is_bottom_edge = strip_y < rect_y1 && rect_y1 < strip_y + Tile::HEIGHT;
+        let is_top_edge = strip_y_f < rect.y0 && rect.y0 < strip_y_end;
+        let is_bottom_edge = strip_y_f < rect.y1 && rect.y1 < strip_y_end;
 
         if is_top_edge || is_bottom_edge {
             let alpha_start = alpha_buf.len() as u32;
-            let y_mask = y_alpha_tile(s, strip_y, rect_y0, rect_y1);
+            let y_mask = y_alpha_tile(s, strip_y, rect.y0, rect.y1);
 
-            // Walk every tile column, AND the per-column x-mask with the
-            // per-row y-mask to get the final per-pixel alpha.
+            // Walk every tile column, multiply the per-column x-coverage with
+            // the per-row y-coverage to get the final per-pixel alpha.
             let mut col = left_tile_x;
             while col + Tile::WIDTH <= x_end {
-                let combined = x_alpha_tile(s, col, rect_x0, rect_x1) & y_mask;
+                let x_mask = x_alpha_tile(s, col, rect.x0, rect.x1);
+                let combined = s.narrow_u16x16(normalized_mul_u8x16(x_mask, y_mask));
                 alpha_buf.extend_from_slice(combined.as_slice());
                 col += Tile::WIDTH;
             }
@@ -123,18 +134,17 @@ fn render_impl<S: Simd>(s: S, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf:
 
 /// Build a column-major x-alpha mask for one tile-width of columns.
 ///
-/// Each column gets `Tile::HEIGHT` lanes, all 0x00 or all 0xFF depending on
-/// whether the column falls inside `[rect_x0, rect_x1)`.
+/// Each column gets `Tile::HEIGHT` identical lanes whose value is the
+/// horizontal coverage of that column by the rect `[rect_x0, rect_x1)`:
+/// 255 for fully covered, 0 for fully outside, and a proportional value
+/// for partially covered edge columns.
 #[inline(always)]
-fn x_alpha_tile<S: Simd>(s: S, tile_x: u16, rect_x0: u16, rect_x1: u16) -> u8x16<S> {
+fn x_alpha_tile<S: Simd>(s: S, tile_x: u16, rect_x0: f64, rect_x1: f64) -> u8x16<S> {
     let mut buf = [0_u8; 16];
     for col in 0..Tile::WIDTH {
-        let px = tile_x + col;
-        let alpha = if px >= rect_x0 && px < rect_x1 {
-            255
-        } else {
-            0
-        };
+        let px = f64::from(tile_x + col);
+        let coverage = (px + 1.0).min(rect_x1) - px.max(rect_x0);
+        let alpha = (coverage.clamp(0.0, 1.0) * 255.0 + 0.5) as u8;
         let base = (col * Tile::HEIGHT) as usize;
         buf[base..base + Tile::HEIGHT as usize].fill(alpha);
     }
@@ -143,19 +153,17 @@ fn x_alpha_tile<S: Simd>(s: S, tile_x: u16, rect_x0: u16, rect_x1: u16) -> u8x16
 
 /// Build a column-major y-alpha mask for one tile row.
 ///
-/// Each of the `Tile::HEIGHT` rows is 0x00 or 0xFF depending on whether it
-/// falls inside `[rect_y0, rect_y1)`. The pattern is identical across all
-/// `Tile::WIDTH` columns.
+/// Each of the `Tile::HEIGHT` rows gets a coverage value based on how much
+/// it overlaps `[rect_y0, rect_y1)`: 255 for fully covered, 0 for fully
+/// outside, and a proportional value for partially covered edge rows.
+/// The pattern is identical across all `Tile::WIDTH` columns.
 #[inline(always)]
-fn y_alpha_tile<S: Simd>(s: S, strip_y: u16, rect_y0: u16, rect_y1: u16) -> u8x16<S> {
+fn y_alpha_tile<S: Simd>(s: S, strip_y: u16, rect_y0: f64, rect_y1: f64) -> u8x16<S> {
     let mut y_mask = [0_u8; 4];
     for row in 0..Tile::HEIGHT {
-        let py = strip_y + row;
-        y_mask[row as usize] = if py >= rect_y0 && py < rect_y1 {
-            255
-        } else {
-            0
-        };
+        let py = f64::from(strip_y + row);
+        let coverage = (py + 1.0).min(rect_y1) - py.max(rect_y0);
+        y_mask[row as usize] = (coverage.clamp(0.0, 1.0) * 255.0 + 0.5) as u8;
     }
     let mut buf = [0_u8; 16];
     for col in 0..Tile::WIDTH as usize {
diff --git a/sparse_strips/vello_common/src/util.rs b/sparse_strips/vello_common/src/util.rs
index ef6aad4a..a02f327a 100644
--- a/sparse_strips/vello_common/src/util.rs
+++ b/sparse_strips/vello_common/src/util.rs
@@ -9,7 +9,7 @@ use fearless_simd::{
 };
 #[cfg(not(feature = "std"))]
 use peniko::kurbo::common::FloatFuncs as _;
-use peniko::kurbo::{Affine, Rect};
+use peniko::kurbo::Affine;
 
 /// Convert f32x16 to u8x16.
 #[inline(always)]
@@ -62,6 +62,19 @@ pub fn normalized_mul_u8x16<S: Simd>(a: u8x16<S>, b: u8x16<S>) -> u16x16<S> {
     (S::widen_u8x16(a.simd, a) * S::widen_u8x16(b.simd, b)).div_255()
 }
 
+/// Check if an affine transform preserves axis alignment.
+///
+/// Returns true when the off-diagonal coefficients are zero (`b ≈ 0, c ≈ 0`),
+/// which covers any combination of scaling, reflection, and translation.
+/// Axis-aligned rectangles remain axis-aligned after this transform, so
+/// `transform_rect_bbox` produces the exact transformed rect rather than a
+/// loose bounding box.
+#[inline]
+pub fn is_axis_preserving(transform: &Affine) -> bool {
+    let [_, b, c, _, _, _] = transform.as_coeffs();
+    b.is_nearly_zero() && c.is_nearly_zero()
+}
+
 /// Check if an affine transform is a pure integer translation.
 ///
 /// Returns true if the transform only contains integer translation (no rotation,
@@ -77,18 +90,6 @@ pub fn is_integer_translation(transform: &Affine) -> bool {
         && (f - f.round()).is_nearly_zero()
 }
 
-/// Check if rect coordinates are all integers (no fractional parts).
-///
-/// The optimized rect path doesn't handle anti-aliasing for fractional edges,
-/// so non-integer coordinates require path-based rendering.
-#[inline]
-pub fn is_integer_rect(rect: &Rect) -> bool {
-    (rect.x0 - rect.x0.round()).is_nearly_zero()
-        && (rect.y0 - rect.y0.round()).is_nearly_zero()
-        && (rect.x1 - rect.x1.round()).is_nearly_zero()
-        && (rect.y1 - rect.y1.round()).is_nearly_zero()
-}
-
 /// Extract scale factors from an affine transform using singular value decomposition.
 ///
 /// Returns a tuple of (`scale_x`, `scale_y`) representing the scale along each axis.

```

taj-p:
We chatted in our 1:1 and think this PR (after removing the requirement for pixel snapping) is actually the best way to go because:

- It gives us the advantages of the blit rect PR
- It reduces complexity
- It doesn't support rotated images, but that's a bit unclear how to support AA properly
- It enables arbitrary paint to work 
- Is closer aligned with our current sparse strip strategy (i.e. it doesn't produce a sharp fork in strategies)

Laurenz will explore doing the AA in the shader to remove the pixel snapping requirement

cc @grebmeg

LaurenzV:
@taj-p Anti-aliasing is supported now. I did some measurements, unfortunately it does seem to have an impact on performance (from my experiments, this is not because of additional CPU work but instead because the fragment shader has to do more work): In a 4k scene with solid color rectangles of varying sizes, I got 29ms for 16750 rectangles with pixel-snapping, this now seems to drop to 41.5 afterwards. Will experiment if this can be improved, but I'm not sure. However, this is for solid colors, so I imagine if you are rendering real images (where we also do bilinear filtering), the difference shouldn't be as large anymore.

nicoburns:
> Anti-aliasing is supported now. I did some measurements, unfortunately it does seem to have an impact on performance (from my experiments, this is not because of additional CPU work but instead because the fragment shader has to do more work)

Would it be possible to have an even-faster-path for content that happens to already be pixel aligned? I imagine that would be quite a common case.

LaurenzV:
We do save the work for pixel aligned rectangles, but for some reason it didn't seem to yield much improvements. I haven't been able to find out why.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -287,11 +322,22 @@ var clip_input_texture: texture_2d<f32>;
 fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
     let x = u32(floor(in.tex_coord.x));
```
taj-p: We needn't calculate this for rects

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -287,11 +322,22 @@ var clip_input_texture: texture_2d<f32>;
...
+        // then multiplying them.
+        // For (maybe?) better performance, we calculate the x and y dimension in a single
+        // pass by packing everything into a vec2.
+        let rect_size = vec2<f32>(f32(in.dense_end & 0xFFFFu), f32(in.dense_end >> 16u));
+        let tc = in.tex_coord;
+        // + 0.5 and -0.5 since the fragment shader positions the coordinates in the center of the pixel.
+        let bottom_and_right = min(tc + 0.5, rect_size - frac.zw);
+        let top_and_left = max(tc - 0.5, frac.xy);
+        let a = clamp(bottom_and_right - top_and_left, vec2(0.0), vec2(1.0));
+        alpha = a.x * a.y;
```
taj-p, line 338: This calculation is occurring for every pixel in a rect even though it need only occur for the edges. We can significantly reduce per pixel work by outputting edge and fill rects instead within schedule.

Something like the below (very rough):

```rs
                FastStripCommand::Rect(r) => {
                    let sx0 = r.x0.floor();
                    let sy0 = r.y0.floor();
                    let sx1 = r.x1.ceil();
                    let sy1 = r.y1.ceil();

                    let base_x = sx0 as u16;
                    let base_y = sy0 as u16;
                    let snapped_w = (sx1 - sx0) as u16;
                    let snapped_h = (sy1 - sy0) as u16;

                    let q = |f: f32| -> u8 { (f * 255.0 + 0.5) as u8 };
                    let frac_l = q(r.x0 - sx0);
                    let frac_t = q(r.y0 - sy0);
                    let frac_r = q(sx1 - r.x1);
                    let frac_b = q(sy1 - r.y1);

                    let has_left = frac_l > 0;
                    let has_top = frac_t > 0;
                    let has_right = frac_r > 0;
                    let has_bottom = frac_b > 0;

                    let left_inset = u16::from(has_left);
                    let top_inset = u16::from(has_top);
                    let right_inset = u16::from(has_right);
                    let bottom_inset = u16::from(has_bottom);

                    let interior_w = snapped_w.saturating_sub(left_inset + right_inset);
                    let interior_h = snapped_h.saturating_sub(top_inset + bottom_inset);

                    // For small rects where the interior is empty, emit a single
                    // instance with full AA fracs (decomposition wouldn't help).
                    if interior_w == 0 || interior_h == 0 {
                        let frac = u32::from(frac_l)
                            | (u32::from(frac_t) << 8)
                            | (u32::from(frac_r) << 16)
                            | (u32::from(frac_b) << 24);
                        let (payload, paint_packed) =
                            Self::process_paint(&r.paint, scene, (base_x, base_y), paint_idxs);
                        draw.0.push(GpuStrip {
                            x: base_x,
                            y: base_y,
                            width: snapped_w,
                            dense_width_or_rect_height: snapped_h,
                            col_idx_or_rect_frac: frac,
                            payload,
                            paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
                        });
                    } else {
                        // Decompose into a pixel-aligned interior rect (rect_frac=0,
                        // skips AA in the fragment shader) plus thin 1px edge strips
                        // that handle anti-aliasing for fractional edges.
                        let interior_x = base_x + left_inset;
                        let interior_y = base_y + top_inset;

                        let (payload, paint_packed) = Self::process_paint(
                            &r.paint,
                            scene,
                            (interior_x, interior_y),
                            paint_idxs,
                        );
                        draw.0.push(GpuStrip {
                            x: interior_x,
                            y: interior_y,
                            width: interior_w,
                            dense_width_or_rect_height: interior_h,
                            col_idx_or_rect_frac: 0,
                            payload,
                            paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
                        });

                        if has_top {
                            let frac = u32::from(frac_l)
                                | (u32::from(frac_t) << 8)
                                | (u32::from(frac_r) << 16);
                            let (payload, paint_packed) = Self::process_paint(
                                &r.paint,
                                scene,
                                (base_x, base_y),
                                paint_idxs,
                            );
                            draw.0.push(GpuStrip {
                                x: base_x,
                                y: base_y,
                                width: snapped_w,
                                dense_width_or_rect_height: 1,
                                col_idx_or_rect_frac: frac,
                                payload,
                                paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
                            });
                        }

                        if has_bottom {
                            let bottom_y = base_y + snapped_h - 1;
                            let frac = u32::from(frac_l)
                                | (u32::from(frac_r) << 16)
                                | (u32::from(frac_b) << 24);
                            let (payload, paint_packed) = Self::process_paint(
                                &r.paint,
                                scene,
                                (base_x, bottom_y),
                                paint_idxs,
                            );
                            draw.0.push(GpuStrip {
                                x: base_x,
                                y: bottom_y,
                                width: snapped_w,
                                dense_width_or_rect_height: 1,
                                col_idx_or_rect_frac: frac,
                                payload,
                                paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
                            });
                        }

                        if has_left {
                            let frac = u32::from(frac_l);
                            let (payload, paint_packed) = Self::process_paint(
                                &r.paint,
                                scene,
                                (base_x, interior_y),
                                paint_idxs,
                            );
                            draw.0.push(GpuStrip {
                                x: base_x,
                                y: interior_y,
                                width: 1,
                                dense_width_or_rect_height: interior_h,
                                col_idx_or_rect_frac: frac,
                                payload,
                                paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
                            });
                        }

                        if has_right {
                            let right_x = base_x + snapped_w - 1;
                            let frac = u32::from(frac_r) << 16;
                            let (payload, paint_packed) = Self::process_paint(
                                &r.paint,
                                scene,
                                (right_x, interior_y),
                                paint_idxs,
                            );
                            draw.0.push(GpuStrip {
                                x: right_x,
                                y: interior_y,
                                width: 1,
                                dense_width_or_rect_height: interior_h,
                                col_idx_or_rect_frac: frac,
                                payload,
                                paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
                            });
                        }
                    }
                }
```

LaurenzV: Makes sense, will try this!

LaurenzV: Unfortunately, this doesn't seem to help much. :( I'm not sure why, but I think the additional time is somehow not coming from the additional workload itself, but from the additional branching or something.

For testing, I applied your code snippet and also changed the shader as follows:
```
if is_rect && in.rect_frac != 0u {
        let frac = unpack4x8unorm(in.rect_frac);
        // Calculate how much of the pixel is actually covered by the rect.
        // We do this by simply calculating the fractions in the x and y direction, and
        // then multiplying them.
        // For (maybe?) better performance, we calculate the x and y dimension in a single
        // pass by packing everything into a vec2.
        let rect_size = vec2<f32>(f32(in.dense_end & 0xFFFFu), f32(in.dense_end >> 16u));
        let tc = in.tex_coord;
        // + 0.5 and -0.5 since the fragment shader positions the coordinates in the center of the pixel.
        let bottom_and_right = min(tc + 0.5, rect_size - frac.zw);
        let top_and_left = max(tc - 0.5, frac.xy);
        let a = clamp(bottom_and_right - top_and_left, vec2(0.0), vec2(1.0));
        alpha = a.x * a.y;
        return vec4<f32>(1.0, 0.0, 0.0, alpha);
    } 
```

So I changed this to return a red color just to be sure that it's working correctly and the code path is only chosen for the border pixels, and it seems to work (but as mentioned, performance is slightly worse actually, I know get around 39ms frame time instead of 37ms):
image

However, if I alter that code path to just always return a red color for that:
```
if is_rect && in.rect_frac != 0u {
        return vec4<f32>(1.0, 0.0, 0.0, 0.5);
} 
```

I now get 33ms, even though the code path for the actually largest part (rendering the inner rectangle parts) remained unchanged. It's possible that the border part simply becomes a bottleneck because for blending, the GPU still needs to ensure that everything is executed in the correct order. But I'm not sure, I don't know much about the internals of GPUs. 😅 In any case, I don't think it's too big of a deal, it should still be very performant! Though maybe not quite as much as the original blit rect approach.

taj-p: The scene I used for testing was tiled non-overlapping solid rects. Even though it's not perfect (with room for improvement), that's a more generally representable scene than areas with dozens of overlaps. 

Are you able to try on that type of scene?

LaurenzV: Can try that 👍

laurenz-canva: Hmm I tried to run a scene with 60k tiled rects of size 5x5 but I'm anyways capping out at 120FPS. 😄 So doesn't seem like I can measure much here

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -167,37 +185,44 @@ struct Config {
...
-    @location(0) @interpolate(flat) paint: u32,
+    @location(0) @interpolate(flat) paint_and_rect_flag: u32,
     // Texture coordinates for the current fragment
     @location(1) tex_coord: vec2<f32>,
     // UV coordinates for the current fragment, used for image sampling
     @location(2) sample_xy: vec2<f32>,
-    // Ending x-position of the dense (alpha) region
+    // For normal strips: ending x-position of the dense (alpha) region.
+    // For rect strips: packed dimensions (width | height << 16).
     @location(3) @interpolate(flat) dense_end: u32,
```
grebmeg: [nit]: now that `dense_end` serves double duty (alpha region end for normal strips, packed `width | height << 16` for rect strips), would it be worth renaming to `dense_end_or_rect_size`? That would match the _or_ convention already used by `widths_or_rect_height` and `col_idx_or_rect_frac` in `StripInstance`.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -555,16 +563,59 @@ impl Scheduler {
...
+                    let (payload, paint_packed) =
+                        Self::process_paint(&r.paint, scene, (x, y), paint_idxs);
+
+                    // Determine the fractional offsets for anti-aliasing and quantize so it
+                    // fits into u8.
+                    let q = |f: f32| -> u8 { (f * 255.0 + 0.5) as u8 };
+                    let frac = u32::from(q(r.x0 - sx0))
+                        | (u32::from(q(r.y0 - sy0)) << 8)
+                        | (u32::from(q(sx1 - r.x1)) << 16)
+                        | (u32::from(q(sy1 - r.y1)) << 24);
```
grebmeg: [nit]: The inline q closure + manual byte-packing is essentially a hand-rolled `pack_unorm4x8` (the CPU-side counterpart of the `unpack4x8unorm` the shader uses to read this value). Could we extract a small helper like:

```rust
fn pack_unorm4x8(v: [f32; 4]) -> u32 {
    let q = |f: f32| -> u8 { (f * 255.0 + 0.5) as u8 };
    u32::from(q(v[0]))
        | (u32::from(q(v[1])) << 8)
        | (u32::from(q(v[2])) << 16)
        | (u32::from(q(v[3])) << 24)
}
```

Then the call site becomes:

```rust
let frac = pack_unorm4x8([r.x0 - sx0, r.y0 - sy0, sx1 - r.x1, sy1 - r.y1]);
```

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -228,51 +253,61 @@ fn vs_main(
...
         }
     }
 
-    // Regular texture coordinates for other render types
-    out.tex_coord = vec2<f32>(f32(instance.col_idx) + x * f32(width), y * f32(config.strip_height));
+    if is_rect {
+        out.tex_coord = vec2<f32>(x * f32(width), y * f32(height));
+    } else {
+        out.tex_coord = vec2<f32>(f32(instance.col_idx_or_rect_frac) + x * f32(width), y * f32(height));
+    }
```
grebmeg: This second `is_rect` branch can be folded into a branchless `select`, since the only difference is whether `col_idx_or_rect_frac` is added as an x-offset:

```wgsl
let col_offset = select(f32(instance.col_idx_or_rect_frac), 0.0, is_rect);
out.tex_coord = vec2<f32>(col_offset + x * f32(width), y * f32(height));
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -555,16 +563,59 @@ impl Scheduler {
...
+
+                    draw.0.push(GpuStrip {
+                        x,
+                        y,
+                        width,
+                        dense_width_or_rect_height: height,
+                        col_idx_or_rect_frac: frac,
+                        payload,
+                        paint_and_rect_flag: paint_packed | RECT_STRIP_FLAG,
+                    });
```
grebmeg: [nit]: the `Path` arm delegates to `generate_gpu_strips_for_fast_path`, but the `Rect` arm inlines ~30 lines of snapping/quantization/bit-packing. Worth extracting into a parallel helper for consistency and to make the fractional quantization math unit-testable?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: I think it'd be nice to avoid running the AA math for interior pixels, on low-tier devices with many rect geometries, I think the per-fragment cost could add up. That said, without proper profiling data showing an actual performance degradation, I think it's not worth adding the complexity of a 5-strip decomposition right now.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -434,40 +455,56 @@ impl Scene {
...
+
+        if self.clip_context.get().is_some() {
+            return false;
+        }
+
+        // We can't handle skewed rectangles.
+        let coeffs = self.render_state.transform.as_coeffs();
+        if coeffs[1].abs() > 1e-5 || coeffs[2].abs() > 1e-5 {
+            return false;
+        }
```
grebmeg, line 483: I'm thinking about the next approaches we can have:

- AABB rect + transparent extend mode (image fills only)
`encode_current_paint` already bakes the rotation into the paint transform, so paint sampling already handles rotation correctly. We could emit the AABB of the rotated rect as an axis-aligned fast-path rect and add a `Transparent` extend mode so that pixels outside the image bounds return zero.

- SDF-based AA for rotated rects
For the general case, we'd need the fragment shader to know the rotation so it can anti-alias against the actual rotated edges. Pack `sin`/`cos`, transform the fragment position back to rect-local space, and compute edge distances roughly somethign like the following:

```wgsl
let local_pos = inv_rotation * (screen_pos - rect_center);
let dx = clamp(min(local_pos.x, rect_w - local_pos.x) + 0.5, 0.0, 1.0);
let dy = clamp(min(local_pos.y, rect_h - local_pos.y) + 0.5, 0.0, 1.0);
alpha = dx * dy;
```

grebmeg (approved):
I really like this fast path approach, LGTM 👍

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Makes sense!

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Can do!

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Good point. 👍

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: Sounds good.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Will add a TODO.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Will add a link to this thread.

PR #1484, opened 2026-02-26
vello_common: Optimize wide tile resetting

laurenz-canva:
Stacked on top of #1479. This PR
- Adds a flag for tracking whether the `Wide` struct actually needs to be resetted. This makes sense because especially for the fast path, it's likely that we aren't actually using coarse rasterization. `reset` has been shown to take up a not insignificant amount of the frame time, so this seems worthy to do.
- It moves code for resetting a single wide tile into a dedicated function.


laurenz-canva:
Hmm yes, that might work, unclear whether it would be worth it though. 😄

PR #1485, opened 2026-02-27
vello_common: Optimize layer pushing for filter layers without clip path

laurenz-canva:
Gives a good boost CPU-side on our filter demo scene:

Before:
image

After:
image

For vello_cpu, this obviously won't help much since the bottleneck lies in the actual filter application, but for vello_hybrid this will be helpful!


grebmeg (approved):
Nice on! 🚀

PR #1486, opened 2026-03-02
Refactor `RenderState` handling

laurenz-canva:
This PR:
- Moves `RenderState` into vello_common.
- Adds tint to the render state.
- Makes vello_cpu and vello_hybrid use the render state internally instead of duplicating all fields.

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -683,12 +683,17 @@ impl RenderContext {
         }
     }
 
-    /// Save the current rendering state.
+    /// Take current rendering state and reset the existing state to its default.
     pub fn take_current_state(&mut self) -> RenderState {
+        core::mem::take(&mut self.state)
+    }
+
+    /// Save a copy of the current rendering state.
+    pub fn save_current_state(&mut self) -> RenderState {
         self.state.clone()
     }
 
-    /// Restore the saved rendering state.
+    /// Restore rendering state.
     pub fn restore_state(&mut self, state: RenderState) {
         self.state = state;
     }


laurenz-canva:
Actually, we could probably change it so that each render context itself stores a `RenderState`, so we don't need to duplicate the fields. Will try that.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -729,24 +666,12 @@ impl Scene {
...
 
     /// Restore rendering state.
     pub fn restore_state(&mut self, state: RenderState) {
-        self.paint = state.paint;
-        self.paint_transform = state.paint_transform;
-        self.stroke = state.stroke;
-        self.transform = state.transform;
-        self.fill_rule = state.fill_rule;
-        self.blend_mode = state.blend_mode;
+        self.render_state = state;
```
grebmeg, line 686: `paint_visible` is not recalculated here after restoring the paint. Since `paint_visible` gates all draw operations, a stale value after restore could either skip draws that should happen or dispatch draws that should be skipped. This was a pre-existing bug, but since we're refactoring this area, maybe worth fixing?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -729,24 +666,12 @@ impl Scene {
...
     pub fn take_current_state(&mut self) -> RenderState {
-        RenderState {
-            paint: self.paint.clone(),
-            paint_transform: self.paint_transform,
-            transform: self.transform,
-            fill_rule: self.fill_rule,
-            blend_mode: self.blend_mode,
-            stroke: core::mem::take(&mut self.stroke),
-        }
+        self.render_state.clone()
```
grebmeg, line 681: This switched from `core::mem::take` to `clone()`. For most fields that's free (Copy types), but `Stroke::dash_pattern` is a `SmallVec<[f64; 4]>` which would allocate on clone if there are `>4` dashes. Probably doesn't matter in practice, but what do you think about something like:

 - Rename this to `save_current_state` to reflect the clone semantics (pairs nicely with restore_state).
 - Keep a `take_current_state` variant that does `core::mem::take(&mut self.state)` for callers that want zero-cost move semantics and don't mind the state being at defaults until restored.

PR #1487, opened 2026-03-02
vello_hybrid: Better support for rendering to RGBA8 textures on targets without native RGBA8

laurenz-canva:
I haven't tested this and unfortunately cannot backport this to my filter branch because they diverge too much, but I think this should work.

Note also that the current `mem::replace` things we have in WebGL don't seem ideal, but this can be figured out in a future PR.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -34,7 +34,7 @@ use crate::{
         },
     },
     scene::Scene,
-    schedule::{LoadOp, RendererBackend, Scheduler, SchedulerState},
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -175,7 +177,8 @@ impl WebGlRenderer {
             "Render size must match drawing buffer size"
         );
 
-        self.render_scene(scene, render_size, true)?;
+        let view_fb = &self.programs.resources.view_framebuffer.clone();
```
LaurenzV, line 180: There is no cost associated with cloning here, right?

LaurenzV:
This is kinda part of the soon-to-arrive filter PR anyway, so will close.

PR #1488, opened 2026-03-02
vello_cpu: Fix off-by-one error in gaussian blur decimation filter

LaurenzV:
Let's take `upscale_x`In the very first iteration, `x` will be set to `src_width - 1`. Therefore, p2 will point to `src_width - 2`. But right now, p1 would point to `src_width` and p0 to `src_width + 1`. Therefore, `src_width - 1` is actually never sampled. I think this should be the right fix.

The same applies to the `upscale_y` function.

See also the `issue_filter_canvas_boundaries` test case, the bottom edge now looks symmetrical to the top edge. The right still looks different to the left one, so I'm still leaving the TODO comment there.

### DIFF
--- a/sparse_strips/vello_cpu/src/filter/gaussian_blur.rs
+++ b/sparse_strips/vello_cpu/src/filter/gaussian_blur.rs
@@ -1093,7 +1093,7 @@ mod tests {
         //   (0,1)=20   (1,1)=60   (2,1)=100  (3,1)=140
         let mut pixmap = pixmap_from_red(4, 2, &[&[0, 40, 80, 120], &[20, 60, 100, 140]]);
 
-        let dst_width = 4u16.div_ceil(2);
+        let dst_width = 4_u16.div_ceil(2);
         downscale_x(&mut pixmap, 4, 2, dst_width, EdgeMode::Duplicate);
 
         // Row 0: p[-1]=0,p[0]=0,p[1]=40,p[2]=80    → (0+0*3+40*3+80+4)>>3       = 25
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/filter/gaussian_blur.rs
```diff
@@ -386,8 +386,8 @@ fn upscale_y(src: &mut Pixmap, src_width: u16, src_height: u16, edge_mode: EdgeM
     for x in 0..src_width {
         // Maintain sliding window of three pixels: prev, current, next
         // This allows us to compute both output pixels that depend on current pixel y
-        let mut p0 = sample_y(src, x, src_height as i32 + 1, src_height, edge_mode);
-        let mut p1 = sample_y(src, x, src_height as i32, src_height, edge_mode);
+        let mut p0 = sample_y(src, x, src_height as i32, src_height, edge_mode);
+        let mut p1 = sample_y(src, x, src_height as i32 - 1, src_height, edge_mode);
```
grebmeg, line 392: Would it be worth adding a test that asserts actual pixel values after upscale with a non-uniform input? The existing tests here all use uniform data, so they wouldn't catch an off-by-one like the one this PR fixes. Something like a 3×1 pixmap with [100, 200, 50] that checks the 6 output pixels against the expected 0.75/0.25 interpolation. And similar to downscale.

LaurenzV:
> One thing I noticed while reviewing: the remaining left-right asymmetry mentioned in the TODO might be caused by downscale_y in downscale fucntion receiving src_width instead of dst_width. After downscale_x compacts the data into 0..dst_width columns, downscale_y still iterates over the full original width, processing stale data in the right-side columns. That stale data can then leak into the upscale pass via edge-mode sampling. Changing src_width to dst_width on that line might fix the right-edge asymmetry.

It doesn't seem to fix it but it also doesn't cause new regressions, so I'll include it!

PR #1489, opened 2026-03-02
vello_hybrid: Add support for adding padding around images

LaurenzV:

- Right now, when rendering images with bilinear sampling, we have to "emulate" this instead of using GPU-native filtering because 1) we only support Pad/Repeat/Reflect extend modes, but no "transparent pixel" extend mode and 2) there might be other pixel data at the border. By having this padding, we can in the future switch to GPU-native bilinear filtering by always allocating a border of 1 (I believe something similar is useful for glyph caching) and adding a new extend mode.
- For filter effects, I very similarly want to have the possibility to just assume transparent pixels outside (although in this case, the necessary padding iwll often be larger than 1), as it allows me to once again just use native bilinear filtering in all cases instead of having to make case distinctions.

This PR lays the foundational work for both use cases. It currently doesn't actually use this padding in the shader, but it can be used in the future. For now, just knowing that a number of transparent pixels exists around the image is enough for me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3977,7 +3977,6 @@ version = "0.0.6"
 dependencies = [
  "bytemuck",
  "fearless_simd",
- "guillotiere",
  "hashbrown 0.16.1",
  "libm",
  "log",
@@ -4059,7 +4058,6 @@ name = "vello_hybrid"
 version = "0.0.6"
 dependencies = [
  "bytemuck",
- "guillotiere",
  "hashbrown 0.16.1",
  "js-sys",
  "log",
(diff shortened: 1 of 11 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -418,14 +428,15 @@ impl WebGlRenderer {
...
                 [
-                    image_resource.offset[0] as u32,
-                    image_resource.offset[1] as u32,
+                    image_resource.offset[0] as u32 - padding,
+                    image_resource.offset[1] as u32 - padding,
                 ],
-                image_resource.width as u32,
-                image_resource.height as u32,
+                image_resource.width as u32 + padding * 2,
+                image_resource.height as u32 + padding * 2,
```
LaurenzV, line 439: I think in theory we don't need to clear the padding, but safe is safe.

grebmeg (approved):

PR #1490, opened 2026-03-03
Try running tests in release mode

laurenz-canva:


LaurenzV:
It seems like the slow compile times offset the improvements in test runtime 😦

PR #1492, opened 2026-03-05
Update `wgpu` to v28.

xStrom:
Wgpu v28 came out in Dec 18 2025, so about time for us to start using it too. Masonry in particular needs it for [an annoying macOS jitter bug](https://github.com/linebender/xilem/pull/1551).

Needed to also bump the MSRV to satisfy the `wgpu` demand of 1.92.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ For now, prefer updating the package-level readmes, e.g. vello/README.md.
 [![Linebender Zulip](https://img.shields.io/badge/Linebender-%23vello-blue?logo=Zulip)](https://xi.zulipchat.com/#narrow/channel/197075-vello)
 [![dependency status](https://deps.rs/repo/github/linebender/vello/status.svg)](https://deps.rs/repo/github/linebender/vello)
 [![Apache 2.0 or MIT license.](https://img.shields.io/badge/license-Apache--2.0_OR_MIT-blue.svg)](#license)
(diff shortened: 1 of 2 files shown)


vello/src/wgpu_engine.rs
```diff
@@ -362,7 +362,7 @@ impl WgpuEngine {
             },
             depth_stencil: None,
             multisample: wgpu::MultisampleState::default(),
-            multiview: None,
+            multiview_mask: None,
             cache: self.pipeline_cache.as_ref(),
```
PoignardAzur, line 366: (Non-blocking) I'm wondering, for structs like these where we're only changing default fields, could we skip the diff in the future by using the `..Default::default()` struct update syntax?

vello_shaders/src/compile/mod.rs
```diff
@@ -85,12 +85,9 @@ pub struct ShaderInfo {
...
-    #[cfg_attr(
-        target_pointer_width = "64",
-        allow(
-            clippy::result_large_err,
-            reason = "Deferred: This is a cold code path."
-        )
+    #[allow(
+        clippy::result_large_err,
+        reason = "Deferred: This is a cold code path."
     )]
```
PoignardAzur, line 91: (Non-blocking) At this point it would probably be more concise to box the error case.

PR #1493, opened 2026-03-05
vello_hybrid: Faster bilinear image rendering

laurenz-canva:
Now that we have the ability to add image padding, we could add a new `draw_image` method that uses the GPU-native bilinear sampler instead of our custom bilinear interpolation method. From my very rough experiment with a single scene of 1k images, this seems to give a pretty good speed boost.

A part of the speed boost also comes from the fact that we need to do less work in the fragment shader itself.

# Zoomed out

https://github.com/user-attachments/assets/5bed74aa-6011-4893-8ae7-f9296766f914

We roughly get the following frame times for each mode:
**Low**: 16.5ms
**Medium (manual bilinear)**: 21.5ms
**Medium (GPU-native bilinear)**: 12.2ms
**High**: 39ms

# Zoomed in

https://github.com/user-attachments/assets/d829c19c-d542-444e-b830-74cdf7974480

**Low**: 27ms
**Medium (manual bilinear)**: 35ms
**Medium (GPU-native bilinear)**: 19.2ms
**High**: 60ms

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -20,7 +20,7 @@
 only break in edge cases, and some of them are also only related to conversions from f64 to f32."
 )]
 
-use crate::render::common::IMAGE_PADDING;
+use crate::render::common::{IMAGE_PADDING, IMAGE_QUALITY_GPU_FAST_PATH};
 use crate::{
     GpuStrip, RenderError, RenderSettings, RenderSize,
     filter::{FilterContext, FilterInstanceData, FilterPassState, FilterPassTarget},
@@ -583,8 +583,16 @@ impl WebGlRenderer {
         let transform = image.transform.as_coeffs().map(|x| x as f32);
         let image_size = pack_image_size(image_resource.width, image_resource.height);
         let image_offset = pack_image_offset(image_resource.offset[0], image_resource.offset[1]);
+
+        // See the comment in the wgpu backend.
+        let quality = if image.use_gpu_fast_path() {
+            IMAGE_QUALITY_GPU_FAST_PATH
+        } else {
+            image.sampler.quality as u32
+        };
+
         let image_params = pack_image_params(
-            image.sampler.quality as u32,
+            quality,
             image.sampler.x_extend as u32,
             image.sampler.y_extend as u32,
             image_resource.atlas_id.as_u32(),
@@ -717,8 +725,10 @@ struct StripUniforms {
     alphas_texture: WebGlUniformLocation,
     /// Clip input texture location.
     clip_input_texture: WebGlUniformLocation,
-    /// Atlas texture location.
-    atlas_texture_array: WebGlUniformLocation,
+    /// Atlas texture location for fragment shader.
+    atlas_texture_array_fs: WebGlUniformLocation,
+    /// Atlas texture location for vertex shader.
+    atlas_texture_array_vs: WebGlUniformLocation,
     /// Encoded paints texture location for fragment shader.
     encoded_paints_texture_fs: WebGlUniformLocation,
     /// Encoded paints texture location for vertex shader.
@@ -1671,7 +1681,8 @@ fn get_strip_uniforms(gl: &WebGl2RenderingContext, program: &WebGlProgram) -> St
     // Get texture uniform locations.
     let alphas_texture_name = render_strips::fragment::ALPHAS_TEXTURE;
     let clip_input_texture_name = render_strips::fragment::CLIP_INPUT_TEXTURE;
-    let atlas_texture_array_name = render_strips::fragment::ATLAS_TEXTURE_ARRAY;
+    let atlas_texture_array_fs_name = render_strips::fragment::ATLAS_TEXTURE_ARRAY;
+    let atlas_texture_array_vs_name = render_strips::vertex::ATLAS_TEXTURE_ARRAY;
     let encoded_paints_texture_fs_name = render_strips::fragment::ENCODED_PAINTS_TEXTURE;
     let encoded_paints_texture_vs_name = render_strips::vertex::ENCODED_PAINTS_TEXTURE;
     let gradient_texture_name = render_strips::fragment::GRADIENT_TEXTURE;
@@ -1685,8 +1696,11 @@ fn get_strip_uniforms(gl: &WebGl2RenderingContext, program: &WebGlProgram) -> St
         clip_input_texture: gl
             .get_uniform_location(program, clip_input_texture_name)
             .unwrap(),
-        atlas_texture_array: gl
-            .get_uniform_location(program, atlas_texture_array_name)
+        atlas_texture_array_fs: gl
+            .get_uniform_location(program, atlas_texture_array_fs_name)
+            .unwrap(),
+        atlas_texture_array_vs: gl
+            .get_uniform_location(program, atlas_texture_array_vs_name)
             .unwrap(),
         encoded_paints_texture_fs: gl
             .get_uniform_location(program, encoded_paints_texture_fs_name)
@@ -1823,10 +1837,37 @@ fn create_texture(gl: &WebGl2RenderingContext) -> WebGlTexture {
     create_texture_inner(gl, WebGl2RenderingContext::TEXTURE_2D)
 }
 
-/// Create a texture array with nearest neighbor sampling and
-/// clamp-to-edge wrapping.
+/// Create a texture array with bilinear sampling and clamp-to-edge wrapping.
 fn create_texture_array(gl: &WebGl2RenderingContext) -> WebGlTexture {
-    create_texture_inner(gl, WebGl2RenderingContext::TEXTURE_2D_ARRAY)
+    let texture = gl.create_texture().unwrap();
+    gl.active_texture(WebGl2RenderingContext::TEXTURE0);
+    gl.bind_texture(WebGl2RenderingContext::TEXTURE_2D_ARRAY, Some(&texture));
+    gl.tex_parameteri(
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_MIN_FILTER,
+        WebGl2RenderingContext::LINEAR as i32,
+    );
+    gl.tex_parameteri(
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_MAG_FILTER,
+        WebGl2RenderingContext::LINEAR as i32,
+    );
+    gl.tex_parameteri(
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_WRAP_S,
+        WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
+    );
+    gl.tex_parameteri(
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_WRAP_T,
+        WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
+    );
+    gl.tex_parameteri(
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_MAX_LEVEL,
+        0,
+    );
+    texture
 }
 
 fn create_texture_inner(gl: &WebGl2RenderingContext, target: u32) -> WebGlTexture {
@@ -2260,8 +2301,14 @@ impl WebGlRendererContext<'_> {
             WebGl2RenderingContext::TEXTURE_2D_ARRAY,
             Some(&self.programs.resources.atlas_texture_array.texture),
         );
-        self.gl
-            .uniform1i(Some(&self.programs.strip_uniforms.atlas_texture_array), 2);
+        self.gl.uniform1i(
+            Some(&self.programs.strip_uniforms.atlas_texture_array_fs),
+            2,
+        );
+        self.gl.uniform1i(
+            Some(&self.programs.strip_uniforms.atlas_texture_array_vs),
+            2,
+        );
 
         // Bind encoded paints texture for image metadata
         self.gl.active_texture(WebGl2RenderingContext::TEXTURE3);
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -18,7 +18,7 @@
 only break in edge cases, and some of them are also only related to conversions from f64 to f32."
 )]
 
-use crate::render::common::IMAGE_PADDING;
+use crate::render::common::{IMAGE_PADDING, IMAGE_QUALITY_GPU_FAST_PATH};
 use crate::{
     GpuStrip, RenderError, RenderSettings, RenderSize,
     filter::{FilterContext, FilterInstanceData, FilterPassState, FilterPassTarget},
@@ -628,8 +628,19 @@ impl Renderer {
         let transform = image.transform.as_coeffs().map(|x| x as f32);
         let image_size = pack_image_size(image_resource.width, image_resource.height);
         let image_offset = pack_image_offset(image_resource.offset[0], image_resource.offset[1]);
+
+        // Values 0-2 represent the peniko `ImageQuality` variants, value 3 (IMAGE_QUALITY_GPU_BILINEAR)
+        // stands for "GPU-native bilinear sampling with transparent padding".
+        // Meaning that if quality is 3, the extend modes will be ignored and are instead used
+        // to distinguish between bilinear sampling and nearest-neighbor sampling.
+        let quality = if image.use_gpu_fast_path() {
+            IMAGE_QUALITY_GPU_FAST_PATH
+        } else {
+            image.sampler.quality as u32
+        };
+
         let image_params = pack_image_params(
-            image.sampler.quality as u32,
+            quality,
             image.sampler.x_extend as u32,
             image.sampler.y_extend as u32,
             image_resource.atlas_id.as_u32(),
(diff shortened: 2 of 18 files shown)


sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -376,52 +380,61 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
-                    encoded_image.extend_modes,
-                    encoded_image.image_padding,
-                );
-            } else if encoded_image.quality == IMAGE_QUALITY_MEDIUM {
-                let final_xy = image_offset + extended_xy - vec2(0.5);
-                sample_color = bilinear_sample(
+
+            if encoded_image.quality == IMAGE_QUALITY_GPU_BILINEAR {
+                let atlas_dims = vec2<f32>(textureDimensions(atlas_texture_array));
+                let sample_xy = (in.sample_xy) / atlas_dims;
```
grebmeg: The atlas dimensions are constant across an entire frame. The normalization `in.sample_xy / atlas_dims` could be done in the vertex shader instead, which would eliminate the per-fragment division and the `textureDimensions` call entirely. You'd compute the normalized uv once per vertex and pass as `in.sample_xy` and let the rasterizer interpolate which is exactly what `textureSample` expects.

grebmeg:
```suggestion
let atlas_dims = vec2<f32>(textureDimensions(atlas_texture_array));
let sample_xy = in.sample_xy / atlas_dims;
```

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -462,6 +463,41 @@ impl Scene {
...
+    ///
+    /// If you want more room for customization, you can also paint an image by setting
+    /// an image paint via `set_paint` and specifying parameters like extend and filtering quality
+    /// manually. However, if you only need bilinear filtering, using this method is recommended
+    /// as it will be much faster.
+    ///
+    /// This method will respect the current affine transformation in place (set via [`Scene::set_transform`],
+    /// but is not affected by the current paint transform (set via [`Scene::set_paint_transform`] in place.
+    pub fn draw_image(&mut self, image: ImageSource, rect: &Rect) {
+        let state = self.save_current_state();
```
grebmeg: Would it be possible to avoid saving/storing state here and instead allow the user to control that? It would be useful in cases where many `draw_image` calls happen in sequence.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -505,8 +504,18 @@ impl Renderer {
         let transform = image_transform.as_coeffs().map(|x| x as f32);
         let image_size = pack_image_size(image_resource.width, image_resource.height);
         let image_offset = pack_image_offset(image_resource.offset[0], image_resource.offset[1]);
+
+        // Values 0-2 represent the peniko `ImageQuality` variants, value 3
+        // stands for "GPU-native bilinear sampling with transparent padding".
+        // Meaning that if quality is 3, the extend modes will be ignored as well.
+        let quality = if image.custom & 1 == 1 {
+            3
```
grebmeg: What about adding a named constant?

```rust
pub(crate) const IMAGE_QUALITY_GPU_BILINEAR: u32 = 3;
```

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -505,8 +504,18 @@ impl Renderer {
         let transform = image_transform.as_coeffs().map(|x| x as f32);
         let image_size = pack_image_size(image_resource.width, image_resource.height);
         let image_offset = pack_image_offset(image_resource.offset[0], image_resource.offset[1]);
+
+        // Values 0-2 represent the peniko `ImageQuality` variants, value 3
+        // stands for "GPU-native bilinear sampling with transparent padding".
+        // Meaning that if quality is 3, the extend modes will be ignored as well.
+        let quality = if image.custom & 1 == 1 {
```
grebmeg: What about having something like this instead?

```rust 
impl EncodedImage {
    const GPU_BILINEAR: u8 = 1 << 0;

    pub fn use_gpu_bilinear(&self) -> bool {
        self.custom & Self::GPU_BILINEAR != 0
    }
}
```

Then the line would look like this:
```rust
let quality = if image.use_gpu_bilinear() {
    // ...
}
```

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -462,6 +463,41 @@ impl Scene {
...
+    /// Draw an image using bilinear filtering.
+    ///
+    /// If you want more room for customization, you can also paint an image by setting
+    /// an image paint via `set_paint` and specifying parameters like extend and filtering quality
+    /// manually. However, if you only need bilinear filtering, using this method is recommended
+    /// as it will be much faster.
+    ///
+    /// This method will respect the current affine transformation in place (set via [`Scene::set_transform`],
+    /// but is not affected by the current paint transform (set via [`Scene::set_paint_transform`] in place.
+    pub fn draw_image(&mut self, image: ImageSource, rect: &Rect) {
```
grebmeg: What about keeping the existing image rendering API as it is today, where you use `set_paint_transform`, `set_paint`, and `fill_rect`?

And then maybe requiring users to opt in to the new, faster path, for example by setting a native sampling flag? Then, in `encode_current_paint`, `custom` could be set as needed before passing it into `encode_into`. That flag would have no effect for `vello_cpu`, so we could remove `hybrid_ref` and use the same tests for all our renderers.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1514,10 +1521,37 @@ fn create_texture(gl: &WebGl2RenderingContext) -> WebGlTexture {
...
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_WRAP_T,
+        WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
+    );
+    gl.tex_parameteri(
+        WebGl2RenderingContext::TEXTURE_2D_ARRAY,
+        WebGl2RenderingContext::TEXTURE_MAX_LEVEL,
+        0,
+    );
+    texture
```
grebmeg, line 1870: Have you had a chance to compare performance in webgl?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Good call, will make that change!

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: I will change it so that we only store the paint transform and paint. I think this is not something that should be upon the user to do though, since I think it's unexpected that `draw_image` would actually change the render state. Let me know if that works for you.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: > And then maybe requiring users to opt in to the new, faster path, for example by setting a native sampling flag?

I think the problem with this is that it should be possible to still use the old image rendering mode (if you want to draw normal image patterns), even if you add a flag for that. I guess you could make the flag configurable with a set/unset method, but I'm not sure how much better this is either. 😄 I think it makes sense to add this mode to vello_cpu as well, though, so that we have feature parity. But not sure if it should happen in this PR.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Will investigate that.

LaurenzV: It doesn't seem to be as straight-forward to get reliable measurements in WebGL, but I'm seeing similar improvements (33ms vs 22ms).

https://github.com/user-attachments/assets/4a3962e8-ea8f-4252-a3a2-56e493cac776

LaurenzV:
@grebmeg If I try this fast path for `ImageQuality::Low` I'm also seeing a drop (from about 19ms to 15ms), so it might be worth doing if we need it for glyph caching. But as mentioned, we probably need to figure out a better API story first 😬 

EDIT: See below, I've implemented this now.

LaurenzV:
@grebmeg I've split this into two commits now, the second commit also adds support for a fast-path for nearest-neighbor sampling. If we don't want that, I can revert the last commit.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -462,6 +463,49 @@ impl Scene {
...
+                alpha: 1.0,
+            },
+        });
+
+        let paint_idx = self.encoded_paints.len();
+        self.fill_rect(rect);
+
+        if let Some(EncodedPaint::Image(img)) = self.encoded_paints.get_mut(paint_idx) {
+            img.set_use_gpu_fast_path();
+        }
```
taj-p: Sorry - just skimming through this PR to understand how it works from an API perspective.

Is this an invariant? Shouldn't this _always_ be `Some(EncodedPaint::Image(img))`? If so, we should `.unwrap` rather than suffer a silent bug

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -462,6 +463,49 @@ impl Scene {
...
+    ///
+    /// If you want more room for customization, you can also paint an image by setting
+    /// an image paint via `set_paint` and specifying parameters like extend and filtering quality
+    /// manually. However, if you only want to draw a simple image without bicubic filtering or
+    /// any special extend mode, it is recommended to use this method as it leverages GPU-native
+    /// capabilities and is therefore significantly faster.
+    ///
+    /// This method will respect the current affine transformation in place (set via [`Scene::set_transform`],
+    /// but is not affected by the current paint transform (set via [`Scene::set_paint_transform`] in place.
+    pub fn draw_image(&mut self, image: ImageSource, rect: &Rect, bilinear: bool) {
```
taj-p, line 505: For my information, why didn't we decide to apply this fast path in `fill_rect` where `bilinear` is passed in as a `RenderSettings` or `set_*` configuration?

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Yes, it should always be. Can change.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: The API can still change, but the problem with passing it to `RenderSettings` is that it means you cannot draw normal pattern images with extend repeat/reflect/pad anymore. We could add a setter for this so that it can change at runtime, but personally it just seems worse to me from an API perspective. I think a dedicated `draw_image` method expresses the intent clearer.

harley-canva:
Using texture filtering hardware I think is the right choice, but I have a few overall concerns with this. I question the value of creating a fast branch, when we really always want to be fast. Leveraging the hardware should be the norm and not the exception that requires a flag or special configuration.

`clamp`/`fract`/`abs`/`floor` should be all we need to get extend/repeat/reflect working with texture filtering. There will be discontinuities at the edge texel, but I think that's an acceptable trade-off. The key is to make sure we sample from the center of the texels at the edge of the texture, not at the edge of the texel, to avoid bilinear filtering with transparent padding. 

Are there any other reasons for not doing extend/repeat/reflect that I've overlooked here? Because my impression is that it's viable and that native texture filtering shoudn't be a branch, it should be the replacement for the manual bilinear approach that relies on 4 `textureLoad` calls. 

For the bicubic case, this can also be done more efficiently with bilinear texture filtering. In this case, this [GPU Gems article](https://developer.nvidia.com/gpugems/gpugems2/part-iii-high-quality-rendering/chapter-20-fast-third-order-texture-filtering) covers the approach. This is something where we may have more reservations about changing the quality, but I'd argue in favor of being high quality and fast across a range of hardware is better place to be. There's a lot of room to reduce the number of registers needed for the uber shader in the texture filtering code just by leveraging the existing sampling hardware.

LaurenzV:
> I question the value of creating a fast branch, when we really always want to be fast. Leveraging the hardware should be the norm and not the exception that requires a flag or special configuration.

I agree, but I think in our case, we would internally _always_ use this new API, because I from what I understood we don't have a need for Pad/Reflect/Repeat? From my understanding, we just need to have the ability to draw simple rectangular images, instead of needing to be able to fill arbitrary shapes with image patterns (which is the case where the Pad/Repeat/Reflect API starts to really become relevant). 

So basically, the existing functionality stays in-place simply because the current imaging model of Vello needs to have support for it (for example, SVG requires this), but in our case, we would always only use this new API (which does mean we need to add this to vello_cpu as well, to have consistency across the two).

> `clamp`/`fract`/`abs`/`floor` should be all we need to get extend/repeat/reflect working with texture filtering. There will be discontinuities at the edge texel, but I think that's an acceptable trade-off. 

I'm not sure I agree that it's an acceptable trade-off. Yes, if you are rendering things like larger images it probably wouldn't be that noticeable, but if you are using the API to render "real" patterns (for example grid lines or other cell patterns, which are often small in size), then I think this could cause some real artifacts, especially when scaling up a lot. For bicubic, this would be even worse because we can sample even further outside of the border. I guess we would need to test but I don't think other renderers like Skia will have artifacts there, so I would be a bit hesitant to go down this route. It would also make cross-testing against vello_cpu much harder, since our current tests trigger many such edge cases.

> The key is to make sure we sample from the center of the texels at the edge of the texture, not at the edge of the texel, to avoid bilinear filtering with transparent padding.

If we want to have the behaviour of clamping instead of transparent padding we could probably add that, but obviously we'd need to do the clamping as well in the fragment shader, while right now, we can always just use the raw coordinates. Might not be a big deal, but as we've seen, on low-tier devices, every operation counts. 😦 

> For the bicubic case, this can also be done more efficiently with bilinear texture filtering.

Yes, the approach in this PR could also be used to accelerate bicubic filtering. But as mentioned, if we try to use this for pad/reflect/repeat as well, this would make artifacts even worse since we will sample even more incorrect pixels at the edge. :(

But I agree it might be worth trying what the performance gains would be (on low-tier devices) if we always use the native sampling path.

harley-canva:
>    if you are using the API to render "real" patterns (for example grid lines or other cell patterns, which are often small in size), then I think this could cause some real artifacts, especially when scaling up a lot

I may be overlooking something, but I can only really see an issue for this when it comes to "repeat" which has a discontinuity. If I understand correctly, then "pad" is just clamping at the texel center for the edge texel, and "reflect" samples to the center of the edge texel and then samples inward towards the texture. If we want "repeat" to work correctly with texture samplers, it would be possible to solve it ideally in the padding of texture in the atlas, or in worst case, a branch at the discontinuity in the shader. I believe it's entirely possible to do this without artifacts, as long as we are precise with texel coordinates.

>    in our case, we would always only use this new API (which does mean we need to add this to vello_cpu as well, to have consistency across the two).

I think that's part of my concern with creating a branch. How can this work for vello_cpu? isn't the correct api for vello_cpu the one that already exists?

>    Might not be a big deal, but as we've seen, on low-tier devices, every operation counts.

I'm not particularly concerned about the cost of clamp/fract/abs/floor ops required to get texture filtering behaving consistently with the current approach, since they are very efficiently implemented in hardware. Switching from textureLoad to textureSample is a huge win in terms of prefetching and reduces a lot of ALU and registers required to do image painting. I would be quite surprised if we could measure a noticeable difference in performance with keeping the extend_mode function in place.

---

I decided to investigate implementing my suggestions, and ~~I realize my mistake with regards to "reflect", it does have sampling issues at the boundaries that I overlooked~~. EDIT: I got it working https://github.com/linebender/vello/pull/1517 

I haven't shifted my view from being concerned with having a fast branch as part of the API. I think this can still be simplified to using `textureSample` for `Pad` + `Reflect`, which can produce correct results, with a smaller set of changes to the shader and webgl/wgpu code and no change to the API. 

We can handle `Repeat` by handling it in the atlas, to be both performant and correct, although that's obviously a larger change. If we did that, we can then remove the branch, and minimize the number of registers needed for bilinear sampling.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -491,6 +492,50 @@ impl Scene {
...
+            },
+        });
+
+        let paint_idx = self.encoded_paints.borrow().len();
+        self.fill_rect(rect);
+
+        if let Some(EncodedPaint::Image(img)) = self.encoded_paints.borrow_mut().get_mut(paint_idx)
+        {
+            img.set_use_gpu_fast_path();
+        }
```
taj-p, line 533: This is an invariant right? This must _always_ be `Some`. In lieu of amending the implementation such that this if isn't required, I think we should `.unwrap` instead of silently failing (if for whatever reason `get_mut` returns None).

LaurenzV: Ah yes, sorry you already remarked that, but I haven't fixed it yet. 😅

LaurenzV:
> My preference is to not add a new API but I'm still learning the nuances of both approaches.

To maybe add another argument in favor of a new API, it seems like at least Skia also has a dedicated [`drawImage` method](https://api.skia.org/classSkCanvas.html#ab37570ba2c305284a79e88ad57a86313) instead of requiring the user to call `drawRect` with an image paint. So we wouldn't be alone in having a dedicated API for that. 😄 Though I admittedly haven't looked into how those are handled in the shader.

taj-p:
Hey @LaurenzV and @harley-canva ! cc @grebmeg 

I performed an investigation on this PR since it seems that we're a bit unsure about direction. I think the findings are valuable and will help direct efforts about where we're spending the time. See https://github.com/linebender/vello/pull/1547.

I'm wondering whether someone can validate that investigation because the data seems to suggest that the bottleneck isn't the extend_mode calculation, but:
- `textureDimension` querying;
- image quality branching; and
- per-pixel frag calculations (that should be moved to the vertex and passed as memory). 

When you put those conclusions together, if we're introducing this new API for performance only, it may be unnecessary. The above requires some review and discussion.

LaurenzV:
Closing since wildly outdated, but we will definitely revisit this soon.

PR #1494, opened 2026-03-05
vello_hybrid: Initial support for filter effects

laurenz-canva:
This is a first stab at implementing support for filter effects in vello_hybrid, ensuring that we have (virtually) complete feature parity with vello_cpu in that regard. In its current state with some caveats, both in terms of performance and feature-completeness, so I will talk about those first.

# Caveats

## Functionality
- I haven’t looked into it yet, but using non-default blend modes in filter layers doesn’t seem to quite work yet (see the failing test case). I will definitely address this, but probably not in this first PR, since it’s a rather uncommon case.
- I haven’t added support for edge modes. Not because it’s fundamentally not doable, but instead because I really wanted to use GPU-native bilinear sampling in this implementation, and this unfortunately only works if we are fine with sampling transparent pixels on the outside. So we can add this in the future, but we would have to add additional code to handle border pixels, which I want to avoid doing in this first PR.
- There are some robustness with the default image atlas settings (where the number of maximum atlases is bounded), since you can easily run out of space if you create dozens of filter layers. But that’s not a fundamental issue with the filter approach, just a configuration issue that we should probably tweak in the future.

## Performance
- For some reason, the wgpu WebGL backend is really slow on Chrome and Safari (not in Firefox). The native WebGL backend works fine, so I haven’t investigated this further.
- There is more room for improving the render times when rendering a single filter element. For example, from some initial experiments I gathered that introducing a `upscale/downscale 4x` filter pass would be worth it since we aren’t performing that many samples, so the reduction in render passes would give a nice boost. However, it results in larger divergences compared to vello_cpu, so I haven’t  pursued this further in this PR.
- Secondly, there is another optimization we can apply when rendering scenes with dozens of filter layers. Right now, each filter layer is rendered and filtered individually, which means that you can quickly end up with hundreds of render passes, which obviously is bad for performance. A potential improvement here is to batch the application of multiple filter effects (as long as the filter layers are independent from each other) in the same render pass. We can do this because we can allocate multiple intermediate scratch buffers in the filter image atlas, so there is no inherent reason why we have to schedule those sequentially. However, this is a bigger change and therefore also reserved for the future, if the need arises.

# Implementation

Fundamentally, this PR tries to copy the overall logic for the filters from vello_cpu. If there are any filter layers, instead of eagerly iterating over all wide tile comments, we iterate over the execution order of the render graph, and render intermediate filter layers into an intermediate texture. Then, we decompose the filters into a number of filter passes, which are then applied using a ping-pong scheme, until we have our final filtered layer, which is then stored in the normal image atlas. From then on, filter layers can be treated as any other image and filled when we hit the corresponding command in the execution order. That's it, it's nothing too crazy, but it definitely took a long while to arrive at this solution!

Below, you can see the PR in action and how it compares to vello_cpu:

Vello CPU:

https://github.com/user-attachments/assets/f86e710d-db9a-4562-918e-87c4026b0b71

Vello Hybrid:

https://github.com/user-attachments/assets/1791eb00-d59c-4b0e-9924-1801910f3338

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/filter.rs
+++ b/sparse_strips/vello_sparse_tests/tests/filter.rs
@@ -17,7 +17,7 @@ use vello_common::paint::Image;
 use vello_common::peniko::{
     BlendMode, Compose, Extend, Gradient, ImageQuality, ImageSampler, LinearGradientPosition, Mix,
 };
-use vello_cpu::color::palette::css::{BLUE, GREEN, RED, TRANSPARENT, YELLOW};
+use vello_cpu::color::palette::css::{BLUE, GREEN, RED, YELLOW};
 use vello_cpu::kurbo::Dashes;
 use vello_dev_macros::vello_test;
 


sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+        let lk = LinearKernel::new(&blur.kernel, blur.kernel_size);
+
+        Self {
+            header: pack_header_with_gaussian_params(
+                filter_type::GAUSSIAN_BLUR,
+                edge_mode_to_gpu(blur.edge_mode),
+                // Note that this could be exceeded in theory, but it would have to be a huge
+                // standard deviation! If it turns out to be a problem we can reserve additional
+                // bits for it in the future.
+                blur.n_decimations as u32,
```
grebmeg, line 281: Would it be worth adding an assert or clamping the value?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+        dest_image_cache: &ImageCache,
+        get_filter_atlas_size: impl Fn(u32) -> [u32; 2],
+        get_image_atlas_size: impl Fn() -> [u32; 2],
+    ) {
+        let s = &mut self.filter_pass_state.borrow_mut();
+        let state = s.deref_mut();
+        state.clear();
+
+        // These unwraps can only panic for filter layers without a bbox, but those are skipped
+        // anyway before even getting here, so unwrap should be safe here.
```
grebmeg, line 886: Could you make the invariant here more explicit? These `unwrap()` calls rely on the fact that every non-empty filter layer in the render graph has a corresponding entry in offsets / filter_textures / image_cache, which is established by `prepare`. A comment or a `debug_assert!` linking to that guarantee would make it much easier for future readers to verify this is safe, right now you have to trace through two other files to convince yourself these won't panic.

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+    /// Calculate the required height for the filter data texture.
+    /// Returns `None` if no filters are present.
+    pub(crate) fn required_filter_data_height(&self, max_texture_dimension_2d: u32) -> Option<u32> {
+        let required_texels = self.total_texels();
+
+        if required_texels == 0 {
+            return None;
+        }
+        let height = required_texels.div_ceil(max_texture_dimension_2d);
+        assert!(
```
grebmeg:
```suggestion
debug_assert!
```
Also, could this return `None`?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+    pub(crate) fn get_filter_data(&self, layer_id: &LayerId) -> Option<&GpuFilterData> {
+        let offset = self.offsets.get(layer_id)?;
+        let index = (*offset / GpuFilterData::SIZE_TEXELS) as usize;
+        self.filters.get(index)
+    }
+
+    pub(crate) fn serialize_to_buffer(&self, buffer: &mut [u8]) {
+        let src = bytemuck::cast_slice::<GpuFilterData, u8>(&self.filters);
+        buffer[..src.len()].copy_from_slice(src);
+    }
```
grebmeg, line 854: Nit: a `debug_assert!` documenting the size expectation would make this contract explicit and help catch buffer-sizing mistakes early:
```suggestion
    pub(crate) fn serialize_to_buffer(&self, buffer: &mut [u8]) {
        let src = bytemuck::cast_slice::<GpuFilterData, u8>(&self.filters);
        debug_assert!(buffer.len() >= src.len(), "filter data buffer too small: {} < {}", buffer.len(), src.len());
        buffer[..src.len()].copy_from_slice(src);
    }
```

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+    /// filter data texture.
+    pub(crate) filters: Vec<GpuFilterData>,
+    /// At what texel offset the filter data for the given layer ID is stored in the texture.
+    pub(crate) offsets: HashMap<LayerId, u32>,
+    /// Data for each filter layer.
+    pub(crate) filter_textures: HashMap<LayerId, FilterLayerData>,
+    /// Image cache for storing filter intermediate textures.
+    pub(crate) image_cache: ImageCache,
+    /// State used for constructing filter passes.
+    pub(crate) filter_pass_state: RefCell<FilterPassState>,
```
grebmeg: Would it be feasible to change `build_filter_passes` to take `&mut self`? If the renderer's borrow structure makes that tricky, an alternative would be to extract `FilterPassState` out of `FilterContext` and pass it alongside, which keeps both the borrow check static and the API clean?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+/// convolution.
+#[expect(clippy::cast_possible_truncation, reason = "safe in this case")]
+const FILTER_ATLAS_PADDING: u16 = MAX_KERNEL_SIZE as u16 / 2;
+
+// Note: Keep these variables and struct layouts in sync with `filters.wgsl`!
+
+// Since we store in RGBA32 texture.
+const BYTES_PER_TEXEL: usize = 16;
+const FILTER_SIZE_BYTES: usize = 48;
+const FILTER_SIZE_U32: usize = FILTER_SIZE_BYTES / 4;
```
grebmeg, line 65: What about adding compile-time assertions, something like the following?
```rust
const _: () = assert!(size_of::<GpuOffset>() == FILTER_SIZE_BYTES);
// and for others
```

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -452,6 +488,79 @@ impl Scheduler {
     }
 
     /// Schedule and render the scene.
+    pub(crate) fn do_scene<R: RendererBackend>(
```
grebmeg, line 494: Could you please add a comment noting that this is similar to `rasterize_with_filters` in `vello_cpu`, just so anyone reading the code is aware of it?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -828,68 +1011,100 @@ impl Scheduler {
...
+                        state,
+                        encoded_paints,
+                        alpha_fill,
                         paint_idxs,
+                        wide_tile_x,
+                        wide_tile_y,
+                        attrs,
                     );
+                }
+                Cmd::PushBuf(LayerKind::Filtered(child_layer_id), blend_into_dest) => {
```
grebmeg, line 1044: Similar to above comment: what about adding a cross-reference comment here to the vello_cpu counterpart `process_layer_tile` in `single_threaded.rs`. Both functions handle the same `Cmd::PushBuf(LayerKind::Filtered(...))` logic, but diverge in how they actually composite.

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+                    let scratch_1_atlas_id = self.image_cache.get(scratch_1).unwrap().atlas_id;
+
+                    // Second scratch buffer needs to live on a different texture than first scratch buffer.
+                    let scratch_2 = self.image_cache.allocate_excluding(
+                        width,
+                        height,
+                        FILTER_ATLAS_PADDING,
+                        Some(AtlasId(scratch_1_atlas_id.as_u32())),
+                    )?;
+                    Some([scratch_1, scratch_2])
```
grebmeg, line 774: The comment says `scratch_2` must differ from `scratch_1`, but it could still land on the same atlas as the initial texture. With 3 atlases and first-fit allocation this is plausible: initial → atlas 0, `scratch_1` → atlas 1, `scratch_2` → atlas 0.

As I understand it, currently this is safe because no pass simultaneously reads initial and writes `scratch_2` in the same render pass? But it's a fragile invariant, any future filter type that reads from initial mid-sequence could trigger a read/write hazard.

Maybe at minimum, we add a debug assertion?
```rust
let scratch_2_atlas_id = self.image_cache.get(scratch_2).unwrap().atlas_id;
debug_assert_ne!(
    scratch_2_atlas_id, initial_atlas_id,
    "scratch_2 must not share an atlas with the initial texture"
);
```

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+            filter_pass_state: RefCell::new(FilterPassState::default()),
+        }
+    }
+
+    /// Deallocates all filter textures from both the filter image cache and the image atlas cache,
+    /// then clears the filter context.
+    ///
+    /// Note that the client is responsible for clearing (with a transparent color) the existing
+    /// images in the atlas, if desired.
+    pub(crate) fn deallocate_all_and_clear_context(&mut self, image_atlas_cache: &mut ImageCache) {
```
grebmeg, line 679: Nit: this runs every frame. For scenes without filters (the common case), we can short-circuit:
```rust
if self.filters.is_empty() {
    return;
}
```

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+    /// Atlas index and region of the initial (unfiltered) image.
+    initial: (u32, [u32; 2]),
+    /// Atlas index and region of the final destination.
+    dest: (u32, IntRect),
+    /// Atlas index and region of each scratch buffers.
+    scratch: [(u32, IntRect); 2],
+    /// Texel offset into the filter data texture.
+    filter_data_offset: u32,
+    /// Full size of the original content region.
+    original_size: [u32; 2],
```
grebmeg: Nit: initial stores `(atlas_idx, offset)` and `(atlas_idx, size)` as `(u32, [u32; 2])` while dest and scratch use `(u32, IntRect)`. The mixed representations are a bit confusing. Maybe let's add a small named struct (somethign like `IntSize`, `IntOffset`) to unify these and `IntRect` internall as well, or at least a comment explaining why initial only needs an offset.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -828,68 +1011,100 @@ impl Scheduler {
...
-                    };
 
-                    draw.push(
-                        gpu_strip_builder
-                            .with_sparse(alpha_fill.width, col_idx)
-                            .paint(payload, paint),
-                    );
+                    match tile.cmds.get(cmd_idx + 1) {
+                        Some(Cmd::PushZeroClip(id)) if *id == *child_layer_id => {
+                            cmd_idx = filtered_ranges.full_range.end - 1;
```
grebmeg, line 1105: Could we leave an explanatory comment here, similar to the comment we have in `process_layer_tile`?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+    fn from(filter: GpuGaussianBlur) -> Self {
+        bytemuck::cast(filter)
+    }
+}
+
+impl From<GpuDropShadow> for GpuFilterData {
+    fn from(filter: GpuDropShadow) -> Self {
+        bytemuck::cast(filter)
+    }
+}
```
grebmeg: [nit]: Maybe something like the following would be more concise, but I’ll leave it to you to decide?

```rust
trait CastToFilterData: bytemuck::Pod {}

impl CastToFilterData for GpuOffset {}
impl CastToFilterData for GpuFlood {}
impl CastToFilterData for GpuGaussianBlur {}
impl CastToFilterData for GpuDropShadow {}

impl<T: CastToFilterData> From<T> for GpuFilterData {
    fn from(filter: T) -> Self {
        bytemuck::cast(filter)
    }
}
```

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1047 @@
...
+        &self.filter_passes
+    }
+
+    pub(crate) fn instances(&self) -> &[FilterInstanceData] {
+        &self.instances
+    }
+}
+
+/// A helper struct making it easier to schedule the rendering passes for filters.
+struct PassScheduler<'a> {
```
grebmeg: [nit]: `PassScheduler` is quite generic — this struct exclusively schedules the multi-pass blur pipeline (`DOWNSCALE` → `BLUR_H` → `BLUR_V` → `UPSCALE`), and its only callers are `GAUSSIAN_BLUR` and `DROP_SHADOW` (which is blur + composite). Its core method is even called emit_blur_sequence.

Something like `BlurPassScheduler` would better convey that this isn't a general-purpose pass scheduler for all filter types — single-pass filters like `OFFSET` and `FLOOD` are handled separately above and never touch this struct.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
```diff
@@ -0,0 +1,361 @@
...
+fn fs_main(in: FilterVertexOutput) -> @location(0) vec4<f32> {
+    let frag_coord = vec2<u32>(in.position.xy);
+    let rel_coord = vec2<f32>(frag_coord - in.dest_offset);
+
+    // See the comment in `vs_main`.
+    if rel_coord.x >= f32(in.dest_size.x) || rel_coord.y >= f32(in.dest_size.y) {
+        return vec4<f32>(0.0);
+    }
+
+    let data = load_filter_data(in.filter_offset);
```
grebmeg: [nit]: `PASS_DOWNSCALE`, `PASS_UPSCALE`, and `PASS_COPY` don't use data at all. Moving `load_filter_data` inside only the cases that need it would save 3 texture loads per fragment for those passes.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
```diff
@@ -0,0 +1,361 @@
...
+fn sample_original(in: FilterVertexOutput, rel_coord: vec2<f32>) -> vec4<f32> {
+    let src_coord = vec2<u32>(vec2<i32>(in.original_offset) + vec2<i32>(rel_coord));
+    return textureLoad(original_tex, src_coord, 0);
+}
+
+// Sample a pixel from the input texture.
+fn sample_input(in: FilterVertexOutput, rel_coord: vec2<f32>) -> vec4<f32> {
+    let src_coord = vec2<u32>(vec2<i32>(in.src_offset) + vec2<i32>(rel_coord));
+    return textureLoad(in_tex, src_coord, 0);
+}
```
grebmeg, line 193: `sample_input` and `sample_original` silently wrap to huge coordinates if `rel_coord` is negative (i32-to-u32 wrap). Currently all callers are safe, but it might be worth a comment on the contract (non-negative `rel_coord` required) to protect against future misuse.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1828,16 +2375,134 @@ impl RendererBackend for RendererContext<'_> {
...
+        let filter_passes = pass_state.filter_passes();
+        if filter_passes.is_empty() {
+            return;
+        }
+
+        let instances = pass_state.instances();
+        let instance_stride = size_of::<FilterInstanceData>() as u64;
+        let total_size = instances.len() as u64 * instance_stride;
+        self.programs.resources.filter_instance_buffer =
+            Programs::create_filter_instance_buffer(self.device, total_size);
```
grebmeg, line 2469: This creates a new GPU buffer on every `apply_filter` call — for scenes with N filter layers, that's N allocations per frame. What about reusing the buffer and only recreating when the required size exceeds the current capacity?

image

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1726,32 +2178,127 @@ struct RendererContext<'a> {
...
+                // 1. Account for the intermediate texture living at an offset within its atlas.
+                // 2. Account for the filter layer bbox not starting at (0, 0).
+                let strip_offset_x = resources.offset[0] as i32
+                    - (filter_textures.bbox.x0() * WideTile::WIDTH) as i32;
+                let strip_offset_y =
+                    resources.offset[1] as i32 - (filter_textures.bbox.y0() * Tile::HEIGHT) as i32;
+
+                let atlas_config_buffer =
+                    self.device
+                        .create_buffer_init(&wgpu::util::BufferInitDescriptor {
```
grebmeg, line 2278: This creates a new uniform buffer and bind group on every strip render pass targeting a filter layer. Since the config is small and the layout is fixed, what about pre-allocating a reusable buffer (with `COPY_DST`) and updating it via `queue.write_buffer`? The bind group can be cached in `GpuResources` and rebuilt only when the underlying textures change.

image

image

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1973,8 +2360,125 @@ impl RendererBackend for WebGlRendererContext<'_> {
...
+                FilterPassTarget::FilterAtlas(idx) => {
+                    let fb = &self.programs.resources.filter_atlas_framebuffers[*idx as usize];
+                    self.gl
+                        .bind_framebuffer(WebGl2RenderingContext::FRAMEBUFFER, Some(fb));
+                    self.gl
+                        .viewport(0, 0, filter_atlas_width as i32, filter_atlas_height as i32);
+                }
+                FilterPassTarget::MainAtlas(idx) => {
+                    let fb =
+                        main_atlas_fb.get_or_insert_with(|| self.gl.create_framebuffer().unwrap());
```
grebmeg: What about caching this framebuffer in `WebGlResources` (like `atlas_render_framebuffer`) to avoid create/delete on every frame. The FBO is reattached via `framebuffer_texture_layer` each pass anyway, so a persistent object should work fine.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1726,32 +2178,127 @@ struct RendererContext<'a> {
...
             return;
         }
         // TODO: We currently allocate a new strips buffer for each render pass. A more efficient
         // approach would be to re-use buffers or slices of a larger buffer.
         self.programs.upload_strips(self.device, self.queue, strips);
 
+        let (view, bind_group): (TextureView, BindGroup) = match target {
+            StripPassRenderTarget::Output(OutputTarget::FinalView) => (
+                self.view.clone(),
+                self.programs.resources.slot_bind_groups[2].clone(),
```
grebmeg: These `.clone()` calls are new on the strip render path and apply to every strip pass (not just filter passes). `TextureView` and `BindGroup` are `Arc`-backed, so each `clone` is an atomic increment. Would it be possible to restructure to borrow references instead?

sparse_strips/vello_common/src/filter/gaussian_blur.rs
```diff
@@ -158,9 +160,59 @@ pub fn compute_gaussian_kernel(std_deviation: f32) -> ([f32; MAX_KERNEL_SIZE], u
...
+
+    /// Apply a new downscale operation.
+    pub fn downscale(&mut self) -> (u16, u16) {
+        self.dim_stack.push((self.width, self.height));
+        self.width = self.width.div_ceil(2);
+        self.height = self.height.div_ceil(2);
+        (self.width, self.height)
+    }
+
+    /// Apply a new upsacle operation.
```
grebmeg:
```suggestion
/// Apply a new upscale operation.
```

sparse_strips/vello_common/src/image_cache.rs
```diff
@@ -67,21 +79,35 @@ impl ImageCache {
...
+        &mut self,
+        width: u32,
+        height: u32,
+        padding: u16,
+    ) -> Result<ImageId, AtlasError> {
+        self.allocate_excluding(width, height, padding, None)
+    }
+
+    /// Allocate an image in the cache, with optional transparency padding
+    /// and  optionally excluding a specific atlas.
```
grebmeg:
```suggestion
    /// and optionally excluding a specific atlas.
```

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1109,17 +1262,111 @@ pub(crate) fn blur_with_edge_mode(ctx: &mut impl Renderer, edge_mode: EdgeMode)
...
+    ctx.set_paint(BLUE);
+    ctx.fill_rect(&Rect::new(10.0, 60.0, 40.0, 90.0));
+
+    ctx.push_filter_layer(filter.clone());
+    ctx.set_paint(YELLOW);
+    ctx.fill_rect(&Rect::new(60.0, 60.0, 90.0, 90.0));
+    ctx.pop_layer();
+}
+
+// Doesn't seem to work in hybrid yet. :(
```
grebmeg, line 1358: Will this be resolved as a follow-up?

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -3,23 +3,28 @@
...
+use vello_common::peniko::{
+    BlendMode, Compose, Extend, Gradient, ImageQuality, ImageSampler, LinearGradientPosition, Mix,
+};
+use vello_cpu::color::palette::css::{BLUE, GREEN, RED, TRANSPARENT, YELLOW};
 use vello_cpu::kurbo::Dashes;
 use vello_dev_macros::vello_test;
 
 // TODO: We are purposefully using multiple of WideTile width/height here, because the implementation
 // currently works incorrectly if it's not the case. Once the issue as been fixed, we should update
 // this test to use normal dimensions.
```
grebmeg, line 26: Is this TODO still accurate?

grebmeg (commented):
This is incredible work! 🚀 Sorry for the delay in reviewing this PR. It’s a lot of meaningful and non-trivial code. I’m posting my comments now, please consider this the first round. After the feedback is addressed, I’d like to do one more round.

grebmeg:
Could we add a new example scene that includes different types of elements, such as a couple of rectangles, paths, gradients, some filters like drop shadows and blurs, etc.?

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: I already have a debug assertion in `pack_header_with_gaussian_params`. You'd need a stnadard deviation of 36636 to hit this, so I think this is save to ignore for our purpose, right?

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Will add comments for that.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: What do you mean could this return `None`? It can return `None` if `required_texels` is zero. Or do you mean it should return `None` if the assertion is hit?

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: Yes, this issue exists on both, CPU and GPU, and is unrelated to this PR.

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: Yes, can look into that. It only seems to be happening for non-default blending so it's probably not super urgent, but it would still be good to figure out what's going on here.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Makes sense!

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Hmm, however, let's say that a scene had filters but in the next frame it doesn't anymore, then we wouldn't clean up the resources from the last scene (only if we hit another frame in the future where filters are used again). I'm not sure if that would be desirable? Since it would keep using space in the image allocator.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
laurenz-canva: Good catch!

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Done.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Done.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: Added them.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: Done.

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: Copy-pasted comments from vello_cpu.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Applied!

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Good point, will try to simplify this.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: The intention is that we can reuse this in case there are other multi-pass filters we want to support in the future, but for now I can rename it.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
laurenz-canva: Makes sense.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Right, yes I should document this better, it's a bit more subtle. 

You are right that `initial image` and `scratch_2` can end up in the same atlas, but this is intentional. The reason why this works is that initial image atlas is (currently) only needed for the final drop shadow pass (where we composite the original content on top of the drop shadow). However, this final pass is always written to the final destination, which lives in the original image atlas, not the filter texture atlases. Depending on which scratch texture the contents currently live on, it can happen that the same texture is bound twice, but that shouldn't be a problem since it's just bound for reading, the write destination always lives elsewhere.

Will add some comments as well as an assertion (I think this is better than debug assertion since it will crash anyway in case this is ever violated).

sparse_strips/vello_hybrid/src/render/wgpu.rs
laurenz-canva: The problem is that this would require us to precompute the size for _all filters_ in the scene and then write all instances into the same buffer. I don't think we can reuse them across multiple `apply_filter` calls since the new results will overwrite the old ones (see my other comment below). It's certainly doable, but I think that's something that can be done as a follow-up. Will add a TODO comment.

sparse_strips/vello_hybrid/src/render/wgpu.rs
laurenz-canva: I think this is the same problem as for the filter instance buffer. If I understood correctly (and this also matches my experiments, since applying the filter instance buffer suggestion caused test failures), all buffers will be written before any of the render passes start. Meaning that if you write to the same buffer multiple times, the old values at the same location will be discarded. Is my understanding correct here? If so, I think we once again would have to write all into a single buffer and then pass the correct slice offset when we need it. 

Which is certainly doable, but I think it would be better in a follow-up. What do you think?

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: The first is not possible easily, I tried this before and as you mentioned I ran into borrowing issues. But I've taken the second approach now.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Added comments.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: I simplified this and hope it's clearer now.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
laurenz-canva: Added two notes!

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Renamed it.

sparse_strips/vello_hybrid/src/render/webgl.rs
laurenz-canva: Added caching for it.

sparse_strips/vello_hybrid/src/render/wgpu.rs
laurenz-canva: Changed it to using references whenever possible.

LaurenzV:
I added a `filter_elements` scene, is that what you had in mind? You can add/remove batches of 10 elements with +/-. Right now, it crashes pretty easily on my machine with 50+ elements because we run out of atlas space, but that's not some limitation of my implementation, but simply the fact taht the default image atlas size is 4096x4096 and we don't allow more than 8 texture layers.

image

sparse_strips/vello_common/src/filter/gaussian_blur.rs
```diff
@@ -158,9 +160,59 @@ pub fn compute_gaussian_kernel(std_deviation: f32) -> ([f32; MAX_KERNEL_SIZE], u
...
+    /// Apply a new downscale operation.
+    pub fn downscale(&mut self) -> (u16, u16) {
+        self.dim_stack.push((self.width, self.height));
+        self.width = self.width.div_ceil(2);
+        self.height = self.height.div_ceil(2);
+        (self.width, self.height)
+    }
+
+    /// Apply a new upscale operation.
+    pub fn upscale(&mut self) -> (u16, u16) {
```
grebmeg, line 207: Maybe it’s worth inlining these methods?

sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -209,8 +209,9 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
         || input_fn_name_str.contains("compose")
         || (input_fn_name_str.contains("blend")
             && !input_fn_name_str.contains("default_blending_only"))
-        || input_fn_name_str.contains("filter")
         || input_fn_name_str.contains("recording")
+        // TODO: Add test annotation instead of hard coding here
+        || input_fn_name_str.contains("filter_varying_depths_clips_and_compositions")
```
grebmeg, line 214: Could you please add the reason why it’s skipped here and near the test?

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1087,6 +1135,111 @@ fn issue_filter_canvas_boundaries(ctx: &mut impl Renderer) {
...
+    ctx.push_filter_layer(filter.clone());
+    ctx.set_paint(VIOLET);
+    ctx.fill_rect(&rect);
+    ctx.pop_layer();
+    ctx.pop_layer();
+    ctx.pop_layer();
+}
+
+#[vello_test(skip_multithreaded, no_ref)]
+fn filter_on_right_of_viewport(ctx: &mut impl Renderer) {
```
grebmeg, line 1231: Could you please add brief comments to this test explaining what we’re verifying and why `no_ref` is used?

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1109,17 +1262,112 @@ pub(crate) fn blur_with_edge_mode(ctx: &mut impl Renderer, edge_mode: EdgeMode)
...
+/// Verifies that content drawn before a filter layer is preserved correctly
+/// underneath the filtered result — in particular that fast-path strips
+/// (used for the background rect) still work alongside the filter path.
+fn filter_over_existing_content_inner<R: Renderer>(ctx: &mut R, draw_bg: impl FnOnce(&mut R)) {
+    ctx.set_paint(GREEN);
+    draw_bg(ctx);
+
+    let filter = Filter::from_primitive(FilterPrimitive::Offset { dx: 0.0, dy: 0.0 });
+    ctx.push_filter_layer(filter);
+    ctx.set_paint(TRANSPARENT);
```
grebmeg: What if the content isn’t actually preserved, but we just don’t notice because it’s transparent? What if we use another transparent color here with a rectangle that’s half the height of the viewport?

sparse_strips/vello_sparse_tests/tests/filter.rs
grebmeg: Yeah, definitely, I’m curious how much work it will actually take to fix that.

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -77,7 +78,11 @@ pub enum RenderError {
     /// TODO: Consider supporting more than a single column of slots in slot textures.
     #[error("No slots available for rendering")]
     SlotsExhausted,
-    // TODO: Consider expanding `RenderError` to replace some `.unwrap` and `.expect`.
+    /// An allocation error occurred while trying to allocate a new image. This can happen
```
grebmeg, line 81: I think the previous TODO is still relevant, isn’t it?

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -1109,17 +1262,112 @@ pub(crate) fn blur_with_edge_mode(ctx: &mut impl Renderer, edge_mode: EdgeMode)
...
+    ctx.pop_layer();
+
+    ctx.set_paint(BLUE);
+    ctx.fill_rect(&Rect::new(10.0, 60.0, 40.0, 90.0));
+
+    ctx.push_filter_layer(filter.clone());
+    ctx.set_paint(YELLOW);
+    ctx.fill_rect(&Rect::new(60.0, 60.0, 90.0, 90.0));
+    ctx.pop_layer();
+}
```
grebmeg, line 1356: Many of the new filter tests added above are already covered, to some extent, by the large `filter_varying_depths_clips_and_compositions` test. That said, we can keep them here for now, especially since hybrid constrainted doesn't run against that large test.

I think it’s fine to leave this as is, but also what about adding a couple more tests that combine filters and clipping?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -828,68 +1014,111 @@ impl Scheduler {
...
+                        alpha_fill,
                         paint_idxs,
+                        wide_tile_x,
+                        wide_tile_y,
+                        attrs,
                     );
+                }
+                // This is roughly equivalent to `process_layer_tile` in vello_cpu.
+                Cmd::PushBuf(LayerKind::Filtered(child_layer_id), blend_into_dest) => {
+                    let filtered_ranges = tile.layer_cmd_ranges.get(child_layer_id).unwrap();
```
grebmeg, line 1045: You moved each arm body into separate functions, but left this one, maybe worth doing the same for it as well?

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Yes, let’s do it in follow-ups. This PR is already quite large.

sparse_strips/vello_hybrid/src/filter.rs
grebmeg: Yep, that makes sense. My main concern here is that `deallocate_all_and_clear_context` is on the hot path in `render_scene`, so it would be good to avoid calling this method, and/or the `clear` methods within it.

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -0,0 +1,1078 @@
...
+pub(crate) struct FilterContext {
+    /// The encoded data for each filter used in the current scene that will be uploaded to the
+    /// filter data texture.
+    pub(crate) filters: Vec<GpuFilterData>,
+    /// At what texel offset the filter data for the given layer ID is stored in the texture.
+    pub(crate) offsets: HashMap<LayerId, u32>,
+    /// Data for each filter layer.
+    pub(crate) filter_textures: HashMap<LayerId, FilterLayerData>,
+    /// Image cache for storing filter intermediate textures.
+    pub(crate) image_cache: ImageCache,
```
grebmeg, line 438: Can we add a comment somewhere explaining that for the filters context, the backed textures are separate and not part of the texture atlas, as is the case for the main image cache?

It would also be good to add some documentation explaining the filters pipeline in vello_hybrid, to make it easier for future maintainers to understand.

sparse_strips/vello_dev_macros/src/test.rs
laurenz-canva: It's already annotated at the top, it is skipped because the test uses non-default blending (like all other methods in this list). Which doesn't work in the `constrained` case.

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: Will add, it's `no_ref` because it only checks for no panicking.

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: The point of the test is to make sure the background is preserved, not the contents of the filter layer. But I can just make it a nearly-transparent blue instead so we can check both!

sparse_strips/vello_hybrid/src/lib.rs
laurenz-canva: Right!

sparse_strips/vello_hybrid/src/schedule.rs
laurenz-canva: The problem here is that we have a couple of `continue`, which require it to live inside of the main while loop. In theory we could have the method return something so we know whether to continue, but I _feel_ like for now the way it is currently is a bit more clear. But feel free to let me know if you feel about it strongly, and then I can change it in a follow-up.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Makes sense, I'll try to add some documentation!

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: You are right, some more unit tests for filters with clip layers would probably be great. Since we recently switched to `push_clip_path` I think it's not that urgent so I'll just add a TODO for now, but will address this! 😄

sparse_strips/vello_common/src/filter/gaussian_blur.rs
laurenz-canva: Done.

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: Done!

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: Changed the test.

sparse_strips/vello_hybrid/src/lib.rs
laurenz-canva: Re-added it.

sparse_strips/vello_sparse_tests/tests/filter.rs
laurenz-canva: Added a TODO.

sparse_strips/vello_hybrid/src/filter.rs
laurenz-canva: Added a high-level description.

laurenz-canva:
Thanks a lot for the thorough review!! Will address follow-ups shortly. 🙂

PR #1495, opened 2026-03-09
fix: Implement clear_view for wgpu renderer and wire up the clear parameter

grebmeg:
The render_scene method's clear parameter was previously ignored (prefixed with _), leaving a TODO to fix it. This PR implements the clear functionality by adding a clear_view helper that issues a dedicated render pass to clear the target to transparent black before drawing.


LaurenzV (approved):
Can we maybe add a TODO to investigate adding tests for this?

PR #1496, opened 2026-03-09
vello_hybrid: Fix slow scenes with many changing gradients

### LINKED ISSUE #1413: vello_hybrid: Gradients that change in each frame lead to sluggish performance
Running the examples here in https://github.com/RobertBrewitz/vello-gradient-perf-research with vello_hybrid (commit hash at the time of testing) https://github.com/linebender/vello/tree/8b4a33284a9231e3a206ecde1224e98471824088 

zulip thread: [#vello > &#91;vello_hybrid&#93; Gradient performance, around 60 "particles".](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/.5Bvello_hybrid.5D.20Gradient.20performance.2C.20around.2060.20.22particles.22.2E/with/571738107)

## TL;DR
Current with `--release`:

Image

No cache and stop_len 64 with `--release`:

Image

## Details

Produces:

Image

And with this diff

```diff
diff --git a/sparse_strips/vello_hybrid/src/gradient_cache.rs b/sparse_strips/vello_hybrid/src/gradient_cache.rs
index 5929eafd..b07f5b4c 100644
--- a/sparse_strips/vello_hybrid/src/gradient_cache.rs
+++ b/sparse_strips/vello_hybrid/src/gradient_cache.rs
@@ -19,11 +19,11 @@ pub(crate) struct GradientRampCache {
     /// Current epoch for LRU tracking.
     epoch: u64,
     /// Cache mapping gradient signature to cached ramps and last access time.
-    cache: HashMap<CacheKey<GradientCacheKey>, (CachedRamp, u64)>,
+    pub cache: HashMap<CacheKey<GradientCacheKey>, (CachedRamp, u64)>,
     /// Packed gradient luts.
-    luts: Vec<u8>,
+    pub luts: Vec<u8>,
     /// Whether the packed luts needs to be re-uploaded.
-    has_changed: bool,
+    pub has_changed: bool,
     /// Maximum number of gradient cache entries to retain.
     retained_count: u32,
     /// SIMD level used for gradient LUT generation.
diff --git a/sparse_strips/vello_hybrid/src/render/wgpu.rs b/sparse_strips/vello_hybrid/src/render/wgpu.rs
index 2a90143f..170e7bb2 100644
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -142,6 +142,9 @@ impl Renderer {
         render_size: &RenderSize,
         view: &TextureView,
     ) -> Result<(), RenderError> {
+        self.gradient_cache.cache.clear();
+        self.gradient_cache.luts.clear();
+        self.gradient_cache.has_changed = true;
         self.prepare_gpu_encoded_paints(&scene.encoded_paints);
         // TODO: For the time being, we upload the entire alpha buffer as one big chunk. As a future
         // refinement, we could have a bounded alpha buffer, and break draws when the alpha
```

Produces:

Image

Also reducing lut size to 64 with this diff:

```diff
diff --git a/sparse_strips/vello_common/src/encode.rs b/sparse_strips/vello_common/src/encode.rs
index 8313c4e8..1eab0737 100644
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -1099,11 +1099,12 @@ fn determine_lut_size(ranges: &[GradientRange]) -> usize {
     // 256 for 2 stops.
     // 512 for 3 stops.
     // 1024 for 4 or more stops.
-    let stop_len = match ranges.len() {
-        1 => 256,
-        2 => 512,
-        _ => 1024,
-    };
+    // let stop_len = match ranges.len() {
+    //     1 => 256,
+    //     2 => 512,
+    //     _ => 64,
+    // };
+    let stop_len = 64;
 
     // In case we have some tricky stops (for example 3 stops with 0.0, 0.001, 1.0), we might
     // increase the resolution.
```

Produces:

Image

LaurenzV:
Closes #1413.

There were two problems in the old implementation:
1) When compacting the LUTs, we had a doubly-nested loop that, for each gradient existing in the cache we iterated over ALL removed gradients to compute the new `lut_start` position for that gradient.
2) When copying LUT data, we would do so pixel-by-pixel instead of copying the whole range at once.

This PR addresses both points by 1) computing a prefix sum for how much `lut_start` needs to be adjusted for a given cached ramp and 2) copying whole ranges at once using `copy_within`.

Before:
As you can see, the scene quickly gets laggy once we hit the compaction code path.

https://github.com/user-attachments/assets/319b7c80-a152-4239-9bea-ecf47f259f28

image

After:
Now, the bottleneck is basically only uploading the texture and computing the LUT on the CPU. Which is still not ideal, but much better than before.

https://github.com/user-attachments/assets/b159b47d-a4ee-4f10-aac3-8909cf0a4d91

https://github.com/user-attachments/assets/f58db4b4-2a46-4c14-a225-6fea11cd1314
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/gradient_cache.rs
+++ b/sparse_strips/vello_hybrid/src/gradient_cache.rs
@@ -165,12 +165,15 @@ impl GradientRampCache {
                 .map(|(key, (_, last_used))| (key, *last_used)),
         );
 
-        // Sort by last_used (ascending) to get LRU entries first
-        entries.sort_by_key(|(_, last_used)| *last_used);
+        let (lesser, median, _) =
+            entries.select_nth_unstable_by_key(count - 1, |(_, last_used)| *last_used);
 
+        // Note that since we use `select_nth_unstable`, the entries themselves are not guaranteed
+        // to be sorted.
         let mut removed = core::mem::take(&mut self.scratch.removed);
         removed.clear();
-        removed.extend(entries.iter().take(count).map(|(key, _)| (*key).clone()));
+        removed.extend(lesser.iter().map(|(key, _)| (*key).clone()));
+        removed.push(median.0.clone());
         self.scratch.entries = reuse_vec(entries);
 
         for key in removed.drain(..) {
@@ -191,7 +194,8 @@ impl GradientRampCache {
             return;
         }
 
-        // Sort by lut_start position (ascending) for efficient processing
+        // See comment in `remove_lru_entries`, the entries are not sorted yet but need to be
+        // for the prefix sum to work correctly.
         ramps_to_remove.sort_by_key(|(_, ramp)| ramp.lut_start);
 
         // Compute a prefix sum of how much the `lut_start` entry of a given cached ramp needs


sparse_strips/vello_hybrid/src/gradient_cache.rs
```diff
@@ -178,58 +182,56 @@ impl GradientRampCache {
     }
 
     /// Remove LUT data for evicted entries with compacting the LUTs vector, and update remaining offsets.
-    fn compact_luts(&mut self, ramps_to_remove: &mut [(CacheKey<GradientCacheKey>, CachedRamp)]) {
+    fn compact_luts(
```
grebmeg, line 188: One thought on `remove_lru_entries`: the full `sort_by_key` there is doing more work than needed. You're sorting all N entries just to take the smallest count, and `compact_luts` immediately re-sorts `ramps_to_remove` by `lut_start` anyway, so the LRU ordering is discarded.

`select_nth_unstable_by_key` (quickselect) would give you the count smallest in O(N) average instead of O(N log N):
```rust
entries.select_nth_unstable_by_key(count - 1, |(_, last_used)| *last_used);
```

sparse_strips/vello_hybrid/src/gradient_cache.rs
grebmeg, line 1: Nice work on the compaction rework! `copy_within` + prefix-sum approach is a big improvement over the byte-at-a-time loop.

One thing I've been thinking about: have you considered whether the gradient cache could eventually use the same atlas infrastructure as images, rather than managing its own packed `Vec<u8>` + compaction?

The idea would be:

- Allocate each gradient LUT as a small rectangular region in the `MultiAtlasManager` (e.g., a 1024-texel LUT becomes a 32×32 tile)
- Eviction becomes `image_cache.deallocate(slot)` — O(1), no compaction at all
- The atlas handles fragmentation internally via `guillotine` free-rect coalescing
- One fewer GPU texture and bind group to manage

The main trade-off is the 1D → 2D shape mismatch: thin 4096×1 strips would fragment the `guillotine` allocator badly, so you'd want to wrap LUTs into square-ish tiles, which adds a small amount of shader complexity for 2D indexing within the tile (though the current shader already does something similar to unwrap the flat gradient texture).

Not suggesting this for this PR — the current approach is a solid incremental fix. But it might be worth keeping in mind as a future direction if gradient-heavy scenes remain a hot spot, since it would eliminate the compaction problem at the root rather than optimizing it.

grebmeg: Another thought on the eviction strategy: the current count-based LRU approach requires collecting all entries, partitioning to find the smallest, then compacting. Have you considered an age-based eviction model instead, similar to what the glyph cache does?

The idea:

- Each entry already tracks `last_used` (epoch). During `maintain()`, entries older than some threshold get evicted directly — no sorting or partitioning needed.
- `hashbrown`'s `extract_if` would let you drain expired entries in a single pass, yielding the removed `(key, CachedRamp)` pairs directly — no scratch vecs, no cloning keys, no `select_nth_unstable`.
Something roughly like:
```rust
fn maintain(&mut self) {
    let threshold = self.epoch.saturating_sub(self.max_age);
    let mut lru_entries = core::mem::take(&mut self.scratch.lru_entries);
    lru_entries.clear();
    lru_entries.extend(
        self.cache
            .extract_if(|_, (_, last_used)| *last_used < threshold)
            .map(|(key, (ramp, _))| (key, ramp))
    );
    if !lru_entries.is_empty() {
        let mut prefix_sum = core::mem::take(&mut self.scratch.prefix_sum);
        self.compact_luts(&mut lru_entries, &mut prefix_sum);
        self.scratch.prefix_sum = prefix_sum;
        self.has_changed = true;
    }
    self.scratch.lru_entries = lru_entries;
}
```

LaurenzV: I like the idea of reusing the image allocator for this! Will add a TODO, happy to implement this in a follow-up.

LaurenzV: And also happy to revisit the eviction strategy (perhaps we can come up with some solution that can be shared between glifo and vello later on)? But think in the interest of keeping this PR as small as possible, it would be better to do this in a follow up.

LaurenzV:
> I think this is a useful feature for testing purposes in vello_example_scenes, so it would be worth including it.

I'll include it in my new benchmarking harness! I think it makes more sense there.

sparse_strips/vello_hybrid/src/gradient_cache.rs
LaurenzV: I've updated the logic now! Feel free to take another look, but I'll merge in the meanwhile.

LaurenzV:
@grebmeg  If you could take another look at the newest commit post-merge just to double check, that would be great.

grebmeg:
> @grebmeg If you could take another look at the newest commit post-merge just to double check, that would be great.

It looks good to me! Thank you!

PR #1497, opened 2026-03-10
vello_hybrid: Faster rotated rectangles (PoC)

laurenz-canva:
As a disclaimer again, the shader code was AI-generated and I haven't reviewed it deeply yet, I just wanted to get something working that passes all currently existing tests to test the feasibility. So not meant for deep review yet. 

The core idea would be that rectangles get an additional `rotation` attribute (this does unfortunately mean that we need to add an additional field to `GpuStrip`). In case this rotation is non-zero, we do additional work in the shader to compute the alphas, taking the rotation into account. The existing fast rectangle path can basically be left untouched.

It took me a while to prototype this, and during the implementation I encountered a lot of unexpected edge cases (the corner pixels of rotated rectangles yielded different opacities than the CPU version, complex paints didn't work from the get go), so I am a bit concerned about the additional complexity that this implementation brings. It would require us to add a lot more test cases for various interactions (I've added some already, but we will need many more), so I'm a bit hesitant of going down this route, unless we really have a compelling reason to optimize for the rotated rectangle case.

With that said, from my experiments, this does give a good speed boost, but it is significantly slower than drawing non-rotated rectangles (which isn't surprising given that there are a lot of computations happening in the current shader. It's possible this can be optimized further, but I didn't want to spend more time on this before we have a clear answer on whether we want this in the first place).

Here's what I get with current main. As can be seen, after just a few thousand rectangles we already drop below the 120FPS threshold, and it only keeps getting worse from then on:

https://github.com/user-attachments/assets/99e1399b-5b75-4dca-a89d-9955ab7c3a4e

With this current (as mentioned, potentially not optimized) implementation, I can get to 40k rectangles before it slowly starts to drop:

https://github.com/user-attachments/assets/50be270e-1f85-4eda-930b-ee2f2e67c329

However, this is still a lot worse than the non-rotated rectangles case, where we can easily get to 100k+ without really noticing anything:

https://github.com/user-attachments/assets/05bed7a4-448d-430d-a90f-a888ba001ee4

Curious to hear your thoughts @grebmeg @taj-p.

### DIFF
--- a/sparse_strips/vello_example_scenes/src/lib.rs
+++ b/sparse_strips/vello_example_scenes/src/lib.rs
@@ -10,6 +10,7 @@ pub mod gradient;
 pub mod image;
 pub mod multi_image;
 pub mod path;
+pub mod rotated_rects;
 pub mod simple;
 pub mod svg;
 pub mod text;
@@ -444,6 +445,7 @@ pub fn get_example_scenes<T: RenderingContext + 'static>(
     scenes.push(AnyScene::new(path::TrickyStrokesScene::new()));
     scenes.push(AnyScene::new(path::FunkyPathsScene::new()));
     scenes.push(AnyScene::new(path::RobustPathsScene::new()));
+    scenes.push(AnyScene::new(rotated_rects::RotatedRectsScene::new()));
 
     scenes.into_boxed_slice()
 }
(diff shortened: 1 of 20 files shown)


LaurenzV:
The consensus during office hours was that this is likely not worth having (and I agree), will close this but it can still be inspected for reference or in case we decide to pick it up again in the future!

PR #1499, opened 2026-03-11
Fix half-pixel offset when sampling images using pixel centers not corners

### LINKED ISSUE #972: Images do not roundtrip

Keavon:
Fixes #972 in Vello and https://github.com/GraphiteEditor/Graphite/issues/3018 in Graphite.

Here are some screenshots (view them at exactly 100% scale) from the Graphite viewport.

Blurry (currently in Vello):

image

Pixel-perfect (this fix):

image

Ground truth pixel-perfect (Graphite's SVG renderer, not Vello):

image

Disclosure: this is discovered and fixed with the help of Claude Code, but tested in Graphite. I have not, however, tested this in isolation outside of Graphite, but that might be something a reviewer could help with.

### DIFF
--- a/vello_shaders/shader/fine.wgsl
+++ b/vello_shaders/shader/fine.wgsl
@@ -1220,7 +1220,8 @@ fn main(
                         for (var i = 0u; i < PIXELS_PER_THREAD; i += 1u) {
                             // We only need to load from the textures if the value will be used.
                             if area[i] != 0.0 {
-                                let my_xy = vec2(xy.x + f32(i), xy.y);
+                                // Use pixel centers (+0.5) rather than pixel corners for correct sampling
+                                let my_xy = vec2(xy.x + f32(i) + 0.5, xy.y + 0.5);
                                 var atlas_uv = image.matrx.xy * my_xy.x + image.matrx.zw * my_xy.y + image.xlat;
                                 atlas_uv.x = extend_mode(atlas_uv.x, image.x_extend_mode, image.extents.x);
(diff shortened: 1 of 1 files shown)


DJMcNab:
I believe that this change is correct - thanks! 

I'm slightly surprised that this does not fix any of the known_issues tests, nor indeed changes any of the other snapshots? I would quite like this to have some test coverage.

I think that this should also have a changelog entry, but I won't block on you adding that.

Keavon:
Is the changelog entry something you could add, if you don't mind? I'm not very familiar with your development processes for doing that effectively. (I assume we're just talking about writing a sentence for that, presumably using a certain style that matches the others in the same phrasing.)

DJMcNab:
Yeah, as I noted the changelog entry isn't blocking. This PR is currently blocked on CI though, and (possibly relatedly) test coverage.

l0uisgrange:
Hey, any news on this matter?

Keavon:
I still have to return to this to add the remaining CI things but haven't had the time (I'm not really familiar with the intricacies of this project's testing system, so I'll need to see how that works and presumably regenerate some tests). If anyone else wants to push those to this branch, that would be appreciated, or I can hopefully get back to it in the next several days. (It's been on my mind, but pivoting context has been challenging.)

xStrom:
I created #1606 which builds upon this PR here and deals with tests and the changelog.

Keavon:
Thank you @xStrom!

PR #1500, opened 2026-03-12
vello_hybrid: Fix issue that causes mismatch between stored render size and assumed render size

LaurenzV:
We already [have code](https://github.com/linebender/vello/blob/ea2a935ec79206fb5dd3d367ac8feb79bffc905b/sparse_strips/vello_hybrid/src/render/webgl.rs#L962) that checks whether the currently stored render size (from the last iteration) doesn't match the new render size, and if so updates everything accordingly. Therefore, simply just don't replace the render size in the first place in `render_to_atlas`. I verified that this change also fixes the issue we encountered during the integration.


LaurenzV:
@taj-p Feel free to merge yourself if you need it.

PR #1501, opened 2026-03-12
vello_sparse_tests: Add support for multi-frame tests

LaurenzV:
The main intention behind this is to have infrastructure in place to better test glyph caching (where there are a range of bugs that only become evident when rendering more than one frame), to detect issues like the one in https://github.com/linebender/vello/pull/1500.


sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -276,9 +302,15 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
                 };
                 use vello_cpu::{RenderContext, RenderMode};
 
-                let mut ctx = get_ctx::<RenderContext>(#width, #height, #transparent, #num_threads, #level, #render_mode, false);
-                #input_fn_name(&mut ctx);
-                ctx.flush();
+                let mut ctx = get_ctx::<RenderContext>(#width, #height, #num_threads, #level, #render_mode, false);
+                for frame in 0..#frame_count {
+                    if frame > 0 {
+                        ctx.reset()
```
grebmeg:
```suggestion
                        ctx.reset();
```

sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -44,6 +44,11 @@ struct Arguments {
     no_ref: bool,
     /// A reason for ignoring a test.
     ignore_reason: Option<String>,
+    /// The number of frames to render. When greater than 1, the test function receives an
+    /// additional `u32` argument indicating the current frame index. The generated test will
+    /// loop `frame_count` times, resetting the context before each frame. Only the final
+    /// frame is checked against the reference image.
+    frame_count: u32,
```
grebmeg, line 51: What about snapshotting all frames into a single long strip and then comparing the whole thing? I think this way we could verify that all frames in the sequence have not changed.

sparse_strips/vello_sparse_tests/tests/basic.rs
```diff
@@ -547,3 +547,34 @@ fn composite_to_pixmap_at_offset() {
...
+
+    if frame == 0 {
+        ctx.set_paint(RED.with_alpha(0.5));
+    } else {
+        ctx.set_paint(BLUE.with_alpha(0.5));
+    }
+
+    ctx.fill_rect(&rect);
+    ctx.pop_layer();
+}
```
grebmeg, line 580: What about adding more tests to check for state leaks, or even better a single test (or unify above tests) where you'd have clip, opacity layer, blends, etc.?

grebmeg (approved):

sparse_strips/vello_dev_macros/src/test.rs
LaurenzV: Not sure I understand the reasoning here?

sparse_strips/vello_dev_macros/src/test.rs
LaurenzV: Hmm, I guess this would be not ideal if we have some tests that rely on many frames to trigger a bug, but I wouldn't mind changing it. @taj-p What do you think?

taj-p: Snapshotting across multiple frames increases our testing coverage in that we are also asserting on intermediate states. Perhaps it can be an additional flag to specify? Depending on what tests you want to add at this stage, we can decide on whether to default on or default off.

The existing test suite seems like it doesn't need it but I can imagine it might being wanted in the future (if we're unsure, we could leave it till then).

sparse_strips/vello_dev_macros/src/test.rs
grebmeg: Should there be semicolons after resetting? Sorry for removing the spaces (now fixed).

sparse_strips/vello_dev_macros/src/test.rs
grebmeg: I don’t mind adding an extra flag for this. As for test coverage, I don’t think we’ll have many tests that check using multiple frames, but I’d definitely prefer to keep snapshotting all frames behind a flag. My thinking is that the final output might be correct even if intermediate frames are not. Anyway, that’s just an assumption.

LaurenzV:
Maybe in the future, this PR has gotten pretty outdated. 😄

PR #1502, opened 2026-03-15
Re attach texture to re-trigger framebuffer completeness

taj-p:
Last week, I did some checking of Vello on a handful of low end devices. This is a bug I found while testing on a Samsung A05s, which uses very old drivers.

(I tested manually on A05s).

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1060,6 +1060,16 @@ impl WebGlPrograms {
                 0,
             );
 
+            #[cfg(debug_assertions)]
+            {
+                let status = gl.check_framebuffer_status(WebGl2RenderingContext::FRAMEBUFFER);
+                debug_assert_eq!(
+                    status,
+                    WebGl2RenderingContext::FRAMEBUFFER_COMPLETE,
+                    "view framebuffer incomplete"
+                );
+            }
+
             self.render_size = new_render_size.clone();
         }
     }


LaurenzV:
It's sad that there is no way to properly test this other than manually trying it out. :( Do you think (in the future) it would make sense to turn the WebGL rendering context into a trait that alternatively "records" the commands we emit, such that we can write tests that ensure correctness of the executed commands?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1043,6 +1043,23 @@ impl WebGlPrograms {
...
+                WebGl2RenderingContext::FRAMEBUFFER,
+                Some(&self.resources.view_framebuffer),
+            );
+            gl.framebuffer_texture_2d(
+                WebGl2RenderingContext::FRAMEBUFFER,
+                WebGl2RenderingContext::COLOR_ATTACHMENT0,
+                WebGl2RenderingContext::TEXTURE_2D,
+                Some(&self.resources.view_texture),
+                0,
+            );
```
grebmeg, line 1061: I’m wondering whether these kinds of bugs are easily catchable by consumers, and whether consumers or we should add some detection strategies to validate the gl state, somethig like ([check_framebuffer_status](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/checkFramebufferStatus) and [get_error](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getError)):
```rust
    let status = gl.check_framebuffer_status(FRAMEBUFFER);
    if status != FRAMEBUFFER_COMPLETE {
        return Err(format!("FBO incomplete: 0x{:X}", status));
    }
    // and/or
    let err = gl.get_error();
    if err != WebGl2RenderingContext::NO_ERROR {
        log::warn!("GL error after framebuffer_texture_2d: 0x{:X}", err);
    }
```

grebmeg: One thing worth considering: right now, the generated GLSL from Naga is only validated at runtime by whatever driver the user's browser happens to ship. That means a Naga upgrade, a shader change, or a subtle GLSL spec-conformance issue can silently break specific devices with no signal in CI. This fix is a good example of the kind of device-specific issue that's hard to catch before it reaches users.

Here are a few things we could add (as follow-up work) to try to systematically catch this class of problem earlier:

- Run [ANGLE's shader_translator](https://chromium.googlesource.com/angle/angle/+/refs/heads/main/src/compiler/translator/) in CI. This is the actual GLSL compiler that Chrome, Edge, and most Android WebGL implementations use. Running it offline on the generated GLSL strings would catch precision qualifier issues, ANGLE-rejected constructs, and uniform block layout mismatches before merge. Since the GLSL is available as Rust string constants at build time, a build-script or test integration would be straightforward.

- Validate with [glslangValidator](https://github.com/KhronosGroup/glslang) as a second opinion. It catches a different class of spec-conformance issues than ANGLE's parser — running both gives good cross-coverage. Lightweight to add alongside option 1.

- Round-trip through SPIR-V and run spirv-val. WGSL → Naga → SPIR-V → spirv-val validates that Naga's internal representation of the shader is consistent. This catches Naga backend bugs that can manifest as driver-specific crashes or miscompiles — a different layer than GLSL text validation.

- Snapshot the generated GLSL and diff on changes. Commit the generated GLSL output and fail CI when it changes unexpectedly. This makes any Naga upgrade or shader edit produce a visible diff — reviewers can spot silent changes to precision qualifiers, variable names, or codegen patterns before they ship. 

That translated code (or webgl itself), could potentially be run on cloud device farms. 

That said, the above are just ideas that first need to be confirmed.

taj-p: Happy to add `getError` and `checkFramebufferStatus` but when `debug_assertions` are enabled only because they can cause synchronous stalls on the thread (causing a flush and roundtrip on the GPU). See https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#avoid_blocking_api_calls_in_production.

It's a good callout. I wonder whether we should enable that debugging on all WebGL calls in debug mode!

> Here are a few things we could add (as follow-up work) to try to systematically catch this class of problem earlier:

These are brilliant ideas! WDYT about discussing how to tackle this problem and sharing these ideas at office hours?

taj-p: Argh! The issue with debug assertions is that the error probably only surfaces in production... but it impacts performance in production. Definitely something we want to chat about in OH.

taj-p: Added behind debug assertions flag in https://github.com/linebender/vello/pull/1502/commits/1020725ca9b72c2bc6b355d4c563bd32fda946d3

PR #1504, opened 2026-03-15
Use matrix primitives in `render_strips.wgsl`

taj-p:
When I tested Vello on my Samsung S10e, I was hitting invalid fragment shader syntax errors due to unspecified precision on float arrays (I'll raise this in WGPU separately).

I overcame this by using matrix primitives in the shader directly, which I think is likely a better implementation than hand rolling matrix calculations for readability/maintenance.

cc @LaurenzV


sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg, line 1: Nice fix! Using `mat2x2`/`vec2` is definitely cleaner than the manual array math.

Curious about the naga side of this: was the root cause that naga omits precision qualifiers on `array<f32, N>` members when transpiling to GLSL ES? i.e., it emits:

```glsl
float transform_[6];
```
instead of:
```glsl
highp float transform_[6];
```

taj-p: Yup! Exactly that!

PR #1505, opened 2026-03-16
fix: hanging in mmark example

fundon:
In my Mac mini M4, when the path elements reached at `70000`, its hanging.

### Before

https://github.com/user-attachments/assets/6889f717-86c8-43d5-8267-632c4584f9d0

### After

https://github.com/user-attachments/assets/22104d62-8889-4d75-b299-4e27ad646f2f

### DIFF
--- a/examples/scenes/src/mmark.rs
+++ b/examples/scenes/src/mmark.rs
@@ -81,9 +81,13 @@ impl TestScene for MMark {
         let mut path = BezPath::new();
         let len = self.elements.len();
         for (i, element) in self.elements.iter_mut().enumerate() {
-            if path.is_empty() {
-                path.move_to(element.seg.start());
+            // Delete old path events.
+            if !path.is_empty() {
+                path.truncate(0); // Should have clear method, to avoid allocations.
             }
+
+            path.move_to(element.seg.start());
+
             match element.seg {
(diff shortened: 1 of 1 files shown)


DJMcNab:
Welcome! There are a couple of things going on here, so I just want to check that we get the right answer here.
The "hang" is a known issue, and is (sort of) intentional behaviour when we run out of memory; this issue in #366. As soon as the scene complexity drops below the memory limit, it should stop hanging.
It currently seems extremely unlikely that we're going to resolve this in Vello Classic, unfortunately. This was (one of) the motivations for developing the ongoing sparse strip approach.

On the other hand, this change is potentially something we want. To give some context, the mmark example is an attempt to recreate the MotionMark benchmark from browserbench (https://browserbench.org/MotionMark/). As such, I'm very cautious about making changes here, as it may make our results be wrong if we make our version easier than the original. Can you describe more carefully what this change is doing? Without this change, do we have duplicated path elements?

fundon:
Thank you for providing the context.

I also sent this repair just to limit the number of path events to 2. In fact, it didn't fix anything. When I tried to expand the path events to 4, 6, 8, the problem reappeared. The problem should be in the GPU calculation.

It is necessary to continue to keep it. Thanks.

PR #1506, opened 2026-03-16
[sparse_strips] Hoist Horner coefficients in flatten_simd

b0nes164:
Currently, `eval_simd` evaluates curves using the standard Bernstein polynomial form directly inside the subdivision loop: 
$P(t) = (1-t)^3 P_0 + 3(1-t)^2 t P_1 + 3(1-t) t^2 P_2 + t^3 P_3$

This requires continuously calculating $(1 - t)$, expanding exponents, and performing multiple vector multiplications for *every* step of $t$. 

This implementation optimizes the evaluation loop in three steps:

1. **Polynomial Conversion**: We algebraically convert the Bernstein equation into a standard power basis: 
   $P(t) = At^3 + Bt^2 + Ct + D$
2. **Loop Hoisting**: Because the coefficients ($A$, $B$, $C$, $D$) rely entirely on the fixed control points, we compute them once *outside* the loop:
   `let coeff_a = (p1 - p2) * 3.0 - p0 + p3;`
3. **Horner's Method**: Inside the loop, we factor out $t$ to completely remove exponent calculations:
   $P(t) = ((A \cdot t + B) \cdot t + C) \cdot t + D$

The standalone `eval_simd` helper is removed.

Before:
```
flatten/Ghostscript_Tiger
time:   [132.20 µs 132.35 µs 132.49 µs]
```

After:
```
flatten/Ghostscript_Tiger
time:   [120.48 µs 120.68 µs 120.88 µs]
change: [-8.8932% -8.6772% -8.4569%] (p = 0.00 < 0.05)
```

Note: This patch was AI assisted.


b0nes164:
Test failures are identical across the backends and are the result of floating point noise when polyfilling to non-simd versions.

My local results, which reproduce the failure on the jobs:
clip_deeply_nested_circles_cpu_multithreaded:
clip_deeply_nested_circles_cpu_multithreaded

clip_deeply_nested_circles_cpu_f32_scalar:
clip_deeply_nested_circles_cpu_f32_scalar

clip_non_isolated_deeply_nested_circles_cpu_multithreaded:
clip_non_isolated_deeply_nested_circles_cpu_multithreaded

clip_non_isolated_deeply_nested_circles_cpu_f32_scalar:
clip_non_isolated_deeply_nested_circles_cpu_f32_scalar

clip_wrong_command_cpu_f32_scalar:
clip_wrong_command_cpu_f32_scalar

clip_wrong_command_cpu_multithreaded:
clip_wrong_command_cpu_multithreaded

tricky_strokes_cpu_f32_scalar:
tricky_strokes_cpu_f32_scalar

tricky_strokes_cpu_multithreaded
tricky_strokes_cpu_multithreaded

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -403,16 +386,24 @@ fn eval_cubics_simd<S: Simd>(simd: S, c: &CubicBez, n: usize, result: &mut Flatt
...
     let (p2_128, p3_128) = split_single(p2p3);
 
+    let coeff_a = (p1_128 - p2_128) * f32x8::splat(simd, 3.0) - p0_128 + p3_128;
+    let coeff_b = (p0_128 - p1_128 * f32x8::splat(simd, 2.0) + p2_128) * f32x8::splat(simd, 3.0);
+    let coeff_c = (p1_128 - p0_128) * f32x8::splat(simd, 3.0);
+    let coeff_d = p0_128;
+
     let iota = f32x8::from_slice(simd, &[0.0, 0.0, 2.0, 2.0, 1.0, 1.0, 3.0, 3.0]);
-    let step = iota * dt;
+    let step = iota * f32x8::splat(simd, dt);
```
LaurenzV: Is this change needed?

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -403,16 +386,24 @@ fn eval_cubics_simd<S: Simd>(simd: S, c: &CubicBez, n: usize, result: &mut Flatt
     let (p0_128, p1_128) = split_single(p0p1);
     let (p2_128, p3_128) = split_single(p2p3);
 
+    let coeff_a = (p1_128 - p2_128) * f32x8::splat(simd, 3.0) - p0_128 + p3_128;
```
LaurenzV: Maybe comment to this PR would be good!

LaurenzV: also I don't thinkl you ened to splat 3 here, I thin you can multiply vectors just with scalars.

LaurenzV: Could we even use `mul_add` here? `(p1_128 - p2_128).mul_add(f32x8::splat(simd, 3.0), p3_128) - p0_128`.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -519,24 +510,24 @@ fn output_lines_simd<S: Simd>(
 
     const IOTA2: [f32; 8] = [0., 0., 1., 1., 2., 2., 3., 3.];
     let iota2 = f32x8::from_slice(simd, IOTA2.as_ref());
-    let x = iota2.mul_add(dx, f32x8::splat(simd, x0));
+    let x = iota2.mul_add(f32x8::splat(simd, dx), f32x8::splat(simd, x0));
     let da = f32x8::splat(simd, ctx.da[i]);
     let mut a = da.mul_add(x, f32x8::splat(simd, ctx.a0[i]));
-    let a_inc = 4.0 * dx * da;
+    let a_inc = f32x8::splat(simd, 4.0 * dx) * da;
```
LaurenzV: Same for those two, are those changes necessary?

LaurenzV (approved):
Nice, LGTM, the math checks out for me! Great improvement, not noticing anything fishy in my PDF test suite.

sparse_strips/vello_common/src/flatten_simd.rs
b0nes164: I've changed this, but does changing to scalars change the compiled code?

LaurenzV: ah no dont think so, it's just a nit! Feel free to disregrad 😄

b0nes164: No, you're right I should have cleaned up the slop changes before uploading it!

b0nes164:
I don't think I have the authority to bypass the CI, but no rush on merging this. Also, do I need to update the affected golden images or is that something that happens automatically?

LaurenzV:
oh yeah, you just need to run with `REPLACE=1 cargo test --release`, sorry if that wasn't clear! After that you can merge.

PR #1508, opened 2026-03-18
Streamline handling of pixel center offsets

LaurenzV:
This resolves a TODO, such that we don't need "do and then undo" the  center shift transform in vello_hybrid. It's not entirely clear to me why it doesn't work for blurred rounded rectangles, but unfortunately I don't really understand the algorithm in the first place, so I can't explore this further in the short term.

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -35,22 +35,6 @@ use peniko::kurbo::common::FloatFuncs as _;
 
 const DEGENERATE_THRESHOLD: f32 = 1.0e-6;
 const NUDGE_VAL: f32 = 1.0e-7;
-const PIXEL_CENTER_OFFSET: f64 = 0.5;
-
-/// How the client performs pixel sampling.
-///
-/// For complex paints like images and gradients, it's important that the value is always sampled
-/// at the center of the pixel. In case the client code uses sampling based on the top-left corner
-/// instead, an additional correctional shift will be encoded in the transform of the paint to
-/// account for that.
-#[derive(Debug, PartialEq, Eq, Copy, Clone)]
-pub enum PixelSampling {
-    /// Assume that the client does sampling from the pixel center by default.
-    Center,
-    /// Assume that the client does sampling from the top-left pixel corner by default.
-    Corner,
-}
-
 #[cfg(feature = "std")]
 fn exp(val: f32) -> f32 {
     val.exp()
@@ -73,13 +57,6 @@ pub trait EncodeExt: private::Sealed {
         paints: &mut Vec<EncodedPaint>,
         transform: Affine,
         tint: Option<Tint>,
-        // Note: We make this configurable because different backends have different requirements.
-        // For vello_cpu, it makes sense to apply the center shift _once_ during encoding, so that
-        // we don't have to manually apply the shift for each wide tile command (since by default,
-        // we will sample from the top-left corner).
-        // However, for vello_hybrid, the fragment shader already calculates the positions such
-        // that they are centered in the pixel, so there is no need to apply an additional shift.
-        pixel_sampling: PixelSampling,
     ) -> Paint;
 }
 
@@ -90,7 +67,6 @@ impl EncodeExt for Gradient {
         paints: &mut Vec<EncodedPaint>,
         transform: Affine,
         _tint: Option<Tint>,
-        pixel_sampling: PixelSampling,
     ) -> Paint {
         // First make sure that the gradient is valid and not degenerate.
         if let Err(paint) = validate(self) {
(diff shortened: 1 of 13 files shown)


sparse_strips/vello_common/src/encode.rs
```diff
@@ -59,6 +59,12 @@ pub trait EncodeExt: private::Sealed {
         paints: &mut Vec<EncodedPaint>,
         transform: Affine,
         tint: Option<Tint>,
+        // Note: We make this configurable because different backends have different requirements.
+        // For vello_cpu, it makes sense to apply the center shift _once_ during encoding, so that
+        // we don't have to manually apply the shift for each wide tile command.
+        // However, for vello_hybrid, the fragment shader already calculates the positions such
+        // that they are centered in the pixel, so we don't need to apply the shift during encoding.
+        apply_center_shift: bool,
```
grebmeg: What about replacing `apply_center_shift: bool` with a named enum to avoid the boolean-trap problem where passing `true` or `false` into a function makes the call site less readable?

This makes every call site self-documenting (e.g., `PixelSampling::Center` vs `true`):

```rust
pub enum PixelSampling {
    Center,
    Corner,
}
```

sparse_strips/vello_common/src/encode.rs
```diff
@@ -1172,7 +1188,7 @@ mod tests {
         };
 
         assert_eq!(
-            gradient.encode_into(&mut buf, Affine::IDENTITY, None),
+            gradient.encode_into(&mut buf, Affine::IDENTITY, None, false),
```
grebmeg: What about adding one test that verifies `apply_center_shift` correctly bakes the `(0.5, 0.5)` or not translation into the encoded transform?

sparse_strips/vello_common/src/encode.rs
```diff
@@ -847,6 +859,10 @@ impl EncodeExt for BlurredRoundedRectangle {
         paints: &mut Vec<EncodedPaint>,
         transform: Affine,
         _tint: Option<Tint>,
+        // I'm not sure why, but for some reason applying this here gives wrong
+        // results for the "blurred_rounded_rect_non" test case. It's possible that the
+        // current algorithm somehow already assumes centered sampling.
+        _apply_center_shift: bool,
```
grebmeg: Could you investigate which case this is? If the algorithm is inherently center-sampled, the comment should document that understanding rather than leaving the uncertainty.

From what I can tell, this is the "algorithm already assumes centered sampling" case. The pixel center offset is hardcoded directly in the math [here](https://github.com/linebender/vello/blob/302fa902ebc5a7605b6664393d43abaf369eee29/sparse_strips/vello_cpu/src/fine/common/rounded_blurred_rect.rs#L193) `v1 = 0.5` in `SimdRoundedBlurredRect::new`, which then feeds into the distance field computation [here](https://github.com/linebender/vello/blob/302fa902ebc5a7605b6664393d43abaf369eee29/sparse_strips/vello_cpu/src/fine/common/rounded_blurred_rect.rs#L140-L147):
```rust
// Equivalent to j + r.v1 - r.v1 * r.height  →  j + 0.5 - 0.5 * height
let y = j - r.v1.mul_sub(r.height, r.v1);
// Equivalent to i + r.v1 - r.v1 * r.width   →  i + 0.5 - 0.5 * width
let x = i - r.v1.mul_sub(r.width, r.v1);
```

sparse_strips/vello_common/src/encode.rs
grebmeg, line 1: One thing I'm curious about: did you consider an alternative where the encoding never bakes the +0.5 shift into the transform, and instead the CPU fine rasterizer applies it at the consumption site? Something like:
```rust
let start_pos = image.transform
    * Point::new(f64::from(start_x) + 0.5, f64::from(start_y) + 0.5);
```

The upside would be keeping `EncodeExt::encode_into`'s signature unchanged (no new parameter), and making the responsibility clear: the consumer that iterates integer pixel coordinates is the one that decides where those integers fall in continuous space. It would also avoid the `BlurredRoundedRectangle` question entirely: no special-casing needed since the +0.5 would be applied uniformly to all paints at the same place.

LaurenzV: Yes, I briefly mentioned this in one of the comments, the reason I didn't do it this way is that we need that this will add two additions to _every_ single wide tile command, instead of only needing it to apply once for the whole path and the transform automatically being propagated to all wide tiles. It's probably not going to have a noticeable impact, but since it seemed like an easy thing to avoid I chose the other way for now. 

But if you prefer this, I'm happy to change it. The only downside is that it requires us to hardcode this change into every single painter kernel, so I'm not sure if it's better in that regard.

sparse_strips/vello_common/src/encode.rs
LaurenzV: That sounds good, can make that change!

sparse_strips/vello_common/src/encode.rs
LaurenzV: Can add this.

sparse_strips/vello_common/src/encode.rs
LaurenzV: I did play around with this, but always got unreasonable test failures after changing. I think it's somehow encoded into the algorithm itself, but I can take another stab at investigating this, maybe I missed something.

LaurenzV: Okay you are right, the fix seemed pretty easy! I had to regenerate one of the snapshots, but I believe this is because the current snapshot is more correct than the previous one. The previous one would apply the 0.5 offset in the current user space (meaning it would be affected by the current scaling/skewing transform), while now it's always applied in pixel space (which should be the correct one).

sparse_strips/vello_common/src/encode.rs
LaurenzV: The cleanest option is to create `sampling_x` and `sampling_y` f32 variables here with the offset applied, and then use it in the given paint samplers. What do you think?: 

```rs
            Paint::Indexed(paint) => {
                let color_buf = &mut self.paint_buf[x * TILE_HEIGHT_COMPONENTS..]
                    [..TILE_HEIGHT_COMPONENTS * width];

                let encoded_paint = &encoded_paints[paint.index()];

                let start_x = self.wide_coords.0 * WideTile::WIDTH + x as u16;
                let start_y = self.wide_coords.1 * Tile::HEIGHT;
```

I think I agree now that it's probably more clean to apply that change in vello_cpu instead. WIll create a new commit that does this, but if you think the older version is better I can revert that.

LaurenzV:
I think I like that new approach better, let me know what you think!

grebmeg (approved):
Nice!! I think it’s much cleaner now, thank you!

LaurenzV:
Thanks for the great idea!

PR #1510, opened 2026-03-18
vello_hybrid: Fix implementation for blending layers inside of filters

laurenz-canva:
The problem was that we were applying `blend_into_dest` when applying the already-filtered layer, instead of when rendering the filter layer itself. This is wrong, because if we have blending _inside_ of the filter layer, then we need to create the temporary slots when rendering the filter layer itself, instead of doing this in the parent layer that just samples from the already rendered filter texture.

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -709,6 +709,7 @@ impl Scheduler {
                     continue;
                 };
 
+                // TODO: Use enum instead of `wrap_surface`?
                 let wrap_surface = matches!(
                     wide_tile.cmds[ranges.full_range.start],
                     Cmd::PushBuf(_, true)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -709,9 +709,10 @@ impl Scheduler {
                     continue;
                 };
 
-                // See the comment in `do_tile`, we currently don't need this for filter layers.
-                // This might change in the future.
-                let wrap_surface = false;
+                let wrap_surface = matches!(
+                    wide_tile.cmds[ranges.full_range.start],
+                    Cmd::PushBuf(_, true)
```
grebmeg, line 715: [nit]: I think it’d be clearer to avoid using a boolean here, it’s hard to read without tracing what the value represents. I’d suggest using a small enum instead.

laurenz-canva: Might make sense, but since it already was that way before I'll just leave a TODO for now.

PR #1511, opened 2026-03-18
vello_common: Support rectangles with fractional offsets in fast path

laurenz-canva:
**Below is the description of the old fix, at the moment of writing the PR has been repurposed to support fractional offsets in the fast rectangle path for CPU, rendering the original change described below useless.**

Our rectangle fast path now also supports anti-aliasing, but the `flush_fast_path` method still accidentally called the method that assumes pixel-aligned rectangles. Hence, when pushing a blend layer, the rectangle would lose it's anti-aliasing.

Before:
image

After: It matches the CPU implementation.

I also noticed that we don't have proper handling for `aliasing_threshold` in the rectangle fast path, which we should probably fix, or at least bail out. Added this as a TODO.

### DIFF
--- a/sparse_strips/vello_common/src/rect.rs
+++ b/sparse_strips/vello_common/src/rect.rs
@@ -21,7 +21,7 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
     dispatch!(level, simd => render_impl(simd, rect, strip_buf, alpha_buf));
 }
 
-/// Generates strip data for a pixel-aligned rectangle.
+/// Generates strip data for an axis-aligned rectangle.
 ///
 /// # Strip layout strategy
 ///
@@ -30,7 +30,7 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
 /// - **Edge rows** (top/bottom of rect): the rect boundary crosses partway
 ///   through the tile vertically, so individual pixels need per-cell alpha.
 ///   We emit a *single wide strip* spanning all tile columns, with alpha =
-///   `x_mask & y_mask` (each is 0x00 or 0xFF, so AND gives the intersection).
+///   `x_alpha` * `y_alpha` (so the intersection of the alpha mask in each direction).
 ///
 /// - **Interior rows**: every pixel in the tile has full vertical coverage,
 ///   so we only need to handle the left and right partial-column edges.


sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -577,8 +581,16 @@ impl Scene {
...
+                    self.strip_generator.generate_filled_path(
+                        rect.to_path(DEFAULT_TOLERANCE),
+                        Fill::NonZero,
+                        Affine::IDENTITY,
+                        // The fast path for rects currently isn't used if an aliasing threshold is set.
+                        None,
+                        &mut strip_storage,
+                        // The fast path for rect doesn't trigger if a clip path was active.
+                        None,
+                    );
```
grebmeg: Instead of replacing `generate_filled_rect_fast` with the full path pipeline here, what would you think about improving `rect::render` so that it'll compute fractional coverage for edge pixels? That would let `flush_fast_path` keep using the fast rect path for all rects, preserving the performance win, while producing correct AA.

The change to `rect.rs` looks fairly contained and should be fairly small . The [existing TODO](https://github.com/linebender/vello/blob/302fa902ebc5a7605b6664393d43abaf369eee29/sparse_strips/vello_common/src/rect.rs#L43) already calls this out.

sparse_strips/vello_common/src/rect.rs
```diff
@@ -40,23 +40,23 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
 ///
 /// The x-alpha masks for the left/right edge tiles are y-independent, so they
 /// are precomputed once and reused across all interior rows.
-// TODO: Consider extending this to handle arbitrary axis-aligned rectangles (with fractional
-// coordinates) by computing partial coverage alpha values for edge pixels instead of
-// binary 0/255.
 fn render_impl<S: Simd>(s: S, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &mut Vec<u8>) {
-    // TODO: Negative rect coordinates are not handled correctly — casting negative
-    // f64 to u16 saturates to 0. The caller currently clamps to the viewport, but if
-    // that changes, this will need signed-integer math or explicit clamping.
```
LaurenzV, line 49: Was this ever a problem in the first place? Since anything on the negative side isn't visible in the first place, so I'm not sure how this would have broken anything.

LaurenzV:
@grebmeg Repurposed this PR to support arbitrary fractional offsets, let me know what you think!

sparse_strips/vello_common/src/rect.rs
```diff
@@ -40,23 +40,24 @@ pub fn render(level: Level, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &
 ///
 /// The x-alpha masks for the left/right edge tiles are y-independent, so they
 /// are precomputed once and reused across all interior rows.
-// TODO: Consider extending this to handle arbitrary axis-aligned rectangles (with fractional
-// coordinates) by computing partial coverage alpha values for edge pixels instead of
-// binary 0/255.
 fn render_impl<S: Simd>(s: S, rect: Rect, strip_buf: &mut Vec<Strip>, alpha_buf: &mut Vec<u8>) {
```
grebmeg, line 43: Now that fractional coordinates are supported, the doc comments above, including file level doc comment, are stale. Maybe "Fast axis-aligned rectangle rendering directly into strips" should be there now? Could you please update them?

grebmeg (approved):
Nice, thanks for the refactor! The performance still looks pretty good.

- Before

Image

- After
Image

sparse_strips/vello_common/src/rect.rs
grebmeg: You're right, it was never actually a problem. Both call sites (`try_fast_rect` clamps to viewport bounds, `generate_filled_rect_fast` intersects with the viewport) guarantee non-negative coordinates before reaching `render_impl`.

PR #1519, opened 2026-03-20
vello_common: Cull geometry outside of the clip

tomcur:
This cheaply calculates a bounding box of a clip path by calculating the bounding box of the path's control points, and uses it for culling during flattening.

This strengthens culling quite a bit: on `main` culling allows us to skip flattening, tiling, sorting and analytic AA for out-of-viewport geometry; this PR now generalizes that to out-of-clip-bbox geometry.

I instrumented the benchmarks to add a Tiger variant with a clip layer that's a 30x30 square at the Tiger's face (in both cases, the viewport is 200x200). This gives about a 5-6x speed up for that particular case.

```
flatten/Ghostscript_Tiger
                        time:   [179.68 µs 179.80 µs 179.94 µs]
flatten/Ghostscript_Tiger_clipped
                        time:   [31.634 µs 31.649 µs 31.664 µs]

tile_aaa/Ghostscript_Tiger
                        time:   [172.45 µs 172.80 µs 173.20 µs]
tile_aaa/Ghostscript_Tiger_clipped
                        time:   [33.857 µs 34.166 µs 34.569 µs]

                        time:   [195.25 µs 195.47 µs 195.72 µs]
render_strips/Ghostscript_Tiger_clipped_simd
                        time:   [56.732 µs 56.749 µs 56.772 µs]
```

The same considerations of "left edge challenges" apply unchanged, in that geometry to the left of the cull bbox impacts winding and cannot be completely elided. We still emit that geometry as straight line segments, skipping flattening. No code changes are required for that, as that's sound even if those lines are inside the viewport: they're outside the clip, and that part of the geometry won't be visible. The winding contributes to the regions inside the clip.

### DIFF
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -15,6 +15,10 @@ You can find its changes [documented below](#007---2026-03-24).
 
 This release has an [MSRV][] of 1.92.
 
+### Optimized
+
+- Performance of clipped drawing by culling geometry outside the clip bounding box during flattening. ([#1519][] by [@tomcur][])
+
 ## [0.0.7][] - 2026-03-24
 
 This release has an [MSRV][] of 1.92.
@@ -184,6 +188,7 @@ See also the [vello_cpu 0.0.1](../vello_cpu/CHANGELOG.md#001---2025-05-10) relea
 [#1478]: https://github.com/linebender/vello/pull/1478
 [#1484]: https://github.com/linebender/vello/pull/1484
 [#1486]: https://github.com/linebender/vello/pull/1486
+[#1519]: https://github.com/linebender/vello/pull/1519
 
 [Unreleased]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.7...HEAD
 [0.0.7]: https://github.com/linebender/vello/compare/sparse-strips-v0.0.6...sparse-strips-v0.0.7


tomcur:
> I'm wondering, since we seem to be converting the bbox to f64 later on anyway, maybe we should just use a Rect in the first place instead of [u16; 4]? What do you think?

Good question. The reason I did not, was to align this with strip coordinates and potentially allow this to be reused more easily in other parts. I suppose if that comes up we could just do the integer AABB at that point. I'll change this to `Rect`.

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -46,13 +49,26 @@ pub(crate) fn flatten<S: Simd>(
...
+    // The culling performed here would remove that geometry. However, as it does not extend above
+    // the strip row, it does not add coarse winding to the strip. Yet, if this is the top part of,
+    // say, a rectangle, this geometry (or coarse winding) is necessary in later stages for the
+    // bottom part of the strip to get filled.
+    //
+    // Therefore, we align `top` to strip row boundaries.
+    let left = cull_bbox[0] as f64;
+    let top = (cull_bbox[1] / Tile::HEIGHT * Tile::HEIGHT) as f64;
+    let right = cull_bbox[2] as f64;
+    let bottom = cull_bbox[3] as f64;
```
tomcur, line 79: There is some subtlety that I hadn't realized before: the culling occurring above the bounding box has to be aligned to tile rows, so this now does that. That also means it's somewhat more natural to express the bbox in `u16` pixel coordinates after all.

sparse_strips/vello_sparse_tests/snapshots/clip_rect_cull_alignment.png
tomcur, line 1: Running the regression test before the [`Align to strip row boundaries, add comment`](https://github.com/linebender/vello/pull/1519/commits/888c9556664081f6e9f269477f54a7bb6a52bada) commit, results in the following.

Image

sparse_strips/vello_common/src/clip.rs
```diff
@@ -92,9 +101,38 @@ impl ClipContext {
...
+        // Intersect with the existing clip bounding box, or the viewport if this is the outermost
+        // clip.
+        if let Some(existing) = self.clip_stack.last() {
+            bbox[0] = bbox[0].max(existing.bbox[0]);
+            bbox[1] = bbox[1].max(existing.bbox[1]);
+            bbox[2] = bbox[2].min(existing.bbox[2]);
+            bbox[3] = bbox[3].min(existing.bbox[3]);
+        } else {
+            bbox[2] = bbox[2].min(strip_generator.width());
+            bbox[3] = bbox[3].min(strip_generator.height());
```
LaurenzV, line 129: Does it not make sense to `max` with 0 here as well?

sparse_strips/vello_common/src/flatten_simd.rs
```diff
@@ -46,13 +49,26 @@ pub(crate) fn flatten<S: Simd>(
...
+    // coordinate space).
+    //
+    // The culling performed here would remove that geometry. However, as it does not extend above
+    // the strip row, it does not add coarse winding to the strip. Yet, if this is the top part of,
+    // say, a rectangle, this geometry (or coarse winding) is necessary in later stages for the
+    // bottom part of the strip to get filled.
+    //
+    // Therefore, we align `top` to strip row boundaries.
+    let left = cull_bbox[0] as f64;
+    let top = (cull_bbox[1] / Tile::HEIGHT * Tile::HEIGHT) as f64;
```
LaurenzV: Could we maybe for extra clarity do:

```suggestion
    let top = ((cull_bbox[1] / Tile::HEIGHT) * Tile::HEIGHT) as f64;
```

sparse_strips/vello_common/src/flatten_simd.rs
LaurenzV, line 79: Why is this not necessary for the bottom side? It seems to me like something similar could happen there? Or am I missing something? If so, could you add another comment? Maybe also another test that tests it for that side to show it works?

sparse_strips/vello_common/src/clip.rs
tomcur: This is now `[u16; 4]` instead of `Rect`, so it's always `>= 0`. (If integer coordinates in the pipeline end up changing to `i16`, and we also do that for clipping, this probably would need the `max`.)

LaurenzV: Ah, right. 🤦

sparse_strips/vello_common/src/flatten_simd.rs
tomcur: I've added a comment and updated the test. In short, this is symmetrical with coarse winding being accounted when geometry crosses the top edge of a strip.

PR #1520, opened 2026-03-21
Update skrifa to 0.41 (and add support for Varc glyphs)

oscargus:
Main point was to update skrifa, but it turns out that Varc glyphs are now supported.

Made an attempt, assuming that it works similar to the rest, but not sure how to test it (nor if it actually works).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1087,9 +1087,9 @@ dependencies = [
 
 [[package]]
 name = "font-types"
-version = "0.11.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1e4d2d0cf79d38430cc9dc9aadec84774bff2e1ba30ae2bf6c16cfce9385a23"
+checksum = "73829a7b5c91198af28a99159b7ae4afbb252fb906159ff7f189f3a2ceaa3df2"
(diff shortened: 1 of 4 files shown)


xStrom:
Let's wait with merging this until we get a sparse strips release out (hopefully within a few days), because otherwise the sparse strips release will have a different skrifa version than Vello Classic, which was released yesterday.

nicoburns (approved):
Releases have now been published :)

DJMcNab:
Thank you! Apologies that this didn't get merged as this PR, I'm not quite sure why that didn't happen. This code was merged in https://github.com/linebender/vello/pull/1520 (edit: lol. Thanks @xstrom)

oscargus:
No worries. I do PRs like this to reduce the number of dependencies in my project, not as a class assignment. ;-)

xStrom:
Just for the record, the code was merged as part of #1594.

nicoburns:
Also for the record: this wasn't initially merged becuase of https://github.com/linebender/vello/pull/1520#issuecomment-4103384553. And then by the time I came to merging it, there were fairly significant conflicts due to `glifo` refactoring. And I decided it would easier to fix that myself rather than asking [oscargus](https://github.com/oscargus) to do it.

PR #1526, opened 2026-03-24
vello_hybrid: Disallow background optimization in case at least one filter layer has been pushed

### LINKED ISSUE #1509: sparse_strips: Rotated drop shadows behave weirdly
Reproduction:
1) Start the newly added example scenes for filters.
2) Press + once to add a new batch of shapes.
3) Zoom into the given square and then pan the viewport.

See also the attached video. Happens for both, vello_cpu and vello_hybrid.

https://github.com/user-attachments/assets/e05dcbba-1726-4536-a6b5-1c50b7ca78b4

LaurenzV:
Fundamentally, the problem is that in the current way commands are stored, we have no way of saying "clear all commands in the root but leave commands from the filter layer untouched". Therefore, the best solution for now is to simply disable the optimization if there has been at least one filter layer.

See here the before vs after:
issue_1509_cpu_f32_neon

Closes #1509.

### DIFF
--- a/sparse_strips/vello_common/src/coarse.rs
+++ b/sparse_strips/vello_common/src/coarse.rs
@@ -1528,6 +1528,9 @@ impl<const MODE: u8> WideTile<MODE> {
                 && width == WideTile::WIDTH
                 && self.n_clip == 0
                 && self.n_bufs == 0
+                // TODO: Relax this condition to only trigger for filter graphs that contain at
+                // least 1 primitive filter that has non-local effects (see
+                // https://github.com/linebender/vello/pull/1526#discussion_r3007377869)
                 && !self.had_filter_layer;
 
             if can_override {


sparse_strips/vello_common/src/coarse.rs
```diff
@@ -1360,6 +1360,8 @@ pub struct WideTile<const MODE: u8 = MODE_CPU> {
     ///
     /// Only used in `HYBRID` mode.
     push_buf_indices: Vec<usize>,
+    /// Whether at least one filter layer has been pushed on this tile.
+    had_filter_layer: bool,
```
taj-p, line 1364: Is it now worth using bit flags for the three bools on wide tile to reduce its size?

LaurenzV: Might be worth considering, but I think the other fields are large enough that I'm not sure if I should really piggy-back this onto the current PR. 😄 As you want!

grebmeg, line 1364: Could we avoid this class of bug entirely by collapsing filter layers into a separate tile command stream?

The real filter work happens at the render graph level: the `FilterLayer` node renders to a pixmap/texture, applies the filter, stores the result, and the root layer composites it back by pattern-matching on `PushBuf(Filtered(...))` and skipping ahead through `layer_cmd_ranges`.

The `had_filter_layer` fix is needed precisely because the tile stream carries these remaining commands that interact with overdraw elimination in unintended ways. And the fix is broader than necessary — it disables overdraw elimination for all filter types, but only spatially-sampling filters (blur, drop-shadow) actually break when prior tile content is erased. A color matrix or flood filter wouldn't care.

What about instead of filter layer commands living inline in the parent's tile cmds (and then needing skip logic, range tracking, and special-case guards), a filter layer gets its own `Wide` tile grid or similar storage (we already have `wtile_bbox`). When you push a filter layer, you start emitting commands into a fresh set of wide tiles. So that the parent's cmds never contain `PushBuf(Filtered)`, `Fill`, `Filter`, `Blend`, `PopBuf` for the child layer. They just contain maybe something like `CompositeFilteredLayer(layer_id)` at the point where the result should be blended in.

I think this should be fundamentally right because a filter layer is a separate rendering surface. It has its own coordinate space (potentially with expanded bounds for spatial filters, eventually with negative coordinates), its own background (transparent), and its own command stream. Mixing its commands into the parent's tile stream and then un-mixing them at render time is the source of all the fragility, the overdraw elimination, the `had_filter_layer` guard, the `in_clipped_filter_layer` flag, the range-skipping state machine in `process_layer_tile`.

This would likely mean:
- Overdraw elimination just works
- `Cmd::Filter` and `had_filter_layer` both go away — the filter is applied in the render graph where it belongs; the tile stream just says "composite the pre-rendered result here"
- Future-proofs for negative coordinates — when `bbox` expansion goes negative, only the render graph's allocation logic changes.
- Spatial vs. non-spatial becomes irrelevant to the tile stream — the distinction only matters inside render graph execution, which already handles it.

LaurenzV: I agree the proper solution is to change how wide tile commands for layers is stored. However, this is a major refactor, hence why I would prefer leaving this for a future PR. :/ Would that be okay? While it’s not ideal, I think this patch should be good enough so that the bug is gone. 

However, I’m happy to add a restriction so that this only triggers for non-local filters.

LaurenzV:
(Nevermind, forgot about the pixel-local thing first, can look into that before the next review)

LaurenzV:
I tried to loosen the restriction to only apply to filters that have non-local effects, but realized that we currently don't have a mechanism to determine whether a filter graph has such primitives. The options would either be to traverse the filter graph each time (which would be somewhat expensive) or to cache it somehow (in which case we need to make sure it stays in sync). 

Since this would probably warrant some more design discussion, I think it's better if I just leave it as a TODO for now. The main filter primitives we are interested in are gaussian blur and drop shadow which are both affected by this, so we wouldn't gain much from this anyway. I hope that's okay.

PR #1529, opened 2026-03-24
vello_hybrid: Remove handling of `undefined` from gradient LUT

### LINKED ISSUE #1528: vello_hybrid: Gradients with undefined locations are lumped together with gradients without undefined locations
See this test case:

```rs
#[vello_test]
    fn demo(ctx: &mut impl Renderer) {
        use vello_common::color::{DynamicColor, palette::css::PURPLE};
        use peniko::ColorStop;
        use smallvec::smallvec;

        // 1) This first draw op will put the gradient into the cache.
        let grad1 = Gradient {
            kind: Radial(RadialGradientPosition {
                start_center: Point::new(-200.0, -200.0),
                start_radius: 5.0,
                end_center: Point::new(-200.0, -200.0),
                end_radius: 35.0,
            }),
            stops: stops_blue_green_red_yellow(),
            ..Default::default()
        };
        ctx.set_paint(grad1);
        ctx.fill_rect(&Rect::new(-250.0, -250.0, -150.0, -150.0));

        // 2) This second draw op should _not_ result in a cache hit, because the gradient
        // can have undefined locations. Therefore, a different LUT will be generated which adds a
        // final transparent stop. Therefore, this gradient must be treated differently
        // than the first one.
        let grad2 = Gradient {
            kind: Radial(RadialGradientPosition {
                start_center: Point::new(30.0, 50.0),
                start_radius: 5.0,
                end_center: Point::new(70.0, 50.0),
                end_radius: 20.0,
            }),
            stops: stops_blue_green_red_yellow(),
            ..Default::default()
        };
        ctx.set_paint(grad2);
        ctx.fill_rect(&Rect::new(10.0, 10.0, 90.0, 90.0));
        
        // 3) In case 2) was not fulfilled, the transparent pixel will instead lan on the first
        // LUT entry of this gradient, which is purple.
        let grad3 = Gradient {
            kind: Radial(RadialGradientPosition {
                start_center: Point::new(-200.0, -200.0),
                start_radius: 5.0,
                end_center: Point::new(-200.0, -200.0),
                end_radius: 35.0,
            }),
            stops: ColorStops(smallvec![
            ColorStop { offset: 0.0, color: DynamicColor::from_alpha_color(PURPLE) },
            ColorStop { offset: 1.0, color: DynamicColor::from_alpha_color(PURPLE) },
        ]),
            ..Default::default()
        };
        ctx.set_paint(grad3);
        ctx.fill_rect(&Rect::new(-250.0, -250.0, -150.0, -150.0));
    }
```
Left is how it should look like, right is how it actually looks like.
Image

LaurenzV:
Closes #1528. 

# Context
Nearly all gradients have a clearly defined value at all position, the only exception being very specific degenerate radial gradients. The behavior in this case is that all positions that are undefined should yield a transparent pixel.

Previously, we had handling for this in the CPU shader, where at sampling time we would explicitly mask out such positions. As part of https://github.com/linebender/vello/pull/1301, a number of performance changes were introduced for gradient handling. Among that, a trick was used that encodes an additional transparent padding pixel in the gradient LUT, allowing to handle that case much better in Vello CPU.

While this is a valid approach, it has two problems:
1) It destroys the assumption that the length of gradient LUTs are a multiple of 2. While for Vello CPU this doesn't matter and it currently also doesn't matter for Vello Hybrid, it could be a very useful property to have in the future, especially if we decide to change how gradients are uploaded in Vello hybrid.
2) It requires us to cache stops based on whether the underlying gradient has potentially undefined locations, which is also unfortunate and the root cause for #1528.

# Solution

This PR revert that change (while still keeping in-tact the other performance changes from that PR), such that gradients LUT's are always a multiple of 2 again and don't need to know anything about undefined locations. From my benchmarks, this doesn't have any notable performance impact for the normal case. In certain cases, it seems to even slightly improve performance. While I haven't measured it, this is likely to make undefined gradients about 2x slower, but since those are very rare edge cases we don't really care about apart from for correctness, I think this is very much a worthy change. Here you can find the change in performance before vs. after this PR:

```
fine/gradient/linear/opaque_u8_neon
                        time:   [359.08 ns 360.57 ns 362.26 ns]
                        change: [-5.0401% -4.6310% -4.2254%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  3 (3.00%) high mild
  5 (5.00%) high severe

                        time:   [475.66 ns 476.60 ns 477.56 ns]
                        change: [-1.3570% -1.0840% -0.8030%] (p = 0.00 < 0.05)
  4 (4.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe

fine/gradient/radial/opaque_conical_u8_neon
                        time:   [577.07 ns 578.02 ns 579.05 ns]
                        change: [-1.2222% -0.8150% -0.3570%] (p = 0.00 < 0.05)
  1 (1.00%) low mild
  5 (5.00%) high mild
  6 (6.00%) high severe

fine/gradient/sweep/opaque_u8_neon
                        time:   [847.68 ns 849.27 ns 850.89 ns]
                        change: [-0.0226% +0.2834% +0.5911%] (p = 0.07 > 0.05)
  1 (1.00%) low mild
  1 (1.00%) high severe

fine/gradient/extend/pad_u8_neon
                        time:   [346.25 ns 346.96 ns 347.77 ns]
                        change: [-0.2873% +0.0627% +0.3827%] (p = 0.72 > 0.05)
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

fine/gradient/extend/repeat_u8_neon
                        time:   [368.20 ns 369.45 ns 371.00 ns]
                        change: [-4.5087% -4.1494% -3.8349%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe

fine/gradient/extend/reflect_u8_neon
                        time:   [417.63 ns 418.73 ns 420.20 ns]
                        change: [-6.1869% -5.5205% -4.6695%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  1 (1.00%) low mild
  8 (8.00%) high mild
  6 (6.00%) high severe

fine/gradient/many_stops_u8_neon
                        time:   [479.07 ns 480.13 ns 481.45 ns]
                        change: [-0.7523% -0.4523% -0.0884%] (p = 0.01 < 0.05)
  1 (1.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe

fine/gradient/transparent_u8_neon
                        time:   [459.30 ns 461.39 ns 463.68 ns]
                        change: [-0.4103% +0.0785% +0.6946%] (p = 0.77 > 0.05)
  1 (1.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe

```

### DIFF
--- a/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
@@ -462,17 +462,17 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
             
             // For radial gradient, calculate distance from center
             let gradient_result = calculate_radial_gradient(grad_pos, radial_gradient);
-            if gradient_result.is_valid {
-                let gradient_color = sample_gradient_lut(
-                    gradient_result.t_value, 
-                    radial_gradient.extend_mode, 
-                    radial_gradient.gradient_start, 
-                    radial_gradient.texture_width
-                );
-                final_color = alpha * gradient_color;
-            } else {
-                final_color = vec4<f32>(0.0, 0.0, 0.0, 0.0);
-            }
+            let gradient_color = sample_gradient_lut(
+                gradient_result.t_value, 
+                radial_gradient.extend_mode, 
+                radial_gradient.gradient_start, 
+                radial_gradient.texture_width
+            );
+            final_color = select(
+                vec4<f32>(0.0, 0.0, 0.0, 0.0),
+                alpha * gradient_color,
+                gradient_result.is_valid
+            );
         } else if paint_type == PAINT_TYPE_SWEEP_GRADIENT {
             let paint_tex_idx = in.paint_and_rect_flag & PAINT_TEXTURE_INDEX_MASK;
             let sweep_gradient = unpack_sweep_gradient(paint_tex_idx);


sparse_strips/vello_common/src/encode.rs
```diff
@@ -764,13 +765,16 @@ pub struct GradientCacheKey {
     pub interpolation_cs: ColorSpaceTag,
     /// Hue direction used for interpolation.
     pub hue_direction: HueDirection,
+    /// Whether the gradient itself can have undefined
+    pub has_undefined: bool,
```
grebmeg: One alternative worth considering: instead of adding `has_undefined` to the cache key, you could handle undefined t-values entirely at the rendering site and remove the extra LUT entry altogether. This would also address the concern you raised about breaking the power-of-2 LUT size property.

The idea:

- LUT stays pure: `GradientLut` never appends a transparent entry. `determine_lut_size` always returns a power-of-2, and the LUT stays that size. No `has_undefined` field on `GradientLut` or `GradientCacheKey`.
- Same stops → same LUT → same cache entry: Since the LUT content is now independent of the gradient's geometry, the cache key only needs (`stops`, `colorspace`, `hue_direction`) — no `has_undefined` needed, and the cache collision simply can't happen.
- CPU painter handles `NaN` at output: The existing split between `gradient_painter` (normal) and `gradient_painter_with_undefined` (radial with undefined regions) already exists. The undefined path uses a `MaskedGradientPainter` that detects `NaN` t-values via `pos.simd_eq(pos)`, clamps the index to 0 (harmless — the value is discarded), samples the LUT, then zeros out the RGBA output for undefined pixels. Same number of SIMD operations, just operating on the output color instead of the index.
GPU shader returns transparent directly: `sample_gradient_lut` early-returns `vec4(0.0)` when `!is_valid` instead of indexing into a sentinel entry.
- This keeps the LUT as a clean power-of-2 color ramp, moves the "undefined → transparent" decision to where it semantically belongs (the renderer), and eliminates the need to distinguish gradients by geometry in the cache.

This branch [`laurenz/grad_cache_exp`](https://github.com/linebender/vello/compare/laurenz/grad_cache...laurenz/grad_cache_exp) is where I’ve experimented with the idea above.

sparse_strips/vello_common/src/encode.rs
grebmeg, line 786:
```suggestion
            && self.has_undefined == other.has_undefined
```

sparse_strips/vello_common/src/encode.rs
```diff
@@ -764,13 +765,16 @@ pub struct GradientCacheKey {
     pub interpolation_cs: ColorSpaceTag,
     /// Hue direction used for interpolation.
     pub hue_direction: HueDirection,
+    /// Whether the gradient itself can have undefined
```
grebmeg: Nit: the sentence is cut off. Consider clarifying what "undefined" refers to.

sparse_strips/vello_common/src/encode.rs
LaurenzV: Yes, I also think it would be better, as I mentioned in the PR description. I think this even is how we had it originally, I believe Valerie changed it at some point to make the CPU version faster, but now that we also use it for the GPU it might be worth it to revert that optimization.

Let me know your preference, if you prefer that I can repurpose the PR to make that change.

LaurenzV:
@grebmeg Repurposed this PR to remove undefined handling from gradient LUT, I hope that matches what you had in mind!

sparse_strips/vello_common/src/encode.rs
LaurenzV: Done!

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -463,14 +462,17 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
+                let gradient_color = sample_gradient_lut(
+                    gradient_result.t_value, 
+                    radial_gradient.extend_mode, 
+                    radial_gradient.gradient_start, 
+                    radial_gradient.texture_width
+                );
+                final_color = alpha * gradient_color;
+            } else {
+                final_color = vec4<f32>(0.0, 0.0, 0.0, 0.0);
+            }
```
grebmeg: This if/else introduces SIMT divergence when threads in a wavefront disagree on `is_valid`, the GPU masks off inactive lanes and executes both paths sequentially. Since `sample_gradient_lut` uses `textureLoad`, the texture fetch runs for all lanes regardless of the branch, so the branch doesn't actually skip any meaningful work.

```suggestion
            let gradient_color = sample_gradient_lut(
                gradient_result.t_value, 
                radial_gradient.extend_mode, 
                radial_gradient.gradient_start, 
                radial_gradient.texture_width
            );
            final_color = select(
                vec4<f32>(0.0, 0.0, 0.0, 0.0),
                alpha * gradient_color,
                gradient_result.is_valid
            );
```

Practically, the savings are modest here since the else path is trivial, but it's a good habit to prefer branchless patterns in fragment-rate shader code as it keeps the pipeline clean and avoids edge cases where the compiler can't reorder instructions across branch boundaries.

grebmeg (approved):

PR #1532, opened 2026-03-24
vello_hybrid: Use a staging belt to reuse allocations in wgpu backend

LaurenzV:
From my understanding: By default, every time we call `write_texture`, wgpu will first copy our data into a _newly_ allocated CPU-based staging buffer, and only then upload the data from the staging buffer to the GPU.

By using a staging belt, we can reuse the temporary buffers used by wgpu across frames, which eliminates unnecessary memory copies from the first frame onward.

The results are indeed much better, and the code (in my opinion) also becomes a lot simpler!

Before: For 1200 frames of GhostScript tiger, around 479ms are spent in `render`
image

After: For 1200 frames of GhostScript tiger, only 155ms are spent in `render`
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -351,9 +351,10 @@ impl Renderer {
         self.programs.prepare(
             device,
             queue,
+            encoder,
             &mut self.gradient_cache,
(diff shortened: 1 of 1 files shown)


LaurenzV:
Fair enough! I'll think about whether I can come up with something.

LaurenzV:
Superseded by #1915.

PR #1534, opened 2026-03-24
Upgrade to wgpu v29. Reduce MSRV to 1.88.

nicoburns:
Upgrade Vello and Vello Hybrid to wgpu v29. Also reduces MSRV to 1.88 as wgpu lowered theirs.

https://github.com/gfx-rs/wgpu/releases/tag/v29.0.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ env:
   # If the compilation fails, then the version specified here needs to be bumped up to reality.
   # Be sure to also update the rust-version property in the workspace Cargo.toml file,
   # plus all the README.md files of the affected packages.
-  RUST_MIN_VER: "1.92"
+  RUST_MIN_VER: "1.88"
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
   RUST_MIN_VER_PKGS: "-p vello -p vello_encoding -p vello_shaders -p vello_api -p vello_common -p vello_cpu -p vello_hybrid"
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ VELLO_STATIC_LOG="vello=trace" VELLO_STATIC_ARGS="--test-scenes" cargo apk run -
 
 ## Minimum supported Rust Version (MSRV)
 
-This version of Vello has been verified to compile with **Rust 1.86** and later.
+This version of Vello has been verified to compile with **Rust 1.88** and later.
 
 Future versions of Vello might increase the Rust version requirement.
 It will not be treated as a breaking change and as such can even happen with small patch releases.
--- a/examples/simple/src/main.rs
+++ b/examples/simple/src/main.rs
@@ -16,7 +16,7 @@ use winit::event::WindowEvent;
 use winit::event_loop::{ActiveEventLoop, EventLoop};
 use winit::window::Window;
 
-use vello::wgpu;
+use vello::wgpu::{self, CurrentSurfaceTexture};
 
 #[derive(Debug)]
 enum RenderState {
@@ -95,12 +95,12 @@ impl ApplicationHandler for SimpleVelloApp {
         event: WindowEvent,
     ) {
         // Only process events for our window, and only when we have a surface.
-        let (surface, valid_surface) = match &mut self.state {
+        let (surface, valid_surface, window) = match &mut self.state {
             RenderState::Active {
                 surface,
                 valid_surface,
                 window,
-            } if window.id() == window_id => (surface, valid_surface),
+            } if window.id() == window_id => (surface, valid_surface, window),
             _ => return,
         };
 
@@ -158,10 +158,22 @@ impl ApplicationHandler for SimpleVelloApp {
                     .expect("failed to render to surface");
 
                 // Get the surface's texture
-                let surface_texture = surface
-                    .surface
-                    .get_current_texture()
-                    .expect("failed to get surface texture");
+                let surface_texture = match surface.surface.get_current_texture() {
+                    CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
+                    CurrentSurfaceTexture::Outdated | CurrentSurfaceTexture::Suboptimal(_) => {
+                        self.context.configure_surface(surface);
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Occluded | CurrentSurfaceTexture::Timeout => {
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Lost => panic!("Surface was lost"),
+                    CurrentSurfaceTexture::Validation => {
+                        panic!("Validation error getting surface")
+                    }
+                };
 
                 // Perform the copy
                 let mut encoder =
--- a/examples/with_winit/src/lib.rs
+++ b/examples/with_winit/src/lib.rs
@@ -43,7 +43,7 @@ use winit::dpi::LogicalSize;
 use winit::event_loop::EventLoop;
 use winit::window::{Window, WindowAttributes};
 
-use vello::wgpu::{self, PipelineCache};
+use vello::wgpu::{self, CurrentSurfaceTexture, PipelineCache};
 
 #[cfg(not(any(target_arch = "wasm32", target_os = "android")))]
 mod hot_reload;
@@ -632,10 +632,22 @@ impl ApplicationHandler<UserEvent> for VelloApp {
                 drop(render_span);
 
                 let texture_span = tracing::trace_span!("Blitting to surface").entered();
-                let surface_texture = surface
-                    .surface
-                    .get_current_texture()
-                    .expect("failed to get surface texture");
+                let surface_texture = match surface.surface.get_current_texture() {
+                    CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
+                    CurrentSurfaceTexture::Outdated | CurrentSurfaceTexture::Suboptimal(_) => {
+                        self.context.configure_surface(surface);
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Occluded | CurrentSurfaceTexture::Timeout => {
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Lost => panic!("Surface was lost"),
+                    CurrentSurfaceTexture::Validation => {
+                        panic!("Validation error getting surface")
+                    }
+                };
                 // Perform the copy
                 // (TODO: Does it improve throughput to acquire the surface after the previous texture render has happened?)
                 let mut encoder =
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -13,7 +13,7 @@ You can find its changes [documented below](#007---2026-03-24).
 
 ## [Unreleased]
 
-This release has an [MSRV][] of 1.98.
+This release has an [MSRV][] of 1.88.
 
 ### Optimized
 
--- a/sparse_strips/vello_common/CHANGELOG.md
+++ b/sparse_strips/vello_common/CHANGELOG.md
@@ -13,7 +13,7 @@ You can find its changes [documented below](#007---2026-03-24).
 
 ## [Unreleased]
 
-This release has an [MSRV][] of 1.92.
+This release has an [MSRV][] of 1.98.
 
 ### Optimized
 
--- a/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
@@ -612,7 +612,7 @@ pub async fn render_scene(scene: Scene, width: u16, height: u16) {
     };
     let surface_texture = match surface.get_current_texture() {
         CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
-        _ => panic!("Error getting surface"),
+        e => panic!("Error getting initial surface: {e:?}"),
     };
     let surface_texture_view = surface_texture
         .texture
--- a/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
@@ -20,6 +20,18 @@ use vello_example_scenes::{AnyScene, image::ImageScene};
 use vello_hybrid::{AtlasConfig, Pixmap, RenderSettings, RenderTargetConfig, Renderer, Scene};
 use wasm_bindgen::prelude::*;
 use web_sys::{Event, HtmlCanvasElement, KeyboardEvent, MouseEvent, WheelEvent};
+use wgpu::{
+    CurrentSurfaceTexture,
+    rwh::{DisplayHandle, HandleError, HasDisplayHandle},
+};
+
+#[derive(Debug)]
+struct OurDisplayHandle;
+impl HasDisplayHandle for OurDisplayHandle {
+    fn display_handle(&self) -> Result<DisplayHandle<'_>, HandleError> {
+        Ok(DisplayHandle::web())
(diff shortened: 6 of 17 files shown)


README.md
```diff
@@ -222,7 +222,7 @@ VELLO_STATIC_LOG="vello=trace" VELLO_STATIC_ARGS="--test-scenes" cargo apk run -
 
 ## Minimum supported Rust Version (MSRV)
 
-This version of Vello has been verified to compile with **Rust 1.86** and later.
+This version of Vello has been verified to compile with **Rust 1.88** and later.
```
nicoburns, line 225: This should have been updated to 1.92 with the wgpu v28 bump but got missed

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -174,7 +186,19 @@ impl AppState {
...
 
-        let surface_texture = self.renderer_wrapper.surface.get_current_texture().unwrap();
+        let surface_texture = match self.renderer_wrapper.surface.get_current_texture() {
+            CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
+            CurrentSurfaceTexture::Occluded
+            | CurrentSurfaceTexture::Timeout
+            | CurrentSurfaceTexture::Outdated
+            | CurrentSurfaceTexture::Suboptimal(_) => {
+                return;
+            }
```
LaurenzV, line 199: Should we really silently return in case those happened? Seems to me like, especially for the example, it's fine to just panic here

sparse_strips/vello_hybrid/examples/winit/src/main.rs
```diff
@@ -326,10 +327,22 @@ impl ApplicationHandler for App<'_> {
                     height: surface.config.height,
                 };
 
-                let surface_texture = surface
-                    .surface
-                    .get_current_texture()
-                    .expect("failed to get surface texture");
+                let surface_texture = match surface.surface.get_current_texture() {
```
LaurenzV, line 352: Same here.

vello/src/util.rs
```diff
@@ -34,7 +34,9 @@ impl RenderContext {
         let flags = wgpu::InstanceFlags::from_build_config().with_env();
         let memory_budget_thresholds = wgpu::MemoryBudgetThresholds::default();
         let backend_options = wgpu::BackendOptions::from_env_or_default();
-        let instance = Instance::new(&wgpu::InstanceDescriptor {
+        let instance = Instance::new(wgpu::InstanceDescriptor {
+            // TODO: consider allowing users to pass display handle when constructing a RenderContext
```
LaurenzV, line 38: Will this cause regressions for existing users if we don't allow this now?

examples/with_winit/src/lib.rs
```diff
@@ -633,10 +633,22 @@ impl ApplicationHandler<UserEvent> for VelloApp {
                 drop(render_span);
 
                 let texture_span = tracing::trace_span!("Blitting to surface").entered();
-                let surface_texture = surface
-                    .surface
-                    .get_current_texture()
-                    .expect("failed to get surface texture");
+                let surface_texture = match surface.surface.get_current_texture() {
```
LaurenzV, line 635: Same as before.

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
waywardmonkeys: People will copy the example.

LaurenzV: But then is the best thing to do really to just return? I don't know, I'm not super familiar with wgpu API. 😄

nicoburns: It's important not to panic here (which we were previously doing!). These errors can and do happen relatively often (we've gotten bug reports for Blitz), but they are transient/recoverable errors so the recommended thing to do is skip the current frame and immediately retry by rendering another frame.

See the WGPU v29 release notes, where they've intentionally changed this API not to use `Result` to push people towards not just calling `unwrap`/`expect` on these errors. https://github.com/gfx-rs/wgpu/releases/tag/v29.0.0

For some of those cases, I'm supposed to "reconfigure the surface", which I have done in the native cases, but I wasn't sure how to do that on the webgl backend. And for the `Lost` case you're apparently supposed to recreate the device in some cases (but I think this happens much less often so I'm happy to just continue panicking here for now).

vello/src/util.rs
nicoburns: My understanding is that this primarily impacts GL backends. Which I think is not relevant at all for Vello Classic. It could potentially affect Vello Hybrid (but the WebGL backend has a separate setup where we do already pass the "web display handle").

So I believe it would be "people running Vello Hybrid with OpenGL on native platforms" that would need this.

PR for this is https://github.com/gfx-rs/wgpu/pull/8782

nicoburns:
Very confused what's going on with CI here: the errors seem to reference code which doesn't exist (either locally or in the Github view of the file). And the commands are passing locally :/ And I've also tried busting the cache (by changing the cache key), which seems to be working because it says "no cache found", but that doesn't seem to help.

@xStrom @DJMcNab Any idea what might be going on here? Am I being an idiot?

DJMcNab:
On mobile, so hard to check. Best guess is that it merges clean, so CI is running on the combined merge.

That behaviour sort of makes sense, but GitHub explains it exceedingly poorly; you have to infer what's going on.

So to rule that out, I'd try a rebase (or a fast forward merge? Idk, I always rebase)

.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ env:
   # If the compilation fails, then the version specified here needs to be bumped up to reality.
   # Be sure to also update the rust-version property in the workspace Cargo.toml file,
   # plus all the README.md files of the affected packages.
```
DJMcNab, line 10: Follow-up then is to add CHANGELOG.md (all 4? of them) to this list.

Needn't be in this PR.

LaurenzV:
I think you probably just haven't rebased onto newest main, which contains the new opaque/alpha strips path.

examples/simple/src/main.rs
```diff
@@ -158,10 +158,22 @@ impl ApplicationHandler for SimpleVelloApp {
...
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Occluded | CurrentSurfaceTexture::Timeout => {
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Lost => panic!("Surface was lost"),
+                    CurrentSurfaceTexture::Validation => {
+                        panic!("Valdidation error getting surface")
```
DJMcNab: Can you also submit the typo to typos after fixing it :)?

examples/simple/src/main.rs
```diff
@@ -158,10 +158,22 @@ impl ApplicationHandler for SimpleVelloApp {
                     .expect("failed to render to surface");
 
                 // Get the surface's texture
-                let surface_texture = surface
-                    .surface
-                    .get_current_texture()
-                    .expect("failed to get surface texture");
+                let surface_texture = match surface.surface.get_current_texture() {
+                    CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
+                    CurrentSurfaceTexture::Outdated | CurrentSurfaceTexture::Suboptimal(_) => {
```
DJMcNab, line 163: There are reasonable arguments for still drawing to a suboptimal surface. I'm worried about the case where suboptimal means "use a different device", which would mean we would loop here (this is unvalidated, to be clear, as I'm on mobile)

Given that this is an example, I don't really care about it.

examples/with_winit/src/lib.rs
```diff
@@ -632,10 +632,22 @@ impl ApplicationHandler<UserEvent> for VelloApp {
...
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Occluded | CurrentSurfaceTexture::Timeout => {
+                        window.request_redraw();
+                        return;
+                    }
+                    CurrentSurfaceTexture::Lost => panic!("Surface was lost"),
+                    CurrentSurfaceTexture::Validation => {
+                        panic!("Valdidation error getting surface")
```
DJMcNab: Same typo

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -32,9 +44,9 @@ impl RendererWrapper {
     async fn new(canvas: HtmlCanvasElement) -> Self {
         let width = canvas.width();
         let height = canvas.height();
-        let instance = wgpu::Instance::new(&wgpu::InstanceDescriptor {
+        let instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
             backends: wgpu::Backends::GL,
-            ..Default::default()
+            ..wgpu::InstanceDescriptor::new_with_display_handle(Box::new(OurDisplayHandle))
```
DJMcNab, line 49: Why not `Box::new(DisplayHandle::web())`? I don't think this example is quite so precious about memory usage. We could give a hint comment if we really wanted to.

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -177,7 +189,19 @@ impl AppState {
...
+            CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
+            CurrentSurfaceTexture::Occluded
+            | CurrentSurfaceTexture::Timeout
+            | CurrentSurfaceTexture::Outdated
+            | CurrentSurfaceTexture::Suboptimal(_) => {
+                return;
+            }
+            CurrentSurfaceTexture::Lost => panic!("Surface was lost"),
+            CurrentSurfaceTexture::Validation => {
+                panic!("Valdidation error getting surface")
```
DJMcNab: Same typo.

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
```diff
@@ -586,7 +610,10 @@ pub async fn render_scene(scene: Scene, width: u16, height: u16) {
         width: width as u32,
         height: height as u32,
     };
-    let surface_texture = surface.get_current_texture().unwrap();
+    let surface_texture = match surface.get_current_texture() {
+        CurrentSurfaceTexture::Success(surface_texture) => surface_texture,
+        _ => panic!("Error getting surface"),
```
DJMcNab:
```suggestion
        e => panic!("Error getting initial surface: {e}"),
```
Maybe?
Non-blocking.

sparse_strips/vello_hybrid/examples/winit/src/render_context.rs
```diff
@@ -73,7 +73,9 @@ impl RenderContext {
         let flags = wgpu::InstanceFlags::from_build_config().with_env();
         let memory_budget_thresholds = wgpu::MemoryBudgetThresholds::default();
         let backend_options = wgpu::BackendOptions::from_env_or_default();
-        let instance = Instance::new(&wgpu::InstanceDescriptor {
+        let instance = Instance::new(wgpu::InstanceDescriptor {
+            // TODO: consider allowing users to pass a display handle when constructing a RenderContext
```
DJMcNab: I think render_context is "private" code, so it doesn't matter if users can or can't do this. That is, if a user needs this, they can add it themselves.

So to clarify, I'd probably just drop this comment, but I'm not blocking on it either way.

vello/src/util.rs
```diff
@@ -133,7 +135,7 @@ impl RenderContext {
         self.configure_surface(surface);
     }
 
-    fn configure_surface(&self, surface: &RenderSurface<'_>) {
+    pub fn configure_surface(&self, surface: &RenderSurface<'_>) {
```
DJMcNab, line 138: Is this finally my impetus for killing this file entirely?

Either way, that shouldn't happen in this PR...

DJMcNab (commented):
All seems reasonable to me, although this is a mobile only review, and I've not yet read the upstream changelog, so won't review it.

If I've not reviewed by 16:30ish Sydney time tomorrow, feel free to ping again here and/or on Zulip.

examples/simple/src/main.rs
nicoburns: https://github.com/crate-ci/typos/issues/1531#issuecomment-4288710066

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
nicoburns: Because that wasn't valid until `wgpu` v29.0.1! But we should be able to change that now :)

vello/src/util.rs
nicoburns: I do now maintain https://docs.rs/wgpu_context as part of AnyRender which can potentially serve as a replacement (it would be annoying if `vello` depended on this, but a dev-dependency would be fine).

sparse_strips/vello_hybrid/examples/winit/src/render_context.rs
nicoburns: Ah, yeah I see that this is just example code.

examples/simple/src/main.rs
nicoburns: Yeah, not sure. Either way, this seems better than panicking (which is what we were doing before). Handling the "should use a different device case" seems like it would be good to do in future (also for `CurrentSurfaceTexture::Lost`), but is probably quite rare in practice?

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
nicoburns: Hmm... actually this still isn't working, even with v29.0.1. I've reinstated this code. Issue is that `DisplayHandle::web()` isn't `Send`/`Sync`. I did have discussions with wgpu maintainers about removing this bound for WASM (which wgpu already does elsewhere), but I don't think this has actually happened yet.

.github/workflows/ci.yml
nicoburns: done.

nicoburns:
> On mobile, so hard to check. Best guess is that it merges clean, so CI is running on the combined merge.
> That behaviour sort of makes sense, but GitHub explains it exceedingly poorly; you have to infer what's going on.
> So to rule that out, I'd try a rebase (or a fast forward merge? Idk, I always rebase)

(this was it. I had rebased on my local `main`, but I didn't have the latest locally - thanks for the help!)

sparse_strips/vello_hybrid/examples/wgpu_webgl/src/lib.rs
DJMcNab: Ah, apologies. Thanks for checking! If there's an upstream issue, we should link to it in a comment (not checked if you're already done this).

sparse_strips/vello_common/CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ You can find its changes [documented below](#007---2026-03-24).
 
 ## [Unreleased]
 
-This release has an [MSRV][] of 1.92.
+This release has an [MSRV][] of 1.98.
```
taj-p:
```suggestion
This release has an [MSRV][] of 1.88.
```

nicoburns:
> If I've not reviewed by 16:30ish Sydney time tomorrow, feel free to ping again here and/or on Zulip.

@DJMcNab ping!

DJMcNab (approved):
I've not re-reviewed this - I'll aim to post-review next week. I've got an extra long weekend, so that won't be before Tuesday. But happy to land to let dependant PRs happen.

nicoburns:
> I've not re-reviewed this - I'll aim to post-review next week. I've got an extra long weekend, so that won't be before Tuesday. But happy to land to let dependant PRs happen.

Cool. In that case I will merge now to avoid conflicts.

PR #1539, opened 2026-03-26
Import glifo into the vello workspace

laurenz-canva:
Not hooked up yet into vello, just the raw import. First commit is a 1:1 copy-paste of the crate, all subsequent commits finetune it to work with the workspace structure.

This is the actual diff: https://github.com/linebender/vello/compare/96c94822...9b070fd

Accompanying parley PR: https://github.com/linebender/parley/pull/586

### DIFF
--- a/glifo/Cargo.toml
+++ b/glifo/Cargo.toml
@@ -9,7 +9,7 @@ rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
 
-publish = false
+publish = false # When enabling, add it to the MSRV package list in CI
 
 [package.metadata.docs.rs]
 all-features = true
--- a/glifo/LICENSE-MIT
+++ b/glifo/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright 2025 the Parley Authors
+Copyright 2025 the Vello Authors
 
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated
--- a/glifo/src/lib.rs
+++ b/glifo/src/lib.rs
@@ -1,7 +1,7 @@
 // Copyright 2026 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! Parley Draw provides APIs for efficiently rendering glyphs and paint styles like underline.
+//! Glifo provides APIs for efficiently rendering glyphs and paint styles like underline.
 //!
 //! ## Features
 //!


laurenz-canva:
I had to update the copyright headers, I presume this is okay?

xStrom:
> I had to update the copyright headers, I presume this is okay?

To Vello? Yeah I think it's ok in this case, because there were very few changes made in the Parley repo and all from known set of people. Otherwise the header verification supports having multiple authorships in the headers like "Copyright 2026 the Vello Authors and the Parley Authors"

sparse_strips/glifo/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+//! - `std` (enabled by default): Get floating point functions from the standard library
+//!   (likely using your target's libc).
+//! - `libm`: Use floating point implementations from `libm`.
+//! - `vello_cpu` (enabled by default): Implements `GlyphRenderer` for Vello CPU's `RenderContext`.
+//! - `png`: Enables PNG support for drawing bitmap glyphs.
+//!
+//! At least one of `std` and `libm` is required.
+
+// TODO: Re-enable once we start proper work on glifo.
+#![allow(clippy::all, reason = "TODO: fix clippy lints")]
```
xStrom: What sort of timeline are you imagining for it? Such a broad `allow` should be very short term, otherwise things will go out of control.

laurenz-canva: I can remove it and fix it if you want, I think it just was a few clippy lints that needed fixing, but I wanted to keep the diff minimal.

xStrom: I'm ok with it staying in for this specific PR. Sounds like you're fine with it being removed soon as a follow-up and that satisfies my concerns.

nicoburns: Can I suggest scoping it to `clippy::collapsible_if`. All of the failing lints are those ones. And we may want to keep that ignore until the impending MSRV reduction, otherwise we'll need to change the code back.

glifo/src/lib.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright 2026 the Vello Authors
```
nicoburns, line 1: Could we have the crate at the top-level rather than under `sparse_strips`? Glifo is not sparse-strips specific.

glifo/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Parley Draw provides APIs for efficiently rendering glyphs and paint styles like underline.
```
xStrom:
```suggestion
//! Glifo provides APIs for efficiently rendering glyphs and paint styles like underline.
```

glifo/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright 2025 the Parley Authors
```
xStrom: As you changed the other headers, might as well change this too.

```suggestion
Copyright 2025 the Vello Authors
```

glifo/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+version = "0.0.0"
+description = "Glifo provides APIs for efficiently rendering text."
+keywords = ["text", "layout"]
+categories = ["gui", "graphics"]
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+
+publish = false
```
xStrom: Let's add a note so we won't forget to add this to the MSRV test when it's time.

```suggestion
publish = false # When enabling, add it to the MSRV package list in CI
```

LaurenzV:
I just copy-pasted the whole folder!

LaurenzV:
@grebmeg If you are happy please feel free to merge it yourself, otherwise I'll do it tomorrow.

PR #1541, opened 2026-03-26
vello_hybrid: Fix filter layers that contain a nested clip layer

LaurenzV:
The problem is as follows:

If we have a single layer with clip and filter, i.e.:
```
push_layer(clip, filter)
do_something()
pop_layer
```

Due to reasons, in the current coarse rasterization logic, this will be represented as (it's possible this can be improved, but unfortunately simplifying coarse rasterization will be hard until we've figured whether we can remove the "old" clipping algorithm):
```
push_layer (filter)
push_layer (clip)
do_something
pop_layer (clip)
pop_layer (filter)
```

However, the problem is that, for filter layers, clip paths need to be applied _after_ the filter was applied. Therefore, when rendering filters, we have this lookahead to check whether the filter layer is followed by a clip path.

The problem is that if we have the following sequence:
```
push_layer(filter)
push_layer(clip)
do_something
pop_layer (clip)
pop_layer (filter)
```

This will get the exact same representation as the first sequence, but semantically it's different because the clip path is not on the same layer as the filter (i.e. the clip path _should_ be applied before the filter is applied). However, due to the fact that they have the same representation, we still apply the lookahead fix.

The solution to this problem is that we apply the lookahead only if the layer id of the nested clip layer has the same ID as the surrounding filter layer.

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
@@ -338,6 +338,8 @@ impl SingleThreadedDispatcher {
                     // The layer was already rendered for filtering, but we skip compositing
                     // since this tile is entirely clipped out.
                     // (PushZeroClip only appears for clipped filter layers)
+                    // See https://github.com/linebender/vello/pull/1541/ for why we
+                    // add the ID check.
                     Some(Cmd::PushZeroClip(id)) if *id == *child_layer_id => {
                         // If we have a zero-clip, it means that the whole layer should not be drawn.
                         // Therefore, we want to skip to the very end so that only `PopBuf` will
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -347,7 +347,7 @@ impl SingleThreadedDispatcher {
                     }
 
                     // Partial clip: push the clip buffer, then composite the filtered layer
-                    Some(Cmd::PushBuf(LayerKind::Clip(_), _)) => {
+                    Some(Cmd::PushBuf(LayerKind::Clip(id), _)) if *id == *child_layer_id => {
```
LaurenzV, line 352: Note to self: Add comment linking to this PR

LaurenzV:
@grebmeg Post-merge comments still welcome!

PR #1546, opened 2026-03-27
vello_hybrid: Avoid rendering into an intermediate framebuffer

LaurenzV:
Confirmed that this gives a good speed boost on my Samsung Galaxy Tab A6 for low-complexity scenes! I also ran the `native_webgl` and `wgpu_webgl` examples and they seem to run fine.

In my view, there were two ways of solving this problem:
- Forcing the user to provide a root transform on `Scene` when rendering with native WebGL. While this would require less code changes, it's IMO pretty ugly because it's not something that the user should have to handle manually.
- Basically what this PR implements, while it requires a bit more code changes, it completely abstracts it away from the user, which IMO is much better!

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -125,6 +125,23 @@ impl WebGlRenderer {
             .dyn_into::<WebGl2RenderingContext>()
             .expect("Context to be a WebGL2 context");
 
+        #[cfg(debug_assertions)]
+        {
+            // If a WebGL context already exists on this canvas, it will be returned instead of
+            // creating a new one with the correct context_options set.
+            // See this comment for why we still care about non-antialiased context:
+            // https://github.com/linebender/vello/pull/1546/changes#r3008692535
+            let context_attributes = gl.get_context_attributes().unwrap();
+            let antialias = js_sys::Reflect::get(&context_attributes, &"antialias".into())
+                .unwrap()
+                .as_bool()
+                .unwrap();
+            debug_assert!(
+                !antialias,
+                "WebGL context must be created with `antialias: false` for vello_hybrid to work correctly."
+            );
+        }
+
         let max_texture_dimension_2d = get_max_texture_dimension_2d(&gl);
         let total_slots: usize = (max_texture_dimension_2d / u32::from(Tile::HEIGHT)) as usize;
         let image_cache = ImageCache::new_with_config(settings.atlas_config);


sparse_strips/vello_sparse_tests/tests/renderer.rs
```diff
@@ -821,6 +821,15 @@ impl Renderer for HybridRenderer {
...
             .unwrap();
+
+        // WebGL framebuffers are y-up, so we need to invert the rows.
+        let row_bytes = width as usize * 4;
+        let height = height as usize;
+        for y in 0..height / 2 {
+            let (a, b) = pixels.split_at_mut((height - 1 - y) * row_bytes);
+            a[y * row_bytes..(y + 1) * row_bytes].swap_with_slice(&mut b[..row_bytes]);
+        }
+
```
LaurenzV, line 832: From my understanding, this code would have been needed in the first place, and the fact we didn't need it is purely accidental. Here, we were setting `READ_FRAMEBUFFER` to the view framebuffer:

https://github.com/linebender/vello/blob/5796226bfafd75fb121e938536ff74bb80a6114f/sparse_strips/vello_hybrid/src/render/webgl.rs#L193-L195

But we were never setting it to the final blitted framebuffer. Therefore, the tests were reading the non-flipped version, and when reading it back to the CPU pixmap it therefore wasn't necessary to flip it again.

taj-p:
Before I dive into review, what's the level of effort to support build time shader flags so we can avoid the runtime branching? (E.g. using wesl)

I suspect build time flags may prove useful for other endeavours too (e.g. debug info, only supporting bilinear sampling to avoid low/med/high branching etc)

LaurenzV:
I’m not sure because I’ve never looked into it, but it seems to me like this is something orthogonal to this PR, no? I don’t think a single select in the vertex shader and blending/layer branch is gonna kill performance for us. 😄 But if you want to I can certainly look into it.

taj-p:
> I’m not sure because I’ve never looked into it, but it seems to me like this is something orthogonal to this PR, no? I don’t think a single select in the vertex shader and blending/layer branch is gonna kill performance for us. 😄 But if you want to I can certainly look into it.

Definitely not going to kill performance. But build time flags reduce complexity in the driver code (`webgl.rs`) and are likely the better way to handle build-time shader configuration (for maintenance, future evolution, etc).

BUT, I just took a closer look at the code and this isn't a build-time flag. It varies at runtime. In the current code, if you're rendering to a slot texture, we don't invert NDC. If you're rendering to the viewport, do you invert NDC. So, because this varies at runtime, it should remain a runtime flag.

I'll take a closer look at the code soon.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -824,8 +728,8 @@ struct ClearSlotsConfig {
     pub slot_height: u32,
     /// Total height of the texture.
     pub texture_height: u32,
-    /// Padding for alignment.
-    pub _padding: u32,
+    /// See [`Config::ndc_y_negate`].
+    pub ndc_y_negate: u32,
```
taj-p: This field is never read in the shader. I recommend keeping it as `_padding`

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: This means that we only update the config when the render size has changed - that assumes the viewport and the atlas texture are not the same size. If they are the same size, we'll never update `negate_ndc`

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1215,7 +1121,7 @@ impl WebGlPrograms {
                     slot_width: u32::from(WideTile::WIDTH),
                     slot_height: u32::from(Tile::HEIGHT),
                     texture_height: u32::from(Tile::HEIGHT) * total_slots,
-                    _padding: 0,
+                    ndc_y_negate: 0,
```
```suggestion
                    _padding: 0,
```

`ndc_y_negate` isn't read in `clear_slots.wgsl`

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -51,8 +51,22 @@ pub struct Config {
...
-    /// Padding to satisfy WebGL's 16-byte alignment requirement for uniform buffers.
-    pub _padding: u32,
+    /// Whether to flip the y-component of the NDC position.
+    ///
+    /// When transpiling a shader with naga, it applies a y-flip transform when
+    /// compiling to GLSL to account for the difference between the y-down
+    /// coordinate of WebGPU and the y-up coordinate system of WebGL framebuffers.
+    ///
+    /// However, for the native WebGL backend, we want to _avoid_ this transform so that we
+    /// can write directly into the user-provided framebuffer, but still ensure that the scene
```
taj-p, line 61: Might be worth mentioning how, without this optimisation, we need to allocate an extra frame buffer, render into that, and then pay a blit cost from that texture to the screen.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -113,42 +113,13 @@ impl WebGlRenderer {
...
-        //      flipped.
-        //   2. A view framebuffer is used as an intermediate render target. The final result is
-        //      blit onto the  default framebuffer reflected to fix the flipped Y axis.
-        // Anti-aliasing causes the blit operation to fail.
-        let context_options = js_sys::Object::new();
-        js_sys::Reflect::set(&context_options, &"antialias".into(), &JsValue::FALSE).unwrap();
-
         let gl = canvas
-            .get_context_with_context_options("webgl2", &context_options)
+            .get_context("webgl2")
```
taj-p: I think we want to keep the AA false here. 

- The sparse strips pipeline already does per pixel AA
- A 4x MSAA default framebuffer uses 4x more memory
- Unsure how double AA will interact
- Are we getting any value from enabling AA here?

I think to keep this PR contained, let's not consider this for now and keep the default of AA disabled. (Perhaps update the comment that AA is disabled for performance, memory, and because we handle our own AA)

taj-p (commented):
Nice looks good! So good to save this huge copy and allocation. Some comments 🙏

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Oh damn, very sorry for missing that. 🙏 Will fix this.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: My bad, this is a remnant from a previous version where I did negate it in there. 😓

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Understood, but in this case, it still can happen that we get an anti-aliased one in case the user previously requested one, right? But now, we don't really care about it anymore and therefore don't need the assertions anymore, right?

sparse_strips/vello_hybrid/src/render/common.rs
LaurenzV: Done!

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: I've applied a temp fix now, but I think we should generalize this so it doesn't happen again in the future, added a TODO for this. In general, I think we should probably add some better abstractions to the WebGL backend to more generally prevent such out-of-sync situations... But if I understood correctly, allocating some time to improve robustness is on our near-term TODO list anyway? 😄

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -132,23 +125,6 @@ impl WebGlRenderer {
...
-        #[cfg(debug_assertions)]
-        {
-            // If a WebGL context already exists on this canvas, it will be returned instead of
-            // creating a new one with the correct context_options set. A cached context with
-            // antialias enabled will cause vello_hybrid to fail silently. This assertion catches
-            // that error early.
-            let context_attributes = gl.get_context_attributes().unwrap();
-            let antialias = js_sys::Reflect::get(&context_attributes, &"antialias".into())
-                .unwrap()
-                .as_bool()
```
ajakubowicz-canva, line 144: This debug assertion may still be valuable for providing a guard such that its easier to migrate from the webgl backend to a webgpu backend with its webgl runtime feature. 

I think `wgpu` will crash if they receive a canvas with antialias set to true.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -187,64 +163,6 @@ impl WebGlRenderer {
         );
 
         self.render_scene(scene, render_size, true)?;
-
-        // Blit the view framebuffer to the default framebuffer (canvas element), reflecting the
```
ajakubowicz-canva, line 191: It's very nice to see this code get removed :D Awesome!

sparse_strips/vello_sparse_tests/tests/renderer.rs
ajakubowicz-canva: Great find! Great great great improvements :D And great documentation and comments!

sparse_strips/vello_hybrid/src/render/webgl.rs
ajakubowicz-canva: > Understood, but in this case, it still can happen that we get an anti-aliased one in case the user previously requested one, right?

Yeah, it definitely is possible that we can get a previously configured context. The only reason I can think of for preventing anti-aliasing is to maintain compatibility with wgpu + their webgl backend which I think requires antialias to be false. ([my comment](https://github.com/linebender/vello/pull/1546#discussion_r3008692535))

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Done!

PR #1549, opened 2026-03-29
[INVESTIGATION]: Front to back rendering of opaque strips

taj-p:
I'm seeing very promising performance results on low tier devices from front to back rendering of opaque strips (owing to early z culling):

Benchmarks were amended to be opaque. The current Vello bench suites largely have alpha.


LaurenzV:
I tihnk this can be closed now right? 😄

PR #1552, opened 2026-03-31
vello_hybrid: Implement sampling external textures (with atlas support)

tomcur:
On top of https://github.com/linebender/vello/pull/1559.

This adds `Scene::draw_texture_rects` that takes an opaque `TextureId`, one or more rectangles that are to be sampled from that texture, and transforms that determine where the result is to be placed in the destination. The `TextureId`s are mapped by the user at render-time, and are not interned by the renderer.

Some rough benchmark info here: https://github.com/linebender/vello/pull/1552#pullrequestreview-4040920322, with the newest numbers in https://github.com/linebender/vello/pull/1552#pullrequestreview-4108776026 for the newest numbers. In that specific benchmark, rendering glyphs with this texture sampling versus current glifo has about ~1.32x rendering throughput (not measuring building `Scene`), and 1.77x end-to-end throughput versus `glyph_run` with Glifo atlasing enabled (i.e., measuring building `Scene` with hot caches along with measuring rendering the Scene).

This is useful for sampling textures without requiring copying into (and taking space in) the internal atlas. This API can also be used by users to draw from an atlas or spritesheet they manage themselves.

The texture is given to the shader by binding it. If the texture needs to change, the previous operations are emitted with a draw call and the next texture is bound. For simplicity, this currently binds only one texture at a time, though we could bind a handful more. I did experiment with that, but that requires some more invasive changes, e.g., it requires re-encoding the paints to track slots. Theoretically we could re-order operations to try to prevent unnecessary re-binding when, e.g., drawing with Texture A, then B, then A could be re-ordered without visual changes to A, A, B, but that goes into deeper scheduler changes.

Bindless support would be useful here.

I've currently only implemented this for wgpu, but the same sort of thing should work for WebGL.

The `Scene` API in this PR already allows quite a bit in practice, but it's not a generalized brush. In the form this now takes (relative to the form of last week), also implementing something like `ImageSource::ExternalTexture` is probably not too difficult. This atlas API (i.e., with low per-instance overhead) is probably simpler and more important to land.

Some of the GPU plumbing (bindings, etc.) is AI-generated code that I still need to give another look later.

Run the example using `cargo run -p vello_hybrid --example texture_rects`, which generates the following.

texture_rects_output


tomcur (commented):
I've pushed a rough benchmark of rendering by sampling a bunch of glyphs from an external texture versus using `glifo` or `glyph_run`s here: https://github.com/tomcur/vello/tree/glyph-benchmark. I pre-rendered [a spritesheet](https://github.com/tomcur/vello/blob/glyph-benchmark/sparse_strips/vello_hybrid/examples/assets/glyphs_roboto_lowercase.png) of 26 lowercase characters using the Roboto font in this repo. The benchmark renders all `17,576` combinations of three characters in a grid (`52,728` glyphs).  Run the benchmark on that branch using `cargo run -rp vello_hybrid --example bench_glyph_pairs`.

(Edit: the latest numbers are here: https://github.com/linebender/vello/pull/1552#pullrequestreview-4108776026.)

On my machine, the numbers look as follows.

- Rendering has about ~1.4x throughput relative to rendering the built `Scene` versus either `glifo` or `glyph_run`; where this path only does a bunch of blits.
- End-to-end this is of course dramatically faster than `glyph_run`, which does almost no caching, so it's about 15x faster than that, and interestingly, it's also about 15x faster than `glifo` with atlasing turned off (so `glifo` and `glyph_run` are very much in the same ballpark when building the scene from scratch without caching).
- If I turn on atlasing with `glifo`, end-to-end this texture rect pipeline is ~1.6x faster, but I'm not sure what type of work `glifo` with atlasing turned on is currently doing (it's rendering blank).

The benchmarks have a few iterations of warm-up (so caches are hot) and then iterates 100 times by default (re-using `Scene` with `Scene::reset`).

```
Benchmarking 17576 triple cells via 52728 texture rects (52728 outline glyphs, viewport 7488x6734)
Glyph atlas cell 30x37, scene cell 72x37, atlas 780x37

texture_rects   frame_total: avg    4.773 ms, best    4.309 ms
glyph_run       frame_total: avg   71.665 ms, best   70.437 ms
glifo           frame_total: avg   72.931 ms, best   71.671 ms
glifo (atlas)   frame_total: avg    7.679 ms, best    7.329 ms
frame_total glyph_run vs texture_rects: 15.01x
frame_total glifo     vs texture_rects: 15.28x
frame_total glifo (atlas) vs texture_rects: 1.61x

texture_rects   render_only: avg    2.735 ms, best    2.651 ms
glyph_run       render_only: avg    4.060 ms, best    3.879 ms
glifo           render_only: avg    4.006 ms, best    3.866 ms
render_only glyph_run vs texture_rects: 1.48x
render_only glifo     vs texture_rects: 1.46x
```

The result looks like this.

All combinations of three lowercase Latin characters laid out in a grid (aaa, aab, aac, ..., zzy, zzz)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -597,11 +734,17 @@ impl Scene {
...
+                    texture_rects.push(r);
+                }
             }
         }
 
         strip_storage.set_generation_mode(GenerationMode::Replace);
         self.strip_path_mode = StripPathMode::CoarseOnly;
+        drop(strip_storage);
+
+        self.push_texture_rects_to_wide_tiles(&texture_rects);
```
taj-p: Does this honour z-ordering? If I render FillPath -> TexRect -> FillPath and then fallback from fast path, is this going to re-oroder to FillPath -> FillPath -> TexRect

tomcur: It may not have. The version in this PR now goes through the ubershader, making this much simpler.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -491,6 +524,135 @@ impl Scene {
...
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "f64→f32 truncation is acceptable for pixel coordinates"
+    )]
+    pub fn draw_texture_rects(
+        &mut self,
+        texture_id: TextureId,
+        quality: ImageQuality,
+        rects: impl IntoIterator<Item = SampleRect>,
+    ) {
```
tomcur, line 548: I've renamed this to `draw_texture_rects` from `fill_texture_rects` in the previous versions, as an `R8` coverage mask may want an instanced method like `fill_coverage_mask_rects`, which multiplies the normal paint by the `SampleRect` coverage masks.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -240,6 +243,9 @@ var<uniform> config: Config;
 @group(1) @binding(0)
 var atlas_texture_array: texture_2d_array<f32>;
 
+@group(1) @binding(1)
+var external_texture: texture_2d<f32>;
```
tomcur, line 247: I believe with wgpu it would be mostly free to bind this as a `texture_2d_array`, and we're then able to reuse the sampling functions instead of copy-pasting them to take a `texture_2d`, but on WebGL that might have a real cost.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -1026,6 +1062,70 @@ fn bicubic_sample(
...
+    let row3 = cx.x * s03 + cx.y * s13 + cx.z * s23 + cx.w * s33;
+    let result = cy.x * row0 + cy.y * row1 + cy.z * row2 + cy.w * row3;
+
+    // Clamp each component to [0,1] and ensure color components don't exceed alpha
+    return vec4<f32>(
+        min(clamp(result.r, 0.0, 1.0), result.a),
+        min(clamp(result.g, 0.0, 1.0), result.a),
+        min(clamp(result.b, 0.0, 1.0), result.a),
+        min(clamp(result.a, 0.0, 1.0), result.a)
+    );
```
waywardmonkeys: This needs the fix to clamp the alpha first and then clamp the others to that. This needs doing on the non-external bicubic code path as well. (I just made this change in Vello Classic's copy of the bicubic code as well and will submit a PR for the non-external version here.)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -491,6 +524,135 @@ impl Scene {
...
+    /// The per-rect transforms are composed with the current
+    /// [scene transform][`Self::set_transform`]. This transform is relative to the local region
+    /// defined by each [`SampleRect`]: i.e., the origin of each [`SampleRect`] is used only to
+    /// determine the region to sample in the source [`TextureId`], and is ignored for determining
+    /// the destination.
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "f64→f32 truncation is acceptable for pixel coordinates"
+    )]
+    pub fn draw_texture_rects(
```
waywardmonkeys, line 543: The doc comment here should lay out some of the other requirements:

* Needs to be pre-multiplied RGBA
* The source_region needs to stay within the texture bounds.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -431,16 +451,41 @@ impl TileEl {
...
-        self.0.push(gpu_strip);
+    fn push(&mut self, gpu_strip: GpuStrip, external_texture_id: Option<TextureId>) {
+        if let Some(texture_id) = external_texture_id {
+            let needs_new_run = self
+                .external_texture_runs
+                .last()
+                .is_none_or(|run| run.texture_id != texture_id);
+            if needs_new_run {
+                let strips_start = if self.external_texture_runs.is_empty() {
+                    0
```
waywardmonkeys, line 477: Codex thinks this is correct, but needs an explanatory comment:

- when a pass has any external run, do_strip_render_pass switches bind group 1 per run
- that bind group still always has the normal atlas at binding 0
- only binding 1 changes, for the external texture
- so atlas-only strips that get covered by a run range still render correctly,
  because they ignore binding 1

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -493,6 +493,10 @@ impl WebGlRenderer {
                     self.encoded_paints[encoded_paint_idx] = gpu_gradient;
                     current_idx += gradient_size_texels;
                 }
+                EncodedPaint::ExternalTexture(_external_texture) => {
+                    // TODO: External textures are not yet supported.
+                    log::warn!("External texture are not yet supported in the WebGL backend");
```
waywardmonkeys, line 577: Codex thinks this can cause a mis-rendering:

There is no encoding and no current_idx += ....

Why that matters:

- every paint records its start index before the match
- if an external texture paint appears, it claims the current index
- but the slot is never populated and never consumed
- the next paint reuses the same index

So if a scene has ExternalTexture followed by a gradient or image, both paints can
point at the same GPU paint texels.

waywardmonkeys (commented):
Should also ad a test case for WGPU: test that `render_with_texture_bindings` returns `RenderError::MissingTextureBinding` when a referenced `TextureId` is absent.

waywardmonkeys:
There's also a chance that `render_to_atlas` will not work due to overwriting the texture bindings (`let bindings = TextureBindings::new();`)

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
tomcur: Nice, brought that change over!

sparse_strips/vello_hybrid/src/scene.rs
tomcur: I've documented these scene requirements on `draw_texture_rects`, and documented wgpu-specific requirements on `TextureBindings::insert`. WebGL should similarly be documented (when we get to it).

sparse_strips/vello_hybrid/src/schedule.rs
tomcur: Good point, I've documented this, though focused somewhat more on the perspective of the scheduler.

sparse_strips/vello_hybrid/src/render/webgl.rs
tomcur: Good point!

tomcur (commented):
The newest benchmark with `vello_hybrid` now using `glifo` is as follows:

```
Benchmarking 17576 triple cells via 52728 texture rects (52728 outline glyphs, viewport 7488x6734)
Glyph atlas cell 30x37, scene cell 72x37, atlas 780x37

texture_rects   frame_total: avg    4.628 ms, best    4.490 ms
glyph_run       frame_total: avg   73.212 ms, best   71.953 ms
glyph_run (atl) frame_total: avg    8.191 ms, best    7.728 ms
frame_total glyph_run       vs texture_rects: 15.82x
frame_total glyph_run (atl) vs texture_rects: 1.77x

texture_rects   render_only: avg    3.040 ms, best    2.863 ms
glyph_run       render_only: avg    4.004 ms, best    3.863 ms
render_only glyph_run vs texture_rects: 1.32x
```

sparse_strips/vello_common/src/encode.rs
```diff
@@ -575,6 +579,26 @@ pub struct EncodedImage {
...
+    /// External texture handle.
+    pub texture_id: TextureId,
+    /// Source region of the texture in texel coordinates.
+    pub source_region: RectU16,
+    /// Sampler parameters.
+    pub sampler: ImageSampler,
+    /// Whether the sampled content may contain non-opaque pixels.
+    pub may_have_opacities: bool,
+    /// Inverse destination transform, mapping scene coordinates to local source-rect space.
+    pub transform: Affine,
```
LaurenzV, line 597: Is it really more intuitive to make the user pass the inverse transform here? Can they not just pass the transform and we invert it ourselves?

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -626,17 +626,23 @@ impl<const MODE: u8> Wide<MODE> {
...
-                        }
+                            Some(EncodedPaint::ExternalTexture(img))
+                                if !img.may_have_opacities
+                                    && img.sampler.alpha == 1.0
+                                    && img.tint.is_none_or(|t| t.color.components[3] >= 1.0) =>
+                            {
+                                FillHint::OpaqueImage
+                            }
+                            _ => FillHint::None,
+                        },
```
LaurenzV, line 645:
```suggestion
                        Paint::Indexed(idx) => match encoded_paints.get(idx.index()) {
                            Some(EncodedPaint::Image(img) | EncodedPaint::ExternalTexture(img))
                                if !img.may_have_opacities
                                    && img.sampler.alpha == 1.0
                                    && img.tint.is_none_or(|t| t.color.components[3] >= 1.0) =>
                            {
                                FillHint::OpaqueImage
                            }
                            _ => FillHint::None,
                        },
```

sparse_strips/vello_common/src/encode.rs
LaurenzV: Nvm, I didn't realize this is the vello-internal representation!

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -490,6 +523,140 @@ impl Scene {
...
+    /// [scene transform][`Self::set_transform`]. This transform is relative to the local region
+    /// defined by each [`SampleRect`]: i.e., the origin of each [`SampleRect`] is used only to
+    /// determine the region to sample in the source [`TextureId`], and is ignored for determining
+    /// the destination.
+    ///
+    /// A texture with the given [`TextureId`] must be supplied at render time. The given
+    /// [source regions][`SampleRect::source_region`] must be within bounds of that texture. The
+    /// texture is treated as premultiplied alpha in the render target's color space. See the
+    /// backend's binding type for more information on texture requirements.
+    #[expect(
```
LaurenzV, line 565: Maybe a comment that this method isn't affected by the current `paint_transform`?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -490,6 +523,140 @@ impl Scene {
...
+        // Because in this API the destination drawn is always the transformed input rect, this
+        // means extend modes don't currently materially impact rendering. In general drawing with
+        // an external texture brush, extend modes would matter, so we still encode them.
+        let x_extend = Extend::Pad;
+        let y_extend = Extend::Pad;
+
+        let can_fast_path = self.strip_path_mode != StripPathMode::CoarseOnly
+            && !self.wide.has_layers()
+            && self.filter.is_none()
+            && self.clip_context.get().is_none();
```
LaurenzV: Can we deduplicate this with the check in `try_fast_rect`?

sparse_strips/vello_sparse_tests/tests/renderer.rs
```diff
@@ -72,6 +75,21 @@ pub(crate) trait Renderer: Sized {
...
     fn height(&self) -> u16;
+    #[allow(
+        dead_code,
+        reason = "external texture snapshot scaffolding is not enabled yet"
+    )]
+    fn register_external_texture(&mut self, pixmap: Arc<Pixmap>) -> TextureId;
+    #[allow(
+        dead_code,
+        reason = "external texture snapshot scaffolding is not enabled yet"
+    )]
```
LaurenzV: Don't we already have tests that use it? Why is it still dead code?

sparse_strips/vello_sparse_tests/tests/renderer.rs
```diff
@@ -272,6 +303,12 @@ pub(crate) struct HybridRenderer {
     texture: wgpu::Texture,
     texture_view: wgpu::TextureView,
     renderer: vello_hybrid::Renderer,
+    external_textures: HashMap<TextureId, wgpu::TextureView>,
+    #[allow(
+        dead_code,
+        reason = "external texture snapshot scaffolding is not enabled yet"
+    )]
```
LaurenzV, line 300: Same here

sparse_strips/vello_sparse_tests/tests/external_texture.rs
```diff
@@ -0,0 +1,202 @@
+// Copyright 2025 the Vello Authors
```
LaurenzV:
```suggestion
// Copyright 2026 the Vello Authors
```

sparse_strips/vello_sparse_tests/tests/external_texture.rs
```diff
@@ -0,0 +1,202 @@
...
+            HEIGHT,
+            false,
+            0,
+            "fallback",
+            RenderMode::OptimizeSpeed,
+            false,
+        );
+        f(&mut ctx);
+        let specific_name = format!("{test_name}_hybrid");
+        // Set `diff_pixels` to `3`: we can't generate a "golden standard" using `f32` on `vello_cpu` as
```
LaurenzV, line 50: Isn't 1 or 2 still enough? Even if the gold standard is generated by hrbid instead of CPU.

sparse_strips/vello_sparse_tests/tests/external_texture.rs
```diff
@@ -0,0 +1,202 @@
...
+        test_name: &str,
+        f: impl FnOnce(&mut HybridRenderer),
+    ) {
+        let mut ctx = get_ctx::<HybridRenderer>(
+            WIDTH,
+            HEIGHT,
+            false,
+            0,
+            "fallback",
+            RenderMode::OptimizeSpeed,
```
LaurenzV:
```suggestion
            RenderMode::OptimizeQuality,
```

Even if it doesn't really change anything for Hybrid, for future-proofing I think it's still better to use this for reference image generation.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -906,7 +955,7 @@ impl Scheduler {
                 continue;
             }
 
-            renderer.render_strips(&draw.0, target, load);
+            renderer.render_strips(&draw.strips, &draw.external_texture_runs, target, load);
```
LaurenzV: Might as well just make the method take &Draw here, no?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -432,16 +452,45 @@ impl TileEl {
...
+    fn push(&mut self, gpu_strip: GpuStrip, external_texture_id: Option<TextureId>) {
+        if let Some(texture_id) = external_texture_id {
+            // Runs are consecutive `GpuStrip`s that can be rendered with a single external texture.
+            // If the external texture changes, a new run is created. The first external texture
+            // encountered creates the first run, which is "retroactively" set to start from the
+            // first `GpuStrip`, such that it can be applied within the same draw call.
+            let needs_new_run = self
+                .external_texture_runs
+                .last()
+                .is_none_or(|run| run.texture_id != texture_id);
```
LaurenzV, line 524: Do we really need a new run if the last one is `None`? Let's say we have

ID 2, None, None, ID 2

shouldn't it be possible to all schedule them in one run, even if the strips in the middle don't actually end up using the texture?

LaurenzV: I imagine this would be pretty common, so probably worth optimizing! (Unless I am missing something)

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -236,6 +300,37 @@ impl Renderer {
...
+            render_size,
+            view,
+            &bindings,
+        )
+    }
+
+    /// Render `scene` with [externally bound textures](`TextureBindings`).
+    ///
+    /// See [`Self::render`] to render without externally bound textures.
+    ///
```
LaurenzV: Do we really need two methods for this? To me, it seems fine to just force the user to do `TextureBindings::new`. It's a bit annoying to change the signature everywhere, but IMO better than having a separate method just for a single argument. 😄 But maybe other people have other thoughts.

LaurenzV: Or it could also be an option, and do unwrap_or(TextureBindings::new)

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1718,18 +1879,44 @@ impl Programs {
...
+            size: Extent3d {
+                width: 1,
+                height: 1,
+                depth_or_array_layers: 1,
+            },
+            mip_level_count: 1,
+            sample_count: 1,
+            dimension: wgpu::TextureDimension::D2,
+            format: wgpu::TextureFormat::Rgba8Unorm,
+            usage: wgpu::TextureUsages::TEXTURE_BINDING | wgpu::TextureUsages::COPY_DST,
```
LaurenzV: Is `COPY_DST` needed here?

LaurenzV (commented):
From my side no big concerns about landing this, but Alex still wants to take a look!

sparse_strips/vello_sparse_tests/tests/external_texture.rs
tomcur: Unfortunately with 1 or 2 there are test failures: it seems the computations on the GPU sometimes end up falling 2 away from the ground truth, but we don't actually know that ground truth, so we need to compare the "fuzzy" renders between different hardware. Those end up disagreeing by up to 3 here (and I suppose it could sometimes even be 4, but that doesn't appear to be happening).

sparse_strips/vello_sparse_tests/tests/renderer.rs
tomcur: Woops, good catch. Using `#[expect(...)]` would have prevented this.

sparse_strips/vello_hybrid/src/render/wgpu.rs
tomcur: Nope, and good catch! I want/need to give the GPU plumbing a bit more of a closer look, it was mostly generated. (I was waiting for this general direction to be mostly approved!)

sparse_strips/vello_common/src/coarse.rs
tomcur: This unfortunately doesn't work as written, because `EncodedPaint::Image` carries an `EncodedImage`, and `EncodedPaint::ExternalTexture` carries an `EncodedExternalTexture`. We could write something like the following, but it reads a bit messily with all the `Options.

```rust
Paint::Indexed(idx) => {
    let opacity_info = match encoded_paints.get(idx.index()) {
        Some(EncodedPaint::Image(img)) => {
            Some((img.may_have_opacities, img.sampler.alpha, img.tint))
        }
        Some(EncodedPaint::ExternalTexture(img)) => {
            Some((img.may_have_opacities, img.sampler.alpha, img.tint))
        }
        _ => None,
    };
    match opacity_info {
        Some((false, 1.0, tint))
            if tint.is_none_or(|t| t.color.components[3] >= 1.0) =>
        {
            FillHint::OpaqueImage
        }
        _ => FillHint::None,
    }
}
```

We could get rid of some with a labeled block and a `break`, but that's sometimes a bit surprising without a loop.

```rust
Paint::Indexed(idx) => 'fill: {
    let (may_have_opacities, sampler_alpha, tint) =
        match encoded_paints.get(idx.index()) {
            Some(EncodedPaint::Image(img)) => {
                (img.may_have_opacities, img.sampler.alpha, img.tint)
            }
            Some(EncodedPaint::ExternalTexture(img)) => {
                (img.may_have_opacities, img.sampler.alpha, img.tint)
            }
            _ => break 'fill FillHint::None,
        };
    if !may_have_opacities
        && sampler_alpha == 1.0
        && tint.is_none_or(|t| t.color.components[3] >= 1.0)
    {
        FillHint::OpaqueImage
    } else {
        FillHint::None
    }
}
```

sparse_strips/vello_hybrid/src/schedule.rs
tomcur: If I understand you correctly, that's what this is doing. For example, if we render 

`None, ID 2, None, None, ID 2, None, ID 3, None, ID 3, None`, then there are two runs: the first run renders the first six strips with texture with ID 2 bound, the second run renders the remaining strips with texture with ID 3 bound.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Oh never mind, I thought I tried it but makes sense! Feel free to disregard then.

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: But ‘needs_new_run’ returns true if the last is None, no? So every None triggers a new run? Or am I missing something?

LaurenzV: Ahh never mind, the none refers to whether there even exists a previous run! My bad 😄

tomcur: Indeed! The runs stay `None` when there are no external textures at all (in which case we bind the placeholder texture).

sparse_strips/vello_hybrid/src/scene.rs
tomcur: Yes, that's a good point. We may at some point want a similar method for using coverage textures, which could allow a regular paint (with a paint transform); e.g., drawing text with a gradient using a coverage atlas. There, a distinction like this becomes important.

As an aside, with the form the shader in this PR currently has, I think we could also wire up these external textures as a normal image source, but that's a bit tangential, and I first wanted to focus on a path with low per-instance overhead for atlas/sprite sheet use.

sparse_strips/vello_sparse_tests/tests/external_texture.rs
```diff
@@ -0,0 +1,202 @@
+// Copyright 2026 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#[cfg(not(target_arch = "wasm32"))]
+mod tests {
```
grebmeg, line 5: My next comment isn’t about the tests, but about the `texture_rects` example you added in the `tomcur:external-textures` branch. It would be great if you could add another example scene to `vello_example_scenes` with similar stuff to showcase its usage and test it in a more interactive environment.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Could you say a bit more about the WebGL cost you're imagining?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -395,9 +402,17 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
-            if encoded_image.quality == IMAGE_QUALITY_HIGH {
+            if encoded_image.source_kind == IMAGE_SOURCE_EXTERNAL {
+                let final_xy = image_offset + extended_xy;
+                sample_color = sample_external_image(
+                    encoded_image.quality,
+                    final_xy,
+                    image_offset,
+                    image_size,
+                );
+            } else if encoded_image.quality == IMAGE_QUALITY_HIGH {
```
grebmeg, line 424: This next point leans a bit toward more specialized shaders, though I think this PR is already a step in that direction. I’m curious whether you considered compiling two strip pipelines instead of branching on `source_kind` per fragment.

Each draw already has a consistent `source_kind` (since the scheduler splits draws on `TextureId` boundaries), so the per-fragment branch and the `source_kind` field on `EncodedImage` seem structurally redundant. It looks like using `set_pipeline` per run would achieve the same effect.

Some potential benefits: removing the per-fragment branch, and being able to drop `EncodedImage::source_kind` and `EXTERNAL_IMAGE_SOURCE_FLAG`, which would simplify paint encoding.

I’m not suggesting this needs to be done now, just interested in your perspective.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -1022,6 +1058,66 @@ fn bicubic_sample(
...
+    let row0 = cx.x * s00 + cx.y * s10 + cx.z * s20 + cx.w * s30;
+    let row1 = cx.x * s01 + cx.y * s11 + cx.z * s21 + cx.w * s31;
+    let row2 = cx.x * s02 + cx.y * s12 + cx.z * s22 + cx.w * s32;
+    let row3 = cx.x * s03 + cx.y * s13 + cx.z * s23 + cx.w * s33;
+    let result = cy.x * row0 + cy.y * row1 + cy.z * row2 + cy.w * row3;
+
+    // Clamp alpha first, then clamp premultiplied color channels against it.
+    let a = clamp(result.a, 0.0, 1.0);
+    return vec4<f32>(clamp(result.rgb, vec3(0.0), vec3(a)), a);
+}
```
grebmeg, line 1117: Could the math for the bilinear and bicubic samplers be extracted into shared functions to reduce duplicated code?

grebmeg (commented):
Hey @tomcur, sorry for the delay in reviewing your PR 🙏 I just got back from vacation yesterday, so it took me a bit of time to dig into what’s going on, and I still need a bit more time.

Re: this PR — great work! I like that it’s not as invasive as I initially thought it might be. There are clear benefits to allowing consumers to use external textures, and this also feels like a solid starting point for specialised shaders and future batching work.

My main concern is performance on low-tier devices. I’ve done some benchmarking, and the results were quite surprising.

I used the [vello_bench2](https://github.com/laurenz-canva/vello_bench2) repo that Laurenz originally created for web benchmarking across devices. I tested on a Samsung Galaxy A05s, connected to my MacBook Pro M1 Max via usb, using `adb` to reverse proxy the vello_bench2 app running on http://localhost:8080. I ran benchmarks and also spent some time in interactive mode with rectangle scenes using both solid and image fills.

For comparison, I used commit 08d19de2 as the “before” state and f8d67c86 (from the `tomcur/glyph-benchmark` branch) as the “after”. Initially, I ran with 3 warmups and 15 iterations and saw somewhat similar trends, but I suspected thermal throttling or other noise, so I increased the workload. I reran both “before” and “after” with 10 warmups and 100 iterations, then used the compare report feature in vello_bench2.

Here’s what I observed:

||||||
---|---|---|---|---
Image|Image|Image|Image|Image|

While testing in interactive mode, I saw a similar trend:

Before|After
---|---
Image|Image
Image|Image

Performance for solid rectangle rendering improved by roughly 50% !!??, while image rendering regressed by about 40%. That wasn’t what I expected, and I don’t have a clear explanation.

The regression in image rendering is a significant concern for low-tier devices. It would be great to investigate this further. As a potential next step, maybe we can consider having two shaders, one with and one without external texture support, and/or feature-gating this functionality?

Also, would it be possible to include a webgl implementation as well?

LaurenzV:
Regarding the benchmarks, I wouldn’t trust those fully because I’ve observed they can sometimes vary a lot, not sure why. 😅 But if you saw similar numbers in interactive mode as well then that’s indeed interesting. 🤔

sparse_strips/vello_hybrid/src/scene.rs
tomcur: What do you think of the following? https://github.com/linebender/vello/pull/1552/commits/d06dcc9434d100ae6a415a0830df330aa8be4e5c

grebmeg:
> Regarding the benchmarks, I wouldn’t trust those fully because I’ve observed they can sometimes vary a lot, not sure why. 😅 But if you saw similar numbers in interactive mode as well then that’s indeed interesting. 🤔

This is what I eventually ended up with after benchmarking the blurred rounded rects PR #1610: the results are pretty unreliable. I used 10 warmups and 100 iterations, so it took quite a while to run, but I kept seeing unexpected data. In the end, I validated performance only in interactive mode, which also showed some strange results.

I think we should investigate this further. My guess is that vello_bench2 is very sensitive to noise, or maybe we're not running something correctly, not sure yet. I can probably spend a bit more time on this tomorrow and rerun the benchmarks one more time.

grebmeg (approved):
Let’s move forward with merging this! I think it’d be better to spend some time introducing [WESL](https://wesl-lang.dev/)￼ preprocessing later, rather than spending more time benchmarking this further, since preprocessing is one of the next steps for shader specialization.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
tomcur: I believe to use a `texture_2d_array` containing a single texture on WebGL, the user actually has to create a singleton array. On wgpu, a 2D texture can be used as a singleton `texture_2d_array` by constructing the correct `TextureView`.

sparse_strips/vello_hybrid/src/schedule.rs
tomcur: Yeah, I think that would've made sense before. There's a bit more logic now because of the opaque strip work that landed on main, so I think we can leave this as is.

sparse_strips/vello_hybrid/src/render/wgpu.rs
tomcur: I've merged it into a single method taking `&TextureBindings`. Constructing the `TextureBindings` is very cheap in the grand scheme of things (and it's reusable).

tomcur:
> Let’s move forward with merging this! I think it’d be better to spend some time introducing [WESL](https://wesl-lang.dev/)￼ preprocessing later, rather than spending more time benchmarking this further, since preprocessing is one of the next steps for shader specialization.

Awesome! Thank you for your careful review and benchmarking.

I've made this mergeable again (and preserved history of this PR, just in case it ever makes sense to look at the specialized shader in the old commits here). I've addressed Laurenz' points. I'll look into adding the example you requested tomorrow and go over your other points.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
tomcur: I've taken a look at sharing, and while some sharing is possible, in the end all the `textureLoads` need to be on the correct type. For bicubic sampling that means either still having 16 `textureLoads` for both versions with the 16 positions extracted somewhere else (increasing the lines of code), or both versions having a loop over the 16 positions and relying on a good optimizing compiler.

Unfortunately, the version with the loop doesn't seem to get optimized well on my machine. It looked something like the following.

```wgsl
let setup = bicubic_setup(coords, image_offset, image_size); // Generates a fixed-size array with positions to sample.
var samples: array<vec4<f32>, 16>;
for (var i = 0u; i < 16u; i = i + 1u) {
    samples[i] = textureLoad(tex, setup.positions[i], atlas_idx, 0);
}
return bicubic_combine(samples, setup.cx, setup.cy) // Combines the samples
```

In short, it seems like there's not too much we can do here.

sparse_strips/vello_sparse_tests/tests/external_texture.rs
tomcur: Done! Run with `cargo run -rp vello_hybrid_winit -- 20` (the scene index is correct at the time of writing).

tomcur:
> Also, would it be possible to include a webgl implementation as well?

I've landed this without webgl (it needs separate approval); theoretically it would be similar to the wgpu one.

Issue #1545, opened 2026-03-27
Performance issue with rendering large images using Vello Classic


johan-overbye:
Hi :) I'm having a performance issue with Vello Classic. I'm filling a full-screen rectangle with a 1835 by 8192 image on an iPad Pro 2025 13", and I'm unable to sustain 120 FPS. I'm not doing anything else with the GPU.

If I scale the image down (proportionally) to a height of 4096, the frame rate recovers to 120 (the display's refresh rate). However if I scale it to 4097, the problem persists, suggesting some kind of threshold at 4096 pixels. What could be going wrong here?

Two other less significant issues:
- `ImageQuality::High` does not yield bicubic filtering.
- Image sizes of more than 8192 silently fail to render. I'd expect an error somewhere.

Tested with revision Vello commit `ea2a935d4c502c` and `--release`.

johan-overbye:
I've been playing around with this some more, and I'm seeing now that the performance of image rendering with Vello Classic seems to just be generally poor on iPadOS.

I tried tessellating my large image to various tile sizes, down to 256 by 256 at the smallest. At every tile size I'm struggling to sustain 120 FPS. With 256x256 tiles, I'm able to render around 65 of them before my app starts to drop frames.

Tested with a newer Vello revision as well, `d4c502c`.

nicoburns:
> the performance of image rendering with Vello Classic seems to just be generally poor on iPadOS

What hardware are you running that on?. I imagine that makes more difference than the OS.

waywardmonkeys:
> > the performance of image rendering with Vello Classic seems to just be generally poor on iPadOS
> 
> What hardware are you running that on?. I imagine that makes more difference than the OS.

He said:

> I'm filling a full-screen rectangle with a 1835 by 8192 image on an iPad Pro 2025 13", and I'm unable to sustain 120 FPS. I'm not doing anything else with the GPU.

That's good hardware.

The issues here are Vello.

In Vello Classic, the textures are copied into an atlas and it starts at 1024x1024 and doubles up to 8192x8192.

During resolve, I think it clears the atlas state, tries to repack, and if something doesn't fit, bumps the size and packs again.

Further, images are re-uploaded into the atlas texture during render.

For the silent drop of textures that are 8192 or larger, if an image can't be packed after the atlas reaches max size, Vello sets the image's atlas location to None (src/resolve.rs:497) and when patching the draw data, it writes zeros so the image won't render (src/resolve.rs:295) and that's a silent failure.

[...]

PR #1558, opened 2026-04-03
vello: Keep image atlas residency across renders

waywardmonkeys:
This changes Vello’s image atlas handling from per-render rebuilds to persistent residency.

Before this series, image atlas placement was recomputed on every resolve, and the renderer recreated and rewrote the atlas texture every render. That meant even steady-state scenes that only moved already-uploaded images still paid repeated atlas upload cost. This series keeps atlas residency metadata alive in `vello_encoding`, keeps the atlas texture proxy alive in `vello`, and only uploads images that are newly inserted or dirtied by atlas growth.

What changed:

* `vello_encoding` now keeps the single-atlas allocator and resident image map across resolves.
* Resident entries are reused by blob id instead of being repacked every render.
* Eviction uses generation-based staleness under allocation pressure before growing the atlas.
* Atlas growth still repacks, but marks residents dirty so only required re-uploads happen afterward.
* The renderer now keeps the atlas texture resident across renders and only recreates it when the atlas size changes.
* Atlas logging is quieter in the steady state, but still reports creation, resize, uploads, and eviction activity.
* Added demo scenes that make residency visible, including a large-image case that roughly fills a 4096x4096 atlas with moving images.

Use this to see the trace log messages:

```
RUST_LOG=vello::render=debug cargo run -p with_winit -- --test-scenes
```

### DIFF
--- a/vello/src/render.rs
+++ b/vello/src/render.rs
@@ -12,6 +12,13 @@ use crate::Scene;
 
 use vello_encoding::{Encoding, Resolver, WorkgroupSize, make_mask_lut, make_mask_lut_16};
 
+#[derive(Clone, Copy, Debug)]
+enum AtlasProxyAction {
+    Created,
+    Reused,
+    Resized,
+}
+
 /// State for a render in progress.
 pub struct Render {
     fine_wg_count: Option<WorkgroupSize>,
@@ -78,9 +85,10 @@ pub(crate) fn render_full(
     scene: &Scene,
     resolver: &mut Resolver,
     shaders: &FullShaders,
+    image_atlas: &mut Option<ImageProxy>,
     params: &RenderParams,
 ) -> (Recording, ResourceProxy) {
-    render_encoding_full(scene.encoding(), resolver, shaders, params)
+    render_encoding_full(scene.encoding(), resolver, shaders, image_atlas, params)
 }
 
 #[cfg(feature = "wgpu")]
@@ -92,10 +100,12 @@ pub(crate) fn render_encoding_full(
     encoding: &Encoding,
     resolver: &mut Resolver,
     shaders: &FullShaders,
+    image_atlas: &mut Option<ImageProxy>,
     params: &RenderParams,
 ) -> (Recording, ResourceProxy) {
     let mut render = Render::new();
-    let mut recording = render.render_encoding_coarse(encoding, resolver, shaders, params, false);
+    let mut recording =
+        render.render_encoding_coarse(encoding, resolver, shaders, image_atlas, params, false);
     let out_image = render.out_image();
     render.record_fine(shaders, &mut recording);
     (recording, out_image.into())
@@ -127,6 +137,7 @@ impl Render {
         encoding: &Encoding,
         resolver: &mut Resolver,
         shaders: &FullShaders,
+        persistent_image_atlas: &mut Option<ImageProxy>,
         params: &RenderParams,
         robust: bool,
     ) -> Recording {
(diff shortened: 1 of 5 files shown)


waywardmonkeys:
As with everything, this was done with AI assistance, but I did review it and we iterated on it some.

waywardmonkeys:
This helps address the performance problem described in #1545.

waywardmonkeys:
We probably need a way to mark images as dirty (mainly `image_override` ones) (or only them?)

vello_encoding/src/image_cache.rs
```diff
@@ -1,26 +1,41 @@
...
+    dirty: bool,
+    last_seen_generation: u64,
+}
+
 pub(crate) struct ImageCache {
     atlas: AtlasAllocator,
-    /// Map from image blob id to atlas location.
-    map: HashMap<u64, (u32, u32)>,
-    /// List of all allocated images with associated atlas location.
+    max_size: i32,
```
raphlinus, line 33: I'd appreciate docstring explaining what max_size and evicted_in_resolve refer to. In the case of max_size, the thing that confused me was simply that it's a square size.

vello_encoding/src/image_cache.rs
```diff
@@ -1,26 +1,41 @@
...
 
 const DEFAULT_ATLAS_SIZE: i32 = 1024;
 const MAX_ATLAS_SIZE: i32 = 8192;
+const EVICT_AFTER_GENERATIONS: u64 = 2;
 
 #[derive(Default)]
 pub struct Images<'a> {
     pub width: u32,
     pub height: u32,
+    pub evicted: usize,
```
raphlinus, line 17: I think this could use a docstring; the other fields are clear enough. Having read the PR, I think this is "the number of images evicted during the resolve pass", but I could be wrong.

Also, unless I'm reading this wrong, it's not actually used in the logic, only for debug logging; might be helpful to mention this.

vello_encoding/src/image_cache.rs
```diff
@@ -32,50 +47,205 @@ impl Default for ImageCache {
...
+    }
+
+    fn repack_to_size(&mut self, size: i32) {
+        let mut atlas = AtlasAllocator::new(size2(size, size));
+        let mut entries: Vec<_> = self.map.drain().collect();
+        entries.sort_by_key(|(id, _)| *id);
+        for (id, mut resident) in entries {
+            let alloc = atlas
+                .allocate(size2(resident.image.width as _, resident.image.height as _))
+                .expect("resident image must fit after atlas growth");
```
raphlinus, line 178: I'm a little worried about panics here. Does guillotiere give guarantees that allocation will succeed if size is sufficient? It seems like fragmentation etc might cause a failure under reasonable circumstances.

vello_encoding/src/image_cache.rs
```diff
@@ -1,26 +1,41 @@
...
 }
 
+#[derive(Clone)]
+struct ResidentImage {
+    image: ImageData,
+    alloc_id: AllocId,
+    x: u32,
+    y: u32,
+    dirty: bool,
+    last_seen_generation: u64,
```
raphlinus, line 28: I'm wondering if "last_used_generation" might be a clearer name, as it evokes "LRU".

vello_encoding/src/resolve.rs
```diff
@@ -487,13 +488,21 @@ impl Resolver {
...
+            }
             // Loop over the images, attempting to allocate them all into the atlas.
             for pending_image in &mut self.pending_images {
                 if let Some(xy) = self.image_cache.get_or_insert(&pending_image.image) {
                     pending_image.xy = Some(xy);
                 } else {
+                    if self.image_cache.can_fit_image(&pending_image.image)
+                        && self.image_cache.evict_stale_entries()
+                    {
+                        continue 'outer;
```
raphlinus, line 504: I think `continue 'outer` will work here, but isn't necessarily optimal; it should be possible to retry the `get_or_insert` and move on. But it makes the logic a bit messier, so not sure it's worth changing.

vello_encoding/src/resolve.rs
```diff
@@ -487,13 +488,21 @@ impl Resolver {
     }
 
     fn resolve_pending_images(&mut self) {
-        self.image_cache.clear();
         'outer: loop {
+            self.image_cache.restart_resolve_pass();
+            for pending_image in &mut self.pending_images {
```
raphlinus, line 493: Pretty sure this isn't needed; on the last iteration of the outer loop, the .xy field will be set for all pending images.

vello_encoding/src/image_cache.rs
```diff
@@ -32,50 +47,205 @@ impl Default for ImageCache {
...
+    pub(crate) fn can_fit_image(&self, image: &ImageData) -> bool {
+        image.width <= self.atlas.size().width as u32
+            && image.height <= self.atlas.size().height as u32
+    }
+
+    pub(crate) fn evict_stale_entries(&mut self) -> bool {
+        let Some(stale_before) = self.generation.checked_sub(EVICT_AFTER_GENERATIONS) else {
+            return false;
+        };
+        let stale_ids: Vec<_> = self
```
raphlinus, line 151: This could be redone without allocating stale_ids. The flow is basically:

```rust
    for (_id, resident) in self.map.extract_if(...) {
        self.atlas.deallocate(resident.alloc_id);
        self.evicted_in_resolve += 1;
    }
```

vello_encoding/src/image_cache.rs
```diff
@@ -32,50 +47,205 @@ impl Default for ImageCache {
...
         }
     }
+
+    pub(crate) fn can_fit_image(&self, image: &ImageData) -> bool {
+        image.width <= self.atlas.size().width as u32
+            && image.height <= self.atlas.size().height as u32
+    }
+
+    pub(crate) fn evict_stale_entries(&mut self) -> bool {
+        let Some(stale_before) = self.generation.checked_sub(EVICT_AFTER_GENERATIONS) else {
```
raphlinus, line 148: Potential optimisation: can return false if `evicted_in_resolve` is nonzero.

raphlinus (approved):
I checked this over fairly carefully. Generally, it looks good, but I did have some comments; my biggest single concern is the potential panic in `repack_to_size`.

Just sketching, there's an alternative to the `dirty` bool in residents. Rather, you can have a resize_gen for the atlas, incremented on each resize. Instead of dirty, track the resize_gen for which the image was last uploaded. Then in the occupied case of `get_or_insert`, push & update resize_gen if resize_gen doesn't match. That avoids the full map iteration in `restart_resolve_pass` and eliminates the need for `finish_resolve`. I personally find this easier to reason about: the resize_gen in a resident identifies which gen it was uploaded to. But basically comes down to a style issue; I did verify the current logic and am satisfied.

waywardmonkeys:
@raphlinus I've updated this per your feedback, I think. Want to take a fresh look?

raphlinus:
Yup, looks like you addressed all my comments. Thanks!

vello/src/render.rs
```diff
@@ -146,11 +157,47 @@ impl Render {
...
+                );
+            } else {
+                log::debug!(
+                    "image atlas {}x{}: {:?}, {} upload(s), {} eviction(s)",
+                    atlas_width,
+                    atlas_height,
+                    atlas_proxy_action,
+                    images.images.len(),
+                    images.evicted
+                );
```
Keavon, line 198: Can we, as consumers of Vello, opt to disable this log output in some way?

PR #1562, opened 2026-04-07
Migrate vello sparse strips renderers to use glifo

LaurenzV:
# Overview
This PR attempts to integrate glifo into vello_hybrid and vello_cpu. As mentioned, the git diff is a bit messy because 1) it's based on an old version of main and 2) some commits that are needed have been cherry-picked, bust should ideally be merged separately (i.e. the other currently open PRs).

# Blockers

PRs I’d consider blocking are:
~#1553~
~#1555~

Semi-blocking is also #1459, but for now we could just accept that (non-cached) bitmap glyphs panic in vello_hybrid and not render them for now.

# Limitations / TODOs
Since this PR by itself is already big by itself, it still some TODOs and definitely doesn’t represent the ideal end state we want to have:
- glifo still has the dependency on vello_common.
- While I agree that there are probably improvements that can be made to the API (see the [Zulip thread](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Hooking.20up.20glifo.20in.20vello_hybrid.2Fvello_cpu/with/583030800)), I have not investigated those yet and just implemented the minimum necessary to get tests to run. For now, core text rendering logic still lives in vello_hybrid/vello_cpu instead of an external crate or somewhere else. I also haven’t introduced any “render from sub-rect of atlas” API, instead just sticking as closely as possible to what Alex previously implemented.

Some other TODOs:
- When caching is enabled, zooming is currently extremely slow. Looking at the profile, it becomes clear that this is because deallocating and reallocating cached glyphs with guillotierre becomes a huge bottleneck with many glyphs. So we either need to explore using etagere for that, or investigate whether there is something else that causes the issue in our case. See also the video:

https://github.com/user-attachments/assets/1692d83f-29e9-436f-9c0b-2a0af2445bd5

- There also seems to be some other bug where some glyphs become corrupted at certain zoom levels, but it's not easily reproducible. I am not sure yet whether this is a regression introduced in this PR or already existed previously. If I find the time I might piggy-back a fix into this PR, otherwise it will be a follow-up:

https://github.com/user-attachments/assets/33ec8245-922b-47d6-b14f-314f11a19217

- It seems like when caching is enabled, the glyphs have some kind of funny "jump" when panning the viewport. I suspect that the subpixel offsets are somehow applied inversely, but I need to look into it. **EDIT: Nevermind, I realized that the answer is very simple, it's because we don't do any subpixel rendering in the y direction at all. Only in the x direction.**

https://github.com/user-attachments/assets/5c847d29-0f77-4620-b669-36742ec62604

So overall, there are definitely more things to investigate, but as you can see the integration seems to work overall!

# Changes
- The biggest change is probably to resources handling. Both renderers now require the user to pass a `Resources` (open for better name suggestions) for certain operations (anything text-related on the scene, and the final rendering operation). In vello_cpu, this basically only contains stuff relevant for glyph caching. In vello_hybrid, this additionally now contains the image cache, which previously lived in the “Renderer”. This allows us to share resources between scene encoding and the final rendering operation. The disadvantage is that it puts the burden on the user to always pass the correct resources, but hopefully this shouldn’t be too hard to get wrong. I don’t really see any better way of handling this otherwise. Perhaps this API can in the future be extended to allow the user passing custom resources (e.g. custom texture views for sampling). In vello_cpu, we could _probably_ get away with hiding this behind `RenderContext`, but it also gets a bit messy and I think having this explicit consistently is better.
- The default size of an atlas on Vello CPU was changed to 4096, I hope that’s a reasonable default? Or should it be smaller?
- I’ve made slight changes to how the glyph run builder API works. In the previous parley version, GlyphRunBuilder was a simple struct with a builder pattern, where you needed to manually the image cache and glyph caches to the method. In order to make it work better with the `Resources` struct used by both renders, I’ve made the builder generic over a backend. This way, the user simply has to call `glyph_run` once, passing the resources to it, and the rest then happens automatically.
- The previous benchmarks had some benches for measuring specific parts of the pipeline (e.g. maintain), since those are now more hidden away it’s not easy to benchmark anymore. Therefore, I’ve changed the benchmarks to simply be “integration” benchmarks for Vello CPU.
- I split up the previous `GlyphCaches` struct into a separate struct for outline/hint caches and atlas caching. The reason for this change is that I want it to be possible to lazily initialize glyph atlases and glyph renderers (especially in vello_cpu), such that the user doesn’t have to pay the cost for constructing those and keeping them in memory, if they don’t make use of the feature at all. This required making it easier to pass no atlas cache to all of the glyph drawing methods, and splitting those two up seemed like the easiest way. This way, you can still use the outline/hint cache even if atlas caching itself is disabled.
- I’m not sure what the best way of handling this is, but obviously we want to exercise the caching path in the tests. Unfortunately, caching simply can result in quite large pixel differences, so reusing existing snapshots is not really an option, and we need dedicated ones for new ones. Right now, I went down the path of simply adding an attribute that parametrizes the test on whether glyph caching should be enabled or not. I’m not sure if the better option here is to just manually hardcode the variants for each test, or something else. Open for suggestions here!

# Other Notes
- Something that make things awkward are recordings. Ideally, glifo would take care of handling and re-exporting all text-related dependencies, and vello_common doesn’t have any of that. However, since the `Recording` type lives in vello_common and we need to have the abilities to replay glyph drawing instructions, some of it _has_ to live in vello_common for now. In particular, I had to move the `Glyph` type to vello_common, in the long term this should probably move back to glifo.
- To expand on that, the fact that we have recordings makes things even harder. On current main, we are able to record normal outline glyphs but will panic for COLR and bitmap glyphs. However that exact setup is unfortunately difficult to replicate now. The reason being that the version currently on main relies on the ability to lower glyph runs into strip commands directly in vello_common. However, now that logic lives in glifo, and vello_common can’t take a dependency on glifo. Therefore, for now I’ve opted to simply record glyph render commmands as they are, but not actually provide any caching.
- For similar reasons, we need to duplicate the glyph run builder API in vello_common, which is also not ideal, but I wouldn't know how to deal with that better.


sparse_strips/vello_sparse_tests/snapshots/glyphs_colr_noto.png
LaurenzV, line 1: Some of the COLR glyph images changed, but that is expected. Previously, in Vello CPU we would first render them into an intermediate pixmap and then render that pixmap. Now, they are rendered directly into the scene (in case caching is disabled).

sparse_strips/vello_sparse_tests/snapshots/glyphs_colr_test_glyphs.png
LaurenzV, line 1: The two glyphs in row 8 that changed are expected changes, see also https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/web_tests/platform/linux/virtual/text-antialias/colrv1-expected.png. The old reference image was wrong there.

nicoburns:
Is it strictly necessary for scene encoding to have access to `Resources`? Would it be possible to defer this work until "render" is called so that `Resources` is just an internal implementation detail?

LaurenzV:
> Would it be possible to defer this work until "render" is called so that Resources is just an internal implementation detail?

I wouldn't see how, because you need to have data about the glyph atlas during scene encoding (such that you know with glyphs can be cached and which ones can't, as well as where the glyph lands in the atlas).

LaurenzV:
Unfortunately, I just realized there is more trouble...

Previously, COLR glyphs were rendered into separate pixmaps in Vello CPU and not at all supported in Vello Hybrid. Now, they are either rendered into the complete atlas page or pixmap. The problem now are destructive blend modes: Our current approach is to simply push new layers in _all_ wide tiles: https://github.com/linebender/vello/blob/01ce47ba052f9aa5e3dfa4979c8b2edd20861345/sparse_strips/vello_common/src/coarse.rs#L809-L821

As you can imagine, this becomes prohibitively expensive, especially if it's a nested layer and you have a huge atlas page. I'm afraid before this PR can be merged, we first need to come up with some strategy... Otherwise, as can be seen in the CI runs, the tests take prohibitively long.

Or, we just temporarily accept that COLR glyphs don't work as they should, and disable the tests...

glifo/src/glyph.rs
```diff
@@ -52,20 +53,6 @@ const BLACK_PACKED: u32 = PremulRgba8 {
...
-    ///
-    /// This ID is specific to the font being used and corresponds to the
-    /// glyph index within that font. It is *not* a Unicode code point.
-    pub id: u32,
-    /// X-offset in run, relative to transform.
-    pub x: f32,
-    /// Y-offset in run, relative to transform.
-    pub y: f32,
-}
-
```
LaurenzV, line 68: As mentioned, this has to live in vello_common for now, unfortunately.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -826,8 +834,6 @@ impl Scene {
 
         self.render_state.reset();
 
-        #[cfg(feature = "text")]
-        self.glyph_caches.as_mut().unwrap().maintain();
```
LaurenzV, line 830: We maintain the glyph cache after each render operation, so removing this should be the right call and there isn't anything else to do here, right?

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -591,21 +643,30 @@ impl RenderContext {
...
             &self.encoded_paints,
-            &self.image_registry,
+            &resources.image_registry,
         );
+        // TODO: We need to figure something out here API-wise. At the moment, the user can
+        // theoretically rasterize the same `RenderContext` multiple times without resetting in-between.
+        // However, if glyph caching is enabled, this method call could now evict that were previously
+        // assumed to exist in `RenderContext`, meaning that if the user rasterizes the same `RenderContext`
+        // again without resetting it, some of the cached glyphs might be stale and not exist anymore.
+        resources.after_render();
```
LaurenzV, line 658: Open for ideas on how to improve that situation.

sparse_strips/vello_dev_macros/src/test.rs
```diff
@@ -316,119 +340,217 @@ pub(crate) fn vello_test_inner(attr: TokenStream, item: TokenStream) -> TokenStr
...
+        invoke_test.clone(),
+    );
+
+    let cached_cpu_f32_fn_name = Ident::new(
+        &format!("{input_fn_name}_cpu_f32_scalar_cached"),
+        input_fn_name.span(),
+    );
+    let cached_cpu_f32_fn_name_str = cached_cpu_f32_fn_name.to_string();
+    let cached_hybrid_fn_name = Ident::new(
+        &format!("{input_fn_name}_hybrid_cached"),
```
LaurenzV, line 487: I will try to address the duplication in a future PR, as this method is getting very unwieldy. For now, I'm just going to add to the pile. 😓

LaurenzV:
Lots of follow ups will be required, but hopefully this is a good start...

glifo/src/renderer.rs
taj-p, line 1: This file has no dependency on Vello CPU nor Vello Hybrid. Let's delete this module and simply call this file `renderer.rs` and move it to `glifo/src/renderer.rs`

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -33,12 +37,57 @@ use vello_common::recording::{
...
-    colr::{ColrPainter, ColrRenderer},
-    glyph::{GlyphCaches, GlyphRenderer, GlyphRunBuilder, GlyphType, PreparedGlyph},
-};
+pub(crate) const DEFAULT_GLYPH_ATLAS_SIZE: u16 = 4096;
+// Why do we need this? The reason is that the way uploaded images work in Vello Hybrid
+// is different from how they work in Vello CPU.
+//
+// In Vello Hybrid, all images, regardless of whether they are user-uploaded images or cached glyphs,
+// are stored in an image atlas at a certain location. An image ID then uniquely resolves to an
+// atlas page index + a location on that page. Whenever we want to cache a new glyph, we simply allocate a location in the image atlas and then return the image ID associated with that location.
```
taj-p: Can we split this line so that it isn't ~200 chars long? I think we tend to keep comment length to around 100 characters before wrapping.

sparse_strips/vello_cpu/src/text.rs
```diff
@@ -3,92 +3,104 @@
...
     /// Shared cache data.
-    pub(crate) inner: GlyphAtlas,
-    /// One `Pixmap` per atlas page, grown on demand. Wrapped in `Arc` so
-    /// callers can cheaply share a page with a render context.
+    pub(crate) inner: GlifoGlyphAtlas,
+    /// One `Pixmap` per atlas page, grown on demand.
+    // It's a bit annoying to have this in an `Arc`, but it needs to be this way. During fine
+    // rasterization, we need to be able to easily clone the atlas page so that it can be shared
+    // across multiple threads. However, we also need to be able to mutate the pixmap to
+    // sync new glyphs. The way this is achieved is by calling `Arc::make_mut` when syncing
```
taj-p: Is this accurate/up to date? I can't find any `make_mut` calls of pixmap? It looks like we `get_mut` below instead

sparse_strips/vello_cpu/src/text.rs
```diff
@@ -3,92 +3,104 @@
...
-    #[inline]
-    pub fn page_pixmap(&self, page_index: usize) -> Option<&Arc<Pixmap>> {
-        self.pixmaps.get(page_index)
-    }
-
     /// Returns a mutable reference to the pixmap for `page_index`.
     #[inline]
-    pub fn page_pixmap_mut(&mut self, page_index: usize) -> Option<&mut Pixmap> {
+    pub(crate) fn page_pixmap_mut(&mut self, page_index: usize) -> Option<&mut Pixmap> {
         self.pixmaps.get_mut(page_index).and_then(Arc::get_mut)
```
taj-p: No need to address in this PR, but if `self.pixmaps.get_mut(page_index)` returns `Some`, then it's an invariant for `Arc::get_mut` to return `Some` too right (so, we should either panic or error if its `None`)?

sparse_strips/vello_hybrid/src/text.rs
```diff
@@ -157,47 +150,263 @@ impl GlyphCache for GpuGlyphAtlas {
...
+    fn push_clip_layer_impl(&mut self, clip: &BezPath) {
+        self.scene.push_layer(Some(clip), None, None, None, None);
+        self.layer_depth += 1;
+    }
+
+    fn push_blend_layer_impl(&mut self, blend_mode: BlendMode) {
+        if self.layer_depth == 0 && blend_mode != BlendMode::default() {
+            self.scene.push_layer(None, None, None, None, None);
+            self.inserted_root_wrapper = true;
+        }
```
taj-p: Why not move this optimization directly into `scene.rs` so that consumers needn't care about this at all?

Then, we can remove this wrapper struct altogether.

sparse_strips/vello_cpu/src/text.rs
```diff
@@ -233,52 +245,281 @@ impl GlyphCache for CpuGlyphAtlas {
...
+                DEFAULT_GLYPH_ATLAS_SIZE,
+                level,
+                render_mode,
+                GlyphCacheConfig::default(),
+            ));
+        }
+    }
+
+    /// Upload all pending bitmaps, rasterize pending outline/COLR glyphs, etc.
+    fn sync_glyph_cache(&mut self) {
```
taj-p, line 136: Can't we use this function to implement the custom `GlyphAtlas::insert` functionality? If so, we can remove the `GlyphAtlas` completely from the `vello_[cpu|hybrid]` consumer - simplifying the traits and logic an external renderer need implement

LaurenzV:
As discussed with Taj, I will be removing the glyph recording functionality, since we don't need it anyway and it allows for a lot of simplifications.

sparse_strips/vello_sparse_tests/snapshots/glyphs_colr_test_glyphs.png
LaurenzV: Actually, I'm not sure this is the case anymore. I looked into it, and it seems like the problem is that the clip bbox does not contain all parts of the glyphs. Previously, we would use the clip bounding box to determine the size of the bitmap to render the glyph into, hence anything that exceeds the clip bbox would get cut off. Now that we render the glyphs directly into the surface, it more or less gets ignored.

In the [COLR spec,](https://learn.microsoft.com/en-us/typography/opentype/spec/colr) it's mentioned that: "The clip box either should be large enough to encompass the color glyph for all instances, or should itself vary such that each instance of the clip box encompasses the instance color glyph."

So I think it's an error in the font itself.

taj-p (commented):
Let's chat in our meeting. I am concerned there is too much complexity leaking out of `glifo` into `vello_[cpu|hybrid]`. 

I think my recommendations are:

- Remove `ColrRenderer`, `AtlasReplayTarget`, `GlyphRenderer`, `GlyphAtlasBackend` and replace them with a `DrawTarget` trait (yes, this is "like" AnyRender / imaging / Vello API, but let's defer that decision for now).
  - My hope is that a renderer need only implement 1 or 2 traits to support glifo.
- Remove the `GlyphCache` trait - can't all the current functionality be implemented by the new resources model and the `GlyphCache` owned within `glifo` itself?

This is some vibe coded ideas I've been thinking about: https://github.com/linebender/vello/compare/laurenz/parley_migration...tajp/parley_migration_suggestions?expand=1.

sparse_strips/vello_cpu/src/text.rs
LaurenzV: Ah yes! It's actually the other way around, `get_mut` is the non-cloning one, and `make_mut` is the cloning one. 😓

sparse_strips/vello_hybrid/src/text.rs
LaurenzV: Because this is only an optimization valid specifically for COLR glyphs. We don't want glyphs to blend with existing page contents (i.e. each glyph itself should be isolated), hence why it's fine to just push an empty layer before doing anything.

On the other hand, if the user says they want to push a new blend layer, then their intention is to blend it with the current backdrop. So you cant just push an empty layer there, because then you are always blending into the transparent backdrop instead of the actual background.

taj-p: Ah yes! Whoops!

sparse_strips/vello_cpu/src/text.rs
LaurenzV: Indeed, but both call-sites currently unwrap here anyway so it doesn't make a difference. But I will change it, for better clarity.

LaurenzV:
@taj-p Thanks for the great suggestion, code is now simpler indeed, let me know what you think! I haven't yet merged the two traits because conceptually they do kind of do different things, but I'm happy to revisit this in the future.

There's probably still a lot more that can be improved, but I hope this is a good start now! Feel free to make any changes you want and merge yourself, or I can do it once you are happy and approved.

sparse_strips/vello_cpu/src/text.rs
LaurenzV: Hmm, could you elaborate on that? The syncing logic is pretty different between Vello CPU and Vello Hybrid, so I wouldn't quite know how to unify those unless we introduce more traits again.

LaurenzV:
I also removed the COLR scene wrapper and just always wrap color glyphs in a push/pop command, which seems like the correct thing to do on both, Vello Hybrid and Vello CPU (regardless of the `default_blending_problem`).

sparse_strips/vello_hybrid/src/text.rs
```diff
@@ -0,0 +1,315 @@
...
+        });
+
+        let padding = u32::from(GLYPH_PADDING);
+        self.refresh_pending_glyph_uploads();
+        for upload in &self.pending_glyph_uploads_scratch {
+            let resource = self.image_cache.get(upload.image_id).unwrap();
+            let dst_x = resource.offset[0] as u32 + padding;
+            let dst_y = resource.offset[1] as u32 + padding;
+            upload_to_atlas(backend, &self.image_cache, upload, dst_x, dst_y);
+        }
```
taj-p: I don't believe we need the `pending_glyph_uploads_scratch` member nor the `refresh_pending_glyph_uploads` method. I think this can be inlined to:

```rs
        const PADDING: u32 = GLYPH_PADDING as u32;

        if let Some(glyph_resources) = self.glyph_resources.as_mut() {
            for upload in glyph_resources.glyph_atlas.drain_pending_uploads() {
                let resource = self.image_cache.get(upload.image_id).unwrap();
                let dst_x = resource.offset[0] as u32 + PADDING;
                let dst_y = resource.offset[1] as u32 + PADDING;
                upload_to_atlas(backend, &self.image_cache, &upload, dst_x, dst_y);
            }
        }
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -179,7 +185,41 @@ impl WebGlRenderer {
...
+            &mut resources.image_cache,
+            render_size,
+            true,
+            RootRenderTarget::UserSurface,
+        )?;
+
+        #[cfg(feature = "text")]
+        {
+            resources.after_render(self, |renderer, rects| {
+                clear_atlas_regions(renderer, rects.iter().cloned());
```
taj-p: Why perform the copy instead of passing `rects` directly?

```diff
diff --git a/sparse_strips/vello_hybrid/src/render/webgl.rs b/sparse_strips/vello_hybrid/src/render/webgl.rs
index 04e4e1ae..91889c47 100644
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -216,7 +216,7 @@ impl WebGlRenderer {
         #[cfg(feature = "text")]
         {
             resources.after_render(self, |renderer, rects| {
-                clear_atlas_regions(renderer, rects.iter().cloned());
+                clear_atlas_regions(renderer, rects);
             });
         }
 
@@ -670,7 +670,7 @@ impl WebGlRenderer {
 #[cfg(feature = "text")]
 fn clear_atlas_regions(
     renderer: &mut WebGlRenderer,
-    rects: impl Iterator<Item = PendingClearRect>,
+    rects: &[PendingClearRect],
 ) {
     // TODO: Similarly to wgpu, maybe this can be done in a more effective
     // way?

```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -603,6 +667,25 @@ impl WebGlRenderer {
     }
 }
 
+#[cfg(feature = "text")]
+fn clear_atlas_regions(
+    renderer: &mut WebGlRenderer,
+    rects: impl Iterator<Item = PendingClearRect>,
+) {
+    // TODO: Similarly to wgpu, maybe this can be done in a more effective
+    // way?
```
taj-p, line 674: 100

sparse_strips/vello_hybrid/src/resources.rs
```diff
@@ -0,0 +1,54 @@
...
+use vello_common::multi_atlas::AtlasConfig;
+
+/// Persistent resources required by Vello Hybrid for rendering.
+#[derive(Debug)]
+pub struct Resources {
+    pub(crate) image_cache: ImageCache,
+    #[cfg(feature = "text")]
+    pub(crate) glyph_prep_cache: GlyphPrepCache,
+    #[cfg(feature = "text")]
+    pub(crate) pending_glyph_uploads_scratch: Vec<PendingBitmapUpload>,
```
taj-p:
```suggestion
```

See above comment - I don't believe this is required

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -33,12 +37,60 @@ use vello_common::recording::{
...
+/// Persistent resources required by Vello CPU for rendering.
+#[derive(Debug, Default)]
+pub struct Resources {
+    pub(crate) image_registry: ImageRegistry,
+    #[cfg(feature = "text")]
+    pub(crate) glyph_prep_cache: GlyphPrepCache,
+    #[cfg(feature = "text")]
+    pub(crate) pending_glyph_uploads_scratch: Vec<PendingBitmapUpload>,
+    #[cfg(feature = "text")]
+    pub(crate) pending_glyph_clear_rects_scratch: Vec<PendingClearRect>,
```
taj-p:
```suggestion
```

We can remove these scratch buffers and intermediate copies by doing something like:

```diff
diff --git a/sparse_strips/vello_cpu/src/render.rs b/sparse_strips/vello_cpu/src/render.rs
index 7363dc5f..701e3295 100644
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -9,8 +9,6 @@ use crate::dispatch::Dispatcher;
 use crate::text::{GlyphAtlasResources, GlyphRunBuilder};
 #[cfg(feature = "text")]
 use glifo::GlyphPrepCache;
-#[cfg(feature = "text")]
-use glifo::atlas::{PendingBitmapUpload, PendingClearRect};
 
 #[cfg(feature = "multithreading")]
 use crate::dispatch::multi_threaded::MultiThreadedDispatcher;
@@ -66,10 +64,6 @@ pub struct Resources {
     pub(crate) image_registry: ImageRegistry,
     #[cfg(feature = "text")]
     pub(crate) glyph_prep_cache: GlyphPrepCache,
-    #[cfg(feature = "text")]
-    pub(crate) pending_glyph_uploads_scratch: Vec<PendingBitmapUpload>,
-    #[cfg(feature = "text")]
-    pub(crate) pending_glyph_clear_rects_scratch: Vec<PendingClearRect>,
     // Will be initialized lazily on first use.
     #[cfg(feature = "text")]
     pub(crate) glyph_resources: Option<GlyphAtlasResources>,
diff --git a/sparse_strips/vello_cpu/src/text.rs b/sparse_strips/vello_cpu/src/text.rs
index d91a262b..74bf4117 100644
--- a/sparse_strips/vello_cpu/src/text.rs
+++ b/sparse_strips/vello_cpu/src/text.rs
@@ -86,14 +86,6 @@ impl GlyphAtlasResources {
     pub(crate) fn maintain(&mut self) {
         self.glyph_atlas.maintain(&mut self.image_cache);
     }
-
-    /// Returns a mutable reference to the pixmap for `page_index`.
-    pub(crate) fn page_pixmap_mut(&mut self, page_index: usize) -> Option<&mut Pixmap> {
-        self.pixmaps.get_mut(page_index).map(|p|
-            // This should never fail when this method is called, hence why we unwrap
-            // explicitly instead of using `and_then`.
-            Arc::get_mut(p).expect("pixmap isn't uniquely owned"))
-    }
 }
 
 /// Ensure a pixmap exists for the given page, creating it if needed.
@@ -142,24 +134,21 @@ impl Resources {
 
     /// Upload all pending bitmaps, rasterize pending outline/COLR glyphs, etc.
     fn sync_glyph_cache(&mut self) {
-        self.pending_glyph_uploads_scratch.clear();
         let glyph_resources = self
             .glyph_resources
             .as_mut()
             .expect("glyph atlas resources must exist before syncing");
-        self.pending_glyph_uploads_scratch
-            .extend(glyph_resources.glyph_atlas.drain_pending_uploads());
 
         // Upload all pending bitmap glyphs to the image atlas.
-        for upload in &self.pending_glyph_uploads_scratch {
+        for upload in glyph_resources.glyph_atlas.drain_pending_uploads() {
+            let page_index = upload.atlas_slot.page_index as usize;
             ensure_page(
                 &mut glyph_resources.pixmaps,
                 glyph_resources.page_width,
                 glyph_resources.page_height,
-                upload.atlas_slot.page_index as usize,
+                page_index,
             );
-            let pixmap = glyph_resources
-                .page_pixmap_mut(upload.atlas_slot.page_index as usize)
+            let pixmap = Arc::get_mut(&mut glyph_resources.pixmaps[page_index])
                 .expect("atlas upload refers to a missing page");
             copy_pixmap_to_atlas(
                 &upload.pixmap,
@@ -200,19 +189,14 @@ impl Resources {
     }
 
     fn clear_evicted_glyph_atlas_regions(&mut self) {
-        self.pending_glyph_clear_rects_scratch.clear();
         let glyph_resources = self
             .glyph_resources
             .as_mut()
             .expect("glyph atlas resources must exist before clearing");
-        self.pending_glyph_clear_rects_scratch
-            .extend(glyph_resources.glyph_atlas.drain_pending_clear_rects());
-
-        for clear in &self.pending_glyph_clear_rects_scratch {
-            let pixmap = glyph_resources
-                .page_pixmap_mut(clear.page_index as usize)
+        for clear in glyph_resources.glyph_atlas.drain_pending_clear_rects() {
+            let pixmap = Arc::get_mut(&mut glyph_resources.pixmaps[clear.page_index as usize])
                 .expect("atlas clear rect refers to a missing page");
-            clear_pixmap_region(pixmap, *clear);
+            clear_pixmap_region(pixmap, clear);
         }
     }
 }
```

sparse_strips/vello_hybrid/src/resources.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[cfg(feature = "text")]
+    pub(crate) pending_glyph_clear_rects_scratch: Vec<PendingClearRect>,
```
taj-p: We can remove this scratch buffer and its intermediate copies by doing something like:

```diff
diff --git a/sparse_strips/vello_hybrid/src/render/webgl.rs b/sparse_strips/vello_hybrid/src/render/webgl.rs
index 91889c47..1fa0691e 100644
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -215,8 +215,8 @@ impl WebGlRenderer {
 
         #[cfg(feature = "text")]
         {
-            resources.after_render(self, |renderer, rects| {
-                clear_atlas_regions(renderer, rects);
+            resources.after_render(self, |renderer, rect| {
+                clear_atlas_region(renderer, rect);
             });
         }
 
@@ -668,22 +668,15 @@ impl WebGlRenderer {
 }
 
 #[cfg(feature = "text")]
-fn clear_atlas_regions(
-    renderer: &mut WebGlRenderer,
-    rects: &[PendingClearRect],
-) {
-    // TODO: Similarly to wgpu, maybe this can be done in a more effective
-    // way?
-    for rect in rects {
-        let padding = u32::from(GLYPH_PADDING);
-        let offset = [
-            u32::from(rect.x).saturating_sub(padding),
-            u32::from(rect.y).saturating_sub(padding),
-        ];
-        let width = u32::from(rect.width) + padding * 2;
-        let height = u32::from(rect.height) + padding * 2;
-        renderer.clear_atlas_region(AtlasId::new(rect.page_index), offset, width, height);
-    }
+fn clear_atlas_region(renderer: &mut WebGlRenderer, rect: &PendingClearRect) {
+    let padding = u32::from(GLYPH_PADDING);
+    let offset = [
+        u32::from(rect.x).saturating_sub(padding),
+        u32::from(rect.y).saturating_sub(padding),
+    ];
+    let width = u32::from(rect.width) + padding * 2;
+    let height = u32::from(rect.height) + padding * 2;
+    renderer.clear_atlas_region(AtlasId::new(rect.page_index), offset, width, height);
 }
 
 /// Contains the WebGL programs and resources for rendering.
diff --git a/sparse_strips/vello_hybrid/src/render/wgpu.rs b/sparse_strips/vello_hybrid/src/render/wgpu.rs
index f1b3d447..88a26c94 100644
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -275,8 +275,8 @@ impl Renderer {
 
         encoded_paints.truncate(scene_paint_count);
         #[cfg(feature = "text")]
-        resources.after_render(self, |renderer, rects| {
-            clear_atlas_regions(queue, renderer, rects.iter().cloned());
+        resources.after_render(self, |renderer, rect| {
+            clear_atlas_region(queue, renderer, rect);
         });
         result
     }
@@ -785,39 +785,32 @@ impl Renderer {
 }
 
 #[cfg(feature = "text")]
-fn clear_atlas_regions(
-    queue: &Queue,
-    renderer: &mut Renderer,
-    rects: impl Iterator<Item = PendingClearRect>,
-) {
-    // TODO: Can we optimize this more?
-    for rect in rects {
-        let byte_count = rect.width as usize * rect.height as usize * 4;
-        renderer.atlas_clear_scratch.resize(byte_count, 0);
-        queue.write_texture(
-            wgpu::TexelCopyTextureInfo {
-                texture: renderer.atlas_texture(),
-                mip_level: 0,
-                origin: wgpu::Origin3d {
-                    x: rect.x as u32,
-                    y: rect.y as u32,
-                    z: rect.page_index,
-                },
-                aspect: wgpu::TextureAspect::All,
+fn clear_atlas_region(queue: &Queue, renderer: &mut Renderer, rect: &PendingClearRect) {
+    let byte_count = rect.width as usize * rect.height as usize * 4;
+    renderer.atlas_clear_scratch.resize(byte_count, 0);
+    queue.write_texture(
+        wgpu::TexelCopyTextureInfo {
+            texture: renderer.atlas_texture(),
+            mip_level: 0,
+            origin: wgpu::Origin3d {
+                x: rect.x as u32,
+                y: rect.y as u32,
+                z: rect.page_index,
             },
-            &renderer.atlas_clear_scratch[..byte_count],
-            wgpu::TexelCopyBufferLayout {
-                offset: 0,
-                bytes_per_row: Some(rect.width as u32 * 4),
-                rows_per_image: None,
-            },
-            Extent3d {
-                width: rect.width as u32,
-                height: rect.height as u32,
-                depth_or_array_layers: 1,
-            },
-        );
-    }
+            aspect: wgpu::TextureAspect::All,
+        },
+        &renderer.atlas_clear_scratch[..byte_count],
+        wgpu::TexelCopyBufferLayout {
+            offset: 0,
+            bytes_per_row: Some(rect.width as u32 * 4),
+            rows_per_image: None,
+        },
+        Extent3d {
+            width: rect.width as u32,
+            height: rect.height as u32,
+            depth_or_array_layers: 1,
+        },
+    );
 }
 
 /// Defines the GPU resources and pipelines for rendering.
diff --git a/sparse_strips/vello_hybrid/src/resources.rs b/sparse_strips/vello_hybrid/src/resources.rs
index 8fdfa932..5a17ce54 100644
--- a/sparse_strips/vello_hybrid/src/resources.rs
+++ b/sparse_strips/vello_hybrid/src/resources.rs
@@ -3,15 +3,10 @@
 
 //! Persistent renderer resources shared across frames.
 
-#[cfg(feature = "text")]
-use alloc::vec::Vec;
-
 #[cfg(feature = "text")]
 use crate::text::GlyphAtlasResources;
 #[cfg(feature = "text")]
 use glifo::GlyphPrepCache;
-#[cfg(feature = "text")]
-use glifo::atlas::{PendingBitmapUpload, PendingClearRect};
 use vello_common::image_cache::ImageCache;
 use vello_common::multi_atlas::AtlasConfig;
 
@@ -22,8 +17,6 @@ pub struct Resources {
     #[cfg(feature = "text")]
     pub(crate) glyph_prep_cache: GlyphPrepCache,
     #[cfg(feature = "text")]
-    pub(crate) pending_glyph_clear_rects_scratch: Vec<PendingClearRect>,
-    #[cfg(feature = "text")]
     pub(crate) glyph_resources: Option<GlyphAtlasResources>,
 }
 
@@ -34,8 +27,6 @@ impl Resources {
             image_cache: ImageCache::new_with_config(AtlasConfig::default()),
             #[cfg(feature = "text")]
             glyph_prep_cache: GlyphPrepCache::default(),
-            #[cfg(feature = "text")]
-            pending_glyph_clear_rects_scratch: Vec::new(),
             // Will be initialized lazily.
             #[cfg(feature = "text")]
             glyph_resources: None,
diff --git a/sparse_strips/vello_hybrid/src/text.rs b/sparse_strips/vello_hybrid/src/text.rs
index 6b705a91..99610f9e 100644
--- a/sparse_strips/vello_hybrid/src/text.rs
+++ b/sparse_strips/vello_hybrid/src/text.rs
@@ -17,6 +17,7 @@ use glifo::{
     AtlasCacher, AtlasSlot, DrawSink, GLYPH_PADDING, Glyph, GlyphAtlas, GlyphCacheConfig,
     GlyphRunBackend, ImageCache,
 };
+
 use peniko::BlendMode;
 use peniko::color::palette::css::BLACK;
 use peniko::color::{AlphaColor, Srgb};
@@ -122,19 +123,14 @@ impl Resources {
     pub(crate) fn after_render<T>(
         &mut self,
         backend: &mut T,
-        mut clear_rects: impl FnMut(&mut T, &[PendingClearRect]),
+        mut clear_rect: impl FnMut(&mut T, &PendingClearRect),
     ) {
-        self.refresh_pending_glyph_clear_rects();
-        clear_rects(backend, &self.pending_glyph_clear_rects_scratch);
-    }
-
-    pub(crate) fn refresh_pending_glyph_clear_rects(&mut self) {
         self.glyph_prep_cache.maintain();
-        self.pending_glyph_clear_rects_scratch.clear();
         if let Some(glyph_resources) = self.glyph_resources.as_mut() {
             glyph_resources.maintain(&mut self.image_cache);
-            self.pending_glyph_clear_rects_scratch
-                .extend(glyph_resources.glyph_atlas.drain_pending_clear_rects());
+            for rect in glyph_resources.glyph_atlas.drain_pending_clear_rects() {
+                clear_rect(backend, &rect);
+            }
         }
     }
 }
```

sparse_strips/vello_hybrid/src/text.rs
```diff
@@ -0,0 +1,315 @@
...
+            #[expect(
+                clippy::cast_possible_truncation,
+                reason = "atlas dimensions are configured to fit in u16"
+            )]
+            let (atlas_width, atlas_height) = {
+                let (width, height) = self.image_cache.atlas_manager().config().atlas_size;
+                (width as u16, height as u16)
+            };
+            // TODO: Use a config optionally provided by the user!
+            self.glyph_resources = Some(GlyphAtlasResources::with_config(
```
taj-p, line 64: Ah, so the user pays the cost of duplicate alpha and strip storage and all the other buffer allocations of a new renderer. That's quite a memory cost. Ideally, these buffers are shared between renderers, but to enable that, we would require deferring alpha/strip evaluation until render time - that is, buffer all draw commands and then, on render, execute the caching glyph pipeline first, reset and pass internal buffer allocations to the screen renderer, and then render the screen.

No action here. Just sharing thoughts on opportunities.

sparse_strips/vello_cpu/src/text.rs
taj-p: I believe this has been addressed in your updates 🙏

glifo/src/colr.rs
```diff
@@ -399,46 +402,58 @@ pub(crate) fn convert_bounding_box(rect: BoundingBox<f32>) -> Rect {
     )
 }
 
-impl ColrRenderer for AtlasCommandRecorder {
+impl DrawSink for AtlasCommandRecorder {
```
taj-p: Shouldn't this impl be in `glifo/src/atlas/commands.rs` so that it lives alongside `AtlasCommandRecorder`? I think that seems like a better home than `colr.rs`

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -237,7 +283,18 @@ impl WebGlRenderer {
             &mut self.programs.resources.stub_atlas_texture_array,
         );
 
-        let result = self.render_scene(scene, &atlas_render_size, false);
+        let mut dummy_image_cache = self
```
taj-p, line 287: It's unfortunate we need to do this, could we leave a TODO to explore removing this by potentially passing `None` or similar to `render_scene` for `image_cache`?

sparse_strips/vello_cpu/src/render.rs
taj-p: Sorry, I ran out of my timebox this morning in reviewing the PR

taj-p (commented):
Sorry, I ran out of time this morning to finish the review. The migration looks good. 

Some of the comments are addressed in https://github.com/linebender/vello/pull/1571 (feel free to merge that PR / cherry pick / whatever you like).

There's a current high memory cost to glyph caching that we're likely going to need to address before being able to consume it. Either by device detection or improving memory efficiency. But that can definitely come later.

glifo/src/interface.rs
```diff
@@ -0,0 +1,67 @@
...
+//! Traits for rendering glyphs and replaying them.
+
+use crate::atlas::{AtlasPaint, AtlasSlot};
+use crate::color::{AlphaColor, Srgb};
+use crate::kurbo::{Affine, BezPath, Rect};
+use crate::peniko::BlendMode;
+use vello_common::paint::{Image, ImageSource, Tint};
+
+/// A sink for low-level glyph drawing commands.
+pub trait DrawSink {
```
taj-p, line 15: Can we leave a comment stating that `DrawSink` is a temporary trait until such time that we replace it with a "standard" imaging API

sparse_strips/vello_cpu/src/render.rs
taj-p: `Resources` can have a generation counter (incremented on `after_render`) and `[before|after]_render` early return if the generation hasn't changed since last render. We'll need `reset` to increment the generation too or similar.

sparse_strips/vello_hybrid/src/resources.rs
LaurenzV: Ah.. sorry about those, I am very certain they were necessary in the past due to borrowing issues, but seems like some of those disappeared as part of the refactors!

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -33,12 +37,57 @@ use vello_common::recording::{
...
 use vello_common::util::is_axis_aligned;
+
 #[cfg(feature = "text")]
-use vello_common::{
-    color::{AlphaColor, Srgb},
-    colr::{ColrPainter, ColrRenderer},
-    glyph::{GlyphCaches, GlyphRenderer, GlyphRunBuilder, GlyphType, PreparedGlyph},
-};
+pub(crate) const DEFAULT_GLYPH_ATLAS_SIZE: u16 = 4096;
+// Why do we need this? The reason is that the way uploaded images work in Vello Hybrid
```
taj-p, line 41: We might be able to avoid this but let's not tackle that now. The PR is in a much better state already.

taj-p (approved):
Considering this PR was supposed to simply factor out the `vello_*` logic into corresponding crates, I think I've overstepped how much improvement we should try make to `glifo` in this PR.

It works and it looks good! 

cLGTM if the prior comments are adopted

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: 100 what? 😄

sparse_strips/vello_hybrid/src/text.rs
LaurenzV: Hmm yeah, I think it would be tricky! We should do some measurements to determine memory impact.

glifo/src/colr.rs
LaurenzV: Makes sense.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Will add a TODO.

glifo/src/interface.rs
LaurenzV: Will add a TODO.

sparse_strips/vello_hybrid/src/resources.rs
LaurenzV: Applied from your branch.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: Applied from your branch.

sparse_strips/vello_hybrid/src/resources.rs
LaurenzV: Applied from your branch.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Applied from your branch.

sparse_strips/vello_hybrid/src/text.rs
LaurenzV: Applied from your branch.

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: Ohhh I generally write this to mean "100% - totally agree!" - as in I strongly agree.

grebmeg:
This is huge work! well done @LaurenzV! 👏

PR #1563, opened 2026-04-09
vello_cpu: Fix artifacts for bicubic image rendering on x86

LaurenzV:
Encountered this while working on #1562.

Basically, the problem is that `f32_to_u8` does not have consistent behavior for negative floating point numbers. Therefore, we need to ensure that the values are always positive. Otherwise, on x86 SIMD, the values will wrap to 255 instead of 0, resulting in artifacts for certain pixels where the pixels would end up as white.

We actually already had a guard for this in the code for bicubic rendering, but the problem is that we were applying `min(alphas)` _after_ clamping. This means that in case the alpha values themselves where negatives, we would end up with negative values again. The solution here is to do the clamping in the very end.

Here is an example of the artifacts that would occur:

image

### DIFF
--- a/sparse_strips/vello_common/src/util.rs
+++ b/sparse_strips/vello_common/src/util.rs
@@ -15,6 +15,8 @@ use peniko::kurbo::common::FloatFuncs as _;
 ///
 /// **Important note: The values need to be between 0.0 and 1.0, otherwise you might
 /// get inconsistent results across different platforms.**
+// We can't guarantee correctness for values < 0.0 due to a restriction in fearless_simd:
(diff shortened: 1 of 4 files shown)


LaurenzV:
(This is required before merging #1562.)

waywardmonkeys (approved):
As long as the alpha is always between 0 and 1, then this is correct, I think.

LaurenzV:
> As long as the alpha is always between 0 and 1

It can be slightly below 0 or above 1, that's why we need to clamp it. But I don't think this should cause any problems here!

PR #1566, opened 2026-04-10
vello_hybrid: Use scissor rects for filter pass aplication

LaurenzV:
Not going to add much more in the description of this PR because I've already documented everything using comments.

On current main, rendering a single filtered rectangle on my Samsung Galaxy Tab A6 only yields 3-4FPS. With this change, I get to 13-14FPS, so a pretty huge speedup. If I use Canvas2D, I get 20FPS, so there is definitely more room for improvement, but I think this should be a good start!

PR #1567, opened 2026-04-10
vello_hybrid: Use scissor rect when rendering into intermediate layers

LaurenzV:
Stacked on top of #1566. Basically the same as there, but instead when rendering the original contents of the filter layer.

I'm not sure why it helps, because it shouldn't be the case that we are actually rendering any content outside of the targeted region. But it consistently gives me 1-3 additional FPS on my Samsung tablet when rendering scenes that contain filters. Are there any disadvantages I should be aware of?

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1995,7 +1995,7 @@ impl WebGlRendererContext<'_> {
         }
         self.programs.upload_strips(self.gl, strips);
 
-        match &target {
+        let scissor_rect = match &target {
             StripPassRenderTarget::Output(OutputTarget::IntermediateTexture(layer_id)) => {
                 let image_id = self
                     .filter_context
@@ -2057,6 +2057,13 @@ impl WebGlRendererContext<'_> {
                     self.programs.strip_uniforms.config_fs_block_index,
                     Some(buf),
                 );
+
+                Some([
+                    resources.offset[0] as i32,
+                    resources.offset[1] as i32,
+                    resources.width as i32,
+                    resources.height as i32,
+                ])
             }
             StripPassRenderTarget::Output(OutputTarget::FinalView) => {
                 self.gl.bind_framebuffer(
@@ -2077,6 +2084,8 @@ impl WebGlRendererContext<'_> {
                     self.programs.strip_uniforms.config_fs_block_index,
                     Some(&self.programs.resources.view_config_buffer),
                 );
+
+                None
             }
             StripPassRenderTarget::SlotTexture(ix) => {
                 self.gl.bind_framebuffer(
@@ -2103,7 +2112,16 @@ impl WebGlRendererContext<'_> {
                     self.programs.strip_uniforms.config_fs_block_index,
                     Some(&self.programs.resources.slot_config_buffer),
                 );
+
+                None
             }
+        };
+
+        if let Some([x, y, width, height]) = scissor_rect {
+            self.gl.enable(WebGl2RenderingContext::SCISSOR_TEST);
+            self.gl.scissor(x, y, width, height);
+        } else {
+            self.gl.disable(WebGl2RenderingContext::SCISSOR_TEST);
         }
 
         // Clear framebuffer if requested.
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -2241,10 +2241,15 @@ impl RendererContext<'_> {
             }
         }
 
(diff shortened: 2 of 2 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p, line 1: There is asymmetry between how the scissor rect is handled in WebGL compared to WGPU. WGPU computes the scissor rect (if any) within the match arms and then applies the scissoring after matching. WebGL applies the scissor rect within the match arms. 

Could we make the two implementations a bit more symmetrical via something like the below?


```diff
diff --git a/sparse_strips/vello_hybrid/src/render/webgl.rs b/sparse_strips/vello_hybrid/src/render/webgl.rs
index 18bdbf4e..4ee9f576 100644
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1995,7 +1995,7 @@ impl WebGlRendererContext<'_> {
         }
         self.programs.upload_strips(self.gl, strips);
 
-        match &target {
+        let scissor_rect = match &target {
             StripPassRenderTarget::Output(OutputTarget::IntermediateTexture(layer_id)) => {
                 let image_id = self
                     .filter_context
@@ -2014,13 +2014,6 @@ impl WebGlRendererContext<'_> {
                 let atlas_height = self.programs.resources.filter_atlas_height;
                 self.gl
                     .viewport(0, 0, atlas_width as i32, atlas_height as i32);
-                self.gl.enable(WebGl2RenderingContext::SCISSOR_TEST);
-                self.gl.scissor(
-                    resources.offset[0] as i32,
-                    resources.offset[1] as i32,
-                    resources.width as i32,
-                    resources.height as i32,
-                );
 
                 let filter_textures = self.filter_context.filter_textures.get(layer_id).unwrap();
                 let strip_offset_x = resources.offset[0] as i32
@@ -2064,9 +2057,15 @@ impl WebGlRendererContext<'_> {
                     self.programs.strip_uniforms.config_fs_block_index,
                     Some(buf),
                 );
+
+                Some([
+                    resources.offset[0] as i32,
+                    resources.offset[1] as i32,
+                    resources.width as i32,
+                    resources.height as i32,
+                ])
             }
             StripPassRenderTarget::Output(OutputTarget::FinalView) => {
-                self.gl.disable(WebGl2RenderingContext::SCISSOR_TEST);
                 self.gl.bind_framebuffer(
                     WebGl2RenderingContext::FRAMEBUFFER,
                     self.programs.resources.view_framebuffer_override.as_ref(),
@@ -2085,9 +2084,10 @@ impl WebGlRendererContext<'_> {
                     self.programs.strip_uniforms.config_fs_block_index,
                     Some(&self.programs.resources.view_config_buffer),
                 );
+
+                None
             }
             StripPassRenderTarget::SlotTexture(ix) => {
-                self.gl.disable(WebGl2RenderingContext::SCISSOR_TEST);
                 self.gl.bind_framebuffer(
                     WebGl2RenderingContext::FRAMEBUFFER,
                     Some(&self.programs.resources.slot_framebuffers[*ix as usize]),
@@ -2112,7 +2112,16 @@ impl WebGlRendererContext<'_> {
                     self.programs.strip_uniforms.config_fs_block_index,
                     Some(&self.programs.resources.slot_config_buffer),
                 );
+
+                None
             }
+        };
+
+        if let Some([x, y, width, height]) = scissor_rect {
+            self.gl.enable(WebGl2RenderingContext::SCISSOR_TEST);
+            self.gl.scissor(x, y, width, height);
+        } else {
+            self.gl.disable(WebGl2RenderingContext::SCISSOR_TEST);
         }
 
         // Clear framebuffer if requested.
```

taj-p, line 2131: Per @nical's below advice, I believe the scissor rect improving performance is not counter intuitive because it is also preventing this clear operation from operating over the entire atlas area (and instead just clearing the required content area). 

> When setting up a render pass, try to avoid LoadOp::Load if the pass will only overwrite pixels and you can afford to lose what was previously stored. That enables the GPU to skip expensive reads from the framebuffer to initialize the on-chip tile memory. The caveat is that if you LoadOp::Clear but only touch a small portion of the render target, the GPU will have to spend extra time clearing the portions on the render target you didn't touch which I've found to be slow on some devices. LoadOp::DontCare is ideal on native but on the web I think that it defaults to LoadOp::Clear (since it's not part of the WebGPU spec).

So, I think we should proceed with this optimisation. (Thanks again Nical! 🙏 )

taj-p (approved):
LGTM! Nice find Laurenz 🎉 🙌

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Good suggestion, thanks!

PR #1568, opened 2026-04-13
vello_hybrid: Adjust atlas config to actual device capabilities

LaurenzV:
Right now, `AtlasConfig` implements `Default` with the following configuration:

```rs
impl Default for AtlasConfig {
    fn default() -> Self {
        Self {
            initial_atlas_count: 1,
            max_atlases: 8,
            atlas_size: (4096, 4096),
            auto_grow: true,
            allocation_strategy: AllocationStrategy::FirstFit,
        }
    }
}
```

Given that this is part of `RenderSettings`, users should be able to assume that they can do `RenderSettings::default` and it just works everywhere. However, this is currently not the case, as we will blindly try to allocate the resources without checking the actual device capabilties.

This PR proposes to change this by treating the config as a strong recommendation, but allowing backends to choose different values based on the capabilities of the device it runs on. I experimented with another approach where we instead introduce an `Auto` struct for some fields to allow the users to explicitly opt in for that, and instead let the backend error out in case the user-requested parameters are not compatible with the device. However, this resulted in a much larger code diff, hence why I think just mutating the settings is the easier approach for now.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -31,8 +31,8 @@ use crate::{
             GPU_ENCODED_IMAGE_SIZE_TEXELS, GPU_LINEAR_GRADIENT_SIZE_TEXELS,
             GPU_RADIAL_GRADIENT_SIZE_TEXELS, GPU_SWEEP_GRADIENT_SIZE_TEXELS, GpuEncodedImage,
             GpuEncodedPaint, GpuLinearGradient, GpuRadialGradient, GpuSweepGradient,
-            pack_image_offset, pack_image_params, pack_image_size, pack_radial_kind_and_swapped,
-            pack_texture_width_and_extend_mode, pack_tint,
+            normalize_atlas_config, pack_image_offset, pack_image_params, pack_image_size,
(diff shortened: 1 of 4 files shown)


taj-p (approved):

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -142,7 +149,14 @@ impl WebGlRenderer {
...
         }
 
+        let mut settings = settings;
         let max_texture_dimension_2d = get_max_texture_dimension_2d(&gl);
+        normalize_atlas_config(
+            &mut settings.atlas_config,
+            max_texture_dimension_2d,
+            get_max_texture_array_layers(&gl),
+            1,
+        );
```
grebmeg, line 159: The change in this PR is a bit controversial. Consider testing the renderer against edge cases to ensure it properly handles failure modes. Also, it now hides from the consumer that its configuration has been adjusted based on device capabilities.

LaurenzV: Would you prefer a different implementation?

grebmeg: It’s fine to leave this as it is for now, but whenever we’re working near this functionality, I’d consider updating it to make the config changes more explicit (or rather consider it's as setup-config) instead of silently applying them.

PR #1569, opened 2026-04-13
Fix clipping in recording replay

AdrianEddy:
`generate_strips_from_commands` (used by `prepare_recording`) was passing `None` for the clip path to the strip generator, while the live draw paths (`fill_path`, `stroke_path`) pass the active scissor clip. 
This caused recorded geometry inside a `push_clip_path` scope to render unclipped on replay. 

This PR fixes that by adding `current_clip_path()` to the `Dispatcher` trait and querying it per-command in the strip generation loop. A `recording_inside_clip` regression test covers both `FillRect` and `FillPath` commands inside a clip scope, verifying geometry is clipped to the scissor rect.

The bug was introduced in #1203.

I think a deeper issue is that clip paths are not part of the recording, but I'm not sure they should be, so I'm keeping this PR minimal


sparse_strips/vello_sparse_tests/tests/recording.rs
```diff
@@ -205,3 +205,50 @@ fn recording_handles_completely_offscreen_content(ctx: &mut impl Renderer) {
     ctx.prepare_recording(&mut recording);
     ctx.execute_recording(&recording);
 }
+
+// Regression: `Scene::generate_strips_from_commands` (used by
```
LaurenzV: Can we remove the comments here and in the method? I think the title should already be enough to explain what's going on.

LaurenzV (approved):

LaurenzV:
Thanks!

PR #1570, opened 2026-04-13
PoC: Remove scene constraints

LaurenzV:
This is an AI-generated PoC, not meant for code review yet, just to discuss the high-level idea. If we decide to pursue this, I'll make sure to clean this up properly and also add more tests.

This PR suggests an idea for how we could get rid of scene constraints, while _almost_ getting the best of both worlds. The key idea here is to keep a second `Wide` struct around. By default if no layers are involved, we just keep accumulating fast strip paths, as was the case before. If there are layers but they only use default blending, as before we switch into "interleaved" mode, where we interleave fast path strips and coarse batches. Finally, if there _is_ at least one layer where we blend into the root layer using a non-default blend mode, we iterate over all batches and coarse rasterize them into the second wide tile. For fast path strips, we can just do the same as before. For coarse batches, we replay all commands in each wide tile into the new `Wide`. Then, we simply swap the two wide tiles and from now on just proceed in `CoarseOnly` mode.

There is one downside: For each coarse batch, we need to iterate over all wide tiles, even if some wide tiles might not have commands in that specific batch. This means that the lazy pushing mechanism essentially won't pay off for all coarse batches up until the first blending operation. However, the upside is that it simplifies our testing infrastructure again, because we don't have to test two separate modes anymore. It also benefits user who don't know ahead of time if there is blending, but in the vast majority of the cases don't use it. I think that it is probably worth it, especially because many workloads do not require blending. Alternatively, we could also try to track the bounding box for each batch and use that, but it might complicate the code a bit more.

Curious to hear what you think @taj-p.


taj-p (commented):
I think this implementation works. It's really quite ingenious. Kudos!

I think I want to discuss with you, Raph, and Alex the direction of this, however. 

The uber shader is only growing (see https://github.com/linebender/vello/pull/1552). We know that uber shaders get in the way of low tier device performance. 

So, if we arrive at a decision of "we want to move away from an uber shader", then the solution here might be to first accumulate all the draw commands, then run them through the pipeline using the best strategy - rather than, what is proposed here, to unwind the command stack and replay them through a second Wide.

So, if we're sticking with the uber shader and immediate evaluation of draw commands: This makes sense.

If we want to move away from uber shader, then it might be worth moving in that direction rather than making it harder to revert from.

To be clear: The uber shader, as is proposed by Raph, may take the form of a "complex" and "simple" shader or it might become a small set of shaders like was described by Thomas in OH.

LaurenzV:
Was just a PoC, I don't think we will land this in this form.

PR #1571, opened 2026-04-13
Tajp/suggestions2

taj-p:


LaurenzV:
Applied all suggestions, thanks!

PR #1572, opened 2026-04-14
Clear hybrid render targets before drawing

waywardmonkeys:
This clears the hybrid WGPU render target before normal scene rendering.

The previous behavior assumed that the target contents were either undefined or irrelevant, which is often true for fresh swapchain images. But `Renderer::render` accepts an arbitrary `TextureView`, and in practice it is also used with reused offscreen targets. In that case, transparent or otherwise untouched regions can retain pixels from an earlier frame.

This PR does two things:

* adds a regression test that renders a full frame, resets the scene, then renders a smaller second frame into the same target
* clears the user-surface render path before drawing so the second frame starts from transparent black

The atlas render path is left unchanged. This fix is only for the normal user-facing render target path.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -278,9 +278,6 @@ impl Renderer {
             &mut encoded_paints,
         )?;
 
-        // TODO: Passing `false` here because wgpu swapchain textures likely have
-        // undefined initial content, making an explicit clear redundant in the common
-        // case. Verify whether there are scenarios where wgpu would need a clear.
         let result = self.render_scene(
             scene,
             device,
@@ -290,7 +287,7 @@ impl Renderer {
             view,
             &resources.image_cache,
             &encoded_paints,
-            false,
+            true,
             RootRenderTarget::UserSurface,
         );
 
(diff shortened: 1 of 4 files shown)


nicoburns:
I've had bug reports for rendering "undefined buffer content" in Blitz (which in practice happened to be a stale render from another unrelated window!). But don't we want to allow for explicitly not clearing so that we can so incremental renders?

waywardmonkeys:
WebGL clears in this case. This leads to wrong output when using it with compositor layers now.

LaurenzV:
Gonna defer to Taj here.

taj-p (approved):
This makes the WGPU implementation consistent with WebGL, which seems like a worthwhile thing to do! 

Ultimately, I think it should be user configured (either a Load/Clear/DontCare enum) - we shouldn't be deciding whether a surface is cleared or not because, as Nico suggests, you could be building an incremental renderer (and it's expensive on low tier devices!).

waywardmonkeys:
> Ultimately, I think it should be user configured (either a Load/Clear/DontCare enum) - we shouldn't be deciding whether a surface is cleared or not because, as Nico suggests, you could be building an incremental renderer (and it's expensive on low tier devices!).

yep. There is a lot of API work to do across Hybrid, CPU, Classic, and Imaging to have a coherent and possibly consistent story.

PR #1574, opened 2026-04-14
vello_hybrid: Properly clean up unused resources

LaurenzV:
Right now, an instance of a renderer creates a number of resources (vertex arrays, cached frame buffers), but they are never actually destroyed when the renderer is destroyed. This is usually not a huge problem because the browser garbage collector is likely to pick them up at some point, but it's still bad practice and there is no guarantee was to when this happens. I noticed this when trying `vello_bench2` on my tablet, where I wanted to create a new fresh renderer for each benchmark, but at some point the browser would just crash. It is also a [best practice to do so](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#delete_objects_eagerly).

This PR proposes to create thin wrapper types around WebGL resources, where we add a `Drop` implementation that destroys them in WebGL. I admittedly haven't tested on my tablet whether this fixes it (because I switched vello_bench2 to a different approach), but conceptually this should work. 🤔 And it doesn't break the existing `native_webgl` example.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -45,8 +45,10 @@ use alloc::vec;
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
 use core::fmt::Debug;
+use core::ops::Deref;
 #[cfg(feature = "text")]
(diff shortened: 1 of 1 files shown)


taj-p (approved):
LGTM!!

LaurenzV:
Will open a new PR for this.

PR #1576, opened 2026-04-14
sparse_strips: Apply stroke adjustment

LaurenzV:
If we have a glyph run with font size 8 and a scaling factor of 8, we instead set the font size to 64 and the scale to 1. However, the stroke width should _still_ be affected by the scaling factor, even if we absorb. Unfortunately, this is not straightforward to do, because at the time where we do absorption, we don't know yet whether we are going to fill or stroke. So this is the only hack I could come up with.

There is one uncertainty: For glyph caching, we only do absorption if there is no skewing and uniform scaling, so this is always safe to do. However, for hinting we also can absorb if there is vertical skewing; I'm not sure if uniformly increasing the stroke width is still the right thing to do here? But anyway, this is hopefully an edge case we can ignore for now.

### DIFF
--- a/glifo/src/glyph.rs
+++ b/glifo/src/glyph.rs
@@ -498,6 +498,18 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone> GlyphRunRenderer<'a, 'b, Gl
         }
     }
 
+    /// Return the scaling factor that should be applied to the stroke width when stroking this
+    /// glyph run.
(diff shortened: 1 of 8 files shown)


taj-p (approved):
> Unfortunately, this is not straightforward to do, because at the time where we do absorption, we don't know yet whether we are going to fill or stroke. So this is the only hack I could come up with.

Seems like a deficiency in the API.

LGTM for now but this certainly provides important context for how this API needs to evolve.

PR #1577, opened 2026-04-14
[Vello Hybrid] Opaque Strip Pass

taj-p:
Follows from [#vello > Early z culling in hybrid via front to back rendering @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Early.20z.20culling.20in.20hybrid.20via.20front.20to.20back.20rendering/near/582436173).

## Intent

Introduces a "front to back" render pass of opaque strips before a "back to front" pass of strips with alpha.

On low tier devices, overdraw (that is, running the fragment shader over a pixel unnecessarily (because it will be later overwritten with an opaque fragment) is particularly hurtful to performance.

Our current pipeline already splits out opaque strips from alpha strips. This PR merely schedules the opaque strips in a separate front-to-back render pass with depth test and write enabled (with blending disabled). Then, the alpha strips are rendered in a back-to-front render pass with depth test enabled (depth write disabled and blending enabled). Or...

1. Front to back opaque strip pass: Depth write ON, depth test ON, blending OFF
2. Back to front alpha strip pass: Depth write OFF, depth test ON, blending ON

## Results

The performance as measured on `vello_bench2` is found here: [#vello > Early z culling in hybrid via front to back rendering @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Early.20z.20culling.20in.20hybrid.20via.20front.20to.20back.20rendering/near/582436173).

When testing in a "real" application, I see roughly 30% improved performance for scenes with large opaque backgrounds (fairly common in the web and design world).

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -150,8 +150,9 @@ pub struct GpuStrip {
     pub payload: u32,
     /// See `StripInstance::paint_and_rect_flag` documentation in `render_strips.wgsl`.
     pub paint_and_rect_flag: u32,
-    /// Painter's-order index used to compute z-depth for early-z rejection on TBDR GPUs.
-    /// Monotonically increasing in back-to-front order within a frame.
+    /// Painter's-order index used to compute z-depth for early-z rejection in shader.
+    /// In other words, the back-most layer has index 0 and every additional layer in front 
+    /// has an incrementing index.
     pub layer_index: u32,
 }
 
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -150,6 +150,9 @@ pub struct GpuStrip {
     pub payload: u32,
     /// See `StripInstance::paint_and_rect_flag` documentation in `render_strips.wgsl`.
     pub paint_and_rect_flag: u32,
+    /// Painter's-order index used to compute z-depth for early-z rejection on TBDR GPUs.
+    /// Monotonically increasing in back-to-front order within a frame.
+    pub layer_index: u32,
 }
 
 /// Different types of GPU encoded paints.
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -151,9 +151,9 @@ pub struct GpuStrip {
     /// See `StripInstance::paint_and_rect_flag` documentation in `render_strips.wgsl`.
     pub paint_and_rect_flag: u32,
     /// Painter's-order index used to compute z-depth for early-z rejection in shader.
-    /// In other words, the back-most layer has index 0 and every additional layer in front
+    /// In other words, the back-most draw has index 0 and every additional draw in front
     /// has an incrementing index.
-    pub layer_index: u32,
+    pub depth_index: u32,
 }
 
 /// Different types of GPU encoded paints.
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -151,7 +151,7 @@ pub struct GpuStrip {
     /// See `StripInstance::paint_and_rect_flag` documentation in `render_strips.wgsl`.
     pub paint_and_rect_flag: u32,
     /// Painter's-order index used to compute z-depth for early-z rejection in shader.
-    /// In other words, the back-most layer has index 0 and every additional layer in front 
+    /// In other words, the back-most layer has index 0 and every additional layer in front
     /// has an incrementing index.
     pub layer_index: u32,
 }
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1995,9 +1995,7 @@ fn create_webgl_resources(
         slot_framebuffers,
         view_framebuffer_override: None,
         depth_cleared_this_frame: false,
-        depth_attachment_array: js_sys::Array::of1(
-            &WebGl2RenderingContext::DEPTH_ATTACHMENT.into(),
-        ),
+        depth_attachment_array: js_sys::Array::of1(&WebGl2RenderingContext::DEPTH.into()),
         max_texture_dimension_2d,
         stub_atlas_texture_array,
         atlas_render_framebuffer: None,
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1421,10 +1421,10 @@ impl WebGlPrograms {
     fn upload_strip_pair(
         &mut self,
         gl: &WebGl2RenderingContext,
-        first: &[GpuStrip],
-        second: &[GpuStrip],
+        opaque_strips: &[GpuStrip],
+        alpha_strips: &[GpuStrip],
     ) {
-        if first.is_empty() && second.is_empty() {
+        if opaque_strips.is_empty() && alpha_strips.is_empty() {
             return;
         }
 
@@ -1433,9 +1433,9 @@ impl WebGlPrograms {
             Some(&self.resources.strips_buffer),
         );
 
-        let first_bytes: &[u8] = bytemuck::cast_slice(first);
-        let second_bytes: &[u8] = bytemuck::cast_slice(second);
-        let total_len = first_bytes.len() + second_bytes.len();
+        let opaque_bytes: &[u8] = bytemuck::cast_slice(opaque_strips);
+        let alpha_bytes: &[u8] = bytemuck::cast_slice(alpha_strips);
+        let total_len = opaque_bytes.len() + alpha_bytes.len();
 
         // Allocate buffer, then write both slices via bufferSubData.
         // We don't want to pay for concatenating the two slices. It's better to
@@ -1445,18 +1445,18 @@ impl WebGlPrograms {
             total_len as i32,
             WebGl2RenderingContext::DYNAMIC_DRAW,
         );
-        if !first_bytes.is_empty() {
+        if !opaque_bytes.is_empty() {
             gl.buffer_sub_data_with_i32_and_u8_array(
                 WebGl2RenderingContext::ARRAY_BUFFER,
                 0,
-                first_bytes,
+                opaque_bytes,
             );
         }
-        if !second_bytes.is_empty() {
+        if !alpha_bytes.is_empty() {
             gl.buffer_sub_data_with_i32_and_u8_array(
                 WebGl2RenderingContext::ARRAY_BUFFER,
-                first_bytes.len() as i32,
-                second_bytes,
+                opaque_bytes.len() as i32,
+                alpha_bytes,
             );
         }
     }
@@ -2137,7 +2137,7 @@ fn initialize_clear_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources)
 
 /// Context for WebGL rendering operations.
 // TODO: Improve buffer management. Currently a single buffer is used per resource, which means that
-// the GPU must finish drawing before the next `upload_strips` can be executed (effectively pausing
+// the GPU must finish drawing before the next `upload_strip_pair` can be executed (effectively pausing
 // execution). Investigate a buffer pool or creating a new buffer per pass.
 struct WebGlRendererContext<'a> {
     programs: &'a mut WebGlPrograms,
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1378,9 +1378,15 @@ impl WebGlPrograms {
         gl.clear(WebGl2RenderingContext::COLOR_BUFFER_BIT);
     }
 
-    /// Upload strip data to GPU.
-    fn upload_strips(&mut self, gl: &WebGl2RenderingContext, strips: &[GpuStrip]) {
-        if strips.is_empty() {
+    /// Upload two strip slices (opaque then alpha) into a single GPU buffer,
+    /// avoiding multiple `bufferData` calls.
+    fn upload_strip_pair(
+        &mut self,
+        gl: &WebGl2RenderingContext,
+        first: &[GpuStrip],
+        second: &[GpuStrip],
+    ) {
+        if first.is_empty() && second.is_empty() {
             return;
         }
 
@@ -1388,12 +1394,32 @@ impl WebGlPrograms {
             WebGl2RenderingContext::ARRAY_BUFFER,
             Some(&self.resources.strips_buffer),
         );
-        let strips_data = bytemuck::cast_slice(strips);
-        gl.buffer_data_with_u8_array(
+
+        let first_bytes: &[u8] = bytemuck::cast_slice(first);
+        let second_bytes: &[u8] = bytemuck::cast_slice(second);
+        let total_len = first_bytes.len() + second_bytes.len();
+
+        // Allocate buffer, then write both slices via bufferSubData to avoid
+        // a temporary concatenation Vec.
+        gl.buffer_data_with_i32(
             WebGl2RenderingContext::ARRAY_BUFFER,
-            strips_data,
+            total_len as i32,
             WebGl2RenderingContext::DYNAMIC_DRAW,
         );
+        if !first_bytes.is_empty() {
+            gl.buffer_sub_data_with_i32_and_u8_array(
+                WebGl2RenderingContext::ARRAY_BUFFER,
+                0,
+                first_bytes,
+            );
+        }
+        if !second_bytes.is_empty() {
+            gl.buffer_sub_data_with_i32_and_u8_array(
+                WebGl2RenderingContext::ARRAY_BUFFER,
+                first_bytes.len() as i32,
+                second_bytes,
+            );
+        }
     }
 }
 
@@ -2012,6 +2038,11 @@ fn create_framebuffer_for_texture(
     framebuffer
 }
 
+const STRIP_STRIDE: i32 = size_of::<GpuStrip>() as i32;
+const STRIP_ATTR_COUNT: i32 = STRIP_STRIDE / 4;
+const _: () = assert!(STRIP_STRIDE == 24, "expected stride of 24");
+const _: () = assert!(STRIP_ATTR_COUNT == 6);
+
 /// Initialize strip VAO.
 fn initialize_strip_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources) {
     gl.bind_vertex_array(Some(&resources.strip_vao));
@@ -2020,12 +2051,7 @@ fn initialize_strip_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources)
         Some(&resources.strips_buffer),
     );
 
-    const STRIDE: i32 = size_of::<GpuStrip>() as i32;
-    const { assert!(STRIDE == 24, "expected stride of 24") };
-    let stride = STRIDE;
-
-    // Configure attributes.
-    for i in 0..6 {
+    for i in 0..STRIP_ATTR_COUNT {
         let location = i as u32;
         let offset = i * 4;
 
@@ -2034,7 +2060,7 @@ fn initialize_strip_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources)
             location,
             1,
             WebGl2RenderingContext::UNSIGNED_INT,
-            stride,
+            STRIP_STRIDE,
             offset,
         );
 
@@ -2291,6 +2317,13 @@ impl WebGlRendererContext<'_> {
         let is_final_view =
             matches!(target, StripPassRenderTarget::Output(OutputTarget::FinalView));
 
+        // Single upload for all strip data (opaque then alpha) to avoid
+        // multiple bufferData calls which are more expensive than attribute rebinding.
+        self.programs
+            .upload_strip_pair(self.gl, opaque_strips, alpha_strips);
+        let opaque_count = opaque_strips.len() as i32;
+        let alpha_count = alpha_strips.len() as i32;
+
         if is_final_view {
             // Clear depth buffer on first use per frame.
             if !self.programs.resources.depth_cleared_this_frame {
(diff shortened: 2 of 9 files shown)


sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1405,11 +1484,11 @@ impl Scheduler {
             GpuStripBuilder::at_slot(slot_idx, cmd.x, cmd.width)
         };
 
-        draw.push(
-            gpu_strip_builder
-                .with_sparse(cmd.width, col_idx)
-                .paint(payload, paint),
-        );
+        draw.push_alpha(gpu_strip_builder.with_sparse(cmd.width, col_idx).paint(
```
taj-p, line 1509: Question for reviewers: should the caller decide whether to push opaque/alpha or should we pass sufficient information to `draw` for it to make the call?

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -150,6 +150,10 @@ pub struct GpuStrip {
     pub payload: u32,
     /// See `StripInstance::paint_and_rect_flag` documentation in `render_strips.wgsl`.
     pub paint_and_rect_flag: u32,
+    /// Painter's-order index used to compute z-depth for early-z rejection in shader.
+    /// In other words, the back-most layer has index 0 and every additional layer in front
+    /// has an incrementing index.
+    pub layer_index: u32,
```
LaurenzV: Initially I thought we could just use `col_idx` for that, but I guess that won't work for the fast-path rectangles. :(

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -306,9 +308,10 @@ fn vs_main(
     let col_offset = select(f32(instance.col_idx_or_rect_frac), 0.0, is_rect);
     out.tex_coord = vec2<f32>(col_offset + x * f32(width), y * f32(height));
 
+    let z = 1.0 - f32(instance.layer_index) / f32(1u << 24u);
```
LaurenzV: Could dividing by such a large number cause precision issues? Would it make sense to transmit the max layer index in the config uniform instead? Although I'm not sure what the performance implications of that would be.

LaurenzV: Ah, there are more than 2^23 floating point numbers between 0.0 and 1.0, so it should be fine? I'm wondering whether it's possible that precision issues on some GPUs could lead to subtle bugs here though...

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -502,6 +536,7 @@ impl Scheduler {
         filter_context: &FilterContext,
         encoded_paints: &[EncodedPaint],
     ) -> Result<(), RenderError> {
+        self.layer_counter = 0;
```
LaurenzV: We probably should have a more principled way for resetting the scheduler... But not in this PR!

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1770,14 +1899,19 @@ fn generate_gpu_strips_for_fast_path(
         let x0 = strip.x;
         let y = strip.y;
 
+        // Since the components of a single strip are not overlapping, we can re-use the same
+        // layer index for all the components (alpha + fill) for a single strip.
+        let layer_index = *layer_counter;
+        *layer_counter += 1;
```
LaurenzV: I'm not sure I understand this, isn't this increasing the layer counter for each strip + sparse fill instead of once for all strips in the path? In general, I'm not sure I understand why we need to pass a mutable reference to `layer_counter`to `generate_gpu_strips_for_fast_path`, can we not do the same as for rectangle commands in `push_direct_strips`, where we increase the layer counter once and use it for all strips of the path?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -906,7 +954,21 @@ impl Scheduler {
...
             }
 
-            renderer.render_strips(&draw.0, target, load);
+            if i == 2 {
+                // Output target: reverse opaque for front-to-back rendering.
+                // Alpha strips stay in natural back-to-front order.
+                let mut opaque = core::mem::take(&mut draw.opaque);
+                opaque.reverse();
+                renderer.render_strips(&opaque, &draw.alpha, target, load);
+                draw.opaque = opaque; // Return allocation for later reuse.
```
LaurenzV:
```suggestion
                let Draw {opaque, alpha} = draw;
                opaque.reverse();
                renderer.render_strips(&opaque, &alpha, target, load);
```

LaurenzV: Maybe add a comment that this will leave the opaque strips in a "dirty" state, but it doesn't matter since we don't use it again anyway?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1207,10 +1282,11 @@ impl Scheduler {
             if let TemporarySlot::Invalid(temp_slot) = tos.temporary_slot {
                 let next_round = depth.is_multiple_of(2);
                 let el_round = tos.round + usize::from(next_round);
+                let layer_index = self.next_layer_index();
```
LaurenzV: This means we allocate a new layer index for every single wide tile, right? Would it maybe instead be possible to allocate a single layer index for the whole filter layer / coarse batch and then use that for all wide tiles? I think this should work, because everything that copies from a slot uses the `push_alpha` anyway, right? 🤔

sparse_strips/vello_hybrid/src/schedule.rs
LaurenzV: I think for now being more explicit and letting the caller handle it would be the best.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1379,6 +1457,7 @@ impl Scheduler {
         attrs: &CommandAttrs,
     ) {
         let depth = state.tile_state.stack.len();
+        let layer_index = self.next_layer_index();
```
LaurenzV: Same comment as before, I don't think every alpha fill command should have to alocate a new layer index, unless I'm missing something.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1461,7 +1550,32 @@ impl Scheduler {
...
+            Paint::Indexed(indexed_paint) => {
+                let paint_id = indexed_paint.index();
+                match encoded_paints.get(paint_id) {
+                    Some(EncodedPaint::Image(img)) => {
+                        !img.may_have_opacities
+                            && img.sampler.alpha == 1.0
+                            && img.tint.is_none_or(|t| t.color.components[3] >= 1.0)
+                    }
+                    Some(EncodedPaint::Gradient(g)) => !g.may_have_opacities,
+                    _ => false,
```
LaurenzV: Can we spell out blurred rectangles explicitly?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1498,14 +1612,16 @@ impl Scheduler {
         // enough "ping-ponging" to resolve all dependencies.
         let next_round = depth.is_multiple_of(2) && depth > 2;
         let round = nos.round.max(tos.round + usize::from(next_round));
+        let layer_index = self.next_layer_index();
```
LaurenzV: Same as before.

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1498,14 +1612,16 @@ impl Scheduler {
...
             let draw = self.draw_mut(round, nos.dest_slot.get_texture());
-            draw.push(
+            draw.push_alpha(
                 GpuStripBuilder::at_slot(nos.dest_slot.get_idx(), 0, WideTile::WIDTH)
-                    .copy_from_slot(temp_slot.get_idx(), 0xFF),
+                    .copy_from_slot(temp_slot.get_idx(), 0xFF, layer_index),
             );
         }
 
+        let layer_index = self.next_layer_index();
```
LaurenzV: Same here.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV, line 1: Getting a lot of these errors when trying it out:

Image

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1375,22 +1417,48 @@ impl WebGlPrograms {
...
 
-    /// Upload strip data to GPU.
-    fn upload_strips(&mut self, gl: &WebGl2RenderingContext, strips: &[GpuStrip]) {
-        if strips.is_empty() {
+    /// Uploads two strip slices (opaque then alpha) into a single GPU buffer.
+    fn upload_strip_pair(
+        &mut self,
+        gl: &WebGl2RenderingContext,
+        first: &[GpuStrip],
+        second: &[GpuStrip],
```
LaurenzV: Maybe we can just preserve the opaque/alpha names here?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1926,6 +1994,10 @@ fn create_webgl_resources(
         slot_textures,
         slot_framebuffers,
         view_framebuffer_override: None,
+        depth_cleared_this_frame: false,
+        depth_attachment_array: js_sys::Array::of1(
+            &WebGl2RenderingContext::DEPTH_ATTACHMENT.into(),
```
LaurenzV: It seems like we need to pass DEPTH instead of DEPTH_ATTACHMENT here to fix above error? https://docs.gl/es3/glInvalidateFramebuffer

Not sure why this doesn't seem to be mentioned in the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/invalidateFramebuffer) though..

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -1199,6 +1216,82 @@ impl Programs {
...
+                        format: strip_formats[i],
+                        blend: Some(BlendState::PREMULTIPLIED_ALPHA_BLENDING),
+                        write_mask: ColorWrites::ALL,
+                    })],
+                    compilation_options: PipelineCompilationOptions::default(),
+                }),
+                primitive: wgpu::PrimitiveState {
+                    topology: wgpu::PrimitiveTopology::TriangleStrip,
+                    ..Default::default()
+                },
```
LaurenzV: Can we deduplicate those as well?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -2299,19 +2429,23 @@ impl Programs {
         }
     }
 
-    /// Upload the strip data by creating and assigning a new `self.resources.strips_buffer`.
-    fn upload_strips(&mut self, device: &Device, queue: &Queue, strips: &[GpuStrip]) {
-        let required_strips_size = size_of_val(strips) as u64;
-        self.resources.strips_buffer = Self::create_strips_buffer(device, required_strips_size);
-        // TODO: Consider using a staging belt to avoid an extra staging buffer allocation.
```
LaurenzV, line 2303: TODO is still relevant I think?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -2299,19 +2429,23 @@ impl Programs {
...
+    /// Uploads two strip slices (opaque then alpha) into a single GPU buffer.
+    fn upload_strip_pair(
+        &mut self,
+        device: &Device,
+        queue: &Queue,
+        first: &[GpuStrip],
+        second: &[GpuStrip],
```
LaurenzV: Same here, I think having explicit names would be ebtter.

LaurenzV (commented):
Nice!! I would like to take another look at the scheduling logic to think about the interaction between layer indices and layers more deeply, but overall looks good to me!

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: SGTM!

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
taj-p: > Ah, there are more than 2^23 floating point numbers between 0.0 and 1.0, so it should be fine?

Yup! The WebGL2 shader uses highp precision here. I would have concerns about using the max layer index because that might be a non-power of 2, resulting in rounding. The division by a power of 2 is an exponent shift only - there should be zero rounding for this line for any layer index between 0 and 2^24. 

I think we want to preserve a form here where we know `z` will always be exactly representable given any `layer_index`.

Added comment in https://github.com/linebender/vello/pull/1577/commits/e72979464d0200c7ac751dd3edef7e7c205b808e

sparse_strips/vello_hybrid/src/render/common.rs
taj-p: That would've been so good!!!

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Yes! Nice catch - fixed in https://github.com/linebender/vello/pull/1577/commits/2f950a5299a368664823417d9814139153392fd0

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Added comment in https://github.com/linebender/vello/pull/1577/commits/e8fe0c83874fdba7ff8a20a7c3857adbdcabe5d2 - good idea 🙏

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: I feel like I've introduced confusion here by referring to this as layer. It is not the same concept as a coarse batch or filter layer as is used in other parts of the pipeline. I'll rename this to `depth_index`.

This index represents z-ordering, so every subsequent draw command from a different path should live in a higher index. Z indices for strips for a single path can be shared (because strips for a path are not overlapping), but we cannot share z indices for two separate strips unless we first check they don't overlap (which I don't think we want to do).

So, it's not possible to share 1 index across a coarse batch (because each command in the batch needs to be drawn on top of the last). 

I assumed that the `2^24` available z indices would give us sufficient room to not worry about any complexity here - that the available room could keep this simple - but I'll have a think about de-duplication strategies. 

I'm going to time box using `attrs_idx` as the z index (and carry any offsets required across interleaved mode).

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: See https://github.com/linebender/vello/pull/1577/changes/BASE..d9f0535e9d2ac4c29fad7780bad30de6ebc993f8#r3089206558

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Thank you for this catch 🙏 . Implemented in https://github.com/linebender/vello/pull/1577/commits/fde9069549daac66fdf0c821b57d605c2a131698

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: Ohhh thank you! What a catch. Fixed in https://github.com/linebender/vello/pull/1577/commits/3b13fee428653f3b6b9844f2e2559e98722a29e3. 

Looks like an FBO requires the `_ATTACHMENT` suffix:

> If the specified framebuffer is a framebuffer object, each element of attachments must be one of GL_DEPTH_ATTACHMENT, GL_STENCIL_ATTACHMENT GL_DEPTH_STENCIL_ATTACHMENT, or GL_COLOR_ATTACHMENTi, where i is between zero and the value of GL_MAX_FRAMEBUFFER_ATTACHMENTS minus one.

> If the specified framebuffer is a default framebuffer, each element of attachments must be one of GL_FRONT_LEFT, GL_FRONT_RIGHT, GL_BACK_LEFT, GL_BACK_RIGHT, GL_AUXi, GL_ACCUM, GL_COLOR, GL_DEPTH, or GL_STENCIL. GL_COLOR, is treated as GL_BACK_LEFT for a double-buffered context and GL_FRONT_LEFT for a single-buffered context. The other attachments identify the corresponding specific buffer. 

Very confusing 😓

sparse_strips/vello_hybrid/src/render/wgpu.rs
taj-p: Yes definitely. Added in https://github.com/linebender/vello/pull/1577/commits/5c29c28365f33b719144072d91d86e959cdc4ed6

sparse_strips/vello_hybrid/src/render/wgpu.rs
taj-p: Agree - fixed in https://github.com/linebender/vello/pull/1577/commits/940e668db500a83850bd9509928544d39977b0c6

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: We came up with a much better idea in DMs: https://github.com/linebender/vello/pull/1577/commits/4ff1645023e55d1422173acde6c843f41d452da9

> So, I think we can make it a lot simpler since only opaque strips write to the depth buffer, so only opaque strips require unique z-values. Alpha strips just test against whatever opaques wrote - they don't need distinct z values (they just need to be greater or equal to the last opaque).

> With this in mind, we can do something like:
> ```rs
> impl DepthCounter {
>   fn next_opaque(...) -> u32 { self.counter += 1; self.counter }
>   fn next_alpha(...) -> u32 { self.counter }
> }
> ```
> Then we use the minimum layers (provided we don't create needless opaque indices when stack depth isn't 1)

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: Agree - fixed in https://github.com/linebender/vello/pull/1577/commits/940e668db500a83850bd9509928544d39977b0c6

sparse_strips/vello_hybrid/src/render/wgpu.rs
taj-p: Deduplicated in https://github.com/linebender/vello/pull/1577/commits/de44c800477263b03b83af3a1fb99343d734af37

taj-p: Great idea - thank you 🙌

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Oops - I referenced the wrong commit above (that's a complicated version). The simpler one is here:

https://github.com/linebender/vello/pull/1577/commits/6bab22b157433eea029555599112775bd0200088

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: Thanks for finding this fix below 🙏

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -760,25 +820,40 @@ impl Scheduler {
         let strip_storage = scene.strip_storage.borrow();
         // Always choose the draw of the final surface, since direct strips are only ever
         // rendered to the final surface.
+        let mut depth = core::mem::take(&mut self.depth);
```
LaurenzV, line 823: Do we really need `mem::take` here?

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1461,7 +1580,33 @@ impl Scheduler {
...
         };
 
-        draw.push(gpu_strip_builder.paint(payload, paint));
+        let strip = gpu_strip_builder.paint(payload, paint, depth_index);
+        if depth == 1 && is_opaque {
+            draw.push_opaque(strip);
+        } else {
+            draw.push_alpha(strip);
+        }
+    }
```
LaurenzV, line 1589: This isn't enough, because the `depth` will also be 1 when rendering a filter layer. See this test, which demonstrates it (would be good if you could add it):

```rs
#[vello_test(skip_multithreaded)]
fn issue_filter_preserves_painter_order_for_opaque_and_alpha(ctx: &mut impl Renderer) {
    let filter = Filter::from_primitive(FilterPrimitive::Offset { dx: 0.0, dy: 0.0 });

    ctx.push_filter_layer(filter);
    ctx.set_paint(AlphaColor::from_rgba8(0, 0, 255, 128));
    ctx.fill_rect(&Rect::new(20.0, 20.0, 80.0, 80.0));
    ctx.set_paint(RED);
    ctx.fill_rect(&Rect::new(30.0, 30.0, 70.0, 70.0));
    ctx.pop_layer();
}
```

We'll likely have to somehow wire information about whether we are in a root or filter layer down there...

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -1461,7 +1580,33 @@ impl Scheduler {
...
+
+    /// Determine if a paint is fully opaque (no transparency).
```
LaurenzV, line 1591:
```suggestion
    /// Determine if a paint is fully opaque.
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1926,6 +1994,8 @@ fn create_webgl_resources(
         slot_textures,
         slot_framebuffers,
         view_framebuffer_override: None,
+        depth_cleared_this_frame: false,
+        depth_attachment_array: js_sys::Array::of1(&WebGl2RenderingContext::DEPTH.into()),
```
LaurenzV, line 1998: Maybe let's leave a comment that in case we want to support this for non-default framebuffers, we need to use DEPTH_ATTACHMENT here? As a reminder for the future.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -2284,14 +2356,92 @@ impl WebGlRendererContext<'_> {
...
+        if is_final_view {
+            // Clear depth buffer on first use per frame.
+            if !self.programs.resources.depth_cleared_this_frame {
+                self.programs.resources.depth_cleared_this_frame = true;
+                self.gl.clear_depth(1.0);
+                self.gl.clear(WebGl2RenderingContext::DEPTH_BUFFER_BIT);
+            }
+
+            self.gl.enable(WebGl2RenderingContext::DEPTH_TEST);
+            self.gl.depth_func(WebGl2RenderingContext::LEQUAL);
```
LaurenzV, line 2381: Does it make sense to move that up? Is clearing guaranteed to work even when depth test is disabled?

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -2329,16 +2414,18 @@ impl RendererContext<'_> {
...
     ) {
-        if strips.is_empty() {
+        if opaque_strips.is_empty() && alpha_strips.is_empty() {
             return;
         }
-        // TODO: We currently allocate a new strips buffer for each render pass. A more efficient
-        // approach would be to re-use buffers or slices of a larger buffer.
-        self.programs.upload_strips(self.device, self.queue, strips);
+        self.programs
+            .upload_strip_pair(self.device, self.queue, opaque_strips, alpha_strips);
```
LaurenzV, line 2426: TODO Still seems relevant?

LaurenzV (commented):
Awesome! 🎉 🎉

LaurenzV (approved):

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: Added note in https://github.com/linebender/vello/pull/1577/commits/e956d675cb8701f061a45da5628029482efb0563

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Great catch! Thank you so much 🙌 !!

Fixed in https://github.com/linebender/vello/pull/1577/commits/eac6cbe3b7b466c635f0bfcd6a9bb50ca03b5498 and added test in https://github.com/linebender/vello/pull/1577/commits/17e234d756b30b68ad22b5134f5d5f5efc807431

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Added in https://github.com/linebender/vello/pull/1577/commits/67580f9432cd38cd9652c0ac0a0eb383dcf335e6

sparse_strips/vello_hybrid/src/render/webgl.rs
taj-p: Definitely. Added in https://github.com/linebender/vello/pull/1577/commits/b9aaceef1ca9968d8c8a173a9e35894eb7307132

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: I tried cloning, splitting the borrow, but they didn't appear as neat / reveal the semantics I wanted compared to this

sparse_strips/vello_hybrid/src/render/wgpu.rs
taj-p: Added back in https://github.com/linebender/vello/pull/1577/commits/4d09457ea3c688cb7b31b840e4b2ccd924a4b8b4

PR #1579, opened 2026-04-15
Add vectorisation to some suspicious blend related missing parts

DJMcNab:
In writing #1578, I stumbled upon these missing instances.

I'm aware of the comment at https://github.com/linebender/vello/blob/6d23bf54894f52ccc72dce1d77c500d8346f53d7/sparse_strips/vello_cpu/src/fine/highp/mod.rs#L339-L342, and so haven't been aggressive as I might otherwise have expected to be.

I probably won't have time to drive this further (i.e. do the benchmarking myself), so feel free to push to this as appropriate.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/highp/blend.rs
+++ b/sparse_strips/vello_cpu/src/fine/highp/blend.rs
@@ -25,6 +25,7 @@ impl<S: Simd> Channels<S> {
 
 // TODO: blending is still extremely slow, investigate whether there is something obvious we are
 // missing that other renderers do.
+#[inline(always)]
 pub(crate) fn mix<S: Simd>(src_c: f32x16<S>, bg: f32x16<S>, blend_mode: BlendMode) -> f32x16<S> {
     if matches!(blend_mode.mix, Mix::Normal) {
         return src_c;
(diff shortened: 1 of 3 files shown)


LaurenzV:
Back then, IIRC I purposefully decided not to inline blending-related methods to reduce potential code bloat, and since I assumed that blending itself is so slow that inlining itself won't give us that much benefit. Definitely open to changing this, but yes we should try to 1) check the impact on binary size and 2) see whether it actually yields any performance improvements.

Though the vectorize calls definitely make sense! I presume especially on x86 this should yield improvements.

DJMcNab:
All of the methods I changed (in theory) wouldn't be optimised properly/at all on x86.
That is, each change needs addressing, but the choice between vectorise and inline(always) is one which needs to be made. I made my choices based on whether the method was already in a vector context.
It would be reasonable to not inline these, and instead insert a vectorise call - i'm happy to make that change, although I still won't have time for benchmarking.

LaurenzV:
Will take a look when I find the time, thanks!

LaurenzV:
Already fixed, we know have plenty of inlines and also `vectorize` calls for blending, which have been confirmed to help a lot on x86.

PR #1580, opened 2026-04-15
vello_hybrid: Let `destroy_image` methods take `Resources` directly instead of `ImageCache`

LaurenzV:
Since `ImageCache` is a private field in `Resources`, clients otherwise have no way of destroying their images!


tomcur (approved):
Good one!

PR #1584, opened 2026-04-17
sparse_strips: Re-enable COLR glyph tests

LaurenzV:
# Overview
This PR attempts to make COLR glyphs work again. To do so, it combines three changes.

## Clip layers
I realize that COLR doesn't actually have a concept of "clip layers"... I'm not sure why we implemented it this way, I think this was before we had the faster non-layer based clipping algorithm. Instead, in COLR there is a separate clipping stack for clip rects / glyphs, and then a layer stack for blending. I've updated the implementation to reflect that, by removing clip layers and using our faster `push_clip_path`.

However, we still need to keep `push_clip_layer` in glifo's imaging model for now, to resolve the second problem mentioned below.

## Destructive blends
Currently, we have the problem that as soon as we have a destructive blend, we iterate over all wide tiles in the _whole_ viewport area and perform blending. As you can imagine, this is prohibitively expensive for COLR glyphs: A glyph only covers a small area, but we end up applying the blend mode to the whole drawing area, which is likely going to be huge, especially when drawing to an atlas.

The fundamental problem is a deficiency in our imaging model: In our current implementation, we assume that a destructive blend always has to apply to the whole drawing area. However, there should be a mechanism to restrict the impact radius of destructive blends. This is also what the long-standing issue #1119 is about.

**This PR does not fix this fundamental issue.** With this PR, if you push a single layer with a clip path _and_ a destructive blend mode on the same layer, the blend mode will still be applied to the whole drawing area.  Instead, what this PR does is if we _first_ push a clip layer and _then_ push another blend layer, the impact radius of the destructive blend will be reduced to the bbox of the previous clip layer. It’s basically a first step towards the end goal, but it's only a start, not the full solution. I hope that we can fully fix this in the future, but it will probably require some deeper changes to coarse rasterization.

In the meanwhile, we ensure that when rendering a COLR glyph, we first push a clip layer that contains the bounding box of the glyph, and then render the glyph as usual, such that our optimization triggers. For now, this is considered a vello-specific hack in glifo and it probably should be removed in the future. But for now this is the easiest way to get COLR glyphs rendering without any extreme performance cliffs.

## Blend optimization
For this change to work, we need to make two passes over the COLR glyph: A first pass to get the actual bounding box (not all COLR glyphs have a clip bbox, and our previous approach of restricting the bbox to upem is wrong, because glyphs are allowed to exceed the rectangle spanned by the upem), and a second pass to do the actual drawing. 

Since we already need this first extra pass, we might as well use this to get some more metadata about the glyph for a second optimization:  Strictly speaking, COLR uses a layer-based rendering model, where the renderer has to push blend layers and merge them together. However, in case there are no default blend modes, we can optimize the whole thing to not use any layers at all. This is possible because for normal src-over compositing, it doesn’t make a difference whether you first render everything into layers and then composite, or just composite everything without using any layers whatsoever, since premultiplied alpha blending is associative.  The vast majority of glyphs do not use any non-default blend modes. Hence, this also brings a small speed boost for the vast majority of glyphs and avoids using coarse rasterization and scheduling for this.

# Remaining problems
Unfortunately, even with those fixes in place, there still are a few issues, mostly performance related. Some of those problems just require some investigation and _presumably_ aren't hard to fix, other's of those unfortunately expose some serious weaknesses/shortcomings in our current system, and I'm not sure what the best way of tackling those are.

## Zoom bug
I'm not sure why, but zooming into the emojis in interactive view seems to result in them getting cut off at the top. It probably should be an easy fix, but needs to be investigated. Happens for both, CPU and Hybrid.

https://github.com/user-attachments/assets/e8c8cf01-aec4-4d19-a938-a85cb6df15d2

## Sluggish performance
While there have been great improvements to the performance of Vello Hybrid, unfortunately it is still clear that for general vector graphics, performance isn't quite there yet, because the CPU-side of things takes up too much time. For COLR glyphs, this starts to become apparent soon once you try displaying multiple dozens of glyphs. Note that these recordings are on my M4 Max with a 1080p screen, so it's probably only gonna get worse on lower-tier devices!

Here is an example of emojis that only use default blending:

https://github.com/user-attachments/assets/ebc15b8f-21b6-4fa1-b3fe-ac1e10523b4e

Looking at the profile yields the following result:
image

6.1% goes into calculating the COLR metrics. However, this should be something we can easily get rid of by just caching it, since it's a very small struct per glyph. Happy to make that change.

However, as you can see everything else is just preparatory CPU work for rendering.

**EDIT**: I just noticed that more than half of the time is spent on `fil_rect`, which presumably doen't use our GPU fast because of clipping. I will look into whether we can use the CPU fast path for rendering rectangles in this case, which should help a lot. It also seems like for push/pop clip glyph we aren't currently caching the outlines, which is probably also worth trying. So maybe it's not as bad as it seems.

Here are some emojis which make more extensive use of blend modes, which seems to be even slower, although from some initial profiling it still doesn't seem to be bottlenecked by the GPU:

https://github.com/user-attachments/assets/34e51a66-9ca0-4da7-87d5-9091d7b3ebc1

## Jiggle when caching is enabled
Some of the performance considerations can partly be mitigated with caching, but there once agains seems to be some jiggling happening:

https://github.com/user-attachments/assets/5c0e1616-51fb-4316-af31-07c4ddf8adad

It might be because we don't du "subpixel" bucketing for cached COLR glyphs, but I haven't investigated this yet.


sparse_strips/vello_sparse_tests/snapshots/glyphs_colr_test_glyphs_cached.png
LaurenzV, line 1: Before:

Image

After:

Image

Skia:

Image

LaurenzV, line 1: Before:

Image

After:

Image

Skia:

Image

sparse_strips/vello_sparse_tests/snapshots/glyphs_colr_test_glyphs.png
LaurenzV, line 1: Same as for cached:

Before:

Image

After:

Image

Skia:

Image

LaurenzV:
Alright, so the current PR descriptiption is a bit outdated now. Here is the current state: 
1) I've opened two PRs (#1585 and #1586), which will dramatically improve the render times. 
2) I've fixed the zoom bug and added a regression test for it.

Once those PRs are merged, I can rebase this PR and I think we can land it more or less in this form (note that it's not ready for review yet, though).

As mentioned, there are **three more things** we can do to further improve performance:
1) Cache the `ColrGlyphInfo` struct.
2) Cull glyphs outside of the viewport more aggressively.
3) Cache the outlines that are constructed as part of building an outline. This might be a bit more difficult to implement in glifo, but should also be doable.

Those should give even better performance boosts, but I will leave those as a follow-up since they are not that critical.

Therefore, after landing all of this, I think the final remaining problem would be figuring out the jitter bug when caching is enabled. But this is probably not super high priority, since we aren't using caching anyway.

cc @taj-p

LaurenzV:
Also, I vibecoded an implementation for 1) and 3). When those two are in, atlas caching makes everything buttery smooth! See the videos below:

1) No caching:

https://github.com/user-attachments/assets/b0a53488-bb98-4b74-8980-35ad3e782846

2) Caching:

https://github.com/user-attachments/assets/56879b12-7312-4c20-bf77-a8e28466bca9

Excited to see all of this landing. 🎉

sparse_strips/vello_hybrid/src/schedule.rs
```diff
@@ -821,33 +821,35 @@ impl Scheduler {
         // Always choose the draw of the final surface, since direct strips are only ever
         // rendered to the final surface.
         let mut depth = core::mem::take(&mut self.depth);
+        // TODO: Also allow the split when rendering to an atlas layer.
```
LaurenzV, line 824: @taj-p Had to apply this fix, otherwise cached glyphs wouldn't render properly. We currently apply the alpha/opaque only when rendering to the root surface, but would still attempt to split it when rendering to an atlas layer.

LaurenzV:
Should be ready for review! I've noted the proposed follow-ups as TODOs in the code.

glifo/src/colr.rs
taj-p, line 44: It's unfortunate we create a new `ColrPainter` for each glyph rather than re-using it - outside of scope of this PR, but maybe something we can tackle another time.

https://github.com/linebender/vello/blob/44703390c1efea65121bb1156a1b557cf2532185/glifo/src/renderer.rs#L232

glifo/src/colr.rs
```diff
@@ -201,16 +288,20 @@ impl ColorPainter for ColrPainter<'_> {
...
 
-        self.painter.push_clip_layer(&transformed);
-        self.layer_count += 1;
+        self.push_clip(&transformed);
     }
 
     fn pop_clip(&mut self) {
-        self.painter.pop_layer();
-        self.layer_count -= 1;
+        if self.pop_stack_entry(ColrStackEntry::ClipPath) {
```
taj-p, line 307: Let's add a debug assert or assert that is only fired in tests here to make sure that this condition only fails for malformed fonts as opposed to a bug in our implementation - i.e., our code should always correctly call `pop_clip` when the top of stack is a clip path for well-formed fonts.

For fonts that specify a clip glyph that doesn't exist in the font, then we expect this to fail, in which case, graceful degradation is fine.

glifo/src/colr.rs
```diff
@@ -153,26 +166,101 @@ impl<'a> ColrPainter<'a> {
...
+            true
+        } else {
+            false
+        }
+    }
+}
+
+pub(crate) struct ColrGlyphInfo {
+    /// A conservative bounding box of the glyph.
+    pub(crate) bbox: Option<Rect>,
```
taj-p, line 199: Is it possible to specify in the documentation what `None` means for `bbox` - that there exists no renderable content?

glifo/src/colr.rs
```diff
@@ -332,52 +426,117 @@ impl ColorPainter for ColrPainter<'_> {
...
+        if !self.skip_blend_layers {
+            self.painter.push_blend_layer(blend_mode);
+            self.stack.push(ColrStackEntry::BlendLayer);
+        }
     }
 
     fn pop_layer(&mut self) {
-        self.painter.pop_layer();
-        self.layer_count -= 1;
+        if !self.skip_blend_layers && self.pop_stack_entry(ColrStackEntry::BlendLayer) {
```
taj-p, line 456: As above, can we add a test assertion that the top of stack is a blend layer when `!self.skip_blend_layers`?

sparse_strips/vello_hybrid/src/schedule.rs
taj-p: Thank you 🙏 ! Nice catch!

sparse_strips/vello_common/src/coarse.rs
```diff
@@ -784,22 +784,31 @@ impl<const MODE: u8> Wide<MODE> {
 
             self.layers_needing_buf_stack.push(layer_kind);
 
-            // We eagerly push buffers for the entire viewport if this is a destructive blend
-            // or if we are (nested in) a clipped filter layer.
+            // We eagerly push buffers for the current active bbox if this is a destructive blend
```
taj-p, line 787: This is a pretty big optimisation for consumers of Vello Hybrid / CPU who use destructive blends generally. I wonder whether it should form part of the public documentation in `scene.rs` or similar - wrap destructive blends in a clip path to avoid unnecessary work.

Ultimately, it's an implementation detail that I'm not sure should be shared with consumers - but, at the same time, it's an optimisation that an external consumer (glifo) is currently relying on.

taj-p (approved):
LGTM!

glifo/src/interface.rs
```diff
@@ -23,12 +23,32 @@ pub trait DrawSink {
     fn fill_path(&mut self, path: &BezPath);
     /// Fill a rectangle with the current paint and transform.
     fn fill_rect(&mut self, rect: &Rect);
+    // Note: `push/pop_clip_layer` are only needed temporarily for Vello renderers so the impact of
+    // destructive blend modes can be limited to a smaller area. Once the limitation is lifted,
+    // we can remove this method and just use `push_clip_path` and `push_blend_layer`.
```
taj-p: Should we suffix this with a `TODO:` instead of a note?

glifo/src/colr.rs
LaurenzV: Will add a TODO.

sparse_strips/vello_common/src/coarse.rs
LaurenzV: Yes, once we rework the clip layer semantics in vello common this can probably become a general optimization.

glifo/src/colr.rs
LaurenzV: Added an assertion in `pop_stack_entry` that triggers only in test mode.

glifo/src/colr.rs
LaurenzV: Added docs.

glifo/src/colr.rs
LaurenzV: Resolved by the same change I made for your other comment.

PR #1585, opened 2026-04-17
vello_common: Optimize fast rectangle path with active clip path

LaurenzV:
Spun off from #1584. For axis aligned rectangles, we can make the rectangle even smaller by intersecting it with the (ceiled) bounding box of the currently active clip path.

This is important for COLR: Conceptually, in COLR we always draw rectangles with the full viewport size, but then clip it to the actual area of the glyph, which is usually very small. By doing so, we save a lot of unnecessary computations.


taj-p (approved):
Makes sense! LGTM

PR #1586, opened 2026-04-17
vello_hybrid: Add support for CPU fast path rectangles

LaurenzV:
While Vello Hybrid already has a fast path for axis-aligned rectangle, that doesn't work when there is a clip path in place. However, in COLR the operation "draw this rectangle with this clip path" is used everywhere. Therefore, we also add the option to fall back to the CPU rectangle fast path for this specific case.

Also adds some more tests for this.


taj-p (approved):
LGTM!

PR #1588, opened 2026-04-20
vello_hybrid: Streamline error handling in the WebGL backend

LaurenzV:
Right now, we have a relatively inconsistent approach of error handling, returning errors in some cases but just unwrapping in other cases. This PR attempts to improve the situation by adding more robust error handling to the WebGL backend. Something similar should be done in the wgpu backend, but that is left for a future PR.

In a few cases, it was often not straightforward to decide whether something should become an error or be unwrapped. I've tried to follow the principle of unwrapping in all cases where it _should_ be impossible to fail (i.e. `max_texture_dimension` returning a non-numeric value), while using errors anywhere else where it could reasonably fail. Alternatively, we could add a final `Unknown(String)` error variant where we combine all errors that should in theory never happen, and thus ensure that the backend never crashes, even if something really unexpected happens.

### DIFF
--- a/sparse_strips/vello_hybrid/src/lib.rs
+++ b/sparse_strips/vello_hybrid/src/lib.rs
@@ -62,7 +62,10 @@ pub mod util;
 pub use render::{AtlasWriter, RenderTargetConfig, Renderer};
 pub use render::{Config, GpuStrip, RenderSize};
 #[cfg(all(target_arch = "wasm32", feature = "webgl"))]
-pub use render::{WebGlAtlasWriter, WebGlRenderer, WebGlTextureWithDimensions};
+pub use render::{
+    WebGlAtlasWriter, WebGlError, WebGlInitError, WebGlRenderer, WebGlResourceType,
+    WebGlTextureWithDimensions,
+};
 pub use resources::Resources;
 pub use scene::{RenderSettings, Scene, SceneConstraints};
 #[cfg(feature = "text")]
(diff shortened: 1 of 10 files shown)


sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -84,12 +87,13 @@ pub enum RenderError {
...
-    /// An allocation error occurred while trying to allocate a new image. This can happen
-    /// if the scene contains filter layers, which need space in the image atlas for intermediate
-    /// storage.
-    #[error("Filter atlas allocation failed: {0}")]
+    /// An allocation in an image atlas failed.
+    #[error("Atlas allocation failed: {0}")]
     AtlasError(#[from] vello_common::multi_atlas::AtlasError),
-    // TODO: Consider expanding `RenderError` to replace some `.unwrap` and `.expect`.
+    /// A WebGL operation failed during rendering.
+    #[cfg(all(target_arch = "wasm32", feature = "webgl"))]
```
LaurenzV, line 94: Not sure if it's a good idea to feature gate an error variant, open for other suggestions! Maybe WebGL errors should just always be available? Or I guess there could be a generic error type where WebGL and wgpu can insert their own internal error type.

LaurenzV:
Hmm, I think if we want to have something like this we need to think a bit more about what a user is supposed to do in case they encounter an error and document that. I think ideally, once any kind of error is encountered the user is expected to not do anything with the renderer anymore and completely discard it. However, if we switch to this result-based API, users might be inclined to just try reusing the context, which is bad because at that point the whole renderer might be in an inconsistent state (for example, pending bitmap uploads will be completely drained even if not all bitmap glyphs have been uploaded.

I think just documenting that "error encountered = don't perform any further operations with the renderer" might be enough, but perhaps this warrants some more discussion.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -113,41 +243,49 @@ pub struct WebGlRenderer {
...
         #[cfg(debug_assertions)]
         {
             // If a WebGL context already exists on this canvas, it will be returned instead of
             // creating a new one with the correct context_options set.
             // See this comment for why we still care about non-antialiased context:
             // https://github.com/linebender/vello/pull/1546/changes#r3008692535
-            let context_attributes = gl.get_context_attributes().unwrap();
+            let context_attributes = gl
+                .get_context_attributes()
+                .expect("WebGL context attributes should be available");
```
LaurenzV, line 280: Note to self: This can probably fail if the WebGL context has been lost.

taj-p:
> Hmm, I think if we want to have something like this we need to think a bit more about what a user is supposed to do in case they encounter an error and document that. I think ideally, once any kind of error is encountered the user is expected to not do anything with the renderer anymore and completely discard it. However, if we switch to this result-based API, users might be inclined to just try reusing the context, which is bad because at that point the whole renderer might be in an inconsistent state (for example, pending bitmap uploads will be completely drained even if not all bitmap glyphs have been uploaded.
> 
> I think just documenting that "error encountered = don't perform any further operations with the renderer" might be enough, but perhaps this warrants some more discussion.

We could poison the renderer similar to how a mutex is poisoned, but that doesn't solve the resources problem 🤔 . I'll raise it in the OH tomorrow

taj-p (commented):
Re-request review once you're ready 🙏

LaurenzV:
Outdated.

PR #1592, opened 2026-04-21
[Vello Hybrid + CPU]: Expose `render_decoration` from Glifo

taj-p:
Mostly Claude Opus 4.6 implemented with some tweaking of tests - simply exposes the `render_decoration` method of `glifo`.


sparse_strips/vello_sparse_tests/snapshots/glyphs_decoration_transformed_cached.png
taj-p, line 1: This looks like a bug. I don't have time left today to debug this. I can investigate tomorrow unless you get to it today

laurenz-canva: I think this should be correct? This is coming from the first glyph run:
image

Since you scale it up by 20, it makes sense that the underline (as well as the gap) become larger as well. Or am I missing something?

laurenz-canva: > // Run-level scale absorbs into font_size

For the purpose of glyph caching it should absorb into font size, yes. However, the underline should just be scaled by 20x, otherwise the size of the underline would not change when zooming in!

LaurenzV (approved):
I disabled hinting completely and made the tests a bit smaller, let me know if that works for you!

Assuming that the glyphs decoration behavior is intended (which I believe it is), it would maybe be good to adjust the test so it doesn't look that ugly. Otherwise LGTM!

sparse_strips/vello_sparse_tests/snapshots/glyphs_decoration_transformed_cached.png
taj-p: Ah yes! Thank you for this. It is correct.

taj-p: Made it less ugly in https://github.com/linebender/vello/pull/1592/commits/28685ca18cffe4fe05aee690afe09e5fe83b0956

PR #1596, opened 2026-04-21
vello_hybrid: Add the `probe` method to the WebGL backend

LaurenzV:
# Problem

Due to the diverse nature of devices, there are many different WebGL drivers out there, some of which can be buggy in certain ways or have certain constraints that are not present in other computers. Because of this, just because our WebGL tests pass in CI, it does not automatically mean that it will run on every other devices.

From our experiments, in most cases the, browser will complain if there is some issue with compiling, linking, or running our shader on a certain device, allowing us to recognize this case and handle it appropriately. However, in other cases it can instead happen that there is a silent failure: There is no explicit error, but the canvas will just stay blank regardless of what was rendered, or be glitchy. This is the worst case, because there isn't really any way of detecting issue automatically.

# Solution
To counteract this problem, this PR proposes to add a new `probe` method to the WebGL backend. As part of this, we start a very small rendering operation where we try to exercise different parts of the rendering pipeline (simple shapes, alpha blending, filters, etc.). After that, we read the pixels back to the CPU and then compare it against a pre-generated gold image that was generated with Vello CPU. We do a pixel-by-pixel comparison (with a threshold of +- 3) and return an error in case the test failed.

All of this is hidden behind feature flags, so that clients who don't need this don't need to unnecessarily pay the cost of embedding the reference image in the binary (which is just 1KB though).

Obviously running this method can be time-consuming and a lot of care needs to be be taken on when to schedule this function call, but it's still useful to have this mechanism for anyone who wants to perform this kind of basic sanity check.

# Notes
I still want to test this properly on some mobile devices to ensure it works, but otherwise the overall PR should be ready for review.


sparse_strips/vello_common/src/probe.rs
```diff
@@ -0,0 +1,396 @@
...
+            Self::Error(ProbeError {
+                expected: expected.into_png().unwrap(),
+                actual: actual.into_png().unwrap(),
+            })
+        }
+    }
+}
+
+/// API necessary to draw the probe scene.
+pub trait ProbeRenderer {
```
LaurenzV, line 94: I also don't like adding yet another "rendering abstraction", but I want to avoid using recordings for this since 1) they use more memory and 2) it's uncertain whether we will keep them in the near future.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -354,6 +326,117 @@ impl WebGlRenderer {
...
+        };
+        let (atlas_width, atlas_height) = atlas_config.atlas_size;
+
+        let mut probe_image_cache = ImageCache::new_with_config(atlas_config);
+        let mut probe_atlas_texture_array =
+            create_atlas_texture_array(&self.gl, atlas_width, atlas_height, 1);
+        core::mem::swap(
+            &mut self.programs.resources.atlas_texture_array,
+            &mut probe_atlas_texture_array,
+        );
```
LaurenzV, line 398: All of these `mem::swap`s and `Option::take` are getting really unwieldy and IMO it's getting harder to keep track of the state, I hope that we can improve this in a future PR. Otherwise, it's just going to be a matter of time until we somewhere forget to properly reset the state. 😦

LaurenzV:
Not ready for in-depth review yet, as I want to clean up `webgl.rs` more tomorrow, but comments on the high-level idea are welcome. @taj-p @ajakubowicz-canva

sparse_strips/vello_common/src/probe.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// The probe matched the bundled reference image.
+    Success,
+    /// The probe failed and returns the expected image and the actual image.
+    Error(ProbeError),
+}
+
+/// Probe failure output.
+#[derive(Debug, Clone)]
+pub struct ProbeError {
+    /// The expected image, encoded as PNG.
```
ajakubowicz-canva: Does probing need to depend on the `png` feature? Could a byte array of rgb be used instead?

ajakubowicz-canva: Also overall this PR is looking good.

LaurenzV: It could, but it would increase the size of the reference image from around 1KB to 10KB, so not sure how nice that is.

taj-p:
Approach looks good! When you're ready, request a review 🙏

sparse_strips/vello_common/src/probe.rs
LaurenzV: I could however change it so that `actual` is sent as a byte array instead of a PNG.

LaurenzV:
Should be ready now I think!

sparse_strips/vello_common/src/probe.rs
taj-p: We could remove this trait by moving the probe module to `vello_hybrid` and using Vello Hybrid's scene directly, thereby using hybrid to generate the reference but then we lose the "source of truth" that Vello CPU provides. 

One could argue we already have that source of truthing via our unit test suite, but introducing that indirection for probe feels awkward.

Happy with this approach if you think it best.

sparse_strips/vello_common/src/probe.rs
taj-p: Wouldn't `png` be larger? Have you checked this? I think the reference images, as raw bytes, may be 10 kB in the raw binary, but it would compress fairly well during transport right? Potentially all the way to 1kB with gzip because the reference looks fairly compressible - whereas you might not see that much with the `.png` file.

I suspect the `png` variant will actually be larger on binary size (especially gzipped) if the consumer makes no use of `png` besides for probe.

taj-p: I tested this. I'm seeing ~55 kB gzipped greater binary size with PNG decoding vs using raw buffer.

The smallest binary size outcome might be for the caller to pass in the raw data and use platform capabilities (e.g. web api) to decode, but that gets complicated fast, so just using the raw buffer (without `png` seems best).

taj-p (approved):
LGTM with comments addressed 🙏

sparse_strips/vello_common/src/probe.rs
LaurenzV: If the whole binary will be zipped then yes, there probably won't be a huge difference in binary size. I'll change it back then.

sparse_strips/vello_common/src/probe.rs
LaurenzV: I think for now it's best to keep it this way, since setting up all of the wgpu stuff just for the update script seems a bit overkill. 😄

LaurenzV:
I've tweaked it now so that we only send/store the raw RGBA data, instead of requiring the png dependency. As commented above, I've left it unchanged that the reference image is generated using Vello CPU instead of Vello Hybrid for now.

Also added a test for ensuring the probe image stays up-to-date.

LaurenzV:
I realize it might be good to have 1-2 more tests that test the depth buffer explicitly. but will leave that for a follow-up PR

taj-p (approved):
LGTM!

PR #1597, opened 2026-04-22
Vello CPU: Add public API to render with offset/stride

### LINKED ISSUE #1382: Support `PixmapMut<'_>` and stride in `vello_cpu`?

nicoburns:
- Fixes https://github.com/linebender/vello/issues/1382

This PR is pretty simple as the logic for doing this already existed (I think added for [`RenderContext::composite_to_pixmap_at_offset`](https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.composite_to_pixmap_at_offset), and this just plumbing through 4 extra `u16` parameters in a bunch of places.

## Todo
- [x] ~~The logic in this PR is unfortunately not quite right as things stand. Testing shows the stride not being accounted for even when using the new method)~~ (looks like I'd just missed a method that needed to be updated)

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
@@ -93,12 +93,19 @@ impl SingleThreadedDispatcher {
         buffer: &mut [u8],
         width: u16,
         height: u16,
+        dst_x: u16,
+        dst_y: u16,
+        dst_buffer_width: u16,
+        dst_buffer_height: u16,
         encoded_paints: &[EncodedPaint],
         image_resolver: &dyn ImageResolver,
     ) {
         use crate::fine::F32Kernel;
         use vello_common::fearless_simd::dispatch;
-        dispatch!(self.level, simd => self.rasterize_with::<_, F32Kernel>(simd, buffer, width, height, encoded_paints, image_resolver));
+        dispatch!(self.level, simd => self.rasterize_with::<_, F32Kernel>(simd, buffer, width, height, dst_x,
+                dst_y,
+                dst_buffer_width,
+                dst_buffer_height,encoded_paints, image_resolver));
     }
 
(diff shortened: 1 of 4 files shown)


sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
nicoburns, line 235: I did not update this instance of `Regions::new`, as I think it's rendering into an intermediate layer and not the output buffer. Does that seems correct?

LaurenzV:
I think I might have a better idea now how to implement this and consolidate it with exsting APIs. Need to try it out, but if it works out I will open a PR soon.

LaurenzV:
I believe this PR is not relevant anymore?

nicoburns:
Correct, this was superceded by https://github.com/linebender/vello/pull/1665

PR #1598, opened 2026-04-22
vello_sparse_shaders: Add binary to generate WebGL shaders locally for easier inspection

LaurenzV:
Currently, `vello_sparse_shaders` automatically generates native WebGL shaders using `naga` via a build script, ensuring that our native WebGL backend can directly consume them. However, there is currently no easy way to inspect what the generated shaders look like, instead of actually checking the build directory.

Therefore, this PR proposes explicitly checking the generated sources into the repository. This change 1) Makes it easier to inspect what the generated WebGL shader actually looks like and most importantly 2) allows us to more easily track potential regressions that might arise from, for example, updating naga to a new version. While there is no expectation for anyone to always manually validate the generated shader after changes, I still think this is worth doing to keep better track of this.

Especially as we tackle tracking down issues on low-tier devices, it will be helpful to be able to more easily inspect the generated shader to see what the issue is.


sparse_strips/vello_sparse_shaders/src/main.rs
taj-p, line 1: WDYT about adding a short header to the file like:

```rs
/// AUTOGENERATED. DO NOT EDIT. 
/// To regenerate, see `//sparse_strips/vello_sparse_shaders/README.md`.
```

sparse_strips/vello_sparse_shaders/README.md
```diff
@@ -20,6 +20,15 @@ Vello Hybrid renderer.
 This crate provides the WGSL programs and build step for GLSL programs that are used by the
 optimized hybrid rendering engine.
 
+Whenever the WGSL shaders are updated, the build script will automatically take care of
+regenerating the GLSL programs. However, in addition to that, a snapshot of the shaders is
+also stored in the `generated_glsl` folder, which is not updated automatically. Before opening
+a PR, you will have to update them. Please use the following command for doing so:
```
taj-p:
```suggestion
also stored in the `generated_glsl` folder, which is not updated automatically. If your PR
edits a shader, you will have to update the snapshots. Please use the following command 
for doing so:
```

taj-p (approved):
Great idea! Nice! LGTM

taj-p (changes requested):
We spoke about this in OH and it introduces too much friction and any issues with GLSL output should be handled upstream.

Raph suggested introducing build-time checks / linting tool to address any known GLSL issues.

We can unblock this PR if we simply output the GLSL shader to stdout

LaurenzV:
I've changed it now so that the shaders are not checked into git, but there is still a tool that allows you to regenerate them into a git-ignored folder such that it's easier to inspect. I hope this is less controversial!

taj-p (approved):

PR #1604, opened 2026-04-27
Avoid using structs in fragment shaders

LaurenzV:
😔 Unfortunately, on a row of older Adreno GPUs, there is a very annoying bug where the numeric precision of structs is always 16-bit, regardless of whether high precision was requested before. See https://github.com/KhronosGroup/WebGL/issues/3351 and also linked issues there.

Those GPUs seem to be pretty common, I found it on at least three different devices on BrowserStack, so seems like we will have to work around it... So in this PR, I removed any structs from fragment shaders where having 32-bit precision is important for correctness. For consistency, maybe we should extend this to just never using structs in fragment shader so we don't accidentally miss anything in the future, but that could be a follow-up. With this change, I'm able to get Vello Hybrid running on the devices in question.

It definitely is ugly and I'd prefer not having to merge this unless strictly required, so ideas are very much welcome.

### DIFF
--- a/sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
@@ -11,6 +11,11 @@
 // for the drop shadow filter, we need to composite the original content on top of the shadow.
 @group(2) @binding(0) var original_tex: texture_2d<f32>;
 
+// VERY IMPORTANT NOTE: Whenever making use of structs in the fragment shader, 
+// make sure that any field could be interpreted using 16-bit precision and still work
+// correctly!! See <ISSUE> for more information.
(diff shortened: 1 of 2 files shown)


LaurenzV:
I just realized I'm not sure if the `image_offset` field in `EncodedImage` is enough with this 16-bit floating points... Might be a bit risky.

taj-p:
If the bug is in the driver, it's unclear whether there's another viable option than something like this. I'm not sure we could implement it in `naga`. Deferring to Alex here - maybe Alex knows another path

waywardmonkeys:
@cwfitzgerald This is the issue I mentioned to you yesterday.

LaurenzV:
I will split this up into two new separate PRs.

PR #1606, opened 2026-04-27
Fix blurry image rendering in Vello Classic.

### LINKED ISSUE #972: Images do not roundtrip

xStrom:
Images were rendered incorrectly due to pixel corners being sampled instead of centers. This PR fixes that.

This is based on #1499. I also applied the same fix to the `IMAGE_QUALITY_HIGH` code path, updated impacted snapshots, and converted relevant known issue tests to regression tests.

Fixes #972.


RagibHasin (approved):
Looks good. Thanks!

PR #1608, opened 2026-04-27
Increase hybrid tolerance for two filter tests.

xStrom:
Both `filter_with_inner_clip` and `filter_drop_shadow_inside_clip` needed a bump in tolerance in order to pass on my RTX 4090 Windows 11 setup.

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/filter.rs
+++ b/sparse_strips/vello_sparse_tests/tests/filter.rs
@@ -1473,7 +1473,7 @@ fn filter_with_non_rect_clip(ctx: &mut impl Renderer) {
     ctx.pop_layer();
 }
 
-#[vello_test(skip_multithreaded)]
+#[vello_test(skip_multithreaded, hybrid_tolerance = 1)]
 fn filter_drop_shadow_inside_clip(ctx: &mut impl Renderer) {
     let filter = Filter::from_primitive(FilterPrimitive::DropShadow {
         dx: 20.0,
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):

sparse_strips/vello_sparse_tests/tests/filter.rs
grebmeg, line 1: I think there's a real problem in our test framework that this PR surfaces: having a single unified `hybrid_tolerance` knob has a few compounding downsides.

- It hides which backend is actually bad. A bump from 1 to 2 here could be wgpu, webgl, or the constrained variant — the threshold doesn't tell us, and neither will future readers.
- It moves the weakest backend's slop onto everyone. Whichever backend needs the loosest bound dictates the bound for all the others, so the stricter backends silently lose coverage.
- It conflates unrelated sources of error. The reason wgpu might deviate (e.g. f32 rounding in the shader) is usually unrelated to why webgl deviates (e.g. naga WGSL→GLSL translation, driver quirks). Folding them into one number makes it impossible to reason about either.
- It makes per-backend regressions invisible. If webgl silently degrades from 1 to 3 but the global tolerance is already 4 (because some other test needed it), nobody notices until webgl crosses 4.

A while back I did a spike on extending `vello_dev_macro` (unfortunately, I didn’t have enough time to finish it; happy for someone else to take it on, otherwise I can come back to it later) — it does add some verbosity, but given the failure modes above, I think the trade-off is worth it. 

cc @LaurenzV

LaurenzV: Hmm yeah it's not ideal, but wouldn't know how to improve this really. 🤔 I think as long as the tolerances are low enough, it should still give us enough confidence that it's just small deviations instead of a bigger problem. But maybe I'm wrong. 😄

grebmeg: What I did in that spike was update the macro so you can configure parameters per backend, while still having common defaults for each parameter. That way, if you want to tune a specific platform's parameter in a test, you can adjust it directly there.

This should also help us gather a lot more data over time, for example, identifying that Windows devices tend to use certain adjustments more frequently, and so on.

Issue #1373, opened 2026-01-22
Vello Hybrid: Support blurred rounded rectangles


DJMcNab:
Currently, the "blurred rounded rectangle" paint type is supported by both Vellos Classic and CPU, but not Vello Hybrid. We should implement it for Vello Hybrid.

The biggest difficulty I can foresee with this task is wiring it through the pipeline. But in theory, it need only be one bit in addition to the current solid colour paint, so hopefully we can find.

The existing WGSL code for blurred rounded rectangles (added in #665) should be pretty much reusable in Vello Hybrid.

nicoburns:
~~I think this should enable the existing blurred rounded rectangle tests `vello_hybrid`. At least https://github.com/linebender/vello/blob/main/vello_tests/tests/snapshot_test_scenes.rs#L111, but perhaps there are others?~~

EDIT: looks like they're already enabled and the skip is conditional on a cfg. Are we just not running GPU tests in CI? (shouldn't there be some kind of snapshot update?)

LaurenzV:
We are running GPU tests in CI. If they are not skipped via an attribute, they are probably skipped in vello_dev_macros instead via pattern matching.

DJMcNab:
Those are the Vello Classic tests...
Also this is an issue, there isn't any associated code...

nicoburns:
Implemented in https://github.com/linebender/vello/pull/1610

PR #1610, opened 2026-04-28
`vello_hybrid`: support blurred rounded rects fast path

waywardmonkeys:
This wires Hybrid through the existing shared `vello_common` blurred rounded rect encoding, adds GPU/WebGL encoded paint upload support, and teaches the sparse strip shader to evaluate the analytic blurred rounded rect paint. Root-level unclipped cases use the existing direct rectangle fast path; other supported cases fall back through the normal strip/coarse path.

Also adds a small Hybrid `winit` demo scene, adapted from the Classic test scene, so the feature can be inspected interactively.

### DIFF
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -11,6 +11,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::cell::RefCell;
 use core::ops::Range;
+use vello_common::blurred_rounded_rect::BlurredRoundedRectangle;
 use vello_common::clip::ClipContext;
 use vello_common::coarse::{MODE_HYBRID, Wide, WideTilesBbox};
 use vello_common::encode::{EncodeExt, EncodedPaint};
@@ -22,6 +23,7 @@ use vello_common::multi_atlas::AtlasConfig;
 use vello_common::paint::{Paint, PaintType, Tint};
 #[cfg(feature = "text")]
 use vello_common::peniko::FontData;
+use vello_common::peniko::color::palette::css::BLACK;
 use vello_common::peniko::{BlendMode, Compose, Fill, Mix};
 use vello_common::render_graph::{RenderGraph, RenderNodeKind};
 use vello_common::render_state::RenderState;
@@ -209,7 +211,7 @@ pub struct Scene {
     // The reason we use `RefCell` here is that during `render`, we need
     // mutable access so we can store additional encoded paints for filtered layers,
     // if applicable.
-    /// Storage for encoded gradient and image paint data.
+    /// Storage for encoded non-solid paint data.
     pub(crate) encoded_paints: RefCell<Vec<EncodedPaint>>,
     /// Whether the current paint is visible (e.g., alpha > 0).
     paint_visible: bool,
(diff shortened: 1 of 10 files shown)


waywardmonkeys:
This was done with the assistance of Codex (5.5, high). I've reviewed it and think it is right, but I don't know all of the hybrid internals to be sure I didn't miss something.

DJMcNab:
Does this address https://github.com/linebender/vello/issues/1373?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -509,6 +513,10 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
                 sweep_gradient.texture_width
             );
             final_color = alpha * gradient_color;
+        } else if paint_type == PAINT_TYPE_BLURRED_ROUNDED_RECT {
+            let paint_tex_idx = in.paint_and_rect_flag & PAINT_TEXTURE_INDEX_MASK;
+            let blurred_rect = unpack_blurred_rounded_rect(paint_tex_idx);
```
LaurenzV, line 518: I can probably just do it myself depending on which PR gets merged first, but just as a note the way we process paint types is likely to change here: https://github.com/linebender/vello/pull/1619/changes

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -1350,6 +1377,67 @@ fn unpack_sweep_gradient(paint_tex_idx: u32) -> SweepGradient {
...
+    let y = clamp(x * 1.1283791671, -100.0, 100.0);
+    let yy = y * y;
+    let z = y + (0.24295 + (0.03395 + 0.0104 * yy) * yy) * (y * yy);
+    return z / sqrt(1.0 + z * z);
+}
+
+// Approximation for the convolution of a gaussian filter with a rounded rectangle.
+fn calculate_blurred_rounded_rect(fragment_pos: vec2<f32>, rect: BlurredRoundedRect) -> vec4<f32> {
+    let local_xy = rect.transform * fragment_pos + rect.translate;
+    let y = local_xy.y - 0.5 * rect.height;
```
LaurenzV, line 1425: Can you maybe add a note that 0.5 and 0.0 correspond to v0 and v1 in the CPU code? And in general, a small comment that this shader is modelled after the vello_cpu variant (instead of vello itself)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -506,20 +508,47 @@ impl Scene {
         }
     }
 
+    fn try_fast_rect(&mut self, rect: &Rect) -> bool {
+        if self.fast_rect_bounds(rect).is_none() {
```
LaurenzV: Now we are doing the computation twice, once here and then again in `try_fast_rect_with_paint`. Could we not change it so that
1) `fast_rect_bounds` just returns a normal Option<kurbo::Rect> (since the coordinates are f64 anyway) and then
2) Turn `try_fast_rect_with_paint` into a method that takes the calculated bounds directly and just pushes it to the fast strips buffer (and rename into something like `push_fast_rect`)?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -540,29 +569,75 @@ impl Scene {
...
+            let inflated_rect = rect.inflate(f64::from(kernel_size), f64::from(kernel_size));
+            let transform = ctx.render_state.transform * ctx.render_state.paint_transform;
+            let paint =
+                blurred_rect.encode_into(&mut ctx.encoded_paints.borrow_mut(), transform, None);
+
+            if ctx.try_fast_rect_with_paint(&inflated_rect, paint.clone()) {
+                return;
+            }
+
+            if is_axis_aligned(&ctx.render_state.transform) && ctx.aliasing_threshold.is_none() {
```
LaurenzV, line 608: I wonder whether all of this can be deduplicated somehow, but maybe better for a follow-up.

LaurenzV (commented):
LGTM overall, from some testing on my tablet I wasn't able to determine any slowdown from adding this additional shader code.

sparse_strips/vello_hybrid/src/scene.rs
waywardmonkeys: Improved now, I think.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
waywardmonkeys: Done.

grebmeg (approved):
I checked on a low-tier device, a Samsung Galaxy A05s, to make sure the PR doesn’t degrade performance, and it looks like it doesn’t. I ran fairly intensive tests with 10 warm-up runs and 100 iterations for each benchmark.

From the results, there doesn’t seem to be any noticeable degradation, although benchmarking GPU performance properly is always tricky... In interactive mode, it feels fine as well.

Ideally, we’d introduce feature gating for such features, but we currently don’t have the infrastructure for that. The long-term plan is to move to more specialized shaders, but until then, I still think it’s better to have the feature gating in place since it doesn’t take much effort to implement. Otherwise, users might not even use some of these features, but still end up paying the performance cost for them.

waywardmonkeys:
> Ideally, we’d introduce feature gating for such features, but we currently don’t have the infrastructure for that. The long-term plan is to move to more specialized shaders, but until then, I still think it’s better to have the feature gating in place since it doesn’t take much effort to implement. Otherwise, users might not even use some of these features, but still end up paying the performance cost for them.

You'll have to sort out the CI cost / impact for this due to how CI checks features. That makes additional features more painful (and it is already bad).

PR #1611, opened 2026-04-29
Remove recordings from sparse strips renderers

LaurenzV:
See the [Zulip](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Removing.20recordings.20from.20vello_common/with/590186223) thread for more discussion. 

Note that this is still pending approval from @grebmeg, so still not 100% settled whether we want to do this. However, in case we do decide to go ahead with this, this is the PR that does it!


grebmeg (approved):
As discussed, recording functionality is currently unused and has been getting in the way of developing other features while we’ve been carrying it for a long time. The original idea was to apply transforms on sparse strips (and the data was promising). After discussing it, we decided to remove it for now and revisit it when the timing is better.

Could you also add a line or two to the CHANGELOG about removing recordings?

PR #1614, opened 2026-05-02
sparse_strips: Don't enable `std` feature when probing is enabled

LaurenzV:
1) There is no reason why enabling the `std` feature should imply enabling `wgpu`. It should be possible to enable std and still only use the WebGL backend.
2) `probe` doesn't need the `std` feature. Initially the probe feature made use of `png` which required that, but now that it just uses raw RGBA bytes this is not needed anymore.


nicoburns (approved):

LaurenzV:
Uff, I guess some more work is needed to figure out CI. 😄

LaurenzV:
I'll repurpose this PR to just not enable the std feature for the probe feature.

PR #1615, opened 2026-05-03
Add two more benchmarks for alpha fill rendering

LaurenzV:
This PR adds two more benchmarks for strip painting in Vello CPU. One for a single strip (since this is probably actually the most common one), and one for strips of medium lengths.

### DIFF
--- a/sparse_strips/vello_bench/src/fine/strip.rs
+++ b/sparse_strips/vello_bench/src/fine/strip.rs
@@ -16,14 +16,32 @@ use vello_cpu::fine::{Fine, FineKernel};
 use vello_dev_macros::vello_bench;
 
 pub fn strip(c: &mut Criterion) {
+    solid_single(c);
     solid_short(c);
+    solid_medium(c);
     solid_long(c);
+}
+
+#[vello_bench]
+pub fn solid_single<S: Simd, N: FineKernel<S>>(b: &mut Bencher<'_>, fine: &mut Fine<S, N>) {
+    let paint = Paint::Solid(PremulColor::from_alpha_color(ROYAL_BLUE));
+    let width = Tile::WIDTH;
+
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_bench/src/fine/strip.rs
```diff
@@ -16,10 +16,20 @@ use vello_cpu::fine::{Fine, FineKernel};
 use vello_dev_macros::vello_bench;
 
 pub fn strip(c: &mut Criterion) {
+    solid_single(c);
     solid_short(c);
+    solid_medium(c);
     solid_long(c);
 }
```
grebmeg, line 23: Have you considered making this a parameterized run instead? The bodies are identical except for the `width`, and criterion's `BenchmarkGroup` + `bench_with_input` would let you define widths in a single array, get a width-vs-time plot for free, and avoid having to register each new width here.

Concretely it'd be nice to collapse to something like:

```rust
const STRIP_WIDTHS: &[usize] = &[1, 2, 4, 8, 16, 32, 64, 128, 256];

#[vello_bench(widths = STRIP_WIDTHS)]
pub fn solid<S: Simd, N: FineKernel<S>>(
    b: &mut Bencher<'_>,
    fine: &mut Fine<S, N>,
    width: usize,
) {
    let paint = Paint::Solid(PremulColor::from_alpha_color(ROYAL_BLUE));
    strip_single(&paint, &[], width, b, fine);
}
```

grebmeg (approved):

sparse_strips/vello_bench/src/fine/strip.rs
LaurenzV: Could do that, but if would require some major changes to the current infrastrcucture, so I will leave that for the future. 😄

PR #1619, opened 2026-05-04
vello_sparse_shaders: Avoid using structs for storing data

LaurenzV:
This is part 1 of implementing a solution to the problem raised in #1604. Basically, instead of unpacking paint data into dedicated structs, we instead just sample the raw texels whever required and then create dedicated getter methods that extract the field in question from that texel. It does make things a bit less readable, but it's necessary to avoid the precision loss that has been raised in that PR.


grebmeg (commented):
- Do I understand correctly that the generated GLSL currently does not emit `highp` on the relevant struct fields, and that this is what naga generates today? My understanding is that the issue may be deeper, though: even if we emit `highp` there, some drivers may still actually demote struct fields to `mediump`?

- Could you please share which devices you observed the issue on and this change fixes it?

- I find the proposed approach is hard to read and maintain. I’d prefer to look for another solution:

  - First, I think we may be able to fix precision qualifier emission in naga. From a quick look, [write_struct_body](https://github.com/gfx-rs/wgpu/blob/34e688f8160b1e132060fe6c216074be2de12174/naga/src/back/glsl/writer.rs#L1529-L1583) does not seem to write each member through `write_type` with a precision prefix, unlike [file-level defaults](https://github.com/gfx-rs/wgpu/blob/34e688f8160b1e132060fe6c216074be2de12174/naga/src/back/glsl/writer.rs#L149-L154), [explicit highp](https://github.com/gfx-rs/wgpu/blob/34e688f8160b1e132060fe6c216074be2de12174/naga/src/back/glsl/writer.rs#L603-L619) on sampler/image declarations, and sampler function parameters. So I believe this might be fixable in naga for the GLSL backend.

  - Second, based on WebGL issue KhronosGroup/WebGL#3351, the problem seems reproducible in fragment shaders specifically. Could we move the struct initialisation/unpacking to the vertex shader instead? I briefly looked at WebRender, and it seems that most “unpack into a struct” logic happens in the vertex shader.

LaurenzV:
> Do I understand correctly that the generated GLSL currently does not emit highp on the relevant struct fields, and that this is what naga generates today? My understanding is that the issue may be deeper, though: even if we emit highp there, some drivers may still actually demote struct fields to mediump?

Yes, the issue _is_ deeper, as you mentioned. I don't remember off the top of my head what naga currently emits (but I believe it just emits the `precision highp float/int;` at the beginning to set the default), but from my experiments even directly setting the precision qualifier on the struct field does not change anything, so that's not the problem. The problem is simply that the driver _always_ uses mediump for struct fields in the fragment shader, regardless of what you set.

If you want, you can play around with the following shader on the given phones. If the device work corrects it should return green, otherwise red. I was not able to adjust the shader to use structs on some way without getting rid of the red color on the given phones.

```
#version 300 es

precision highp float;
precision highp int;

struct Pair {
    uint f0;
};

layout(location = 0) out vec4 out_color;

void main() {
    Pair p = Pair(0x00010000u);
    uint mode = p.f0 >> 16;
    float green = (mode == 1u) ? 1.0 : 0.0;
    float red = 1.0 - green;
    out_color = vec4(red, green, 0.0, 1.0);
}
```

> Could you please share which devices you observed the issue on and this change fixes it?

There were a couple, but some for example include Google Pixel 5, Redmi Note 11 and Samsung Galaxy Tab A9. So it seems to affect Adreno GPU's in the low 600 range, maybe also 500 and 400, I haven't checked. But it affects many devices.

Also note that if you want to test it you need to do it in conjunction with #1620, just this PR alone might not fix it if I remember correctly.

> First, I think we may be able to fix precision qualifier emission in naga.

As mentioned, from my experiments just adding precision qualifiers doesn't help here. You would need some kind of "flatten structs" functionality in naga, and that doesn't seem to exist (and seems very hard, if at all possible, to implement, given that structs can probably be arbitrarily nested?)

> Could we move the struct initialisation/unpacking to the vertex shader instead? I briefly looked at WebRender, and it seems that most “unpack into a struct” logic happens in the vertex shader.

From my experiments, even structs that are passed from the vertex shader to the fragment shader are affected by this. This is the whole reason why #1620 is also needed to fix the issue. It seems like you  literally cannot use structs in fragment shaders. :/ Feel free to try to double check my findings if you want to, would be good to have confirmation that you also weren't able to find a workaround, but I've spent a lot of time on it, so I'm feeling pretty certain that I didn't miss anything here.

LaurenzV:
So, I tried it again now. As mentioned, just adding `highp` to the struct field doesn't change anything. The following HTML page still shows up red for me:

```html
<!doctype html>
<html>
<head>
  <style>
    html, body, canvas {
      width: 100%;
      height: 100%;
      margin: 0;
      display: block;
      overflow: hidden;
    }
  </style>
</head>
<body>
  <canvas></canvas>
  <script>

void main() {
    vec2 positions[3] = vec2[3](
        vec2(-1.0, -1.0),
        vec2( 3.0, -1.0),
        vec2(-1.0,  3.0)
    );
    gl_Position = vec4(positions[gl_VertexID], 0.0, 1.0);
}
`;

precision highp float;
precision highp int;

struct Pair {
    highp uint f0;
};

layout(location = 0) out vec4 out_color;

void main() {
    Pair p = Pair(0x00010000u);
    uint mode = p.f0 >> 16;
    float green = (mode == 1u) ? 1.0 : 0.0;
    float red = 1.0 - green;
    out_color = vec4(red, green, 0.0, 1.0);
}
`;

    function shader(type, source) {
      const result = gl.createShader(type);
      gl.shaderSource(result, source);
      gl.compileShader(result);
      return result;
    }

    gl.attachShader(program, shader(gl.VERTEX_SHADER, vertexSource));
    gl.attachShader(program, shader(gl.FRAGMENT_SHADER, fragmentSource));
    gl.linkProgram(program);
    gl.useProgram(program);

    function draw() {
      const scale = devicePixelRatio;
      canvas.width = innerWidth * scale;
      canvas.height = innerHeight * scale;
    }

    addEventListener("resize", draw);
    draw();
  </script>
</body>
</html>

```

Struct unpacking in the vertex shader also doesn't work, the following still shows up red:
```html
<!doctype html>
<html>
<head>
  <style>
    html, body, canvas {
      width: 100%;
      height: 100%;
      margin: 0;
      display: block;
      overflow: hidden;
    }
  </style>
</head>
<body>
  <canvas></canvas>
  <script>

precision highp float;
precision highp int;

struct Pair {
    highp uint f0;
};

flat out Pair v_pair;

void main() {
    vec2 positions[3] = vec2[3](
        vec2(-1.0, -1.0),
        vec2( 3.0, -1.0),
        vec2(-1.0,  3.0)
    );
    v_pair = Pair(0x00010000u);
    gl_Position = vec4(positions[gl_VertexID], 0.0, 1.0);
}
`;

precision highp float;
precision highp int;

struct Pair {
    highp uint f0;
};

flat in Pair v_pair;
layout(location = 0) out vec4 out_color;

void main() {
    uint mode = v_pair.f0 >> 16;
    float green = (mode == 1u) ? 1.0 : 0.0;
    float red = 1.0 - green;
    out_color = vec4(red, green, 0.0, 1.0);
}
`;

    function shader(type, source) {
      const result = gl.createShader(type);
      gl.shaderSource(result, source);
      gl.compileShader(result);
      return result;
    }

    gl.attachShader(program, shader(gl.VERTEX_SHADER, vertexSource));
    gl.attachShader(program, shader(gl.FRAGMENT_SHADER, fragmentSource));
    gl.linkProgram(program);
    gl.useProgram(program);

    function draw() {
      const scale = devicePixelRatio;
      canvas.width = innerWidth * scale;
      canvas.height = innerHeight * scale;
    }

    addEventListener("resize", draw);
    draw();
  </script>
</body>
</html>
```

However, removing the wrapper struct and just passing `f0` directly works and now gives me a green color:
```html
<!doctype html>
<html>
<head>
  <style>
    html, body, canvas {
      width: 100%;
      height: 100%;
      margin: 0;
      display: block;
      overflow: hidden;
    }
  </style>
</head>
<body>
  <canvas></canvas>
  <script>

precision highp float;
precision highp int;

flat out uint v_f0;

void main() {
    vec2 positions[3] = vec2[3](
        vec2(-1.0, -1.0),
        vec2( 3.0, -1.0),
        vec2(-1.0,  3.0)
    );
    v_f0 = 0x00010000u;
    gl_Position = vec4(positions[gl_VertexID], 0.0, 1.0);
}
`;

precision highp float;
precision highp int;

flat in uint v_f0;
layout(location = 0) out vec4 out_color;

void main() {
    uint mode = v_f0 >> 16;
    float green = (mode == 1u) ? 1.0 : 0.0;
    float red = 1.0 - green;
    out_color = vec4(red, green, 0.0, 1.0);
}
`;

    function shader(type, source) {
      const result = gl.createShader(type);
      gl.shaderSource(result, source);
      gl.compileShader(result);
      return result;
    }

    gl.attachShader(program, shader(gl.VERTEX_SHADER, vertexSource));
    gl.attachShader(program, shader(gl.FRAGMENT_SHADER, fragmentSource));
    gl.linkProgram(program);
    gl.useProgram(program);

    function draw() {
      const scale = devicePixelRatio;
      canvas.width = innerWidth * scale;
      canvas.height = innerHeight * scale;
    }

    addEventListener("resize", draw);
    draw();
  </script>
</body>
</html>
```

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -389,106 +396,118 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
+                    i32(image_atlas_index),
                     image_offset,
                     image_size,
-                    encoded_image.extend_modes,
-                    encoded_image.image_padding,
+                    image_extend_modes,
+                    image_padding,
                 );
-            } else if encoded_image.quality == IMAGE_QUALITY_MEDIUM {
+            } else if get_image_quality(image_raw0) == IMAGE_QUALITY_MEDIUM {
```
grebmeg: `get_image_quality(image_raw0)` is called twice, while the other image getters were hoisted into `image_offset`/`image_size`/`image_extend_modes`/`image_atlas_index`/`image_padding`. Could you add it to the precompute block for consistency?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -389,106 +396,118 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
...
                     vec2<u32>(final_xy),
-                    i32(encoded_image.atlas_index),
+                    i32(image_atlas_index),
                     0,
                 );
             }
 
-            let is_multiply = bool(encoded_image.tint_mode);
+            let image_tint = get_image_tint(image_raw2);
+            let is_multiply = bool(get_image_tint_mode(image_raw2));
```
grebmeg: Both `get_image_tint` and `get_image_tint_mode` re-extract `raw2.y` and apply the same `packed_tint != 0u` `select`. Could the tint branch be precomputed at the call site, similar to how `image_offset`/`image_size`/etc. are precomputed above?

```wgsl
let packed_tint = image_raw2.y;
let has_tint = packed_tint != 0u;
let image_tint = select(vec4<f32>(1.0), unpack4x8unorm(packed_tint), has_tint);
let is_multiply = has_tint && image_raw2.z != 0u;
```

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -330,6 +334,9 @@ var alphas_texture: texture_2d<u32>;
 @group(0) @binding(2)
 var clip_input_texture: texture_2d<f32>;
 
+// Note: When extending the fragment shader, do not create new structs.
+// See also https://github.com/linebender/vello/pull/1604 for more information on why.
```
grebmeg: Could you please relocate this to the file-level docblock? Maybe add something like an uppercase **WARNING**, or another more appropriate marker, and expand the explanation a bit.

Specifically, mention that this issue affects buggy Adreno GPUs (for example, some devices like the Pixel 5), where `f32`/`u32` values stored in struct fields get demoted to `f16`/`u16` in fragment shaders. Please also keep the link to the issue. Maybe also mention the alternatives we tried that didn’t end up solving the issue properly, and add a note that contributions or better solutions are welcome if someone finds a cleaner approach.

I think this would make it much easier for anyone reading the shader to quickly understand why we avoid using structs there.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -874,46 +881,55 @@ fn encoded_paint_coord(flat_idx: u32) -> vec2<u32> {
...
+fn load_encoded_paint_texel(paint_tex_idx: u32, texel_offset: u32) -> vec4<u32> {
+    return textureLoad(
+        encoded_paints_texture,
+        encoded_paint_coord(paint_tex_idx + texel_offset),
+        0,
     );
-    let translate = vec2<f32>(bitcast<f32>(texel1.w), bitcast<f32>(texel2.x));
+}
+
+fn get_image_quality(raw0: vec4<u32>) -> u32 { return raw0.x & 0x3u; }
```
grebmeg: With removing structs we got rid of field documentation too. I think it's quite clear from the fucntions name what they do, but some of them still deserve documentation (e.g. `get_radial_f_is_swapped`, `get_radial_fp1`, etc.). It might be 100% clear what these are about, so it would be nice to leave some comments.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -843,29 +873,6 @@ fn blend_mix(cb: vec3<f32>, cs: vec3<f32>, mode: u32) -> vec3<f32> {
 const TINT_MODE_ALPHA_MASK: u32 = 0u;
 const TINT_MODE_MULTIPLY: u32 = 1u;
```
grebmeg, line 900: The previous struct definition documented each field (`quality: u32`, `extend_modes: vec2<u32>`, etc.) and the bit packing was explained in the `unpack_encoded_image` body. After this change the layout is reconstructed across 9 separate getters with no inline doc, and the only place describing the encoding is `pack_image_params` in `vello_hybrid/src/render/common.rs`. Could we add a short header block above this cluster of getters with the bit layout, plus `// Must match GpuEncodedImage in vello_hybrid/src/render/common.rs`? That gives a contributor a single place to look when extending the encoding and at least documents the contract that's now implicit.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
grebmeg: Also, what do you think about panicking with an explicit message when generating glsl shaders if a `struct` is used?

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -296,10 +296,14 @@ fn vs_main(
 
         if paint_type == PAINT_TYPE_IMAGE {
             let paint_tex_idx = instance.paint_and_rect_flag & PAINT_TEXTURE_INDEX_MASK;
-            let encoded_image = unpack_encoded_image(paint_tex_idx);
+            let image_raw0 = load_encoded_paint_texel(paint_tex_idx, 0u);
+            let image_raw1 = load_encoded_paint_texel(paint_tex_idx, 1u);
+            let image_raw2 = load_encoded_paint_texel(paint_tex_idx, 2u);
```
grebmeg: What about using a `texel` or `texl` postfix for this? I think that would make things a bit more understandable and better match what the `load_encoded_paint_texel` helper returns.

grebmeg (approved):
> So, I tried it again now. As mentioned, just adding highp to the struct field doesn't change anything. The following HTML page still shows up red for me:

> Feel free to try to double check my findings if you want to, would be good to have confirmation that you also weren't able to find a workaround, but I've spent a lot of time on it, so I'm feeling pretty certain that I didn't miss anything here.

Thanks for trying my suggestion, but you’re right, there doesn’t seem to be a good visible workaround here. I’ve also spent some time looking for alternatives, but unsuccessfully. After reading through the changes one more time, I don’t find them nearly as unclear as I did at first, so I’m in favor of landing this to allow some devices with buggy Adreno GPUs to render properly. Thanks for your work!

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: It sounds nice in principle, but how should we detect that? Just checking whether the shader contains the "struct" substring? Iterating over the naga IR? I guess this kind of plays into the "linter" thing we talked about. Not sure if it should be a follow-up.

grebmeg: This would definitely be better as a follow-up. #1659 is the kind of validation check I had in mind, and if you agree with that implementation, we can iterate from there. Feel free to tweak anything as needed.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Done.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Done.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Added the doc comment at the top of both files.

LaurenzV: I think we can remove them once your PR is merged though, since it will be checked in CI anyway.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: I've added such a comment block now. I have to say I'm not a suuper huge fan of having too many such comments because it creates a lot of additional churn when refactoring to also update all comments, but I can understand that it's useful to have an overview as well.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Replaced `raw` with `texel`.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
LaurenzV: Re-added those doc comments.

PR #1620, opened 2026-05-04
vello_sparse_shaders: Flatten fragment shader arguments

LaurenzV:
This is a continuation to #1619 and builds on top of it. With this PR in place, Vello Hybrid runs on Google Pixel 5 and Redmi Note 11 without any problems!

### DIFF
--- a/sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
+++ b/sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
@@ -166,27 +166,31 @@ fn vs_main(
 // Sample a pixel from the original texture.
 // Note: `rel_cord` needs to be positive and must not exceed the width/height of the image
 // that is to be sampled.
-fn sample_original(in: FilterVertexOutput, rel_coord: vec2<f32>) -> vec4<f32> {
-    let src_coord = vec2<u32>(vec2<i32>(in.original_offset) + vec2<i32>(rel_coord));
+fn sample_original(original_offset: vec2<u32>, rel_coord: vec2<f32>) -> vec4<f32> {
+    let src_coord = vec2<u32>(vec2<i32>(original_offset) + vec2<i32>(rel_coord));
     return textureLoad(original_tex, src_coord, 0);
 }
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
```diff
@@ -148,27 +148,31 @@ fn vs_main(
 // Sample a pixel from the original texture.
 // Note: `rel_cord` needs to be positive and must not exceed the width/height of the image
 // that is to be sampled.
-fn sample_original(in: FilterVertexOutput, rel_coord: vec2<f32>) -> vec4<f32> {
```
grebmeg, line 169: Similar to #1619, could you please update the file-level doc with more details about the issue and explicitly forbid its usage? Also, consider adding small tests or simply panicking when a glsl shader is generated.

grebmeg (approved):
I guess this is also causing buggy behaviour in older Adreno drivers? Could you please share a bit more detail about the issue and include it in the file-level doc? Actually, I just realised that since we have two gwsl shaders, it might make more sense to create an additional README file with descriptions of those bugs.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
LaurenzV: Same comment as in the other PR, nice idea in principle but how would you want to detect this? If it's also via the naga IR, it seems like a good follow-up instead where we add both of those lints. I'm just not sure how difficult it will be with regards to edge cases and whether it makes sense to piggyback onto this PR.

grebmeg: The initial discussion [here](https://github.com/linebender/vello/pull/1619#discussion_r3263631906)

LaurenzV: Extended the warning to specifically also warn against the arguments of the fragment shader entry point.

PR #1621, opened 2026-05-04
vello_common: Add tests for lazy buffer pushing

LaurenzV:
Realized that we don't have tests for this; probably good to have!


LaurenzV:
Since coarse rasterization is likely to be superseded soon I'll just close this.

PR #1622, opened 2026-05-05
vello_hybrid: Optimize full viewport rectangles that act as a background

LaurenzV:
In a lot of cases, it is common to draw a full viewport-sized (or lager) rectangle which acts as a background. For example, when rendering a PDF you usually always render them against a white background. This PR adds an optimization such that if this case is detected, instead of actually drawing a fully-sized rectangle we instead piggyback the clear operation that is applied at the start of each render operation to clear the canvas to the given color. While it might always provide a notable benefit since drawing rectangles is relatively cheap, as we have noticed, on low tier devices any saved fragment shader invocations can potentially be very helpful!

### DIFF
--- a/sparse_strips/vello_hybrid/src/scene.rs
+++ b/sparse_strips/vello_hybrid/src/scene.rs
@@ -20,7 +20,7 @@ use vello_common::filter_effects::Filter;
 use vello_common::kurbo::{Affine, BezPath, Rect, Shape, Stroke};
 use vello_common::mask::Mask;
 use vello_common::multi_atlas::AtlasConfig;
-use vello_common::paint::{Paint, PaintType, Tint};
+use vello_common::paint::{Paint, PaintType, PremulColor, Tint};
 #[cfg(feature = "text")]
 use vello_common::peniko::FontData;
 use vello_common::peniko::color::palette::css::BLACK;
@@ -225,6 +225,8 @@ pub struct Scene {
     pub(crate) render_graph: RenderGraph,
     /// Current filter effect applied to individual draw operations.
     filter: Option<Filter>,
+    /// The background color of the scene.
+    pub(crate) background: Option<PremulColor>,
     /// A buffer that stores the strips of path drawing calls that are rendered directly
     /// to the surface, bypassing coarse rasterization.
     pub(crate) fast_strips_buffer: FastStripsBuffer,
@@ -316,6 +318,7 @@ impl Scene {
             layer_id_next: 0,
             render_graph,
             filter: None,
+            background: None,
             fast_strips_buffer: FastStripsBuffer::default(),
             strip_path_mode: StripPathMode::FastOnly,
             coarse_batch_splits: Vec::new(),
@@ -481,6 +484,10 @@ impl Scene {
             return;
         }
 
+        if self.try_set_background_from_rect(rect) {
+            return;
+        }
+
         if self.try_fast_rect(rect) {
             return;
         }
@@ -513,6 +520,37 @@ impl Scene {
         let paint = self.encode_current_paint();
         self.push_fast_rect(bounds, paint);
         true
+    }
+
+    #[inline]
+    fn try_set_background_from_rect(&mut self, rect: &Rect) -> bool {
+        let PaintType::Solid(color) = &self.render_state.paint else {
+            return false;
+        };
+
+        if !self.fast_strips_buffer.commands.is_empty()
+            || !self.constraints.use_default_blending_only()
+            || color.components[3] != 1.0
+            || self.filter.is_some()
+            || self.wide.has_layers()
+            || self.wide.has_coarse_batches()
+            || self.clip_context.get().is_some()
+            || !is_axis_aligned(&self.render_state.transform)
+        {
+            return false;
+        }
+
+        let transformed_rect = self.render_state.transform.transform_rect_bbox(*rect);
+        if transformed_rect.x0 <= 0.0
+            && transformed_rect.y0 <= 0.0
+            && transformed_rect.x1 >= f64::from(self.width)
+            && transformed_rect.y1 >= f64::from(self.height)
+        {
+            self.background = Some(PremulColor::from_alpha_color(*color));
+            return true;
+        }
+
+        false
     }
 
     #[expect(
@@ -929,6 +967,7 @@ impl Scene {
         self.encoded_paints.borrow_mut().clear();
 
         self.render_state.reset();
+        self.background = None;
 
         self.fast_strips_buffer.clear();
         self.strip_path_mode = StripPathMode::FastOnly;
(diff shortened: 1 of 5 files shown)


LaurenzV:
I guess for the blending code path this once again isn't as straight-foward. 😣 Putting this into draft for now.

LaurenzV:
I've repurposed this now to make the detection automatic instead of requiring the user to call a `set_background` method.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -515,6 +522,37 @@ impl Scene {
...
+        };
+
+        if !self.fast_strips_buffer.commands.is_empty()
+            || !self.constraints.use_default_blending_only()
+            || color.components[3] != 1.0
+            || self.filter.is_some()
+            || self.wide.has_layers()
+            || self.wide.has_coarse_batches()
+            || self.clip_context.get().is_some()
+            || !is_axis_aligned(&self.render_state.transform)
```
upsuper, line 538: Theoretically we don't need this condition. It could be any transformation as long as the transformed rect contains all four corners of the viewport. This could be done by either
- applying the inverse transformation to four corners of viewport then calculate whether they are all within this rect, or
- check the transformed rect as a polygon contains all four corners.

(This precise check isn't necessary for our use case though. In our use case this optimization would be able to apply even if it says the transformation must be identity.)

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -515,6 +522,37 @@ impl Scene {
...
+            return false;
+        }
+
+        let transformed_rect = self.render_state.transform.transform_rect_bbox(*rect);
+        if transformed_rect.x0 <= 0.0
+            && transformed_rect.y0 <= 0.0
+            && transformed_rect.x1 >= f64::from(self.width)
+            && transformed_rect.y1 >= f64::from(self.height)
+        {
+            self.background = Some(PremulColor::from_alpha_color(*color));
```
upsuper, line 549: I'm not familiar with how Vello works, but I wonder if we are putting it as a background here, wouldn't we need to also clear existing paint below somehow (or prevent new paint below it to be added)? Otherwise the z-order could be wrong. I couldn't find any such code in this PR. Or is it the `fast_strips_buffer` check ensures no other command exists yet?

(If `fast_strips_buffer` contains all existing paints below, it might be useful we just clear it here as oppose to prevent this optimization from happening?)

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: I guess that's true, but not sure it's worth the trouble. 😄 I already tested it on our side and as soon as default blending scene constraints are enabled, the optimization kicked off on our side.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Hmm, yeah we _might_ be able to add such an optimization, I'd have to think about it a bit more though, so would prefer leaving this for a follow-up. But yes, in this case checking `fast_strips_buffer` is empty and `self.wide.has_coarse_batches()` is enough to determine that nothing has been drawn so far, so for now this should be safe.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -481,6 +484,10 @@ impl Scene {
             return;
         }
 
+        if self.try_set_background_from_rect(rect) {
```
grebmeg, line 487: Could you please elaborate on why you abandoned the idea of having a separate method? I see several benefits in having an explicit API: zero detection logic (the 9-condition guard in `try_set_background_from_rect` disappears entirely), which is important because we want `fill_rect` to stay as fast as possible. It also works under any constraint setting, without requiring a `default_blending_only` gate, and avoids fragility around conditions like “must be the first draw”, “must cover the viewport exactly”, or “must be axis-aligned”. Most importantly, the user can express intent explicitly. The renderer can then simply always read `scene.background` and use it as the clear color.

What we’re effectively trying to do here is detect whether the current rect fills the entire viewport area. I think we should keep this logic simple and let the consumer decide that explicitly instead of relying on heuristics.

grebmeg: Also, there are some duplicated checks between `try_set_background_from_rect` and `try_fast_rect`.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: I think it’d be interesting to do a spike in the future and see what that kind of optimisation would look like, and to have a look at the numbers. Overall though, we already have similar optimisations in place in both areas: in coarse, we discard all commands before the fill that cover the entire wide tile, while in the fast path we rely on front-to-back depth buffer rendering to discard draws behind existing geometry. That said, as I mentioned, it’d still be interesting to see whether the background setup between calls ends up clearing previously pushed commands.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: > Could you please elaborate on why you abandoned the idea of having a separate method?

This was mainly motivated by @upsuper's comment, I was a bit skeptical first but I think I agree now that it's better to not add additional API surface for this and just have this happen automatically instead.

> which is important because we want fill_rect to stay as fast as possible

I very highly doubt that this will have any impact on performance. The method is inlined and in the vast majority of the cases we will already short-circuit on the first condition `!self.fast_strips_buffer.commands.is_empty()` (I purposefully put this check first since it's the most likely one to be true), so this should not matter at all, I think.

> It also works under any constraint setting, without requiring a default_blending_only gate

Unfortunately it doesn't, I tried making it work with that but it caused issue with tests that blend into the root (because then we can't apply the background to the user surface directly), and I wasn't able to find a solution that didn't require changing a lot of code. 

> Most importantly, the user can express intent explicitly.

Yeah, I agree that from this perspective having the `set_background` API is nicer.

LaurenzV: If you think `set_background` is better then let's stick to that. As mentioned, I'm not quite sure if it can be made to work nicely with blending though. I would have to try it. If it doesn't work, we will likely have to shelve this change until we manage to get rid of the scene constraints thing.

grebmeg: I think I’d be in favour of landing this in some form rather than shelving it. I also don’t think these two ideas contradict each other: having `set_background` and automatic background detection.

From a user’s perspective, I like that skia has a `clear` method, which is similar in many ways to ours. That method makes it obvious how to set a background colour and that previous commands are discarded. This is actually why I lean more towards having a similar `clear` method, because it communicates the intent more clearly, including the fact that it resets prior commands.

Moreover, if the consumer has, as I think they should for performance reasons, some data structure representing the scene as AABB hierarchies, then they should already know the background and be able to use that method explicitly. If they do not, we could still rely on background detection, which likely would not degrade performance significantly. However, those 9 conditions do make the code more fragile.

Because of that, I’d recommend splitting this work into two parts:

1. (this PR) Introduce a clear method, which for now we only allow as the first command.
2. (in the future) Add background detection separately, and later support `clear` in arbitrary positions with proper propagation of the background into the wide-tile composition path.

LaurenzV: But how to deal with the fact that it doesn't work with no scene constraints? Should we just fallback to rendering the background as a viewport-sized rectangle then?

grebmeg: I think that could work — our current background optimization for tiles with opaque images should still behave correctly by reseting the tiles' commands to only background color, though probably not for colors with opacity? I’m more thinking about propagating the color when using `clear` to wide tiles — we could essentially just set the background color on each wide tile and clear the commands?

LaurenzV:
@grebmeg I ran a test on my very slow tablet. I drew a viewport-width solid white rectangle, and a partially-transparent half-height yellow rectangle on top of that. With current main, I'm getting around 35-40FPS. With the background optimization, I'm getting around 50-55FPS. _However_, I believe that a big reason for this is the uber shader, and I don't think it's unlikely that the difference might not be as high anymore once we move to specialized shaders.

WDYT about shelving this PR for now so we can focus on more immediate stuff and then revisiting whether such an optimization would be worth it once we have specialized shaders? Or do you think, given the immediate benefit this seems to have, it's worth pursuing this now?

LaurenzV:
I think this will be superseded by https://github.com/linebender/vello/pull/1783. Using that method, clients can set their own color to clear the canvas to. While it does sound interesting to have automatic detection for this, I think from an API perspective it's cleaner to leave this to the configuration of the canvas clear behavior instead.

PR #1623, opened 2026-05-06
sparse_strips: Add `ColorMatrix` filter

waywardmonkeys:
This adds `ColorMatrix` filter support to the sparse-strip renderers.

* Adds CPU fast paths for:
  *  identity matrix no-op
  * tracking `may_have_transparency` during the color-matrix pass
  * opaque premultiplied input pixels
  * `/ 255.0` normalization via reciprocal multiply

## Performance Notes

I profiled a temporary release-mode harness applying `matrices::SEPIA` repeatedly over a 1024x1024 pixmap for 20 seconds.

After the reciprocal multiply change:

* Opaque input: `3610` full-pixmap passes
* Translucent input: `3233` full-pixmap passes

### DIFF
--- a/sparse_strips/vello_cpu/src/filter/color_matrix.rs
+++ b/sparse_strips/vello_cpu/src/filter/color_matrix.rs
@@ -1,0 +1,147 @@
+// Copyright 2026 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Color matrix filter implementation.
+
+use super::FilterEffect;
+use super::pixel::{clamp_unit, premul_rgba8_to_straight_f32, straight_f32_to_premul_rgba8};
+use crate::layer_manager::LayerManager;
+use vello_common::filter::color_matrix::ColorMatrix;
+use vello_common::peniko::color::PremulRgba8;
+use vello_common::pixmap::Pixmap;
+
+impl FilterEffect for ColorMatrix {
+    fn execute_lowp(&self, pixmap: &mut Pixmap, _layer_manager: &mut LayerManager) {
+        apply_color_matrix(pixmap, &self.matrix);
+    }
+
+    fn execute_highp(&self, pixmap: &mut Pixmap, _layer_manager: &mut LayerManager) {
+        apply_color_matrix(pixmap, &self.matrix);
+    }
+}
+
+fn apply_color_matrix(pixmap: &mut Pixmap, matrix: &[f32; 20]) {
+    for pixel in pixmap.data_mut() {
+        *pixel = apply_color_matrix_to_pixel(*pixel, matrix);
+    }
+
+    pixmap.recompute_may_have_transparency();
+}
+
+fn apply_color_matrix_to_pixel(pixel: PremulRgba8, matrix: &[f32; 20]) -> PremulRgba8 {
+    let [r, g, b, a] = premul_rgba8_to_straight_f32(pixel);
+
+    let out_r = apply_row(matrix, 0, r, g, b, a);
+    let out_g = apply_row(matrix, 5, r, g, b, a);
+    let out_b = apply_row(matrix, 10, r, g, b, a);
+    let out_a = apply_row(matrix, 15, r, g, b, a);
+
+    straight_f32_to_premul_rgba8(out_r, out_g, out_b, out_a)
+}
+
+fn apply_row(matrix: &[f32; 20], offset: usize, r: f32, g: f32, b: f32, a: f32) -> f32 {
+    clamp_unit(
+        matrix[offset] * r
+            + matrix[offset + 1] * g
+            + matrix[offset + 2] * b
+            + matrix[offset + 3] * a
+            + matrix[offset + 4],
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use vello_common::filter_effects::matrices;
+
+    #[test]
+    fn identity_preserves_premultiplied_pixel() {
+        let pixel = PremulRgba8 {
+            r: 128,
+            g: 64,
+            b: 32,
+            a: 128,
+        };
+
+        assert_eq!(
+            apply_color_matrix_to_pixel(pixel, &matrices::IDENTITY),
+            pixel
+        );
+    }
+
+    #[test]
+    fn grayscale_uses_unpremultiplied_color_channels() {
+        let pixel = PremulRgba8 {
+            r: 128,
+            g: 0,
+            b: 0,
+            a: 128,
+        };
+
+        assert_eq!(
+            apply_color_matrix_to_pixel(pixel, &matrices::GRAYSCALE),
+            PremulRgba8 {
+                r: 27,
+                g: 27,
+                b: 27,
+                a: 128,
+            }
+        );
+    }
+
+    #[test]
+    fn matrix_offsets_can_create_color_from_transparent_black() {
+        let matrix = [
+            0.0, 0.0, 0.0, 0.0, 0.5, //
+            0.0, 0.0, 0.0, 0.0, 0.25, //
+            0.0, 0.0, 0.0, 0.0, 0.0, //
+            0.0, 0.0, 0.0, 0.0, 0.5,
+        ];
+
+        assert_eq!(
+            apply_color_matrix_to_pixel(PremulRgba8::from_u32(0), &matrix),
+            PremulRgba8 {
+                r: 64,
+                g: 32,
+                b: 0,
+                a: 128,
+            }
+        );
+    }
+
+    #[test]
+    fn execution_updates_pixmap_transparency_flag() {
+        let mut pixmap = Pixmap::from_parts_with_opacity(
+            alloc::vec![PremulRgba8 {
+                r: 255,
+                g: 0,
+                b: 0,
+                a: 255,
+            }],
+            1,
+            1,
+            false,
+        );
+        let mut layer_manager = LayerManager::new();
+        let matrix = [
+            1.0, 0.0, 0.0, 0.0, 0.0, //
+            0.0, 1.0, 0.0, 0.0, 0.0, //
+            0.0, 0.0, 1.0, 0.0, 0.0, //
+            0.0, 0.0, 0.0, 0.0, 0.5,
+        ];
+        let filter = ColorMatrix::new(matrix);
+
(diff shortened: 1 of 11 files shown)


waywardmonkeys:

It did some of the boring stuff but also wrote benchmarks, ran them under a profiler, disassembled code to verify certain things were happening, and carried out a number of experiments to see if some stuff would result in appreciable performance improvements or not.

waywardmonkeys:
After talking to @raphlinus, Codex and I did another optimization that was amended into the 3rd commit in this series to avoid a lot of the un-premultiply / re-premultiply work with compatible matrices.

The fast path applies matrices directly in premultiplied space when they:

* preserve alpha exactly
* only mix RGB into RGB
* have no RGB offsets

That covers `GRAYSCALE` and `SEPIA`, but not things like `ALPHA_TO_BLACK`.

Benchmark using the same 20s 1024x1024 SEPIA release harness:

| Path | Before | After |
| --- | ---: | ---: |
| Opaque | 3610 passes | 5705 passes |
| Translucent | 3233 passes | 5706 passes |

waywardmonkeys:
Some of this code didn't really belong in Vello, so I've gone and created a new crate in the `color` repo, `color_operations`: https://github.com/linebender/color/pull/221

I've got a version of this set of changes that builds on that instead, and I'll update this PR with it later today.

grebmeg:
> Some of this code didn't really belong in Vello, so I've gone and created a new crate in the `color` repo, `color_operations`: [linebender/color#221](https://github.com/linebender/color/pull/221)
> 
> I've got a version of this set of changes that builds on that instead, and I'll update this PR with it later today.

Hey @waywardmonkeys, sorry for the slow review. Now that I’ve got review capacity back, I’m ready to take a look. Before I start, just wanted to confirm whether this PR is ready for review or if you’re still planning to make more updates to it?

waywardmonkeys:
@grebmeg I think most of this apart from the color math is useful to take a look at and make sure there's no objections to it overall.

The color math here is fine, but really belongs elsewhere, hence https://github.com/linebender/color/pull/221.  Unfortunately, Tom brought up some good points over there and maybe some people who care about this topic should have a meeting to discuss things.

If you'd like to set up such a meeting with me, please feel free to do so.

I won't need much notice about having it, as long as I'm available.

waywardmonkeys:
FYI: @tomcur  (See above)

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
```diff
@@ -114,7 +116,17 @@ fn load_filter_data(texel_offset: u32) -> GpuFilterData {
     let t0 = textureLoad(filter_data, vec2((texel_offset     ) % w, (texel_offset     ) / w), 0);
     let t1 = textureLoad(filter_data, vec2((texel_offset + 1u) % w, (texel_offset + 1u) / w), 0);
     let t2 = textureLoad(filter_data, vec2((texel_offset + 2u) % w, (texel_offset + 2u) / w), 0);
-    return GpuFilterData(array(t0.x, t0.y, t0.z, t0.w, t1.x, t1.y, t1.z, t1.w, t2.x, t2.y, t2.z, t2.w));
+    let t3 = textureLoad(filter_data, vec2((texel_offset + 3u) % w, (texel_offset + 3u) / w), 0);
+    let t4 = textureLoad(filter_data, vec2((texel_offset + 4u) % w, (texel_offset + 4u) / w), 0);
+    let t5 = textureLoad(filter_data, vec2((texel_offset + 5u) % w, (texel_offset + 5u) / w), 0);
```
grebmeg, line 121: With adding three more texture loads, I think it’s time to address [this comment](https://github.com/linebender/vello/blob/f1da2a22c25bbca11b48aed8d1cdb9000baa7223/sparse_strips/vello_hybrid/src/filter.rs#L212-L215) and introduce a variable-stride index for filter data. This would be similar to how sparse strips are handled in `encoded_paints`, where `paint_idxs` stores the texel offset for where the current filter starts in the texture. Would it be possible to do this, if not in this PR, then at least as a follow-up?

sparse_strips/vello_cpu/src/filter/color_matrix.rs
```diff
@@ -0,0 +1,316 @@
...
+            may_have_transparency |= transformed.a != 255;
+            *pixel = transformed;
+        }
+    } else {
+        for pixel in pixmap.data_mut() {
+            let transformed = apply_color_matrix_to_pixel(*pixel, matrix);
+            may_have_transparency |= transformed.a != 255;
+            *pixel = transformed;
+        }
+    }
```
grebmeg, line 47: Color matrix is a great candidate for SIMD here 😅 We haven’t wired filters up with SIMD yet, but it’d be nice to do at some point since it could bring significant performance improvements. Just sharing my thoughts, it might be worth adding a TODO somewhere so we can revisit this later.

grebmeg: Actually, I just realized we also haven’t wired up the faster non-spatial filter path at all, things like color matrix, opacity, flood, basically filters that only require per-pixel work.

I don’t think that blocks this PR since we already have the same situation with `Flood`, but I wanted to highlight it in case you’re aiming for significantly better performance. Right now those filters still go through the spatial filter path, which allocates a pixmap and does some additional work that could be avoided entirely. This could likely be handled much more efficiently [here](https://github.com/linebender/vello/blob/f1da2a22c25bbca11b48aed8d1cdb9000baa7223/sparse_strips/vello_cpu/src/fine/mod.rs#L580-L587) in vello_cpu fine.

sparse_strips/vello_cpu/src/filter/color_matrix.rs
```diff
@@ -0,0 +1,316 @@
...
+    clamp_unit(
+        matrix[offset] * r
+            + matrix[offset + 1] * g
+            + matrix[offset + 2] * b
+            + matrix[offset + 3] * a
+            + matrix[offset + 4],
+    )
+}
+
+fn is_premul_compatible_matrix(matrix: &[f32; 20]) -> bool {
```
grebmeg, line 93: Would it be worth inlining this and the functions above as well?

sparse_strips/vello_sparse_tests/tests/filter.rs
```diff
@@ -83,6 +83,24 @@ fn filter_offset_nested(ctx: &mut impl Renderer) {
     ctx.pop_layer();
 }
 
+#[vello_test(skip_multithreaded, width = 120, height = 80, hybrid_tolerance = 1)]
+fn filter_color_matrix_full_sepia(ctx: &mut impl Renderer) {
+    let filter = Filter::from_primitive(FilterPrimitive::ColorMatrix {
+        matrix: matrices::SEPIA,
```
grebmeg, line 89: Could you add at least one snapshot test for an alpha-affecting matrix (e.g. `matrices::ALPHA_TO_BLACK`) or a matrix with non-zero offsets (the 5th column in a row)?

sparse_strips/vello_cpu/src/filter/color_matrix.rs
```diff
@@ -0,0 +1,316 @@
...
+        && matrix[8] == 0.0
+        && matrix[9] == 0.0
+        && matrix[13] == 0.0
+        && matrix[14] == 0.0
+        && matrix[15] == 0.0
+        && matrix[16] == 0.0
+        && matrix[17] == 0.0
+        && matrix[18] == 1.0
+        && matrix[19] == 0.0
+}
```
grebmeg, line 105:
```suggestion
fn is_premul_compatible_matrix(matrix: &[f32; 20]) -> bool {
    let row_r = &matrix[0..5];
    let row_g = &matrix[5..10];
    let row_b = &matrix[10..15];
    let row_a = &matrix[15..20];
    // RGB rows must not depend on alpha (col 3) or carry a constant offset (col 4):
    // applying the matrix directly to premultiplied channels then yields the same
    // result as unpremultiply -> matrix -> premultiply.
    let rgb_alpha_independent = row_r[3] == 0.0
        && row_r[4] == 0.0
        && row_g[3] == 0.0
        && row_g[4] == 0.0
        && row_b[3] == 0.0
        && row_b[4] == 0.0;
    // And alpha must be preserved exactly.
    let alpha_preserved = row_a[0] == 0.0
        && row_a[1] == 0.0
        && row_a[2] == 0.0
        && row_a[3] == 1.0
        && row_a[4] == 0.0;
    rgb_alpha_independent && alpha_preserved
}
```

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
```diff
@@ -311,6 +323,34 @@ fn convolve(
     return color;
 }
 
+fn color_matrix_row(data: GpuFilterData, base: u32, color: vec4<f32>) -> f32 {
```
grebmeg, line 326: Just wanted to pay you attention that unfortunately we had to change the way how we write wgsl shaders and avoid using structs (PRs #1619 #1620), so this's likly why you have merge conflicts.

grebmeg (approved):
I think this PR is in pretty good shape overall. It might be worth adding the optimizations I mentioned [here](https://github.com/linebender/vello/pull/1623#discussion_r3278530050) so we can avoid the additional memory overhead for each filter.

There are also a few other things we haven’t wired up yet, but hopefully we’ll be able to tackle those over time as well.

Thanks for introducing the color filter support! 🎉

grebmeg:
> The color math here is fine, but really belongs elsewhere, hence [linebender/color#221](https://github.com/linebender/color/pull/221). Unfortunately, Tom brought up some good points over there and maybe some people who care about this topic should have a meeting to discuss things.
> 
> If you'd like to set up such a meeting with me, please feel free to do so.
> 
> I won't need much notice about having it, as long as I'm available.

I have a feeling it might be useful to start the discussion on Zulip first, so we can all better explore the problem space together?

I think this is really important, especially since in the future we’ll likely need to consider not just sRGB. I’m not even sure how much of the discussion so far has touched on that. I also honestly don’t know how actively the `color` crate is being developed or what direction it’s heading in, so I’m definitely keen to hear more thoughts on it.

waywardmonkeys:
I don't currently know what to do about the color operations code.

grebmeg:
I think this PR is quite obsolete now, and I’m curious what the blocker was for merging it. Should we close it for now and reimplement it on top of the latest `main` when needed?

PR #1625, opened 2026-05-06
vello_hybrid: fast-path premultiplied color matrices

waywardmonkeys:
This adds a `vello_hybrid` shader fast path for color matrices that can be applied directly in premultiplied-alpha space.

For RGB-only, alpha-preserving matrices with no RGB offsets, `ColorMatrix` no longer needs to unpremultiply each sampled pixel, apply the matrix in straight-alpha space, then premultiply again. Instead, hybrid marks those matrices in the encoded filter header and the shader applies the RGB rows directly to premultiplied RGB, clamps the result to `[0, alpha]`, and preserves alpha.

This covers matrices such as grayscale and sepia while keeping the existing general path for matrices with RGB offsets, alpha reads, or alpha changes.

### DIFF
--- a/sparse_strips/vello_common/src/filter/color_matrix.rs
+++ b/sparse_strips/vello_common/src/filter/color_matrix.rs
@@ -1,0 +1,21 @@
+// Copyright 2026 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! The color matrix filter.
+
+/// Matrix-based color transformation filter.
+///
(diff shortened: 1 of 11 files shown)


waywardmonkeys:

waywardmonkeys:
This is a draft as it builds upon PR #1623, adding an optimization present there for the CPU renderer to the Hybrid renderer.

grebmeg:
I think this PR is quite obsolete now, and I’m curious what the blocker was for merging it. Should we close it for now and reimplement it on top of the latest `main` when needed?

PR #1626, opened 2026-05-07
Add method to mark image override textures as dirty

raphlinus:
Followup to #1558, which caches images. There's a risk that we'll render stale cached date from dynamic external textures, so we provide a method to mark those as dirty.

Authored with substantial assistance from Claude Code.

### DIFF
--- a/vello/src/lib.rs
+++ b/vello/src/lib.rs
@@ -544,6 +544,15 @@ impl Renderer {
         }
     }
 
+    /// Marks `image` as dirty in the atlas cache, so its override texture will be re-copied on
+    /// the next render.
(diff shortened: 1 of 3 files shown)


nicoburns (approved):
I have tested this with Blitz, and can confirm that:
- Latest `main` never renders updates to override textures (unless you resize the window)
- This PR fixes that issue.

raphlinus:
Bruce has mentioned there are other problems, which I am willing to believe. I haven't analyzed it carefully.

nicoburns:
Superceded by https://github.com/linebender/vello/pull/1638

PR #1627, opened 2026-05-07
[hybrid]: Truncate gradient luts after upload

taj-p:
Like the WGPU implementation, shown below, we need to truncate the gradient luts:

https://github.com/linebender/vello/blob/3a021344a27f2c00553efcbc88a1f833edcbbbed/sparse_strips/vello_hybrid/src/render/wgpu.rs#L2394

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -1539,6 +1539,7 @@ impl WebGlPrograms {
 
         // Take ownership of the luts to avoid copying, then resize for texture padding.
         let mut luts = gradient_cache.take_luts();
+        let old_luts_len = luts.len();
         luts.resize(total_capacity, 0);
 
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1560,7 +1561,8 @@ impl WebGlPrograms {
         )
         .unwrap();
 
-        // Restore the luts back to the cache.
+        // Restore the luts back to the cache
```
```suggestion
        // Restore the luts back to the cache.
```

grebmeg (approved):

LaurenzV:
I'm curious how this bug surfaced? Is there a way of having a test for this?

taj-p:
> I'm curious how this bug surfaced? Is there a way of having a test for this?

I couldn't think of one that didn't expose internal information. I don't believe this is a cause for any bugs AFAICT. I was inspecting this code while triaging a different issue

LaurenzV:
Ah, I see!

PR #1628, opened 2026-05-07
glifo: Add font embolden

jrmoulton:
This PR bumps kurbo to latest main although it is currently a patch since peniko depends on the released kurbo 0.13. I can open a PR to peniko if we need to bump kurbo there first. 

This PR adds support for font embolden by getting the value through to the new kurbo `expand_path`. The default embolden value is (0, 0). The chosen miter limit is 4.0 and the tolerance is 0.1. I'm not sure if these are the best values to use or if we should maybe make them configurable but these seem to be reasonable defaults. 

@nicoburns

### DIFF
--- a/glifo/src/atlas/key.rs
+++ b/glifo/src/atlas/key.rs
@@ -63,6 +63,10 @@ pub struct GlyphCacheKey {
     pub context_color: AlphaColor<Srgb>,
     /// Pre-packed context color (premultiplied RGBA8 as u32) used in Hash/Eq.
     pub context_color_packed: u32,
+    /// Synthetic embolden amount. Only non-zero for outline glyphs.
+    pub embolden_x_bits: u64,
+    /// Synthetic embolden amount. Only non-zero for outline glyphs.
+    pub embolden_y_bits: u64,
     /// Variation coordinates for variable fonts.
     pub var_coords: SmallVec<[NormalizedCoord; 4]>,
 }
@@ -82,6 +86,8 @@ impl GlyphCacheKey {
         fractional_x: f32,
         context_color: AlphaColor<Srgb>,
         context_color_packed: u32,
+        embolden_x: f64,
+        embolden_y: f64,
         var_coords: &[NormalizedCoord],
     ) -> Self {
         Self {
@@ -93,6 +99,8 @@ impl GlyphCacheKey {
             subpixel_x: quantize_subpixel(fractional_x),
             context_color,
             context_color_packed,
+            embolden_x_bits: embolden_x.to_bits(),
+            embolden_y_bits: embolden_y.to_bits(),
             var_coords: SmallVec::from_slice(var_coords),
         }
     }
@@ -112,6 +120,8 @@ impl Hash for GlyphCacheKey {
         self.hinted.hash(state);
         self.subpixel_x.hash(state);
         self.context_color_packed.hash(state);
+        self.embolden_x_bits.hash(state);
+        self.embolden_y_bits.hash(state);
     }
 }
 
@@ -125,6 +135,8 @@ impl PartialEq for GlyphCacheKey {
             && self.size_bits == other.size_bits
             && self.hinted == other.hinted
             && self.context_color_packed == other.context_color_packed
+            && self.embolden_x_bits == other.embolden_x_bits
+            && self.embolden_y_bits == other.embolden_y_bits
     }
 }
 
@@ -196,15 +208,16 @@ mod tests {
     #[test]
     fn test_key_equality() {
         let packed = pack_color(BLACK);
-        let key1 = GlyphCacheKey::new(1, 0, 42, 16.0, true, 0.3, BLACK, packed, &[]);
-        let key2 = GlyphCacheKey::new(1, 0, 42, 16.0, true, 0.3, BLACK, packed, &[]);
+        let key1 = GlyphCacheKey::new(1, 0, 42, 16.0, true, 0.3, BLACK, packed, 0.0, 0.0, &[]);
+        let key2 = GlyphCacheKey::new(1, 0, 42, 16.0, true, 0.3, BLACK, packed, 0.0, 0.0, &[]);
         assert_eq!(key1, key2);
     }
 
     #[test]
     fn test_outline_colr_bitmap_keys_never_collide() {
         let packed = pack_color(BLACK);
-        let outline_key = GlyphCacheKey::new(1, 0, 42, 16.0, false, 0.0, BLACK, packed, &[]);
+        let outline_key =
+            GlyphCacheKey::new(1, 0, 42, 16.0, false, 0.0, BLACK, packed, 0.0, 0.0, &[]);
         let colr_key = GlyphCacheKey {
             font_id: 1,
             font_index: 0,
@@ -214,6 +227,8 @@ mod tests {
             subpixel_x: SUBPIXEL_COLR,
             context_color: BLACK,
             context_color_packed: packed,
+            embolden_x_bits: 0,
+            embolden_y_bits: 0,
             var_coords: SmallVec::new(),
         };
         let bitmap_key = GlyphCacheKey {
@@ -225,6 +240,8 @@ mod tests {
             subpixel_x: SUBPIXEL_BITMAP,
             context_color: BLACK,
             context_color_packed: packed,
+            embolden_x_bits: 0,
+            embolden_y_bits: 0,
             var_coords: SmallVec::new(),
         };
         assert_ne!(outline_key, colr_key);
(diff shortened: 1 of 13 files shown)


jrmoulton:
I've also added support in vello classic

waywardmonkeys:
@jrmoulton We want to make sure we're happy with this ... have some screenshots again? and then we'd do the publish of `kurbo`, `peniko`, update this PR, land it, and prep for a Vello family release.

jrmoulton:
I'll get some screenshots posted

jrmoulton:
I've exposed the join type, miter limit, and tolerance.

glifo/src/atlas/key.rs
```diff
@@ -63,6 +64,16 @@ pub struct GlyphCacheKey {
...
+    /// Synthetic embolden amount. Only non-zero for outline glyphs.
+    pub embolden_x_bits: u64,
+    /// Synthetic embolden amount. Only non-zero for outline glyphs.
+    pub embolden_y_bits: u64,
+    /// Join style for synthetic embolden. Only meaningful for outline glyphs.
+    pub embolden_join_bits: u8,
+    /// Miter limit for synthetic embolden. Only meaningful for outline glyphs.
+    pub embolden_miter_limit_bits: u64,
+    /// Tolerance for synthetic embolden. Only meaningful for outline glyphs.
+    pub embolden_tolerance_bits: u64,
```
LaurenzV: Could we not cast to f32 first and then store as u32 instead of u64? This would help with keeping the memory footprint of the glyph cache key smaller.

glifo/src/atlas/key.rs
```diff
@@ -196,15 +235,60 @@ mod tests {
     #[test]
     fn test_key_equality() {
         let packed = pack_color(BLACK);
-        let key1 = GlyphCacheKey::new(1, 0, 42, 16.0, true, 0.3, BLACK, packed, &[]);
-        let key2 = GlyphCacheKey::new(1, 0, 42, 16.0, true, 0.3, BLACK, packed, &[]);
+        let key1 = GlyphCacheKey::new(
```
LaurenzV, line 245: I think this constructor is even more unwieldy now than it already was before now, would it make sense to group all emboldening-related keys into a custom struct?

glifo/src/glyph.rs
```diff
@@ -722,6 +757,30 @@ impl<'a, B> GlyphRunBuilder<'a, B> {
...
+        self.run.font_embolden_miter_limit = miter_limit;
+        self
+    }
+
+    /// Set the tolerance for synthetic embolden.
+    pub fn font_embolden_tolerance(mut self, tolerance: f64) -> Self {
+        self.run.font_embolden_tolerance = tolerance;
+        self
+    }
+
```
LaurenzV: Similarly, would it make sense to just create a `FontEmboldenSettings` struct and just a single setter for it?

LaurenzV:
Also, I think it would be great to have one or two tests for this in `vello_sparse_tests`.

glifo/src/atlas/key.rs
jrmoulton: I tried it but it seemed to be an awkward fit to construct the a struct just as a grouping for the cache key. It might be better to remove the new function and directly construct the key struct. That would be less unwieldy and ambiguous but since the new function was already there I opted to use it.

glifo/src/glyph.rs
jrmoulton: I don't think so. This already has a fluent api for setting different properties. The naming is a bit unfortunate but I don't think adding a new struct with probably another fluent api would be better.

jrmoulton: But I can do that if we decide to

LaurenzV: I think it would be my preference, but yeah let's see what other's say.

>  This already has a fluent api for setting different properties.

Yes but they are all different properties, while the font embolden ones all refer to the same feature basically. Just like we also have a single `Stroke` struct instead of methods like `set_stroke_width`, `set_stroke_miter_limit`, etc.

glifo/src/glyph.rs
```diff
@@ -1563,9 +1651,30 @@ struct OutlineKey {
...
+}
+
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "Cache keys intentionally store embolden parameters at f32 precision."
+)]
+fn f32_bits(value: f64) -> u32 {
+    (value as f32).to_bits()
+}
+
```
taj-p, line 1673: Should these be inline always?

vello_encoding/src/glyph_cache.rs
```diff
@@ -191,8 +210,25 @@ impl GlyphCacheSession<'_> {
         } else {
             DrawSettings::unhinted(self.size, self.coords)
         };
-        outline.draw(draw_settings, &mut path).ok()?;
-        if path.finish(false) == 0 {
+        let n_path_segments = if self.embolden != Diagonal2::new(0.0, 0.0) {
+            let mut path = BezPathOutline(BezPath::new());
```
taj-p: Should we maintain a reusable BezPath on the session so that we don't pay this allocation for every glyph?

glifo/src/glyph.rs
```diff
@@ -1754,6 +1872,16 @@ impl<'a> OutlineCacheSession<'a> {
 
                 drawing_buf.reuse();
                 outline_glyph.draw(draw_settings, &mut drawing_buf).unwrap();
+                if embolden != Diagonal2::new(0.0, 0.0) {
+                    drawing_buf.path = kurbo::expand_path(
+                        &drawing_buf.path,
```
taj-p, line 1870: It's unfortunate that the new method doesn't have a version that doesn't allocate. 

It looks like `expand_path` always allocates a new `BezPath` internally - so we're paying an allocation per glyph when zero is possible.

glifo/src/glyph.rs
taj-p: If you're looking for another opinion, I think I prefer the separate struct version rather than inclining because it feels like how our other API is designed (collect into structs rather than inlining and flattening fields)

taj-p (commented):
So good!!! Left some comments 🙏

Would be nice to have some tests as Laurenz suggested above

glifo/src/atlas/key.rs
jrmoulton: @taj-p @LaurenzV any more thoughts on this? I've added a new struct to group the font embolden on the glyph run builder but we can't use the same struct in the cache key because the cache key now uses f32. New struct or leave it as is?

jrmoulton: actually I think I can just move the handling into the new function. that should work fine

jrmoulton: Yeah I was overthinking that

jrmoulton:
I added a snapshot test

glifo/src/glyph.rs
taj-p: I had a go of adding a re-useable context here https://github.com/linebender/kurbo/pull/582

nicoburns:
@jrmoulton Git is complaining about this branch for me because the snapshot is not in git-lfs. I think you need to install Git LFS and then recommit the snapshot.

nicoburns:
This is looking decent to me:

Left: Firefox on macOS. Right: Blitz on macOS with Vello Classic + this PR.

Embolden amount is computed as:

```rust
let embolden = kurbo::Vec2::new((0.015125 * font_size).min(0.3), (0.0121 * font_size).min(0.3));
```

where `font_size` is the font size in *logical* pixels.

Screenshot 2026-05-10 at 23 53 23

Unless I'm missing something, this PR doesn't seem to wire through embolden into the `vello_hybrid` or `vello_cpu` APIs, so I am unable to test there.

Blitz PR for anyone who wants to try this out: https://github.com/DioxusLabs/blitz/pull/428
Command to run is `cargo run -rp browser`

jrmoulton:
> Unless I'm missing something, this PR doesn't seem to wire through embolden into the vello_hybrid or vello_cpu APIs, so I am unable to test there.

The glyph run builder is provided by glifo. You can look at the snapshot test that uses vello cpu. 

> where font_size is the font size in logical pixels.

I had thought it was logical pixels but when I was testing I found it to be device pixels. I could have been wrong though

nicoburns:
> I had thought it was logical pixels but when I was testing I found it to be device pixels. I could have been wrong though

To be clear: I'm applying the formula to the unscaled font size, and applying the result as an embolden amount in physical pixels.

jrmoulton:
@nicoburns I think I've fixed the git lfs issue

nicoburns (approved):
@jrmoulton This needs the kurbo crates.io patch removing, and an explicit dependency on `kurbo` v0.13.1 in both `vello_encoding` and `glifo` but has otherwise been judged to be good to land.

nicoburns:
CI issues seems to be https://github.com/Swatinem/rust-cache/issues/341

PR #1629, opened 2026-05-08
glifo: Always draw non-hinted glyphs at upem size

LaurenzV:
# Problem

Right now, we _always_ draw glyph outlines using skrifa at the given font size we are currently rendering. This causes two problems:
1) It is expensive, because when the user zooms we keep redrawing the glyph outlines at increasingly larger font sizes, meaning that we don't get any cache hits.
2) It also causes cache thrashing, since we try to cache the same glyph outlines at all kinds of different sizes.

# Solution

The insight is that in the common case of having an unhinted and not-stroked glyph, it should be totally fine to just always outline the glyph at upem size and then scale it to the appropriate font size with a simple transform. I believe internally skrifa does something slightly different to keep compatibility with freetype, but in general this shouldn't cause any problems apart from small pixel differences (@dfrg can you confirm?).

# Performance

When continuously zooming on our vello text scene, I get the following:

Before:
image

After:
image

As you can see, the skrifa outlining part completely disappears because each glyph is only outlined once and everything can therefore stay in the cache. The glyph cache maintenance part also goes away since there is no cache thrashing anymore.


LaurenzV:
Hmm, it is a bit unfortunate that basically all tests had to be regenerated. I would really be curious to hear from @dfrg first if he sees any issue with doing this.

dfrg:
No issues at all. In fact, I intended to implement this optimization for unhinted glyphs in vello classic but never got around to it.

There are some older fonts where hinting is required for correct rasterization so I would keep that in mind if the aim is production quality. See googlefonts/fontations#1303

edit: I suppose “no issues” was a bit strong. More like “one minor issue…”

LaurenzV:
Thanks a lot for confirming! Right, I remember that, though the issue is orthogonal to this PR. This could be fixed in a follow-up if we want to.

glifo/src/glyph.rs
```diff
@@ -563,6 +568,7 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone> GlyphRunRenderer<'a, 'b, Gl
         buffer: f32,
     ) -> impl Iterator<Item = Rect> + 'c {
         let font_ref = self.prepared_run.font.as_skrifa();
+        let upem: f32 = font_ref.head().map(|h| h.units_per_em()).unwrap().into();
```
grebmeg: `upem` and the cache-size decision are pure functions of the prepared run, but they're computed independently in `draw_glyphs` and here. Storing the resolved `cache_size` (and ideally `upem`) on `PreparedGlyphRun` during `prepare_glyph_run` would prevent these two consumers from drifting and remove the redundant `font_ref.head()` lookup.

glifo/src/renderer.rs
```diff
@@ -358,7 +365,7 @@ fn insert_and_render_outline(
         return CacheResult::UnsupportedTransform;
     }
 
-    let bounds = path.bounding_box();
+    let bounds = path.bounding_box().scale_from_origin(scale);
```
grebmeg: `Arc<BezPath>` is now shared across many `font_sizes`, so we'll keep recomputing the same precise `bbox` on every atlas miss during zoom. Since `OutlineEntry` is already keyed once per upem-space path, caching path.`bounding_box()` (or even just exposing the existing rough bbox) on the entry would let this site become a `bounds.scale_from_origin(scale)` over a precomputed value. Maybe somethign like the next would help to achieve that:

```rust
// In create_outline_glyph:
let bbox = cached.path.bounding_box(); // fold into OutlineEntry once
GlyphType::Outline(GlyphOutline {
    path: Arc::clone(cached.path),
    bbox, // upem-space, scale at use site
    scale,
})
```

glifo/src/glyph.rs
```diff
@@ -951,9 +961,45 @@ fn create_outline_glyph<'a>(
...
+}
+
+impl GlyphScaleProperties {
+    fn new(
+        draw_font_size: f32,
+        upem: f32,
+        hinting_instance: Option<&HintingInstance>,
+        style: Style,
+    ) -> Self {
+        if hinting_instance.is_some() || style == Style::Stroke {
```
grebmeg: [nit]: Maybe consider decoupling this from `HintingInstance` and just using `hinted: bool` instead.

glifo/src/glyph.rs
```diff
@@ -331,6 +333,8 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone> GlyphRunRenderer<'a, 'b, Gl
             // On a miss we keep both for reuse in the outline branch below.
             let outline_transform =
                 calculate_outline_transform(glyph, draw_props, hinting_instance);
+            let scale_props =
```
grebmeg: Maybe consider moving it closer to where it’s used otherwise it's computed for every glyph in the loop?

glifo/src/glyph.rs
```diff
@@ -617,7 +626,7 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone> GlyphRunRenderer<'a, 'b, Gl
                 glyph.id,
                 self.prepared_run.font.data.id(),
                 self.prepared_run.font.index,
-                draw_props.font_size,
+                scale_props.cache_size,
```
grebmeg, line 699: If you pass `upem` here as well, I think you can communicate the intent a bit more clearly when building `draw_settings`. Essentially, you could write something like the following:

```rust
let size_param = if scale_props.cache_size == upem {
    Size::unscaled()
} else {
    Size::new(scale_props.cache_size)
};
DrawSettings::unhinted(size_param, var_key.0)
```

grebmeg (approved):
It’s a nice improvement, @LaurenzV! Well done! 👏 Thanks for highlighting the areas to focus on in the screenshot 🙏  How much improvement are we seeing in the numbers here?

LaurenzV:
> How much improvement are we seeing in the numbers here?

I don't remember the numbers, but seems like ~20-25%? It's definitely a big improvement though!

Also thanks for the comments, will try to address them soon!

glifo/src/renderer.rs
LaurenzV: Makes total sense, thanks!

glifo/src/glyph.rs
LaurenzV: Hmm I've not changed this for now because we already have soo many arguments in that function. Feel free to let me know if you feel strongly about it and then I'll do it in a follow-up.

glifo/src/glyph.rs
LaurenzV: Done!

glifo/src/glyph.rs
LaurenzV: Done!

glifo/src/glyph.rs
LaurenzV: Storing the `cache_size` on `PreparedGlyphRun` is not quite possible because at that point we don't know yet whether we are filling or stroking (unless we precompute both, which seems a bit ugly to me too 😄).

Will precompute the upem though, as you suggested.

PR #1630, opened 2026-05-08
glifo: Add basic sanity unit tests for atlas caching

LaurenzV:
As was evidenced in #1605, it is pretty easy to make changes to the caching logic and then accidentally break the whole caching mechanism.

To hopefully provide some safe-guarding against this, this PR adds 6 unit tests, each ensuring that the different configurations for outline/bitmap/COLR glyphs result in the expected usage of the cache. In case atlas caching is disabled, the cache should stay untouched. In case it's enabled, the glyphs should actually end up landing in the cache.

### DIFF
--- a/glifo/src/glyph.rs
+++ b/glifo/src/glyph.rs
@@ -2191,6 +2191,7 @@ mod tests {
         let mut run = GlyphRun {
             font: font.clone(),
             font_size: 20.0,
+            font_embolden: Default::default(),
             transform: Affine::translate((0.0, 20.0)),
             glyph_transform: None,
             normalized_coords: &[],
--- a/glifo/src/glyph.rs
+++ b/glifo/src/glyph.rs
@@ -2191,7 +2191,7 @@ mod tests {
         let mut run = GlyphRun {
             font: font.clone(),
             font_size: 20.0,
-            font_embolden: Default::default(),
+            font_embolden: FontEmbolden::default(),
             transform: Affine::translate((0.0, 20.0)),
             glyph_transform: None,
             normalized_coords: &[],
--- a/glifo/src/glyph.rs
+++ b/glifo/src/glyph.rs
@@ -1592,14 +1592,6 @@ impl OutlinePen for OutlinePath {
 /// the need for updates only to align Skrifa versions.
 pub type NormalizedCoord = i16;
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    const _NORMALISED_COORD_SIZE_MATCHES: () =
-        assert!(size_of::<skrifa::instance::NormalizedCoord>() == size_of::<NormalizedCoord>());
-}
-
 /// Caches used for glyph rendering.
 ///
 /// Contains renderer-agnostic caches (outline paths, hinting instances)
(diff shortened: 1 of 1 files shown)


glifo/src/glyph.rs
```diff
@@ -1931,3 +1923,218 @@ fn x_y_advances(transform: &Affine) -> (Vec2, Vec2) {
...
+        let font = test_font(kind);
+        let glyph = test_glyph(&font, kind);
+        let mut resources = TestResources::default();
+
+        draw_test_glyph(&font, glyph, false, &mut resources);
+
+        assert_eq!(resources.glyph_atlas.len(), 0);
+        assert_eq!(resources.glyph_atlas.cache_hits(), 0);
+        assert_eq!(resources.glyph_atlas.cache_misses(), 0);
+    }
```
grebmeg, line 2250: Let's maybe draw the same glyph twice and assert `cache_hits == N`, `len == 1` (still one entry), `cache_misses` unchanged on the second draw — is ~3 extra lines per helper and increases the assertions' value. It would also catch any future bug where the cache key isn't stable across draws.

glifo/src/glyph.rs
```diff
@@ -1931,3 +1923,218 @@ fn x_y_advances(transform: &Affine) -> (Vec2, Vec2) {
...
+    fn ensure_cache(kind: TestGlyphKind) {
+        let font = test_font(kind);
+        let glyph = test_glyph(&font, kind);
+        let mut resources = TestResources::default();
+
+        draw_test_glyph(&font, glyph, true, &mut resources);
+
+        assert_eq!(resources.glyph_atlas.len(), 1);
+        assert_eq!(resources.glyph_atlas.cache_hits(), 0);
+        assert!(resources.glyph_atlas.cache_misses() > 0);
```
grebmeg, line 2231: This leaves the test blind to any regression that adds extra lookups, what about having `cache_misses() == N` instead of `cache_misses() > 0`?

glifo/src/glyph.rs
```diff
@@ -1931,3 +1923,218 @@ fn x_y_advances(transform: &Affine) -> (Vec2, Vec2) {
...
+    fn bitmap_glyph_is_cached_when_atlas_cache_is_enabled() {
+        ensure_cache(TestGlyphKind::Bitmap);
+    }
+
+    #[cfg(feature = "png")]
+    #[test]
+    fn bitmap_glyph_is_not_cached_when_atlas_cache_is_disabled() {
+        ensure_no_cache(TestGlyphKind::Bitmap);
+    }
+}
```
grebmeg, line 2294: Maybe it’s worth adding tests for stroke outlines to verify that they are not cached, regardless of whether caching is enabled or disabled?

grebmeg (approved):
LGTM, thanks for adding tests!

glifo/src/glyph.rs
LaurenzV: I added a second drawing round to the tests.

glifo/src/glyph.rs
LaurenzV: The problem is that we actually _do_ have different behavior. For COLR, we will end up with one more cache miss because we are also performing the speculative outline check. So is it okay for now if I just add a comment for that for now?

grebmeg: Oh, yeah, sure, missed that! Thanks!

glifo/src/glyph.rs
LaurenzV: Done!

PR #1632, opened 2026-05-09
vello: Support brush transforms for glyph runs

### LINKED ISSUE #1631: Glyph rendering with gradient brush
Hi, I can't find an example that uses a gradient brush for glyph/text rendering. I am experimenting with it and can't seem to get it to render correctly. Could it be because we can't assign a brush transform for glyph runs?

I may be completely in the wrong here, currently just seeking help!

waywardmonkeys:
Add a `brush_transform` field to encoded `GlyphRun` and expose it through `DrawGlyphs`. The resolver now emits an additional transform immediately before the glyph-run `PATH` marker, so gradient and image brushes use the run paint transform while per-glyph geometry transforms remain unchanged.

Thread the option through the scenes text helper and add smoke snapshot coverage using a transformed linear gradient across rendered text.

BREAKING CHANGE: Low-level callers constructing `vello_encoding::GlyphRun` literals must initialize `brush_transform`, usually to `None`.

Fixes #1631.


raphlinus (approved):
Looks good, I didn't find any problems. Thanks!

PR #1634, opened 2026-05-10
vello_common: Inline the `push_row` closure

LaurenzV:
I just noticed that #1368 did seem to regress tiling performance a bit, so I let codex try figure out how to improve it again. Here’s what it came up with (will split up into multiple PRs):

Times before #1368:
```
tile_aaa/Ghostscript_Tiger
                        time:   [107.84 µs 110.30 µs 112.78 µs]
                        change: [-15.968% -14.447% -12.990%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Times with current main:
```
tile_aaa/Ghostscript_Tiger
                        time:   [122.76 µs 123.48 µs 124.19 µs]
                        change: [+30.113% +31.913% +33.719%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (4.00%) high mild
```

The first optimization, implemented in this PR, is simply to use a macro instead of a closure, so that we don't need to invoke a function call (seems like the closure is not inlined in this case. I think I was the one who suggested this, so that's my bad!). This gives us the following results:

```
tile_aaa/Ghostscript_Tiger
                        time:   [101.27 µs 103.40 µs 105.80 µs]
                        change: [-18.378% -16.744% -15.246%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -703,90 +703,94 @@ impl Tiles {
                     let w_start_base = dx_dir << WINDING_SHIFT;
                     let w_end_base = not_dx_dir << WINDING_SHIFT;
 
-                    let mut push_row = |y_idx: u16,
-                                        row_top_y: f32,
-                                        row_bottom_y: f32,
-                                        w_start: u32,
-                                        w_end: u32,
-                                        w_single: u32| {
-                        let row_top_x = p0_x + (row_top_y - p0_y) * x_slope;
-                        let row_bottom_x = p0_x + (row_bottom_y - p0_y) * x_slope;
+                    let mut push_row = {
+                        #[inline(always)]
+                        |y_idx: u16,
+                         row_top_y: f32,
+                         row_bottom_y: f32,
+                         w_start: u32,
+                         w_end: u32,
+                         w_single: u32| {
+                            let row_top_x = p0_x + (row_top_y - p0_y) * x_slope;
+                            let row_bottom_x = p0_x + (row_bottom_y - p0_y) * x_slope;
 
-                        let row_left_x = f32::min(row_top_x, row_bottom_x).max(line_left_x);
-                        let row_right_x = f32::max(row_top_x, row_bottom_x).min(line_right_x);
+                            let row_left_x = f32::min(row_top_x, row_bottom_x).max(line_left_x);
+                            let row_right_x = f32::max(row_top_x, row_bottom_x).min(line_right_x);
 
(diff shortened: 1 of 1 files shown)


LaurenzV:
Github is making the diff look more complicated than it really is, here is how it looks on my end:

image

image
image

nicoburns:
Would it be possible to use an `#[inline(always)]` function along with a macro that calls that function (and provides the closed-over data as parameter(s))? The motivation for doing it that way is to avoid having a long function with meaningful context defined as a macro (which can make debugging, formatting, rust-analyzer, etc less effective).

LaurenzV:
The problem is that quite a few local variables are needed here (IIRC, on my phone right now so can’t check), so that would be a very ugly function :/

b0nes164:
Another possibility is a `LineContext` member struct that lives on `Tiles` and holds all the dx dy etc. ? And then we could move the macros into inlined private member functions?

LaurenzV:
We probably could but from my initial experiment this would require precomputing some variables which might not be needed in all branches (for example for the common case where a line is fully contained within a single tile), so that seems a bit unfortunate as well. But it might not be too bad, so if there is a strong preference I could try it.

tomcur (commented):
In general I would also prefer to try to avoid long macros.

DJMcNab:
Have we benchmarked just marking `push_row` itself as `#[inline(always)]`? The pattern for that is `let push_row = { #[inline(always)] |...| {} }`

That is, it needs to be in a block (or possibly parentheses)?

LaurenzV:
Genius!!! Thanks, I didn't know that's possible. Yes, I'm seeing a similar speedup here (this time on my M1 Pro, hence why it's a bit slower than my numbers before):

Before:
```
tile_aaa/Ghostscript_Tiger
                        time:   [164.26 µs 167.42 µs 172.11 µs]
                        change: [+28.996% +32.455% +37.039%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (6.00%) high severe
```

After:
```
tile_aaa/Ghostscript_Tiger
                        time:   [126.38 µs 127.25 µs 128.27 µs]
                        change: [-26.620% -24.173% -22.157%] (p = 0.00 < 0.05)
                        Performance has improved.
```

LaurenzV:
Here's the diff again for easier inspection:

image
image
image

LaurenzV:
@b0nes164 please take another look!

tomcur (approved):
Looks great!

PR #1635, opened 2026-05-10
vello_common: Add a fast path in tiling for lines fully contained in the viewport

LaurenzV:
Based on #1634. `push_row` is quite expensive, because a lot of logic is needed to account for culling. This PR adds a fast path for the common case where the line is completely contained in the viewport. This gets me to:

```
tile_aaa/Ghostscript_Tiger
                        time:   [89.356 µs 90.769 µs 92.179 µs]
                        change: [-3.0779% -0.9448% +1.1767%] (p = 0.39 > 0.05)
  4 (8.00%) high mild
```

Tested this against my PDF test suite and also in interactive mode.

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -742,6 +742,7 @@ impl Tiles {
                             let row_top_x = p0_x + (row_top_y - p0_y) * x_slope;
                             let row_bottom_x = p0_x + (row_bottom_y - p0_y) * x_slope;
 
+                            // TODO: Evaluate whether we need the second max/min.
                             let row_left_x = f32::min(row_top_x, row_bottom_x).max(line_left_x);
                             let row_right_x = f32::max(row_top_x, row_bottom_x).min(line_right_x);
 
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_common/src/tile.rs
```diff
@@ -803,30 +830,84 @@ impl Tiles {
...
+                    // otherwise would need to be made viewport culling work.
+                    if line_left_x >= 0.0 && line_right_x < tile_columns as f32 {
+                        if !is_start_culled {
+                            let y = f32::from(y_top_tiles);
+                            let row_bottom_y = (y + 1.0).min(line_bottom_y);
+                            let row_bottom_x = if row_bottom_y == line_bottom_y {
+                                line_bottom_x
+                            } else {
+                                p0_x + (row_bottom_y - p0_y) * x_slope
+                            };
```
b0nes164, line 831: I believe once you do this you don't need the min/max clamping to catch NaN's anymore, because this will actually quite gracefully catch the perfectly horizontal case? (And since a perfectly horizontal line is always contained to a single row, no other guards are neccessary.)

Can you add a TODO for this?

b0nes164 (changes requested):
LGTM modulo one comment!

b0nes164:
Are the failures on ubuntu flakes? I'm rerunning the jobs.

b0nes164 (approved):

PR #1638, opened 2026-05-13
vello: mark override image atlas entries dirty

waywardmonkeys:
Expose an explicit renderer method for callers to refresh atlas entries when an override texture's contents change. Also dirty resident entries when overrides are inserted or removed so switching between CPU image data and override textures refreshes the atlas.


waywardmonkeys:
This supersedes #1626

nicoburns (approved):
This one is also working for me. And the additional changes make sense to me.

PR #1639, opened 2026-05-14
vello_hybrid: Let `TextureBindings` own its `TextureView`s

tomcur:
In https://github.com/linebender/vello/pull/1552 we introduced `TextureBindings` taking `&'a TextureView`, but that requires users to keep the `TextureView` around somewhere other than this binding struct for the duration of the render call, and makes it hard to hold onto `TextureBindings` across frames, which makes the API somewhat awkward to use (thanks Nico for raising this!). These views are relatively cheap to clone with a refcount bump, so we can just store them owned instead.

Alternatively, we could introduce something like

```rust
pub trait TextureBindings {
    fn get(&self, id: TextureId) -> Option<&TextureView>;
}
```

allowing the user to abstract over their own storage, and we could provide a hash map implementation as we have now and perhaps an implementation for `Fn(TextureId) -> Option<&TextureView>`. We'd probably take it as `&dyn TextureBindings` then.

In the meantime, the form in this PR should be an improvement.

### DIFF
--- a/sparse_strips/vello_hybrid/examples/winit/src/main.rs
+++ b/sparse_strips/vello_hybrid/examples/winit/src/main.rs
@@ -389,12 +389,12 @@ impl ApplicationHandler for App<'_> {
                         .create_command_encoder(&wgpu::CommandEncoderDescriptor {
                             label: Some("Vello Render to Surface pass"),
                         });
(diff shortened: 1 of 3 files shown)


nicoburns (approved):
LGTM.

If we did add a trait like proposed, then potentially we could keep `TextureBindings` around and implement the trait for `&TextureBindings`.

PR #1641, opened 2026-05-14
Add explicit kurbo dependency to glifo and vello_encoding

nicoburns:
Ensures we are requiring kurbo `0.13.1` which we need for font embolden functionality.

The import chages were required by linebender's linting settings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1281,6 +1281,7 @@ dependencies = [
  "core_maths",
  "foldhash 0.2.0",
  "hashbrown 0.17.0",
+ "kurbo 0.13.1",
  "log",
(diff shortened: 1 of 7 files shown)


LaurenzV (approved):

nicoburns:
These tests are failing, but only in the merge queue checks :/

filter_transformed_blur_hybrid_constrained
filter_transformed_blur_hybrid
filter_drop_shadow_zero_offset_hybrid_constrained
filter_drop_shadow_zero_offset_hybrid

waywardmonkeys (changes requested):
We should do the peniko release and not do this here.

PR #1642, opened 2026-05-15
Add headless multi-layer symbol rendering example using Vello Classic

qianxiaoming:
This PR adds a new headless rendering example demonstrating how to use Vello Classic crate to render complex multi-layer  symbols into a PNG image.

**Features demonstrated**

* Polygon, polyline, and point symbol rendering
* Multi-pass layered road styling
* Gradient fills and hatch patterns
* Dashed strokes and clipping layers
* Affine transforms and scene composition
* PNG export through offscreen rendering

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3549,6 +3549,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "symbols"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "env_logger",
+ "futures-intrusive",
+ "png",
+ "pollster",
+ "vello",
+]
+
+[[package]]
 name = "syn"
 version = "2.0.117"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "examples/scenes",
     "examples/simple",
     "examples/with_winit",
+    "examples/symbols",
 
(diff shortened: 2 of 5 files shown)


DJMcNab:
Welcome! It isn't completely clear from your PR description what motivation this example serves (beyond that which our existing `headless` example also shows). As this is quite a lot of code to review, such guidance would be helpful.

A potential motivation could be a simpler demonstration of headless rendering than our existing `headless` example; specifically in avoiding the use of the `scenes` crate, which does obscure how to actually use the drawing somewhat. However, I suspect that (edit - added later: the fact that there's such a large amount of code make that less likely)

I do also think that we would be well served by an impressive demo scene showing our capabilities. I'm definitely excited to see something like that if that's the case. In that case I'd recommend that we should mention this example in the README, up to potentially even replacing the existing Tiger demonstration image.

No matter the motivation, I'd quite like to see an image of what this outputs in the PR description, which will help discussion later at [#office hours > 2026-05-15](https://xi.zulipchat.com/#narrow/channel/359642-office-hours/topic/2026-05-15/with/595205741) (which I unfortunately won't be attending).

qianxiaoming:
I tried using Vello as the underlying rendering engine for my map application. In this example, I attempted to draw some common map symbols, which were actually tests I did while learning to use Vello. I'm not sure if my approach is appropriate, as I generally use Vello as a drawing tool similar to QPainter, which is the GUI drawing tool in QT.
symbols

PR #1647, opened 2026-05-15
CI: run tests with `--release`

nicoburns:


xStrom (approved):
Comparing the runs, looks like this would cut down the actual test running time by a very significant margin.

| Platform | Old | New | Delta |
| --- | --- | --- | --- |
| Windows | 496s | 343s | -30% |
| macOS | 321s | 87s | -73% |
| Ubuntu | 710s | 146s | -80% |

Our build caching is currently job based anyway, so test jobs don't share cache with clippy jobs etc. So `--release` artifacts being incompatible with debug ones isn't an immediate concern.  However, it would mean that we couldn't unify those caches even if we wanted to. That said, we can approach that problem when cache uniting is actually working and/or our GitHub 10GB cache limit no longer cuts it.

For now, I think this is worth merging and trying out for real. Initial results are promising.

nicoburns:
I'll merge this after the releases and we can try it, but feel free to merge it sooner if you want to.

PR #1651, opened 2026-05-15
Sync Glifo readme with Linebender standards.

xStrom:
Synced header/footer to template and now generating the content with `cargo-rdme`.

### DIFF
--- a/glifo/src/lib.rs
+++ b/glifo/src/lib.rs
@@ -1,9 +1,18 @@
 // Copyright 2026 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! Glifo provides APIs for rendering and caching glyphs in a backend-agnostic way.
+//! Glifo provides APIs for efficiently rendering glyphs and paint styles like underline.
 //!
-//! ## Features
+//! # Goals
+//!
+//! Glifo is under rapid development. Consider it experimental for now. Its goals are to:
+//!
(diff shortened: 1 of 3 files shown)


glifo/src/lib.rs
```diff
@@ -1,9 +1,18 @@
...
 
-//! Glifo provides APIs for rendering and caching glyphs in a backend-agnostic way.
+//! Glifo provides APIs for efficiently rendering glyphs and paint styles like underline.
 //!
-//! ## Features
+//! # Goals
+//!
+//! Glifo is under rapid development. Consider it experimental for now. Its goals are to:
+//!
+//! - Provide an API surface that accepts glyphs and their positions and renders them to a surface.
```
nicoburns, line 10: This kinda makes it sounds like it's for rendering a "string" or "paragraph" of text which I don't think glifo will do. It's for rendering (or just outlining/scaling/hinting) single-glyphs (but potentially in batches, and into an atlas which is why this is technically true - I'd actually expect glifo to choose the positions in that case though).

nicoburns (approved):
Seems good enough. I don't necessarily 100% agree with all the goals, but I don't think we currently have consensus on that. So we might just need to put something for now and iterate later.

xStrom:
The goals were in the readme before, I didn't make any attempt to change them. We can adjust them for future releases as we continue to learn what we want.

PR #1652, opened 2026-05-15
Prepare for publishing.

xStrom:
* Vello 0.9.0
* Sparse Strips 0.0.8
* Glifo 0.1.0

Also removed the triple-date header from changelogs, as it's low value but high friction and we haven't missed it in any of the other changelogs where we've already removed it.


nicoburns (approved):
LGTM!

PR #1654, opened 2026-05-16
Experiment with whether removing vello_hybrid debug assertions improve CI running times

LaurenzV:


LaurenzV:
Hmm nope, doesn't seem like that's the root cause for slow vello_hybrid tests.

grebmeg:
On which platforms are we currently experiencing slow tests? I think I saw some issues on Linux in the past.

LaurenzV:
Now that the PR to use `--release` for tests was merged, the bottleneck still seems to be Windows, it still takes 16 minutes: https://github.com/linebender/vello/actions/runs/26001467199/job/76425339689

For some reason, hybrid tests run very slowly there. I'm not sure if it's because there is software emulation for the GPU.

PR #1655, opened 2026-05-16
sparse_strips: Make tile size configurable, Variant 1

LaurenzV:
We've on various occasions discussed that we want to experiment with different tile sizes. This is especially relevant for vello_hybrid, but also at least worth trying for vello_cpu.

Before we can do that, the first step is to make tile sizes more configurable. Right now, the width and height of 4 is hardcoded as an associated constant in vello_common. This PR changes it so that they become const generics instead, such that different crates can choose their own sizes. In order to keep the diff small for now, we simply define type aliases in vello_hybrid and vello_cpu that link to the existing 4x4 tile. The main algorithms (tiling, strip rendering) for now also still assume 4x4, changing this will be a follow-up PR.

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -205,7 +205,7 @@ impl CulledWindings {
 /// the compilation target.
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
-pub struct Tile {
+pub struct Tile<const WIDTH: usize = 4, const HEIGHT: usize = 4> {
     // The field ordering is important.
     //
     // The given ordering (variant over little and big endian compilation targets), ensures that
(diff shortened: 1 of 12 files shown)


LaurenzV:
Do you guys think it makes sense to turn those into a generic instead with a sealed trait with variants for 4x4, 8x8 and 16x16 provided such that 1) Clients cannot accidentally create other tile sizes and 2) Later on when we will have to wire the generic through `Tiles` and `StripGenerator`, we only have 1 generic parameter instead of 2? I still think it makes sense to keep the distinction between `Tile::WIDTH` and `Tile::HEIGHT ` internally, but not sure how much sense it makes to expose non-uniform tile sizes to the outside.

LaurenzV:
I opened a second version here #1656. Curious to hear your thoughts @grebmeg @tomcur @taj-p.

I _think_ I prefer the variant in this PR.

b0nes164:
So here's a thought: what if we switch to 8x8 for the tiles, but then we push two 4-high strips? Slightly wasteful on the command generation, but I think you could test the perf without changing everything else.

LaurenzV:
I mean perhaps we could, but I don't see the benefit of doing that. 🤔 If we are planning to rewriting tiling and strip rendering to work with 8x8 and 16x16 (which is planned for a follow-up after we've decided on how to make the tile size configurable).

b0nes164:
The main benefit is that you can isolate the tile size changes and test for correctness without worrying about everything else downstream.

LaurenzV:
I'm in the process of trying to implement the core kernel functions in vello_common to be generic, and am realizing that just trying to make the tiles themselves generic without actually trying to implement the rest doesn't make a whole lot of sense. Will close this PR and open a new one once I have a fully working solution.

PR #1656, opened 2026-05-16
sparse_strips: Make tile size configurable, Variant 2

LaurenzV:
See #1655, this is the second variant that I described in the comments. I'm not sure which one would be preferable.


LaurenzV:
See the comment in #1655

PR #1658, opened 2026-05-18
vello_cpu: Support custom atlas size

fundon:
- Add `ImageCache` into `Resources`, so the atlas size can be customized
- Expose `image_cache` `pixmaps` `glyph_atlas_mut` methods, users can read glyphs
- Simplify the initialization of `GlyphCacheKey`

### DIFF
--- a/glifo/src/atlas/cache.rs
+++ b/glifo/src/atlas/cache.rs
@@ -430,10 +430,10 @@ impl GlyphAtlas {
         self.cache_misses = 0;
     }
 
-    /// Reset the frame counter.
+    /// Resets the frame counter.
     ///
-    /// This is used when `vello_cpu` renders a large amount of text
-    /// and the old image is being grabbed by the new one.
+    /// This is used when `vello_cpu` renders a large amount of text to prevent
+    /// the new image from overwriting stale content from the previous one.
     pub fn clear_tick(&mut self) {
         self.serial = 0;
     }
--- a/glifo/src/atlas/cache.rs
+++ b/glifo/src/atlas/cache.rs
@@ -429,6 +429,13 @@ impl GlyphAtlas {
         self.cache_hits = 0;
         self.cache_misses = 0;
     }
+
+    /// Reset the frame counter.
+    ///
+    /// This is used when `vello_cpu` renders a large amount of text and the old image is being grabbed by the new one.
+    pub fn clear_tick(&mut self) {
+        self.serial = 0;
+    }
 }
 
 /// Queue a clear rect covering the full padded region of an evicted atlas slot.
--- a/glifo/src/atlas/cache.rs
+++ b/glifo/src/atlas/cache.rs
@@ -432,7 +432,8 @@ impl GlyphAtlas {
 
     /// Reset the frame counter.
     ///
-    /// This is used when `vello_cpu` renders a large amount of text and the old image is being grabbed by the new one.
+    /// This is used when `vello_cpu` renders a large amount of text
+    /// and the old image is being grabbed by the new one.
     pub fn clear_tick(&mut self) {
         self.serial = 0;
     }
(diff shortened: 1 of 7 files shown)


LaurenzV:
I can get behind configuring the atlas size, but not why you want to expose before_render/after_render? This is supposed to be abstracted away from the user.

LaurenzV:
Oh, I see now this is in parley… I believe this is still using an older version of vello CPU, so the code hasn’t been updated yet. It should be possible to remove all of this.

fundon:
Another scenario is that after rendering, we can pass `atlas.get(glyph_key)`, and then we can upload the atlas to our custom GPU texture.

fundon:
> It should be possible to remove all of this.

I hope there is a more concise way. After rendering, cache it into GlyphAtlas, and subsequent users can easily obtain data from `GlyphAtlas::get`.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -71,6 +76,13 @@ impl Resources {
         Self::default()
     }
 
+    /// Returns this [`Resources`] with a new [`AtlasConfig`].
+    #[cfg(feature = "text")]
+    pub fn with_atlas_config(mut self, atlas_config: AtlasConfig) -> Self {
+        self.image_cache = ImageCache::new_with_config(atlas_config);
+        self
+    }
```
LaurenzV: I don't think this should be a builder-pattern like method, otherwise you can create  a new `Resources` and then basically invalidate it by overriding the existing image cache. I think I would just prefer a `new_with` method.

LaurenzV (commented):
LGTM overall.

LaurenzV:
By the way, I do want to warn you that the atlas caching API is still experimental and not meant for production use yet. While it seems to work, there are still a number of issues to resolve, especially around performance when zooming.

sparse_strips/vello_cpu/src/render.rs
fundon: The first version also uses `new_with` to create new instance. :)

fundon:
> By the way, I do want to warn you that the atlas caching API is still experimental and not meant for production use yet. While it seems to work, there are still a number of issues to resolve, especially around performance when zooming.

Ok. I'll give feedback if there are any problems. Thanks.

glifo/src/atlas/cache.rs
```diff
@@ -429,6 +429,14 @@ impl GlyphAtlas {
         self.cache_hits = 0;
         self.cache_misses = 0;
     }
+
+    /// Resets the frame counter.
+    ///
+    /// This is used when `vello_cpu` renders a large amount of text to prevent
+    /// the new image from overwriting stale content from the previous one.
+    pub fn clear_tick(&mut self) {
```
fundon, line 437: Is there a better way to avoid this situation?

glifo/src/atlas/key.rs
```diff
@@ -193,7 +189,7 @@ pub(crate) fn pack_color(color: AlphaColor<Srgb>) -> u32 {
     reason = "result is clamped to SUBPIXEL_BUCKETS-1 which fits in u8"
 )]
 #[inline]
-fn quantize_subpixel(frac: f32) -> u8 {
+pub fn quantize_subpixel(frac: f32) -> u8 {
```
fundon, line 192: Expose this method, so we can create keys freely outside of glifo.

sparse_strips/vello_cpu/src/text.rs
```diff
@@ -100,17 +96,19 @@ fn ensure_page(
 }
 
 impl Resources {
-    pub(crate) fn prepare_glyph_cache(&mut self, render_mode: RenderMode) {
+    /// Prepares the glyph cache.
+    pub fn prepare_glyph_cache(&mut self, render_mode: RenderMode) {
```
fundon, line 100: See https://gist.github.com/fundon/2fbf7f6b5d685bb8ed87174fcdafe867#file-fixed_pos_main-rs-L181-L188

Write to cache first, then process text.

PR #1659, opened 2026-05-19
vello_sparse_shaders: Lint against struct values in fragment shaders

grebmeg:
Drafted as a follow-up to [this comment](https://github.com/linebender/vello/pull/1619#discussion_r3269440318).

Some old Adreno GPU drivers silently downgrades the precision of struct values in fragment shaders to 16 bits, regardless of the precision the shader requested. The last few commits flattened the existing fragment shaders to dodge the bug; this commit prevents the workaround from silently regressing.

`assert_no_structs_in_fragment_shader` parses the WGSL with naga, walks the call graph reachable from the `@fragment` entry point, and rejects any function argument, return value, local variable, `Compose`, or `ZeroValue` of struct type. Struct-typed uniform globals and vertex IO structs are unaffected because their fields are only ever read as scalars/vectors inside the fragment shader.

The lint runs both at build time (from `compile_wgsl_shader`, which the build script invokes for every shipped WGSL) and as a unit test (`every_shipped_shader_passes_the_lint`) that runs in CI.

See #1604 #1619 #1620 for context.

Generated with AI assistance.

### DIFF
--- a/sparse_strips/vello_sparse_shaders/src/compile.rs
+++ b/sparse_strips/vello_sparse_shaders/src/compile.rs
@@ -8,7 +8,7 @@ use naga::{
     valid::{Capabilities, ValidationFlags, Validator},
 };
 
-use crate::lint::assert_no_structs_in_fragment_shader;
+use crate::lint::lint;
 use crate::types::{CompiledGlsl, ReflectionMap, Stage};
 
 #[allow(
@@ -16,9 +16,6 @@ use crate::types::{CompiledGlsl, ReflectionMap, Stage};
     reason = "False positive as compile_wgsl_shader is used at build time."
 )]
 /// Compiles the given wgsl source into GLSL using [naga].
-///
-/// `shader_name` is only used to make diagnostics more actionable when the
-/// [`assert_no_structs_in_fragment_shader`] lint fires.
 pub(crate) fn compile_wgsl_shader(
     source: &str,
     shader_name: &str,
@@ -27,10 +24,7 @@ pub(crate) fn compile_wgsl_shader(
 ) -> CompiledGlsl {
     let module = wgsl::parse_str(source).unwrap();
 
-    // Guard against accidentally reintroducing struct values in fragment shaders. Some
-    // Adreno GPU drivers downgrade these to 16-bit precision; see
-    // https://github.com/linebender/vello/pull/1604.
-    assert_no_structs_in_fragment_shader(shader_name, &module);
+    lint(shader_name, &module);
 
     let info = Validator::new(ValidationFlags::all(), Capabilities::default())
         .subgroup_stages(naga::valid::ShaderStages::all())
--- a/sparse_strips/vello_sparse_shaders/src/compile.rs
+++ b/sparse_strips/vello_sparse_shaders/src/compile.rs
@@ -8,6 +8,7 @@ use naga::{
     valid::{Capabilities, ValidationFlags, Validator},
 };
 
+use crate::lint::assert_no_structs_in_fragment_shader;
 use crate::types::{CompiledGlsl, ReflectionMap, Stage};
 
 #[allow(
@@ -15,12 +16,21 @@ use crate::types::{CompiledGlsl, ReflectionMap, Stage};
     reason = "False positive as compile_wgsl_shader is used at build time."
 )]
 /// Compiles the given wgsl source into GLSL using [naga].
+///
+/// `shader_name` is only used to make diagnostics more actionable when the
+/// [`assert_no_structs_in_fragment_shader`] lint fires.
 pub(crate) fn compile_wgsl_shader(
     source: &str,
+    shader_name: &str,
     vertex_entry: &str,
     fragment_entry: &str,
 ) -> CompiledGlsl {
     let module = wgsl::parse_str(source).unwrap();
+
+    // Guard against accidentally reintroducing struct values in fragment shaders. Some
+    // Adreno GPU drivers downgrade these to 16-bit precision; see
+    // https://github.com/linebender/vello/pull/1604.
+    assert_no_structs_in_fragment_shader(shader_name, &module);
 
     let info = Validator::new(ValidationFlags::all(), Capabilities::default())
         .subgroup_stages(naga::valid::ShaderStages::all())
--- a/sparse_strips/vello_sparse_shaders/src/lint.rs
+++ b/sparse_strips/vello_sparse_shaders/src/lint.rs
@@ -1,0 +1,307 @@
+// Copyright 2026 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Lints for the WGSL shaders.
+//!
+//! At the moment a single lint is enforced: no user-defined struct value may appear
+//! anywhere in code reachable from a `@fragment` entry point. A row of older Adreno
+//! GPU drivers silently downgrade the numeric precision of structs in fragment shaders
+//! to 16 bits, regardless of the precision the shader actually requested. The
+//! workaround is to flatten all data into scalar or vector primitives. See
+//! <https://github.com/linebender/vello/pull/1604> and the issues linked from it.
+//!
+//! Struct-typed uniform globals (e.g. `Config`) and vertex IO structs are allowed
+//! because their fields are only ever read as scalars/vectors inside the fragment
+//! shader — the struct itself is never materialised in fragment-shader storage.
+
+use std::collections::{BTreeSet, VecDeque};
+
+use naga::{Block, Expression, Function, Handle, Module, ShaderStage, Statement, Type, TypeInner};
+
+/// Asserts that no struct value is referenced from the fragment entry point of
+/// `wgsl_source`. Panics with a detailed message listing each violation if the rule
+/// is broken. See the module-level documentation for background.
+///
+/// Modules without a fragment entry point are accepted silently.
+pub(crate) fn assert_no_structs_in_fragment_shader(shader_name: &str, module: &Module) {
+    let Some(entry_point) = module
+        .entry_points
+        .iter()
+        .find(|ep| ep.stage == ShaderStage::Fragment)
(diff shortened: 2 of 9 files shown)


LaurenzV:
Haven't checked in detail but approach seems good overall, let's use this PR as the basis then once the others have been merged!

ajakubowicz-canva (commented):
Drive by (sorry didn't have much time) – moving towards a lint like this is excellent!

The following shader is not caught by the lint. Should it be caught?

```rs
    const SHADER_LOADS_UNIFORM_STRUCT: &str = r#"
struct Config { width: u32, height: u32 }

@group(0) @binding(0) var<uniform> config: Config;

@vertex
fn vs_main() -> @builtin(position) vec4<f32> {
    return vec4<f32>(0.0);
}

@fragment
fn fs_main() -> @location(0) vec4<f32> {
    let copied = config;
    return vec4<f32>(f32(copied.width), f32(copied.height), 0.0, 1.0);
}
"#;


    #[test]
    fn rejects_whole_uniform_struct_load_in_fragment_shader() {
        assert_no_structs_in_fragment_shader(
            "loads_uniform_struct",
            &parse(SHADER_LOADS_UNIFORM_STRUCT),
        );
    }
```

LaurenzV:
> Should it be caught?

I tested it with a Google Pixel 5 and it seems to be broken indeed, so probably yes. The question is how hard it is to implement...

sparse_strips/vello_sparse_shaders/src/compile.rs
```diff
@@ -8,20 +8,30 @@ use naga::{
...
 use crate::types::{CompiledGlsl, ReflectionMap, Stage};
 
 #[allow(
     dead_code,
     reason = "False positive as compile_wgsl_shader is used at build time."
 )]
 /// Compiles the given wgsl source into GLSL using [naga].
+///
+/// `shader_name` is only used to make diagnostics more actionable when the
+/// [`assert_no_structs_in_fragment_shader`] lint fires.
```
LaurenzV: I'm wondering whether the comment is a bit superfluous? I don't think we have to justify every argument of a method 😄

sparse_strips/vello_sparse_shaders/src/compile.rs
```diff
@@ -8,20 +8,30 @@ use naga::{
...
+    shader_name: &str,
     vertex_entry: &str,
     fragment_entry: &str,
 ) -> CompiledGlsl {
     let module = wgsl::parse_str(source).unwrap();
 
+    // Guard against accidentally reintroducing struct values in fragment shaders. Some
+    // Adreno GPU drivers downgrade these to 16-bit precision; see
+    // https://github.com/linebender/vello/pull/1604.
+    assert_no_structs_in_fragment_shader(shader_name, &module);
```
LaurenzV: What do you think about creating a `lint` method and calling it in there so we can group them into that method in the future in case we decide to add more such lints?

sparse_strips/vello_sparse_shaders/src/lint.rs
```diff
@@ -0,0 +1,307 @@
...
+//! At the moment a single lint is enforced: no user-defined struct value may appear
+//! anywhere in code reachable from a `@fragment` entry point. A row of older Adreno
+//! GPU drivers silently downgrade the numeric precision of structs in fragment shaders
+//! to 16 bits, regardless of the precision the shader actually requested. The
+//! workaround is to flatten all data into scalar or vector primitives. See
+//! <https://github.com/linebender/vello/pull/1604> and the issues linked from it.
+//!
+//! Struct-typed uniform globals (e.g. `Config`) and vertex IO structs are allowed
+//! because their fields are only ever read as scalars/vectors inside the fragment
+//! shader — the struct itself is never materialised in fragment-shader storage.
```
LaurenzV: I think I would prefer just having that description directly above `assert_no_structs_in_fragment_shader` so in case we add future lints we don't need to put everything at the top of the file. For the module description just the first sentence seems enough to me.

sparse_strips/vello_sparse_shaders/src/lint/no_structs_in_fragment.rs
```diff
@@ -0,0 +1,307 @@
...
+pub(crate) fn assert_no_structs_in_fragment_shader(shader_name: &str, module: &Module) {
+    let Some(entry_point) = module
+        .entry_points
+        .iter()
+        .find(|ep| ep.stage == ShaderStage::Fragment)
+    else {
+        return;
+    };
+
+    let mut violations = Vec::new();
```
LaurenzV, line 33: Arguably this is something that should be passed from the parent `lint` function so it can be shared across multiple lint functions in the future.

sparse_strips/vello_sparse_shaders/src/lint.rs
```diff
@@ -0,0 +1,307 @@
...
+         (e.g. unpack a `vec4<u32>` texel directly with top-level `get_*` helpers).\n\
+         See https://github.com/linebender/vello/pull/1604 for context.\n\n\
+         Violations:\n"
+    );
+    for violation in &violations {
+        message.push_str("  - ");
+        message.push_str(violation);
+        message.push('\n');
+    }
+    panic!("{message}");
```
LaurenzV: Up to you but I would also suggest doing the message concatenation in the parent function and writing the function in a way such that it comes natural to add more lint functions in the future.

sparse_strips/vello_sparse_shaders/src/lint/no_structs_in_fragment.rs
```diff
@@ -0,0 +1,307 @@
...
+        if is_struct(module, local.ty) {
+            violations.push(format!(
+                "{function_label}: local variable `{}` has struct type `{}`",
+                local.name.as_deref().unwrap_or("<unnamed>"),
+                type_name(module, local.ty),
+            ));
+        }
+    }
+    for (_, expr) in func.expressions.iter() {
+        match expr {
```
LaurenzV, line 111: As Andrew mentioned, I suppose we also have to check for `GlobalVariable` and `LocalVariable` and check whether they are structs?

sparse_strips/vello_sparse_shaders/src/lint.rs
```diff
@@ -0,0 +1,307 @@
...
+            "with_struct_helper",
+            &parse(SHADER_WITH_STRUCT_HELPER),
+        );
+    }
+
+    // The same Adreno bug also fires when the fragment entry point takes a struct as
+    // its input parameter (every field of `in` gets silently downgraded to 16-bit).
+    // The flattening of `fs_main(in: FilterVertexOutput)` into individual `@location`
+    // parameters in commit 07ef807b was the workaround; this test pins down that
+    // shape so it cannot regress.
```
LaurenzV:
```suggestion
    // parameters in commit 07ef807b was the workaround; 
```

Last sentence seems superfluous to me

sparse_strips/vello_sparse_shaders/src/lint.rs
```diff
@@ -0,0 +1,307 @@
...
+        let shader_dir = manifest_dir.join("shaders");
+        let shaders =
+            crate::shader_info::load_shader_infos(&shader_dir).expect("load WGSL shaders");
+        assert!(
+            !shaders.is_empty(),
+            "expected at least one shader in {shader_dir:?}"
+        );
+        for shader in shaders {
+            assert_no_structs_in_fragment_shader(&shader.name, &parse(&shader.wgsl_source));
+        }
```
LaurenzV: If you implement my `lint` function suggestion I would suggest making this more general so it calls the whole lint function instead of just one specific lint.

sparse_strips/vello_sparse_shaders/src/compile.rs
```diff
@@ -8,20 +8,30 @@ use naga::{
...
     source: &str,
+    shader_name: &str,
     vertex_entry: &str,
     fragment_entry: &str,
 ) -> CompiledGlsl {
     let module = wgsl::parse_str(source).unwrap();
 
+    // Guard against accidentally reintroducing struct values in fragment shaders. Some
+    // Adreno GPU drivers downgrade these to 16-bit precision; see
+    // https://github.com/linebender/vello/pull/1604.
```
LaurenzV:
```suggestion
```

I think we don't need to repeat the explanation if we add the doc comment to the method as I suggested.

LaurenzV (approved):
LGTM! Up to you, but I would also suggest removing the warning comments in the filter and render strips shader now? https://github.com/linebender/vello/blob/f1da2a22c25bbca11b48aed8d1cdb9000baa7223/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl#L4-L12

Perhaps it's just me, but I think it's unnecessary to keep repeating the same comment across different locations, it adds a lot of redundancy, especially if more lints are added in the future; having it once above the actual lint function should be enough. And since this is now caught in CI, I don't think we need to keep the explicit hint in the actual shader.

sparse_strips/vello_sparse_shaders/src/lint.rs
```diff
@@ -0,0 +1,307 @@
...
+pub(crate) fn assert_no_structs_in_fragment_shader(shader_name: &str, module: &Module) {
+    let Some(entry_point) = module
+        .entry_points
+        .iter()
+        .find(|ep| ep.stage == ShaderStage::Fragment)
+    else {
+        return;
+    };
+
+    let mut violations = Vec::new();
```
grebmeg: I went with `Vec<LintReport>` aggregated in the parent rather than literally threading `&mut Vec<String>` — same idea (no per-pass aggregation, parent owns it) but each pass also carries its own summary/explanation alongside its violations, so a future lint with a different rationale won't need to share a generic header.

grebmeg:
Thanks @ajakubowicz-canva @LaurenzV! 🙏 I’ve addressed your comments and also moved the specific lint into its own file, so it should now be easy to add new lints if needed.

PR #1663, opened 2026-05-22
ci: cancel superseded workflow runs for PR pushes

grebmeg:
Adds a concurrency block to `.github/workflows/ci.yml` that cancels an in-flight workflow run when a newer commit is pushed to the same pull request. This avoids burning runner minutes on commits that have already been superseded.

Any concerns with adding this?

[Control the concurrency of workflows and jobs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,11 @@ on:
     branches:
       - main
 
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
I would be in favour of this

PR #1664, opened 2026-05-23
fix: preserve image atlas across patchless resolves

waywardmonkeys:
Vello's renderer keeps a persistent GPU image atlas, while the resolver tracks logical atlas residency and dirty image uploads. Patchless resolves used to return Images::default(), which made the renderer treat an image-free frame as a 0x0 atlas and replace a previously populated atlas with a fresh 1x1 texture.

The resolver still kept resident image entries marked clean. If a later scene reused the same image, it returned cached atlas coordinates but scheduled no upload. When the renderer grew the atlas back to its previous size, that new GPU texture did not contain the resident image, so image brushes, including glyph image brushes, sampled blank content.

Patchless resolves now begin an image-cache resolve and return the cache's resident atlas state when there are resident images. That clears the per-resolve upload list while preserving the GPU atlas those clean images sample from. Fresh patchless resolves with no resident images still report no atlas, so image-free frames do not allocate the default image atlas just to draw solid content.

Shader hot reload had the same class of failure: it dropped the GPU atlas by replacing the engine while keeping clean resolver residency. Reset resolver state when reloading shaders, and carry image overrides into the new engine so the next image frame uploads into the fresh atlas.

### DIFF
--- a/vello_encoding/src/resolve.rs
+++ b/vello_encoding/src/resolve.rs
@@ -187,8 +187,17 @@ impl Resolver {
     ) -> (Layout, Ramps<'a>, Images<'a>) {
         let resources = &encoding.resources;
         if resources.patches.is_empty() {
+            // This resolve has no late-bound image patches, so it has no image uploads.
+            // If previous resolves left resident images in the cache, keep reporting the
+            // atlas size so the renderer can reuse the GPU atlas those clean images sample
+            // from.
+            self.image_cache.begin_resolve();
             let layout = resolve_solid_paths_only(encoding, packed);
-            return (layout, Ramps::default(), Images::default());
+            return (
+                layout,
+                Ramps::default(),
+                self.image_cache.images_for_patchless_resolve(),
+            );
         }
         let patch_sizes = self.resolve_patches(encoding);
         self.resolve_pending_images();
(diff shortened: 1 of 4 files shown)


waywardmonkeys:

It was found while updating `imaging` to use Vello 0.9.

This was a result of changes that I introduced for Vello 0.8.

We should do a 0.9.1 release with this fix. (And let's see if there are other fixes...)

waywardmonkeys:
I kicked off a group of adversarial subagents and it found a couple of further things wrong.

PR updated.

vello_encoding/src/resolve.rs
```diff
@@ -650,3 +659,55 @@ fn size_to_words(byte_size: usize) -> u32 {
...
     len + (len.wrapping_neg() & (alignment as usize - 1))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use peniko::{Blob, ImageAlphaType, ImageBrush, ImageFormat};
+    use std::sync::Arc;
+
+    fn image(id_byte: u8, width: u32, height: u32) -> ImageData {
```
raphlinus, line 669: Name "id_byte" is confusing to me, it's not an id, it seems to be the image contents.

vello_tests/tests/regression.rs
```diff
@@ -144,6 +152,187 @@ fn text_stroke_width_zero() {
...
+        Color::from_rgb8(30, 30, 34),
+        None,
+        &Rect::new(0.0, 0.0, 256.0, 256.0),
+    );
+    scene
+}
+
+#[test]
+#[cfg_attr(skip_gpu_tests, ignore)]
+fn glyph_image_brush_survives_image_free_render() {
```
raphlinus, line 237: This seems to be doing an awful lot of the harness work from scratch as opposed to using existing harnesses. In particular, it seems to duplicate a lot of logic in `get_scene_image()`. Is there a particular reason that wasn't reused?

raphlinus (approved):
Sorry for the delay in reviewing; even though it's not a big patch, I needed to get my head back around the issues. I can see now what it fixes, and the tests capture that. A couple of nits in the comments which don't block landing, just possible opportunities to clean up tests.

francisdb:
Hit this on 0.10 in a downstream project ([cuelight](https://github.com/francisdb/cuelight/pull/18)): a looping show passes through a scene that draws only paths, and every image and text bitmap renders blank from then on.

Workaround until this lands, in case it helps others: always include one transparent 1x1 image in the scene, so no resolve is patchless and the atlas is never dropped.

Would be great to see this merged and released.

PR #1665, opened 2026-05-23
vello_cpu: Unify render APIs

### LINKED ISSUE #1382: Support `PixmapMut<'_>` and stride in `vello_cpu`?

LaurenzV:
Supersedes #1597, closes #1382.

# Motivation

Right now, we have two rendering APIs: `render_to_pixmap` and `composite_to_pixmap_at_offset`. We are about to add a third one in #1597. Having this many APIs is very confusing, and after thinking about this more closely I realized that really, what we have been doing is adding methods that do similar things but just with different knobs enabled/disabled. I think it would be much cleaner if we had _one single_ API entry point for rendering to a pixmap, and instead expose the available knobs as a simple settings struct.

# Implementation

Therefore, this PR proposes to:
1) Expose **one single** `render` method that can be used to render the contents of a `RenderContext` into `Pixmap`-like struct.
2) Introduce a `PixmapMut` struct to allow constructing a render target from a mutable borrowed byte slice. Let `render` always take a ` PixmapMut`, allowing to pass both, a custom buffer or an owned pixmap as the render target.
3) Introduce a `CompositeMode` enum that defines whether the contents of a pixmap should be completely replaced (by default), or whether source-over compositing should be used.
4) Also introduce a `PixelFormat` struct, to possibly allow `Bgra8` in the future, which is useful for MacOS.
5) `RenderMode` has been moved from being stored in `RenderContext` to instead being a setting during rasterization. I initially stored this in `RenderContext` because I thought it might happen that the setting will also be needed during scene construction, but this turned out to not be the case in the end.
5) Previously, the implicit assumption was that `RenderContext` and `Pixmap` always need to have the same size, although it "accidentally" was possible for them to _not_ have the same size, and some people have made use of that. This PR relaxes the conditions imposed on the size, explicitly allowing size mismatches and also documents what the intended semantics are for each.

# Testing
I've added some new tests for the expected behavior. Existing tests pass. I've also ran `vello_cpu_winit` with the different scenes and didn't notice any regressions.

**TODO: I still want to run this version against my PDF test suite, but this shouldn't block a review.**

### DIFF
--- a/sparse_strips/vello_common/src/pixmap.rs
+++ b/sparse_strips/vello_common/src/pixmap.rs
@@ -29,6 +29,51 @@ pub struct Pixmap {
     may_have_transparency: bool,
 }
 
+/// A mutable view into premultiplied RGBA8 pixmap data.
+#[derive(Debug)]
+pub struct PixmapMut<'a> {
+    /// Width of the pixmap in pixels.
+    width: u16,
+    /// Height of the pixmap in pixels.
+    height: u16,
+    /// Buffer of the pixmap in RGBA8 format.
+    buf: &'a mut [u8],
+}
+
+impl<'a> PixmapMut<'a> {
+    /// Create a new mutable pixmap view.
+    ///
+    /// Returns `None` if `buf` is not exactly `width * height * 4` bytes long.
+    pub fn new(width: u16, height: u16, buf: &'a mut [u8]) -> Option<Self> {
+        if buf.len() == usize::from(width) * usize::from(height) * 4 {
+            Some(Self { width, height, buf })
+        } else {
+            None
+        }
+    }
+
+    /// Return the width of the pixmap.
+    pub fn width(&self) -> u16 {
+        self.width
+    }
+
+    /// Return the height of the pixmap.
+    pub fn height(&self) -> u16 {
+        self.height
+    }
+
+    /// Returns a mutable reference to the underlying data as premultiplied RGBA8 bytes.
+    pub fn data_mut(&mut self) -> &mut [u8] {
+        self.buf
+    }
+}
+
+impl<'a> From<&'a mut Pixmap> for PixmapMut<'a> {
+    fn from(pixmap: &'a mut Pixmap) -> Self {
+        pixmap.as_mut()
+    }
+}
+
 impl Pixmap {
     /// Create a new pixmap with the given width and height in pixels.
     ///
@@ -293,6 +338,10 @@ impl Pixmap {
         &self.buf
     }
 
+    // TODO: Now that we have `as_mut`, maybe we don't need the
+    // mutable methods. If we add a `PixmapRef` we can also remove the
+    // non-mutable ones.
+
     /// Returns a mutable reference to the underlying data as premultiplied RGBA8.
     ///
     /// The pixels are in row-major order.
@@ -314,6 +363,15 @@ impl Pixmap {
     /// `[r, g, b, a]`.
     pub fn data_as_u8_slice_mut(&mut self) -> &mut [u8] {
         bytemuck::cast_slice_mut(&mut self.buf)
+    }
+
+    /// Return a mutable view into this pixmap's pixel data.
+    pub fn as_mut(&mut self) -> PixmapMut<'_> {
+        PixmapMut {
+            width: self.width,
+            height: self.height,
+            buf: bytemuck::cast_slice_mut(&mut self.buf),
+        }
     }
 
     /// Sample a pixel from the pixmap.
--- a/sparse_strips/vello_cpu/README.md
+++ b/sparse_strips/vello_cpu/README.md
@@ -29,7 +29,7 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 [RenderContext::glyph_run]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.glyph_run
 [RenderMode::OptimizeSpeed]: https://docs.rs/vello_cpu/latest/vello_cpu/enum.RenderMode.html#variant.OptimizeSpeed
 [RenderMode::OptimizeQuality]: https://docs.rs/vello_cpu/latest/vello_cpu/enum.RenderMode.html#variant.OptimizeQuality
-[`RenderContext::render_to_pixmap`]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.render_to_pixmap
+[`RenderContext::render`]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.render
 [`Pixmap`]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.Pixmap.html
 
 <!-- cargo-rdme start -->
@@ -43,15 +43,15 @@ Vello CPU is being developed as part of work to address shortcomings in Vello.
 
 To use Vello CPU, you need to:
 
-- Create a [`RenderContext`][], a 2D drawing context for a fixed-size target area.
+- Create a [`RenderContext`][], a 2D drawing context for a fixed-size scene area.
 - For each object in your scene:
   - Set how the object will be painted, using [`set_paint`][RenderContext::set_paint].
   - Set the shape to be drawn for that object, using methods like [`fill_path`][RenderContext::fill_path],
     [`stroke_path`][RenderContext::stroke_path], or [`glyph_run`][RenderContext::glyph_run].
-- Render it to an image using [`RenderContext::render_to_pixmap`][].
+- Render it to an image using [`RenderContext::render`][].
 
 ```rust
-use vello_cpu::{RenderContext, Resources, Pixmap, RenderMode};
+use vello_cpu::{RasterizerSettings, RenderContext, Resources, Pixmap};
 use vello_cpu::{color::{palette::css, PremulRgba8}, kurbo::Rect};
 let width = 10;
 let height = 5;
@@ -64,7 +64,7 @@ let mut target = Pixmap::new(width, height);
 // While calling `flush` is only strictly necessary if you are rendering using
 // multiple threads, it is recommended to always do this.
 context.flush();
-context.render_to_pixmap(&mut resources, &mut target);
+context.render(&mut target, &mut resources, RasterizerSettings::default());
 
 let expected_render = b"\
     0000000000\
--- a/sparse_strips/vello_cpu/README.md
+++ b/sparse_strips/vello_cpu/README.md
@@ -51,7 +51,7 @@ To use Vello CPU, you need to:
 - Render it to an image using [`RenderContext::render`][].
 
 ```rust
-use vello_cpu::{RasterizerSettings, RenderContext, Resources, Pixmap};
+use vello_cpu::{RenderContext, Resources, Pixmap};
 use vello_cpu::{color::{palette::css, PremulRgba8}, kurbo::Rect};
 let width = 10;
 let height = 5;
@@ -64,7 +64,7 @@ let mut target = Pixmap::new(width, height);
 // While calling `flush` is only strictly necessary if you are rendering using
 // multiple threads, it is recommended to always do this.
 context.flush();
-context.render(&mut target, &mut resources, RasterizerSettings::default());
+context.render(&mut target, &mut resources);
 
 let expected_render = b"\
     0000000000\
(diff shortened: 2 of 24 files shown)


LaurenzV:
@nicoburns If you want to feel free to give it a spin.

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -316,6 +365,15 @@ impl Pixmap {
...
     }
 
+    /// Return a mutable view into this pixmap's pixel data.
+    pub fn as_mut(&mut self) -> PixmapMut<'_> {
+        PixmapMut {
+            width: self.width,
+            height: self.height,
+            buf: bytemuck::cast_slice_mut(&mut self.buf),
+        }
+    }
```
nicoburns, line 375: This should probably be an implementation of the stdlib `AsMut` trait

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -29,6 +29,51 @@ pub struct Pixmap {
...
 
+/// A mutable view into premultiplied RGBA8 pixmap data.
+#[derive(Debug)]
+pub struct PixmapMut<'a> {
+    /// Width of the pixmap in pixels.
+    width: u16,
+    /// Height of the pixmap in pixels.
+    height: u16,
+    /// Buffer of the pixmap in RGBA8 format.
+    buf: &'a mut [u8],
```
nicoburns, line 40: Probably not blocking, but it's a little weird that `Pixmap` uses `PremulRgba8` whereas `PixmapMut` uses `u8` (my preference would be `u8` everywhere).

sparse_strips/vello_cpu/src/dispatch/mod.rs
```diff
@@ -72,23 +73,10 @@ pub(crate) trait Dispatcher: Debug + Send + Sync {
...
-        dst_buffer_width: u16,
-        dst_buffer_height: u16,
-        render_mode: RenderMode,
+        target: PixmapMut<'_>,
+        scene_width: u16,
+        scene_height: u16,
+        settings: RasterizerSettings,
         encoded_paints: &[EncodedPaint],
         image_resolver: &dyn ImageResolver,
     );
```
nicoburns, line 82: Why are `dst_x` and `dst_y` in `RasterizerSettings`, but `scene_width` and `scene_height` are passed standalone? I think they should all be passed in the same way (I don't particularly mind which).

nicoburns (commented):
I'm onboard with the general concept, but I think it could use some tweaks.

Eventually I would keen on:
- Adding PixmapRef
- Moving all of the Pixmap types into their own crate (so that things like `glifo` could use them without depending on `vello_common)

sparse_strips/vello_cpu/src/dispatch/mod.rs
LaurenzV: Because dst_x/dst_y is passed by the user, while the scene width/height is only passed internally directly from the render context.

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: This is not possible because the trait returns a &mut T, while we return a custom wrapper type that only conceptually represents a mutable reference to the pixmap.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -621,35 +670,66 @@ impl RenderContext {
...
+        let mut target = target.into();
+        let target_fully_covered = settings.offset == (0, 0)
+            && self.width >= target.width()
+            && self.height >= target.height();
+        // If the scene covers the whole pixmap than packing will take care
+        // of clearing everything anyway, so no reason to clear it explicitly
+        // here.
+        if settings.composite_mode == CompositeMode::Replace && !target_fully_covered {
+            target.data_mut().fill(0);
+        }
```
grebmeg, line 734: I wouldn’t expect the target would be cleared beforehand when using `CompositeMode::Replace`. I’d interpret that mode as “replace the rendered region with the scene”, not “wipe everything and then draw the scene”. Also, don’t we already clear regions later with `fine.clear`? Is `Replace` equivalent to the `Copy` blend mode? If so, could we just use the two Porter-Duff operators here and leave it up to the consumer to decide whether the pixmap should be cleared beforehand?

LaurenzV: We could change the semantics. It just seems a bit unexpected to me that if the render context is smaller than the pixmap, any pixels that aren’t touched by the scene are left stale (which might also be relevant for our internal use case, in that case we would have to manually clear the pixmap before rendering into it).

LaurenzV: > Also, don’t we already clear regions later with fine.clear?

We do, but if the scene doesn't cover the whole pixmap, as I mentioned this will leave the outside regions in the existing pixmap untouched. This is why this action is only triggered if `!target_fully_covered`.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -82,6 +85,62 @@ impl Resources {
...
+    /// Whether to prioritize speed or quality when rendering.
+    ///
+    /// For most cases (especially for real-time rendering), it is highly recommended to set
+    /// this to [`RenderMode::OptimizeSpeed`]. If color accuracy is a more significant concern,
+    /// then you can set this to [`RenderMode::OptimizeQuality`].
+    ///
+    /// Currently, the only difference this makes is that when choosing [`RenderMode::OptimizeSpeed`],
+    /// rasterization will happen using u8/u16,
+    /// while [`RenderMode::OptimizeQuality`] will use a f32-based pipeline.
+    pub quality: RenderMode,
```
LaurenzV: Note to self: This should probably be renamed to `render_mode`.

nicoburns:
I would also ideally like to test this out, but I'm onboard with the API in principle.

LaurenzV:
Anything needed from my side to facilitate this, or do you mean “I’ll test this out once I find the time to do so”?

nicoburns:
I mean “I’ll test this out once I find the time to do so”. I would also be happy for this to land before that if you're confident in it.

LaurenzV:
I agree that the pixmap API should probably be refactored a bit (I'm leaning towards drawing inspiration from tiny-skia here), but I don't want to make this PR harder to review than it already is, so I will leave this for a follow-up.

nicoburns (approved):
I definitely have some quibbles with the API, but I've tested this against Blitz and it works fine (no unexpected surprises). So happy for this to land :)

nicoburns:
> I agree that the pixmap API should probably be refactored a bit (I'm leaning towards drawing inspiration from tiny-skia here), but I don't want to make this PR harder to review than it already is, so I will leave this for a follow-up.

I'd quite like to see `Pixmap`/`PixmapMut`/`PixmapRef` become it's own mini crate. That could then be used by `glifo` and could help remove the `vello_common` dep from `glifo`.

LaurenzV:
Yeah, as I mentioned, happy to iterate further in a future PR!

(Will still wait for Alex to have had a chance to take a look before merging.)

sparse_strips/vello_cpu/README.md
```diff
@@ -29,7 +29,7 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 [RenderContext::glyph_run]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.glyph_run
 [RenderMode::OptimizeSpeed]: https://docs.rs/vello_cpu/latest/vello_cpu/enum.RenderMode.html#variant.OptimizeSpeed
 [RenderMode::OptimizeQuality]: https://docs.rs/vello_cpu/latest/vello_cpu/enum.RenderMode.html#variant.OptimizeQuality
-[`RenderContext::render_to_pixmap`]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.render_to_pixmap
+[`RenderContext::render`]: https://docs.rs/vello_cpu/latest/vello_cpu/struct.RenderContext.html#method.render
```
grebmeg, line 32: I think we should also reflect this change in the CHANGELOG.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -622,35 +671,66 @@ impl RenderContext {
...
+    ///    be cleared, not just the area covered by the scene). One potential reason for doing this
+    ///    is that certain platforms, for example macOS, require a specific byte stride for buffers.
+    ///    For example, let's say that a byte stride of 128 is imposed by the platform, but the actual
+    ///    size of the scene you are drawing is only 20x20. In this case, you can create a pixmap
+    ///    of size 32x20, and the last 12 columns are essentially treated as padding.
+    ///
+    /// 3. In case the width/height of the pixmap is _smaller_ than the offset + width/height of the
+    ///    scene, then anything that exceeds the pixmap boundaries is simply cut off. This can be useful
+    ///    if for some reason you only want to rasterize a small cut-out of the original scene.
+    pub fn render<'a>(
```
grebmeg: To make this less invasive for consumers, who would otherwise need to add `RasterizerSettings::default()` in most cases, what do you think about providing a `render(&mut pixmap, &mut resources)` method that uses `CompositeMode::Replace` by default, and renaming the current method to `render_with(&mut pixmap, &mut resources, settings)`?

sparse_strips/vello_cpu/src/render.rs
grebmeg: > We could change the semantics. It just seems a bit unexpected to me that if the render context is smaller than the pixmap, any pixels that aren’t touched by the scene are left stale (which might also be relevant for our internal use case, in that case we would have to manually clear the pixmap before rendering into it).

Personally, I would expect `Replace` to behave as a region-local operator: the rendered region overwrites the destination, while pixels outside that region are left untouched. If the caller really wants a whole-buffer wipe, they can do that explicitly on their side with `pixmap.data_as_u8_slice_mut().fill(0)`. That makes the behaviour opt-in, explicit, and preserves the same cost.

There may also be cases where you want to avoid the cost of `SrcOver`, especially with unpacking wide tile buffers, but still want to replace only the rendered region without clearing the whole buffer, for example when rendering into a preprocessed image, frame, or another existing target.

grebmeg (commented):
LGTM

> TODO: I still want to run this version against my PDF test suite, but this shouldn't block a review.

Before merging, could you please verify this if you haven’t already?

grebmeg (approved):

LaurenzV:
> Before merging, could you please verify this if you haven’t already?

Already did yesterday, works fine!

sparse_strips/vello_cpu/README.md
LaurenzV: Okay sure, let me add an entry!

sparse_strips/vello_cpu/src/render.rs
LaurenzV: Makes sense,will do that

LaurenzV:
> Personally, I would expect Replace to behave as a region-local operator: the rendered region overwrites the destination, while pixels outside that region are left untouched. If the caller really wants a whole-buffer wipe, they can do that explicitly on their side with pixmap.data_as_u8_slice_mut().fill(0). That makes the behaviour opt-in, explicit, and preserves the same cost.

Hmm, okay. Let's do it this way: I'll add a TODO for this now to avoid additional churn for #1701, and once that has landed I will try to make that change on top of that. I _think_ it shouldn't be too hard, but I want to avoid making rebasing harder than it needs to be. 😄 I hope that's okay with you.

PR #1666, opened 2026-05-24
vello_hybrid: Fix fast rect rendering at u16 edges

dipeshbabu:
## Summary

- use wider arithmetic for fast-rect tile bounds at the u16 coordinate edges
- add regression tests for right and bottom edge rectangles ending at u16::MAX

## Testing

- cargo test -q -p vello_common render_edge_row_at_u16
- git diff --check


dipeshbabu:
@raphlinus can you review this pr?

LaurenzV:
Hello, sorry for leaving this PR uncommented for so long. I'll try to take a look soon. 

However, I think the problem is kind of that there are unfortunately _many_ parts of the code that currently break for the edge case of `u16::MAX` (some of which might be hard/impossible to properly fix), so I'm not sure if this PR by itself will solve all the problems. However, it might still be an improvement, so if it is we can probably still merge it.

LaurenzV:
Could you please rebase onto newest main? Then we shoudl be able to merge it.

LaurenzV:
Thanks!

LaurenzV (approved):

PR #1668, opened 2026-05-25
glifo: Fix interaction with complex paints

LaurenzV:
As part of our migration of glifo, we've basically completely broken the interaction between glyphs and complex paint transforms. This PR makes two changes to restore the expected behavior:
1) In a glyph run, we keep track of the original paint transform and later on apply a correction to make sure that each glyph is drawn with the correct paint transform.
2) We disable glyph caching for glyphs with complex paints since our image tinting API only supports solid color fills.

I added some more new tests to prevent future regressions. Thanks to @waywardmonkeys for finding this.

### DIFF
--- a/glifo/src/glyph.rs
+++ b/glifo/src/glyph.rs
@@ -2114,7 +2114,7 @@ mod tests {
     use crate::peniko::Blob;
     use crate::peniko::color::{AlphaColor, Srgb};
     use alloc::sync::Arc;
-    use vello_common::paint::{Image, ImageId, ImageSource, Tint};
+    use vello_common::paint::{Image, ImageId, ImageSource, PaintType, Tint};
 
     const _NORMALISED_COORD_SIZE_MATCHES: () =
         assert!(size_of::<skrifa::instance::NormalizedCoord>() == size_of::<NormalizedCoord>());
@@ -2136,6 +2136,8 @@ mod tests {
 
     #[derive(Default)]
     struct NoopRenderer;
+
+    static BLACK_PAINT: PaintType = PaintType::Solid(BLACK);
 
     struct TestResources {
         renderer: NoopRenderer,
@@ -2201,6 +2203,10 @@ mod tests {
             BLACK
         }
 
+        fn current_paint(&self) -> &PaintType {
+            &BLACK_PAINT
+        }
+
         fn atlas_image_source(&self, atlas_slot: &AtlasSlot) -> ImageSource {
             ImageSource::opaque_id(ImageId::new(atlas_slot.page_index))
         }
@@ -2248,11 +2254,13 @@ mod tests {
             AtlasCacher::Disabled
         };
 
+        let transform = Affine::translate((0.0, 20.0));
         let mut run = GlyphRun {
             font: font.clone(),
             font_size: 20.0,
             font_embolden: FontEmbolden::default(),
-            transform: Affine::translate((0.0, 20.0)),
+            transform,
+            absolute_paint_transform: transform,
             glyph_transform: None,
             normalized_coords: &[],
             hint: false,
--- a/glifo/src/glyph.rs
+++ b/glifo/src/glyph.rs
@@ -394,6 +394,7 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone> GlyphRunRenderer<'a, 'b, Gl
             // On a miss we keep both for reuse in the outline branch below.
             let outline_transform =
                 calculate_outline_transform(glyph, draw_props, hinting_instance);
+            let outline_draw_transform = outline_transform.pre_scale(scale_props.draw_scale);
 
             // We assume that the backend calculates the absolute paint transform
             // by concatenating scene transform and (relative) paint transform.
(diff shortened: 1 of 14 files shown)


waywardmonkeys:
I will leave the actual review for someone else, but you should probably include at least one test using an image brush as well.

waywardmonkeys:
Also, when I implemented brush transforms for glyphs for Vello Classic recently, I had some interesting tests that might be worth taking a peek at.

LaurenzV:
Sure, can do that!

glifo/src/glyph.rs
```diff
@@ -150,6 +151,8 @@ pub(crate) struct PreparedGlyph<'a> {
     pub(crate) glyph_type: GlyphType<'a>,
     /// The global transform of the glyph.
     pub(crate) transform: Affine,
+    /// The transform of the paint, relative to [`PreparedGlyph::transform`].
```
grebmeg: I think this is subtly wrong, since it’s actually relative to `transform.pre_scale(scale)`.

glifo/src/glyph.rs
```diff
@@ -150,6 +151,8 @@ pub(crate) struct PreparedGlyph<'a> {
     pub(crate) glyph_type: GlyphType<'a>,
     /// The global transform of the glyph.
     pub(crate) transform: Affine,
```
grebmeg: I think `transform` is a bit overloaded across this file. Would it be better to rename it to something like the following?

```rust
    /// Per-glyph outline transform: maps the draw-unit glyph outline
    /// (after font-size absorption) to scene coordinates.
    pub(crate) outline_transform: Affine,
```

glifo/src/glyph.rs
```diff
@@ -1358,6 +1379,8 @@ pub struct GlyphRun<'a> {
     font_embolden: FontEmbolden,
     /// Global transform.
     transform: Affine,
+    /// Absolute paint transform for the glyph run.
+    absolute_paint_transform: Affine,
```
grebmeg: Nit: "absolute" here doesn't refer to a world frame, just "the composition in scene space, before glifo's per-glyph rewrite". Renaming to `scene_paint_transform` would pair naturally with `relative_paint_transform` (which is in glyph-local space) and make the invariant readable: `scene_paint_transform == outline_draw_transform * relative_paint_transform`.

glifo/src/glyph.rs
```diff
@@ -389,6 +396,16 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone> GlyphRunRenderer<'a, 'b, Gl
...
+            let outline_draw_transform = outline_transform.pre_scale(scale_props.draw_scale);
+
+            // We assume that the backend calculates the absolute paint transform
+            // by concatenating scene transform and (relative) paint transform.
+            // (This is currently the case for Vello CPU / Vello Hybrid, but will
+            // also be assumed to be the case for any other potential backend.)
+            // Therefore, we can calculate the relative paint transform for
+            // the glyph by pre-concatenating it with the inverted outline transform.
+            let relative_paint_transform =
+                outline_draw_transform.inverse() * absolute_paint_transform;
```
grebmeg: When the outline cache hits (the common case for repeated characters), the code has already paid for an `Affine::inverse()` and `Affine` * `Affine` multiply that gets discarded. What about putting the computation into the outline-cache-miss branch, since that's the only place a `relative_paint_transform` produced here is consumed?

grebmeg: I’m worried we’re introducing quite a lot of math here on the hot path, especially for a general solid fill, since this math isn’t even needed in that case, can this be avoided at least for solid colors? And more generally, are there ways to avoid doing this math per glyph? One idea would be to factor out the run-constant linear part so that we end up with significantly fewer computations overall, although that would increase the complexity here.

Another idea, which seems preferable, would be to push this responsibility into the renderer, perhaps by introducing `set_paint_transform_absolute(t: Affine)` on `GlyphRenderer` which would flip the responsibility: glifo calls it once per run with `absolute_paint_transform`, the renderer stores it as an optional `Option<Affine>` on its `RenderState`, and every paint-encoding site uses that override directly when set instead of composing `transform · paint_transform`. Glifo can then freely mutate `state.transform` per glyph without any compensation, because the paint basis is pinned. The per-glyph math drops to zero, the inverse disappears entirely.

glifo/src/glyph.rs
LaurenzV: Right that's my bad, I initially based this branch on an old version of main where we didn't have this yet, and after rebasing I forgot to update the doc comment. 😦

glifo/src/glyph.rs
LaurenzV: Can do, will rename some other plain `transform`s while I'm at it.

glifo/src/glyph.rs
LaurenzV: 🤔 Not sure I fully understand this one, but willl make the change anyway. 🙂

glifo/src/glyph.rs
LaurenzV: I will move it down to the very bottom. I'm not against trying to optimize it in the future (if it turns out that this is indeed a bottleneck), but for now I think it's good to prioritize getting the fix for the bug out, since it's pretty bad. 😄

LaurenzV:
Added a test for image brushes as well.

grebmeg (approved):

glifo/src/glyph.rs
grebmeg: I agree that’s pretty bad, and it makes sense not to do it in this PR as well. But I’m curious to see the inverse optimized, since it’s not a cheap operation, and I think we can do more here in terms of performance.

PR #1671, opened 2026-05-26
vello_hybrid: Make probe more asynchronous

LaurenzV:
# Context

Right now, the probing API is fully synchronous: We first construct the scene, render into using WebGL, and then cause a full pipeline stall by directly requesting a read-back from GPU to CPU memory. While it's incredibly hard to get reliable measurements, this does seem to incur a large latency on low-tier devices. Here are some numbers (again, they can vary wildly from run to run for reasons unknown to me, but overall they are in the same ballpark):

Vivo Y21:
image

Xiaomi RedNote 11:
image

# Solution

Therefore, this PR proposes to switch to a slightly different approach: It makes use of WebGL Sync objects and pixel pack buffers to ensure that the vast majority of the work can happen asynchronously in the background without stalling the CPU. On the CPU side, we can then simply continuously poll the Sync object (for example in a RAF callback) to see whether the operation has finished (by calling `clientWaitSync with a timeout of 0, such that it returns immediately in case it's not done yet. I've done some crude measurements to confirm that this is very cheap). If it's not done yet, we can simply do other stuff, so the whole point is to avoid blocking while rendering is still in progress.

When applying this, basically the only synchronous parts are 1) scene construction + submitting commands to the GPU (`start_probe` in the screenshots below) and 2) doing the final read-back from the pixel pack buffer to the CPU (`readback` in the screenshots below). While the overall time can be longer sometimes, the synchronous parts seem to be much faster than before, therefore this seems like a win to me. Especially if we plan on extending the probe scene in the future.

Vivo Y21:
image

Xiaomi RedNote 11:
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -515,6 +515,9 @@ impl WebGlRenderer {
 
     #[cfg(feature = "probe")]
     fn probe_inner(&mut self) -> Result<WebGlPendingProbe, RenderError> {
+        // IMPORTANT NOTE: When making any changes to the probe, make sure to
+        // unignore and rerun the "webgl_probe_succeeds" test locally.
+
         let _state_guard = WebGlStateGuard::with_config(
             &self.gl,
             WebGlStateConfig {
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -613,6 +613,8 @@ impl WebGlRenderer {
             atlas_texture_array: probe_atlas_texture_array,
         };
 
+        // We do this here instead of above such that in case the render result is not
+        // valid, we still properly restore the state (e.g. the old atlas texture array).
         render_result?;
 
         let pending = launch_probe(&self.gl, probe_resources, width, height);
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -130,11 +130,6 @@ pub struct WebGlPendingProbe {
     gl: WebGl2RenderingContext,
     sync: Option<WebGlSync>,
     buffer: Option<WebGlBuffer>,
-    // Just to make sure they are kept alive until the probe
-    // finished. This is _probably_ not necessary because WebGl is a synchronous API
-    // so the `delete_` commands should only be executed once rendering has finished, but
-    // safe is safe.
-    _resources: WebGlProbeResources,
     width: u16,
     height: u16,
 }
@@ -2039,7 +2034,6 @@ fn launch_probe(
         gl: gl.clone(),
         sync: Some(sync),
         buffer: Some(pixel_pack_buffer),
-        _resources: resources,
         width,
         height,
     }
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -48,6 +48,8 @@ use bytemuck::{Pod, Zeroable};
 use core::fmt::Debug;
 #[cfg(feature = "text")]
 use glifo::{GLYPH_PADDING, PendingClearRect};
+#[cfg(feature = "probe")]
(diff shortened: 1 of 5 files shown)


LaurenzV:
I don't know why, but CI doesn't seem to like running the test. :/ I will have to look into it tomorrow...

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -119,6 +123,130 @@ pub struct WebGlRenderer {
...
+#[cfg(feature = "probe")]
+#[derive(Debug)]
+pub struct WebGlPendingProbe {
+    gl: WebGl2RenderingContext,
+    sync: Option<WebGlSync>,
+    buffer: Option<WebGlBuffer>,
+    // Just to make sure they are kept alive until the probe
+    // finished. This is _probably_ not necessary because WebGl is a synchronous API
+    // so the `delete_` commands should only be executed once rendering has finished, but
+    // safe is safe.
```
ajakubowicz-canva: Could this be expanded upon. Is my assumption that this comment means "kept alive via a reference to prevent garbage collection of resources"?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -119,6 +123,130 @@ pub struct WebGlRenderer {
...
+
+#[cfg(feature = "probe")]
+impl WebGlPendingProbe {
+    /// Try to finish the probe.
+    ///
+    /// In case the result is not available yet, a new pending probe object will be returned
+    /// which can be checked again in the future. Otherwise, the probe result or an error will be
+    /// returned.
+    pub fn try_finish(mut self) -> Result<WebGlProbeStatus, WebGlProbeError> {
+        let status = self.gl.client_wait_sync_with_u32(
```
ajakubowicz-canva, line 186: Luckily `clientWaitSync` is supported by [Safari 15](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/clientWaitSync#browser_compatibility).

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -474,21 +605,19 @@ impl WebGlRenderer {
...
+            gl: self.gl.clone(),
+            framebuffer: probe_framebuffer,
+            texture: probe_texture,
+            atlas_texture_array: probe_atlas_texture_array,
         };
 
-        self.gl.delete_framebuffer(Some(&probe_framebuffer));
-        self.gl.delete_texture(Some(&probe_texture));
-        Ok(pixels)
+        render_result?;
```
ajakubowicz-canva, line 616: Is it worth a comment that we check for a valid render result here because `WebGlProbeResources` drop frees resources?

Alternatively could the `WebGlProbeResources` be created above `render_scene` such that the `?` can be applied right there.

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: So basically, if we don't keep it around it would be dropped after probe_inner is done: https://github.com/linebender/vello/blob/a7916e4901028e8f380cdd268c63ab0cb0900c63/sparse_strips/vello_hybrid/src/render/webgl.rs#L616-L620

I was just a bit worried if it's a problem if the drop is already triggered at that point, however, I'm pretty sure it's fine because as mentioned in the comment, WebGL is conceptually synchronous, meaning that if the `fence_sync` and the transfer to pixel pack buffer is called before the end of the method, it is guaranteed that those operations have finished before the deletions are triggered. If you agree with that, then happy to revert that part!

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Yes this is the reason, will add a comment.

ajakubowicz-canva (approved):
Nice!

sparse_strips/vello_hybrid/src/render/webgl.rs
ajakubowicz-canva: I am happy to remove (also, I would really hope that tests would catch something like this if it's wrong after removal)

PR #1672, opened 2026-05-26
glifo: Make use of the outline cache for COLR glyphs

LaurenzV:
With #1629 merged, we now have a very solid cache for outlines in place! However, we were not using it at all for COLR glyphs. It is particularly important for those because COLR glyphs are often composed of the same base glyphs, just with different transforms/positions. This PR wires up the outline cache into COLR rendering. The results can be seen below:

Before: Zooming on a scene with ~700 COLR glyphs takes around 30ms:

https://github.com/user-attachments/assets/148cc62b-94da-4a2e-be1a-9580c4d1693e

After: Zooming on the same scene takes around 23ms:

https://github.com/user-attachments/assets/0a55ab12-4ea3-44fb-be29-f0501c36155a

### DIFF
--- a/glifo/src/colr.rs
+++ b/glifo/src/colr.rs
@@ -6,18 +6,22 @@
 use crate::atlas::commands::AtlasPaint;
 use crate::color::Srgb;
 use crate::color::{AlphaColor, DynamicColor};
-use crate::glyph::{GlyphColr, OutlinePath};
+use crate::glyph::{
+    CachedOutline, FontEmbolden, FontInfo, GlyphColr, OutlineCacheSession, OutlinePath,
+    VarLookupKey,
+};
 use crate::interface::DrawSink;
 use crate::kurbo::{Affine, Point, Rect, Shape};
 use crate::peniko::{self, BlendMode, ColorStops, Compose, Extend, Gradient, Mix};
 use crate::util::FloatExt;
+use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 use peniko::{LinearGradientPosition, RadialGradientPosition, SweepGradientPosition};
 use skrifa::color::{Brush, ColorPainter, ColorStop, CompositeMode, Transform};
 use skrifa::instance::LocationRef;
-use skrifa::outline::{DrawSettings, OutlineGlyphCollection, pen::ControlBoundsPen};
+use skrifa::outline::OutlineGlyphCollection;
 use skrifa::raw::TableProvider;
 use skrifa::raw::types::BoundingBox;
 use skrifa::{FontRef, GlyphId, MetadataProvider};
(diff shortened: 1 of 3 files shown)


glifo/src/colr.rs
```diff
@@ -276,18 +322,17 @@ impl ColorPainter for ColrPainter<'_> {
...
-        let _ = outline_glyph.draw(
-            DrawSettings::unhinted(skrifa::instance::Size::unscaled(), self.colr_glyph.location),
-            &mut self.clip_outline,
-        );
-
-        // Note that the bbox will become stale, but we don't need it anyway here.
+        // TODO: We can make use of `DrawSink::set_transform` to let the client
+        // take care of applying it, so we can avoid copying the path.
+        self.clip_outline.path.extend(outline.iter());
         self.clip_outline.path.apply_affine(self.cur_transform());
```
grebmeg, line 336: I agree, it’d be nice to optimize this away!

glifo/src/colr.rs
```diff
@@ -475,21 +520,14 @@ impl ColorPainter for GlyphInfoExtractor<'_> {
...
-        let mut outline_bbox = ControlBoundsPen::default();
-
-        // TODO: Make it possible to use the outline cache for this.
-        let Some(outline_glyph) = self.outline_glyphs.get(glyph_id) else {
-            return;
+        let outline_bbox = {
+            let Some(outline) = self.get_outline(glyph_id) else {
+                return;
+            };
+            outline.bbox
```
grebmeg, line 527: I noticed this changes the bbox semantics in a subtle way. The old `ControlBoundsPen` returned a control-hull bbox, but `outline.bbox` here is `BezPath::bounding_box()`, which evaluates precise curve extrema. It looks like skrifa doesn't flatten — it forwards `quad_to`/`curve_to` to the pen with the raw control points, and `ControlBoundsPen` just `min`/`max`es those coordinates without any derivative solve, so the two genuinely differ whenever a control point sits outside the swept curve region.

Tighter is arguably more correct for the COLR pixmap sizing fallback in `calculate_colr_metrics` (the clip path itself bounds the rendered content, not the control hull), but it’s a real behavior change that didn’t affect us previously, so I thought it was worth mentioning here.

glifo/src/glyph.rs
```diff
@@ -1219,7 +1265,7 @@ fn calculate_colr_metrics(
     };
 
     // TODO: Cache this across frames.
-    let colr_info = get_colr_info(font_ref, color_glyph, location);
+    let colr_info = get_colr_info(font_ref, color_glyph, location, outline_cache, font_info);
```
grebmeg, line 1268: Currently, the COLR paint tree is walked twice: first by `GlyphInfoExtractor` to compute `bbox` and `has_non_default_blend`, then again by `ColrPainter` for the actual painting. Ideally, we should avoid this duplication. I’m thinking about two possible optimizations here:
- Skip the `bbox` walk when the COLR table already provides `clip_box`. When it returns Some, `colr_info.bbox` is never used, so the entire `GlyphInfoExtractor` walk ends up doing outline lookups unnecessarily. Only `has_non_default_blend` would require the walk, which can be implemented as lightly as possible with an early return as soon as the first layer with a non-default blend is found. I think this optimisation makes sense if `clip_bbox` is mostly accessible.
- Cache `ColrGlyphInfo` across frames.

grebmeg (approved):
LGTM 🚀 Really nice perf optimisation!

Would it be possible, maybe as a follow-up, to add the scene you used in the video to the example scenes?

glifo/src/colr.rs
LaurenzV: Yeah, you are right. 👍 I could slightly expand the bbox just to be sure, but honestly I don't see this as being a problem. 😄

glifo/src/glyph.rs
LaurenzV: Yes, caching `ColrGlyphInfo` has been on my radar. The main reason I haven't done this yet is because we already have 2 LRU caches in the codebase, and I didn't want to add yet another one, but I will look into whether there is an easier way!

LaurenzV:
> Would it be possible, maybe as a follow-up, to add the scene you used in the video to the example scenes?

Can do that, the problem is that it requires downloading the full COLR emoji font (I don't think we want to check it in), but sounds like a reasonable ask.

PR #1673, opened 2026-05-26
Increase coarse winding histogram to i16

b0nes164:
Unfortunately this regresses performance. I ran it a couple times to be sure with the same results.

On tile_aaa/Ghostscript_Tiger (no culling):

Before:
time:   [40.833 µs 41.494 µs 42.172 µs]
                                       
After:
time:   [47.977 µs 48.304 µs 48.724 µs]

On tile_aaa_shift50/Ghostscript_Tiger (with culling):

Before:
time:   [34.752 µs 34.898 µs 35.071 µs]

After:
time:   [38.035 µs 38.533 µs 39.224 µs]

### DIFF
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -45,7 +45,7 @@ pub const MAX_LINES_PER_PATH: u32 = 1 << (32 - INT_MASK_SHIFT);
 pub struct CulledWindings {
     /// Fractional winding coverage for each individual scanline in a row.
     pub partial: Vec<[f32; Tile::HEIGHT as usize]>,
-    // Note that this will cause issues if we have windings greater/less than i8,
+    // Note that this will cause issues if we have windings greater/less than i6,
     // but this should only occur in pathological cases.
     /// Accumulated integer winding deltas for each tile row.
     pub coarse: Vec<i16>,
--- a/sparse_strips/vello_common/src/tile.rs
+++ b/sparse_strips/vello_common/src/tile.rs
@@ -45,7 +45,7 @@ pub const MAX_LINES_PER_PATH: u32 = 1 << (32 - INT_MASK_SHIFT);
 pub struct CulledWindings {
     /// Fractional winding coverage for each individual scanline in a row.
     pub partial: Vec<[f32; Tile::HEIGHT as usize]>,
-    // Note that this will cause issues if we have windings greater/less than i6,
+    // Note that this will cause issues if we have windings greater/less than i16,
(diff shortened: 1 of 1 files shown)


LaurenzV:
Well that is pretty unfortunate. 😔

LaurenzV:
Do we know where it's coming from? Is it from the clearing when resetting it?

b0nes164:
> Do we know where it's coming from? Is it from the clearing when resetting it?

I'm not sure. Theoretically, on the non-culling case, there should be almost no impact, because 1) nothing is adding into the histogram 2) the histogram is never cleared after the constructor.

My only guess is that the change in data type is large enough that we're spilling registers somewhere/it's stopping something from getting inlined.

LaurenzV:
Sad, I'll try to look into it myself as well when I find some time. Thanks!

sparse_strips/vello_common/src/tile.rs
```diff
@@ -45,10 +45,10 @@ pub const MAX_LINES_PER_PATH: u32 = 1 << (32 - INT_MASK_SHIFT);
 pub struct CulledWindings {
     /// Fractional winding coverage for each individual scanline in a row.
     pub partial: Vec<[f32; Tile::HEIGHT as usize]>,
-    // Note that this will cause issues if we have windings greater/less than i8,
+    // Note that this will cause issues if we have windings greater/less than i6,
```
nicoburns: Typo `i6` -> `i16`

LaurenzV:
@b0nes164 What machine were you running on? I tried it on M4 Max and I'm not seeing any change in performance. 🤔

b0nes164:
I'm on an M4 Pro?

LaurenzV:
Well that’s certainly weird 😅

b0nes164:
I think it has to land regardless for correctness. It'll probably show up in your perf data if it's a real issue?

nicoburns:
> I think it has to land regardless for correctness. It'll probably show up in your perf data if it's a real issue?

Yeah, I think this is likely a blocker for landing an upgraded Vello in Servo. Generally we can accept incorrect rendering for corner cases like this, but a crash is not ideal!

LaurenzV:
I think it should be fine to land, but it might be good to get a benchmark run from someone with an x86 machine as well.

b0nes164:
I have old intel uhd 610?630? + ubuntu at home, can test tonight

b0nes164:
About ~300us with culling, ~400us without, no change between i8/i16.

LaurenzV (approved):
Ok let's do this then!

PR #1674, opened 2026-05-27
vello_hybrid: Use RAII pattern for handling WebGL resources

LaurenzV:
This is a new PR, which does basically the same as #1574 but is based on the newest main, which has had quite a few changes. To explain the motivation again:

# Motivation

Right now, the WebGL backend of Vello Hybrid creates various resources, such as buffers, textures, etc., but we never actually destroy them once they go out-of-scope. This means that we are essentially relying on the garbage collector to properly clean those up. This _might_ be fine if we are only creating a single Vello Hybrid instance, but 1) it's bad practice and 2) might cause problems in case you want to have the ability to create multiple Vello Hybrid (for example to recreate it after a WebGL context loss).

# Implementation

Basically, what this PR proposes is to create a wrapper type for all resources, each resource implementing a trait that defines how to create the resource and how to drop it. We then make use of the `Drop` implementation of `Resource` to manually delete the resource again. This way, as soon as the variable goes out-of-scope (since it's not used anymore), the bound resource will be deleted as well. This way, we don't have to worry about manually deleting them anymore.

I should note that there is a small footgun we need to be careful about: It is still possible to write code that for example assumes that some resource has been bound previously and then acts on that. It could happen that we create a new resource, bind it in WebGL, then the variable goes out-of-scope and some other function later on still acts on that binding, assuming that it's still alive. I'm not sure what would happen in such a case, but it's certainly not good. As far as I can tell, we currently don't have any such code (I rewrote the `upload_data_to_rgba32_texture` method to let it take care of binding the resource it needs explicitly), but it's still something we should keep in mind when extending the backend in the future. I'm not sure if there is any way of preventing such code at compile-time via a different design. Nevertheless, I think this is a good start.

# Testing

I have tried the WebGL example scenes on my machine and also ran vello_bench2 on a couple of devices, and didn't observe any problems.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -46,6 +46,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
 use core::fmt::Debug;
+use core::ops::Deref;
 #[cfg(feature = "text")]
 use glifo::{GLYPH_PADDING, PendingClearRect};
 use resource::{Buffer, FragmentShader, Framebuffer, Program, Texture, VertexArray, VertexShader};
@@ -1974,7 +1975,7 @@ fn launch_probe(
     );
     gl.bind_framebuffer(
         WebGl2RenderingContext::FRAMEBUFFER,
-        Some(&resources.framebuffer),
+        Some(framebuffer.deref()),
     );
     gl.read_pixels_with_i32(
         0,
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -46,6 +46,8 @@ use alloc::vec;
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
 use core::fmt::Debug;
+#[cfg(feature = "probe")]
+use core::ops::Deref;
 #[cfg(feature = "text")]
 use glifo::{GLYPH_PADDING, PendingClearRect};
 use resource::{Buffer, FragmentShader, Framebuffer, Program, Texture, VertexArray, VertexShader};
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -575,8 +575,11 @@ impl WebGlRenderer {
             ),
         );
 
-        let previous_view_framebuffer = self.programs.resources.view_framebuffer_override.take();
-        self.programs.resources.view_framebuffer_override = Some(probe_framebuffer);
+        let previous_view_framebuffer = self
+            .programs
+            .resources
+            .view_framebuffer_override
+            .replace(probe_framebuffer);
         let render_result = self.render_scene(
             &scene,
             &mut probe_image_cache,
@@ -1284,8 +1287,8 @@ impl WebGlPrograms {
             // Replace the old resources
             self.resources.atlas_texture_array = new_atlas_texture_array;
             // Cached FBOs were attached to the old texture; drop them so we recreate on next use.
-            let _ = self.resources.atlas_render_framebuffer.take();
-            let _ = self.resources.filter_main_atlas_framebuffer.take();
+            self.resources.atlas_render_framebuffer = None;
+            self.resources.filter_main_atlas_framebuffer = None;
         }
     }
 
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -456,38 +457,33 @@ impl WebGlRenderer {
             ),
         );
 
-        let previous_view_framebuffer = self
-            .programs
-            .resources
-            .view_framebuffer_override
-            .replace(probe_framebuffer.clone());
+        let previous_view_framebuffer = self.programs.resources.view_framebuffer_override.take();
+        self.programs.resources.view_framebuffer_override = Some(probe_framebuffer);
```
grebmeg: Can we use this instead?

```suggestion
        let previous_view_framebuffer = self
            .programs
            .resources
            .view_framebuffer_override
            .replace(probe_framebuffer);
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1166,12 +1159,8 @@ impl WebGlPrograms {
...
             self.resources.atlas_texture_array = new_atlas_texture_array;
             // Cached FBOs were attached to the old texture; drop them so we recreate on next use.
-            if let Some(fb) = self.resources.atlas_render_framebuffer.take() {
-                gl.delete_framebuffer(Some(&fb));
-            }
-            if let Some(fb) = self.resources.filter_main_atlas_framebuffer.take() {
-                gl.delete_framebuffer(Some(&fb));
-            }
+            let _ = self.resources.atlas_render_framebuffer.take();
+            let _ = self.resources.filter_main_atlas_framebuffer.take();
```
grebmeg: Would the following make the intent clearer?

```suggestion
self.resources.atlas_render_framebuffer = None;
self.resources.filter_main_atlas_framebuffer = None;
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -2832,9 +2797,9 @@ impl RendererBackend for WebGlRendererContext<'_> {
                         .programs
                         .resources
                         .filter_main_atlas_framebuffer
-                        .get_or_insert_with(|| self.gl.create_framebuffer().unwrap());
+                        .get_or_insert_with(|| Framebuffer::new(self.gl));
                     self.gl
-                        .bind_framebuffer(WebGl2RenderingContext::FRAMEBUFFER, Some(fb));
+                        .bind_framebuffer(WebGl2RenderingContext::FRAMEBUFFER, Some(&*fb));
```
grebmeg: Can this be simplified a bit?

```suggestion
                        .bind_framebuffer(WebGl2RenderingContext::FRAMEBUFFER, Some(fb));
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -3178,3 +3144,173 @@ fn upload_data_to_rgba32_texture(
...
+    }
+
+    // `Resource` intentionally does not implement `Clone`. There should
+    // only be a single handle to a given resources, such that it has
+    // unique ownership and we don't end up deleting the same resource
+    // twice.
+    impl<T: GlResource> Resource<T> {
+        pub(super) fn new(gl: &WebGl2RenderingContext) -> Self {
+            let raw =
+                T::create(gl).unwrap_or_else(|| panic!("failed to create WebGL {}", T::LABEL));
```
grebmeg, line 3375: I’m a bit concerned about adding more static utf bytes embedded in the wasm binary through strings in `expect` and `panic`, both in this PR and previously. I don’t have a strong answer here, other than using strings sparingly and possibly adding a post-processing step for production builds to strip this data.

grebmeg (approved):

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Hmm yeah maybe! 😄

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: Oops, thanks for catching this!

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: I'm not sure about this, those are just a few bytes per string so I don't think it should matter that much? 🤔 But in any case, if we want to do something about it, let's do it in a follow-up that fixes it for all locations.

grebmeg: It’s fine for now, but this can grow little by little across multiple PRs over time and it could become a problem. I think the preprocessing step added downstream would fix it, but let’s postpone that work for the future.

PR #1676, opened 2026-05-27
vello_hybrid: Two tweaks to the native WebGL example

LaurenzV:
First, this PR adjusts the README  to encourage using the SIMD version by default when running the example, since it's significantly faster.

Next, panning and zooming has always been a bit broken in the WebGL example. When panning, the scene would move much slower than the cursor, and zooming would also not always zoom into the actual position of the mouse cursor.

https://github.com/user-attachments/assets/e8a9b768-bd6d-42d9-8854-6412bf916e95

I tried to have Codex fix it and it seems to work. I've not tried to understand this more deeply, but since this is not production code I hope this is fine to land without too much bikeshedding.

https://github.com/user-attachments/assets/c43e8cec-a91c-44bf-b9c3-d33da517de93

### DIFF
--- a/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
@@ -432,7 +432,6 @@ pub async fn run_interactive(canvas_width: u16, canvas_height: u16) {
             );
         }) as Box<dyn FnMut(_)>);
         let opts = web_sys::AddEventListenerOptions::new();
-        opts.set_passive(false);
         canvas
             .add_event_listener_with_callback_and_add_event_listener_options(
                 "wheel",
--- a/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
+++ b/sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
@@ -34,6 +34,22 @@ impl RendererWrapper {
     }
 }
 
+fn client_to_canvas_px(canvas: &HtmlCanvasElement, client_x: f64, client_y: f64) -> Vec2 {
+    let rect = canvas.get_bounding_client_rect();
+    let width = rect.width().max(1.0);
+    let height = rect.height().max(1.0);
+    let x = ((client_x - rect.left()) / width).clamp(0.0, 1.0);
+    let y = ((client_y - rect.top()) / height).clamp(0.0, 1.0);
+    Vec2::new(x * canvas.width() as f64, y * canvas.height() as f64)
+}
+
+fn client_delta_to_canvas_px(canvas: &HtmlCanvasElement, delta_x: f64, delta_y: f64) -> Vec2 {
+    let rect = canvas.get_bounding_client_rect();
+    let scale_x = canvas.width() as f64 / rect.width().max(1.0);
+    let scale_y = canvas.height() as f64 / rect.height().max(1.0);
+    Vec2::new(delta_x * scale_x, delta_y * scale_y)
+}
+
 /// State that handles scene rendering and interactions
 struct AppState {
     scenes: Box<[AnyScene<Scene>]>,
@@ -43,6 +59,7 @@ struct AppState {
     transform: Affine,
     mouse_down: bool,
     last_cursor_position: Option<Vec2>,
+    last_drag_client_position: Option<Vec2>,
     width: u32,
     height: u32,
     renderer_wrapper: RendererWrapper,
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -397,11 +423,22 @@ pub async fn run_interactive(canvas_width: u16, canvas_height: u16) {
...
+            app_state.borrow_mut().handle_wheel(
+                event.client_x() as f64,
+                event.client_y() as f64,
+                event.delta_y(),
+                event.ctrl_key(),
+                event.delta_mode(),
+            );
         }) as Box<dyn FnMut(_)>);
+        let opts = web_sys::AddEventListenerOptions::new();
+        opts.set_passive(false);
```
ajakubowicz-canva: Why set_passive to `false` explicitly? Based on the usage of the event callback, I feel like explicitly setting to `true` could be better.

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
```diff
@@ -34,6 +34,22 @@ impl RendererWrapper {
...
+    let rect = canvas.get_bounding_client_rect();
+    let width = rect.width().max(1.0);
+    let height = rect.height().max(1.0);
+    let x = ((client_x - rect.left()) / width).clamp(0.0, 1.0);
+    let y = ((client_y - rect.top()) / height).clamp(0.0, 1.0);
+    Vec2::new(x * canvas.width() as f64, y * canvas.height() as f64)
+}
+
+fn client_delta_to_canvas_px(canvas: &HtmlCanvasElement, delta_x: f64, delta_y: f64) -> Vec2 {
+    let rect = canvas.get_bounding_client_rect();
```
ajakubowicz-canva, line 47: Both `canvas.get_bounding_client_rect()` can trigger a browser layout (if something else changed and dirtied browser layout). This can be bad for performance but I _think_ this should be fine. (Usually layout calculations like this are cached if max performance is desired).

For what it is worth, I think the problem is that the full screen canvas has a different resolution based on the browsers `devicePixelRatio`.

Quick glossary for this explanation.
 - logical pixels is what CSS provides. E.g. I made a 10px by 10px div.
 - physical pixels is what the browser renders on the screen. On a `2` devicePixelRatio` display, that same 10px by 10px div will be rendered as 20 physical pixels by 20 physical pixels.

Why does this matter?
Because `canvas.get_bounding_client_rect()` returns **logical** pixels. Similarly `client_x` is in **logical** pixels.

Thus you get a fractional amount of width and height for where the cursor is. However, the `canvas.width` and `canvas.height` properties are in **physical** pixels. Hence, on a display with `2` devicePixelRatio, a canvas logical pixels such as what `get_bounding_client_rect` returns may be something like 100px by 100px. But, for the canvas visuals to look good, the actual amount of physical drawn pixels will mean that the `canvas.width()` and `canvas.height()` respectively would now be 200px by 200px.

Anyways – this is why this code works and is robust. Technically - it would be more performant to cache the bounding client rect unless it is resized, but this is fine.

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
LaurenzV: Let me ask codex. 😆 Sorry about that, as mentioned I haven't tried to understand the semantics of this PR, will look into it and see whether it can just be removed.

ajakubowicz-canva (approved):
Non production code so nothing blocking. Only little comments. Great stuff and great improvements!

sparse_strips/vello_hybrid/examples/native_webgl/src/lib.rs
ajakubowicz-canva: No stress: My reference is:
 - https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#passive

In the web standard a little lower in this document: https://dom.spec.whatwg.org/#interface-eventtarget
I've taken a screenshot and attached.
Screenshot 2026-05-28 at 11 33 24 pm

ajakubowicz-canva: Hence - I recon it's worth setting to true if it's left in, but since no browser scrolling is actually happening, probably fine to just delete it.

LaurenzV: I removed the call now.

PR #1680, opened 2026-05-28
Tweaks to WASM CI for sparse strips

LaurenzV:
This PR introduces two changes to our CI for testing sparse strips in CI:

1) Instead of having one single job, we split them up into three jobs: The first job for checking the examples, the second for running the tests without WASM and the third for running the tests with WASM. This cuts the "bottleneck" time for this test specifically from around 13 minutes to around 6-7 minutes. This won't have an immediate effect since the overall bottleneck is still Windows CI, but it's an easy change and will likely be worth it in the future once we reduce the time of Windows CI.

2) Use MacOS-based runners instead of Ubuntu for running the WASM tests. When working on #1671, I have encountered the problem that the probe test does not pass anymore [due to some weird timeout](https://github.com/linebender/vello/actions/runs/26457569847/job/77896141078).

image

I've spent more than a day trying to debug this now, but have not gotten further than concluding that for some reason, invoking asynchronous methods like `setTimeout` or `requestAnimationFrame` causes those timeouts to occur, regardless of what happens in there (even when removing the actual probing, it still happened when I invoked those methods). Locally everything works fine, and apparently it also works fine when switching to a MacOS-based runners. I'm fairly certain that this is some weirdness with the Github CI runners and not with probing failing completely on Linux (though I admittedly haven't tested it since I don't have a linux laptop), so I think this is the easiest thing to do for now. I'd also like to get to the bottom of this, but I've already spent more time on this than I can really justify, so I hope that this is fine doing for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,9 +321,19 @@ jobs:
         run: cargo test --doc --workspace --locked --all-features --no-fail-fast
 
   test-stable-wasm:
-    name: cargo test (wasm32)
+    name: cargo test (wasm32, ${{ matrix.name }})
     needs: prime-lfs-cache
-    runs-on: ubuntu-latest
+    # We use MacOS because for some reason the Ubuntu-based runners
+    # have trouble handling things like `requestAnimationFrame` or
+    # `setTimeout` in our testing code.
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        include:
+          - name: default
+            rustflags: ""
+          - name: simd128
+            rustflags: "-Ctarget-feature=+simd128"
     steps:
       - uses: actions/checkout@v6
         # We intentionally do not use lfs: true here, instead using the caching method to save LFS bandwidth.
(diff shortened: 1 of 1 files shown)


ajakubowicz-canva (approved):
Nice!

.github/workflows/ci.yml
```diff
@@ -321,9 +321,19 @@ jobs:
         run: cargo test --doc --workspace --locked --all-features --no-fail-fast
 
   test-stable-wasm:
-    name: cargo test (wasm32)
+    name: cargo test (wasm32, ${{ matrix.name }})
     needs: prime-lfs-cache
-    runs-on: ubuntu-latest
+    # We use MacOS because for some reason the Ubuntu-based runners
+    # have trouble handling things like `requestAnimationFrame` or
+    # `setTimeout` in our testing code.
```
ajakubowicz-canva, line 328: This magic is quite unfortunate but I agree that being pragmatic makes sense and it's not worth burning all the time on this.

LaurenzV:
@raphlinus @DJMcNab I think I will need one of you again to merge this to circumvent the bug where CI endlessly waits for a check that isn't going to happen 😦

xStrom:
That's not a bug, it's a feature. The old test job is marked as mandatory and removing it from the CI script doesn't remove that requirement.

Anyway, I removed that requirement and will mark these new ones as required.

PR #1682, opened 2026-05-28
Export webgl features even on non-wasm32

jesses-canva:
We need to be able to reference these without gating our code on `#[cfg(target_arch = "wasm32")]`.

Tested with

```
cargo hack check -p vello_encoding --feature-powerset
cargo hack check -p vello --feature-powerset
cargo hack check -p glifo --feature-powerset --at-least-one-of std,libm
cargo hack check -p vello_common --feature-powerset --at-least-one-of std,libm
cargo hack check -p vello_hybrid --feature-powerset --at-least-one-of std,libm
cargo hack check -p vello_cpu --feature-powerset --at-least-one-of std,libm --at-least-one-of u8_pipeline,f32_pipeline
```

### DIFF
--- a/sparse_strips/vello_cpu/Cargo.toml
+++ b/sparse_strips/vello_cpu/Cargo.toml
@@ -35,7 +35,7 @@ std = ["vello_common/std", "glifo?/std"]
 # Use floating point implementations from libm.
 libm = ["vello_common/libm", "glifo?/libm"]
 # Allow loading Pixmap from PNG, and drawing png glyphs.
-png = ["dep:png", "vello_common/png", "glifo?/png"]
+png = ["std", "dep:png", "vello_common/png", "glifo?/png"]
 # Enable multi-threaded rendering.
 multithreading = [
     "std",
--- a/sparse_strips/vello_hybrid/src/render/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/mod.rs
@@ -11,18 +11,18 @@
 pub(crate) mod common;
 #[cfg(feature = "probe")]
 mod probe;
-#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+#[cfg(feature = "webgl")]
 mod webgl;
 #[cfg(feature = "wgpu")]
 mod wgpu;
 
 pub use common::{Config, GpuStrip, RenderSize};
 
-#[cfg(all(target_arch = "wasm32", feature = "webgl", feature = "probe"))]
+#[cfg(all(feature = "webgl", feature = "probe"))]
 pub use vello_common::probe::{Probe, ProbeResult};
-#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+#[cfg(feature = "webgl")]
 pub use webgl::{WebGlAtlasWriter, WebGlRenderer, WebGlTextureWithDimensions};
-#[cfg(all(target_arch = "wasm32", feature = "webgl", feature = "probe"))]
+#[cfg(all(feature = "webgl", feature = "probe"))]
 pub use webgl::{WebGlPendingProbe, WebGlProbeError, WebGlProbeStatus};
 #[cfg(feature = "wgpu")]
 pub use wgpu::{AtlasWriter, RenderTargetConfig, Renderer, TextureBindings};
--- a/sparse_strips/vello_hybrid/src/render/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/mod.rs
@@ -6,7 +6,7 @@
 //! ## Renderer Backends
 //!
 //! - `wgpu` contains the default renderer backend, leveraging `wgpu`.
-//! - `webgl` contains a WebGL2 backend specifically for `wasm32` if the `webgl` feature is active.
+//! - `webgl` contains a WebGL2 backend if the `webgl` feature is active.
 
 pub(crate) mod common;
 #[cfg(feature = "probe")]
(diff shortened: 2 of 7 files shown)


sparse_strips/vello_hybrid/src/render/mod.rs
```diff
@@ -11,18 +11,18 @@
 pub(crate) mod common;
```
grebmeg, line 11: Could you please update the module documentation related to webgl above? I think it’s a bit misleading now, since `wasm32` is not used here anymore.

sparse_strips/vello_hybrid/src/render/mod.rs
```diff
@@ -11,18 +11,18 @@
...
 #[cfg(feature = "probe")]
 mod probe;
-#[cfg(all(target_arch = "wasm32", feature = "webgl"))]
+#[cfg(feature = "webgl")]
 mod webgl;
```
grebmeg, line 15: What do you think about adding the following `debug_assert!` in `WebGlRenderer::new_with`, right next to `maybe_warn_about_webgl_feature_conflict`?

```rust
debug_assert!(
    cfg!(target_arch = "wasm32"),
    "`WebGlRenderer` can only be constructed when targeting `wasm32`",
);
```

grebmeg (approved):
Thanks @jesses-canva!

LGTM if LGT @LaurenzV

sparse_strips/vello_cpu/Cargo.toml
```diff
@@ -35,7 +35,7 @@ std = ["vello_common/std", "glifo?/std"]
 # Use floating point implementations from libm.
 libm = ["vello_common/libm", "glifo?/libm"]
 # Allow loading Pixmap from PNG, and drawing png glyphs.
-png = ["dep:png", "vello_common/png", "glifo?/png"]
+png = ["std", "dep:png", "vello_common/png", "glifo?/png"]
```
LaurenzV, line 38: Is this change really needed? Seems weird that vello_cpu would be affected by this. 🤔

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -250,6 +250,16 @@ impl WebGlRenderer {
...
+        #[allow(
+            clippy::assertions_on_constants,
+            reason = "intentional guard against non-wasm32 use"
+        )]
+        {
+            debug_assert!(
+                cfg!(target_arch = "wasm32"),
+                "`WebGlRenderer` can only be constructed when targeting `wasm32`",
+            );
+        }
```
LaurenzV, line 262:
```suggestion
        const {
            assert!(
                cfg!(target_arch = "wasm32"),
                "`WebGlRenderer` can only be constructed when targeting `wasm32`",
            );
        }
```

Could we do this instead? 1) Make it a normal assert so it also triggers in release build and 2) make it a const block so the assertion triggers at compile time? Or do we require the ability to compile it, just not run it?

jesses-canva: We need to be able to compile it, otherwise the `cargo check` that VS Code runs wont work, so this has to be a runtime assertion.

sparse_strips/vello_cpu/Cargo.toml
jesses-canva: It would be an existing issue that was surfaced by the `--feature-powerset` testing I did (mentioned in OP). Let me find the exact command to reproduce the error.

LaurenzV:
Thanks!

sparse_strips/vello_cpu/Cargo.toml
jesses-canva: Okay it was to fix this which was broken on `main` (before this merged)

```
$ cargo check --no-default-features --features f32_pipeline,libm,png -p vello_cpu
error[E0432]: unresolved import `vello_common::peniko::kurbo::common::FloatFuncs`
   --> sparse_strips/vello_cpu/src/filter/drop_shadow.rs:25:5
    |
 25 | use vello_common::peniko::kurbo::common::FloatFuncs as _;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `FloatFuncs` in `common`
    |
note: found an item that was configured out
   --> /Users/jesses/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/kurbo-0.13.1/src/common.rs:117:1
    |
117 | / define_float_funcs! {
118 | |     fn abs(self) -> Self => fabs/fabsf;
119 | |     fn acos(self) -> Self => acos/acosf;
120 | |     fn atan2(self, other: Self) -> Self => atan2/atan2f;
...   |
137 | |     fn trunc(self) -> Self => trunc/truncf;
138 | | }
    | |_^ the item is gated here
    = note: this error originates in the macro `define_float_funcs` (in Nightly builds, run with -Z macro-backtrace for more info)

...
```

Adding `"std"` to the features of `png` fixed it:

```diff
diff --git a/sparse_strips/vello_cpu/Cargo.toml b/sparse_strips/vello_cpu/Cargo.toml
index c7081b21..45ecf32a 100644
--- a/sparse_strips/vello_cpu/Cargo.toml
+++ b/sparse_strips/vello_cpu/Cargo.toml
@@ -35,7 +35,7 @@ std = ["vello_common/std", "glifo?/std"]
 # Use floating point implementations from libm.
 libm = ["vello_common/libm", "glifo?/libm"]
 # Allow loading Pixmap from PNG, and drawing png glyphs.
-png = ["dep:png", "vello_common/png", "glifo?/png"]
+png = ["std", "dep:png", "vello_common/png", "glifo?/png"]
 # Enable multi-threaded rendering.
 multithreading = [
     "std",
```

```
$ cargo check --no-default-features --features f32_pipeline,libm,png -p vello_cpu
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
```

PR #1653, opened 2026-05-16
vello_cpu: Add u8 fast path for some blend modes

LaurenzV:
Mostly generated with codex, but I did look at it myself and make some adjustments, so I hope it's good now. Since we do have quite a few tests for blending (both manual ones as well as via COLR), not too concerned about correctness issues here.

Note that this does not address https://github.com/linebender/vello/pull/1579 yet so it's possible this won't have much effect on AVX2. However, on NEON I'm seeing 4x-5x speedups for blending now:

```
fine/blend/normal_u8_neon
                        time:   [40.096 ns 40.297 ns 40.521 ns]
                        change: [-1.1748% +0.8265% +3.1392%] (p = 0.45 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe

fine/blend/multiply_u8_neon
                        time:   [976.39 ns 978.78 ns 981.34 ns]
                        change: [-80.790% -80.728% -80.664%] (p = 0.00 < 0.05)
                        Performance has improved.

fine/blend/screen_u8_neon
                        time:   [1.0140 µs 1.0167 µs 1.0199 µs]
                        change: [-80.496% -80.407% -80.320%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

fine/blend/overlay_u8_neon
                        time:   [1.3631 µs 1.3667 µs 1.3701 µs]
                        change: [-74.682% -74.592% -74.499%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

fine/blend/darken_u8_neon
                        time:   [1.1359 µs 1.1385 µs 1.1412 µs]
                        change: [-77.273% -77.197% -77.125%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

fine/blend/lighten_u8_neon
                        time:   [1.1535 µs 1.1557 µs 1.1582 µs]
                        change: [-77.013% -76.936% -76.857%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

fine/blend/color_dodge_u8_neon
[...]

PR #1684, opened 2026-05-29
vello_cpu: Fix potential overflows in new blending code

LaurenzV:
#1653 added a new fast path for performing blending with u8/u16. Something that was missed here was that in many cases, we can exceed the technically allowed maximum of 255. In the previous f32 path, this wasn't a problem because converting a float larger than 255.0 to u8 it will automatically be clamped to, but this is not the case when doing all of blending using u8. I noticed this when trying to use the newest version of vello_cpu in my PDF renderer, where some pixels turned black.

This PR fixes this by
1) Ensuring all pixels are clamped to 255.
2) Additionally, ensuring that all RGB values are <= the alpha, which is necessary for premultiplied colors.

With this fix applied, I didn't notice any other regressions in my test suite.

First commit adds failing tests, second commit fixes the issue.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/lowp/blend.rs
+++ b/sparse_strips/vello_cpu/src/fine/lowp/blend.rs
@@ -222,3 +222,78 @@ fn with_src_alpha<S: Simd>(simd: S, rgb: u8x32<S>, src_c: u8x32<S>) -> u8x32<S> 
 
     (rgb & !alpha_mask) | (src_c & alpha_mask)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::peniko::Compose;
+    use vello_common::fearless_simd::Fallback;
+
+    fn lowp_first_pixel_for_mix(blend: Mix, src: [u8; 4], bg: [u8; 4]) -> [u8; 4] {
+        let simd = Fallback::new();
+        let src = u8x32::from_slice(simd, &src.repeat(8));
+        let bg = u8x32::from_slice(simd, &bg.repeat(8));
+        let blend_mode = BlendMode::new(blend, Compose::SrcOver);
+        let res = mix(src, bg, blend_mode);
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_cpu/src/fine/lowp/blend.rs
```diff
@@ -213,12 +219,91 @@ fn hard_light_inner<S: Simd>(src_c: u8x32<S>, bg_c: u8x32<S>, condition: u8x32<S
...
+        let lowp = lowp_first_pixel_for_mix(mix, src, bg);
+        let highp = highp_first_pixel_for_mix(mix, src, bg);
+
+        for (component, (&lowp, &highp)) in lowp.iter().zip(highp.iter()).enumerate() {
+            let delta = lowp.abs_diff(highp);
+            assert!(
+                delta <= MAX_DELTA,
+                "{mix:?} component {component} differed by {delta}: lowp={lowp}, highp={highp}, src={src:?}, bg={bg:?}"
+            );
+        }
```
waywardmonkeys, line 291: The old behavior could produce invalid premultiplied colors, but still within the MAX_DELTA = 2, so that exact failure isn't picked up by this test, I think.

I think explicit assertions that `low[0..3]` are all `<= lowp[3]` and perhaps that `lowp[3] == src[3]` would be useful and have helped pick up the failure that is being fixed before.

waywardmonkeys (commented):
A small request for one of the tests ... also, some of other fast paths changed but didn't get tested here. Not sure if there's anything behind that.

sparse_strips/vello_cpu/src/fine/lowp/blend.rs
LaurenzV: Done, added an explicit assertion for the alpha part. Also added the remaining tests, everything should be covered now.

waywardmonkeys (approved):

PR #1687, opened 2026-05-30
Version bump for glifo + sparse strips

LaurenzV:


waywardmonkeys (approved):
As long as the PR numbers are correct, this looks good to me.

PR #1678, opened 2026-05-27
Fix feature propagation

Shnatsel:
**Background:** 

On x86 the SIMD code will not lower into the SIMD instructions it's supposed to unless it is inside a function annotated with `#[target_feature(enable = ...)]`. See [here](https://github.com/linebender/fearless_simd/tree/main/fearless_simd#inlining) for more info.

It is possible that some paths just get inlined enough thanks to LLVM heuristics, but that situation is very fragile and may cause wild swings on benchmarks due to unrelated changes. So it is a good idea tighten this up even if it gave no change in performance. My particular interest is in landing https://github.com/linebender/fearless_simd/pull/236 which will certainly alter the heuristic inlining.

Subsumes #1579

**What this PR does:**

1. Makes the `#[vello_bench]` macro run the inner function under `vectorize` so that the required target features are actually set and the SIMD code can lower into the instructions it's supposed to
2. Mass-converts all functions generic over `Simd` to be `#[inline(always)]` and calls `vectorize` inside them to create an inlining barrier. 

I also tried to just annotate everything generic over Simd with `#[inline(always)]` to see what would happen, but that made compile times blow up.

**Performance**

Gains: tile_aaa improves 1.5x to 2x, blending improves 2x to 20x, minor gains in glyphs, 

Losses: some single/short benchmarks are slower. ~~fine/strip/solid_medium_u8_avx2 and fine/blend/normal_u8_avx2 regress by 20%~~ cannot reproduce, probably measurement fluke

Nonsense: something is very wrong with rounded_blurred_rect, it is 3x slower under AVX2 than under SSE4.2 even on main.

<details><summary>Full run on x86 (Zen4)</summary>
<p>

```

allocator/alloc_1000_varied_8_128
                        time:   [284.81 µs 284.91 µs 285.01 µs]
                        change: [-2.8860% -2.8003% -2.7298%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  10 (10.00%) low severe
  2 (2.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe

allocator/alloc_until_full_32x32
                        time:   [325.59 µs 325.62 µs 325.66 µs]
                        change: [-2.0880% -2.0624% -2.0350%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  5 (5.00%) high severe

[...]

PR #1688, opened 2026-05-30
sparse_strips: Better inlining of SIMD kernels for improved x86 performance

LaurenzV:
This is an alternative to #1678, while trying to be a bit more surgical about where to add inlines and where not. I measured this on my x86 machine, and similarly got 85%-95% better performance for blending. Image rendering of high quality also improved by 85%. Other stuff mostly unchanged.

#1678 applied vectorization in the benchmark functions as well, but this should not be needed because we just call `fill` and any method inside of that will have a corresponding vectorize call, so that should be enough. This way, we also don't have to inline it.

Commit 1 just applies inline to three kernel methods (although it seems like they mostly already where inlined).

Commit 2 adds vectorize methods to the `new` methods of the painter structs. I decided to choose this route instead of force inlining and forcing the parent to do the vectorization because those methods are only called very few times anyway, so this seems like the better route. The compiler can then decide itself whether to inline or not.

Commit 3 makes sure that the `next` methods of all iterators are inlined. In this case, it is the responsibility of the parent to make sure vectorization is in place. We want these to be inlined because those will be called for each pixel chunk, so it's important to not have function call overhead here.

Commit 4 Applies fixes to make sure target features are properly propagated for blending.

Commit 5 applies some more small fixes that I missed before.

There is obviously a bit of bikeshedding you could do about in which cases you want the method to contain `vectorize` and in which cases you force inline it and have the parent be responsible for providing the vectorization, but I think the current state should be good for now.

### DIFF
--- a/sparse_strips/vello_common/src/clip.rs
+++ b/sparse_strips/vello_common/src/clip.rs
@@ -230,6 +230,7 @@ pub fn intersect(
 ///
 /// This is all that this method does. It just looks more complicated as the logic for iterating
 /// in lock step is a bit tricky.
+#[inline(always)]
 fn intersect_impl<S: Simd>(
     simd: S,
     path_1: PathDataRef<'_>,
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -985,6 +985,7 @@ pub trait FromF32Color: Sized + Debug + Copy + Clone {
 impl FromF32Color for f32 {
     const ZERO: Self = 0.0;
 
+    #[inline(always)]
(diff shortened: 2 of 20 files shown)


Shnatsel (approved):
Looks good by an large.

Could you also add comments on functions that have neither vectorize() nor inline(always) that this is deliberate and that the callee contains a call to vectorize()?

Here's a list of remaining functions without annotations:

encode.rs (line 781): u8_lut, f32_lut
multi_threaded.rs (line 361): rasterize_with
single_threaded.rs (line 129): rasterize_with, rasterize_with_filters, process_layer_tile, rasterize_simple, composite_at_offset_with

single_threaded can probably use one comment instead of annotating every function.

PR #1689, opened 2026-05-30
Adapt to opaque mask representation

Shnatsel:
This makes vello compatible with both current fearless_simd release and the git version that removed block-based ops from masks.

This makes experimentation with the git version easier.


tomcur (approved):
Thanks!

PR #1691, opened 2026-05-31
vello_common: Use binary search to determine the start index in clipping

LaurenzV:
Based on #1690. I think this should mostly be self-explanatory with the comments. This brings the clipping time in that particular example from 4.4% to 2.6%:

image

### DIFF
--- a/sparse_strips/vello_common/src/clip.rs
+++ b/sparse_strips/vello_common/src/clip.rs
@@ -375,6 +375,7 @@ fn intersect_impl<S: Simd>(
     }
 }
 
+#[inline(always)]
 fn first_strip_at_or_after(strips: &[Strip], strip_y: u16) -> usize {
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_common/src/clip.rs
```diff
@@ -361,6 +374,12 @@ fn intersect_impl<S: Simd>(
     }
 }
 
+fn first_strip_at_or_after(strips: &[Strip], strip_y: u16) -> usize {
```
taj-p, line 379: Should we set this to inline always?

taj-p (approved):

sparse_strips/vello_common/src/clip.rs
LaurenzV: Sure!

PR #1696, opened 2026-06-03
Sparse Strips: Ignore unimplemented filters rather than panicking

nicoburns:
Changes made

- [x] Make Vello CPU (singlethreaded) ignore unimplemented filters instead of panicking
- [x] Make Vello Hybrid ignore unimplemented filters instead of panicking (replaces filters with 0.0 offset filter)
- [x] Make Vello CPU (multithreaded) ignore filter layers rather than panicking

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/mod.rs
+++ b/sparse_strips/vello_cpu/src/fine/mod.rs
@@ -651,7 +651,6 @@ impl<S: Simd, T: FineKernel<S>> Fine<S, T> {
             }
             Cmd::PushZeroClip(_) | Cmd::PopZeroClip => {
                 // These commands are handled by the dispatcher and should not reach fine rasterization
-                unreachable!();
             }
             Cmd::BatchEnd => {}
         }
--- a/sparse_strips/vello_hybrid/src/filter.rs
+++ b/sparse_strips/vello_hybrid/src/filter.rs
@@ -36,7 +36,7 @@ use vello_common::filter::drop_shadow::DropShadow;
 use vello_common::filter::flood::Flood;
 use vello_common::filter::gaussian_blur::{DecimationSizer, GaussianBlur, MAX_KERNEL_SIZE};
 use vello_common::filter::offset::Offset;
-use vello_common::filter_effects::EdgeMode;
+use vello_common::filter_effects::{EdgeMode, Filter, FilterPrimitive};
 use vello_common::kurbo::{Affine, Vec2};
 use vello_common::paint::{ImageId, ImageSource};
 use vello_common::peniko::{ImageQuality, ImageSampler};
@@ -770,7 +770,12 @@ impl FilterContext {
                 let width = wtile_bbox.width_px() as u32;
                 let height = wtile_bbox.height_px() as u32;
 
-                let instantiated = PreparedFilter::new(filter, transform);
+                let instantiated = PreparedFilter::new(filter, transform).unwrap_or_else(|| {
+                    // Removing filters from the graph is awkward, so we replace unsupported filters
+                    // with a zero-distance offset filter which has no effect on the output.
+                    let f = Filter::from_primitive(FilterPrimitive::Offset { dx: 0.0, dy: 0.0 });
+                    PreparedFilter::new(&f, transform).expect("Offset filter is known to be valid")
+                });
                 let gpu_filter = GpuFilterData::from(&instantiated);
                 let is_multi_pass = gpu_filter.is_multi_pass();
 
--- a/sparse_strips/vello_common/src/filter/mod.rs
+++ b/sparse_strips/vello_common/src/filter/mod.rs
@@ -34,16 +34,16 @@ pub enum PreparedFilter {
 
 impl PreparedFilter {
     /// Build a new prepared filter for the given transform.
-    pub fn new(filter: &Filter, transform: &Affine) -> Self {
+    pub fn new(filter: &Filter, transform: &Affine) -> Option<Self> {
         // Multi-primitive filter graphs are not yet implemented.
         if filter.graph.primitives.len() != 1 {
-            unimplemented!("Multi-primitive filter graphs are not yet supported");
+            return None;
         }
 
         match &filter.graph.primitives[0] {
             FilterPrimitive::Flood { color } => {
                 let flood = Flood::new(*color);
-                Self::Flood(flood)
+                Some(Self::Flood(flood))
             }
             FilterPrimitive::GaussianBlur {
                 std_deviation,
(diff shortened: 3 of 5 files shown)


sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -651,7 +651,6 @@ impl<S: Simd, T: FineKernel<S>> Fine<S, T> {
             }
             Cmd::PushZeroClip(_) | Cmd::PopZeroClip => {
                 // These commands are handled by the dispatcher and should not reach fine rasterization
-                unreachable!();
             }
```
nicoburns, line 654: Possibly this should be removed? But presumably something earlier in the pipeline would need to be tweaked to make that work properly?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -770,7 +770,11 @@ impl FilterContext {
                 let width = wtile_bbox.width_px() as u32;
                 let height = wtile_bbox.height_px() as u32;
 
-                let instantiated = PreparedFilter::new(filter, transform);
+                let Some(instantiated) = PreparedFilter::new(filter, transform) else {
+                    // If we don't support the specified filter, we skip the node.
+                    // Later on, we skip nodes with no associated filter layer gracefully.
+                    continue;
```
nicoburns: Turns out this does not in fact skip these nodes gracefully in the current implementation.

sparse_strips/vello_cpu/src/fine/mod.rs
LaurenzV: Don't worry about it, this will go away in my soon-to-land rewrite.

sparse_strips/vello_hybrid/src/filter.rs
nicoburns: I've switched to replacing this with a Offset filter with 0.0 offset. Bit of a hack, but it eliminates the panics.

taj-p:
What's the context for this change? Wouldn't this mean a consumer has no way of knowing whether their rendering is correct or not without referencing the implementation?

nicoburns:
> What's the context for this change? Wouldn't this mean a consumer has no way of knowing whether their rendering is correct or not without referencing the implementation?

My high-level requirement here is that Blitz doesn't panic when it visits a webpage that specifies a filter that vello doesn't support. And it also needs to do that while pushing rendering through an abstraction (AnyRender) that supports Skia (which does implement all of the filters).

The status quo is that a consumer has no way of knowing whether their rendering will panic without referencing the implementation? I think the difference here may be between known vs. user generated content. If you can validate your inputs then panicking is potentially useful. If your inputs are runtime generated then you're liable to end up in a situation where your app works fine locally but crashes in production!

---

The really critical part of this is removing the panic from the multi-threaded variant of Vello CPU. This is because that causes the set of filters that triggers a panic to change within the same crate (just by someone setting a feature flag), and there is no reliable way for me to prevent or detect that feature flag being set.

(I can split this out into it's own PR if need be)

I was able to work around the other part (where the renderers crash on unimplemented filters) in the AnyRender backend because the set of filters that work is known (by reading the source code), so I can filter them out before passing them into Vello. This doesn't seem ideal to me (and doesn't help users using Vello's API directly), but wouldn't be blocking for me/Blitz.

LaurenzV:
If @taj-p has no concerns with removing the panics, fine with me.

However, I think it might be easier to simply replace the filter with `None` when doing push_layer in case the backend doens't support it, then it shouldn't be necessary to for example remove the `unreachable` in the PopZeroClip branch. It would also be better if `PreparedFilter` didn't return `None`, instead just putting a log warning in case the filter graph is larger than one and then just process the first filter in the chain.

taj-p:
Oof. This is a hard one. Blitz doesn't know what renderer features a given website will support and we don't expose state for you to understand whether we're multithreaded or not. This means that either Blitz panics at runtime when a filter isn't supported or must ignore all filters for the Vello backend. Both options aren't great. Panics at runtime are awful and incorrect output from the Vello backend isn't also great for consumers who need accurate outputs.

The current state is not great.

> The status quo is that a consumer has no way of knowing whether their rendering will panic without referencing the implementation?

The current implementation signals whether the user has broken the contract of a caller (using an unimplemented filter) by runtime panicking. The proposal is to hide that signal to avoid the runtime panic. But, then the caller has no way of knowing whether their rendering is correct or not. 

I think this is the part I have worry about - I don't want callers to be in a state where they don't know whether a rendering is correct or not. I think this is a philosophical principle we've heretofore abided by - all consumers should know whether we think their rendering is correct.

As soon as we push that requirement to callers, then callers enter this really brittle state of needing to pore over source code / changelogs to understand contracts.

The current state isn't great - panicking at runtime and definitely should be better - but I'd rather keep some signal than remove it altogether.

> The really critical part of this is removing the panic from the multi-threaded variant of Vello CPU. This is because that causes the set of filters that triggers a panic to change within the same crate (just by someone setting a feature flag), and there is no reliable way for me to prevent or detect that feature flag being set.

Would an alternative option here be to expose that state for the `AnyRender` backend to filter on? Consumers could feasibly decide whether they want you to `Error` on unimplemented filters or noop (as is proposed) and this can be handled at that layer?

The other option, albeit more complex, is to pass a flag to Vello on whether to return `Error` or noop on unimplemented render features. I'm not sure I'm in favour of this option, but I'll defer to others on preference. (We could return `Error`s here, but then we wouldn't get the noop behaviour you're looking for).

taj-p:
We could feature flag the panic or not panic decision as an alternative low effort solution if that helps

nicoburns:
The really critical thing for me is the ability to avoid panicking in the vello-cpu-with-multithreading. This is trickier than the other cases, because in this case the panicking is enabled by a feature flag (the `multithreading` feature), and it is currently impossible for Blitz to determine whether something else in the crate graph might have enabled that feature flag. I have put up a PR that only disables the panic based on the multithreading feature flag (https://github.com/linebender/vello/pull/1702)

I think runtime detection could also work for this. If `vello_cpu::Renderer` exposed a method to detect whether multithreading is enabled then the AnyRender backend could use that to remove filters from the it's input. That would solve the issue for AnyRender consumers (but I still don't think panicking is a great experience for users in general!)

I guess a dedicated feature flag could work too so long as it was opt-in.

---

Finally, this would be more work to change, but I think the idiomatic Rust thing to do in this case would be for the render method to return `Result`. Perhaps optionally with some kind of `fail_fast: bool` runtime configuration option that allows the user to choose whether Vello should attempt to continue to render the rest of the scene or exit early in the case of failure.

LaurenzV:
I think this has been superseded by #1703 for now. Of course, we still want to implement a proper fix in the future.

PR #1697, opened 2026-06-04
sparse_strips: Fix opaque image optimizations when blending/masks are active

LaurenzV:
A while ago, we added optimizations for fully opaque images. On the one hand, during coarse rasterization, an image that covers the full wide tile can trigger clearing all previous commands. On the other hand, during fine rasterization we also use a fast path for images that don't have any transparency. However, those fast paths are not valid if there is either a mask or a blend mode active. This PR fixes it.

First commit adds the (wrong-looking) tests, second commit fixes the issue and updates the tests.


taj-p (approved):
Nice catch!

PR #1700, opened 2026-06-07
[classic] Support more than 256 bins

b0nes164:
Changes binning to accept more than 256 bins.

After intersecting with the clip, each workgroup processes the bins in blocks of 256. Although this can be a little slow, for almost all cases the binning will be resolved in <= 2 blocks.

### DIFF
--- a/vello_shaders/shader/binning.wgsl
+++ b/vello_shaders/shader/binning.wgsl
@@ -74,7 +74,7 @@ fn main(
         return;
     }
 
-    // Read inputs and determine coverage of bins
+    // Read inputs and determine coverage of bins for the draw object global_id.x
     let element_ix = global_id.x;
     var x0 = 0;
     var y0 = 0;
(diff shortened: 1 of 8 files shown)


b0nes164:
Bruce or Raph, could you TAL? I *think* the wasm failure is unrelated?

vello_shaders/shader/binning.wgsl
```diff
@@ -113,69 +113,94 @@ fn main(
...
-            x = x0;
-            y += 1;
+
+    // Loop over the blocks of bins in chunks of N_TILE (256)
+    for (var block_start = 0u; block_start < n_bins; block_start += N_TILE) {
+        let next_block = block_start + N_TILE;
+
+        // Write coverage for the current block
+        for (var y_offset = y0_width; y_offset < y1_width; y_offset += width_in_bins) {
+            let start_bin = u32(y_offset + x0);
```
raphlinus: It seems like you could make this more efficient by taking the max with block_start here, and the min with next_block for end_bin, removing the need for the if check inside the for loop.

vello_shaders/shader/binning.wgsl
```diff
@@ -113,69 +113,94 @@ fn main(
...
+        sh_chunk_offset[local_id.x] = chunk_offset;
+
+        let header_ix = wg_id.x * aligned_n_bins + cur_bin_ix;
+        bin_header[header_ix].element_count = element_count;
+        bin_header[header_ix].chunk_offset = chunk_offset;
+        workgroupBarrier();
+
+        // Output the element indices to bin_data
+        for (var y_offset = y0_width; y_offset < y1_width; y_offset += width_in_bins) {
+            let start_bin = u32(y_offset + x0);
```
raphlinus: Same as above.

raphlinus (approved):
Thanks for this! It turned out to be quite a bit easier than I was fearing.

Inline is a suggestion which I think should improve performance, but I'm not sure how much impact it'll have.

raphlinus:
The CI failure seems unrelated. You should be able to rerun the failed job, but it's something we should fix. Also, I should mention I did use Fable to help me review.

PR #1701, opened 2026-06-08
vello_cpu: A full rewrite of the frontend

LaurenzV:
# Notes
I apologize and recognize that this PR is very hard to review, as it changes so much and touches many different files. However, I fundamentally also don’t see a way of splitting this up into multiple smaller PRs, as it’s basically a full rewrite of most core parts of Vello CPU, which are very interdependent and hard to split out. Therefore, the best way to review this PR is probably not to focus too much on individual file diffs, but instead really treat this as a from-scratch implementation and trying to understand how the individual components work, and how this maps to how Vello CPU previously worked.

This PR also closes the long-standing issue #1119, although only for Vello CPU, not Vello Hybrid.

# TLDR
To summarize what this PR does in a few sentences and why: It completely weans Vello CPU off our current coarse rasterization strategy (basically anything that relates to `Wide` and `WideTile` in `vello_common`), instead using a simpler scheme that is specific to Vello CPU. To put it briefly, the reason this new coarse rasterization scheme is simpler is that 1) It only splits vertically, but does not perform horizontal splitting into buckets of width 256 and 2) It removes the old clipping algorithm, instead moving to a much simpler, although slower, layer clipping scheme. Since we now have the faster `push_clip_path` method, I don’t see this as a concern.

The core achievements of this refactor are 1) Overall less complicated logic (well, there _is_ still a lot of complexity, but in my opinion the main complexity only stems from handling filter layers correctly; if we didn’t have to do that, like is the case currently on main, I believe things in this PR would become much simpler), 2) Filter layers are now supported properly, even at viewport edges or for content outside of the viewport (!!) and 3) While this was not a core motivation for this PR, it turns out that performance actually also improves all across the board!

# Motivation
As was already discussed on various occasions (e.g. [Zulip]([#vello > Removing coarse-rasterization based clipping](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Removing.20coarse-rasterization.20based.20clipping/with/592893644))), our current coarse rasterization algorithm has been a pretty big pain point in Vello Hybrid. On the one hand because it hamstrings performance, but on the other hand also because it adds a lot of code complexity, making it prone to bugs that are challenging to fix. The fact that we have the specialized layer clipping algorithm and also inline filter layer rendering commands into the wide tiles leads to a lot of very hard to grasp workarounds for edge cases, for example [here](https://github.com/linebender/vello/blob/dd9a6cce233a09fd5fff3678b3fff49675a7767d/sparse_strips/vello_cpu/src/dispatch/single_threaded.rs#L318-L389). Therefore, we have decided that we are likely to move Vello Hybrid away from coarse rasterization. However, there isn’t really much point in just doing that, because as long as Vello CPU still depends on that implementation, we can’t get rid of it and are stuck with it! Therefore, in order to really get this through, we also need to rethink how we handle rendering on the Vello CPU side. This PR proposes a way of achieving this.

# Core algorithm
Let’s compare how Vello CPU worked up until now, and how things work with this PR. 

On current main, there are two stages: Coarse rasterization and fine rasterization. As the user submits fill/stroke commands, we first generate the strips, then turn those strips into render commands that are bucketed into wide tiles of size 256x4, and during fine rasterization we simply iterate over each wide tile and execute the rendering commands, in the end packing the fine buffer with the composited results into the pixmap. As mentioned, a very big problem with this approach are filter layers. Conceptually, some filter layers require the ability to sample neighboring pixels. Because of this, you fundamentally have to render filter layers _as a whole_ first, apply the filter to the _whole_ pixmap, and then sample from that. This is fundamentally incompatible with our approach of splitting the render target up into chunks of 256x4 and trying to render each part independently. Because of this, we use this very brittle approach where inside of each wide tile we store what range of commands belong to each layer, and in the end we try to extract those to render filter layers separately, and skip over those command ranges when rendering the parent layer. To put it briefly, the fact that we try to inline filter layer commands into a single stream with the commands of the parent layer causes a lot of pain. It’s also incompatible with the fact that filter layers might need a larger viewport than our main render target. Because of this, we currently have the bug where filters are not properly applied at the edges of the viewport.

With this PR, the overall flow is slightly different: As the user submits fll/stroke commands, we still generate the strips eagerly. However, instead of already splitting them up into pixmap-local tiles, we simply create a flat buffer of “recorded commands” that just store the range of strips that belong to that command. It’s a simple 1D vector of commands that does not try to turn them into bucketed render commands. Because of this, implementing filter layers is actually quite easy: We simply push a new stack to the context, and all render commands that belong to this filter layer will be stored separately from the commands of the root layer. This way, we don’t conflate the render commands of different render targets, and we can also easily account for the differences in viewport size between different layers.

Only once we get to fine rasterization do we actually start doing some form of bucketing of commands. However, this happens _per root layer_, so each filter layer and the root layer will be bucketed independently from each other. This happens using a `CommandBucketer` which can be resized and therefore reused when rasterizing multiple layers. A core difference to current main is that we don’t split by 256x4 anymore. Instead, we simply split by viewport_width x 4. This means that we don’t do any horizontal splitting anymore. My rationale behind this change is that doing this kind of splitting significantly increases coarse rasterization time and also makes the logic more complex. The original reasoning _for_ doing this is the argument that doing compositing in a smaller 256x4 buffer increases cache locality. However, my benchmarks strongly suggest that this is either not true, or any gain in performance that comes from that is offset by the increased complexity of coarse rasterization. Therefore, I think that this is overall a good change to make!

Above is a rough summary of how the core flow changes. However, there are some more cool implementation details that are worth elaborating on (for example the depth buffer!), hence I will summarize them more extensively below. I will not dive into the nitty-gritty implementation details because I’ve already tried to explain those in the form of doc comments.

# Core changes

## vello_common
- Filter bounds are now split into filter_expansion and source_expansion, so Vello CPU can distinguish the area where filtered output may appear from the extra source area that must be rendered to compute the filter correctly. This distinction is important for correctly rendering drop shadows at the edge of the viewport.
- Added a RetainVec structure that conceptually allows arbirarily resizing 2D arrays without losing the underlying allocations.
- Some other small helpers.

## vello_cpu
- As mentioned, the CPU pipeline now has an explicit recording stage before doing bucketing of commands. Each root/filter layer gets their own recorded command stream, so they are no longer interleaved into the parent command stream.
- As also mentioend further above, the old 256x4 wide-tile coarse path is replaced by a CommandBucketer which buckets only by strip row.
- The clipping algorithm has been replaced by a different approach instead where we simply use bounded blend commands to implement the same logic. The only downside is that we don’t have the optimization from existing coarse rasterization where we save the work of a blending operation for inner 256x4 wide tiles. But this optimization was problematic in its own way, so I think this is a good thing. And as mentioned, if people want fast clipping they should use `push_clip_path` instead.
- The fine command model was also revamped by consolidating different commands. Instead of many specialized commands, row commands are now mostly PaintFill, LayerFill (the previous BlendFill), PushBuf, and PopBuf. Filter layers will be treated as normal image fills.
- Inspired by @taj-p excellent optimization in Vello Hybrid, I’ve introduced a new depth buffer optimization to save unnecessary pixel work in certain situations. It’s basically the equivalent of the `background` optimization that current exists for wide tiles, except that we can affort applying it at a finer granularity than before (128 pixels vs. 256 pixels before).
- Since we don’t split commands horizontally anymore, fine rasterization now happens in full row bands. We use the depth buffer to reduce compositing work and also be smarter about which parts of the buffer we actually need to clear before packing (as it turns out from some profiling I did, this can be pretty expensive for larger resolutions).
- The layer manager that was used for filter layers is gone and instead replaced by a simpler FilterContext.

I _think_ this summarizes the main changes. Obviously there are lots more implementation details, but those are better to understand from the code itself.

# How to review this PR
As mentioned, I don’t think it makes much sense to review this PR with just file diffs, because most files have been completely rewritten. Instead, I would rather recommend just looking at the files themselves with a fresh mind. I personally recomment the following order:
- Any changes to vello_common. Those should be relatively independent from the rest.
- Changes to `[render.rs](http://render.rs/)` as well as the new `[record.rs](http://record.rs/)`. Those additions relate to the very first part of the new pipeline.
- Any new file in `coarse`, as those are also relatively independent from the rest, in particular the new depth buffer.
- And then finally, the changes in `fine` as well as `dispatch`.

# Filter layers
Filter layers work correctly now, see the video below!

https://github.com/user-attachments/assets/8efac8c0-a4c9-42d9-99de-7d1585613433

# Performance
Primarily, as part of this refactor my main concern was not introducing _regressions_ in performance. I was unsure whether the new recording step that precedes coarse rasterization will add significant overhead. However, to my surprise, the opposite is actually the case! Most likely thanks to less overhead during coarse rasterization, I’m seeing lots of time improvements, especially for larger shapes.

## Blend2D benchmarks
First, you can check the old/new results from the Blend2D benchmark harness here: https://laurenzv.github.io/vello_chart/

As you can see, for FillRectA and FillRectU we are now much faster at larger geometry sizes, getting very close to Blend2D performance! For smaller shape sizes, the results are overall the same. Multi-threading is too noisy to really draw a conclusion, but results mostly seem unchanged.

## M4 Max benchmarks
Next, i ran some benchmarks on my M4 Max laptop usign different work loads. Two workloads are normal vector graphics, six of the workloads are designs that I downloaded from Canva with different characteristics (much text, many images, etc.).

As you can see, we are seeing decent gains in many cases, especially for larger sizes.

| Workload | Size | Old total | New total | Speedup |
| --- | ---: | ---: | ---: | ---: |
| gs-tiger | 200x200 | 0.527 ms | 0.514 ms | +2.5% |
|  | 500x500 | 1.063 ms | 0.992 ms | +6.7% |
|  | 1000x1000 | 2.030 ms | 1.846 ms | +9.1% |
|  | 3000x3000 | 6.927 ms | 5.582 ms | +19.4% |
| coat-of-arms | 165x200 | 1.982 ms | 1.962 ms | +1.0% |
|  | 413x500 | 2.897 ms | 2.850 ms | +1.6% |
|  | 827x1000 | 4.321 ms | 4.073 ms | +5.8% |
|  | 2480x3000 | 11.129 ms | 9.694 ms | +12.9% |
| design-1 | 200x113 | 0.169 ms | 0.165 ms | +2.5% |
|  | 500x281 | 0.504 ms | 0.481 ms | +4.5% |
|  | 1000x563 | 1.476 ms | 1.417 ms | +4.0% |
|  | 3000x1688 | 10.516 ms | 9.993 ms | +5.0% |
| design-2 | 200x113 | 0.061 ms | 0.058 ms | +5.1% |
|  | 500x281 | 0.204 ms | 0.195 ms | +4.4% |
|  | 1000x563 | 0.651 ms | 0.631 ms | +3.1% |
|  | 3000x1688 | 5.137 ms | 5.037 ms | +1.9% |
| design-3 | 200x113 | 0.498 ms | 0.486 ms | +2.4% |
|  | 500x281 | 1.114 ms | 1.094 ms | +1.8% |
|  | 1000x563 | 3.194 ms | 3.111 ms | +2.6% |
|  | 3000x1688 | 37.511 ms | 32.900 ms | +12.3% |
| design-4 | 200x113 | 0.595 ms | 0.575 ms | +3.3% |
|  | 500x281 | 0.850 ms | 0.825 ms | +3.0% |
|  | 1000x563 | 1.432 ms | 1.359 ms | +5.2% |
|  | 3000x1688 | 5.362 ms | 4.834 ms | +9.8% |
| design-5 | 200x113 | 0.641 ms | 0.630 ms | +1.7% |
|  | 500x281 | 0.750 ms | 0.728 ms | +3.0% |
|  | 1000x563 | 0.986 ms | 0.945 ms | +4.2% |
|  | 3000x1688 | 2.489 ms | 2.281 ms | +8.3% |

## Samsung Galaxy Tab A6.
I did some very basic experimentation with my tablet to see whether the improvements also hold on WASM. And it seems like they do:

Current main:
- GhostScript Tiger: 122ms
- Coat of Arms: 244ms
- Design 1: 175ms

This PR:
- GhostScript Tiger: 113ms
- Coat of Arms: 228ms
- Design 1: 166ms

## Intel i9 14990k
Here are the numbers, overall a similar pattern! :)

| Workload | Size | Old total | New total | Speedup |
| --- | ---: | ---: | ---: | ---: |
| gs-tiger | 200x200 | 1.601 ms | 1.564 ms | +2.3% |
|  | 500x500 | 2.876 ms | 2.768 ms | +3.8% |
|  | 1000x1000 | 4.944 ms | 4.720 ms | +4.5% |
|  | 3000x3000 | 15.161 ms | 14.065 ms | +7.2% |
| coat-of-arms | 165x200 | 5.233 ms | 5.246 ms | -0.2% |
|  | 413x500 | 7.371 ms | 7.330 ms | +0.6% |
|  | 827x1000 | 10.575 ms | 10.232 ms | +3.2% |
|  | 2480x3000 | 24.654 ms | 22.689 ms | +8.0% |
| design-1 | 200x113 | 0.442 ms | 0.434 ms | +1.8% |
|  | 500x281 | 1.363 ms | 1.282 ms | +6.0% |
|  | 1000x563 | 4.158 ms | 3.887 ms | +6.5% |
|  | 3000x1688 | 28.928 ms | 27.022 ms | +6.6% |
| design-2 | 200x113 | 0.142 ms | 0.138 ms | +2.9% |
|  | 500x281 | 0.527 ms | 0.512 ms | +2.8% |
|  | 1000x563 | 1.770 ms | 1.705 ms | +3.7% |
|  | 3000x1688 | 14.368 ms | 13.434 ms | +6.5% |
| design-3 | 200x113 | 1.453 ms | 1.438 ms | +1.0% |
|  | 500x281 | 3.225 ms | 3.153 ms | +2.3% |
|  | 1000x563 | 12.384 ms | 12.114 ms | +2.2% |
|  | 3000x1688 | 87.415 ms | 76.070 ms | +13.0% |
| design-4 | 200x113 | 1.603 ms | 1.594 ms | +0.5% |
|  | 500x281 | 2.440 ms | 2.389 ms | +2.1% |
|  | 1000x563 | 4.001 ms | 3.882 ms | +3.0% |
|  | 3000x1688 | 14.783 ms | 13.315 ms | +9.9% |
| design-5 | 200x113 | 1.578 ms | 1.571 ms | +0.4% |
|  | 500x281 | 1.959 ms | 1.926 ms | +1.7% |
|  | 1000x563 | 2.601 ms | 2.495 ms | +4.1% |
|  | 3000x1688 | 6.578 ms | 6.013 ms | +8.6% |

# Disadvantages
There are some disadvantages with this PR, but I think they are all acceptable (for now):
- ~~It is unclear to me how much memory usage will be impacted. Previously, strips were only retained on a per-path basis, now we retain the strips of _all_ paths. It’s possible that there is higher memory usage now, since we retain _all_ strips of all paths with our new recording approach. However, since we don’t split commands horizontally anymore, we have less overall bucketed commands, so this should provide some reduction. If we really wanted to, I believe it should also be possible to push down the size to 8 bytes. So I don’t think this is a blocking concern.~~ Turns out that the opposite is actually the case! I checked the used memory when rendering GS Tiger at 4K in vello_cpu_winit, and the max memory usage goes from about 100MB to 92MB!
- I’ve already covered that clip layers might be slower if they affect a large area, but that should also not matter since `push_clip_path` is the recommended clipping approach anyway.
- I have decided to temporarily disable support for the `EdgeMode` in gaussian blurs, as things got a bit more complicated when implementing this. It’s definitely possible, but this PR is already big enough. The feature is not supported in Hybrid either, so I hope it’s not a big deal.
- Relatedly, we do have the problem now that there is a mismatch between Hybrid and CPU due to Hybrid still having the incorrect filter behavior. Therefore, I’ve had to temporarily disable some tests for Hybrid. We still have many other tests that still work, so I’m hoping this is not a big deal as a temporary measure.

# Testing
I've made sure that
1. Existing tests pass (execpt for the expected differences).
2. All of the interactive scenes seem to still work correctly.
3. My PDF test suite passes with this branch.

Hopefully this will give us enough confidence that we don’t accidentally introduce any regressions. :)

# Future work

There is also some interesting follow-up work that is unlocked if this PR is merged, but in the interest of not blowing the PR up further I won't work on this now.
- It should now be much easier to make `RenderContext` resizable. This means that even if you resize our window, you can just reuse the same `RenderContext` and don't have to allocate a new one.
- I think another interesting direction to explore is whether we can completely get rid of bucketing, instead rendering the strips directly into the Pixmap. I did some very brief experimentation with this, and while it seemed like it might improve performance for smaller sizes (since we can completetly get rid of packing, which can take up a lot of time), as soon as the render target is very large (> 1k), it becomes very noticeable that the lack of cache locality makes fine rasterization overall a lot slower. I'm not gonna pursue this further as part of this PR, but I think it's still an interesting direction to explore in the future.
- It should also be easier to implement some optimizations like inlining unnecessary push/pop layer commands, something I also want to look into!


sparse_strips/vello_sparse_tests/snapshots/clip_clear.png
LaurenzV, line 1: This diff is expected and how it should look like. The `Clear` blend should be constrained by the clip path of the layer.

https://github.com/linebender/vello/issues/1119#issuecomment-3106309203

sparse_strips/vello_sparse_tests/snapshots/filter_drop_shadow_offscreen.png
LaurenzV, line 1: New test that demonstrates drop shadow works even if the shape is above the viewport.

sparse_strips/vello_sparse_tests/snapshots/filter_rotated_blur.png
LaurenzV, line 1: The diff here is that the green part at the bottom got darker. However, I believe this is expected behavior and seems to be a bug on current main relating to the clipping logic. I tried making all green on main, and this is what I get:

Image

The one at the bottom is indeed much lighter than the other ones, even though there really shouldn't be a reason since the other ones are all the same in the darker green shade.

sparse_strips/vello_sparse_tests/snapshots/issue_filter_canvas_boundaries.png
LaurenzV, line 1: Here it's also expected that the transparent part disappears.

sparse_strips/vello_bench/src/fine/pack.rs
```diff
@@ -4,70 +4,39 @@
...
 use vello_common::tile::Tile;
-use vello_cpu::fine::{Fine, FineKernel, SCRATCH_BUF_SIZE};
-use vello_cpu::region::Regions;
+use vello_cpu::fine::{Fine, FineKernel};
 use vello_dev_macros::vello_bench;
 
+// 256 was the original width of a wide tile, hence we are keeping it for consistency.
+const TILE_BUFFER_SIZE: usize = 256 * Tile::HEIGHT as usize * 4;
+
 pub fn pack(c: &mut Criterion) {
```
LaurenzV, line 17: Packing has changed now so that we always handle as many 4-pixel chunks as possible and only the final bit with `pack_tail`, so it doesn't make a lot of sense to keep those as separate benchmarks now.

sparse_strips/vello_common/src/clip.rs
```diff
@@ -160,39 +155,6 @@ impl ClipContext {
     }
 }
 
-/// Compute a conservative bounding box for the transformed path by computing the bounding box of
-/// the transformed control points.
-///
-/// If `path` is empty, this returns an infinite, inversed [`Rect`] (`left` > `right` and `top` > `bottom`).
-fn control_point_bbox(path: &BezPath, transform: Affine) -> Rect {
```
LaurenzV, line 167: This got moved into the `util` module instead, since it's now used from other places as well.

sparse_strips/vello_common/src/util.rs
```diff
@@ -124,3 +129,185 @@ pub fn extract_scales(transform: &Affine) -> (f32, f32) {
...
+
+/// A type that can be cleared.
+pub trait Clear {
+    /// Clear the object to its default state.
+    fn clear(&mut self);
+}
+
+/// A resizable vector that retains inner elements upon resizing.
+#[derive(Debug)]
+pub struct RetainVec<T> {
```
LaurenzV, line 187: This it the new `RetainVec` abstraction briefly introduced in the PR description.

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
```diff
@@ -345,6 +304,116 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
...
+
+        let loaded = simd.load_interleaved_128_u32x16(casted).to_bytes();
+        let (loaded_lo, loaded_hi) = simd.split_u8x64(loaded);
+        let (loaded_1, loaded_2) = simd.split_u8x32(loaded_lo);
+        let (loaded_3, loaded_4) = simd.split_u8x32(loaded_hi);
+
+        let (dest0, rest0) = row0.split_at_mut(Tile::WIDTH as usize * COLOR_COMPONENTS);
+        let (dest1, rest1) = row1.split_at_mut(Tile::WIDTH as usize * COLOR_COMPONENTS);
+        let (dest2, rest2) = row2.split_at_mut(Tile::WIDTH as usize * COLOR_COMPONENTS);
+        let (dest3, rest3) = row3.split_at_mut(Tile::WIDTH as usize * COLOR_COMPONENTS);
```
LaurenzV, line 371: This is slightly changed compared to what we had before, but it was necessary to retain (nearly) the same performance. And it's now not limited to a multiple of 256 for the width, but just 4!

sparse_strips/vello_cpu/src/fine/lowp/mod.rs
```diff
@@ -345,6 +304,116 @@ impl<S: Simd> FineKernel<S> for U8Kernel {
...
+    let mut row3 = &row3[..row_len];
+
+    for col in scratch.as_chunks_mut::<CHUNK_LENGTH>().0.iter_mut() {
+        // Note: We experimented with using u32 vs. f32 for this, but it seems like for some reason
+        // f32 works better on M1, while on M4 they are the same. Probably worth doing more
+        // benchmarks on different systems.
+        let (src0, rest0) = row0.split_at(Tile::WIDTH as usize * COLOR_COMPONENTS);
+        let (src1, rest1) = row1.split_at(Tile::WIDTH as usize * COLOR_COMPONENTS);
+        let (src2, rest2) = row2.split_at(Tile::WIDTH as usize * COLOR_COMPONENTS);
+        let (src3, rest3) = row3.split_at(Tile::WIDTH as usize * COLOR_COMPONENTS);
```
LaurenzV, line 437: By applying the same trick for `unpack`, I was actually able to double performance compared to what we had before! This is now about the same speed as `pack_block`.

LaurenzV:
I think this is ready for review now! 🎉 🎉  Very excited about this.

sparse_strips/vello_cpu/src/record.rs
```diff
@@ -0,0 +1,535 @@
...
+            self.push_filter_layer(props, filter_plan);
+            return;
+        }
+
+        let id = self.push_layer_metadata(RecordedLayer::regular(props));
+        self.push_render_cmd(RecordedCmd::PushLayer { id });
+        self.root_layer_stack.push(PushedRootLayer {
+            id,
+            previous_root_layer: self.active_root_layer,
+        });
```
grebmeg, line 348: For readability, it would be helpful to have a `push_regular_layer` method, similar to `push_filter_layer`.

sparse_strips/vello_sparse_tests/snapshots/filter_extreme_blur_2.png
grebmeg, line 1: I’m seeing a fairly significant difference in the blur results, but I’m not sure what’s causing it. As far as I can tell, the blur logic hasn’t changed, yet the difference is quite noticeable.

Do you have any ideas why this might be happening here and in the other tests as well?

https://github.com/user-attachments/assets/9b905ba8-d64a-443e-a9dc-6c3371cafbdd

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -239,37 +243,55 @@ impl RenderContext {
     }
 
     fn encode_current_paint(&mut self) -> Paint {
+        let transform = self.effective_paint_transform();
```
grebmeg: Since you don’t need this for solid colors, it might be worth leaving it only in the match arms for the other cases?

sparse_strips/vello_cpu/src/record.rs
```diff
@@ -0,0 +1,535 @@
...
+    /// For example, if the filter layer contains a rect spanning (200, 200) to (300, 300)
+    /// with a blur that has a radius exceeding the rectangle 40 pixels on each side, the pixmap
+    /// bbox will be (160, 160) to (340, 340).
+    ///
+    /// See the comments in [`FilterLayerPlacement::new`] for more information.
+    pub(crate) pixmap_bbox: RectU16,
+    /// Rectangle in the parent layer's coordinate space the filtered pixmap is composited into.
+    ///
+    /// See the comments in [`FilterLayerPlacement::new`] for more information.
+    pub(crate) composite_bbox: RectU16,
```
grebmeg: Maybe `dest_bbox` would be a bit more descriptive here?

sparse_strips/vello_cpu/src/record.rs
```diff
@@ -0,0 +1,535 @@
...
+    pub(crate) layers: Vec<RecordedLayer>,
+    /// A pool for reusable `Vec<RecordedCmd>` allocations.
+    cmd_pool: VecPool<RecordedCmd>,
+    /// The layer whose command stream is currently treated as the root.
+    ///
+    /// This is either the real scene root (`None`) or an active filter layer
+    /// (`Some`).
+    active_root_layer: Option<LayerId>,
+    /// Stack of currently pushed root layers.
+    root_layer_stack: Vec<PushedRootLayer>,
```
grebmeg: I think `root` is a bit misleading here. Is it accurate to say that `active_root_layer` can only ever have a filter layer? If so, it might be clearer to rename `active_root_layer` to `active_filter_layer`, and `root_layer_stack` to simply `layer_stack`. Also, replacing `PushedRootLayer` with something like next could make the flow easier to understand.

```rust
#[derive(Debug)]
struct OpenLayer {
    id: LayerId,
    /// The filter layer that was active when this layer was pushed; restored as
    /// the recording target when a filter layer is popped.
    enclosing_filter_layer: Option<LayerId>,
}
```

sparse_strips/vello_cpu/src/record.rs
```diff
@@ -0,0 +1,535 @@
...
+        });
+    }
+
+    pub(crate) fn pop_layer(&mut self) -> PoppedLayer {
+        let root_layer = self.root_layer_stack.pop().unwrap();
+        let id = root_layer.id;
+        let bbox = self.layers[id.get()].bbox;
+        match &mut self.layers[id.get()].kind {
+            RecordedLayerKind::Regular => {
+                self.record_bbox(|| bbox);
```
grebmeg: If this layer is clipped, should we intersect `bbox` with `clip.bbox` here and maybe for filter layer too?

sparse_strips/vello_cpu/src/record.rs
```diff
@@ -0,0 +1,535 @@
...
+                bbox.union(RectU16::new(strip_x1, row_y, fill_x1, row_y1));
+            }
+        }
+    }
+
+    bbox
+}
+
+#[cfg(test)]
+mod tests {
```
grebmeg, line 482: `strip_bbox` is well covered, but `CommandRecorder`'s active-root switching and `FilterLayerPlacement::new` have no direct tests. A few cheap unit tests (nested + sibling filter layers, placement with/without source shift) i think it'd be worth adding here.

sparse_strips/vello_cpu/src/record.rs
```diff
@@ -0,0 +1,535 @@
...
+fn strip_bbox(strips: &[Strip], viewport_width: u16) -> RectU16 {
+    let mut bbox = RectU16::INVERTED;
+
+    // Need at least one strip (and the sentinel one).
+    if strips.len() < 2 {
+        return bbox;
+    }
+
+    // TODO: It _feels_ like this "iterating over strips code" should be possible to
+    // deduplicate with other locations (i.e. the code used to generate fill commands).
```
grebmeg, line 444: Yeah, exactly. I think we should generate fill's `bbox` while the strips are being generated, without requiring another pass — would this be possible?

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -0,0 +1,828 @@
...
+        assert_eq!(
+            self.viewport.x0 % Tile::WIDTH,
+            0,
+            "viewport origin must be tile-width aligned",
+        );
+        assert_eq!(
+            self.viewport.y0 % Tile::HEIGHT,
+            0,
+            "viewport origin must be tile-height aligned",
+        );
```
grebmeg: What about making this `debug_assert`?

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -0,0 +1,828 @@
...
+                bucketer.push_fill(fill, attrs_idx, draw_id);
+            },
+            |bucketer, fill| {
+                let row_idx = fill.row_idx;
+                bucketer.ensure_row_layers(row_idx);
+                bucketer.rows[row_idx].push_cmd(RenderCmd::PaintFill(PaintFill::new(
+                    fill.span,
+                    Some(fill.alpha_idx),
+                    attrs_idx,
+                )));
```
grebmeg, line 600: Maybe, similar to `push_fill`, we could create `push_alpha_fill` method?

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -0,0 +1,828 @@
...
+
+    pub(crate) fn push_layer(&mut self, props: &LayerProps) {
+        let parent_bbox = *self.clip_bboxes.last().unwrap();
+        let bbox = props
+            .clip_path
+            .as_ref()
+            .map(|clip| {
+                let clip_bbox = clip.bbox.relative_to_origin(self.viewport_origin());
+                clip_bbox.intersect(parent_bbox)
+            })
```
grebmeg, line 353: If I understand everything right there's a coordinate-space mismatch here: `clip.bbox` is in scene coordinates, but `clip_bboxes` is viewport-local (starts at (0,0)). This only works when the viewport origin is (0,0) — for filter layers with a nonzero pixmap origin, fills inside the clip get dropped. Repro:

```rust
    #[test]
    fn clips_fills_correctly_inside_nonzero_origin_viewport() {
        // Viewport spans scene (32, 32)..(96, 96): local space is 64x64, origin (32, 32).
        let mut bucketer = CommandBucketer::new(RectU16::new(32, 32, 96, 96));
        // Clip bbox in scene coordinates: (40, 32)..(72, 96) => local (8, 0)..(40, 64).
        bucketer.push_layer(&clipped_layer_props(RectU16::new(40, 32, 72, 96)));

        // A 32px-wide alpha strip at scene (40, 32) => local (8, 0), fully inside the clip.
        let strips = [
            Strip::new(40, 32, 0, false),
            Strip::new(72, 32, 32 * u32::from(Tile::HEIGHT), false),
        ];
        bucketer.generate_fill(&strips, &fill_attrs(Paint::Solid(color(RED))), &[]);

        let row = &bucketer.rows()[0];
        assert_eq!(
            row.render_cmds.len(),
            2,
            "expected PushBuf + PaintFill, got {:?}",
            row.render_cmds
        );
        assert!(matches!(row.render_cmds[0], RenderCmd::PushBuf));
        assert!(matches!(
            row.render_cmds[1],
            RenderCmd::PaintFill(cmd)
                if cmd.span.pixel_x() == 8
                    && cmd.span.pixel_width() == 32
                    && cmd.alpha_idx() == Some(0)
        ));
    }
```

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -0,0 +1,828 @@
...
+
+        let origin = self.viewport_origin();
+        let strip_x = |strip: &Strip| {
+            // TODO: Maybe there is a better way we can handle this?
+            if strip.is_sentinel() {
+                strip.x
+            } else {
+                strip.x.saturating_sub(origin.0)
+            }
+        };
```
grebmeg: It looks like `saturating_sub` could clamp strips that are above the viewport origin to row 0 instead of skipping them? If I read it right, clip paths aren't included in the layer bbox, so a clip larger than the content might have strips above the filter layer's pixmap — and those would land in row 0 as an extra LayerFill, compositing that row twice. I think the next test should repro this issue:

```rust
    #[test]
    fn culls_clip_strips_above_viewport_origin() {
        // Viewport spans scene (0, 32)..(64, 96): origin (0, 32), local space 64x64.
        let mut bucketer = CommandBucketer::new(RectU16::new(0, 32, 64, 96));

        let alpha = u32::from(Tile::HEIGHT);
        let strips = [
            // Content: 16px alpha strip at scene (0, 32) => local row 0.
            Strip::new(0, 32, 0, false),
            Strip::new(16, 32, 16 * alpha, false),
            // Clip strip above the viewport origin (scene y = 0): covers nothing visible.
            Strip::new(0, 0, 16 * alpha, false),
            Strip::new(16, 0, 32 * alpha, false),
            // Clip strip at scene y = 32 => local row 0: the real coverage.
            Strip::new(0, 32, 32 * alpha, false),
            Strip::new(16, 32, 48 * alpha, false),
        ];

        bucketer.push_layer(&clipped_layer_props_with_strips(
            RectU16::new(0, 0, 16, 40),
            2..6,
        ));
        bucketer.generate_fill(&strips[0..2], &fill_attrs(Paint::Solid(color(RED))), &[]);
        bucketer.pop_layer(&strips);

        let row = &bucketer.rows()[0];
        assert_eq!(
            count_layer_fills(&row.render_cmds),
            1,
            "row 0 must be composited exactly once, got {:?}",
            row.render_cmds
        );
    }
```

sparse_strips/vello_cpu/src/coarse/depth.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// aligned to whole depth buckets.
+    Regular(Span),
+    /// An opaque span that is aligned and can therefore be rendered front-to-back with
+    /// depth-buffer write enabled.
+    Opaque(BucketRange),
+}
+
+/// Splits a tile-aligned span into regular edge spans and a depth-trackable
+/// opaque middle span.
+pub(crate) fn split_opaque_span(span: Span, mut segment: impl FnMut(DepthSegment)) {
```
grebmeg, line 80: The tile-alignment requirement is documented but unchecked, and it's enforced only by convention over in bucketer.rs. An unaligned span makes the edge segments expand outward (e.g. `Span::new(2, 130)` produces segments covering [0, 132) — 2 px of overpaint). Currently unreachable, but cheap to guard at the boundary where a violation would be introduced:

```rust
    debug_assert!(
        span.pixel_x().is_multiple_of(Tile::WIDTH)
            && span.pixel_end().is_multiple_of(Tile::WIDTH),
        "`split_opaque_span` requires a tile-aligned span"
    );
```

sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -443,535 +439,650 @@ pub trait FineKernel<S: Simd>: Send + Sync + 'static {
...
-            Cmd::PopBuf => {
-                self.blend_buf.pop();
+            RenderCmd::PushBuf => {
+                // TODO: Instead of always resizing to the full viewport width, very strongly
+                // consider only resizing to the horizontal bounding box of the layer. This should
+                // save quite a bit of time if we have many layers with small contents inside.
+                let mut buf = self.buffer_pool.take();
+                buf.resize(self.blend_buffers[0].len(), T::Numeric::ZERO);
+                self.blend_buffers.push(buf);
             }
```
grebmeg: Since the pool clears returned buffers, this resize re-zeroes the full viewport width on every `PushBuf`, even when the layer only covers a small span of the row. For wide canvases with many small layers that's a lot of memset bandwidth (64 KB/row/layer at 4096 px, u8). `PushBuf` could carry the layer span (it fits in the 16-byte `RenderCmd`) and we'd only need to zero `scratch_range(span)` — everything outside it is never read by LayerFill, would that work?

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -142,342 +114,200 @@ impl SingleThreadedDispatcher {
...
+            if pixmap_bbox.is_empty() {
+                continue;
+            }
 
-        let mut cmd_idx = ranges.render_range.start;
-        while cmd_idx < ranges.render_range.end {
-            let cmd: &Cmd = &wtile.cmds[cmd_idx];
+            let width = pixmap_bbox.width();
+            let height = pixmap_bbox.height();
+            let mut pixmap = Pixmap::new(width, height);
```
grebmeg, line 248: The next idea is not for this PR, but I think about having a pool of N pixmaps and also about reusing fine, depth buffer, and regions somehow. I think it’s feasible, we’d just need to figure out how to pass all of that around and where it should be initialized. I believe you already have a TODO for this somewhere.

Leaving this comment because I think it’s an important optimization opportunity that we can explore further.

sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -443,535 +439,650 @@ pub trait FineKernel<S: Simd>: Send + Sync + 'static {
...
+                        + usize::from(x - cmd.span.pixel_x()) * Tile::HEIGHT as usize;
+                    &alphas[alpha_offset..]
+                });
+
+                // Avoid using depth buffer if it's trivially skippable,
+                // since it's generally cheaper to not use it at all than consult it just to be
+                // returned the same span.
+                if !row.can_skip_depth(span, attrs.draw_id) {
+                    depth.for_each_visible_run(span, attrs.draw_id, |span| {
+                        self.paint_fill(span, attrs, resources, alphas);
```
grebmeg: I think I found a rendering issue here, and it might be caused by this line.

When an AA strip wider than one depth bucket is partially occluded by a later opaque fill, the visible part right of the occluder renders with what looks like coverage values from the left end of the strip. My guess: the alphas slice is offset once for the full command span, but `for_each_visible_run` can split the span around occluded buckets — so sub-spans after the first read coverage from the wrong position. `layer_fill` seems to handle this by re-offsetting per sub-span, so maybe the same is needed here?

Repro that shows the difference (all `vello_cpu` variants show the artifact; `vello_hybrid`, which has no depth-culling path, renders it as expected):

```rust
#[vello_test(width = 512, height = 16)]
fn opaque_rect_partially_occluding_aa_edge(ctx: &mut impl Renderer) {
    // Hypotenuse crosses strip row y in 8..12 over the full width, producing one
    // long AA strip. The rect's interior covers depth buckets [128, 384), splitting
    // the strip into visible runs [0, 128) and [384, 512).
    let mut triangle = BezPath::new();
    triangle.move_to((0.0, 8.0));
    triangle.line_to((512.0, 12.0));
    triangle.line_to((0.0, 12.0));
    triangle.close_path();
    ctx.set_paint(DARK_BLUE);
    ctx.fill_path(&triangle);
    ctx.set_paint(RED);
    ctx.fill_rect(&Rect::new(96.0, 8.0, 416.0, 12.0));
}
```

Image

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -554,98 +366,85 @@ impl Dispatcher for SingleThreadedDispatcher {
...
         aliasing_threshold: Option<u8>,
         mask: Option<Mask>,
-        filter: Option<Filter>,
+        filter_data: Option<FilterData>,
     ) {
-        // Allocate a new unique layer ID.
-        self.layer_id_next += 1;
+        if let Some(filter_data) = &filter_data {
+            self.push_filter_viewport(filter_data.source_padding);
+        }
```
grebmeg, line 398: I might be wrong here, but it looks like when we swap the strip generator to the shifted/expanded filter viewport, `self.clip_context` keeps the clip strips it already generated in the unshifted parent space. If a non-isolated clip is still active when a filter layer starts, the fill/stroke methods below would then intersect shifted content with unshifted clip strips — which I think would move the visible clip boundary by `source_shift`. I haven't fully traced it, but a quick experiment seems to confirm the issue (the image below current vello_cpu vs vello_hybrid diff:

```rust
#[vello_test(skip_multithreaded, width = 100, height = 100)]
fn filter_outer_clip_path_with_blur(ctx: &mut impl Renderer) {
    ctx.push_clip_path(&Rect::new(0.0, 0.0, 50.0, 100.0).to_path(0.1));
    ctx.push_filter_layer(Filter::from_primitive(FilterPrimitive::GaussianBlur {
        std_deviation: 4.0,
        edge_mode: EdgeMode::None,
    }));
    ctx.set_paint(RED);
    ctx.fill_rect(&Rect::new(10.0, 10.0, 90.0, 40.0));
    ctx.pop_layer();
    ctx.pop_clip_path();
}
```

Image

grebmeg (commented):
Really great to see how easier it is now to understand the code, even with new concepts like the depth buffer and its own bucketing! The general approach sounds promising and enables many simplifications and optimizations: filter layers are no longer inlined into the parent stream, there is no more wide-tile grid, filter layers are properly isolated and can be composited correctly into the parent layer, and there are also improvements such as the coarse depth buffer and allocation pooling, and etc. Really great work!

This PR is not simple to review, but I tried my best to go through it thoroughly and found a couple of issues that I think are concerning.

sparse_strips/vello_cpu/src/record.rs
LaurenzV: Yes that's right. Makes sense, will rename this.

sparse_strips/vello_cpu/src/record.rs
LaurenzV: I think so, but the problem is I want to avoid doing this in `render_strips` for now, because this way we have to pay the cost of computing the bbox for all paths. The way it is currently, it only happens for paths that are in an active layer. I think I would prefer keeping it this way for now and exploring it in a follow-up so I can properly evaluate the performance impact.

sparse_strips/vello_cpu/src/coarse/bucketer.rs
LaurenzV: Will do, though since `bucket_commands` is only called once per root/filter layer I think the performance impact is negligible.

sparse_strips/vello_cpu/src/record.rs
LaurenzV: I don't think this is valid to do for filter layers (and I'm also getting a test failure when trying this, which is similar to what we had before):

image

The problem is that the bbox is used to determine the size of the pixmap itself. However, if we have a layer with clip + filter, the clip should not be affecting the application region of the filter.

It might be possible for normal layer, however, I think it makes more sense to really treat this bounding box as a source of truth for the bounding box of all strips that are present in the recording of the layer. During bucketing, we do [make sure to clip the commands against the clip bbox](https://github.com/linebender/vello/blob/1e7aef2c296c93d63ead2802db48be439dcf6ba8/sparse_strips/vello_cpu/src/coarse/bucketer.rs#L606), so there also isn't much of a benefit of doing it at this stage.

sparse_strips/vello_sparse_tests/snapshots/filter_extreme_blur_2.png
LaurenzV: So, it turns out this is because we are sizing pixmaps differently than before. Previously, filter pixmaps always were snapped to wide tile coordinates, so the width had to be a multiple of 256, and the height of 4. Now, we are only snapping to tile coordinates, so both have to be a multiple of 4. Because of this, depending on what the size of the intermediate pixmap is, you end up with a different sequence of pixmap decimations during the gaussian blur. If I change this branch to instead snap to wide tile coordinates again, I'm getting a pixel-perfect match again to the snapshot from main, so pretty sure that this is the issue. A bit unfortunate that the gaussian blur is so sensitive to this, but not much we can do I think. 😄

With the following diff, I'm getting a perfect pixel match compared to main:
```diff
diff --git a/sparse_strips/vello_cpu/src/record.rs b/sparse_strips/vello_cpu/src/record.rs
index d18856eb..ea3abbf3 100644
--- a/sparse_strips/vello_cpu/src/record.rs
+++ b/sparse_strips/vello_cpu/src/record.rs
@@ -41,6 +41,7 @@ use crate::peniko::BlendMode;
 use crate::util::VecPool;
 use alloc::vec::Vec;
 use core::ops::Range;
+use vello_common::coarse::WideTile;
 use vello_common::filter_effects::Filter;
 use vello_common::geometry::RectU16;
 use vello_common::mask::Mask;
@@ -181,15 +182,17 @@ impl FilterLayerPlacement {
         // large of a pixmap we actually need to allocate. Also, as mentioned
         // in [`FilterLayerPlan::new`], we need to ensure the pixmap itself is
         // also a multiple of the tile width / tile height.
-        let pixmap_bbox = bbox
-            .expand(filter_plan.filter_padding)
-            .snap_to_tile_coordinates();
+        let (shift_x, shift_y) = filter_plan.source_shift();
+        let unshifted_bbox = bbox.relative_to_origin((shift_x, shift_y));
+        let pixmap_bbox = translate_bbox(
+            old_style_filter_bbox(unshifted_bbox, filter_plan),
+            (shift_x, shift_y),
+        );
 
         // Remember that in `RenderContext`, we eagerly shift everything drawn by `source_shift`
         // to conservatively ensure that everything that might be needed for the filter is in the
         // viewport area. Therefore, when compositing the filter layer back, we need to undo that
         // shift.
-        let (shift_x, shift_y) = filter_plan.source_shift();
         // For example, if `shift_x` is 20 and `pixmap_bbox.x0` is 4,
         // shifting the pixmap back would place its left edge at -16. Since we
         // start compositing at x=0, we need to skip the first 16 pixels
@@ -264,6 +267,38 @@ impl FilterData {
     }
 }
 
+fn old_style_filter_bbox(bbox: RectU16, filter_plan: &FilterData) -> RectU16 {
+    fn snap_down(value: u16, step: u16) -> u16 {
+        (value / step) * step
+    }
+
+    fn snap_up(value: u16, step: u16) -> u16 {
+        value.checked_next_multiple_of(step).unwrap_or(u16::MAX)
+    }
+
+    let padding = filter_plan.filter_padding;
+    let left = snap_up(padding.x0, WideTile::WIDTH);
+    let right = snap_up(padding.x1, WideTile::WIDTH);
+    let top = snap_up(padding.y0, Tile::HEIGHT);
+    let bottom = snap_up(padding.y1, Tile::HEIGHT);
+
+    RectU16::new(
+        snap_down(bbox.x0, WideTile::WIDTH).saturating_sub(left),
+        snap_down(bbox.y0, Tile::HEIGHT).saturating_sub(top),
+        snap_up(bbox.x1, WideTile::WIDTH).saturating_add(right),
+        snap_up(bbox.y1, Tile::HEIGHT).saturating_add(bottom),
+    )
+}
+
+fn translate_bbox(bbox: RectU16, offset: (u16, u16)) -> RectU16 {
+    RectU16::new(
+        bbox.x0.saturating_add(offset.0),
+        bbox.y0.saturating_add(offset.1),
+        bbox.x1.saturating_add(offset.0),
+        bbox.y1.saturating_add(offset.1),
+    )
+}
+
 #[derive(Debug, Default)]
 pub(crate) struct CommandRecorder {
     /// The commands of the root layer.
```

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Oof, good catch. Let me think about how to solve this best.

sparse_strips/vello_cpu/src/record.rs
LaurenzV: Will add some tests.

sparse_strips/vello_cpu/src/coarse/bucketer.rs
LaurenzV: Were you looking at an old version perhaps? This should already have been fixed in https://github.com/linebender/vello/pull/1701/changes/ed50a518a4158b73f25e2a1ade1230fba03b4577, your test also seems to pass on the newest version. Regression test was already added in that commit.

LaurenzV: I guess it doesn't hurt to add your test as well!

sparse_strips/vello_cpu/src/fine/mod.rs
LaurenzV: Sounds doable, lemme look into this!

sparse_strips/vello_cpu/src/coarse/depth.rs
LaurenzV: Will add this.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Not sure if we have a TODO for this exactly, will add one just to be sure.

sparse_strips/vello_cpu/src/fine/mod.rs
LaurenzV: Thanks, good catch. Will fix this and make the behavior consistent with how layer fills are handled.

sparse_strips/vello_sparse_tests/snapshots/filter_extreme_blur_2.png
grebmeg: So, are you saying this happens because the pixmap width and height we previously received are now different, causing a different number of decimation passes to be applied, which then produces a different blur output since the blur is being performed on a differently sized pixmap?

sparse_strips/vello_cpu/src/record.rs
grebmeg: I agree with keeping the simple option here, but I think it would be good to address this later. Maybe when a filter layer is detected, we could run another `render_strips_with_bbox` instead of `render_strips`.

sparse_strips/vello_sparse_tests/snapshots/filter_extreme_blur_2.png
LaurenzV: Yes

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Fixed now by implementing the clip path replay mechanism I mentioned to you.

sparse_strips/vello_cpu/src/render.rs
```diff
@@ -242,34 +246,59 @@ impl RenderContext {
...
+        self.state.transform
+    }
+
+    fn effective_paint_transform(&self) -> Affine {
+        self.effective_path_transform() * self.state.paint_transform
+    }
+
+    pub(crate) fn push_root_transform(&mut self, transform: Affine) {
+        self.root_transforms.push(transform * self.root_transform());
+    }
```
grebmeg: nit: `push_root_transform`'s param is named `transform`, but it's composed with the current root (`transform * self.root_transform()`) rather than set as the root. Suggest renaming the param to `relative_transform` (or the method to `prepend_root_transform`) to make the composition clear.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -142,342 +114,207 @@ impl SingleThreadedDispatcher {
...
+            // #[cfg(all(debug_assertions, feature = "std", feature = "png"))]
+            // save_filtered_layer_debug(&pixmap, id);
+
+            filter_ctx.set_layer(id, pixmap);
+        }
+
+        filter_ctx
+    }
+
+    fn push_filter_viewport(&mut self, filter_data: &FilterData) {
```
grebmeg: nit: `push_filter_viewport` is a bit vague — it actually swaps in a larger, padded `StripGenerator` so filter effects (blur/shadow) can render past the viewport. Could we rename to `push_filter_surface`?

sparse_strips/vello_cpu/src/clip.rs
```diff
@@ -0,0 +1,180 @@
...
+        for raw_clip in &self.raw_clips {
+            self.context.push_clip(
+                self.path_elements[raw_clip.path.clone()].iter().copied(),
+                strip_generator,
+                raw_clip.fill_rule,
+                active_shift * raw_clip.transform,
+                raw_clip.aliasing_threshold,
+            );
+        }
+    }
```
grebmeg, line 179: Because the shift is always an integer, I think it’s worth exploring whether we can simply apply an integer translation to the already-clipped context?

Another idea is to avoid translating the content into the surface altogether, and instead make the surface aware that it has a non-zero origin. The StripGenerator/surface could carry an origin: `(i32, i32)` (or a `dx/dy` offset), so nothing would actually need to be shifted. However, the compositing side, strip addressing, and related logic would need to take that origin into account. While this would likely require more changes, it could potentially eliminate the need to rebuild clipped contexts?

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -554,98 +373,85 @@ impl Dispatcher for SingleThreadedDispatcher {
...
     fn reset(&mut self) {
-        // Clear all rendering state to prepare for a new scene.
-        self.wide.reset();
-        self.clip_context.reset();
+        // Bucketer will be reset on demand, so no need to reset it here.
+        self.clip_state.reset();
+        self.recorder.reset();
+        self.strip_generator_stack.clear();
+        self.strip_generator.reset();
         self.strip_generator.reset();
```
grebmeg: `strip_generator.reset()` is called twice here — looks like a copy-paste slip.

sparse_strips/vello_cpu/src/render.rs
LaurenzV: Will rename it to `relative_transform`.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Will rename it in both, `ClipState` and `SingleThreadedDispatcher`.

sparse_strips/vello_cpu/src/clip.rs
LaurenzV: > Because the shift is always an integer, I think it’s worth exploring whether we can simply apply an integer translation to the already-clipped context?

As mentioned, the problem here is that the clip path has already been clipped to the original viewport. So if you shift the clip path to the bottom-right, any information from the top-left of the original clip path has already been lost.

So yes, we can definitely explore the negative coordinates thing, but I would suggest we leave this for the future. 🙂

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
LaurenzV: Thanks!

grebmeg (approved):
Thanks for your work! It’s really impressive how well everything fits together. Rewrites like this are rarely easy, especially when so many moving parts are involved, so it’s quite an achievement to get it all working while keeping the entire test suite passing!

LGTM — I think this is ready to land. Could you also add a few lines to the CHANGELOG describing the rewrite?

PR #1702, opened 2026-06-08
[cpu] Ignore filters rather than panicking when mulithreading is enabled

nicoburns:
This PR disables panicking only for multithreaded Vello CPU (where filters aren't implemented at all)


LaurenzV (approved):
I guess let's still wait for @taj-p, but from my side this is fine. Hopefully this will only be a short-term workaround and we can actually add support for multi-threaded filters soon.

taj-p (approved):
I think if we have to choose between being strict with our contracts - that is, panicking when a consumer hasn't read the documentation (forcing them to adhere to calling contracts) - or silent - that is, doing nothing and relying on consumers reading documentation - I think, philosophically, it's much better to be strict. 

I think renderers should strive for correctness first. It's surprising that valid API calls result in essential noops for outputs.

---

The alternative pragmatic option to this solution is expanding the API surface of RenderContext to expose the required state (as @nicoburns and I previously suggested). Since, we expect this to be relatively short term, it's probably preferable to keep the API surface small and suffer the debt since @LaurenzV is happy to pay it.

LaurenzV:
Why is the probe reference changing here? That doesn’t make a lot of sense to me

nicoburns:
I'm really not sure. But it was failing. Agree that it doesn't make much sense.

LaurenzV:
Oh right, because just checking if the multi threaded feature is enabled is not enough. Because if the feature is enabled(ehich it is for the tests) but you only choose a single thread at runtime, it will still choose the single-threaded dispatcher.

PR #1703, opened 2026-06-08
vello_cpu: Add a `is_multi_threaded` method

LaurenzV:
As an alternative to #1702.


nicoburns (approved):
It could probably be public rather than `#[doc(hidden)]`? But this looks like it would solve my issue.

LaurenzV:
I thought it makes sense since it's only supposed to be temporary, but yeah I guess I can just make it public.

PR #1708, opened 2026-06-13
vello_hybrid: Remove outdated debug assertion

### LINKED ISSUE #1707: vello_hybrid: panics on fully transparent colors ("Color fields with 0 alpha are reserved for clipping")
`vello_hybrid` panics on the first frame containing a fully transparent paint:

```
panicked at vello_hybrid-0.0.9/src/schedule.rs:1823:
Color fields with 0 alpha are reserved for clipping
```

Fully transparent colors are legal and routine in CSS content — Tailwind apps emit them constantly (`background-color: rgba(0,0,0,0)` as hover-transition base states, `transparent` gradient stops). Hit via blitz/Dioxus Native, where vello_hybrid is the default renderer: a real-world app aborts on its very first frame.

Could the scheduler skip zero-alpha draws (they're no-ops visually) instead of asserting?

LaurenzV:
Fixes #1707. I'm still not sure how they are able to trigger this assertion, but since it's anyway outdated we should still remove it.

This assertion was introduced in https://github.com/linebender/vello/pull/957. Back then, in the shader the information whether to use a slot or not was stored in the same field that stores the RGBA color for solid fills:

image

However, this is not the case anymore. Now we have two separate flags, `payload` which stores the RGBA values and `paint_and_rect_flag` which stores what kind of paint is actually used (including whether slots should be used). 

https://github.com/linebender/vello/blob/da699dec733ee5aa54b6f5f0e351fa6b894bdf56/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl#L217-L220

Therefore, this debug assertion is not relevant anymore.

### DIFF
--- a/sparse_strips/vello_hybrid/src/schedule.rs
+++ b/sparse_strips/vello_hybrid/src/schedule.rs
@@ -1820,10 +1820,6 @@ impl Scheduler {
         match paint {
             Paint::Solid(color) => {
                 let rgba = color.as_premul_rgba8().to_u32();
-                debug_assert!(
-                    has_non_zero_alpha(rgba),
-                    "Color fields with 0 alpha are reserved for clipping"
-                );
                 let paint_packed = (COLOR_SOURCE_PAYLOAD << 30) | (PAINT_TYPE_SOLID << 27);
                 ProcessedPaint {
                     payload: rgba,
--- a/sparse_strips/vello_sparse_tests/snapshots/issue_1707_transparent_solid_fill.png
+++ b/sparse_strips/vello_sparse_tests/snapshots/issue_1707_transparent_solid_fill.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:56f57423e0ff37d0284ae02f3dfa6ceffad893302f86b5aa4d17dabb5ba1b25d
+size 82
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
good to have the history on this one. We talked about this briefly in the past, I think, so glad to see it go away.

waywardmonkeys:
Since I hit this in the past, that means that there's some code here that isn't currently covered by tests in `vello_hybrid`. Now that this is going away, it might be good to add a test that covers this path to help ensure that it remains working.

LaurenzV:
Tbh I'm not quite sure where it's coming from because we, at least from the brief look I took, we are bailing out before painting with an invisible paint. But I'll try to take another look.

waywardmonkeys:
Is it related to the paint not actually being invisible due to the blend mode? There was a bug in one of them renderers before about that too, but I don't know if that was fixed. :)

LaurenzV:
It seems like you can trigger it with something like `Color::from_rgb8(0, 0, 0).with_alpha(0.001)`, so a non-zero alpha value that gets rounded to 0 when converting to u8.

taj-p (approved):

UMCEKO:
> Since I hit this in the past, that means that there's some code here that isn't currently covered by tests in `vello_hybrid`. Now that this is going away, it might be good to add a test that covers this path to help ensure that it remains working.

I can write the tests and if they still fail fix the relevant part after finals.

LaurenzV:
No worries, I’ve already added one, but thanks!

PR #1714, opened 2026-06-23
sparse_strips: Distinguish between sentinel strips and zero-width delimiter strips

LaurenzV:
This is based on top of #1705.

# Context

In the very early days, the so-called "sentinel strip" was simply used as to indicate "the final strip in a sequence of strips". It provided a way for us to avoid special casing handling the last real strip. In order to determine the width of a strip, we need to look at the `alpha_idx` of the next strip. So having this sentinel strip allows us to treat the last real strip like any other strip.

Starting with commit https://github.com/linebender/vello/commit/4a75eced and then also https://github.com/linebender/vello/commit/1b24e77b, we started to essentially have two versions of sentinel strips: One that, as was the original version, is used to terminate a sequence of strips and make it possible to _not_ have to special case the last strip. And another kind of sentinel strip which simply indicates "fill this area horizontally until the end of the viewport".

# Problem
For fine rasterization, it would be easier if we can assume that a strip never exceeds the viewport. This is generally the case, except in the case where such row-level sentinel strips are emitted, because their x position is set to `u16::MAX`. Therefore, during fine rasterization, we have to manually clamp strips to the viewport width.

# Solution
This PR proposes to solve this problem by
1) Adding a constructor for "sentinel strips" which is really just used terminate a sequence of strips.
2) Instead of also using sentinel strips as a marker of "fill this row across the whole viewport width", instead explicitly allow using zero-width strips as a terminator in the row, and make sure that this terminating strip itself is placed exactly at the viewport boundary.

# TODO
I still need to run this through my PDF test suite, but overall I think this is ready for review.

### DIFF
--- a/sparse_strips/vello_common/src/clip.rs
+++ b/sparse_strips/vello_common/src/clip.rs
@@ -324,13 +324,11 @@ fn intersect_impl<S: Simd>(
         cur_y += 1;
     }
 
-    // Push the sentinel strip, if one wasn't already pushed.
-    if !target.strips.last().is_some_and(Strip::is_sentinel) {
-        target.strips.push(Strip::new(
-            u16::MAX,
+    // Push the sentinel strip if the intersection is not empty.
+    if !target.strips.is_empty() {
+        target.strips.push(Strip::sentinel(
             end_y * Tile::HEIGHT,
             target.alphas.len() as u32,
-            false,
         ));
     }
 }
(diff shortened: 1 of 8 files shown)


b0nes164:
Taking a look now; in the problem section, you note that we want the behavior where "strips never exceed the viewport." I'm assuming that, exceeding the viewport by rounding up to the tile size is fine?

Edit: After reviewing I can see why this easier and it should be fine.

b0nes164 (approved):
LGTM

sparse_strips/vello_common/src/strip.rs
```diff
@@ -123,10 +129,10 @@ impl Strip {
             if should_fill(windings.coarse[row] as i32) {
                 let y_pos = row as u16 * Tile::HEIGHT;
                 strips.push(Self::new(0, y_pos, alphas.len() as u32, false));
+                // TODO: Would be nice to get rid of this, but the current clipping code only
+                // allows zero-width strips as a row terminator, not in-between.
                 alphas.extend([255_u8; Tile::HEIGHT as usize * Tile::WIDTH as usize]);
-                // TODO: Clamp to the scene width instead of u16::MAX; in the future there might
-                // not be clamping on the x.
-                strips.push(Self::new(u16::MAX, y_pos, alphas.len() as u32, true));
+                strips.push(Self::new(viewport_width, y_pos, alphas.len() as u32, true));
```
grebmeg, line 135: Since row-end delimiters now carry `viewport_width` and the only `u16::MAX` strip is the sentinel (`fill_gap == false`), I think [this branch](https://github.com/linebender/vello/blob/ab64bb3dfb020e4f833acfe9030a9a71eaf0c9b1/sparse_strips/vello_common/src/coarse.rs#L1279-L1284) is unreachable — the enclosing `if next_strip.fill_gap()` can never see `x2 == u16::MAX`. If it's so, maybe consider dropping it?

sparse_strips/vello_common/src/clip.rs
```diff
@@ -512,42 +510,53 @@ impl<'a> Iterator for RowIterator<'a> {
...
-            width,
-            alphas,
-        }))
+            // Zero-width strips only act as markers for cheaply delimiting the width
+            // of filled regions, but are not actually relevant for clipping. This is assuming that
+            // zero-width strips can only appear at the end of a row, see the comment in
+            // `Strip::emit_culled_background`.
+            if width == 0 {
+                continue;
+            }
```
grebmeg, line 555: What about hardening it with a debug assertion, just in case a future producer ever emits a zero-width strip mid-row?

```suggestion
            if width == 0 {
              debug_assert!(
                self.next_strip().is_sentinel()
                  || self.next_strip().strip_y() != self.strip_y,
                  "zero-width strips must only appear at the end of a row"
              );
              continue;
            }
```

grebmeg (approved):

sparse_strips/vello_common/src/strip.rs
laurenz-canva: Yeah I think you are right!

sparse_strips/vello_common/src/clip.rs
laurenz-canva: Good call. 👍

PR #1715, opened 2026-06-24
sparse_strips: Implement inverse blurred rounded rectangles

### LINKED ISSUE #1374: Consider support for "inverse" blurred rounded rectangles
It can be desired to support blurred rounded rectangles where the inside of the rectangle is transparent, with the outside being opaque. Imagine for example that this is a blurred "cutout" into an infinite plane. This is not supported by our current blurred rounded rectangle code, as the "infinite outside extent" colour is always transparent. This would be used in "inset" box shadows.
I believe that this would be implemented as a paint type equivalent to `1-brr`, where `brr` is the current blurred rounded rectangle paint type.

Before implementing this, we will need to confirm that this is actually the correct semantics for an inset box shadow, because otherwise this paint type would likely not have any practical use.

[#vello > Inset box shadows](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Inset.20box.20shadows/with/569523012)

See also #1373 for implementing blurred rounded rectangles in Vello Hybrid

nicoburns:
Partially fixes #1374 (fixes it for Vello Hybrid and Vello CPU)

Adds an `inverse: bool` parameter to the `fill_blurred_rounded_rect` method on `vello_hybrid::Scene` and `vello_cpu::RenderContext`. When the `inverse` flag is set, the region that gets painted with by the blurred rounded rect fill is inverted. This can be used to implement inset box shadows.

Generated with Opus 4.8 High. Some manual review, but I don't claim to understand in particular the GPU code. The tests have been reviewed and look good to me.

Feedback on whether PRs like this are helpful is welcome.

### DIFF
--- a/sparse_strips/vello_common/src/blurred_rounded_rect.rs
+++ b/sparse_strips/vello_common/src/blurred_rounded_rect.rs
@@ -20,5 +20,5 @@ pub struct BlurredRoundedRectangle {
     ///
     /// When `true`, the paint is fully opaque outside the blurred rectangle and fades to
     /// transparent inside it. This is useful for implementing inset box shadows.
-    pub inverse: bool,
+    pub invert: bool,
 }
--- a/sparse_strips/vello_common/src/blurred_rounded_rect.rs
+++ b/sparse_strips/vello_common/src/blurred_rounded_rect.rs
@@ -16,4 +16,9 @@ pub struct BlurredRoundedRectangle {
     pub radius: f32,
     /// The standard deviation of the blur effect.
     pub std_dev: f32,
+    /// Whether to paint the inverse (`1 - alpha`) of the blur coverage.
+    ///
+    /// When `true`, the paint is fully opaque outside the blurred rectangle and fades to
+    /// transparent inside it. This is useful for implementing inset box shadows.
+    pub inverse: bool,
 }
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -883,7 +883,7 @@ pub struct EncodedBlurredRoundedRectangle {
     ///
     /// When `true`, the paint is fully opaque outside the blurred rectangle and fades to
     /// transparent inside it. This is useful for implementing inset box shadows.
-    pub inverse: bool,
+    pub invert: bool,
     /// The base color for the blurred rectangle.
     pub color: PremulColor,
     /// A transform that needs to be applied to the position of the first processed pixel.
@@ -958,7 +958,7 @@ impl EncodeExt for BlurredRoundedRectangle {
             r1,
             std_dev_inv,
             min_edge,
-            inverse: self.inverse,
+            invert: self.invert,
             color: PremulColor::from_alpha_color(self.color),
             w,
             h,
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -879,6 +879,11 @@ pub struct EncodedBlurredRoundedRectangle {
     pub height: f32,
     /// An component for computing the blur effect.
     pub r1: f32,
+    /// Whether to paint the inverse (`1 - alpha`) of the blur coverage.
+    ///
+    /// When `true`, the paint is fully opaque outside the blurred rectangle and fades to
+    /// transparent inside it. This is useful for implementing inset box shadows.
+    pub inverse: bool,
     /// The base color for the blurred rectangle.
     pub color: PremulColor,
     /// A transform that needs to be applied to the position of the first processed pixel.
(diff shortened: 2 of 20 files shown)


nicoburns:
@DJMcNab @xStrom This looks like a CI failure unrelated to this PR to me (`error: failed to transform intralinks: rust toolchain not installed: nightly`). Any ideas?

sparse_strips/vello_common/src/blurred_rounded_rect.rs
```diff
@@ -16,4 +16,9 @@ pub struct BlurredRoundedRectangle {
     pub radius: f32,
     /// The standard deviation of the blur effect.
     pub std_dev: f32,
+    /// Whether to paint the inverse (`1 - alpha`) of the blur coverage.
+    ///
+    /// When `true`, the paint is fully opaque outside the blurred rectangle and fades to
+    /// transparent inside it. This is useful for implementing inset box shadows.
+    pub inverse: bool,
```
LaurenzV: I'm wondering whether using `invert`, the verb, would be better here? Up to you.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -879,6 +879,11 @@ pub struct EncodedBlurredRoundedRectangle {
     pub height: f32,
     /// An component for computing the blur effect.
     pub r1: f32,
+    /// Whether to paint the inverse (`1 - alpha`) of the blur coverage.
+    ///
+    /// When `true`, the paint is fully opaque outside the blurred rectangle and fades to
+    /// transparent inside it. This is useful for implementing inset box shadows.
+    pub inverse: bool,
```
LaurenzV: Same here.

sparse_strips/vello_cpu/src/fine/common/rounded_blurred_rect.rs
```diff
@@ -64,7 +66,10 @@ impl<S: Simd> Iterator for BlurredRoundedRectFiller<S> {
 
     #[inline(always)]
     fn next(&mut self) -> Option<Self::Item> {
-        let next = self.alpha_calculator.next().unwrap();
+        let mut next = self.alpha_calculator.next().unwrap();
+        if self.inverse {
+            next = f32x8::splat(next.simd, 1.0) - next;
+        }
```
LaurenzV: I'm wondering whether using a mask would be better here... But I doubt it will have a measurable impact, so let's leave it this way.

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -235,8 +235,8 @@ pub(crate) struct GpuBlurredRoundedRect {
     pub transform: [f32; 6],
     /// Premultiplied color packed as RGBA8 unorm (`pack4x8unorm` layout).
     pub color: u32,
-    /// Padding for 16-byte alignment.
-    pub _padding0: u32,
+    /// Whether to paint the inverse (`1 - alpha`) of the blur coverage (`0` = normal, `1` = inverse).
```
LaurenzV:
```suggestion
    /// Whether to paint the inverse (`1 - alpha`) of the blur coverage.
```

I think the last part is a bit superfluous.

sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl
```diff
@@ -1335,6 +1335,9 @@ fn get_blurred_rounded_rect_translate(texel1: vec4<u32>) -> vec2<f32> {
 /// Premultiplied rectangle color.
 fn get_blurred_rounded_rect_color(texel1: vec4<u32>) -> vec4<f32> { return unpack4x8unorm(texel1.z); }
 
+/// Whether to paint the inverse (`1 - alpha`) of the blur coverage (`0` = normal, `1` = inverse).
```
LaurenzV:
```suggestion
/// Whether to paint the inverse (`1 - alpha`) of the blur coverage.
```

LaurenzV (approved):
I don't really know much about how inset box shadows work, but seems like a reasonable thing to do and if it gives the visual result you expect, LGTM!

waywardmonkeys:
> @DJMcNab @xStrom This looks like a CI failure unrelated to this PR to me (`error: failed to transform intralinks: rust toolchain not installed: nightly`). Any ideas?

https://xi.zulipchat.com/#narrow/channel/147932-chatter/topic/cargo.20rdme.20update.20has.20CI.20impacts/with/606454912

sparse_strips/vello_cpu/src/fine/common/rounded_blurred_rect.rs
nicoburns: Yeah, I did wonder whether a conditional in a SIMD pipeline was a bad idea, but my SIMD isn't good enough to know what the alternative is.

PR #1717, opened 2026-06-26
Update selected dependencies

oscargus:
First commit: update jni to remove lots of transitive (and old) dependencies. Plus a few others.
Second commit: a few more, new duplicate versions, but also got rid of some and fewer (lines) in total. Separate since there may be more doubts here, but feel free to squash.

(Motivation: be able to update some of these in egui.)

The rewrite of with_winit to work with jni 0.22 was done in collaboration with AI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -447,9 +447,9 @@ checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "chrono"
-version = "0.4.44"
+version = "0.4.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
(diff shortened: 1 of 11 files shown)


oscargus:
I now realize that parley 0.11 is released although it is not yet on crates.io. I can update this once it is generally available since it will enable newer harfrust etc. (Or I am not really getting the dependency trail...)

nicoburns (changes requested):
Parley 0.11.0 has now been released. This should also bump the version of Skrifa to match the Fontations versions used in Parley.

LaurenzV (commented):
Could you please undo the fearless_simd change? I want to rerun the benchmarks before doing this one because of the performance implications, as there were major changes in that release.

oscargus:
Sure! Now parley has also been updated to 0.11.

nicoburns:
@oscargus Have you tested this on Android?

oscargus:
No, not as in running the affected example. I can see if I can setup an environment to run it.

LaurenzV:
See https://github.com/linebender/vello/pull/1733#issuecomment-4957702635.

PR #1720, opened 2026-07-01
fix: incorrect blur in Gaussian blur decimation

grebmeg:
### Problem
When zooming a Gaussian-blurred element, the blur strength does not change smoothly. It stays smooth across a range of zoom, then abruptly jumps to a different "level" of blur, and this repeats at each zoom threshold. 

### Root Cause
The multi-scale blur decimates the image with N power-of-two downsample passes, applies a small Gaussian kernel at the coarsest level, then upsamples back to full resolution. The planner (`plan_decimated_blur`) budgets variance per level so the total matches the target `σ`. Each level actually blurs twice — once on the `[1,3,3,1]/8` downscale (variance `0.75`) and again on the bilinear upscale, whose reconstruction is equivalent to the same `[1,3,3,1]/8` filter (another `0.75`). The budget only accounted for the downscale, and did so with an incorrect grid conversion (`0.75 × 4 = 3.0`). The net effect is that it over-subtracted, leaving the kernel under-powered so the pipeline under-blurred by `0.5·(4ⁿ − 1)`. 

Because that deficit grows with the decimation count N, the produced blur falls further behind the target at every level:

levels N | deficit 0.5·(4ⁿ − 1) (device px²)
-- | --
1 | 1.5
2 | 7.5
3 | 31.5
6 | 2 047.5
7 | 8 191.5

Crossing an N boundary while zooming therefore snaps the effective blur. For example, at the N=6→7 boundary (target σ ≈ 143):

at target σ ≈ 143 | old (3.0) effective σ	| new (1.5) effective σ
---|---|---
just below boundary (N = 6)|≈ 134|≈ 141
just above boundary (N = 7)|≈ 112|≈ 145

With the old value the effective blur dropped ~134 → ~112 across the boundary (a visible sharpening step); with the fix it stays ≈ the requested σ on both sides.

### Solution
Subtract the true per-level contribution of both passes — `0.75 (downscale) + 0.75 (upscale) = 1.5` in current-grid units — instead of `3.0`. In the decimation variance budget this makes the produced variance match the target `σ²` at every level (up to minor rounding/edge/non-Gaussian residuals), so the effective blur is continuous across level boundaries and tracks the requested strength as you zoom. 

Before | After
---|---
<video src=https://github.com/user-attachments/assets/cd497eec-25c1-4eaa-92cb-d4ffaa4a2dba/>|<video src=https://github.com/user-attachments/assets/86fddc33-e047-4dcd-ad6e-9dc1bb276ef1/>

---


sparse_strips/vello_common/src/filter/gaussian_blur.rs
```diff
@@ -108,13 +108,15 @@ pub fn plan_decimated_blur(std_deviation: f32) -> (usize, [f32; MAX_KERNEL_SIZE]
...
+    // Each decimation level blurs the image *twice* over the full round trip, and both passes
+    // must be subtracted from the budget so the final result matches the target σ:
+    // 1. The downscale applies a [1,3,3,1]/8 binomial filter (variance 0.75 in the current grid).
+    // 2. The matching upscale reconstruction ([0.25,0.75] bilinear) is equivalent to the same
+    //    [1,3,3,1]/8 filter, contributing another 0.75 in the current grid.
+    // So a level removes 0.75 + 0.75 = 1.5 of variance (in current-grid units) before the 2×
+    // downsampling rescales the remaining variance by 0.25 (= 1/2²) into the next grid.
     while remaining_variance > 4.0 {
-        remaining_variance = (remaining_variance - 3.0) * 0.25;
+        remaining_variance = (remaining_variance - 1.5) * 0.25;
```
angusj-canva, line 120: I've gone through the maths, and this change looks good to me!

Tiny comment nitpick:

The "[0.25,0.75] equivalent to [1,3,3,1]/8" part of the comment is a bit of a big step. 
- I've only understood it by inserting imaginary black pixels between each pixel then applying [1,3,3,1]/8 then doubling everything to account for the dimming of all the black pixels which gets you to the alternating [0.25,0.75], [0.75, 0.25] pattern.
- Maybe there's a simpler way to see the equivalence and I'm missing it?
- Alternatively, thinking of the downscaled pixel's centres as between the higher resolution pixels' centres, you can directly calculate the variance:

```
|   |   |   |      < pre-downscaling pixel centres
  |       |	       < post-downscaling pixel centres
        ^ this pixel's value when upscaling is
  ^ 0.25 * this pixel which is 1.5 pixels away +
          ^ 0.75 * this pixel which is 0.5 pixels away
```
(the weighted mean position is the position of the pixel we're producing a value for, so distance from the pixel and distance from the mean are the same)
So variance is `0.25*(1.5)^2 + 0.75*(0.5)^2 = 0.75`

sparse_strips/vello_example_scenes/src/filter_blur_circle.rs
```diff
@@ -0,0 +1,69 @@
+// Copyright 2025 the Vello Authors
```
LaurenzV:
```suggestion
// Copyright 2026 the Vello Authors
```

sparse_strips/vello_example_scenes/src/filter_blur_circle.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+impl ExampleScene for FilterBlurCircleScene {
+    fn render<T: RenderingContext>(
+        &mut self,
+        ctx: &mut T,
+        _resources: &mut T::Resources,
+        root_transform: Affine,
+    ) {
```
LaurenzV, line 37: Up to you, but I think the two comments here are superfluous.

LaurenzV (approved):
Test changes LGTM assuming the math checks out.

sparse_strips/vello_common/src/filter/gaussian_blur.rs
grebmeg: Good point — I’ve updated the comment to use the direct variance calculation instead. Not sure it captures everything you had in mind, but hopefully it makes the `0.75` contribution easier to follow.

sparse_strips/vello_example_scenes/src/filter_blur_circle.rs
grebmeg: Thanks

sparse_strips/vello_example_scenes/src/filter_blur_circle.rs
grebmeg: I agree.

Fixed.

PR #1722, opened 2026-07-02
Add a variation of the into_png function exposing the encoder via callback

ratmice:
I wanted to modify some stuff in the png header, like setting pixel dimensions. This seemed like the simplest fix,
because it doesn't require us to make an abstraction layer for every knob in the png encoder someone might want to twist. If we hate this approach, then not a lot of time wasted.


LaurenzV:
Hmm, sorry, but I'm not sure this makes a lot of sense. This method is really just meant as a quick convenience method to get a PNG from the pixmap. If you want more customization, I think you you should instead get the raw RGBA bytes and do the encoding yourself. It's just 7 lines of code, so it's probably not too bad to copy-paste, no? 🙂

ratmice:
Reasonable enough.

PR #1723, opened 2026-07-03
vello_cpu: Use multiply-add for radial gradient calculations

LaurenzV:
```
                        time:   [647.35 ns 647.86 ns 648.38 ns]
                        change: [-5.4014% -5.1969% -4.9594%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
```

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/common/gradient/radial.rs
+++ b/sparse_strips/vello_cpu/src/fine/common/gradient/radial.rs
@@ -61,12 +61,12 @@ impl<S: Simd> SimdGradientKind<S> for SimdRadialKind<S> {
 
         match &self.inner {
             SimdRadialKindInner::Radial { bias, scale } => {
-                let radius = (x_pos * x_pos + y_pos * y_pos).sqrt();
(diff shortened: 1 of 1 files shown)


tomcur (approved):

raphlinus:
One question about this: mul_add can be a pretty bad performance hit on baseline x86_64 compiles. What we really want is algebraic_mul and algebraic_add. Alternatively, we could cook up our own cfg tree that does * and + on x86_64 without fma target feature enabled, and mul_add on more modern cfg's. (I'm not talking about multiversioning here).

LaurenzV:
In fearless_simd, this just uses a normal multiply+add fallback on SSE4 and a native FMA on AVX2, so I think it should be fine? Unless I’m missing something 😄

raphlinus:
Oh yes, never mind. For some reason when I was looking at the code I thought it was scalar math, for which my previous comment would be valid. But if these are all simd vectors, we're good.

PR #1724, opened 2026-07-04
sparse_strips: Reduce `powf` calls for rounded blurred rects

LaurenzV:
Someone with better math skills should double check this, but it makes sense to me.

```
lstampfl@MacBookPro vello_bench % cargo bench --bench main -- rounded --baseline main
fine/rounded_blurred_rect/with_transform_u8_neon
                        time:   [6.0929 µs 6.1034 µs 6.1149 µs]
                        change: [-55.059% -54.952% -54.849%] (p = 0.00 < 0.05)
                        Performance has improved.
```

For GPU, it doesn't seem to have much of an effect, but for consistency I still ported it over.

Done with assistance of GPT 5.5, high.


LaurenzV:
image
Diffs are just a few small 1 pixel changes

LaurenzV:
@nicoburns in case you want faster blurred rounded rects, this would be useful! In case you want to take a look. 😄

PR #1727, opened 2026-07-06
vello_hybrid: Extract probing code into its module

LaurenzV:
We already had a custom module for this, but didn't actually use it! Having to tweak the visibilities this way is a bit unfortunate, but I think it's good to have it live in its own module, especially since the WebGL module is already 3k+ lines long...

This PR was done with just manual copy-pasting, without AI-assistance, so no danger of accidental logic changes.


grebmeg (approved):
That makes sense! LGTM

PR #1730, opened 2026-07-08
vello_cpu: Make `paint_*` methods consume `self`

LaurenzV:
~~Based on top of #1728.~~

A painter is supposed to be initialized and then only used once, since each time we change the location we need to reinitialize it. This PR changes the signature so this intention is properly encoded in the `paint_*` methods.

This PR was done with assistance of GPT 5.6-Sol.

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/mod.rs
+++ b/sparse_strips/vello_cpu/src/fine/mod.rs
@@ -1108,7 +1108,7 @@ pub(crate) struct FineRenderParams {
 /// may convert between types as needed.
 pub trait Painter: Sized {
     /// Paint pixel data into a u8 buffer (values in 0-255 range).
-    fn paint_u8(self, buf: &mut [u8]);
+    fn paint_u8(mut self, buf: &mut [u8]);
 
     /// Paint pixel data into an f32 buffer (values in 0.0-1.0 range).
     fn paint_f32(self, buf: &mut [f32]);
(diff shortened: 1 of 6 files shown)


sparse_strips/vello_cpu/src/fine/mod.rs
```diff
@@ -1105,12 +1105,12 @@ pub(crate) struct FineRenderParams {
 ///
 /// Note: Some painters may only efficiently support one numeric type. The implementation
 /// may convert between types as needed.
-pub trait Painter {
+pub trait Painter: Sized {
```
nicoburns, line 1109: Is this supposed to be `pub`?

nicoburns (approved):
Given that these are internal types, I see no reason not to do this if it compiles.

grebmeg (approved):

sparse_strips/vello_cpu/src/fine/mod.rs
LaurenzV: Good point, I think it might have been public in the best due to being needed in `vello_bench`, but I'm not sure.

LaurenzV:
Will rebase this onto main instead since the other PR is probably going to take a while.

sparse_strips/vello_cpu/src/fine/mod.rs
LaurenzV: Ah, nvm it's needed because `FineKernel` needs to be public because of the benches.

PR #1732, opened 2026-07-09
vello_cpu: Flush `MultiThreadedDispatcher` before resetting

### LINKED ISSUE #1731: vello_cpu 0.0.9 panics with multithreading feature
I am running in Servo with 
[candle-repeat-var.html](https://github.com/user-attachments/files/29840752/candle-repeat-var.html)

If we enable multithreading feature for vello_cpu, there is a chance to panic. It seems to be a race condition in 
MultiThreadedDispatcher::reset() with coarse_task_receiver dropped early while workers are still working.

See also https://github.com/servo/servo/issues/46404

yezhizhen:
Right now, `reset()` drops `task_sender` and `recorded_command_receiver` without
waiting for in-flight workers to finish: https://github.com/linebender/vello/blob/cd852dc06bc0f9011385fc63a4852b0fc853f4b1/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L599-L600

This would panic if a worker is still sending.

Fixes: https://github.com/linebender/vello/issues/1731
Testing: Run the example in https://github.com/linebender/vello/issues/1731 for 10 minutes without panic.

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -587,6 +587,10 @@ impl Dispatcher for MultiThreadedDispatcher {
     }
 
     fn reset(&mut self, width: u16, height: u16) {
+        if self.task_sender.is_some() {
+            self.flush();
+        }
+
         // Bucketer will be reset lazily during rasterization with the active viewport.
         self.clip_context.reset();
         self.recorder.reset();
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -587,9 +587,7 @@ impl Dispatcher for MultiThreadedDispatcher {
     }
 
     fn reset(&mut self, width: u16, height: u16) {
-        if self.task_sender.is_some() {
-            self.flush();
-        }
+        self.flush();
 
         // Bucketer will be reset lazily during rasterization with the active viewport.
         self.clip_context.reset();
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -1240,7 +1240,7 @@ mod tests {
                 ..Default::default()
             },
         );
-        
+
         // Note: This test only works if we draw enough rectangles
         // to trigger a batch send.
         for _ in 0..300 {
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -1227,6 +1227,29 @@ mod tests {
         ctx.render_with(&mut pixmap, &mut resources, rasterizer_settings);
     }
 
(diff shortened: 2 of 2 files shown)


yezhizhen:
I don't know if a test is required or how to write it in vello. Do you have any suggestion pls @nicoburns?
cc @xiaochengh

LaurenzV:
Let's add this test to `render.rs`:

```rs
    #[test]
    fn multithreaded_reset_with_pending_tasks() {
        use crate::RenderSettings;

        let mut ctx = RenderContext::new_with(
            100,
            100,
            RenderSettings {
                num_threads: 4,
                ..Default::default()
            },
        );
        
        // Note: This test only works if we draw enough rectangles
        // to trigger a batch send.
        for _ in 0..300 {
            ctx.fill_rect(&Rect::new(0.0, 0.0, 100., 100.0));
        }

        ctx.reset();
    }
```

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
```diff
@@ -587,6 +587,10 @@ impl Dispatcher for MultiThreadedDispatcher {
     }
 
     fn reset(&mut self, width: u16, height: u16) {
+        if self.task_sender.is_some() {
+            self.flush();
+        }
```
LaurenzV:
```suggestion
        self.flush();
```

This _should_ be enough I think, `flush` itself already has an idempotency check, so if the task senter is none flush won't actually run. Let's try it.

yezhizhen: Thanks for review! Makes sense.
It is more about invariant that `task_sender == None` => `flushed == true` than the idempotency tho.

LaurenzV: Yeah, perhaps this should be improved. but doesn't have to be in this PR.

yezhizhen:
> Let's add this test to `render.rs`:
> 
> ```rust
> #[cfg(feature = "multithreading")]
>     #[test]
>     fn multithreaded_reset_with_pending_tasks() {
>         use crate::RenderSettings;
> 
>         let mut ctx = RenderContext::new_with(
>             100,
>             100,
>             RenderSettings {
>                 num_threads: 4,
>                 ..Default::default()
>             },
>         );
>         
>         // Note: This test only works if we draw enough rectangles
>         // to trigger a batch send.
>         for _ in 0..300 {
>             ctx.fill_rect(&Rect::new(0.0, 0.0, 100., 100.0));
>         }
> 
>         ctx.reset();
>     }
> ```

I thought the test would be flaky (that it may succeed with small chance even before the patch), but I ran 300 times and it always fails.

So the test is robust, and I'm still checking why.

LaurenzV:
It probably could be flaky in theory, but in reality rendering 300 rectangles is probably always going to be slower than whatever happens on the main thread. And now with the fix, it should never fail anyway.

LaurenzV (approved):

LaurenzV:
Thanks!

PR #1733, opened 2026-07-09
Update selected dependencies

oscargus:
Closes #1717 

Same as #1717, but without the jni update as I cannot get an Android dev environment up.

Skrifa is now at 0.44, but many of the other dependencies are not yet there, so I decided to stay at 0.43.2.


LaurenzV:
I made a new one since it was quicker than requesting changes here (#1741), thanks regardless!

PR #1734, opened 2026-07-09
vello_hybrid: Deallocate filter regions at the end of each frame instead of the start of next

LaurenzV:
Currently, the `probe` method piggybacks an existing `WebGlRenderer` and does some `mem::swap`s in an effort to not dirty the existing renderer state. Unfortunately, it turns out that this is not enough. Currently, filter handling is written in such a way that at the beginning of the render operation (i.e. when calling `render_scene`), we check the filter context and clear/deallocate image regions _from the last frame_. Obviously, this doesn't work well with probing: If we rendered a filter in the previous frame and are about to render the probe image next, it thinks it still needs to deallocate an image from the last frame, and if the image IDs happen to be the same it will instead deallocate the image that is needed by the probe itself. The fact that filter regions are currently deallocated this way is more than dubious, but I won't bother fixing this now since it will go away with the Vello Hybrid rewrite.

However, I also don't want to fix this issue with a bandaid fix by just trying to `mem::swap` the filter context and any other associated state that we missed, since this is still fragile. Therefore, we instead simply make `probe` create its own `WebGlRenderContext` and therefore leave any other existing contexts untouched. I initially didn't do this because I wanted to avoid having to allocate those additional buffers, but I now think that this is the best and cleanest way.

As a nice side effect, this also _seems_ to fix the CI flakiness, so we can run the probe in CI again.

I verified on a number of low-tier devices that this does not seem to regress the probe time itself.

This PR was done with assistance of GPT5.6-Sol.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -402,13 +402,6 @@ impl WebGlRenderer {
         clear: bool,
         root_output_target: RootRenderTarget,
     ) -> Result<(), RenderError> {
-        if !self.filter_context.filter_textures.is_empty() {
-            self.programs.clear_filter_atlas_textures(&self.gl);
-        }
-
-        self.filter_context
-            .deallocate_all_and_clear_context(image_cache);
-
         let mut encoded_paints = scene.encoded_paints.borrow_mut();
         let original_scene_paint_count = encoded_paints.len();
 
@@ -458,6 +451,10 @@ impl WebGlRenderer {
             &encoded_paints,
         )?;
 
+        // TODO: Most likely we will want to clean up even if rendering fails. The current assumption
+        // is that if something fails, the user will create a new renderer instead of trying
+        // to reuse the existing one. This should likely be changed in the future.
+
         // See: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_invalidateframebuffer
         // We want to indicate to the GPU driver that we won't read the depth buffer again
         // until the next clear. This enables the GPU to avoid storing depth tiles back to VRAM.
@@ -476,6 +473,12 @@ impl WebGlRenderer {
 
         encoded_paints.truncate(original_scene_paint_count);
         self.gradient_cache.maintain();
+
+        if !self.filter_context.filter_textures.is_empty() {
+            self.programs.clear_filter_atlas_textures(&self.gl);
+        }
+        self.filter_context
+            .deallocate_all_and_clear_context(image_cache);
 
         Ok(())
     }


sparse_strips/vello_hybrid/src/render/probe.rs
```diff
@@ -67,8 +83,24 @@ impl WebGlRenderer {
...
     /// anything in-between mostly happens asynchronously.
-    pub fn probe(&mut self) -> Result<WebGlPendingProbe, WebGlProbeError> {
-        self.probe_inner().map_err(WebGlProbeError::Render)
+    pub fn probe() -> Result<WebGlPendingProbe, WebGlProbeError> {
+        let (width, height) = vello_common::probe::canvas_size();
+        let canvas = web_sys::window()
+            .and_then(|window| window.document())
+            .and_then(|document| document.create_element("canvas").ok())
+            .and_then(|element| element.dyn_into::<HtmlCanvasElement>().ok())
+            .ok_or(WebGlProbeError::CanvasCreationError)?;
```
LaurenzV: @taj-p Is this fine to do? Can we assume that this will always yield a canvas from our side?

sparse_strips/vello_hybrid/src/render/probe.rs
```diff
@@ -67,27 +83,27 @@ impl WebGlRenderer {
...
+        canvas.set_width(u32::from(width));
+        canvas.set_height(u32::from(height));
+
+        let renderer = Self::new_with(
+            &canvas,
+            RenderSettings {
+                atlas_config: PROBE_ATLAS_CONFIG,
+                ..RenderSettings::default()
             },
         );
```
grebmeg: I think this is the biggest concern in this PR, since webgl contexts are a limited resource. You mentioned this will go away with the rewrite, but still would it be worth just adding a temporary bandaid fix?

I can think of a few possible approaches:

- the one you mentioned: add another swap for `filter_textures` (it might be the other resources wouldn’t need swapping);
- move filter image cache deallocation to the end of `render_scene`;
- make deallocation unable to free Ids from the wrong cache by associating each cache with a generation or identity.

My concern is that if a user opens a page with multiple vello_hybrid instances running, we could see more webgl context losses, especially since other parts of the page may also consume additional contexts. What do you think?

LaurenzV: Yeah sure, we can do 1) or 2). Not so sure about 3) for now. If we can convince ourselves that 1) is enough, this is probably the easiest. But I would need to take some time to look into whether this is actually enough again.

grebmeg:
Is this ready for me to review again?

LaurenzV:
Yes sorry!

grebmeg (approved):
> Yes sorry!

No worries! Thanks for the fix!

PR #1735, opened 2026-07-09
vello_hybrid: Get specific WebGL error in probe

LaurenzV:
Instead of just returning a generic error, get the specific WebGL error code and return it.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/probe.rs
+++ b/sparse_strips/vello_hybrid/src/render/probe.rs
@@ -9,7 +9,7 @@ use crate::render::webgl::{
 };
 use crate::schedule::RootRenderTarget;
 use crate::{RenderError, RenderSize, Scene, WebGlRenderer};
-use alloc::{format, string::String, sync::Arc};
+use alloc::{borrow::Cow, format, sync::Arc};
 use core::ops::Deref;
 use thiserror::Error;
 use vello_common::filter_effects::Filter;
@@ -232,7 +232,7 @@ impl WebGlPendingProbe {
     }
 }
 
-fn webgl_error_name(error: u32) -> String {
+fn webgl_error_name(error: u32) -> Cow<'static, str> {
     let name = match error {
         WebGl2RenderingContext::NO_ERROR => "NO_ERROR",
         WebGl2RenderingContext::INVALID_ENUM => "INVALID_ENUM",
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render/probe.rs
```diff
@@ -213,11 +213,26 @@ impl WebGlPendingProbe {
         Probe::from_actual(pixmap)
     }
 
-    fn finish_failure(&mut self) -> WebGlProbeError {
-        WebGlProbeError::ReadbackFailed
+    fn finish_failure(&self) -> WebGlProbeError {
+        WebGlProbeError::FinishFailed(webgl_error_name(self.gl.get_error()))
```
grebmeg: Could we store the raw `u32` in `FinishFailed` instead of eagerly building a `String`? Something like:

```rust
#[error("probe failed to finish: {}", webgl_error_name(*.0))]
FinishFailed(u32),
```

LaurenzV: Done.

LaurenzV:
Changed this now so it can be merged directly without depending on #1734.

sparse_strips/vello_hybrid/src/render/probe.rs
```diff
@@ -38,9 +38,9 @@ pub enum WebGlProbeError {
     /// Rendering the probe scene failed.
     #[error("probe render failed: {0}")]
     Render(RenderError),
-    /// Probe readback failed.
-    #[error("probe readback failed")]
-    ReadbackFailed,
+    /// Finishing the probe failed.
+    #[error("probe failed to finish: {}", webgl_error_name(*.0))]
```
grebmeg, line 42: [nit]: Up to you, but maybe we could save a small amount of memory by showing the raw `u32` values here instead. Right now we'd store each webgl error name as a `&'static str` as `webgl_error_name` currently returns a `String`, which means it takes the `&'static str`, allocates a new buffer on the heap, and copies the bytes into it? Returning a `Cow<'static, str>` instead might avoid that allocation in the common case. Something like:

```rust
fn webgl_error_name(error: u32) -> Cow<'static, str> {
    let name = match error {
        WebGl2RenderingContext::NO_ERROR => "NO_ERROR",
        WebGl2RenderingContext::INVALID_ENUM => "INVALID_ENUM",
        WebGl2RenderingContext::INVALID_VALUE => "INVALID_VALUE",
        WebGl2RenderingContext::INVALID_OPERATION => "INVALID_OPERATION",
        WebGl2RenderingContext::INVALID_FRAMEBUFFER_OPERATION => "INVALID_FRAMEBUFFER_OPERATION",
        WebGl2RenderingContext::OUT_OF_MEMORY => "OUT_OF_MEMORY",
        WebGl2RenderingContext::CONTEXT_LOST_WEBGL => "CONTEXT_LOST_WEBGL",
        _ => return Cow::Owned(format!("UNKNOWN_WEBGL_ERROR ({error:#06x})")),
    };
    Cow::Borrowed(name)
}
```

grebmeg (approved):

sparse_strips/vello_hybrid/src/render/probe.rs
LaurenzV: Changed this.

PR #1736, opened 2026-07-09
vello_common: Apply image sampler opacity instead of panicking

AdrianEddy:
`Image::encode_into` currently does `unimplemented!("Applying opacity to image commands")` whenever `ImageSampler.alpha != 1.0`, so any image drawn with less-than-full opacity panics the renderer.

This folds `sampler.alpha` into the paint's straight-alpha tint (scaling the tint's alpha channel, synthesizing a white `Multiply` tint when none exists) and resets `sampler.alpha = 1.0`. It's output-equivalent for both tint modes, since the strip shader computes `sample_premul * tint_premul` (`Multiply`) and `tint_premul * sample.alpha` (`AlphaMask`) — pre-scaling the tint's alpha scales the result uniformly.

Rebased now that #1791 has landed: the transparent-tint GPU fix that used to be the second commit here is gone, since #1791 covers it (with a nicer encoding), so this is just the encoder-side fold. The `sampler.alpha != 1.0` arm of `has_opacity` is dropped along with it — after the fold the sampler alpha is always `1.0`, and the tint's alpha carries the opacity.

**Tests:** 4 encoder unit tests (both tint modes, α = 0, scaling an existing tint), plus an `image_sampler_alpha` snapshot test covering untinted, tinted, α = 1 and α = 0 on both the CPU and hybrid renderers.

### DIFF
--- a/sparse_strips/vello_common/src/encode.rs
+++ b/sparse_strips/vello_common/src/encode.rs
@@ -10,7 +10,7 @@ use crate::color::{ColorSpaceTag, HueDirection, Srgb, gradient};
 use crate::geometry::RectU16;
 use crate::kurbo::{Affine, Point, Vec2};
 use crate::math::{FloatExt, compute_erf7};
-use crate::paint::{Image, ImageSource, IndexedPaint, Paint, PremulColor, Tint};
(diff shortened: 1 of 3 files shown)


LaurenzV:
Created a new one at #1880 based on yours since I was unable to push to your branch, thanks for the fix!!

PR #1737, opened 2026-07-09
vello_hybrid: Partial (damage-region) rendering via RenderRegion

AdrianEddy:
Adds partial (damage-region) rendering to `vello_hybrid`: a `RenderRegion` parameter on `Renderer::render` that confines drawing to a set of damage rectangles, so a frame where only a small area changed only pays for that area.

```rust
renderer.render(
    scene, resources, device, queue, &mut encoder, &render_size, &view, &bindings,
    ClearSettings::DontClear,            // or Rects { .. } when the damage isn't repainted opaquely
    RenderRegion::Rects(&damage_rects),  // pairwise-disjoint damage rects
)?;
```

- `RenderRegion::Full` draws the whole target; existing callers just pass it.
- `RenderRegion::Rects` takes a set of pairwise-disjoint rects. Each root pass replays its draw sequence once per rect under that rect's scissor — scissoring is dynamic pass state, so the target load is still paid once. Pixels outside the rects are preserved byte-exact; pixels inside are byte-identical to a full render of the same scene. A diagonal pan's two edge strips, or two dirty widgets in opposite corners, are a single render call.
- The scheduler culls root strips that fall entirely outside every rect before they reach the GPU, so the per-render CPU cost scales with the damaged region instead of the full scene. Intermediate layer and atlas targets are never culled or scissored — the root pass samples them.
- Rects are clamped to the target, and a region that clamps entirely away draws nothing rather than silently falling back to a full render. The `partial_renders` / `culled_strips` counters let damage-tracking callers assert the machinery actually engaged.

This sits on top of #1783 (the first commit here is that PR's) and pairs with its `ClearSettings`: a pass-level load op ignores scissors per WebGPU, so damage rendering uses `DontClear` when the scene repaints the region opaquely, or a `Rects` clear over the same rectangles when it doesn't.

Byte-exactness between a partial and a full render relies on the same content taking the same rasterization path either way — #1784 makes that hold for rects under clips. Construction-time culling (clip-based, per the earlier review discussion) is a follow-up on top of these.

Tests: `partial_render.rs` pins scissor confinement, damage-preserve byte equality against a full render, and disjoint multi-rect composition with the gap between the rects preserved, plus unit tests for the clamping and the scheduler cull. The full snapshot suite passes unchanged.

The WebGL backend is untouched for now: the scheduler plumbing is shared, but `Load` on a default drawing buffer needs `preserveDrawingBuffer`, which deserves its own follow-up with the caveats documented.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/wgpu.rs
+++ b/sparse_strips/vello_hybrid/src/render/wgpu.rs
@@ -22,7 +22,8 @@ use crate::draw::ExternalTextureRun;
 use crate::render::common::IMAGE_PADDING;
 use crate::util::RangedSlice;
 use crate::{
-    ClearSettings, GpuStrip, LayersConfig, RenderError, RenderSettings, RenderSize, Resources,
+    ClearSettings, GpuStrip, LayersConfig, RenderError, RenderRegion, RenderSettings, RenderSize,
+    Resources,
     blend::{BlendStrip, GpuBlendInstance},
     copy::GpuCopyInstance,
     filter::{FilterContext, FilterInstanceData, FilterPassPlan},
@@ -150,6 +151,37 @@ impl TextureBindings {
     }
 }
 
+/// Clamp the [`RenderRegion`] to the `size` render target so the draw-time
+/// scissor never exceeds the render attachment.
+///
+/// [`RenderRegion::Full`] yields an empty set (no scissoring). For
+/// [`RenderRegion::Rects`], rects are clamped and dropped when they become
+/// empty; if none survive (including an empty input set), a single zero-area
+/// sentinel is returned instead: the strip cull then discards every root strip,
+/// so the render draws nothing rather than silently falling back to a full,
+/// unscissored render.
+fn clamp_region_to_target(region: RenderRegion<'_>, size: &RenderSize) -> Vec<RectU16> {
+    let RenderRegion::Rects(rects) = region else {
+        return Vec::new();
+    };
+    let w = size.width.min(u32::from(u16::MAX)) as u16;
+    let h = size.height.min(u32::from(u16::MAX)) as u16;
+    let mut clamped: Vec<RectU16> = rects
+        .iter()
+        .filter_map(|r| {
+            let x0 = r.x0.min(w);
+            let y0 = r.y0.min(h);
+            let x1 = r.x1.min(w);
+            let y1 = r.y1.min(h);
+            (x1 > x0 && y1 > y0).then(|| RectU16::new(x0, y0, x1, y1))
+        })
+        .collect();
+    if clamped.is_empty() {
+        clamped.push(RectU16::new(0, 0, 0, 0));
+    }
+    clamped
+}
+
 /// Vello Hybrid's Renderer.
 #[derive(Debug)]
 pub struct Renderer {
@@ -167,6 +199,14 @@ pub struct Renderer {
     layers_config: LayersConfig,
     #[cfg(feature = "text")]
     atlas_clear_scratch: Vec<u8>,
+    /// Number of [`Self::render`] calls that drew a partial
+    /// ([`RenderRegion::Rects`]) region. Lets damage-tracking callers assert
+    /// that partial rendering actually happened rather than silently falling
+    /// back to a full render.
+    partial_renders: u64,
+    /// Number of root-destined strips skipped by the damage-region cull across
+    /// all renders. Lets damage-tracking callers assert the cull engaged.
+    culled_strips: u64,
 }
 
 impl Renderer {
@@ -210,7 +250,24 @@ impl Renderer {
             layers_config: layer_config,
             #[cfg(feature = "text")]
             atlas_clear_scratch: Vec::new(),
+            partial_renders: 0,
+            culled_strips: 0,
         }
+    }
+
+    /// Number of [`Self::render`] calls that drew a partial
+    /// ([`RenderRegion::Rects`]) region. Lets damage-tracking callers assert
+    /// the partial path actually engaged.
+    pub fn partial_renders(&self) -> u64 {
+        self.partial_renders
+    }
+
+    /// Number of root-destined strips skipped by the damage-region cull across
+    /// all renders (the fast-path coverage, gap, and rectangle quads that fell
+    /// entirely outside the [`RenderRegion::Rects`] set). Lets damage-tracking
+    /// callers assert the cull actually engaged.
+    pub fn culled_strips(&self) -> u64 {
+        self.culled_strips
     }
 
     /// Render `scene`.
@@ -220,6 +277,12 @@ impl Renderer {
     /// requirements on the bound texture views.
     ///
     /// To render without any texture bindings, you can pass an empty [`TextureBindings`].
+    ///
+    /// For damage-region rendering, pass the damaged rects as
+    /// [`RenderRegion::Rects`]: only that region is redrawn (byte-identical to a
+    /// full render of the same scene) and the rest of the target is preserved.
+    /// Pair it with [`ClearSettings::DontClear`], or [`ClearSettings::Rects`]
+    /// over the same rects when the scene does not repaint the region opaquely.
     pub fn render(
         &mut self,
         scene: &Scene,
@@ -231,7 +294,12 @@ impl Renderer {
         view: &TextureView,
         texture_bindings: &TextureBindings,
         clear: ClearSettings<'_>,
+        region: RenderRegion<'_>,
     ) -> Result<(), RenderError> {
+        if !matches!(region, RenderRegion::Full) {
+            self.partial_renders += 1;
+        }
+
         #[cfg(feature = "text")]
         {
             resources.before_render(
@@ -273,6 +341,7 @@ impl Renderer {
             &resources.image_cache,
             &scene.encoded_paints,
             clear,
+            region,
             RootTarget::UserSurface,
             texture_bindings,
         );
@@ -370,6 +439,7 @@ impl Renderer {
             &dummy_image_cache,
             encoded_paints,
             ClearSettings::DontClear,
+            RenderRegion::Full,
             RootTarget::AtlasLayer,
             texture_bindings,
         );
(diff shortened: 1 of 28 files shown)


sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -89,6 +89,46 @@ pub struct RenderTargetConfig {
...
+    /// re-rendered, and every root strip pass is scissored to it. The rect is
+    /// clamped to `render_size`.
+    ///
+    /// Intermediate passes (slot textures, filter atlases) are not scissored —
+    /// their texels reach the target only through the scissored root passes.
+    /// The depth buffer is renderer-internal per-render scratch: it is still
+    /// fully cleared at the first root pass regardless of the scissor, which is
+    /// correct because scissored draws cannot write depth (or color) outside
+    /// the region, and depth never carries state across renders.
+    Region([u32; 4]),
```
taj-p: I think this pattern lacks the type of flexibility a "damage region aware" consumer will need.

For example, a typical use case for damage regions is panning. But, a diagonal pan (which creates a damage region along two edges) would, via this API, require 2 separate `render` calls - 2 draw calls where one would do.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -89,6 +89,46 @@ pub struct RenderTargetConfig {
...
 }
 
+/// Per-render options for [`Renderer::render_with_options`].
+///
+/// [`RenderOptions::Full`] (the default) is exactly [`Renderer::render`]: clear
+/// the whole target, then rasterize everywhere. The enum shape makes the three
+/// valid modes explicit and the "clear everything but only draw a region"
+/// footgun unrepresentable.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub enum RenderOptions {
```
taj-p: On low tier mobile devices, it's much better to `wgpu::LoadOp::DontCare` when you know you'll be overwriting those pixels with some opaque background. This avoids a clear and an expensive `wgpu::LoadOp::Load` which pays expensive reads from the framebuffer to initialize memory.

sparse_strips/vello_hybrid/src/render/wgpu/mod.rs
taj-p, line 1: How do you plan to use this API?

One potential issue with this approach is that culling happens at the end of the pipeline. It's much better, at scene construction, for a consumer to `push_clip_path` the damage regions so they don't pay per-pixel costs for pixels that will eventually be thrown away at rasterization time.

For example, if I render an large image as the background. When I construct my scene, without clipping, we'll prepare all the strips for the entire viewport (re-sizing a scene is expensive, so we generally keep the same one around). Then, if I only want to render a small region of it, we'll use `RenderOptions` to provide culling after - nominally - a lot of the expensive work is over.

But, if you're rendering the same scene (without any scene mutations) between renders, then this approach could be worthwhile. Is that what you're doing?

I'm actually wondering whether a better approach here is to push the complexity to the consumer. We could expose the clear pipeline for ergonomics (that is, the consumer can `renderer.clear_rect(...)`) and then the consumer simply specifies at render time whether they want to `LoadOp::[Clear(RGBA)|DontCare|Load]`. 

My preference for Vello is for it to just focus on painting pixels as fast as it can and leave "compositor-like" smarts to the higher level.

taj-p (commented):
Nice! Thanks for the contribution - really appreciate it 🙏 . I left some questions.

sparse_strips/vello_hybrid/src/render/wgpu/mod.rs
taj-p: I.e. a lot of the complexity in Vello is reduced if a consumer instead:

```rs
scene.push_clip_path(/* Damage region rect */);
// build scene
scene.pop_clip_path();

renderer.clear_rect(...., /* Damage region rect */); /* May be unnecessary if // build scene pushes opaque bg

renderer.render(scene, ... , LoadOp::Load);
```

Or alternatively,

```rs
scene.push_clip_path(/* Damage region rect */);
// Push opaque background
// build scene
scene.pop_clip_path();

renderer.render(/* texture target */, scene, ... , LoadOp::DontCare);

// Copy Damage region rect from texture to surface.
```

sparse_strips/vello_hybrid/src/render/wgpu.rs
LaurenzV: Yes, this should be a &[RectU16] instead.

LaurenzV (commented):
Mostly echoing what Taj said. The ability to define whether and what regions should be cleared is fine and important, but please remove the other culling stuff you added, you should just push clip paths in your damage regions if you want this.

Also, I noticed that you added a loot of comments (also in your other PRs), most of which seem superfluous to me. It would be great if you could cut away most of them and only keep the ones that really explains something non-obvious. 🙏  Documentation for struct fields is fine, but for example the `Region` one is also overly long. I think this one can for example be described in just 1-2 sentences, something like "Only clear regions in the root target that are contained within the rectangle" (or rectangles if we add multiple).

Thank you!

sparse_strips/vello_hybrid/src/render/wgpu.rs
LaurenzV: Although the lifetime might be annoying. Let’s see.

AdrianEddy:
I reshaped the PR to the consumer-driven decomposition suggested by @taj-p:

- `RenderConfig { load: TargetLoad, scissors: &[RectU16] }` - the scissors are a set of pairwise-disjoint damage rects, so a diagonal pan's two edge strips are one render call. Each root pass runs its draw sequence once per rect under that rect's scissor (dynamic pass state, so the target load is still paid once).
- `TargetLoad::{Clear, Load, DontCare}` rides the first root strip pass's `LoadOp` - there's no standalone clear pass anymore, and `DontCare` forwards wgpu's opaque-coverage token for the tiled-GPU case you described. Your "render to scratch + copy the rect" sketch works on top of this today for full-target scratches; region-sized scratches would additionally need a render offset in vello, which I think is worth a separate issue.
- `clear_rects` is the caller-driven region clear (one scissored-quad pass for N rects), used with `Load` when the scene won't repaint the damage opaquely.

In my engine, the scene is rebuilt every frame - my retained scene graph culls at encode time, so on a partial frame the vello `Scene` only ever receives content intersecting the damage. 


> One potential issue with this approach is that culling happens at the end of the pipeline. It's much better, at scene construction, for a consumer to push_clip_path the damage regions so they don't pay per-pixel costs for pixels that will eventually be thrown away at rasterization time.

I tried this first, and it initially failed to produce same results, because with any clip active, rects (including image and glyph quads) leave the GPU analytic rect path and go through CPU strip generation + mask intersection — and the two rasterizers quantize fractional-edge anti-aliasing differently. The analytic path quantizes the edge position to unorm8 and computes coverage in the shader; the strip path computes exact f32 coverage and quantizes after. At half-ties they round in opposite directions, so a full-viewport integer clip that removes nothing still changed rendered bytes by ±1 LSB at fractional edges. My damage pipeline treats "partial render == full render, byte-for-byte" as a hard invariant, so that ruled the clip recipe out as it stood.

Rather than drop the idea, the second commit makes it byte-exact: clip paths that are disjoint, axis-aligned integer-rectangle sets are detected in ClipContext, and rect content clamps to the set analytically instead of demoting (an integer clip edge is exactly a 0/255 coverage step, so this is provably identical to the mask intersection). `Scene::set_cull_hint` covers the scene-construction cost for everything else by culling flattening/strip generation to the damage bounds — including line segments now, which previously never flatten-culled at all. There's a GPU test (clip_exact.rs) pinning clipped-vs-unclipped byte equality for rects, texture rects, paths and strokes.

Nice side effect: this speeds up every rectangular clip viewport (scrolling containers), damage or not — on a clip-heavy stress scene in my engine it's ~2.3× on scene encode and ~2× on the frame.

taj-p:
> I reshaped the PR to the consumer-driven decomposition suggested by @taj-p:
> 
>  ...

@AdrianEddy - we really love your work. You're opening multiple valuable PRs that moves our roadmap forwards and squashes bugs. This contribution is truly valued.

We've received some feedback that it's hard to parse the PR descriptions and your responses because they feel very LLM heavy. This is our [LLM policy](https://linebender.org/wiki/llm-policy/).

---

> I tried this first, and it initially failed to produce same results, because with any clip active, rects (including image and glyph quads) leave the GPU analytic rect path and go through CPU strip generation + mask intersection — and the two rasterizers quantize fractional-edge anti-aliasing differently.

I wonder if this is a bug in strip generation / GPU analytic rect path and so, instead of building an API to side step the bug, we should fix the bug and propose the best API assuming a "no bug" implementation.

> Scene::set_cull_hint

This is a really cool idea. I love that it enables us to avoid paying flattening / strip generation costs!! I'm wondering whether this optimisation could happen by capturing "rect" clip paths (maybe `push_clip_rect` or similar) and using that for culling, so we don't introduce both culling and clip concepts.

Separately, I think the renderer side culling / clear changes can be a separate PR to the culling that occurs on the front end of the pipeline

LaurenzV:
> I wonder if this is a bug in strip generation / GPU analytic rect path and so, instead of building an API to side step the bug, we should fix the bug and propose the best API assuming a "no bug" implementation.

I agree, my preference would be to first find out what exactly causes the difference between the two rect paths and see whether we can fix that. If this can be fixed, then I don't see an immediate need for adding all of the new culling code, thought something in that direction would still be nice to have for performance reasons (see below).

> This is a really cool idea. I love that it enables us to avoid paying flattening / strip generation costs!! I'm wondering whether this optimisation could happen by capturing "rect" clip paths (maybe push_clip_rect or similar) and using that for culling, so we don't introduce both culling and clip concepts.

I would like to avoid having an explicit `set_cull_hint` method, but I agree that we should have a `clip_rect` method. A first step would be to just use the rect fast path there, and in a second step we might be able to special-case integer-snapped clip rectangles for the GPU. However, I think this requires some more thinking about the design and there are some tricky aspects to it (because some parts of the code assumed tile-alignment), so it's something I would prefer to implement myself.

> Separately, I think the renderer side culling / clear changes

Yes, I agree with that. I might try to sketch out a proposal on the weekend, if time permits.

AdrianEddy:
@taj-p and @LaurenzV thanks for the reviews, it's been very useful to improve this feature even more.

I've now reshaped this PR to build on top of LaurenzV's #1783.

Additionally, I created two new PRs fixing the pixel divergence on two different planes: #1784 is exactly the fix you asked for + a similar one in glifo: #1785

The fast-path optimization part of this PR - I opened that as a separate PR (#1786), because even after fixing the pixel divergence, I think it's still useful to get the better performance.

As for the LLMs: Yes, all these PRs are generated with the help of Claude (I included that info in the PR description, but looks like it was later overwritten - I added it back now)
However, all these changes are driven by real-world usage (I have the real engine - it's not theoretical) and I have the goal to achieve the best possible performance in drawing UIs.

I'm dedicated to make these PRs as useful as possible to everyone and to align with your vision/design/APIs and roadmaps so I'm very happy to go with any changes you propose so that it's a win-win for everyone involved.

I'm also very meticulous when it comes to the implementations itself - using smartest models, multiple design/review rounds from different models as well (Opus/Fable/Codex). I agree that they like to create too long comments and PR descriptions so I'll try to make them more digestible from now on. 

Could I write these PRs without LLMs? No, I'm not familiar enough with the vello internals. But I do understand these changes, I know what they do and what they are for. 

I hope it's fine to continue opening these PRs for useful things, because I really like vello and what it achieves and I'm happy to help it make even better. Feel free to tell me that's something is not useful or if you have any objections at any point though, I'm not here to create useless slop.

PR #1738, opened 2026-07-09
vello_hybrid: Cache depth textures per render size

AdrianEddy:
The renderer recreates its `Depth24Plus` texture whenever `RenderSize` differs from the previous `render()` call. A frame that interleaves offscreen layer renders (at a different size) with the main render therefore reallocates the depth texture on every size switch, every frame (~8 MB for a 1080p depth texture). Separately, atlas-target renders allocated a full-atlas depth texture (~67 MB at the default 4096² atlas) that no pass ever reads.

This adds a per-size `DepthTextureCache` (byte-budgeted, default 32 MB; evict-largest-first with an LRU tiebreak; working-set protection driven by `depth_cache_frame_tick` so a size the frame cycles through is never churned). Depth is now ensured lazily and only for `RootRenderTarget::UserSurface` renders — atlas renders no longer allocate it at all.

New accessors: `depth_recreations`, `depth_cache_bytes`, `set_depth_cache_budget`, `depth_cache_frame_tick`, `clear_depth_cache`.

**Tests:** the victim-selection logic is extracted as a pure `select_eviction_victim` and unit-tested (largest-first, LRU tiebreak, working-set protection).

### DIFF
--- a/sparse_strips/vello_hybrid/CHANGELOG.md
+++ b/sparse_strips/vello_hybrid/CHANGELOG.md
@@ -17,6 +17,7 @@ This release has an [MSRV][] of 1.88.
 - Inverse blurred rounded rectangles through the new `invert` parameter on `Scene::fill_blurred_rounded_rect`, enabling inset box shadows. ([#1715][] by [@nicoburns][])
 - `Resources::new_with_config` for configuring the image and glyph atlas used by resources. ([#1750][] by [@grebmeg][])
 - `FilterPrimitive::DropShadowOnly` for rendering a drop shadow without compositing the original input over it. ([#1763][] by [@LaurenzV][])
+- Per-render-size depth-texture cache (`Renderer::depth_cache_frame_tick`, `set_depth_cache_budget`, `clear_depth_cache`, `depth_recreations`, `depth_cache_bytes`) that avoids reallocating the depth texture on every size switch in alternating-size render loops. ([#1738][] by [@AdrianEddy][])
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


LaurenzV:
Can we not just keep it as it is and only grow the depth texture in both dimensions instead of ever shrinking it? Yes it's still not ideal (rendering to a 4kx1 and then 1x4k texture would give us a 4kx4k texture in total), but I'm afraid trying to cache them is a bit overkill. I know the current API is still far from ideal, but I think this would be a compromise in the meanwhile? Unless I'm missing something. But I don't think the depth texture needs to match the target texture exactly in size, right?

AdrianEddy:
WebGPU specs says that the depth texture **does** need to match the target exactly: https://gpuweb.github.io/gpuweb/#abstract-opdef-gpurenderpassdescriptor-valid-usage
> 7. For each [view](https://gpuweb.github.io/gpuweb/#dom-gpurenderpasscolorattachment-view) in non-null members of this.[colorAttachments](https://gpuweb.github.io/gpuweb/#dom-gpurenderpassdescriptor-colorattachments) and this.[depthStencilAttachment](https://gpuweb.github.io/gpuweb/#dom-gpurenderpassdescriptor-depthstencilattachment).[view](https://gpuweb.github.io/gpuweb/#dom-gpurenderpassdepthstencilattachment-view), if present, the [[[renderExtent]]](https://gpuweb.github.io/gpuweb/#dom-gputextureview-renderextent-slot) must match.

and `wgpu` rejects this with [AttachmentsDimensionMismatch](https://github.com/gfx-rs/wgpu/blob/1a3a4cc27d8082dfd07beaf76dcf3dfb78dda321/wgpu-core/src/command/render.rs#L937-L943) error

LaurenzV:
Ah damn, that's unfortunate....

LaurenzV:
This PR will be superseded by #1810. It moves the responsibility of providing (and managing) depth textures to the consumer. Therefore, Vello Hybrid doesn't need to manage them anymore, and users can implement their own logic for this.

PR #1739, opened 2026-07-09
vello_hybrid: Atlas texture lifecycle fixes and external-render accessors

AdrianEddy:
Two related atlas lifecycle fixes for hosts that manage image/glyph atlases and rasterize vector content into atlas layers, plus the accessors external atlas driving needs:

- **`destroy_image` ordering fix** — destruction is now deferred to the start of the next `render` call, where the freed slot is zeroed via `queue.write_texture` and returned to the allocator. This solves both orderings: an already-encoded (but not yet submitted) render still samples the image intact (queued texture writes run before the command buffers of the *next* submit), and a slot reused within the same frame is no longer wiped after its fresh upload (queue writes run in call order).
- **`render_to_atlas` slot clearing** — a new `clear_rect: Option<RectU16>` clears the destination slot to transparent before compositing (recorded after the atlas grows, before the scene render), so content rendered into a *reused* slot doesn't blend over stale pixels — the atlas is composited with `SrcOver` under the freshly-allocated-slots-are-transparent invariant.
- **External-render accessors** — `Resources::image_cache`/`image_cache_mut` and `Scene::transform`, so external code can drive `render_to_atlas`.

**API note (breaking):** `Renderer::destroy_image` now takes `&Resources` and an `ImageId`; the `encoder` parameter is gone and `resources` is no longer `&mut` (the actual deallocation happens inside the next `render`).

**Tests:** two GPU regression tests — `atlas_slot_reuse_survives_same_frame_destroy` (destroy → reuse → render must show the new image) and `destroyed_image_survives_already_encoded_render` (destroy after encoding but before submitting — the encoded render must still sample the image).

### DIFF
--- a/sparse_strips/vello_hybrid/CHANGELOG.md
+++ b/sparse_strips/vello_hybrid/CHANGELOG.md
@@ -19,11 +19,11 @@ This release has an [MSRV][] of 1.88.
 
 ### Changed
 
-- `Renderer::destroy_image` no longer takes an `encoder`; it clears the freed atlas slot through the `queue`. ([#1739][] by [@AdrianEddy][])
+- `Renderer::destroy_image` no longer takes an `encoder` or `queue`, and takes `&Resources`; the destruction is deferred to the start of the next `render` call. ([#1739][] by [@AdrianEddy][])
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


AdrianEddy:
Rebased onto current main. I also removed `reset_atlas_textures` from this PR: after #1794 tied `Resources` construction to its `Renderer`, resetting the atlas array without also replacing the paired Resources is exactly the desync that pairing is meant to prevent, and rebuilding the pair covers the memory-pressure case fine. What remains is the deferred `destroy_image` fix, the `render_to_atlas clear_rect`, and the accessors. PR description updated to match.

PR #1740, opened 2026-07-11
vello_common: Remove failing test

LaurenzV:
This test fails since #1714, the reason is that since the PR we ensure that no zero-width strips can be emitted at places other than the row end. 

```
---- clip::tests::row_iterator_zero_width_alpha_region stdout ----

thread 'clip::tests::row_iterator_zero_width_alpha_region' (957770) panicked at sparse_strips/vello_common/src/clip.rs:549:17:
zero-width strips must only appear at the end of a row


failures:
    clip::tests::row_iterator_zero_width_alpha_region
```

This is true for our current strip generation code, so the assertion is correct, but of course in a unit test it's still possible to manually construct such strip. The reason this wasn't caught in the PR is that we run tests in release mode now. :( We should probably consider running at least the `vello_common` (and maybe `vello_cpu`) unit tests in debug mode in CI.

### DIFF
--- a/sparse_strips/vello_common/src/clip.rs
+++ b/sparse_strips/vello_common/src/clip.rs
@@ -846,28 +846,6 @@ mod tests {
         run_test(expected, path, cover);
     }
 
-    #[test]
-    fn row_iterator_zero_width_alpha_region() {
-        let mut path = StripStorage::default();
-        path.strips.push(Strip::new(8, 0, 0, false));
-        path.strips.push(Strip::new(16, 0, 0, true));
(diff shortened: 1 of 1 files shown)


DJMcNab:
It seems more likely to me that you'd want to test in release mode but with debug assertions enabled?

LaurenzV:
The problem is I think we have some debug assertions that are really slow (for example in vello hybrid Inthink there is one with quadratic behavior), and the whole point of running tests in release mode was to ensure that CI isn’t too slow. But it’s worth trying!

grebmeg (approved):

PR #1741, opened 2026-07-13
Update dependencies

LaurenzV:
Same as #1733, but updated to newet main and made sure the lock file is still up-to-date. The original author's branch seems to be locked, so I had to create a new branch.


nicoburns (approved):

oscargus:
Nice! (Sorry for locking the branch. Not on purpose...)

PR #1744, opened 2026-07-14
sparse_strips: Move some code from `vello_cpu` to `vello_common`

LaurenzV:
This is the first PR in a series of three (perhaps four) PRs that serve as a preparation for the Vello Hybrid rewrite.

This PR is easiest reviewed commit by commit, though there isn't that much to review in the first place. Apart from adjusting a few descriptions, it's mostly just moving some code around.

The first 7 commits were done by manually copy-pasting code into the right location. The last commit that fixes compiler issues were done with AI assistance.

### PR train

1. #1744 — Move some code from `vello_cpu` to `vello_common` 👈
2. #1745 — Add more new/refactored code into `vello_common`
3. #1746 — Generalize the command recorder abstraction and add more metadata
4. #1751 — Introduce `Padding` and streamline clip handling

PR #1745, opened 2026-07-14
sparse_strips: Add more new/refactored code into `vello_common`

LaurenzV:
This PR is based on #1744 and introduces some actual changes, once again in preparation for the Vello Hybrid rewrite.

The first commit adds some utility types like `Offset` and `Size`, which will layer on be used by Vello Hybrid (they are currently partially duplicated there, but I will just remove them later on). In theory they could just live in Vello Hybrid, but since we already have `RectU16` in vello_common I figured this is the best place.

The second commit resolved a TODO and creates a new generic method that allows generating fill/alpha-fill sequences from a slice of strips. We also use this for the existing `strip_bbox` function now. I'm aware that this will result in a slight slowdown since we will do the clipping even though it isn't necessary to clip anything since we are just calculating the bbox, but I think it's better to accept this cost for less code duplication.

The third commit extracts all of the transforms we currently have into a custom struct, so that, once again, they can more easily be used by Vello Hybrid.

Fourth commit introduces a new `ViewportState` abstraction, that abstracts all of the complexity of handling different viewport so that Vello Hybrid can use this as well.

Fifth commit adds a way of querying whether an encoded paint has any opacities directly.

### PR train

2. #1745 — Add more new/refactored code into `vello_common` 👈
3. #1746 — Generalize the command recorder abstraction and add more metadata
4. #1751 — Introduce `Padding` and streamline clip handling


sparse_strips/vello_common/src/geometry.rs
```diff
@@ -117,6 +209,193 @@ impl RectU16 {
...
+impl From<RectU16> for SizeU16 {
+    fn from(rect: RectU16) -> Self {
+        Self::from_wh(rect.width(), rect.height())
+    }
+}
+
+// TODO: Remove these types once we've completely moved to u16 everywhere in Vello Hybrid.
+
+/// An offset represented by two 32-bit unsigned integers.
+#[repr(C)]
```
LaurenzV, line 237: These types are likely temporary and we will use u16 where possible. Unfortunately, right now we kind of convert between u16 and u32 in all kinds of different places even though it could be more unified, but I want to leave this cleanup to the future.

sparse_strips/vello_common/src/encode.rs
```diff
@@ -551,6 +553,16 @@ pub enum EncodedPaint {
 }
 
 impl EncodedPaint {
+    /// Returns whether this encoded paint is guaranteed to only produce opaque pixels.
+    pub fn is_opaque(&self) -> bool {
+        match self {
+            Self::Gradient(gradient) => !gradient.may_have_transparency,
+            Self::Image(image) => !image.may_have_transparency,
+            Self::ExternalTexture(_) => false,
```
LaurenzV: TODO: Need to implement this for external textures as well.

sparse_strips/vello_common/src/geometry.rs
```diff
@@ -3,6 +3,84 @@
...
+use bytemuck::{Pod, Zeroable};
+use core::num::TryFromIntError;
+use core::ops::Add;
+
+/// A size represented by two 16-bit unsigned integers.
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Pod, Zeroable, PartialEq, Eq)]
+pub struct SizeU16(pub [u16; 2]);
+
+impl SizeU16 {
```
grebmeg, line 16: As I understand it, the `U32` types below will be removed later? If so, do you think it's still worth introducing generic `Size<T>` and `Rect<T>` types?

sparse_strips/vello_common/src/geometry.rs
```diff
@@ -98,12 +176,26 @@ impl RectU16 {
 
     /// Return this rectangle relative to `origin`, clamping negative coordinates to zero.
     #[inline(always)]
-    pub const fn relative_to_origin(self, origin: (u16, u16)) -> Self {
+    pub fn relative_to_origin(self, origin: (u16, u16)) -> Self {
```
grebmeg, line 183: Could we add a couple of small tests for this?

sparse_strips/vello_common/src/geometry.rs
```diff
@@ -3,6 +3,84 @@
...
+    fn from(value: [u16; 2]) -> Self {
+        Self(value)
+    }
+}
+
+impl Add for SizeU16 {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        Self::from_wh(self.width() + rhs.width(), self.height() + rhs.height())
```
grebmeg: I don't think this should ever overflow in our cases, but in theory it could. Maybe it's worth adding a short comment noting that it wraps on overflow in release

sparse_strips/vello_common/src/encode.rs
grebmeg: Could we have a single method here that indicates whether opaque paint optimizations can be used in cpu/hybrid?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -25,6 +27,176 @@ pub struct Strip {
...
+    #[inline(always)]
+    pub const fn pixel_rect(self) -> RectU16 {
+        let rect = self.tile_rect();
+
+        RectU16::new(
+            rect.x0.saturating_mul(Tile::WIDTH),
+            rect.y0.saturating_mul(Tile::HEIGHT),
+            rect.x1.saturating_mul(Tile::WIDTH),
+            rect.y1.saturating_mul(Tile::HEIGHT),
+        )
```
grebmeg: Could we construct the pixel-space `RectU16` directly here? Something like:

```rust
RectU16::new(
    self.tile_x0.saturating_mul(Tile::WIDTH),
    self.tile_y.saturating_mul(Tile::HEIGHT),
    self.tile_x1.saturating_mul(Tile::WIDTH),
    self.tile_y
        .saturating_add(1)
        .saturating_mul(Tile::HEIGHT),
)
```

sparse_strips/vello_common/src/strip.rs
```diff
@@ -25,6 +27,176 @@ pub struct Strip {
...
+    /// Return this segment's pixel-space rectangle shifted by `shift`.
+    #[inline(always)]
+    pub fn shift(self, shift: (i32, i32)) -> RectU16 {
+        self.pixel_rect().shift(shift)
+    }
+}
+
+/// Iterate over all fill and alpha-fill regions formed by the sequence of strips,
+/// within the tile-unit bounds indicated by `viewport`.
+pub fn for_each_fill_segment<C>(
```
grebmeg: Could we add a few small tests for this?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -25,6 +27,176 @@ pub struct Strip {
...
+    strips: &[Strip],
+    viewport: RectU16,
```
grebmeg: Would `tile_bounds` be a more appropriate name here?

sparse_strips/vello_common/src/clip.rs
```diff
@@ -231,8 +231,8 @@ impl ClipState {
         self.context.get()
     }
 
-    /// Push a new filter surface.
-    pub fn push_filter_surface(
+    /// Push a new filter viewport.
+    pub fn push_filter_viewport(
```
grebmeg: I was trying to find an alternative to `viewport` because it easily gets confused with the rendering target viewport, and it's quite a broad term. Also, this `viewport` is expressed in tile units. What do you think about using a more specific name here and elsewhere, such as `tile_viewport`?

sparse_strips/vello_common/src/strip.rs
grebmeg: I think `for_each` is more idiomatic for an iterator method. What do you think about something like `visit_strip_fill_segments_in_viewport`?

sparse_strips/vello_common/src/strip.rs
```diff
@@ -25,6 +27,176 @@ pub struct Strip {
...
+                    fill: StripFillSegment {
+                        tile_x0,
+                        tile_x1,
+                        tile_y,
+                    },
+                    // Make sure to recalculate the index in case we had to clip.
+                    alpha_idx: strip.alpha_idx()
+                        + u32::from(tile_x0 - strip_tile_x0)
+                            * u32::from(Tile::WIDTH)
+                            * u32::from(Tile::HEIGHT),
```
grebmeg, line 173: Is this the part you mentioned in the PR description that would slightly hurt performance?

sparse_strips/vello_common/src/transforms.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn reset_paint_transform(&mut self) {
+        self.paint_transform = Affine::IDENTITY;
+    }
+
+    // Unlike [`Self::effective_path_transform`], this intentionally does not apply
+    // the root transform because clipping handles filter viewport shifts separately.
+    /// Return the transform used for non-isolated clip paths.
+    pub fn clip_path_transform(&self) -> Affine {
+        self.transform
+    }
```
grebmeg, line 76: I'm not sure it's worth adding a duplicate method with a different name here. Why not just use the `transform` method?

sparse_strips/vello_common/src/transforms.rs
```diff
@@ -0,0 +1,136 @@
...
+        // shift everything back when actually compositing the rendered filter layer.
+        let relative_transform = filter_data.map_or(Affine::IDENTITY, |filter_data| {
+            let (shift_x, shift_y) = filter_data.source_shift();
+
+            Affine::translate((f64::from(shift_x), f64::from(shift_y)))
+        });
+
+        self.transforms
+            .push(relative_transform * self.root_transform());
+    }
```
grebmeg: `RootTransforms` currently mixes two responsibilities. Could we keep `RootTransforms` independent of filters by passing the relative `Affine` into `push_root`? Deriving the transform from `FilterData` feels like filter-layer policy, while this type only needs to manage transform composition. That would make the abstraction more reusable and easier to test.

```rust
pub fn push_root(&mut self, relative_transform: Affine) {
    self.transforms
        .push(relative_transform * self.root_transform());
}

let relative_transform = filter_data.as_ref().map_or(
    Affine::IDENTITY,
    |data| Affine::translate(data.source_shift()),
);
self.root_transforms.push_root(relative_transform);

```

sparse_strips/vello_common/src/transforms.rs
```diff
@@ -0,0 +1,136 @@
...
+impl RootTransforms {
+    /// Create a new root transform stack.
+    pub fn new() -> Self {
+        Self {
+            transforms: smallvec![Affine::IDENTITY],
+        }
+    }
+
+    /// Return the currently active root transform.
+    pub fn root_transform(&self) -> Affine {
```
grebmeg, line 101: I think `cumulative_transform` would be a better name, wouldn't it? When I read `root_transform`, I think of a single fixed top-level transform. In reality, this represents the cumulative transform, so `root_transform` feels a bit misleading from that point of view. Would the Cumulative prefix make the intent clearer for this type?

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -628,8 +628,28 @@ impl CommandBucketer {
...
+        debug_assert_eq!(
+            origin.0 % Tile::WIDTH,
+            0,
+            "viewport x origin must be tile-width aligned",
+        );
+        debug_assert_eq!(
+            origin.1 % Tile::HEIGHT,
+            0,
+            "viewport y origin must be tile-height aligned",
+        );
```
grebmeg: nit: There are multiple `debug_assert_eq!` calls here. Would it make sense to move them into a dedicated helper function?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -351,12 +360,12 @@ impl Scene {
             PaintType::Solid(s) => s.into(),
             PaintType::Gradient(g) => g.encode_into(
                 &mut self.encoded_paints.borrow_mut(),
-                self.render_state.transform * self.render_state.paint_transform,
+                *self.transforms().transform() * *self.transforms().paint_transform(),
```
grebmeg: Could this be implemented as a separate method on `Transforms`, maybe something like `scene_paint_transform()`?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -402,7 +411,7 @@ impl Scene {
             let paint = ctx.encode_current_paint();
             ctx.fill_path_with(
                 path,
-                ctx.render_state.transform,
+                *ctx.transforms().transform(),
```
grebmeg: `transforms().transform()` is used everywhere, and it reads a bit awkwardly. What do you think about adding a dedicated `scene_transform()` method to replace these usages?

sparse_strips/vello_common/src/transforms.rs
```diff
@@ -0,0 +1,136 @@
...
+    // the root transform because clipping handles filter viewport shifts separately.
+    /// Return the transform used for non-isolated clip paths.
+    pub fn clip_path_transform(&self) -> Affine {
+        self.transform
+    }
+}
+
+/// Stack of root transforms.
+#[derive(Debug)]
+pub struct RootTransforms {
```
grebmeg, line 81: Maybe we could also add a few tests, mainly to showcase how the cumulative transform mechanism works?

grebmeg:
Thanks for your work! Everything looks much tidier now 🤩 I left a few comments, but they're mostly minor.

sparse_strips/vello_common/src/encode.rs
LaurenzV: What do you mean? Isn't it the same across CPU/Hybrid? I just need to make sure to read the field from the external texture.

sparse_strips/vello_common/src/geometry.rs
LaurenzV: I also considered this but probably not. For now it's easier to just keep it this way. 😄

sparse_strips/vello_common/src/geometry.rs
LaurenzV: Will add two tests.

sparse_strips/vello_common/src/geometry.rs
LaurenzV: I'll turn it into a checked add.

sparse_strips/vello_common/src/strip.rs
LaurenzV: Sure, why not!

sparse_strips/vello_common/src/strip.rs
LaurenzV: Yeah basically anything that intersects the viewport, because we don't need this when we are just calculating the bounding box.

sparse_strips/vello_common/src/transforms.rs
LaurenzV: I think it does make sense, precisely because it is special-cased. In case we ever change how clip path transforms are calculated, we just need to change the inside of the method without updating callers.

sparse_strips/vello_common/src/clip.rs
LaurenzV: Can't say I'm a huge fan, but I also don't have better ideas. 😄 Will rename it.

LaurenzV: Hmm, or maybe `root_viewport` would be better?

sparse_strips/vello_common/src/transforms.rs
LaurenzV: Makes sense.

sparse_strips/vello_common/src/transforms.rs
LaurenzV: Hmm, I can change, but I think `root_transform` makes sense. It represents the root transform for the currently active filter viewport that should be used as the basis for any subsequent path/paint transforms.

sparse_strips/vello_cpu/src/coarse/bucketer.rs
LaurenzV: Good idea!

sparse_strips/vello_common/src/encode.rs
LaurenzV: Ah, I presume you meant something like this? 😄 : https://github.com/linebender/vello/pull/1745/changes/90705e6cd8c3cbcda7306699d221024980c10b22

sparse_strips/vello_common/src/clip.rs
grebmeg: IMO: `root_viewport` seems misleading because filter viewports can be nested; it isn’t necessarily the root.

sparse_strips/vello_common/src/transforms.rs
grebmeg: That makes sense. In that case, `root_transform` sounds good to me. Could we clarify the docs to say that it is the root transform for the currently active filter viewport, including shifts inherited from nested filters?

sparse_strips/vello_common/src/encode.rs
grebmeg: Yes, sort of — if that makes sense in this context.

grebmeg:
Is this one ready for re-review too?

sparse_strips/vello_common/src/clip.rs
LaurenzV: By root viewport, I don’t mean the actual root of the layer tree, but instead “a new viewport that acts as a root because it establishes a new coordinate system for all child layers”. I don’t know, maybe it’s not worth bikeshedding, but tile_viewport doesn’t sound very obvious to me either 😄

LaurenzV:
In general yes, just haven’t requested yet because of the name thing, but I can can just go with tile_viewport if you want.

grebmeg:
> In general yes, just haven’t requested yet because of the name thing, but I can can just go with tile_viewport if you want.

This isn't critical — I'll leave the decision to you.

grebmeg (approved):
Nice! LGTM!

sparse_strips/vello_common/src/clip.rs
LaurenzV: Okay, I've just gone with root viewport for now, I hope it's not a problem.

PR #1746, opened 2026-07-14
vello_common: Generalize the command recorder abstraction and add more metadata

LaurenzV:
This PR is based on #1745. It changes the semantics of the command recorder, turning it into something more akin to a scene graph (we should probably rename it in the future!).

Up until now, the command recorder was written in a way that is only "useful" to Vello CPU. Most notably, normal layers would be folded into their parent stream, and only filter layers had an explicit representation as an independent unit. For Vello Hybrid, this is not a suitable representation, because _all_ layers need to be represented explicitly and be rendered independently from the other layers.

To achieve this goal, this PR implements some changes, split up into two commits.

The first commit represents a refactor of the overall representation. Most notably:

- It is generic over an unspecified draw type. We need this because, in addition to normal path fills as used by Vello CPU, Vello Hybrid also supports fast path rects, which have a different inner representation. Therefore, the actual draw type need to be able to be specified by each backend independently.
- Instead of only treating filter layers as having their own root commands, now every kind of layer has their own command stream. Vello CPU will internally still ensure that they are inlined, but the crucial point is that this assumption is not baked into the generic representation anymore, so Vello Hybrid can do something different here.
- Instead of having four commands "Fill`, "PushLayer", "FilterLayerFill" and "PopLayer" , we represent the whole scene as a tree of nodes, where each node represent a number of batched draw commands, followed by an optional layer invocation. Basically, we interleave normal draws and layer fills in a way such that they can be iterated over more easily. The fact that we represent a batch of draw commands as a simple range of draw commands instead of a flat array of inlined draw commands helped tremendously when building the new Vello Hybrid scheduler. Also, all draw commands now end up in a global buffer which we can reuse more easily across frames. The only downside is one more layer of indirection per batch for Vello CPU, which is completely negligible.
- Since each layer now only stores a `Vec<Node>`, which is likely to be very small, I decided to remove the `VecPool` for this. I don't think it's worth it trying to reuse those across frames.

The second commit ensures that we store some more metadata about the scene, which is needed for Vello Hybrid. Yes, we do pay some additional cost for Vello CPU here, but it's basically nothing.

### PR train

2. #1745 — Add more new/refactored code into `vello_common`
3. #1746 — Generalize the command recorder abstraction and add more metadata 👈
4. #1751 — Introduce `Padding` and streamline clip handling


sparse_strips/vello_common/src/record.rs
```diff
@@ -32,75 +32,49 @@
...
 
-/// Metadata for a recorded layer.
+/// Metadata and child nodes for a recorded layer.
 #[derive(Debug)]
 pub struct RecordedLayer {
     /// Properties of the layer.
     pub props: LayerProps,
-    /// Kind-specific layer data.
+    /// The child nodes of the layer.
+    pub nodes: Vec<Node>,
```
LaurenzV: Perhaps I should turn this into a smallvec of capacity 2-3 so we avoid the allocations for the common case?

grebmeg: I agree that a `SmallVec` makes sense here. I'd start with an inline capacity of 2, since that should cover the common case of a draw batch plus a single child, while avoiding making every `RecordedLayer` larger than necessary.

sparse_strips/vello_common/src/record.rs
```diff
@@ -145,102 +117,125 @@ pub enum RecordedLayerKind {
...
-pub struct CommandRecorder {
-    /// The commands of the root layer.
-    pub root_cmds: Vec<RecordedCmd>,
+/// Recorder for a scene description.
+#[derive(Debug)]
+pub struct CommandRecorder<D> {
+    /// Tile-aligned dimensions of the root scene.
+    pub scene_size: SizeU16,
+    /// The nodes of the root layer.
+    pub nodes: Vec<Node>,
```
grebmeg, line 156: Have you considered representing the root as a separate layer containing its own nodes? That would make the recording a more uniform tree-like structure, with the root as the tree root and all other layers as children. It might simplify traversal and remove some root-specific handling, although the root’s distinct rendering semantics may outweigh that benefit.

sparse_strips/vello_common/src/record.rs
```diff
@@ -145,102 +117,125 @@ pub enum RecordedLayerKind {
...
 }
 
-/// Recorder for fills and nested layers.
-#[derive(Debug, Default)]
-pub struct CommandRecorder {
-    /// The commands of the root layer.
-    pub root_cmds: Vec<RecordedCmd>,
+/// Recorder for a scene description.
+#[derive(Debug)]
+pub struct CommandRecorder<D> {
```
grebmeg, line 152: Would `SceneRecorder` be a more accurate name now? `CommandRecorder` suggests a flat command stream, but this type records a structured scene containing root nodes, nested layers, draws, bounds, and rendering metadata. The existing doc comment already describes it as a "recorder for a scene description".

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -284,56 +286,70 @@ impl CommandBucketer {
...
-                        self.push_layer(props);
-                    }
+        for node in nodes {
+            for RecordedFill {
+                thread_idx,
+                strip_range,
+                paint,
+                blend_mode,
+                mask,
+            } in &draws[node.draws.start as usize..node.draws.end as usize]
```
grebmeg: Would it make sense to add a helper on `Node`, such as `draws_in(&self, draws)`, that returns this slice? It would hide the repeated `u32`-to-`usize` conversion and centralize how a node’s draw range is resolved:

```rust
for fill in node.draws_in(draws) {
    // ...
}
```

sparse_strips/vello_cpu/src/dispatch/mod.rs
```diff
@@ -9,11 +9,49 @@ use crate::RasterizerSettings;
...
+use vello_common::geometry::RectU16;
 use vello_common::mask::Mask;
 use vello_common::paint::{ImageResolver, Paint};
 use vello_common::pixmap::PixmapMut;
+use vello_common::record::Drawable;
+use vello_common::strip::Strip;
+use vello_common::util::strip_bbox;
+
+#[derive(Debug)]
+pub(crate) struct RecordedFill {
```
grebmeg: Would it make sense to move `RecordedFill` into a small top-level record module? It is the CPU-specific payload shared between dispatch and coarse bucketing, rather than part of the `Dispatcher` interface.

sparse_strips/vello_cpu/src/dispatch/single_threaded.rs
```diff
@@ -195,14 +196,11 @@ impl SingleThreadedDispatcher {
...
-            paint,
-            blend_mode,
-            mask,
-            0,
-        );
+        let strip_range = strip_start..strip_end;
+        let draw = RecordedFill::new(0, strip_range.clone(), paint, blend_mode, mask);
+
+        self.recorder
+            .push_draw(draw, &self.strip_storage.strips[strip_range]);
```
grebmeg: nit:
```suggestion
        let strip_range = strip_start..strip_end;
        let strips =
            &self.strip_storage.strips[strip_range.start..strip_range.end];
        let draw = RecordedFill::new(0, strip_range, paint, blend_mode, mask);
        self.recorder.push_draw(draw, strips);
```

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -284,56 +286,70 @@ impl CommandBucketer {
...
+                        // Regular layers are inlined and bucketed into the same command stream.
+                        self.push_layer(props);
+                        self.bucket_commands(
+                            &layer.nodes,
+                            draws,
+                            layers,
+                            strips,
+                            encoded_paints,
+                            filter_ctx,
+                        );
```
grebmeg, line 331: Should we be concerned about the recursion here? Could deeply nested regular layers overflow the stack, and can we avoid that with an iterative work stack?

grebmeg (approved):
LGTM! I especially like the `Node` abstraction — it makes draw and layer ordering much clearer.

sparse_strips/vello_common/src/record.rs
LaurenzV: Yeah, in Vello Hybrid, the root requires some specific handling (because of blending). In my rewrite branch, the special casing hasn't really been a problem, so I think it's better to keep it!

sparse_strips/vello_common/src/record.rs
LaurenzV: Yes, as mentioned, I think we could even just name it `SceneGraph`! However, I don't want to load this PR more than necessary. I will add a TODO!

sparse_strips/vello_cpu/src/coarse/bucketer.rs
LaurenzV: Sure, can add this!

sparse_strips/vello_cpu/src/coarse/bucketer.rs
LaurenzV: I suppose for adverserial scenarios. But I don't think this is urgent, let me add a TODO! You'd probably have to go hundreds (or at least multiple dozens) of layers deep to trigger this. But good catch!

PR #1747, opened 2026-07-15
fix: don't allocate filter atlas textures for scenes without filters

grebmeg:
### Problem
On the vello_hybrid backends (webgl2 and wgpu), a renderer allocated filter scratch atlas texture(s) on its first frame even when the scene contained no filters — 64 MiB per texture at the default 4096² config, and 128 MiB on the wasm/wgpu path where the initial atlas count is forced to 2.

### Root Cause
`RenderSettings` exposed a single `atlas_config` shared by both the image/glyph atlas and the filter atlas. The number of filter scratch textures created (`maybe_resize_filter_atlas_textures` on webgl, `FilterAtlasState::ensure_count` on wgpu) is driven by the filter cache's `atlas_count()`, which equals `initial_atlas_count`. Because the shared config defaulted to `initial_atlas_count = 1` (and 2 on wasm for the GLES D2/D2Array workaround), the filter atlas always started with at least one atlas and materialized full-size scratch textures on the first render, regardless of whether any filter ran.

### Solution
Split `atlas_config` into two independent configs on `RenderSettings`: `image_atlas_config` (defaults unchanged) and `filter_atlas_config` (defaults to `initial_atlas_count = 0`). Each backend normalizes them separately — the image config keeps its `min_initial_atlas_count` (1, or 2 on wasm), while the filter config uses 0, since filter scratch textures are individual 2D textures and are not subject to the `D2Array` workaround. With `auto_grow`, filter atlases are now created lazily on the first actual filter pass, so filterless scenes allocate zero filter textures on both backends.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -205,12 +205,12 @@ impl WebGlRenderer {
             max_texture_array_layers,
             1,
         );
-        // Filter scratch textures are individual 2D textures (not a D2Array), so they can start
-        // at zero and only be allocated on demand when a scene actually uses filters.
         normalize_atlas_config(
             &mut settings.filter_atlas_config,
             max_texture_dimension_2d,
             max_texture_array_layers,
+            // Filter scratch textures are individual 2D textures (not a D2Array), so they can
+            // start at zero and only be allocated on demand when a scene actually uses filters.
             0,
         );
         let total_slots: usize = (max_texture_dimension_2d / u32::from(Tile::HEIGHT)) as usize;
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -198,11 +198,20 @@ impl WebGlRenderer {
 
         let mut settings = settings;
         let max_texture_dimension_2d = get_max_texture_dimension_2d(&gl);
+        let max_texture_array_layers = get_max_texture_array_layers(&gl);
         normalize_atlas_config(
-            &mut settings.atlas_config,
+            &mut settings.image_atlas_config,
             max_texture_dimension_2d,
-            get_max_texture_array_layers(&gl),
+            max_texture_array_layers,
             1,
+        );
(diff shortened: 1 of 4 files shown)


sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -168,18 +168,23 @@ impl SceneConstraints {
...
     /// - How images are allocated across multiple atlases
     /// - Whether new atlases are automatically created when needed
     ///
     /// Adjusting these settings can affect memory usage and rendering performance
     /// depending on your application's image usage patterns.
-    pub atlas_config: AtlasConfig,
+    pub image_atlas_config: AtlasConfig,
+    /// The configuration for the filter scratch texture atlas.
+    ///
+    /// Filters (e.g. blurs) render intermediate results into their own set of scratch atlas
```
LaurenzV, line 185: I mean it's not quite true since we _do_ use the image atlas as well, but let's ignore that for now. 😆

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -198,12 +198,21 @@ impl WebGlRenderer {
...
         normalize_atlas_config(
-            &mut settings.atlas_config,
+            &mut settings.image_atlas_config,
             max_texture_dimension_2d,
-            get_max_texture_array_layers(&gl),
+            max_texture_array_layers,
             1,
         );
+        // Filter scratch textures are individual 2D textures (not a D2Array), so they can start
+        // at zero and only be allocated on demand when a scene actually uses filters.
```
LaurenzV: Maybe add this comment directly above the 0?

LaurenzV (approved):
LGTM, especially since it's just temporary.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: Good point — filters do sample from the image atlas too. Reworded to "separate from the image/glyph atlas" to describe the scratch textures being a distinct allocation without overclaiming full independence.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Done

PR #1748, opened 2026-07-15
Run tests with debug assertions enabled

LaurenzV:


LaurenzV:
The Windows CI run does take longer in this PR (20 minutes vs. 12 minutes), but looking at the logs this seems to be because the dependency builds are simply not cached. I spot-checked the test runtimes and they look about the same, so I think this should be fine!

grebmeg:
@DJMcNab what do you think of this?

DJMcNab:
I don't know. If it's making ci build times that much longer, that seems pretty bad.
But as a thing to want, it seems reasonable enough.

LaurenzV:
The longer build times just seem to be because dependencies aren't cached (yet).

grebmeg:
> I don't know. If it's making ci build times that much longer, that seems pretty bad.
But as a thing to want, it seems reasonable enough.

Yes, I guess this is unfortunate for the debug assertions checks. Maybe there's a middle ground where some optimizations are enabled while keeping the debug assertions?

> The longer build times just seem to be because dependencies aren't cached (yet).

Could we verify that it's being cached by comparing the execution times of consecutive CI checks?

LaurenzV:
I think the cache doesn't trigger until it's actually merged. I'm pretty sure it's just that though, if you compare the Windows CI runs from main and from this PR, this PR is building way more dependencies. I think we could just merge this PR and then see how it goes. If CI time doesn't go down, we can revert.

DJMcNab:
We definitely do save the cache in branches - see e.g .
image

Although wow, these caches are massive, so we really hit the limit...

grebmeg:
> I think we could just merge this PR and then see how it goes. If CI time doesn't go down, we can revert.

> Although wow, these caches are massive, so we really hit the limit...

If I understand correctly, once we hit the cache limit, jobs will use the least recently used cache? Overall, we'd end up consuming a more execution time, but I'm personally fine with that for now, especially if we still have a buffer in our usage? I think we can experiment with it and see how it goes over the next few days or weeks. I'd like vello to get more CI coverage, and I'm happy to trade a bit more execution time for that.

grebmeg:
@DJMcNab are you okay with landing this?

DJMcNab:
Yeah, it's fine by me. Luckily, none of my work is blocked by Vello's CI times!

grebmeg (approved):

grebmeg:
@LaurenzV what about merging this, then monitoring how much benefit it gives us and reconsidering later if needed?

LaurenzV:
Sure, will keep an eye on it!

LaurenzV:
Doesn't seem like it's working... @DJMcNab Do you think we could try clearing up the cache to see whether that's the issue? Or should there still be sufficient storage space?

LaurenzV:
Ok it seems to have worked just now: https://github.com/linebender/vello/actions/runs/31773248831/job/94683323800

No idea....

PR #1749, opened 2026-07-15
vello_cpu: Don't unset flushed status in `MultiThreadedDispatcher.reset`

yezhizhen:
Follow up to #1732

This adds additional workload and would introduce some extra flush when task queue is empty.
cc @xiaochengh

### DIFF
--- a/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
+++ b/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
@@ -596,7 +596,6 @@ impl Dispatcher for MultiThreadedDispatcher {
         self.allocation_group.clear();
         self.batch_cost = 0.0;
         self.task_idx = 0;
-        self.flushed = false;
         self.layer_depth = 0;
         self.task_sender = None;
         self.recorded_command_receiver = None;


sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
yezhizhen, line 590: We added this in #1732 so we no longer need L599.

LaurenzV (approved):

sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs
LaurenzV: Yeah, it's also set to `false` instead of `true`.

LaurenzV:
Sorry, but I just realized that this is wrong I think. Later on in `reset`, we reinitialize the workers, which means that the render context becomes unflushed again. Or am I missing something? I missed this in the review because I didn't check the later body of `reset`. 😦

yezhizhen:
But we only need to set  `flushed` to `false` when we register task, which is already the case?

https://github.com/linebender/vello/blob/b377de1be0f93ba2d1c651e3d654b66f1107a720/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L237-L238

LaurenzV:
Yeah but consider something like this:

```rust
    fn multithreaded_render_empty_frame_after_reset() {
        use crate::RenderSettings;

        let mut ctx = RenderContext::new_with(
            100,
            100,
            RenderSettings {
                num_threads: 4,
                ..Default::default()
            },
        );
        let mut resources = Resources::new();
        let mut pixmap = Pixmap::new(100, 100);

        ctx.fill_rect(&Rect::new(0.0, 0.0, 100.0, 100.0));
        ctx.flush();
        ctx.render(&mut pixmap, &mut resources);

        ctx.reset();
        ctx.flush();
        ctx.render(&mut pixmap, &mut resources);
    }
```

After the first reset, we are setting flushed to true but we are still initializing the workers. If we don't ever submit any render tasks afterwards, flushed will stay true and any subsequent `flush` call therefore early aborts, so we never properly terminate before rendering and get a crash.

yezhizhen:
Makes sense. Sorry for the trouble :(

LaurenzV:
All good, no worries, I missed it myself. Was the original change motivated by some issue you encountered or just a performance optimization?

yezhizhen:
> Makes sense. Sorry for the trouble :(

Yeah your example looks like a deadlock. The worker is blocked to either receive new task or dropped sender, while `render` try to install new task. I wasn't familiar with lifecycle here :(

>  Was the original change motivated by some issue you encountered or just a performance optimization?

Just a performance optimization, rest all good.

LaurenzV:
Another option might be to just lazily initialize the workers when a new command is submitted. Will think about it.

PR #1750, opened 2026-07-16
vello_hybrid: Add `Resources::new_with_config`

grebmeg:
### Summary
The image atlas is currently configured in two independent places: the renderer's settings (which size the GPU atlas texture) and the `ImageCache` owned by `Resources` (which is what actually allocates images and reports errors). When these two disagree, the `AtlasError::TextureTooLarge` error is never surfaced — the too-large check runs against the wrong config, so uploading an image that exceeds the intended atlas size silently proceeds instead of failing.

This PR adds a public `Resources::new_with_config(AtlasConfig)` so callers can configure the atlas that Resources owns, letting both configs be kept in sync. It also adds a browser WebGL test that sets up a small (10×10) atlas and asserts that uploading a larger image fails with `TextureTooLarge`.

This is a band-aid: it makes the error reachable, but the caller is still responsible for passing matching configs in two places. A proper fix should unify this, likely a single source of truth via a unified resource-management approach, and is left for follow-up.


sparse_strips/vello_hybrid/src/resources.rs
```diff
@@ -23,8 +23,17 @@ pub struct Resources {
 impl Resources {
     /// Create a new set of renderer resources.
     pub fn new() -> Self {
+        Self::new_with_config(AtlasConfig::default())
+    }
+
+    /// Create a new set of renderer resources with a custom image/glyph atlas configuration.
```
LaurenzV, line 29: Let's please add the TODO here that we need to fix this, so we don't forget about it. 

Also, we later on normalize the config in the renderer against the device limits, so it's possible they still don't match, but yes let's leave that for a follow-up.

sparse_strips/vello_hybrid/examples/native_webgl/tests/small_atlas_webgl.rs
```diff
@@ -0,0 +1,92 @@
+// Copyright 2026 the Vello Authors
```
LaurenzV: I'm not sure this test will run in CI if we put it here. Can we not put it into `vello_sparse_tests`, still as a custom module?

Also, let's add a TODO that `upload_image` doesn't unwrap in the first place. Since this is what should fail in the first place, can we not reduce the test to just check that `upload_image` doesn't work? Like, there is no point in having any of the stuff after this for that specific test (i.e. drawing it into the scene), no?

LaurenzV (commented):
Yeah, we definitely need to fix this, for now let's just go with this I guess!

LaurenzV (approved):

sparse_strips/vello_hybrid/src/resources.rs
grebmeg: Sure, done.

sparse_strips/vello_hybrid/examples/native_webgl/tests/small_atlas_webgl.rs
grebmeg: Moved this into `vello_sparse_tests` as a new webgl-only `image_atlas` module. Reduced it to just asserting `upload_image` fails with `TextureTooLarge`. Also added a TODO at the `allocate(...).unwrap()`. (FWIW the old `native_webgl/tests` location did run in CI via the `native-webgl` `wasm-pack` job, but agreed it's cleaner in the dedicated test crate.)

grebmeg (commented):
Thanks for the review and the helpful feedback! Yes, this is a temporary fix — we'll need to come back and implement a better long-term solution.

PR #1751, opened 2026-07-16
sparse_strips: Introduce `Padding` type and streamline clip handling

LaurenzV:
This is part 4 in the prep commit series, based on top of #1746.

The first commit introduces a new `PaddingU16` type. On current main, we use kurbo `Rect` to represent the padding that is derived from  a filter. For example, a Rect of (-2, -2, 2, 2,) represents an outward padding of 2 into each direction. This makes somewhat sense, however, we use that same idea to represent a padding using `RectU16`. I think in the latter case, it doesn't make that much sense anymore. A RectU16 with dimensions (2, 2, 2, 2) represents the same idea as for `Rect`. But in reality, this is simply a zero-sized rectangle. It doesn't make a lot of sense to use `RectU16` to represent the idea of a filter padding. Therefore, this commit adds an explicit padding type so that the semantics are clearer. Methods like `intersect` and `union` that exist on `RectU16` don't really make much sense if the rectangle represents a padding, like it currently can. 

I think using `Rect` in the first place is also questionable, but at least left/top padding is represented as negative coordinates, so it makes a bit more sense. This change is necessary because the follow-up commit changes how zero-sized rectangles are handled, but for example for offset filters, it would be valid to have a padding in only one direction, while all other sides are 0.

The second commit introduces some changes to how we handle layer clips. To put it briefly, due to the fact that strips always have a fixed height of `Tile::HEIGHT` and a width with a multiple of `Tile::WIDTH`, we often need to ensure that bounding boxes are snapped to tile coordinates. Previously, this was done in various places at different times, so very inconsistently. For Vello CPU, we basically only needed this for filter layers, but for Vello Hybrid we will need to make sure that this is also the case for normal layers.

- First, we instead of computing the bbox of clips based on the control points, we simply compute it based on the strips instead. Admittedly, this is probably going to be more expensive for rectangular clips, but on the other hand it's good if we can unify this. But if desired, I can change this back and just snap the control point bbox to tile coordinates again.
- `RectU16::INVERTED` is now only used as an internal accumulator to represent "no bounding box yet". However, when returning it to the user we always return an empty rectangle instead. This way, we don't have to special case this everywhere.
- The definition of `RectU16::intersect` is changed such that it cannot produce negative rectangles anymore, also aligning with the kurbo implementation.
- `Drawable::bbox` must return tile-aligned bounding boxes.
- Some other small fixes, for example to ensure that `snap_to_tile_coorindates` preserved empty bboxes.

As a consequence, vello_cpu (and in the future vello_hybrid) do not need to do any snapping anymore. All of this is taken care of at the recorder level. There's some more optimizations we should do in the future to prevent zero-sized nodes from being emitted in the first place, but this change by itself is already large enough that I'm going to leave this as a TODO!

I've tested the combination of all 4 PRs against my PDF suite, and didn't see any regressions!

### PR train

2. #1745 — Add more new/refactored code into `vello_common`
3. #1746 — Generalize the command recorder abstraction and add more metadata
4. #1751 — Introduce `Padding` and streamline clip handling 👈

### DIFF
--- a/sparse_strips/vello_common/src/filter/mod.rs
+++ b/sparse_strips/vello_common/src/filter/mod.rs
@@ -12,8 +12,9 @@ use crate::filter::flood::Flood;
 use crate::filter::gaussian_blur::{GaussianBlur, transform_blur_params};
 use crate::filter::offset::Offset;
 use crate::filter_effects::{Filter, FilterPrimitive};
-use crate::geometry::RectU16;
+use crate::geometry::{PaddingU16, RectU16};
 use crate::kurbo::{Affine, Rect, Vec2};
+use crate::tile::Tile;
 use crate::util::RectExt;
 
 pub mod drop_shadow;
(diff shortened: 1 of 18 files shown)


sparse_strips/vello_common/src/filter/mod.rs
```diff
@@ -169,35 +176,38 @@ pub struct FilterData {
...
             // used to shift the contents when rendering into the render context, and the
             // final pixmap bbox (which is derived from `filter_expansion` will be snapped
             // separately. However, not snapping here causes larger mismatches with Vello Hybrid
             // since the size of the final pixmap determines in which way we decimate for the
             // gaussian blur filter. Therefore, we keep this for compatibility.
-            let expansion = expansion.snap_to_tile_coordinates();
+            fn snap_up(value: f64, step: u16) -> u16 {
+                let step = f64::from(step);
+                ((value / step).ceil() * step) as u16
+            }
```
grebmeg: Could we reuse the existing `snap_up` helper from `util.rs` here instead of duplicating the rounding logic? Perhaps it could be moved to `math.rs` and made `pub(crate)` if `util` isn’t the right home.

sparse_strips/vello_common/src/filter/mod.rs
```diff
@@ -169,35 +176,38 @@ pub struct FilterData {
...
```
grebmeg: Should we explicitly verify the input invariant here? Something like:

```rust
debug_assert!(
    expansion.x0 <= 0.0
        && expansion.y0 <= 0.0
        && expansion.x1 >= 0.0
        && expansion.y1 >= 0.0,
    "filter expansion must contain the origin"
);
```

sparse_strips/vello_cpu/src/coarse/bucketer.rs
```diff
@@ -214,7 +214,13 @@ impl CommandBucketer {
...
 
     fn bbox_span(bbox: RectU16) -> Span {
-        Span::new(bbox.x0, bbox.x1.saturating_sub(bbox.x0))
+        // Bbox might be empty vertically but not horizontally. In this case,
+        // it should still be considered a zero-sized span, though.
+        if bbox.is_empty() {
+            Span::new(bbox.x0, 0)
+        } else {
+            Span::new(bbox.x0, bbox.x1 - bbox.x0)
+        }
```
grebmeg, line 224: Could we avoid creating spans for empty bounding boxes, maybe by culling the empty layer earlier?

grebmeg (approved):
LGTM

sparse_strips/vello_common/src/filter/mod.rs
LaurenzV: Done.

sparse_strips/vello_common/src/filter/mod.rs
LaurenzV: Done.

sparse_strips/vello_cpu/src/coarse/bucketer.rs
LaurenzV: Yes, the plan is to remove such layers in the recording stage. But this is for a future PR, so for now I'll just add a TODO.

PR #1757, opened 2026-07-20
feat: lazily allocate image atlas textures

grebmeg:
Make image atlas allocation lazy by default. WebGL and WGPU now use tiny placeholder textures until the first upload, then promote and grow them while preserving existing layers and WGPU GLES compatibility.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/probe.rs
+++ b/sparse_strips/vello_hybrid/src/render/probe.rs
@@ -123,9 +123,19 @@ impl WebGlRenderer {
         let mut probe_image_cache = ImageCache::new_with_config(atlas_config);
         let mut probe_atlas_texture_array =
             create_atlas_texture_array(&self.gl, atlas_width, atlas_height, 1);
+        let mut probe_atlas_size = (atlas_width, atlas_height);
+        let mut probe_atlas_layer_count = 1;
         core::mem::swap(
             &mut self.programs.resources.atlas_texture_array,
             &mut probe_atlas_texture_array,
+        );
+        core::mem::swap(
+            &mut self.programs.resources.atlas_size,
+            &mut probe_atlas_size,
+        );
+        core::mem::swap(
+            &mut self.programs.resources.atlas_layer_count,
+            &mut probe_atlas_layer_count,
         );
 
         let probe_image = Arc::new(vello_common::probe::probe_image_pixmap());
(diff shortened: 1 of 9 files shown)


sparse_strips/vello_hybrid/src/render/probe.rs
```diff
@@ -123,10 +123,20 @@ impl WebGlRenderer {
         let mut probe_image_cache = ImageCache::new_with_config(atlas_config);
         let mut probe_atlas_texture_array =
             create_atlas_texture_array(&self.gl, atlas_width, atlas_height, 1);
+        let mut probe_atlas_size = (atlas_width, atlas_height);
+        let mut probe_atlas_layer_count = 1;
         core::mem::swap(
             &mut self.programs.resources.atlas_texture_array,
             &mut probe_atlas_texture_array,
         );
+        core::mem::swap(
```
LaurenzV, line 132: Uff... 😓 It's fine for now, but we definitely need to figure something out to make the probe independent from the main renderer, soon. Otherwise, it's only a matter of time that we forget to swap something and everything breaks.

sparse_strips/vello_hybrid/src/render/wgpu.rs
```diff
@@ -182,30 +182,15 @@ impl Renderer {
...
-        let min_initial_atlas_count = 2;
-        #[cfg(not(target_arch = "wasm32"))]
-        let min_initial_atlas_count = 1;
         let max_texture_array_layers = device.limits().max_texture_array_layers;
         normalize_atlas_config(
             &mut settings.image_atlas_config,
             max_texture_dimension_2d,
             max_texture_array_layers,
-            min_initial_atlas_count,
+            0,
```
LaurenzV: I think all invocations of this function just pass 0 now? If so, we might as well just omit the parameter completely.

LaurenzV (approved):
LGTM for now, but let's prioritize refactoring the WebGL and WGPU backends as soon as we can. 🙏 I feel like they are pretty fragile right now.

sparse_strips/vello_hybrid/src/render/probe.rs
grebmeg: Agreed. I’ll keep this PR scoped to lazy atlas allocation and let's handle isolating the probe resources in follow-up work.

sparse_strips/vello_hybrid/src/render/wgpu.rs
grebmeg: Good catch! Removed.

PR #1758, opened 2026-07-21
fix: share renderer resources across example scenes

grebmeg:
Previous buggy behavior:

https://github.com/user-attachments/assets/29c527da-23a6-451f-bf19-b8676aef2d5d

Share one resource set per renderer/device across example scenes, preventing independent allocators from aliasing the same GPU atlas regions. Images are uploaded once, and WebGL examples synchronize the current scene with `?page=N`.


LaurenzV (approved):

LaurenzV:
Sorry for the delay!

PR #1759, opened 2026-07-21
vello_hybrid: A complete rewrite of the core logic

LaurenzV:
Also, I know the number of added/removed lines looks scary, but a significant part of this is just tests and also shuffling existing code around. 🙂 But obviously it's still a big PR! Initially, the plan was to try to split it up, but after discussing with Alex we've come to the conclusion that it makes more sense to deliver it as a whole "package".

# Summary
This PR implements a major rewrite of Vello Hybrid. The WebGL and WGPU backends as well as some core aspects of old Vello Hybrid (like how paints and filter are encoded remain mostly the same), but the core algorithm for interpreting and scheduling scene commands has been completely reworked. More information on what motivates this change can be found in an adjacent [Zulip thread](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/Removing.20coarse-rasterization.20based.20clipping/with/590142140) as well as in some render office hours transcripts, but to summarize it very briefly:

Vello Hybrid is currently based on a so-called “coarse rasterization” approach, where render work is split up into wide tiles that can be rendered independently. One of the core motivations for doing so was to enable doing blending with bounded memory. However, as Vello Hybrid evolved, it turned out that the approach is not great, for two reasons. First, the CPU overhead for doing coarse rasterization is so big that it basically diminishes any of the performance characteristics we would expect from a GPU-based renderer. Secondly, coarse rasterization is fundamentally incompatible with filters, where the whole layer needs to be rendered on its own first. This goes against the core idea of corse rasterization, where the viewport is split up and it is assumed that each wide tile can be rendered independently. 

Because of the above,
1. Vello Hybrid now basically has two separate paths, the old coarse-rasterization based path and a new “fast strips path”, which attempts to regain the performance benefits for simple scenes without layers.
 2.  In order to support filters, we had to resort to _a lot_ of hacks to make them work with coarse rasterization, especially to support them in combination with layer clipping. The code is incredibly complex and hard to reason about, making it hard to maintain it. There are also a few open issues with bugs. In addition, the current approach also makes it impossible to support filter layers that extend over the root viewport; currently, any filter effect that isn't in the main viewport area is cut off.

Vello CPU was also plagued by similar problems, which we’ve fixed by rewriting it to use a simpler coarse rasterization approach (see #1701), in my opinion with great success! 

This PR does the same, but for Vello Hybrid. Unlike Vello CPU, which does still need some form of coarse rasterization to work efficiently, for Vello Hybrid we do not need this at all. Therefore, it's completely removed. Instead, we just submit the "soup of strips" directly to the GPU and let it take care of the rest. From my testing, not only does it solve our existing problems, but it also greatly improves performance of layers, especially on low-tier devices, and makes the overall architecture of Vello Hybrid more consistent, better tested, much easier to reason about and more memory-efficient for scenes with many filters.

# Showcases
I will try to add some showcases tomorrow.

# Changes

I will not dive into the real "meat" of this PR too much here, because I've already extensively documented how it works in the code. But I will still give a list of the overall changes, also give a suggested order for reviewing this PR.

## vello_common

Thanks to the preparatory PRs that have already been merged, the only changes are updating some references to files in comments.

## vello_sparse_shaders

I've renamed the shader files to make their naming a bit more consistent. The blending logic has been extracted into a separate shader since it's a completely independent stage now. I've also added a new `copy` shader and repurposed the existing `clear_slots` shader. Other than that, I also did some renaming of variables in the existing shaders to make them consistent. Apart from that, everythin should be the same.

## vello_sparse_tests

With this PR, we can finally get rid of the scene constraints, as we can completely transparently handle scenes where the root is a blend target, all while still retaining the best speed benefits for both cases! Apart from that, I've also added a bunch of new test cases to get more coverage for the new code.

## vello_hybrid
Staring with the easier part, I've deleted the old `schedule` module, since it's replaced with a new algorithm. The fast rect path has been extracted into a new `rect` module and paint-handling into `paint`. 

The new modules `blend` and `copy` are for GPU-side preparation of those stages. I've also made changes to `filter`, but the core idea of how filters are handled is the same, except that it's completely integrated into the new layer-scheduling mechanism. As a consequence, we do not pollute the image atlas for filters anymore! `scene` has been rewritten to use the new recording mechanism, and the new `target` module provides. In `util`, I've added a new abstraction to allow viewing non-contiguous ranges of slices in a buffer as a single contiguous ones.

Above are the easy changes, but from then on things get more tricky, and I'm unsure what the best way to review them is. 😄 I would probably start by reading the doc comments of all modules in `schedule` to get the basic idea behind the new scheduling algorithm. Then, it probably makes sense to go bottom-up, by reviewing independent modules like `draw` (which should be familiar to already-existing code), `target`,  followed by `allocate`, `execute` and `cursor`. Those basically form the building blocks of the new scheduling algorithm and can kind of be understood independently. After that, I'd recommend reading through `round`, which forms our representation of an independent set of render passes executed as a unit. And finally, the biggest part is in `schedule/mod.rs`, which ties all of the other elements together to implement what we need.

# Follow-ups

- With this PR merged, coarse rasterization in `vello_common` is essentially dead code. However, I will wait until this PR is merged before actually deleting the module, as there are some other places that still use `WideTile::WIDTH`, and I don’t want to make the diff larger than it already is!

This PR was done with assistance of GPT 5.6 Sol, but it goes without saying that I’ve heavily guided and thoroughly reviewed this implementation, to make sure it’s robust and lives up to our standards. It’s possible that a comment might have gotten lost here and there as part of moving code around, but I tried my best to find and fix those before opening this PR. I’ve also gone through a lot of iterations to arrive at this point, so it’s possible there is some accidental or unnecessary complexity somewhere that I missed.


sparse_strips/vello_sparse_tests/tests/compose.rs
```diff
@@ -389,12 +413,12 @@ fn compose_non_isolated_xor(ctx: &mut impl Renderer) {
     compose_non_isolated(ctx, Compose::Xor);
 }
 
-#[vello_test]
+#[vello_test(skip_hybrid)]
```
LaurenzV, line 416: Note that these were also disabled before, but with a wildard in `vello_dev_macros`. Now, we do support some non-isolated compose methods, but destructive ones are still not supported, hence why we need to ignore some.

sparse_strips/vello_sparse_shaders/shaders/filters.wgsl
```diff
@@ -304,55 +308,52 @@ const VERTICAL: vec2<f32> = vec2<f32>(0.0, 1.0);
...
-    @location(2) @interpolate(flat) src_size: vec2<u32>,
-    @location(3) @interpolate(flat) dest_offset: vec2<u32>,
+    @location(0) @interpolate(flat) filter_data_offset: u32,
+    @location(1) @interpolate(flat) source_origin: vec2<u32>,
+    @location(2) @interpolate(flat) source_size: vec2<u32>,
+    @location(3) @interpolate(flat) dest_origin: vec2<u32>,
     @location(4) @interpolate(flat) dest_size: vec2<u32>,
-    @location(5) @interpolate(flat) dest_atlas_size: vec2<u32>,
-    @location(6) @interpolate(flat) original_offset: vec2<u32>,
-    @location(7) @interpolate(flat) original_size: vec2<u32>,
```
LaurenzV, line 314: `dest_atlas_size` and `original_size` were actually unused here.

sparse_strips/vello_sparse_tests/tests/compose.rs
grebmeg: Thanks for the comment! How difficult do you think it would be to support this in the future?

sparse_strips/vello_hybrid/src/paint.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/// Source used to populate a strip's paint payload.
+#[derive(Clone, Copy)]
+enum PaintPayload {
+    /// Premultiplied RGBA value for a solid paint.
+    Solid(u32),
+    /// Scene-space position used to evaluate a non-solid paint.
+    Position,
+}
```
grebmeg: Could `PackedPaint` and `PaintPayload` be moved to the top?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1066,6 +1089,55 @@ impl WebGlPrograms {
...
+
+        for (index, textures) in self.resources.layer_textures.iter_mut().enumerate() {
+            let required_page_count = layer_pages[index];
+
+            // Note: Currently, `texture_size` only grows across frames, so if this condition
+            // is true it means that the new required size is larger than what we currently have.
+            if current_size != texture_size {
+                // TODO: Drop old layer explicitly before creating new one to ensure WebGL
+                // driver has the chance to reduce peak memory usage instead of having to keep both
+                // textures alive at the same time?
```
LaurenzV, line 1127: Should I do this in this PR? I'm not sure if it's gonna help in practice, but it shouldn't hurt either? My idea would have been to maybe have a small wrapper that uses an Option inside, and then take out the old object before inserting the new one.

LaurenzV:
@grebmeg I think from my side I'm done now. I'm not super happy with adding more complexity to `webgl`  and `wgpu`, but no point in trying to refactor this in this PR. 🙂 

Apart from that, the main two other changes were:
- Encoded paints don't need to be in a RefCell anymore, because we don't mutate them during rendering.
- The other bigger change is that the texture limit is now not enforced during scheduling anymore, but before we start rendering. This means that a schedule might happily allocate more textures (which is cheap, it's just adding another guillotierre atlas), but we abort rendering if it would require us to allocate more textures than permitted. This change was made because the scheduler can't/shouldn't really know about what textures are currently allocated in the backend (and how they are distributed), so it's better to move that responsibility to the backend.

AdrianEddy:
I threw this PR into Fable 5 review, maybe it's something useful:

## 🔴 Confirmed bugs

### Bug 1 — Blended child extending outside a clipped parent: panic or silent memory corruption

`compose_layer` (`schedule/mod.rs`) computes `blend_bbox` from the child's bbox (non-destructive)
or the parent's bbox (destructive), intersecting only with the **child's** clip — never with
`parent_region.layer_bbox`. The recorder breaks the child ⊆ parent containment invariant when the
parent has a clip path: it unions the child bbox into the parent *before* shrinking the parent to
its clip (`record.rs`, `pop_layer`). `LayerTextureRegion::texture_rect` (`target.rs:260`) then
hits its documented precondition violation:

- **Left/top overhang** → `checked_sub().unwrap()` **panics** — debug *and* release, both backends.
- **Right/bottom overhang** → no panic; the blend + scratch copy-back **silently write outside the
  parent's allocation** into neighboring live atlas regions (visual corruption).

Repro (32×8 scene):

```rust
push_layer(clip = [24,0,32,8]) {
    fill([24,0,32,8]);
    push_layer(Mix::Multiply / Compose::SrcOver) {
        fill([0,0,16,8]);
    }
}
```

A second variant: an empty *destructive* child inside a filter parent with non-zero source shift
(child bbox = `dest_bbox` ⊄ parent `pixmap_bbox`). Existing tests miss this because
`clipped_away_blend` uses a fully disjoint clip, which early-outs on the empty intersection.

**Suggested fix:** in `compose_layer`, intersect `blend_bbox` *and* the child sample window with
`parent_region.layer_bbox` before emitting the `BlendOp` (release the child and skip if empty) —
mirroring what `push_layer_fill`/`push_rect` already do on the draw path.

### Bug 2 — `Cursor::release` assert fires on a legal scene under atlas pressure

The empty-`blend_bbox` path releases the child at the **stale** `child_layer.ready.round`.
Between the child being scheduled and the compose, the parent's lazy `ensure_layer_target` can
advance the cursor (draining pending releases for space), after which `Cursor::release` panics:
*"cannot release an allocation in a round already passed by the cursor"* (a real `assert!`,
active in release builds).

Worked repro: page size = `8 + 2×FILTER_ATLAS_PADDING`; parent P (odd parity) with an 8×8 draw;
child C = **filter** layer (even, offset filter) with a Multiply blend and a clip disjoint from
its content. Filter layers skip the bbox ∩ clip cull in `prepare_node`, so C schedules fully; its
full-page odd temporary forces the parent's allocation to drain releases (cursor → round 1); the
compose then finds an empty blend bbox and releases C at round 0.

**Fix constraint:** naively clamping the release round to `cursor.current_round()` trades this
panic for an execute-time `unwrap` panic in `clear_passes` — the clear must land in a round that
*binds* the released page. The fix needs to route the release through a binding-compatible
schedule point. A fix for Bug 1 (intersect + release-if-empty) flows through this same path and
must respect the same constraint.

---

## 🟡 Performance items

1. **Per-draw non-default blend is O(N) render-pass sets.** Each draw under
   `set_blend_mode(non-default)` becomes an isolated layer, and same-parent blends serialize
   strictly one per round (`SchedulePoint::after` is strict), so N Multiply draws ⇒ ≥N rounds of
   blend + copy + clear, plus `root_is_blend_target` forcing the whole root through an
   intermediate texture. Caveat: the old hybrid never executed draw-level blends at all — it
   silently rendered them as SrcOver (the blend mode died in `FillAttrs`; the old
   `COLOR_SOURCE_BLEND` path only served *layer* blends). So this is the cost of new correctness,
   not a regression — but the cliff deserves documentation and eventually a dst-read fast path.
   Per-draw isolation itself is correct-by-necessity: merging adjacent blended draws would be
   semantically wrong where they overlap.
2. **wgpu backend creates fresh GPU buffers every pass** — strips per draw pass (in-code TODO),
   plus untracked `create_buffer_init` per blend/copy/clear/filter pass. The WebGL backend reuses
   buffer objects; wgpu should use a staging belt or per-frame arena.
3. **Cursor drains releases parity-blind** (acknowledged TODO; `page_growth` shows it): an
   unsatisfiable request advances rounds even when all pending releases are the wrong parity, then
   creates a page anyway. Empty rounds are free at execute time, but the monotone cursor means
   every later allocation serializes after the drained frontier — lost batching proportional to
   spill events. The parity-aware check is cheap and also shrinks (not closes) the Bug 2 window.
4. **Depth-texture churn:** `maybe_update_config_buffer` recreates the depth texture whenever
   `render_size` changes, and `render_to_atlas` flips render_size to atlas dims and back — any
   frame with glyph-atlas work allocates **two depth textures**, one of them (atlas-sized) never
   used since atlas roots run depthless.
5. **Opaque external-texture strips drop their texture runs** (conditional): opaque-routed strips
   record no `ExternalTextureRun`, and `opaque_pass` binds no external texture — an
   `ExternalTexture` paint with `may_have_transparency == false` on the root would sample the
   wrong binding. Whether the public API can currently produce that combination needs a check;
   if yes, it's a wrong-texture render.

## 🟠 API / cross-renderer semantics

6. **`set_blend_mode` destructive panic diverges from `vello_cpu`:** the PR panics on destructive
   composes; current CPU is a plain setter and renders them. The old hybrid accepted and silently
   mis-rendered. Worth aligning (panic in both, or error in both) — this is a new runtime panic
   on previously-accepted input.
7. **Filter + non-default blend on the same draw renders differently in CPU vs hybrid:** CPU
   pushes a default-blend filter layer (the blend degenerates against the transparent layer
   backdrop); hybrid pushes one layer carrying both (blends the *filtered* result against the
   parent). Hybrid's semantics are arguably the intended ones — but the renderers now disagree.

## ⚪ Minor

- `MAX_ROUNDS = 40_000` deadlock guard panics (in-code TODO to make it an error).
- `schedule_layer` recursion can overflow the stack on pathologically deep layer trees (TODO;
  vello_cpu flattened theirs).
- `render_to_atlas`'s encoder-ordering contract is doc-only — calling it between recording and
  submitting a `render()` encoder would leave the shared config buffer holding atlas dims for the
  surface submission (latent; internal call sites are sequenced correctly).
- Paint offsets silently truncate at 2²⁶ texels (`gpu_offset & PAINT_TEXTURE_INDEX_MASK`).
- Typos: "We could an optimization here that early exits early" (`schedule/mod.rs`), "since we now
  that advancing" → *know* (`cursor.rs`), "texthre page 2" (test comment); `}   else {` formatting
  in `blend.wgsl`; `LayerPass::texture_parity` dead-code allow.

## ✅ Verified safe (concerns checked and refuted)

- **`payload_at` underflow/panic** — refuted: `visit_strip_fill_segments` clamps every emitted
  segment exactly to tile bounds, all bboxes reaching `to_tile_bounds` are tile-aligned on the PR
  head, and `RectU16::shift` clamps rather than wraps.
- **wgpu bind-group/config staleness across resizes and render/render_to_atlas alternation** —
  refuted: `view_config_buffer` is only ever written in place; every cache embedding a recreated
  resource is invalidated at the recreation site; atlas root draws are double-gated onto the
  Rgba8 pipeline.
- **Blend shader** — byte-identical math to the canonical `vello_shaders` blend;
  opacity-on-premultiplied-before-compose matches `vello_cpu` exactly; the clip-coverage lerp
  `mix(backdrop, composed, cov)` matches CPU's `compose_inner` verbatim and is a **correctness
  improvement** over the old GPU path (which coverage-scaled the source — wrong for destructive
  composes).
- **Filter pipeline** — round-binding uniqueness, ping-pong parity across batched multi-filter
  steps, drop-shadow scratch preservation, and the `min(original, dest + 6)` vertex-quad
  replacement for the old scissor are all exactly right; the 6 px padding bound is tight (worst
  blur read = dest + 5) but provably sufficient, and the padding-ring transparency invariant holds
  inductively across frames.
- **Scheduler invariants** — full stage-lattice check passed (incl. even-parent/odd-child,
  two children on different opposite-parity pages, root draws after same-round blends); the
  global `opaque_strips.reverse()` + single early depth pass is correct for all
  opaque/alpha/layer-fill interleavings; every clear lands in a round that binds the cleared page.
- **Recording** — `push_layer` ordering identical to CPU; per-draw bbox is lazy (zero cost for
  root draws); the rect fast path *is* clip-gated (a diff-hunk artifact hid the guard) and the PR
  additionally fixes an old aliasing-threshold hole in that gating.

LaurenzV:
Thank you, I'll take a look at the two bugs tomorrow.

The other stuff looks mostly expected for now, but I'll check again tomorrow and I might add some more TODOs so they are properly tracked. The external texture one is wrong _I think_, because currently external textures can never end up in the opaque pass, but I've definitely not considered this, so will also add a note to the existing TODO so we don't forget about it!

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+//! use the even group and odd-depth layers use the odd group. Recorded layer depths start at one,
+//! so a simple chain labeled from `L0` can render by ping-ponging between one page from each group:
+//!
+//! ```text
+//! L0 (odd)
+//! `-- L1 (even)
+//!     `-- L2 (odd)
+//!         `-- L3 (even)
+//!             `-- L4 (odd)
+//! ```
```
grebmeg, line 40:
```suggestion
Root surface (not an intermediate layer)
└── L0 — depth 1, odd page  — rendered 5th
    └── L1 — depth 2, even page — rendered 4th
        └── L2 — depth 3, odd page  — rendered 3rd
            └── L3 — depth 4, even page — rendered 2nd
                └── L4 — depth 5, odd page  — rendered 1st
```

+ below example too

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+//!
+//! After `L1` is composed, `L0` has an odd allocation containing that result. This allocation must
+//! remain live while `L2` is scheduled. Descending into `L2` then reaches `L3`, which also needs an
+//! odd allocation. `L0` and `L3` may share an odd atlas page if both regions fit; otherwise `L3`
+//! requires a second odd page. Together with the even page used by `L2`, this means the branch can
+//! require three intermediate textures. Lazy allocation minimizes the overlap, but cannot remove
+//! allocations whose contents are simultaneously live. However, it should become apparent that,
+//! even for complexly nested layer graphs, the number of allocations kept alive at the same time
+//! is basically as small as possible and never exceeds the maximum layer depth across the whole
+//! scene graph.
```
grebmeg, line 77: Did you consider rendering the globally deepest subtree first, while still preserving painter’s order for composition? That could improve pass batching in some cases, but later sibling results would need to stay live longer and might increase texture usage. Was this trade-off explored, or is minimizing allocation lifetimes the reason for the ordered depth-first traversal?

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+//! round, we **never** attempt to backtrack to find a spot for a layer allocation in a previous
+//! round, in an attempt to minimize the number of render passes. This is accepted as a downside of
+//! the current algorithm in the interest of keeping the already non-trivial logic as simple as
+//! possible. The concrete contents and texture bindings of a round are described in the [`round`]
+//! module.
+//!
+//! For example, consider a single 30x30 texture page. We first allocate a 5x5 region in round 0,
+//! followed by a 30x30 region that must wait until the first allocation is released and therefore
+//! advances the cursor to round 1. If we then request another 5x5 region, the scheduler does not
+//! reconsider round 0, even if that region could have fit beside the first 5x5 allocation.
```
grebmeg, line 99: Do you think it would be worth exploring backtracking as a future optimization? Reconsidering earlier rounds could reduce render-pass count in cases like this, although I understand it would add significant scheduling complexity. I

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+        let child_region = child_layer.sample_region;
+
+        // For non-destructive blend modes, choose the (smaller) child bbox as the
+        // affected region. Otherwise, we need to choose the (larger) parent bbox, since
+        // any pixel that doesn't lie in the child needs to be cleared.
+        let mut blend_bbox = if blend_mode.is_destructive() {
+            parent_region.layer_bbox
+        } else {
+            child_region.layer_bbox
+        };
```
grebmeg, line 611: I think what @AdrianEddy found in "bug 1" is that we clip the child’s sample region to the parent before constructing the `BlendOp`? If the parent clip makes `parent_region` smaller than `child_region`, `GpuBlendInstance::new` can panic when translating the child via `parent_region.texture_rect(...)`. Maybe the test below would help reproduce the issue:

```rust
#[test]
fn blend_region_stays_within_clipped_parent() {
    let mut case = SceneCase::new(32, 8);
    case.layer_with(
        Some(Rect::new(12.0, 0.0, 20.0, 8.0)),
        None,
        None,
        |case| {
            case.layer_with(
                None,
                Some(BlendMode::new(Mix::Multiply, Compose::SrcOver)),
                None,
                |case| case.draw(Rect::new(4.0, 0.0, 16.0, 8.0), 0.5),
            );
        },
    );
    let scheduled = case.schedule_root();
    let blend = scheduled.storage.buffers.blend_ops.first().unwrap();
    assert_eq!(blend.blend_bbox, blend.parent_region.layer_bbox);
    // This must not panic.
    let _ = crate::blend::GpuBlendInstance::new(
        blend,
        None,
        SizeU16::new(64),
    );
}
```

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+        &mut self,
+        mut layer: OpenLayer<'a>,
+        rounds: &mut Rounds,
+    ) -> Result<ScheduledLayer, RenderError> {
+        // Overall we follow a similar flow to `schedule_root` here.
+
+        for cmd in layer.cmds {
+            // First make sure that the child node is scheduled, in case it exists. Unlike for root
+            // layers, we need to make sure to do this _before_ pushing any draws.
+            // TODO: Similarly to Vello CPU, flatten this to avoid stack overflows for deep layers
```
grebmeg, line 396: 💯

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+            sample_region: layer.sample_placement.resolve_sample_region(region),
+            allocation: target.allocation,
+            ready: target.schedule_state.ready,
+        };
+
+        Ok(scheduled)
+    }
+
+    /// Resolve and schedule the optional child layer referenced by a command node.
+    fn prepare_node(
```
grebmeg: Why not call it something like `schedule_child_layer`?

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+        RoundStage::draw(self.texture.target.texture_parity)
+    }
+
+    fn round_bindings(&self) -> RoundBindings {
+        RoundBindings::new(self.texture.target)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
grebmeg: What about moving this tests into a separate file?

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -0,0 +1,1831 @@
...
+            cursor: Cursor::new(Atlases::new(texture_size)),
+            texture_size,
+            storage,
+        }
+    }
+
+    /// Build the complete dependency-ordered schedule.
+    fn build(mut self) -> Result<Schedule, RenderError> {
+        let mut rounds = Rounds::default();
+        self.schedule_root(&mut rounds)?;
```
grebmeg, line 312: Could we add a debug/test-only validator for the completed schedule? The scenario tests verify particular layouts, but a validator would make every test also check the scheduler’s general structural invariants.

For example, something like this could live in `rounds.rs` (simpified example):
```rust
#[cfg(any(test, debug_assertions))]
impl Rounds {
    pub(super) fn validate(&self, buffers: &ScheduleBuffers) {
        for round in &self.rounds {
            for pass in round.layer_passes(buffers) {
                assert!(
                    round.texture_binding
                        .layer_id(pass.texture_parity)
                        .is_some(),
                    "layer pass has no bound target"
                );
            }
            for clear in round.clear_passes() {
                assert_eq!(
                    round.texture_binding
                        .layer_id(clear.target.texture_parity),
                    Some(clear.target),
                    "clear targets an unbound page"
                );
            }
        }
    }
}
```

sparse_strips/vello_hybrid/src/schedule/round.rs
```diff
@@ -0,0 +1,755 @@
...
+        *page_count = (*page_count).max(required);
+    }
+
+    pub(super) fn ensure_exists(&mut self, round_idx: usize) {
+        while self.rounds.len() <= round_idx {
+            self.rounds.push(Round::default());
+        }
+    }
+
+    pub(super) fn resolve_binding_point(
```
grebmeg: Could we benchmark this linear scan with a scene that produces hundreds of rounds and frequent texture-binding conflicts?

grebmeg: Also, "resolve … point" also permanently bumps `layer_page_counts` as a side effect. That's surprising — maybe there’s a more appropriate name?

sparse_strips/vello_hybrid/src/schedule/mod.rs
grebmeg: Or maybe it could live in test_support.rs

sparse_strips/vello_hybrid/src/draw.rs
```diff
@@ -0,0 +1,692 @@
...
+
+impl Draw {
+    #[inline]
+    fn push(
+        &mut self,
+        strips: &mut Vec<GpuStrip>,
+        gpu_strip: GpuStrip,
+        external_texture_id: Option<TextureId>,
+    ) {
+        if let Some(texture_id) = external_texture_id {
```
grebmeg, line 42: If I understand it correctly atlas-backed images provide `None` here, so they never terminate the current external-texture run. Because each run implicitly ends at the next run or at the end of the draw the sequence: `External A → Atlas → External A`
is represented as one run starting at zero. The backend therefore binds external texture A for all three strips, causing the atlas image to sample from the wrong texture. The leading case `Atlas → External A` has the same problem because the first external run is forced to start at zero.

This test reproduces it:

```rust
#[test]
fn texture_runs_preserve_atlas_transition_between_external_images() {
    let texture = TextureId(10);
    let encoded = [external(texture), atlas_image()];
    let resolver = PaintResolver::new(&encoded, &[0, 0]);
    let mut case =
        DrawCase::new(RootTarget::UserSurface, RectU16::new(0, 0, 16, 8));
    let mut draw = Draw::default();
    for (x, paint_index) in [(0.0, 0), (4.0, 1), (8.0, 0)] {
        case.rect(&mut draw, rect(x), indexed(paint_index), resolver);
    }
    assert_eq!(
        represented_texture_sources(&draw),
        [Some(texture), None, Some(texture)],
    );
}

fn atlas_image() -> EncodedPaint {
    EncodedPaint::Image(EncodedImage {
        source: ImageSource::opaque_id(ImageId::new(0)),
        sampler: ImageSampler {
            x_extend: Extend::Pad,
            y_extend: Extend::Pad,
            quality: ImageQuality::Low,
            alpha: 1.0,
        },
        may_have_transparency: true,
        transform: Affine::IDENTITY,
        x_advance: Vec2::new(1.0, 0.0),
        y_advance: Vec2::new(0.0, 1.0),
        tint: None,
    })
}
fn represented_texture_sources(draw: &Draw) -> Vec<Option<TextureId>> {
    let mut sources = vec![None; draw.strip_ranges.len()];
    for (index, run) in draw.external_texture_runs.iter().enumerate() {
        let end = draw
            .external_texture_runs
            .get(index + 1)
            .map_or(draw.strip_ranges.len(), |next| next.strips_start);
        sources[run.strips_start..end].fill(Some(run.texture_id));
    }
    sources
}

```

grebmeg (commented):
This is a MEGA rewrite! Thanks for all your hard work! I haven’t finished the full review yet, but I’m sharing a few comments for now.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -1114,46 +908,37 @@ impl Scene {
...
             self.pop_layer();
+            result
         } else {
-            f(self);
+            f(self)
         }
     }
 
     /// Reset scene to default values.
     pub fn reset(&mut self) {
```
grebmeg, line 938: Hmm, should we recompute `paint_visible` here? Resetting after transparent paint currently leaves subsequent default-painted draws invisible.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -702,34 +563,32 @@ impl Scene {
...
+            let mut strip_storage = self.strip_storage.borrow_mut();
+            let strip_start = strip_storage.strips.len();
+            self.viewport_state
+                .with_generator_and_clip(|strip_generator, clip_path| {
+                    generate(strip_generator, &mut strip_storage, clip_path);
+                });
+            strip_start..strip_storage.strips.len()
+        };
+
+        let draw = RecordedDraw::new_path(strips.clone(), paint);
```
grebmeg, line 592: Could we skip recording when generation produced no strips?

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -14,152 +14,92 @@ use core::cell::RefCell;
...
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "recorded fast rectangles are clipped to the u16 viewport"
+    )]
+    fn bbox(&self, strips: &[Strip]) -> Option<RectU16> {
+        match self {
+            // TODO: The bbox trait should take a reference to _all_ strips, and then we select
+            // the correct strips here. Otherwise, callers of this method always have to make sure
+            // to slice the strips correctly themselves.
+            Self::Path(_) => strip_bbox(strips),
```
grebmeg, line 91: Could we avoid the extra strip scan for layered paths here? For N segments, generation writes all N, `strip_bbox` reads them again, and scheduling later reads them a third time.

Maybe strip generation could accumulate and return the bbox while emitting segments, then store it on `RecordedPath`? Roughly somthing like this:

```rust
struct RecordedPath {
    strips: Range<usize>,
    bbox: Option<RectU16>,
    paint: Paint,
}
let bbox = generate(strip_generator, &mut strip_storage, clip_path);
let path = RecordedPath {
    strips: strip_start..strip_storage.strips.len(),
    bbox,
    paint,
};
```

Then `Drawable::bbox` becomes O(1):
```rust
Self::Path(path) => path.bbox,
```
WDYT?

sparse_strips/vello_sparse_tests/tests/compose.rs
LaurenzV: I'm not sure. Fundamentally, the problem is that the area needs to be limited to the actual painted area of the path. Perhaps we can simulate it with clip paths somehow.

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: I haven't thought about rendering the globally deepest subtree first, but I did think about what other options there are in general. But yes, the main problems are 1) more unpredictable texture usage and 2) overall complexity in trying to come up and implement a different algorithm. The current traversal is more straightforward and also gives us the best in terms of memory usage, which is something we care about.

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: I don't think it's worth exploring. The great thing about the current approach is that we have one global set of atlas (per texture). If we were to implement some form of backtracking, we would likely have to retain those atlases _for each already scheduled round_. And if you backtrack because you found some space in an earlier round, you would have to retrospectively reapply the atlas allocations to all other atlases between the found round and the current round. So yeah, I don't think this is something we would want. 😄 What we have right now is comparatively simple and gives good results in most cases.

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: Yup. I somehow still had the assumption in my head that children bboxes are always strictly within the parent ones, but of course that isn't the case if there is a clip path on the parent! Will fix this and add tests.

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: That's indeed a better name. 😄

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: Makes sense.

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: Sure, sounds good!

sparse_strips/vello_hybrid/src/schedule/round.rs
LaurenzV: Hmm, what exactly would you like to see in the benchmarks? It's a big vague, because you will get different results depending on how much and what exactly you allocate. 😄 It is definitely somewhat unfortunate because it does grow linearly with the number of allocated pages, but since we are scheduling on units of layers / draw batches, it shouldn't matter in practice, unless of course you purposefully construct such a scene!

LaurenzV: Regarding the rename, how about we remove `layer_page_counts` completely and just derive the required pages in the end in a linear scan over the rounds? And we could rename to `bind_at_or_after` if you think this is better.

sparse_strips/vello_hybrid/src/draw.rs
LaurenzV: If I understand your concern correctly, this is working as intended. The external texture and image atlas array are two separate bindings:

https://github.com/linebender/vello/blob/9198c8aba6ab6e07eadb327959129f9524d435c1/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl#L249-L253

Which one to sample from is decided at runtime:
https://github.com/linebender/vello/blob/9198c8aba6ab6e07eadb327959129f9524d435c1/sparse_strips/vello_sparse_shaders/shaders/render_strips.wgsl#L441-L459

Therefore, it should be fine to collapse them into a single external texture run. I can try to add an explicit test case for this?

sparse_strips/vello_hybrid/src/schedule/mod.rs
LaurenzV: Done!

sparse_strips/vello_hybrid/src/paint.rs
LaurenzV: Done

LaurenzV:
@grebmeg I think your comments are addressed now, except for the benchmark and layer page counts thing.

Regarding the Fabel findings:

> Blended children extending outside clipped parents can panic or corrupt neighboring atlas allocations.

This should be fixed now.

> Releasing a clipped-away blended child at a stale round can panic under atlas pressure.

This should also be fixed now.

> Per-draw non-default blending creates at least one serialized blend round per draw.

That's expected, there is no faster way to implement this kind of blending.

> The WGPU backend creates fresh GPU buffers for many individual passes.

That's for follow ups.

> The cursor advances through pending releases without considering texture parity.

As correctly noted, this is already documented. Might be a follow up, but it's not critical.

> Rendering glyph atlases can recreate two depth textures per frame despite atlas rendering being depthless.

I think that's already an issue on main, but obviously needs fixing.

> Opaque external-texture strips would lose their texture binding if external textures could enter the opaque pass.

This can't happen because external textures don't support opaque right now, but it's a good call out and easy to forget, so I added a TODO.

> Destructive draw-level blend modes panic in Hybrid but remain accepted by Vello CPU.

This is documented in the method signature.

> Combining a filter with a non-default draw blend produces different results in Hybrid and Vello CPU.

Added a TODO on the Vello CPU side.

> The `MAX_ROUNDS` deadlock guard panics instead of returning an error.

Also expected for now.

> Recursive layer scheduling can overflow the stack for pathologically deep layer trees.

Already reflected in a TODO.

> Correct `render_to_atlas` encoder ordering is documented but not enforced.

For now the method is not intended to be called from outside, so also out of scope for this PR.

> Paint-data offsets silently truncate when they exceed the 26-bit shader field.

That's fine.

> Several comments, shader formatting, and a dead-code allowance need minor cleanup.

Applies those changes.

Thanks again!

sparse_strips/vello_hybrid/src/schedule/round.rs
grebmeg: I meant a synthetic worst-case benchmark to check how scheduling time scales as rounds and texture conflicts increase.

Deriving the page counts from the final rounds sounds cleaner, and `bind_at_or_after` works for me.

sparse_strips/vello_hybrid/src/draw.rs
grebmeg: Thanks, you’re right. I had assumed that binding an external texture replaced the atlas binding, but both are bound simultaneously and the shader chooses the correct source for each strip. An explicit test would be helpful.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Let’s handle this as a follow-up, please?

sparse_strips/vello_hybrid/src/schedule/execute.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        // TODO: Currently, we upload data for layers when executing each round. It might be
+        // worth exploring whether we can upload the data once in the beginning, and then
+        // reference it via slices.
+
+        // We iterate over each round separately.
+        for round in self.iter() {
+            // For each round, we first draw to the even layer texture, then to the odd layer
+            // texture. The order is important because odd layers can depend on draws to the even
+            // texture in the same round.
```
grebmeg:
```suggestion
// Draw the even layer texture before the odd one. A child layer always lives in the opposite
// group from its parent, and only odd-texture draws sample the even texture (never the reverse
// within a round), so this order guarantees those pixels are ready when we read them.
```

sparse_strips/vello_hybrid/src/schedule/execute.rs
```diff
@@ -0,0 +1,378 @@
...
+        root_output_target: RootTarget,
+        buffers: &ScheduleBuffers,
+        filter_plan: &mut FilterPassPlan,
+        texture_size: SizeU16,
+    ) {
+        // This is the core loop that ties everything together.
+
+        // TODO: Currently, we upload data for layers when executing each round. It might be
+        // worth exploring whether we can upload the data once in the beginning, and then
+        // reference it via slices.
```
grebmeg, line 109: I think this is worth doing — could we do as a follow-up? The strips already live in one contiguous buffer with each pass being ranges into it (`RangedSlice`), yet every pass currently allocates a fresh GPU buffer in `upload_strip_pair` and gather-copies into it. For deeply nested scenes that's a `create_buffer` per pass per frame, and mid-frame allocation is the costly bit. Uploading `DrawBuffers::strips` once and referencing slices removes that churn.

sparse_strips/vello_hybrid/src/schedule/execute.rs
```diff
@@ -0,0 +1,378 @@
...
+use crate::{GpuStrip, blend::BlendStrip};
+use vello_common::geometry::{RectU16, SizeU16};
+
+/// A backend for executing GPU render passes.
+pub(crate) trait Backend {
+    /// Execute the opaque pass against the root target with the given strips. If this method is
+    /// ever called, it's called before any of the other ones and only once.
+    ///
+    /// The strips are guaranteed to be non-empty.
+    fn opaque_pass(&mut self, strips: &[GpuStrip]);
```
grebmeg: nit: `opaque_pass` is the only method here named after the strips it draws rather than the action it performs — the rest are verbs (`draw_pass`, `clear_pass`, `blend_pass`, `filter_pass`). It's also not really a distinct kind of pass; it's a draw, just with the root's opaque strips + the front-to-back depth optimization.

Could we rename it `opaque_draw_pass`? Keeps the `_pass` suffix, restores the verb, and makes it read as a specialized draw rather than a sibling concept to blend/filter.

grebmeg (approved):
Amazing, solid work! I really like the new scheduler, particularly the separation between scene recording, scheduling, and backend execution! The first-class handling of filters just ❤️

We’ve already discussed this, but I’d like to highlight it again — the main trade-off seems to be memory predictability. The previous system used two fixed slot textures and flushed completed rounds when slots ran out, allowing slots to be reused and keeping intermediate memory bounded. The new system creates additional intermediate layer textures as needed. Applications can limit their number using `max_textures`, but its default value is `None`, so the default configuration does not impose a count limit. If `max_textures` cannot satisfy a scene, rendering returns `AtlasError::NoSpaceAvailable` instead of retrying (or by using a specific option for memory-constrained environments) with a lower-memory strategy that subdivides or spills intermediate results. 

For the future, I’d like us to explore how this new design could split work while respecting a hard memory limit on intermediate textures, allowing it to run reliably even on “potato” devices.

LGTM if LGT @taj-p

sparse_strips/vello_hybrid/src/draw.rs
LaurenzV: Already added one in https://github.com/linebender/vello/pull/1759/commits/6e70ad37ff8dcbd6ac8193011a1d27c539edeb2b!

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Probably doable. However, I want to avoid updating `render_strips` now. I will leave this as a TODO.

sparse_strips/vello_hybrid/src/schedule/execute.rs
LaurenzV: Yes, I'll move it up my priority list. 👍

sparse_strips/vello_hybrid/src/schedule/execute.rs
LaurenzV: Yeah good find!

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Added recomputation and a test.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Added a skip + a test checking that.

PR #1760, opened 2026-07-22
vello_common: Compute the transparency hint for peniko images

tronical:
The conversion built the Pixmap via from_parts(), which conservatively marks every image as possibly transparent. As a result, images supplied as peniko::ImageData never hit the FillHint::OpaqueImage fast path in the coarse rasterizer (skipping per-pixel blending and occluding commands underneath), even when they are fully opaque.

The conversion already visits every pixel to premultiply, so track whether any alpha value is not 255 in the same loop and construct the Pixmap with from_parts_with_opacity().

This PR was done with assistance of Claude Code.


LaurenzV (approved):
Thanks!

PR #1762, opened 2026-07-23
vello_common: Remove the old coarse rasterization logic

LaurenzV:
Based on #1759. Once that is merged, we can remove the code of the previous coarse rasterization code! 🎉 Render graph is also deleted as part of this, as `CommandRecorder` (which I will probably rename in the future) is kind of our new render graph now.  Benchmarks just hardcode the same width for now, and other parts of the code that referenced coarse either were deleted or repurposed.

The only thing I'm unsure about is tests. For now, I just chose to rename them and add some notes, but I'm not sure if we should just delete them outright? They might cause confusion because they don't really test anything that is still relevant for the current architecture, but on the other hand it also doesn't hurt keeping them.


sparse_strips/vello_bench/benches/main.rs
```diff
@@ -5,12 +5,9 @@
...
 
 use criterion::{criterion_group, criterion_main};
-use vello_bench::{allocator, coarse, fine, flatten, glyph, integration, sort, strip, tile};
+use vello_bench::{allocator, fine, flatten, glyph, integration, sort, strip, tile};
 
 criterion_group!(allocator_bench, allocator::allocator);
-criterion_group!(coarse_bench, coarse::coarse);
-criterion_group!(coarse_with_layer, coarse::coarse_with_layer);
-criterion_group!(coarse_layer_4k, coarse::coarse_with_layer_large_viewport);
 criterion_group!(fine_solid, fine::fill);
```
grebmeg, line 11: Did you get a chance to benchmark the new design against the previous one?

sparse_strips/vello_example_scenes/src/lib.rs
```diff
@@ -359,8 +357,6 @@ pub struct AnyScene<T: RenderingContext> {
     key_handler_fn: KeyHandlerFn,
     /// The status query function.
     status_fn: StatusFn,
-    /// Whether to show the wide tile columns overlay.
-    show_widetile_columns: bool,
 }
```
grebmeg, line 359: Can we show horizontal strip rows here and in the toy instead of wide tiles?

grebmeg (approved):
Nice! LGTM

> The only thing I'm unsure about is tests. For now, I just chose to rename them and add some notes, but I'm not sure if we should just delete them outright? They might cause confusion because they don't really test anything that is still relevant for the current architecture, but on the other hand it also doesn't hurt keeping them.

I think it doesn’t hurt to keep them for now. Maybe we can remove them later. What about moving all the wide-tile-specific tests into a separate file, with file-level comments explaining their purpose?

sparse_strips/vello_bench/benches/main.rs
LaurenzV: You mean the new design of Vello Hybrid? Or what benchmark do you mean exactly?

As previously mentioned, filters improve dramatically. I just tried GhostScript tiger on my potato tablet and it's the same performance, which is not surprising because the fast strips path basically stayed the same.

sparse_strips/vello_example_scenes/src/lib.rs
LaurenzV: You mean the generated fill and alpha fill segments right? Sure!

LaurenzV:
I'll change it so the wide tile tests are in a custom module.

sparse_strips/vello_example_scenes/src/lib.rs
LaurenzV: Ah sorry, you mean just the grid in the example scenes and segments in vello toy probably. WIll update.

LaurenzV: Hmm, I tried this but I don't think it makes a lot of sense. This is what it look like before:

Image

And this is how it looks like now 😅 So I'll revert this for now

Image

PR #1763, opened 2026-07-23
sparse_strips: Add a "DropShadowOnly" filter variant

LaurenzV:
This is based on top of #1759. It adds a variation of drop shadow that doesn't composite the original layer on top of the shadow. CC @angusj-canva

Skia has something similar: https://api.skia.org/classSkImageFilters.html#a91354167ad35b6a0d95b6f1fc23f3aea

### DIFF
--- a/sparse_strips/vello_common/src/filter_effects.rs
+++ b/sparse_strips/vello_common/src/filter_effects.rs
@@ -20,6 +20,7 @@
 //! - `Flood` - Solid color fill
 //! - `GaussianBlur` - Gaussian blur filter
 //! - `DropShadow` - Drop shadow effect (compound primitive)
+//! - `DropShadowOnly` - Drop shadow effect without the original input
 //! - `Offset` - Translation/shift (single primitive)
 //!
 //! **Note:** Currently only single primitive filters are supported. Filter graphs with
@@ -410,6 +411,21 @@ pub enum FilterPrimitive {
         /// Default is `EdgeMode::None` per SVG spec.
         edge_mode: EdgeMode,
     },
+    /// Same as [`FilterPrimitive::DropShadow`], but without compositing the original
+    /// layer on top of it.
+    DropShadowOnly {
+        /// Horizontal offset of the shadow in pixels. Positive values shift right.
+        dx: f32,
+        /// Vertical offset of the shadow in pixels. Positive values shift down.
+        dy: f32,
+        /// Blur standard deviation for the shadow. Larger values create softer shadows.
+        std_deviation: f32,
+        /// Color applied to the blurred, offset input alpha mask. The input's color
+        /// channels are ignored, and this color's alpha controls shadow opacity.
+        color: AlphaColor<Srgb>,
+        /// Edge mode for handling boundaries during blur operation.
+        edge_mode: EdgeMode,
+    },
     //
     // ============================================================
     // TODO: The following filter primitives are not yet implemented
@@ -577,9 +593,14 @@ impl FilterPrimitive {
                 dx,
                 dy,
                 ..
+            }
+            | Self::DropShadowOnly {
+                std_deviation,
+                dx,
+                dy,
+                ..
             } => {
-                // Drop shadow = blur + offset + composite with original
-                // The expansion rect encompasses both the blur and the offset
+                // The expansion rect encompasses both the blur and the offset.
                 let blur_radius = (*std_deviation * 3.0) as f64;
                 let dx = *dx as f64;
                 let dy = *dy as f64;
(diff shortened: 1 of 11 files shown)


sparse_strips/vello_common/src/filter_effects.rs
```diff
@@ -20,6 +20,7 @@
 //! - `Flood` - Solid color fill
 //! - `GaussianBlur` - Gaussian blur filter
 //! - `DropShadow` - Drop shadow effect (compound primitive)
+//! - `DropShadowOnly` - Drop shadow effect without the original input
```
LaurenzV, line 23: Technically not a SVG filter I think, but should be fine to keep here, right?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -33,6 +33,7 @@ pub(crate) const FILTER_ATLAS_PADDING: u16 = MAX_KERNEL_SIZE as u16 / 2;
 const BYTES_PER_TEXEL: usize = 16;
 const FILTER_SIZE_BYTES: usize = 48;
 const FILTER_SIZE_U32: usize = FILTER_SIZE_BYTES / 4;
+const COMPOSITE_ORIGINAL_SHIFT: u32 = 13;
```
LaurenzV, line 36: I noticed that for the other ones, we just inlined the constants. We should change that in a follow-up.

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -318,10 +320,14 @@ impl GpuFilterData {
...
 
+    pub(crate) fn composite_original(&self) -> bool {
+        (self.data[0] >> COMPOSITE_ORIGINAL_SHIFT) & 1 != 0
+    }
+
     pub(crate) fn needs_copy_pass(&self) -> bool {
         // For drop shadows, we need to retain the original rendered layer because in the end
         // we need to composite it _on top_ of the actual shadow.
-        self.filter_type() == filter_type::DROP_SHADOW
+        self.filter_type() == filter_type::DROP_SHADOW && self.composite_original()
```
LaurenzV, line 342: Would it be better to introduce a separate filter type for `DROP_SHADOW_ONLY` here?

sparse_strips/vello_sparse_shaders/shaders/filter.wgsl
```diff
@@ -41,7 +42,8 @@ const MAX_TAPS_PER_SIDE: u32 = 3u;
 //   bits [5:6]   = edge_mode     (2 bits, only for blur filters), currently ignored.
 //   bits [7:10]  = n_decimations (4 bits, only for blur filters), only read on the CPU side.
 //   bits [11:12] = n_linear_taps (2 bits, only for blur filters)
-//   bits [13:32] = reserved for future use
+//   bit  [13]    = composite_original (only for drop shadows), only read on the CPU side.
```
LaurenzV, line 45: We should rework this so CPU-only fields don't need to be uploaded, but once again that's for a follow-up. Just noticed this now.

sparse_strips/vello_cpu/src/filter/drop_shadow.rs
```diff
@@ -83,15 +85,21 @@ fn apply_drop_shadow(
...
 
-/// Apply shadow color and composite with original.
+/// Apply the shadow color and optionally composite with the original.
 ///
 /// The shadow has already been offset and blurred, so this simply applies
-/// the shadow color to the alpha channel and composites using source-over.
-fn compose_shadow_direct(shadow: &Pixmap, dst: &mut Pixmap, color: AlphaColor<Srgb>) {
+/// the shadow color to the alpha channel and, when requested, composites the
+/// original over it using source-over.
+fn write_shadow_direct(
```
grebmeg: maybe `write_colored_shadow`?

sparse_strips/vello_hybrid/src/filter.rs
```diff
@@ -284,7 +286,7 @@ impl From<&DropShadow> for GpuDropShadow {
                 edge_mode_to_gpu(shadow.edge_mode),
                 shadow.n_decimations as u32,
                 lk.n_taps as u32,
-            ),
+            ) | (u32::from(shadow.composite_original) << COMPOSITE_ORIGINAL_SHIFT),
```
grebmeg: Could we add a compile-time check that the existing packer never uses the `composite_original` bit? Something like:

```rust
const COMPOSITE_ORIGINAL_MASK: u32 = 1 << COMPOSITE_ORIGINAL_SHIFT;

const _: () = assert!(
    pack_header_with_gaussian_params(31, 3, 15, 3) & COMPOSITE_ORIGINAL_MASK == 0
);
```

This requires making `pack_header_with_gaussian_params` a `const fn`. Because its fields are independently shifted and ORed, maximum values cover every bit it can use, catching future overlap at compile time without runtime overhead.

sparse_strips/vello_hybrid/src/filter.rs
grebmeg: I would keep the current option. `DropShadow` and `DropShadowOnly` share the same GPU data layout and offset/blur pipeline; only the final pass differs.

grebmeg (approved):
Thank you! ❤️

sparse_strips/vello_hybrid/src/filter.rs
LaurenzV: Added this.

PR #1764, opened 2026-07-24
ci: check sparse strips WASM binary sizes

grebmeg:
Adds reproducible browser-ready builds for the cpu, webgl, and wgpu sparse-strips variants. CI now enforces committed WASM size limits to detect binary-size regressions.


sparse_strips/README.md
```diff
@@ -27,6 +27,16 @@ This directory contains the core crates for the Vello rendering. Each crate serv
...
+## WebAssembly builds
+
+Browser-ready non-SIMD and SIMD128 builds are available in the [`web`](web)
+directory. Run the three SIMD128 modules, report their raw and gzip-compressed
+sizes, and enforce the raw size limits from the repository root with:
+
+```bash
+./sparse_strips/web/check_wasm_sizes.sh
+```
+
```
LaurenzV: This makes it sound like we are providing pre-built WASM binaries in this directory for others to use. If I'm understanding this is just a dev-only thing for us to verify the size, I'm not sure whether it "deserves" a section in the `sparse_strips` README. I think the README just in `web` should be enough. But up to you, at the very least I'd try to rephrase this section a bit. 😄

sparse_strips/web/check_wasm_sizes.sh
```diff
@@ -0,0 +1,66 @@
...
+    fi
+}
+
+printf '\nBuild profile: SIMD128, opt-level=3, fat LTO, codegen-units=1\n'
+printf '%-22s %10s %10s %10s\n' \
+    "WebAssembly module" "Raw" "Raw limit" "Gzip"
+check_size \
+    "vello_cpu" \
+    "$target_dir/sparse-strips-wasm/vello_cpu/simd/vello_cpu_bg.wasm" \
+    4456448 # 4.25 MiB
```
LaurenzV: What's the current size, does this contain some buffer already? Just so we don't have to raise the ceiling right in the next PR 😄

LaurenzV: Ah I just checked. That's quite a lot, maybe 4MiB would be a good starting point then? Up to you!

Image

.github/workflows/ci.yml
```diff
@@ -201,6 +201,31 @@ jobs:
...
+        with:
+          tool: wasm-bindgen@0.2.114
+
+      - name: restore cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name != 'merge_group' }}
+
+      - name: check sparse strips wasm sizes
+        run: ./sparse_strips/web/check_wasm_sizes.sh
```
LaurenzV, line 227: It's on purpose this will only check the SIMD variant, not non-SIMD, right? Just double checking, I think it makes sense to avoid stalling CI.

LaurenzV (approved):

sparse_strips/README.md
grebmeg: Good point — I rephrased this to clarify that `web` contains development tooling, while generated modules are written under `target/sparse-strips-wasm` and are not pre-built artifacts in the directory.

sparse_strips/web/check_wasm_sizes.sh
grebmeg: I tightened the initial thresholds: `3.6 MiB` for cpu and webgl, and `7 MiB` for wgpu.

.github/workflows/ci.yml
grebmeg: Yes, this is intentional. The size-check script explicitly builds only the SIMD variant to keep the CI job duration down; non-SIMD builds remain available through `build_wasm.sh` when needed.

PR #1768, opened 2026-07-27
vello_common: Improve filter bounds for drop shadows

LaurenzV:
Currently, the filter bounds of a drop shadow are called additively by combinining the expansion of the blur radius and potentially the shift. If the blur radius is 20, we always expand by 20 in each direction, regardless of the position of the drop shadow.

However, there is an easy optimization to apply here: In case the offset is for example (20, 20) and the blur radius is only 10, there is actually no need to expand the top/left side at all, because the expansion of the blur radius is completely absorbed by the shift. The following videos demonstrate the behavior before vs. after. As you can see, in the before case we always unconditionally expand to the top-left by the blur radius, regardless of where the drops shadow is positioned.

Before:

https://github.com/user-attachments/assets/de7787ba-cf69-4ec0-84f0-0d552ef62fbd

After:

https://github.com/user-attachments/assets/46989f54-36ba-4472-bf55-67b8a4d79d8d

### DIFF
--- a/sparse_strips/vello_common/src/filter_effects.rs
+++ b/sparse_strips/vello_common/src/filter_effects.rs
@@ -578,8 +578,7 @@ impl FilterPrimitive {
     pub fn filter_expansion(&self) -> Rect {
         match self {
             Self::GaussianBlur { std_deviation, .. } => {
-                // Gaussian blur expands uniformly by 3*sigma (covers 99.7% of distribution)
-                let radius = (*std_deviation * 3.0) as f64;
+                let radius = blur_radius(*std_deviation);
                 Rect::new(-radius, -radius, radius, radius)
             }
             Self::Offset { dx, dy } => {
@@ -600,16 +599,15 @@ impl FilterPrimitive {
                 dy,
                 ..
             } => {
-                // The expansion rect encompasses both the blur and the offset.
-                let blur_radius = (*std_deviation * 3.0) as f64;
-                let dx = *dx as f64;
-                let dy = *dy as f64;
+                let blur_radius = blur_radius(*std_deviation);
+                let dx = f64::from(*dx);
+                let dy = f64::from(*dy);
 
                 Rect::new(
-                    -(blur_radius + (-dx).max(0.0)),
-                    -(blur_radius + (-dy).max(0.0)),
-                    blur_radius + dx.max(0.0),
-                    blur_radius + dy.max(0.0),
+                    (dx - blur_radius).min(0.0),
+                    (dy - blur_radius).min(0.0),
+                    (dx + blur_radius).max(0.0),
+                    (dy + blur_radius).max(0.0),
                 )
             }
             // Most other filters don't expand bounds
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_common/src/filter_effects.rs
```diff
@@ -600,16 +599,15 @@ impl FilterPrimitive {
...
 
                 Rect::new(
-                    -(blur_radius + (-dx).max(0.0)),
-                    -(blur_radius + (-dy).max(0.0)),
-                    blur_radius + dx.max(0.0),
-                    blur_radius + dy.max(0.0),
+                    (dx - blur_radius).min(0.0),
+                    (dy - blur_radius).min(0.0),
+                    (dx + blur_radius).max(0.0),
+                    (dy + blur_radius).max(0.0),
```
grebmeg, line 610: I think this can cut off drop shadows when the X and Y scales are different.

The bounds and the rendered blur are scaled differently:

1. The bounds calculate a local `3 × sigma` rectangle here, then transform it with [`transform_rect_bbox`](https://github.com/linebender/vello/blob/f2187fcbcd7db7a9d32b63059e46086841618ca8/sparse_strips/vello_common/src/filter_effects.rs#L202-L209).
2. The renderer averages the X and Y scales in [`transform_blur_params`](https://github.com/linebender/vello/blob/f2187fcbcd7db7a9d32b63059e46086841618ca8/sparse_strips/vello_common/src/filter/gaussian_blur.rs#L15-L32), then uses that through [`transform_shadow_params`](https://github.com/linebender/vello/blob/f2187fcbcd7db7a9d32b63059e46086841618ca8/sparse_strips/vello_common/src/filter/drop_shadow.rs#L86-L110) and [`PreparedFilter::new`](https://github.com/linebender/vello/blob/f2187fcbcd7db7a9d32b63059e46086841618ca8/sparse_strips/vello_common/src/filter/mod.rs#L39-L91).

For `scale(2.0, 0.5)` and sigma `4`, the bounds reserve a vertical radius of `12 × 0.5 = 6`. The renderer uses the average scale `1.25`, producing a radius of `4 × 1.25 × 3 = 15`.

The resulting padding is used by [`FilterData::new`](https://github.com/linebender/vello/blob/f2187fcbcd7db7a9d32b63059e46086841618ca8/sparse_strips/vello_common/src/filter/mod.rs#L187-L245), so this difference can lead to clipping.

These tests reproduce the problem:

```rust
fn assert_contains(outer: Rect, inner: Rect) {
    assert!(
        outer.x0 <= inner.x0
            && outer.y0 <= inner.y0
            && outer.x1 >= inner.x1
            && outer.y1 >= inner.y1,
        "{outer:?} does not contain {inner:?}"
    );
}

fn non_uniformly_scaled_drop_shadow_expansions() -> (Rect, Rect, Rect, Rect) {
    let dx = 20.0;
    let dy = 0.0;
    let std_deviation = 4.0;
    let transform = Affine::scale_non_uniform(2.0, 0.5);

    let filter = Filter::from_primitive(FilterPrimitive::DropShadow {
        dx,
        dy,
        std_deviation,
        color: RED,
        edge_mode: EdgeMode::None,
    });

    let actual_filter = filter.filter_expansion(&transform);
    let actual_source = filter.source_expansion(&transform);

    let (dx, dy, std_deviation) =
        transform_shadow_params(dx, dy, std_deviation, &transform);
    let blur_radius = f64::from(std_deviation * 3.0);
    let dx = f64::from(dx);
    let dy = f64::from(dy);

    let expected_filter = Rect::new(
        (dx - blur_radius).min(0.0),
        (dy - blur_radius).min(0.0),
        (dx + blur_radius).max(0.0),
        (dy + blur_radius).max(0.0),
    );
    let expected_source = Rect::new(
        (-dx - blur_radius).min(0.0),
        (-dy - blur_radius).min(0.0),
        (-dx + blur_radius).max(0.0),
        (-dy + blur_radius).max(0.0),
    );

    (
        actual_filter,
        expected_filter,
        actual_source,
        expected_source,
    )
}

#[test]
fn transformed_filter_expansion_contains_executed_drop_shadow_support() {
    let (actual, required, _, _) =
        non_uniformly_scaled_drop_shadow_expansions();

    assert_contains(actual, required);
}

#[test]
fn transformed_source_expansion_contains_executed_drop_shadow_support() {
    let (_, _, actual, required) =
        non_uniformly_scaled_drop_shadow_expansions();

    assert_contains(actual, required);
}
```

Both tests fail because the calculated vertical bounds are `-6..6`, while the rendered blur requires `-15..15`.

grebmeg: I believe this finding predates the change, so if it makes sense, we could address it as a follow-up in the future.

grebmeg (approved):

sparse_strips/vello_common/src/filter_effects.rs
laurenz-canva: Will do in a follow-up!

PR #1770, opened 2026-07-27
sparse_strips: Document caveats more clearly

LaurenzV:
Since we want to make a 0.1 release, we should start advertising the current state and capabilities more prominently but also add more explicit warnings for current limitations.

### DIFF
--- a/sparse_strips/vello_cpu/src/lib.rs
+++ b/sparse_strips/vello_cpu/src/lib.rs
@@ -57,9 +57,9 @@
 //! assert_eq!(&result, expected_render);
 //! ```
 //!
-//! Feel free to take a look at some further
+//! See the
 //! [examples](https://github.com/linebender/vello/tree/main/sparse_strips/vello_cpu/examples)
-//! to better understand how to interact with Vello CPU's API,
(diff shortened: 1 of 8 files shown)


sparse_strips/vello_cpu/src/lib.rs
```diff
@@ -82,20 +82,26 @@
...
+//! # Current state
+//!
+//! Vello CPU is a solid CPU-only 2D renderer with broad, reliable feature
+//! support. It provides excellent performance across a wide range of workloads,
+//! with optimized SIMD implementations for all major architectures. The
+//! renderer is still under active development, however, and a few limitations
+//! remain:
+//!
+//! - Some features are not yet supported and may panic, including filters during multi-threaded
+//!   rendering and complex filter graphs.
```
nicoburns: It would be nice to be comprehensive here (I think this list might be, but isn't word as if it is)

nicoburns (approved):

sparse_strips/vello_cpu/src/lib.rs
LaurenzV: Yes should be exhaustive, will reword it.

PR #1771, opened 2026-07-28
vello_hybrid: Check WebGL errors before returning

LaurenzV:
I think this is a good thing to do for additional safety. I'm not sure whether the same should be added to `render_to_atlas`, but since it's only intended to be called internally right now for glyph caching I haven't done this here now, but open to changing this.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl.rs
@@ -50,9 +50,7 @@ use crate::{
         RootTarget, TextureParity,
     },
 };
-use alloc::sync::Arc;
-use alloc::vec;
-use alloc::vec::Vec;
+use alloc::{borrow::Cow, format, sync::Arc, vec, vec::Vec};
 use core::fmt::Debug;
 #[cfg(feature = "text")]
 use glifo::{GLYPH_PADDING, PendingClearRect};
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -299,6 +312,14 @@ impl WebGlRenderer {
             });
         }
 
+        // Note that this might yield errors that existed before Vello started
+        // rendering, but still better to err on the side of caution and return
+        // an error if there is one.
+        let webgl_error = self.gl.get_error();
```
LaurenzV, line 318: If I understood correctly there can be multiple queued errors, but I think it's enough to just get the first one?

taj-p: Isn't this a sync API? This API will cause a flush and full round trip to fetch errors from the GPU: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#avoid_blocking_api_calls_in_production.

So, if a consumer needs to render multiple scenes per frame, I think this could negatively impact their performance.

I think this can even impact performance if the consumer needs to run anything on the main thread after `.render` because now they must wait for a full GPU round trip before invoking their next line of JS 🤔 .

taj-p: It might be safe for a consumer to run at the very end of their frame (I'm genuinely unsure!), but I'm not sure that's something Vello knows about.

LaurenzV: Oh no. 🙈 Thanks for pointing this out, then we shouldn't do this.

PR #1776, opened 2026-07-29
sparse_strips: Version bump

LaurenzV:


sparse_strips/vello_cpu/CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 ## [Unreleased]
 
-TODO: Before making a 0.0.10 (or 0.1.0) release, resolve the following issue:
```
LaurenzV, line 13: I forgot about this, but I don't want to address this now, we can still change the exact semantics in the future if we want. There is nothing wrong with what we currently have, we just weren't sure whether it's the best API. I've moved the TODO to `CompositeMode` instead.

nicoburns (approved):

PR #1778, opened 2026-07-31
feat: improve atlas allocation diagnostics

grebmeg:
Improve atlas allocation diagnostics by reporting configured limits, requested and maximum sizes, available space, per-layer utilization, free rectangles, and fragmentation. Introduce dedicated intermediate-texture errors for oversized allocations and texture-count limits instead of reusing generic atlas errors with different semantics.

### DIFF
--- a/sparse_strips/vello_common/src/multi_atlas.rs
+++ b/sparse_strips/vello_common/src/multi_atlas.rs
@@ -139,7 +139,10 @@ impl MultiAtlasManager {
     /// Create a new atlas and return its ID.
     pub fn create_atlas(&mut self) -> Result<AtlasId, AtlasError> {
         if self.atlases.len() >= self.config.max_atlases {
-            return Err(AtlasError::AtlasLimitReached);
+            return Err(AtlasError::AtlasLimitReached {
+                max_atlases: self.config.max_atlases,
+                diagnostics: AtlasSpaceDiagnostics::Unavailable,
+            });
         }
 
         let atlas_id = AtlasId::new(self.next_atlas_id());
@@ -170,7 +173,12 @@ impl MultiAtlasManager {
     ) -> Result<AtlasAllocation, AtlasError> {
         // Check if the image is too large for any atlas
         if width > self.config.atlas_size.0 || height > self.config.atlas_size.1 {
-            return Err(AtlasError::TextureTooLarge { width, height });
+            return Err(AtlasError::TextureTooLarge {
+                width,
+                height,
+                max_width: self.config.atlas_size.0,
+                max_height: self.config.atlas_size.1,
+            });
         }
 
         // Try allocation based on strategy
@@ -185,6 +193,79 @@ impl MultiAtlasManager {
             AllocationStrategy::RoundRobin => {
                 self.allocate_round_robin(width, height, exclude_atlas_id)
             }
+        }
+    }
+
+    fn space_diagnostics(
+        &self,
+        width: u32,
+        height: u32,
+        exclude_atlas_id: Option<AtlasId>,
+    ) -> AtlasSpaceDiagnostics {
+        let mut atlases = Vec::new();
+
+        for atlas in &self.atlases {
+            if Some(atlas.id) == exclude_atlas_id {
+                continue;
+            }
+
+            let mut free_area = 0_u64;
+            let mut free_rectangle_count = 0;
+            let mut largest_free_width = 0;
+            let mut largest_free_height = 0;
+            let mut largest_free_area = 0_u64;
+            atlas.allocator.for_each_free_rectangle(|rect| {
+                let rect_width = rect.width() as u32;
+                let rect_height = rect.height() as u32;
+                let rect_area = u64::from(rect_width) * u64::from(rect_height);
+                free_area += rect_area;
+                free_rectangle_count += 1;
+
+                if rect_area > largest_free_area {
+                    largest_free_area = rect_area;
+                    largest_free_width = rect_width;
+                    largest_free_height = rect_height;
+                }
+            });
+
+            atlases.push(AtlasLayerDiagnostics {
+                atlas_id: atlas.id,
+                total_area: u64::from(atlas.stats.total_area),
+                free_area,
+                free_rectangle_count,
+                largest_free_width,
+                largest_free_height,
+            });
+        }
+
+        AtlasSpaceDiagnostics::Allocation {
+            width,
+            height,
+            atlas_width: self.config.atlas_size.0,
+            atlas_height: self.config.atlas_size.1,
+            max_atlases: self.config.max_atlases,
+            atlases,
+        }
+    }
+
+    fn no_space_available(
+        &self,
+        width: u32,
+        height: u32,
+        exclude_atlas_id: Option<AtlasId>,
+    ) -> AtlasError {
+        AtlasError::NoSpaceAvailable(self.space_diagnostics(width, height, exclude_atlas_id))
+    }
+
+    fn atlas_limit_reached(
+        &self,
+        width: u32,
+        height: u32,
+        exclude_atlas_id: Option<AtlasId>,
+    ) -> AtlasError {
+        AtlasError::AtlasLimitReached {
+            max_atlases: self.config.max_atlases,
+            diagnostics: self.space_diagnostics(width, height, exclude_atlas_id),
         }
     }
 
@@ -210,6 +291,9 @@ impl MultiAtlasManager {
 
         // Try creating a new atlas if auto-grow is enabled
         if self.config.auto_grow {
+            if self.atlases.len() >= self.config.max_atlases {
+                return Err(self.atlas_limit_reached(width, height, exclude_atlas_id));
+            }
             let atlas_id = self.create_atlas()?;
             let atlas = self.atlases.last_mut().unwrap();
             if let Some(allocation) = atlas.allocate(width, height) {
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_hybrid/src/lib.rs
LaurenzV:
```suggestion
    /// An image atlas allocation failed.
```

We are only using it for images now, right?

sparse_strips/vello_hybrid/src/lib.rs
```diff
@@ -119,6 +119,30 @@ pub enum RenderError {
...
+    /// A render requires more intermediate textures than configured.
+    #[error(
+        "Render requires {required} intermediate textures, exceeding the configured maximum of {max}"
+    )]
+    IntermediateTextureLimitReached {
+        /// The number of intermediate textures required by the render.
+        required: usize,
+        /// The configured maximum number of intermediate textures.
+        max: usize,
+    },
```
LaurenzV, line 154: Maybe let's create an IntermediateTexture variant with an inner enum that contains `TooLarge` and `LimitReached`?

sparse_strips/vello_hybrid/src/schedule/mod.rs
```diff
@@ -251,11 +250,16 @@ impl IntermediateTextureRequirements {
         self,
         existing: IntermediateTextureAllocations,
         max_textures: Option<usize>,
-    ) -> Result<(), AtlasError> {
+    ) -> Result<(), RenderError> {
```
LaurenzV: And here we just return an `IntermediateError`.

sparse_strips/vello_hybrid/src/schedule/allocate.rs
```diff
@@ -128,6 +132,10 @@ impl LayerAllocationRequest {
             region,
         }
     }
+
+    pub(super) fn allocation_size(self) -> SizeU32 {
+        SizeU32::from(self.region.size) + u32::from(self.region.padding) * 2
```
LaurenzV: Can we put this logic  in `RegionProps`, just call it here and then also reuse the method in line 204? So that the same logic is not encoded in two places.

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -141,6 +141,8 @@ impl LayersConfig {
                 return Err(AtlasError::TextureTooLarge {
```
LaurenzV: I think we should use `IntermediateTextureTooLarge` here right?

sparse_strips/vello_common/src/multi_atlas.rs
```diff
@@ -188,6 +196,79 @@ impl MultiAtlasManager {
...
+        &self,
+        width: u32,
+        height: u32,
+        exclude_atlas_id: Option<AtlasId>,
+    ) -> AtlasError {
+        AtlasError::AtlasLimitReached {
+            max_atlases: self.config.max_atlases,
+            diagnostics: self.space_diagnostics(width, height, exclude_atlas_id),
+        }
+    }
```
LaurenzV: If you think we need this it's fine for me, but I imagine that calling `space_diagnostics _could_ be a bit expensive if there are lots of allocated rectangles? I haven't looked at how this is implemented. But overall probably still negligible.

LaurenzV: After looking at the code it seems very cheap though.

sparse_strips/vello_common/src/multi_atlas.rs
```diff
@@ -210,6 +291,9 @@ impl MultiAtlasManager {
 
         // Try creating a new atlas if auto-grow is enabled
         if self.config.auto_grow {
+            if self.atlases.len() >= self.config.max_atlases {
+                return Err(self.atlas_limit_reached(width, height, exclude_atlas_id));
+            }
```
LaurenzV: It seems a bit superfluous to call this here when `create_atlas` also does the check, no? Can `create_atlas` not just switch to return diagnostics?

sparse_strips/vello_common/src/multi_atlas.rs
```diff
@@ -333,6 +417,9 @@ impl MultiAtlasManager {
 
         // Try creating a new atlas if auto-grow is enabled
         if self.config.auto_grow {
+            if self.atlases.len() >= self.config.max_atlases {
```
LaurenzV: Same here.

sparse_strips/vello_common/src/multi_atlas.rs
LaurenzV: Ah okay, we need the original method because there's not always a context available... Would it be possible to just call `create_atlas` and use `map_err` instead?

sparse_strips/vello_common/src/multi_atlas.rs
```diff
@@ -188,6 +196,79 @@ impl MultiAtlasManager {
         }
     }
 
+    fn space_diagnostics(
+        &self,
+        width: u32,
+        height: u32,
+        exclude_atlas_id: Option<AtlasId>,
```
LaurenzV: What's the point of excluding an atlas from the statistics themselves? Wouldn't it be better to still have the statistics for all atlases?

By the way, I just realized that the whole `excluding` thing was needed for the old filter system, but since this is gone now we don't need this anymore. I'll remove it in a follow-up.

sparse_strips/vello_common/src/multi_atlas.rs
```diff
@@ -395,24 +482,183 @@ impl core::fmt::Debug for MultiAtlasManager {
...
+
+    /// Calculate layer fragmentation as a percentage from 0 to 100.
+    pub fn fragmentation_percentage(&self) -> f64 {
+        if self.free_area == 0 {
+            0.0
+        } else {
+            let largest_free_area =
+                u64::from(self.largest_free_width) * u64::from(self.largest_free_height);
+            (1.0 - largest_free_area as f64 / self.free_area as f64) * 100.0
+        }
```
LaurenzV, line 601: What's the rationale for having such a metric?

LaurenzV (commented):
I don't really understand the fragmentation percentage, but otherwise LGTM!

LaurenzV (approved):

sparse_strips/vello_hybrid/src/lib.rs
grebmeg: Done.

sparse_strips/vello_hybrid/src/lib.rs
grebmeg: Done.

sparse_strips/vello_hybrid/src/schedule/mod.rs
grebmeg: Done.

sparse_strips/vello_hybrid/src/schedule/allocate.rs
grebmeg: Good catch — moved it to `RegionProps::allocation_size` and both call sites now use it.

sparse_strips/vello_hybrid/src/render/common.rs
grebmeg: Yes, good catch!

sparse_strips/vello_common/src/multi_atlas.rs
grebmeg: Agreed. `space_diagnostics` is only reachable from the two error constructors, so it never runs on a successful allocation, and guillotiere's `for_each_free_rectangle` is a linear scan over the node list with no allocation. The only heap allocation is the per-layer `Vec`, bounded by `max_atlases`.

sparse_strips/vello_common/src/multi_atlas.rs
grebmeg: Good point — both sites now just call `create_atlas` and use `map_err` to attach the allocation diagnostics, so the limit check lives only in `create_atlas`.

sparse_strips/vello_common/src/multi_atlas.rs
grebmeg: Done

sparse_strips/vello_common/src/multi_atlas.rs
grebmeg: You're right, diagnostics should describe the whole atlas array — otherwise the reported layer count and aggregate utilization disagree with reality. Dropped `exclude_atlas_id` from `space_diagnostics` and the two error constructors, so all layers are reported.

sparse_strips/vello_common/src/multi_atlas.rs
grebmeg: The metric distinguishes two failure modes that utilization alone can't tell apart. An allocation needs one contiguous free rectangle, not just enough total free area — so when it fails, either the layer is genuinely full (low fragmentation, little free area), in which case raising `max_atlases` or the atlas size helps, or the layer has plenty of free area chopped into pieces too small for the request (high fragmentation), in which case extra capacity won't help and the real issue is packing/ordering. 0% means all free space is one contiguous rectangle.

Worth noting it reflects `guillotiere`'s own partition rather than idealized geometry — it merges free siblings on deallocation but only along its guillotine tree, so adjacent-but-unrelated rectangles stay split. That's the useful reading, since it's what the allocator can actually serve. Happy to drop it if you still think it's more confusing than helpful.

LaurenzV: I don't mind, if you think it will be useful then let's keep it.

PR #1779, opened 2026-07-31
vello_hybrid: Add better statistics for probing

LaurenzV:
This PR adds a convenience method to get some more details about why the probe exactly failed.

This PR was done with assistance of GPT Sol 5.6

### DIFF
--- a/sparse_strips/vello_common/src/probe.rs
+++ b/sparse_strips/vello_common/src/probe.rs
@@ -98,7 +98,7 @@ pub struct ProbeStatistics {
     /// Width and height of the actual probe image.
     pub actual_size: (u16, u16),
     /// Number of pixels whose channels differ by more than the probe tolerance.
-    pub different_pixel_count: usize,
+    pub different_pixel_count: u32,
     /// Largest absolute difference between corresponding red, green, blue, and alpha channels.
     pub max_channel_discrepancy: [u8; 4],
     /// Bitmask identifying probe features containing a pixel outside the probe tolerance.
--- a/sparse_strips/vello_common/src/probe.rs
+++ b/sparse_strips/vello_common/src/probe.rs
@@ -99,8 +99,8 @@ pub struct ProbeStatistics {
     pub mismatched_dimensions: bool,
     /// Number of pixels whose channels differ by more than the probe tolerance.
     pub different_pixel_count: usize,
-    /// Largest absolute difference between corresponding channels.
-    pub max_channel_discrepancy: u8,
+    /// Largest absolute difference between corresponding red, green, blue, and alpha channels.
+    pub max_channel_discrepancy: [u8; 4],
     /// Bitmask identifying probe features containing a pixel outside the probe tolerance.
     ///
     /// Bit `n` corresponds to the [`ProbeFeature`] whose discriminant is `n`.
@@ -133,10 +133,12 @@ impl ProbeResult {
             .enumerate()
         {
             if expected[3] != 0 || actual[3] != 0 {
-                for (expected, actual) in expected.iter().zip(actual) {
-                    statistics.max_channel_discrepancy = statistics
-                        .max_channel_discrepancy
-                        .max(expected.abs_diff(*actual));
+                for (max_discrepancy, (expected, actual)) in statistics
+                    .max_channel_discrepancy
+                    .iter_mut()
+                    .zip(expected.iter().zip(actual))
+                {
+                    *max_discrepancy = (*max_discrepancy).max(expected.abs_diff(*actual));
                 }
             }
 
@@ -547,6 +549,7 @@ mod tests {
 
         set_channel(&mut actual, 1, 0, 249);
         set_channel(&mut actual, 5, 1, 0);
+        set_channel(&mut actual, 5, 3, 100);
 
         let result = ProbeResult { expected, actual };
         let statistics = result.statistics();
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_common/src/probe.rs
```diff
@@ -63,6 +63,63 @@ pub struct ProbeResult {
...
+    /// Whether the expected and actual images have different dimensions.
+    pub mismatched_dimensions: bool,
+    /// Number of pixels whose channels differ by more than the probe tolerance.
+    pub different_pixel_count: usize,
+    /// Largest absolute difference between corresponding channels.
+    pub max_channel_discrepancy: u8,
+    /// Bitmask identifying probe cells containing a pixel outside the probe tolerance.
+    ///
+    /// Bit `n` corresponds to the `n`th probe cell in rendering order.
+    pub difference_mask: u32,
```
grebmeg: Could we expose stable, named identifiers for the probe elements? Consumers currently have to copy the private `ELEMENTS` order to decode this mask. Reordering elements or re-enabling Filter would shift bit meanings, causing historical telemetry to be interpreted as failures of different features.

```rust
#[repr(u8)]
#[derive(Clone, Copy)]
pub enum ProbeFeature {
    SolidRect = 0,
    AlphaBlending = 1,
    Gradient = 2,
    // ...
}
impl ProbeStatistics {
    pub fn differs(&self, feature: ProbeFeature) -> bool {
        self.difference_mask & (1_u32 << feature as u8) != 0
    }
}
```

sparse_strips/vello_common/src/probe.rs
```diff
@@ -63,6 +63,63 @@ pub struct ProbeResult {
...
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct ProbeStatistics {
+    /// Number of probe cells represented by [`Self::difference_mask`].
+    pub element_count: usize,
+    /// Whether the expected and actual images have different dimensions.
+    pub mismatched_dimensions: bool,
+    /// Number of pixels whose channels differ by more than the probe tolerance.
+    pub different_pixel_count: usize,
+    /// Largest absolute difference between corresponding channels.
+    pub max_channel_discrepancy: u8,
```
grebmeg: Could `max_channel_discrepancy` be `[u8; 4]` for RGBA? A single maximum hides whether the failure comes from color or alpha, while per-channel maxima would make telemetry much more actionable. This may also fit into the padding currently needed before `difference_mask`, so the in-memory size might not be increases.

sparse_strips/vello_common/src/probe.rs
```diff
@@ -63,6 +63,63 @@ pub struct ProbeResult {
     pub actual: ProbeImage,
 }
 
+/// Summary of the differences between the expected and actual probe images.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct ProbeStatistics {
+    /// Number of probe cells represented by [`Self::difference_mask`].
+    pub element_count: usize,
+    /// Whether the expected and actual images have different dimensions.
+    pub mismatched_dimensions: bool,
```
grebmeg: Since the reference size is fixed, could we report `actual_size` instead of `mismatched_dimensions`? The mismatch boolean can be derived by comparing it with `canvas_size()`, while the actual dimensions explain what went wrong and make standalone telemetry more useful.

sparse_strips/vello_common/src/probe.rs
```diff
@@ -63,6 +63,63 @@ pub struct ProbeResult {
     pub actual: ProbeImage,
 }
 
+/// Summary of the differences between the expected and actual probe images.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct ProbeStatistics {
+    /// Number of probe cells represented by [`Self::difference_mask`].
+    pub element_count: usize,
```
grebmeg: Could `element_count` use `u8` instead of `usize`? Its serialized size would then be consistent across platforms, and the `u32` difference mask already limits the representation to 32 elements.

sparse_strips/vello_common/src/probe.rs
```diff
@@ -63,6 +63,63 @@ pub struct ProbeResult {
...
+    /// Whether the expected and actual images have different dimensions.
+    pub mismatched_dimensions: bool,
+    /// Number of pixels whose channels differ by more than the probe tolerance.
+    pub different_pixel_count: usize,
```
grebmeg: Since the probe reference is fixed at 51×51 pixels, could `different_pixel_count` be `u16`? Its maximum value is only 2,601, so usize is larger than necessary and platform-dependent.

grebmeg (approved):
Thank you! LGTM once comments addressed.

sparse_strips/vello_common/src/probe.rs
LaurenzV: makes sense!

sparse_strips/vello_common/src/probe.rs
LaurenzV: Sure! I think it's fine to keep it as a separate field for now for clarity,  clients can serialize it in different ways if they want to.

sparse_strips/vello_common/src/probe.rs
LaurenzV: Makes sense.

sparse_strips/vello_common/src/probe.rs
LaurenzV: I'll make it `u32`, just in case we ever make the probe larger.

PR #1780, opened 2026-07-31
vello_cpu: Fix horizontal extend mode for filtered images

ShiroKSH:
## Summary

- use the X-axis extend mode for horizontal filtered-image taps
- cover differing X/Y extend modes with a focused SIMD regression test

## Root cause

The `ImageSampler` migration wired the horizontal helper to `y_extend`. As a result, bilinear and bicubic CPU sampling used the Y mode on both axes when the modes differed.

## Checks

- `cargo test --locked -p vello_cpu`
- `cargo test --locked -p vello_cpu --all-features`
- `cargo clippy -p vello_cpu --locked --all-targets --all-features -- -D warnings`
- `cargo check --locked -p vello_cpu --no-default-features --features libm,f32_pipeline`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/sparse_strips/vello_cpu/src/fine/common/image.rs
+++ b/sparse_strips/vello_cpu/src/fine/common/image.rs
@@ -221,7 +221,7 @@ impl<S: Simd, const QUALITY: u8> Iterator for FilteredImagePainter<'_, S, QUALIT
                 extend(
                     self.simd,
                     x_positions + $offsets[$idx],
-                    self.data.image.sampler.y_extend,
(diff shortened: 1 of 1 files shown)


LaurenzV:
Looks good, sorry for the delay, will try to merge tomorrow! Just going to add one more visreg test for it.

LaurenzV:
I've opened a new PR in #1803 since there was some weird issue with git LFS when attempting to push to your branch, but I made sure to include your original commit for attribution. We will likely just merge that instead, thanks!

PR #1781, opened 2026-07-31
vello_hybrid: Avoid keeping two versions of a resized texture in memory at the same time

LaurenzV:
As mentioned in the comment, I'm not sure whether this actually helps in practice (I'm not sure how to measure that in the first place), but in theory this _should_ help, and it certainly shouldn't come with any downsides.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -70,6 +70,7 @@ impl<T> ScratchTexture<T> {
         &self.0
     }
 
+    #[cfg(feature = "webgl")]
     pub(crate) fn get_mut(&mut self) -> &mut T {
         &mut self.0
     }
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -69,11 +69,6 @@ impl<T> ScratchTexture<T> {
     pub(crate) fn get(&self) -> &T {
         &self.0
     }
-
-    #[cfg(feature = "webgl")]
-    pub(crate) fn get_mut(&mut self) -> &mut T {
-        &mut self.0
-    }
 }
 
 /// Scratch allocations reused while rendering a frame.
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -69,6 +69,10 @@ impl<T> ScratchTexture<T> {
     pub(crate) fn get(&self) -> &T {
         &self.0
     }
+
+    pub(crate) fn get_mut(&mut self) -> &mut T {
+        &mut self.0
+    }
 }
 
 /// Scratch allocations reused while rendering a frame.
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -56,10 +56,7 @@ use crate::{
 use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
-use core::{
-    fmt::Debug,
-    ops::{Deref, DerefMut},
-};
+use core::fmt::Debug;
 #[cfg(feature = "text")]
 use glifo::{GLYPH_PADDING, PendingClearRect};
 use hashbrown::HashMap;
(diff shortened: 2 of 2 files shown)


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -998,6 +1001,48 @@ struct WebGlIntermediateTexture {
...
+        }
+    }
+
+    fn resize(&mut self, create: impl FnOnce() -> WebGlIntermediateTexture) {
+        // Explicitly invoke the destructor of the old texture before creating
+        // the new one. The hope is that this will allow the WebGL driver to recognize
+        // that the memory of the old texture can be deallocated before allocating the
+        // new one, reducing peak memory consumption. It has not been validated whether
+        // this actually makes any difference in practice, but it shouldn't hurt either.
+        drop(self.inner.take());
```
taj-p: Do we need to use delete texture for this to work properly? 

https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/deleteTexture

LaurenzV: Yes but  `Texture` is just a type alias for our `Resource<WebGLTexture>`:

https://github.com/linebender/vello/blob/b377de1be0f93ba2d1c651e3d654b66f1107a720/sparse_strips/vello_hybrid/src/render/webgl.rs#L3372

And `WebGlTexture` has is a resource that has a `Drop` impl:

https://github.com/linebender/vello/blob/b377de1be0f93ba2d1c651e3d654b66f1107a720/sparse_strips/vello_hybrid/src/render/webgl.rs#L3211-L3296

So that should be fine unless I'm missing something, right?

taj-p: Ohhhh yes. Perfect. Thanks

taj-p (approved):

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -998,6 +1001,48 @@ struct WebGlIntermediateTexture {
...
+
+        self.inner = Some(create());
+    }
+}
+
+impl Deref for ResizableTexture {
```
grebmeg: Could we avoid introducing `ResizableTexture` here? Since resized textures don’t preserve their contents, we can drop the old allocations before rebuilding the vector:

```rust
if size_changed {
    let page_count = textures.len().max(required_page_count);
    textures.clear();
    textures.extend(
        (0..page_count).map(|_| create_intermediate_texture(gl, texture_size)),
    );
} else {
    textures.extend(
        (textures.len()..required_page_count)
            .map(|_| create_intermediate_texture(gl, texture_size)),
    );
}
```

The scratch texture already uses an `Option`, so it can follow the same pattern:

```rust
if recreate_scratch {
    self.resources.scratch_texture = None;
    self.resources.scratch_texture = Some(ScratchTexture::new(
        create_intermediate_texture(gl, texture_size),
    ));
}
```

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -1122,27 +1167,30 @@ impl WebGlPrograms {
             // Note: Currently, `texture_size` only grows across frames, so if this condition
             // is true it means that the new required size is larger than what we currently have.
             if size_changed {
-                // TODO: Drop old layer explicitly before creating new one to ensure WebGL
-                // driver has the chance to reduce peak memory usage instead of having to keep both
-                // textures alive at the same time?
                 for texture in textures.iter_mut() {
```
grebmeg: Another idea would be to call `gl.flush()` after dropping all the old textures. This may encourage earlier cleanup, although it doesn’t guarantee that the memory will be reclaimed immediately. It would be interesting to validate, but I understand that measuring this reliably isn’t easy.

grebmeg (approved):

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: I’m so silly haha 😄 Thanks for finding this

sparse_strips/vello_hybrid/src/render/webgl.rs
LaurenzV: After discussion in private we'll avoid this for now.

grebmeg (approved):

PR #1783, opened 2026-07-31
vello_hybrid: Make clear behavior more customizable

LaurenzV:
This PR attempts to split out the part of #1737 that relates to controlling how to clear the existing contents of the main framebuffer. I’ve gone through a lot more iterations than I initially expected, but this is where I landed now. It’s slightly different than the previously proposed API, but the idea is the same.

First, there was the question of whether there should be an explicit `clear_rects` method or whether it should be part of the render method. I believe the latter is the better option: It makes it impossible for clients to accidentally forget about clearing the output and forces a conscious decision every time `render` is called, which I think is the better option.

Inititially, I had a `LoadOp` enum similar to the original PR, but I believe there isn’t much point in having it: `LoadOp::Load` is basically the same as just not clearing anything in both backends, and, if my thinking is correct, we cannot really use `DontCare` for wgpu. The [documentation](https://docs.rs/wgpu/latest/wgpu/enum.LoadOp.html#variant.DontCare) of the mode basically says this option should only be chosen if all pixels are overwritten. However, even if the user _thinks_ that all pixels are overwritten, it doesn’t necessarily mean that this is internally the case the first time we load the root render target and render to it. This is because not all parts of all shapes might end up actually being executed in the opaque draw pass or a first alpha render pass. 

In order for the user to use it correctly, we’d have to leak internal implementation details of Vello Hybrid, which I’d like to avoid. Therefore, settling on a single enum that specifies the clear behavior seemed like the best choice here.

Looking forward to hearing your thoughts!

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -15,7 +15,8 @@ use crate::filter::FILTER_ATLAS_PADDING;
 use crate::scene::{LayersConfig, MemorySettings, RecordedDraw};
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
-use vello_common::geometry::SizeU16;
+use vello_common::color::{AlphaColor, Srgb};
+use vello_common::geometry::{RectU16, SizeU16};
 use vello_common::record::CommandRecorder;
 
 // GPU paint structure sizes in texels (1 texel = 16 bytes for RGBA32Uint texture format).
@@ -31,6 +32,42 @@ pub(crate) const GPU_BLURRED_ROUNDED_RECT_SIZE_TEXELS: u32 =
 // TODO: If we want to use native bilinear sampling for uploaded images,
 // we can pass 1 instead of 0 here.
 pub(crate) const IMAGE_PADDING: u16 = 0;
+
+/// Controls how the output target is cleared before drawing.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ClearSettings<'a> {
+    /// Do not clear the target.
+    DontClear,
+    /// Clear the complete viewport to `color`.
+    Viewport {
+        /// The clear color.
+        color: AlphaColor<Srgb>,
+    },
+    /// Clear each rectangle to `color`.
+    Rects {
+        /// The clear color.
+        color: AlphaColor<Srgb>,
+        /// The rectangles to clear.
+        rects: &'a [RectU16],
+    },
+}
+
+impl ClearSettings<'_> {
+    pub(crate) fn clear_color(self) -> Option<AlphaColor<Srgb>> {
+        match self {
+            Self::DontClear => None,
+            Self::Viewport { color } | Self::Rects { color, .. } => Some(color),
+        }
+    }
+}
+
+impl Default for ClearSettings<'_> {
+    fn default() -> Self {
+        Self::Viewport {
+            color: AlphaColor::TRANSPARENT,
+        }
+    }
+}
 
 /// Texture limits reported by the rendering device.
 #[derive(Debug, Clone, Copy)]
@@ -96,7 +133,7 @@ pub(crate) struct ScratchBuffers {
     pub(crate) copy_instances: Vec<GpuCopyInstance>,
 }
 
-/// Per-instance data for clearing a rectangle in an intermediate texture.
+/// Per-instance data for clearing a rectangle in a render target.
 #[repr(C)]
 #[derive(Debug, Copy, Clone, Pod, Zeroable)]
 pub(crate) struct GpuClearInstance {
@@ -106,6 +143,8 @@ pub(crate) struct GpuClearInstance {
     pub(crate) size: [u32; 2],
     /// Width and height of the target texture.
     pub(crate) target_size: [u32; 2],
+    /// Premultiplied clear color.
+    pub(crate) color: u32,
 }
 
 impl MemorySettings {
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -15,7 +15,8 @@ use crate::filter::FILTER_ATLAS_PADDING;
 use crate::scene::{LayersConfig, MemorySettings, RecordedDraw};
 use alloc::vec::Vec;
 use bytemuck::{Pod, Zeroable};
-use vello_common::geometry::{SizeU16, SizeU32};
+use vello_common::color::{AlphaColor, Srgb};
+use vello_common::geometry::{RectU16, SizeU16, SizeU32};
 use vello_common::record::CommandRecorder;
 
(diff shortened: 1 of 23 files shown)


AdrianEddy:
This shape works well for the damage-rendering use case, thanks for splitting it out.

One suggestion for the wgpu `clear_pass_inner`: the `ClearSettings::Rects` pass records `LoadOp::Load` + `StoreOp::Store` over the whole target with no scissor, so on tile-based GPUs clearing a handful of small rects still loads and stores every tile of the target. Setting the pass scissor to the union of the target-clamped clear rects avoids that, and it's behavior-neutral since the instance quads never write outside the union. It would also bring the wgpu path in line with the WebGL backend, which already issues scissored clears.

Separately, `clear_pass_inner` creates a fresh instance buffer per call via `create_buffer_init` (that cost predates this PR — the layer-clear path on main does the same). [I have a change](https://github.com/AdrianEddy/vello/commit/f02f951f9021299658c12a2dea4ea93c6bdcb8ec) that stages instances into a persistent pooled buffer with a per-render cursor; I'd send that as a follow-up PR once this lands rather than grow this one.

Additionally, the tests are failing/hanging on this PR (and all my PRs based on this one)

LaurenzV:
> Setting the pass scissor to the union of the target-clamped clear rects avoids that, and it's behavior-neutral since the instance quads never write outside the union.

Do we have any evidence that it actually helps? Judging from the [documentation](https://docs.rs/wgpu/29.0.3/wgpu/struct.RenderPass.html#method.set_scissor_rect), all it does it discard fragments outside of the rects, so I'm not sure it would bring any benefits with regard to load and store? I admittedly don't know much about how GPUs work internally, but would be good to see a source backing this up before adding this.

> Separately, clear_pass_inner creates a fresh instance buffer per call via create_buffer_init (that cost predates this PR — the layer-clear path on main does the same).

Yes, that's for the future.

> Additionally, the tests are failing/hanging on this PR (and all my PRs based on this one)

Yup, this is why I opened #1787, this should fix that. It's unrelated to this PR. So that will have to be merged first.

LaurenzV:
That failure on Ubuntu is certainly intriguing... Not sure if it's a wgpu bug or whether we are doing something wrong.

clear_area_rects_use_requested_color_hybrid

AdrianEddy:
I investigated the Ubuntu CI failure:

**It's a Mesa/llvmpipe driver bug.** llvmpipe's *linear rasterizer fast path* packs the blend constant for RGBA-ordered render targets in the wrong byte order — `(A,R,G,B)` instead of `(R,G,B,A)`. The rect-clear draw is the only thing in the whole vello suite that meets all the trigger conditions at once.

## The trigger (all three required)

1. **Axis-aligned quad geometry** — a TriangleStrip quad or a two-triangle list quad. A single triangle takes the general path and is correct.
2. **`BlendFactor::Constant`** — One/Zero blending and no-blend are both correct; it's specifically the constant fetch.
3. **The linear path must be permitted**: single RGBA8/BGRA8 color target, **no depth/stencil attachment**, single viewport ([lp_state_derived.c:229-241](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/gallium/drivers/llvmpipe/lp_state_derived.c)). The rect-clear pass is vello's only depthless pass — every strip pass carries depth, so nothing else ever hits it.

## What this means for the PR

Pinning or upgrading Mesa in CI won't help — the bug spans at least 25.2.8→26.0.8. The robust fix is to drop `set_blend_constant` and carry the color **per-instance in `GpuClearInstance`** (unpacked in the shader). That also kills the dynamic state and matches what the WebGL backend needs anyway.

LaurenzV:
Thanks! That's unfortunate, in that case I guess passing it dynamically is the right choice.

sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -32,6 +33,42 @@ pub(crate) const GPU_BLURRED_ROUNDED_RECT_SIZE_TEXELS: u32 =
...
+        /// The clear color.
+        color: AlphaColor<Srgb>,
+    },
+    /// Clear each rectangle to `color`.
+    Rects {
+        /// The clear color.
+        color: AlphaColor<Srgb>,
+        /// The rectangles to clear.
+        rects: &'a [RectU16],
+    },
```
grebmeg, line 52: Would it make sense to also support a separate color for each rectangle while keeping this shared-color variant? For example, an additional `ColoredRects { rects: &[ClearRect] }` variant could handle per-rectangle colors without making callers repeat the same color when all rectangles should match.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -362,6 +362,7 @@ impl WebGlRenderer {
         resources: &mut Resources,
         render_size: &RenderSize,
         texture_bindings: &WebGlTextureBindings,
+        clear: ClearSettings<'_>,
```
grebmeg, line 421: In #1850 we detect the scene’s root background, while this PR controls how the render target is initialized through viewport and rectangle clears. They are not exactly the same concept, but they are closely related: an opaque viewport clear provides background-opacity information, while a detected full-viewport background could potentially be implemented as an efficient clear. Would it make sense to develop these as stacked PRs, or combine them, so both features are designed around one coherent model?

sparse_strips/vello_hybrid/src/render/common.rs
LaurenzV: I mean we could do that in theory, but I don't really see the use case here, when would you want to clear different regions of the viewport with different colors?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Hmm, I mean in principle I agree it would be nice to have a coherent story here. I also wouldn't mind stacking. However, I don't think that Vello Hybrid and Vello CPU necessarily need to have the same API here: The capabilities exposed by the devices (GPU vs CPU) are different, and therefore it makes sense to expose them differently. For example, clear rects make a lot of sense on GPU because it's easy to implement, but not so much on CPU. 

In general, my idea would have been to also add `RootBackground` to Vello Hybrid, and then basically override the clear settings provided by the user. For example, if the user wants to clear with a transparent color but we know the scene has a full white background, we can just override it to clear with a white color instead. I think something like this would be doable.

~But if you think it requires more discussion I understand, it's just that not merging #1850 in some form will block us from getting RGB output out of Vello (unless we add a hacky API where the user can force discarding the alpha output)~ (nvm! this shouldn't be blocking anymore, it would just make it faster). And again, we should probably get some input from @upsuper on how we can handle the backgrounds on our side.

LaurenzV:
Hmm, by the way, it seems like the background method currently is not properly affected by destructive blends. Will mark this as draft until I fix this.

LaurenzV:
Superseded by #1869.

PR #1787, opened 2026-08-01
vello_cpu: Initialize multi-threaded dispatcher lazily

LaurenzV:
See #1749 for more context. This turned out to actually be easier than I thought. `register_task`, which is used by all operations that dispatch in MT, actually already has logic for lazily initializing the context in case it's not initialiazed yet:

https://github.com/linebender/vello/blob/b377de1be0f93ba2d1c651e3d654b66f1107a720/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L237-L242

Therefore, we just don't need to take care at all of initializing the render context when constructing the dispatcher for the first time, or when resetting it.

However, we need to add a new guard in `flush_tasks` to make sure that a `push_clip_path` that is called before anything else doesn't result in a panic, since it calls `flush_tasks` unconditionally:

https://github.com/linebender/vello/blob/b377de1be0f93ba2d1c651e3d654b66f1107a720/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L702-L717

CC @yezhizhen


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -1236,6 +1236,51 @@ mod tests {
...
+                num_threads: 1,
+                ..Default::default()
+            },
+        );
+        let clip = Rect::new(0.0, 0.0, 50.0, 50.0).to_path(0.1);
+
+        // Just make sure we don't panic.
+        ctx.push_clip_path(&clip);
+        ctx.flush();
+    }
```
yezhizhen, line 1284:
```suggestion
        ctx.pop_clip_path(&clip);
        ctx.flush();
    }
```

yezhizhen (approved):
Can we also test https://github.com/linebender/vello/blob/b377de1be0f93ba2d1c651e3d654b66f1107a720/sparse_strips/vello_cpu/src/dispatch/multi_threaded.rs#L719-L722?

grebmeg (approved):

yezhizhen:
Any chance we can get 0.1.1 release since it patches a panic for just released 0.1.0?

LaurenzV:
Unfortunately we've already had breaking changes since then in `vello_common`. :( So it might have to wait until the next minor bump, which might not be too far off. But if it's really urgent maybe we can prioritize it more.

yezhizhen:
No worry. Thanks for doing this! The panic in wpt tests are gone with this patch :)
https://github.com/servo/servo/pull/46915#issuecomment-5186959752

jschwe:
> Unfortunately we've already had breaking changes since then in vello_common. :( So it might have to wait until the next minor bump, which might not be too far off. But if it's really urgent maybe we can prioritize it more.

Would you consider backporting this patch onto a `0.1` branch and publishing that as 0.1.1? We don't allow depending on git dependencies anymore in servo, and the `0.1` release does otherwise seem to fix a few issues, so it would be nice to be able to update.

LaurenzV:
I can ask, but would just a 0.2 release work, too?

jschwe:
Sure, that would work too.

LaurenzV:
Opened a PR.

Issue #204, opened 2022-11-17
Plan for glyph rendering


raphlinus:
This issue is branched from a question raised https://github.com/linebender/xilem/pull/1 about our plan for glyph rendering, as that is the responsibility of piet-gpu rather than other layers in the stack.

For the first cut, all glyph rendering will happen as dynamic vector graphics rendering, getting the outlines from the font (currently through swash but possibly migrating to Oxidize) and assembling them into scene fragments. At first we will also not support hinting at all, nor stem darkening, nor RGB subpixel rendering. Results will be pretty good on high-dpi screens, but definitely suboptimal on low-dpi. Performance may be a problem on very low-spec GPUs as well.

The roadmap has a number of improvements. One is to support a glyph cache, which by itself doesn't change rendering quality, but will improve performance greatly (especially on low-spec GPUs, in which pulling textures from the glyph atlas will outperform rendering vectors). The infrastructure for glyph (which involves tracking the transform at which a glyph will be displayed) also enables hinting (which is fairly well implemented by swash already).

When hinting is enabled, it will be a configurable option. It is desirable for UI text (and you'd want it for displaying the text in a text editor), but it is *not* desirable when the scene is more dynamic. Cases where hinting would cause "shimmering" effects include pinch-to-zoom (or any other transform than scrolling) and animation of variable font parameters.

An important quality improvement (especially on macOS) is stem darkening. The best way to implement this is to make coarse path rasterization capable of offset as well as flattening (the same logic will support both stroke rendering and stem darkening). I've figured out the math for this and have a prototype (in JS) but haven't yet implemented it as a compute shader; there are some tricky aspects, especially dealing with joins at corners. This improvement is orthogonal to glyph caching.

Lastly, to support low-dpi screens (especially Windows) we would want to implement RGB subpixel rendering, which is mostly a change in fine rasterization.

With a glyph cache in place, we have a choice. Either this cache is filled by doing path rendering in piet-gpu, or by calling into platform glyph rendering (particularly DirectWrite on Windows). The latter would allow pixel-perfect matching to platform text, but at a performance cost and with other tradeoffs (having two pipelines rather than one always causes problems). My current inclination is to render all the glyphs ourselves, but this is something we can revisit. Note that one option for platform glyph rendering is to take a dependency on font-kit.

canadaduane:
I'd just like to cross-pollinate slimsag's work here in the Zig language on curve primitives:

[...]

PR #1790, opened 2026-08-02
sparse_strips: CoverageContrast, an opt-in coverage transfer for glyph sharpening

AdrianEddy:
## Summary

Adds `CoverageContrast`, an opt-in coverage transfer applied to alpha-mask tints (glyph coverage) before tinting:

```
a' = a + c·a(1−a)(2a−1)   // steepen: lerp(identity, smoothstep)
       + w·a(1−a)         // weight bias: raises the 0.5 crossing by w/4
```

The default is bit-exact off: every consumer branches on `is_none()` before evaluating any new math, so existing output does not change by a single byte unless a renderer opts in via `RenderContext::set_glyph_coverage_contrast` / `Scene::set_glyph_coverage_contrast`.

## Motivation

Vello's analytic coverage is the exact area of the pixel square covered by the outline — at a straight edge, a linear ramp with slope exactly 1.0 alpha/px. That is the correct answer to "how much of this pixel is ink," but it is not what any mainstream text stack displays: every one of them deliberately steepens (and, in some regimes, thickens) the transition, because the exact ramp reads as slightly soft, most visibly on low-DPI displays.

This gap is called out in Vello's own planning documents:

- [vision.md, "Improving rendering quality"](https://github.com/linebender/vello/blob/main/doc/vision.md): *"a theoretically 'correct' approach to gamma often yields text and hairline strokes that appear weak and spindly … it's likely that various heuristics will need to be applied to get good results. (Note that stem darkening is one approach used specifically for text rendering …)"*
- [Plan for glyph rendering #204](https://github.com/linebender/vello/issues/204) notes the initial glyph strategy supports neither hinting-adjacent contrast boosts nor stem darkening, with results "suboptimal on low-dpi displays."
- The theme goes back to [*A crate I want: 2d graphics* (2018)](https://raphlinus.github.io/rust/graphics/2018/10/11/2d-graphics.html): *"to match high quality desktop rendering both RGB subpixel rendering and gamma correct blending is desirable."*

This PR is a small, opt-in implementation of exactly that predicted heuristic, in the factoring that fits the sparse-strips architecture.

## Relation to the stem-darkening roadmap thread

[roadmap_2023.md](https://github.com/linebender/vello/blob/main/doc/roadmap_2023.md) and [*Raph's reflections and wishes for 2023*](https://raphlinus.github.io/personal/2022/12/31/raph-2023.html) proposed solving stem darkening geometrically, via a combined flatten + offset operation — *"the same mechanism can apply stroke thickening to glyphs while retaining very high quality."* The geometric half of that vision has since landed: `kurbo::expand_path` plus glifo's `FontEmbolden` (#1628).

This PR adds the complementary coverage-space half, and the two compose rather than compete:

- `FontEmbolden` offsets outlines in em space, can add arbitrary weight, and participates in the glyph cache key — correct for a *global* weight change, but a *per-text-color* weight (see polarity below) would rasterize and cache every glyph once per color class.
- The `w` term here is a sub-pixel, device-space adjustment to *sampled* coverage at fine/shader time — bounded by a fraction of a pixel, matched to the per-pixel blending artifact it corrects, and free to vary per draw: light and dark text share one cached rasterization at zero atlas churn.

## Prior art in other renderers

| Renderer | Mechanism | Relation to this PR |
|---|---|---|
| Skia | `SkMaskGamma`: mask contrast `a + k·a(1−a)` plus luminance-dependent gamma LUTs | the `w` term is exactly Skia's contrast form, minus the LUTs (see below) |
| DirectWrite | enhanced contrast `a(k+1)/(ak+1)`, `k` gated by text luminance (visible in [Windows Terminal's](https://github.com/microsoft/terminal/blob/main/src/renderer/atlas/dwrite_helpers.hlsl) and [Zed's](https://github.com/zed-industries/zed/blob/main/crates/gpui_windows/src/alpha_correction.hlsl) shaders) | same family to first order; its gate boosts *dark* text because DWrite's blend model is gamma-corrected — in an sRGB-blending renderer the boost belongs on *light* text, which is the direction implemented here |
| FreeType / CoreText | stem darkening / font smoothing: geometric outline emboldening | the geometric complement; already available via `FontEmbolden` |
| Qt Quick (distance-field text) | `smoothstep` over a constant 1.2 device-px band → peak slope 1.25 alpha/px | `c = 0.5` reproduces its output to within one 8-bit level across the core of the edge ramp |

The polarity rationale for `w`: compositing in sRGB-encoded space (which vello_cpu/vello_hybrid do) decodes a half-covered pixel darker than the physically correct blend — dark-on-light text reads slightly bolder, light-on-dark slightly thinner. `resolve_for_color` scales the stored `w` by the text color's approximate relative luminance, so black text keeps the weight-free curve bit-exactly and white text receives the full stored strength.

Compared to the Skia/DWrite implementations, the factoring here is deliberately simpler and colorspace-agnostic: one closed-form curve on coverage *before* tinting, no baked LUTs coupled to a destination gamma, CPU and GPU evaluating identical 8-bit-quantized parameters.

## Design notes

- **Provable invariant instead of per-pixel clamps.** With `t = a − ½` the derivative is `1 + c/2 − 6c·t² − 2w·t`, concave, minimized at `a = 1` where it equals `1 − c − w`. `from_bits` therefore caps `w` at the contrast's headroom (`c + w ≤ 1`); under that invariant the curve is monotone with range [0, 1], so none of the three kernels needs a clamp.
- **Bit-exactness guarantees.** `NONE` is a branch, not neutral math; and `w = 0` leaves `c`-only output bit-identical because the appended weight term is exactly `+0.0`. Both are pinned by tests.
- **Wire format.** The packed tint mode word carries the strengths in bits 8–23; `NONE` contributes zero bits, so paints that don't opt in encode byte-identically. The previously whole-word shader mode compare is masked accordingly.
- **glifo integration.** A default-implemented `GlyphRenderer::glyph_coverage_contrast` hook; the atlas outline-glyph path stamps the luminance-resolved value. Bitmap/COLR glyphs (own color, no coverage mask) and atlas-miss outline fills (linear coverage) are unaffected — the latter is a documented limitation shared with hinting and subpixel-offset bucketing.

## What this deliberately does not attempt

- **Gamma-correct (linear) blending** — vision.md's own caution applies: linear blending makes text spindly *without* compensation, so it is a much larger, coupled change. This curve is compatible with such a future: it operates on coverage pre-tint, so only the calibration constant would move.
- **RGB subpixel rendering** — incompatible with transparent-layer compositing, as the roadmap notes.
- **Backdrop-aware correction** — the backdrop is unknowable at paint time in a compositing renderer; the same limitation applies to Skia and DirectWrite.

## Testing

- `vello_common/tests/coverage_contrast.rs`: property tests for bit-exact identity, fixed endpoints, symmetry of the `c` term, exhaustive monotonicity/range sweeps (including dense probing at the capped boundary where the derivative vanishes), the headroom cap, the Skia-form equivalence of the `w` term, luminance resolution (endpoints, gray = ¼, Rec. 709 ordering, NaN/out-of-range handling), and u8/f32 agreement.
- `pack_tint` unit test pinning the wire layout.
- Two new cross-backend snapshot tests (`c`-only and `c + w`) through the existing spritesheet-tint harness, so the CPU fine stages and the WESL shader are pinned against the same references.
- `cargo fmt`, `clippy --all-targets`, and rustdoc are clean.

## Breaking change

`Tint` gained a `contrast` field (`CoverageContrast::NONE` preserves the previous behavior); changelog entries note it under "Changed" for `vello_common` and "Added" for `vello_cpu`, `vello_hybrid`, and `glifo`.

<sup>This PR was generated by Claude</sup>

### DIFF
--- a/glifo/CHANGELOG.md
+++ b/glifo/CHANGELOG.md
@@ -14,7 +14,7 @@ This release has an [MSRV][] of 1.88.
 
 ### Added
 
-- `GlyphRenderer::glyph_coverage_contrast` (default `CoverageContrast::NONE`), applied to atlas-cached outline glyphs with the weight term resolved per draw against the tint color's luminance. ([#1790][] by [@AdrianEddy][])
+- `GlyphRenderer::glyph_coverage_contrast` (default `CoverageContrast::NONE`), applied to solid-painted outline glyphs with the weight term resolved per draw against the text color's luminance. Atlas-cached glyphs receive the transfer through their alpha-mask tint at sample time; glyphs drawn directly from their outlines receive the same transfer at fill time through the new `GlyphRenderer::fill_glyph_path` hook, so appearance does not depend on whether a glyph is cached. ([#1790][] by [@AdrianEddy][])
 
 ## [0.3.0][] - 2026-08-07
 
--- a/glifo/CHANGELOG.md
+++ b/glifo/CHANGELOG.md
@@ -12,6 +12,10 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 This release has an [MSRV][] of 1.88.
 
+### Added
+
+- `GlyphRenderer::glyph_coverage_contrast` (default `CoverageContrast::NONE`), applied to atlas-cached outline glyphs with the weight term resolved per draw against the tint color's luminance. ([#1790][] by [@AdrianEddy][])
+
 ## [0.2.0][] - 2026-07-29
 
 This release has an [MSRV][] of 1.88.
@@ -55,6 +59,8 @@ Glifo moved to the Vello repo in [#1539][] and was prepared for release by [@con
 [#1672]: https://github.com/linebender/vello/pull/1672
 [#1774]: https://github.com/linebender/vello/pull/1774
 
+[#1790]: https://github.com/linebender/vello/pull/1790
+[@AdrianEddy]: https://github.com/AdrianEddy
 [Unreleased]: https://github.com/linebender/vello/compare/glifo-v0.2.0...HEAD
 [0.2.0]: https://github.com/linebender/vello/compare/glifo-v0.1.1...glifo-v0.2.0
 [0.1.1]: https://github.com/linebender/vello/compare/glifo-v0.1.0...glifo-v0.1.1
--- a/glifo/CHANGELOG.md
+++ b/glifo/CHANGELOG.md
@@ -14,7 +14,7 @@ This release has an [MSRV][] of 1.88.
 
 ### Added
 
-- `GlyphRenderer::glyph_coverage_contrast` (default `CoverageContrast::NONE`), applied to atlas-cached outline glyphs with the weight term resolved per draw against the tint color's luminance. ([#1790][] by [@AdrianEddy][])
+- `GlyphRenderer::glyph_coverage_contrast` (default `CoverageContrast::NONE`), applied to solid-painted outline glyphs with the weight term resolved per draw against the text color's luminance. Atlas-cached glyphs receive the transfer through their alpha-mask tint at sample time; glyphs drawn directly from their outlines receive the same transfer at fill time through the new `GlyphRenderer::fill_glyph_path` hook, so appearance does not depend on whether a glyph is cached. ([#1790][] by [@AdrianEddy][])
 
 ## [0.2.0][] - 2026-07-29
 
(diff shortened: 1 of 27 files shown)


AdrianEddy:
Rebased onto current main. #1791 changed the no-tint encoding to `(u32::MAX, Multiply)`, and the shader compares the whole mode word with `==`. With the contrast strengths packed into bits 8-23 that comparison would break, so the tint mode is now read through a low-byte mask (`get_image_tint_mode`), and the `pack_tint` test pins that the no-tint encoding carries no contrast bits.

PR #1791, opened 2026-08-02
vello_hybrid: Properly handle tints with opacity 0

LaurenzV:
This PR fixes a bug where setting the tint to `0` doesn't actually make the image fully transparent. It's extracted from #1736, but only contains the tinting fix without the `ImageSampler` opacity change, to make it easier to review. 

However, this PR actually uses a different approach: We simply encode no tint as `u32::MAX` (all lanes to 1.0) and use the multiply mode, which means that the color of the pixel will just always stay the same. This way, we don't need to introduce a third `None` mode and can just keep the existing ones.

The operation that does the actual multiplication multiplies in both cases:

```
final_color = alpha * select(
    image_tint * sample_color.a,
    sample_color * image_tint,
    is_multiply
);
```

So by doing this, we are also not incurring more arithmetic operations or anything, it should be the same in terms of efficiency.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -597,7 +597,7 @@ pub(crate) fn pack_image_params(
 /// Pack an optional [`Tint`](vello_common::paint::Tint) into a (`tint_color_u32`, `tint_mode_u32`) pair for the GPU.
 ///
 /// The tint color is premultiplied before packing into a u32 in the same layout
-/// as WGSL `pack4x8unorm`. No tint is encoded as opaque white in multiply mode.
+/// as WGSL `pack4x8unorm`.
 #[inline(always)]
 pub(crate) fn pack_tint(tint: Option<vello_common::paint::Tint>) -> (u32, u32) {
     match tint {
--- a/sparse_strips/vello_hybrid/src/render/common.rs
+++ b/sparse_strips/vello_hybrid/src/render/common.rs
@@ -594,26 +594,18 @@ pub(crate) fn pack_image_params(
     (atlas_index << 6) | (extend_y << 4) | (extend_x << 2) | quality
 }
 
-const GPU_TINT_MODE_NONE: u32 = 0;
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_hybrid/src/render/common.rs
```diff
@@ -598,16 +597,15 @@ pub(crate) fn pack_image_params(
...
+/// as WGSL `pack4x8unorm`.
 #[inline(always)]
 pub(crate) fn pack_tint(tint: Option<vello_common::paint::Tint>) -> (u32, u32) {
     match tint {
         Some(t) => {
             let color = t.color.premultiply().to_rgba8().to_u32();
             (color, t.mode.as_u32())
         }
-        None => (0, 0),
+        None => (u32::MAX, vello_common::paint::TintMode::Multiply.as_u32()),
```
grebmeg, line 611: Could we move the explanation from the PR descriptiong here as a comment? It’s not immediately obvious that `u32::MAX` unpacks to opaque white, making multiplication an identity operation for `None`.

sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -726,3 +726,24 @@ fn image_spritesheet_tinted(ctx: &mut impl Renderer) {
         cursor_x += glyph.width;
     }
 }
+
+#[vello_test]
+fn image_fully_transparent_tint(ctx: &mut impl Renderer) {
```
grebmeg, line 731: Would it be possible to add a test for tinting an external texture as well?

grebmeg (approved):

sparse_strips/vello_sparse_tests/tests/image.rs
LaurenzV: We also don't have tests for tinting with `Multiply`, so let me just add a TODO to increase test coverage for external textures in general, since it's not really related to this fix specifically.

sparse_strips/vello_hybrid/src/render/common.rs
LaurenzV: Done.

PR #1792, opened 2026-08-03
feat: support external textures in WebGL backend

grebmeg:
Adds external texture sampling to the native WebGL backend, bringing `Scene::draw_texture_rects` to parity with wgpu. Introduces `WebGlTextureBindings`, validates bindings during paint preparation, and switches bound textures between scheduled strip runs.

Extends the standard `vello_test` harness with `hybrid_reference` so hybrid can generate snapshots when CPU rendering is unsupported. External texture tests now use this shared harness and run against WebGL in browsers.


sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -84,6 +86,11 @@ const GPU_PAINT_PLACEHOLDER: GpuEncodedPaint = GpuEncodedPaint::LinearGradient(G
     transform: [0.0; 6],
 });
 
+const EXTERNAL_IMAGE_SOURCE_FLAG: u32 = 1 << 14;
+
+/// Texture unit the strip program samples external textures from.
+const EXTERNAL_TEXTURE_UNIT: u32 = 5;
```
LaurenzV, line 90: Feel free to ignore, but as far as I can tell all other texture units are currently just inlined. For consistency, maybe we should just inline it as well now and in a follow-up create constants for _all_ of them?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -84,6 +86,11 @@ const GPU_PAINT_PLACEHOLDER: GpuEncodedPaint = GpuEncodedPaint::LinearGradient(G
     transform: [0.0; 6],
 });
 
+const EXTERNAL_IMAGE_SOURCE_FLAG: u32 = 1 << 14;
```
LaurenzV: Should we move this into the method where it's used? Since it's just one.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -716,6 +796,30 @@ impl WebGlRenderer {
...
+        let transform = texture.transform.as_coeffs().map(|x| x as f32);
+        let region = texture.source_region;
+        let image_size = pack_image_size(region.width(), region.height());
+        let image_offset = pack_image_offset(region.x0, region.y0);
+        let image_params = pack_image_params(
+            texture.sampler.quality as u32,
+            texture.sampler.x_extend as u32,
+            texture.sampler.y_extend as u32,
+            0,
+        ) | EXTERNAL_IMAGE_SOURCE_FLAG;
```
LaurenzV: Since the bit layout is internal to `pack_image_params`, I think it would be better if the method takes a boolean and then `pack_image_params` assigns the flag based on it. This way all the bit-fiddling happens within a single location.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -118,6 +125,52 @@ pub struct WebGlRenderer {
     layers_config: LayersConfig,
 }
 
+/// Runtime bindings for [externally owned textures](`TextureId`) sampled by texture-rect draws.
+#[derive(Debug, Default, Clone)]
+pub struct WebGlTextureBindings {
```
LaurenzV, line 128: Maybe just `TextureBindings` to make it consistent with wgpu backend?

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -2397,15 +2501,95 @@ fn initialize_strip_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources)
...
+            self.bind_external_texture(None);
+        }
+
+        self.set_strip_attrib_offset(0);
+    }
+
+    /// Draw `count` strip instances starting at `first_instance`.
+    ///
+    /// WebGL2 has no base-instance argument, so the vertex attributes are offset to the first
+    /// requested instance before issuing the instanced draw.
```
LaurenzV, line 2549: nit: This is already documented in `draw_strips`, so no need to add it twice I think? Or maybe remove it from `draw_strips` instead.

sparse_strips/vello_hybrid/src/render/webgl.rs
```diff
@@ -2882,6 +3034,7 @@ impl Backend for WebGlRendererContext<'_> {
         self.strip_pass_inner(
             strips,
             RangedSlice::empty(),
+            &[],
```
LaurenzV, line 3035: Just as a note, as mentioned in a different TODO there is no way to set an external texture as being opaque. Therefore, if we want to make use of this, maybe we need to implement this as well? Unless we are fine just treating all images as possibly transparent for now.

sparse_strips/vello_sparse_tests/tests/external_texture.rs
```diff
@@ -82,322 +79,290 @@ mod tests {
...
-        f: impl FnOnce(&mut HybridRenderer),
-    ) {
-        let mut ctx = get_ctx::<HybridRenderer>(
-            WIDTH,
-            HEIGHT,
-            false,
-            0,
-            "fallback",
-            RenderMode::OptimizeQuality,
+    #[vello_test(width = 96, height = 96, skip_cpu, hybrid_reference)]
```
LaurenzV: I think the only case where we will ever have hybrid be the gold standard is if it's not supported by CPU. Should we maybe rename it to `hybrid_only` and have that imply `skip_cpu`? So that only one attribute is needed.

LaurenzV (approved):
LGTM. I think it would be worth it to expand the test coverage a bit (especially the interaction with opaque fast strips path), but doesn't have to be in this PR.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Good point about consistency. I’d prefer to keep this one named because the same unit must stay synchronized between the per-run binding and sampler setup. We can normalize the remaining texture units in a follow-up.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: I think this is mostly a matter of taste. I prefer the standalone named constant because an inline `1 << 14` isn’t self-describing, while `EXTERNAL_IMAGE_SOURCE_FLAG` clearly communicates what the bit represents. A method-local named constant would also work, but offers little benefit beyond narrower scope.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Agreed. I added an `is_external` parameter so the complete bit layout is now handled by `pack_image_params`, and removed the duplicate flag constants from both backends.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: I kept the `WebGl` prefix to follow the existing backend naming pattern (`Renderer`/`WebGlRenderer` and `AtlasWriter`/`WebGlAtlasWriter`). It also avoids a name collision when the `webgl` and `wgpu` features are enabled together, since both types are re-exported at the crate root.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Agreed. I removed the duplicated explanation from `draw_strips` and kept it on `draw_strip_range`, where the offset is applied.

sparse_strips/vello_hybrid/src/render/webgl.rs
grebmeg: Yes, treating all external textures as potentially transparent is intentional for now. Since `may_have_transparency` is always true, they cannot enter the opaque pass, so passing no external-texture runs here is correct. Supporting an opacity hint and the corresponding opaque-path bindings would be a useful follow-up optimization.

sparse_strips/vello_sparse_tests/tests/external_texture.rs
grebmeg: Agreed. I replaced `skip_cpu`, `hybrid_reference` with `hybrid_only`, which now implies both behaviors. This removes the flexibility to run cpu renderer against a hybrid-generated reference, but we currently have no such cases and can reintroduce that option if needed.

grebmeg (commented):
Thanks for the review! I added two tests exercising external textures alongside opaque strips: one for opaque/alpha ordering and another for multiple external texture runs after opaque strips.

PR #1794, opened 2026-08-03
vello_hybrid: Tie `Resources` to a specific renderer instance

LaurenzV:
The `AtlasConfig` struct defines the settings that should be used for the image atlas array as well as intermediate textures. Unfortunately, right now, those are passed in two different places. **Currently, the user needs to manually ensure that they are the same, otherwise silent corruption will happen.** In addition, even if the user does so, the two settings might run out of sync because the renderer needs to normalize the settings to the capabilities of the device.

The reason why the information is needed in two different places is glyph caching. On the one hand, `Renderer` needs it to actually allocate appropriately sized textures. On the other hand, `Scene` needs it because glyph caching is currently tightly coupled with it. In the future, we can hopefully disentangle this, but it's a bigger project.

Therefore, as a stopgap measure, I made the following change: The configuration is now only supplied **once** during renderer construction. The constructor will normalize the settings, and then generate and return a new `Resources` struct based on this instead of having the user construct it. This way, it's always guaranteed that the two match. The downside is that it forces the user to create the renderer before the scene (at least if they want to use glyph caching), but I think this is acceptable for now.

I also don't really like the constructor returning a tuple, but for now that's by far the better option than users accidentally misusing the API and having silent failures.

### DIFF
--- a/sparse_strips/vello_hybrid/CHANGELOG.md
+++ b/sparse_strips/vello_hybrid/CHANGELOG.md
@@ -18,6 +18,8 @@ This release has an [MSRV][] of 1.88.
 
 ### Changed
 
+- Breaking change: `Renderer::new`, `Renderer::new_with`, `WebGlRenderer::new`, and `WebGlRenderer::new_with` now return the renderer and its associated `Resources` as a tuple. The given `Resources` should only be used in conjunction with the renderer instance that generated it. 
+- Breaking change: `Scene::new_with` now takes a `Level` instead of `RenderSettings`.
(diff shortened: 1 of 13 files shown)


LaurenzV:
Another option would be to just store `Resources` on the renderer itself and expose a `resources_mut` function that the user can call.

grebmeg (approved):
Thanks! LGTM. Could you please update CHANGELOG.md to document this breaking change?

RobertBrewitz:
This change made life more difficult :frowning_face: :laughing: :heart:

LaurenzV:
I mean, how did you handle this before? 🤔  In principle, nothing has changed in this regard, you've always needed one set of resources per renderer for correctness.

As mentioned, the API is definitely not ideal and will change in the future, but it's the best we can do for now until this is resolved.

PR #1796, opened 2026-08-03
Switch to proposed load_four_interleaved() API in fearless_simd

Shnatsel:
Update to https://github.com/linebender/fearless_simd/pull/298

This removes a whole bunch of redundant split/combine ops.

Marking as draft because this is a breaking change and the matching fearless_simd version is not released yet.

MSRV check on CI fails because recent fearless_simd needs 1.89 while vello MSRV is 1.88

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1084,9 +1084,8 @@ dependencies = [
 
 [[package]]
 name = "fearless_simd"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76258897e51fd156ee03b6246ea53f3e0eb395d0b327e9961c4fc4c8b2fa151a"
+version = "0.6.0"
+source = "git+https://github.com/Shnatsel/fearless_simd.git?branch=load-interleaved-4-way#f62652d90cbc423dc967a5ea6b05158da388519a"
 dependencies = [
  "libm",
 ]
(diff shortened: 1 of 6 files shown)


LaurenzV:
That's just for store, right? What about load? The code snippets updated here are not that important, but the `load_interleaved_128_u32x16` in `pack_block` is super important, so we need to make sure that one doesn't regress.

LaurenzV:
Ahh nvm, it's there, I missed it. Let me benchmark.

LaurenzV:
Okay, benchmarks are looking good.

grebmeg:
Great work! One small ask: with these internal changes, until we have the regression infrastructure in place, could we include some relevant benchmark results in the PR showing the before-and-after perf?

LaurenzV:
Sure, before we actually update fearless_simd next time I’ll benchmark everything 👍The main reason it hasn’t happened yet is I still need to figure out what to do about AVX512 (because we can’t run it in CI).

Shnatsel:
You can't benchmark AVX-512 in CI, but you can run programs that use it via Intel's emulator: https://github.com/linebender/fearless_simd/blob/afd9fad01a02291b3e42b9fd2527563f84a79677/.github/workflows/ci.yml#L223-L282

LaurenzV:
Yeah but I’m pretty sure this will be too slow for CI. 😄 Or does the emulator JIT somehow?

Shnatsel:
I don't know the implementation details, since the emulator is proprietary, but `sde64 -icl -- cargo test --release --tests --lib --bins` (AVX-512 emulation, running tests excluding doctests) in vello_cpu/ completes in 5 seconds on my machine, assuming the tests are compiled already outside the emulator.

LaurenzV:
Interesting, thanks, will explore this then.

Shnatsel:
Closing in favor of #1802

PR #1797, opened 2026-08-04
vello_hybrid: Move backends into their own folders

LaurenzV:
I'm planning to split up the backend files into multiple modules soon (they are 3k+ lines!). The first step is to create the directories in the first place and move the files into them.


grebmeg (approved):
Nice! LGTM!

PR #1800, opened 2026-08-06
vello_hybrid(webgl): Mali-safe image atlas allocation and growth

grebmeg:
### Problem
- Scenes without alphas, encoded paints, gradients, or filters leave the corresponding WebGL textures incomplete, making affected draw calls non-renderable and producing repeated warnings.
- Growing the image atlas can also hang the Android WebView compositor on Mali-G52 and cause an ANR, especially with larger atlas sizes.

### Root Cause
- Optional data textures had no backing storage until their first upload.
- For atlas growth, the working hypothesis is a Mali-G52/WebView driver ordering issue. Reallocating and copying a large texture array immediately before compositing creates a timing window where queued GPU work has not yet been submitted. Larger atlases widen this window. Replacing the initial 1x1 atlas placeholder also triggers the compositor hang.

### Solution
- Allocate minimal backing storage for optional data textures during initialization.
- Resize the initial atlas placeholder in place to preserve its GL texture identity. When replacing an atlas containing data, unbind and flush before allocation as a precaution, copy the existing layers, then perform the critical trailing `gl.flush()` that prevents the Mali-G52 compositor hang. This uses `gl.flush()` rather than synchronously waiting for GPU completion.

### DIFF
--- a/sparse_strips/vello_hybrid/CHANGELOG.md
+++ b/sparse_strips/vello_hybrid/CHANGELOG.md
@@ -24,8 +24,7 @@ This release has an [MSRV][] of 1.88.
 
 ### Fixed
 
-- The alphas, encoded paints, gradient and filter data textures of the `webgl` backend are now allocated at creation. Previously a scene without alphas, encoded paints, gradients or filters left the corresponding texture incomplete, which made every draw call that bound it non-renderable and emitted a render warning per draw. (by [@grebmeg][])
-- The `webgl` image atlas now grows without hanging the Android WebView (Mali-G52) GL compositor. Growing from the lazily-allocated 1x1 placeholder (`initial_atlas_count == 0`) now re-specifies the backing store in place on the existing `TEXTURE_2D_ARRAY` instead of allocating a replacement texture and deleting the placeholder, and growing an atlas that already holds data unbinds the atlas and `glFlush`es around the reallocation. Previously, either could hang the compositor mid-session, causing an ANR. (by [@grebmeg][])
+- Fixed a bug in the `webgl` backend that could cause an ANR on Mali-G52 GPUs. (by [@grebmeg][])
 
 ## [0.1.0][] - 2026-07-29
 
--- a/sparse_strips/vello_hybrid/CHANGELOG.md
+++ b/sparse_strips/vello_hybrid/CHANGELOG.md
@@ -22,6 +22,11 @@ This release has an [MSRV][] of 1.88.
 - Breaking change: `Scene::new_with` now takes a `Level` instead of `RenderSettings`.
 - Breaking change: `WebGlRenderer::render` and `WebGlRenderer::render_to_atlas` now take a `WebGlTextureBindings`, providing the external textures referenced by the scene. Pass `&WebGlTextureBindings::new()` if the scene does not use any.
 
+### Fixed
+
+- The alphas, encoded paints, gradient and filter data textures of the `webgl` backend are now allocated at creation. Previously a scene without alphas, encoded paints, gradients or filters left the corresponding texture incomplete, which made every draw call that bound it non-renderable and emitted a render warning per draw. (by [@grebmeg][])
+- The `webgl` image atlas now grows without hanging the Android WebView (Mali-G52) GL compositor. Growing from the lazily-allocated 1x1 placeholder (`initial_atlas_count == 0`) now re-specifies the backing store in place on the existing `TEXTURE_2D_ARRAY` instead of allocating a replacement texture and deleting the placeholder, and growing an atlas that already holds data unbinds the atlas and `glFlush`es around the reallocation. Previously, either could hang the compositor mid-session, causing an ANR. (by [@grebmeg][])
+
 ## [0.1.0][] - 2026-07-29
 
 This release has an [MSRV][] of 1.88.
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -1284,6 +1284,7 @@ impl WebGlPrograms {
             // texture name is required on Mali-G52 under the Android WebView GL compositor: replacing
             // the stub object mid-session hangs the native compositor, causing an ANR, and flushing
             // around the swap does not help.
+            gl.active_texture(WebGl2RenderingContext::TEXTURE0);
             gl.bind_texture(
                 WebGl2RenderingContext::TEXTURE_2D_ARRAY,
                 Some(&self.resources.atlas_texture_array.texture),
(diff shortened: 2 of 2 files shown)


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1264,20 +1264,60 @@ impl WebGlPrograms {
...
+                0,
+                WebGl2RenderingContext::RGBA,
+                WebGl2RenderingContext::UNSIGNED_BYTE,
+                None,
+            )
+            .unwrap();
+            self.resources.atlas_texture_array.size = WebGlTextureSize { width, height };
+            self.resources.atlas_layer_count = required_atlas_count;
+            // Cached FBOs were attached to the 1x1 stub store; drop them so we recreate on next use.
+            self.resources.atlas_render_framebuffer = None;
```
LaurenzV: This seems to be in the else branch as well, so we can just lift it out to the end I think?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1264,20 +1264,60 @@ impl WebGlPrograms {
...
+        let (width, height) = self.resources.atlas_size;
+
+        if current_atlas_count == 0 {
+            // Growing from the 1x1 placeholder allocated for `initial_atlas_count == 0`: there is no
+            // atlas data to preserve, so re-specify the backing store in place on the existing
+            // texture object rather than creating a new one and deleting the stub. Preserving the GL
+            // texture name is required on Mali-G52 under the Android WebView GL compositor: replacing
+            // the stub object mid-session hangs the native compositor, causing an ANR, and flushing
+            // around the swap does not help.
+            gl.bind_texture(
```
LaurenzV, line 1288: Would it be good to explicitly set the active texture unit?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1264,20 +1264,60 @@ impl WebGlPrograms {
...
+            self.resources.atlas_layer_count = required_atlas_count;
+            // Cached FBOs were attached to the 1x1 stub store; drop them so we recreate on next use.
+            self.resources.atlas_render_framebuffer = None;
+        } else {
+            // Growing an atlas that already holds data: allocate a new, larger `TEXTURE_2D_ARRAY`,
+            // copy the existing layers across, then swap. On Mali-G52 under the Android WebView GL
+            // compositor, `glFlush`ing after the copy is required before the next composite;
+            // otherwise, the compositor hangs and causes an ANR. Unbinding the atlas and flushing
+            // before allocation are precautionary.
+            gl.bind_texture(WebGl2RenderingContext::TEXTURE_2D_ARRAY, None);
```
LaurenzV, line 1315: This will only work if the currently active texture unit contains the 2D array, right? We probably should set `active_texture` explicitly as well?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1264,20 +1264,60 @@ impl WebGlPrograms {
...
+            gl.flush();
```
LaurenzV, line 1316: Is this flush necessary, too? Just wondering because you added a comment below, but not here.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: As far as I can tell, we bind to texture unit 0 when writing/copying to the 2D array, and texture unit 2 when sampling from it to the strip shader. Which texture unit is this `None` call supposed to affect?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Probably texture unit 0 to make it consistent with `write_to_atlas_layer`. Not sure.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Nevermind, it's mentioned further above. 🤦

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2223,8 +2263,27 @@ fn create_texture_inner(
     texture
 }
 
+/// Create a 1x1 RGBA32UI placeholder texture.
+///
+/// See [`create_placeholder_rgba8_texture`] for why the initial allocation matters.
```
LaurenzV: nit, but that should rather be a plain comment above `upload_data_to_rgba32_texture` instead of a doc comment on the method.

sparse_strips/vello_hybrid/CHANGELOG.md
```diff
@@ -22,6 +22,11 @@ This release has an [MSRV][] of 1.88.
...
 - Breaking change: `WebGlRenderer::render` and `WebGlRenderer::render_to_atlas` now take a `WebGlTextureBindings`, providing the external textures referenced by the scene. Pass `&WebGlTextureBindings::new()` if the scene does not use any.
 
+### Fixed
+
+- The alphas, encoded paints, gradient and filter data textures of the `webgl` backend are now allocated at creation. Previously a scene without alphas, encoded paints, gradients or filters left the corresponding texture incomplete, which made every draw call that bound it non-renderable and emitted a render warning per draw. (by [@grebmeg][])
+- The `webgl` image atlas now grows without hanging the Android WebView (Mali-G52) GL compositor. Growing from the lazily-allocated 1x1 placeholder (`initial_atlas_count == 0`) now re-specifies the backing store in place on the existing `TEXTURE_2D_ARRAY` instead of allocating a replacement texture and deleting the placeholder, and growing an atlas that already holds data unbinds the atlas and `glFlush`es around the reallocation. Previously, either could hang the compositor mid-session, causing an ANR. (by [@grebmeg][])
```
LaurenzV: That seems way too specific to me. Let's just merge both into a single sentence that we fixed a bug that could cause ANR on Mali-G52 GPUs? I don't think we need to expose all the details in a user-facing changelog.

LaurenzV (approved):
Thanks so much for looking into this!! It would still be interesting to do more digging into this, I still feel like we are a bit fuzzy on what _exactly_ causes the hand for those texture arrays, but we've already spent enough time on this. It's not great to have all these workarounds, but my hope here is that we can remove all of this once we move to external textures, anyway.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Good catch — lifted both `atlas_layer_count` and `atlas_render_framebuffer = None` out of the branches.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Good point!

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Good catch! I agree we should use better state hygiene here and explicitly specify which texture unit we’re modifying.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: The first flush is precautionary. In theory, it submits any queued image uploads before we copy the old atlas into its replacement. I didn’t observe any difference during testing, but I think it may still be useful. It doesn’t wait for completion, though the driver processes the commands in order. The final flush is the one experimentally required to prevent the Mali-G52 compositor hang.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Agreed

sparse_strips/vello_hybrid/CHANGELOG.md
grebmeg: Agreed — merged both entries into a single user-facing sentence.

PR #1802, opened 2026-08-06
Upgrade to fearless_simd v0.7 development state plus PR 300

Shnatsel:
Shows the impact of https://github.com/linebender/fearless_simd/pull/300 and other recent changes to fearless_simd

This is performance-neutral on x86 AVX2 on the Ghostscript tiger benchmarks.

AVX-512 is not comparable since main doesn't have AVX-512 support.

The new widen/narrow API is more boilerplate, but can work with hardware-width vectors such as `S::f32s` and can widen 512-bit vectors into two 512-bit vectors, which is inexpressible in the API that immediately combines the halves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1084,9 +1084,8 @@ dependencies = [
 
 [[package]]
 name = "fearless_simd"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76258897e51fd156ee03b6246ea53f3e0eb395d0b327e9961c4fc4c8b2fa151a"
+version = "0.6.0"
+source = "git+https://github.com/Shnatsel/fearless_simd.git?branch=load-interleaved-4-way#f62652d90cbc423dc967a5ea6b05158da388519a"
 dependencies = [
  "libm",
 ]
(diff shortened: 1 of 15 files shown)


LaurenzV:
NEON benchmarks are looking good as well.

T-256:
unrelated to this PR, but I couldn't find any discussion to make simd optional feature (no simd types in final binary), is there any?
should it configurable from vello_common side or fearless_simd  side?

I'm aware of fearless_simd support for [disable multiversioning](https://docs.rs/fearless_simd/latest/fearless_simd/#multiversioning-on-x86) but it doesn't help much for binary-size concerns:
> As a last resort, you can turn off multiversioning for specific SIMD instruction sets by passing --cfg disable_dispatch_sse4_2, --cfg disable_dispatch_avx2, or --cfg disable_dispatch_avx512 in RUSTFLAGS. These configuration flags only control automatic multiversioning. Disabling one does not remove its token type, its [Simd](https://docs.rs/fearless_simd/latest/fearless_simd/trait.Simd.html) implementation, or explicit [kernel](https://docs.rs/fearless_simd/latest/fearless_simd/macro.kernel.html) support; for example, an Avx2 token can still be used to call an AVX2 kernel when the CPU supports it.

LaurenzV:
Have you tried it? I'm pretty sure this should help with binary size, this was one of the main reasons to introduce this. It doesn't remove the tokens, but it should remove all instantiated functions used via the `dispatch` macro, which is what we use in vello.

LaurenzV:
Superseded by #1853.

PR #1803, opened 2026-08-06
vello_cpu: Fix horizontal extend mode for filtered images

LaurenzV:
Supersedes #1780, I was unable to push to the original branch due to some git LFS issue, so reopening it here.


grebmeg (approved):

sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -89,6 +89,25 @@ fn image_pad_x_pad_y(ctx: &mut impl Renderer) {
     repeat(ctx, Extend::Pad, Extend::Pad);
 }
 
+// TODO: Doesn't work correctly in Hybrid due to a difference in sampling behavior.
```
grebmeg, line 92: Yeah, the hybrid option doesn't look very good here

Image

PR #1804, opened 2026-08-06
sparse_strips: Bump to version 0.2.0

LaurenzV:
Was requested by the servo folks in https://github.com/linebender/vello/pull/1787#issuecomment-5173097081. Unfortunately can't be a patch bump because of already-landed breaking changes.

Release date is set for tomorrow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1353,7 +1353,7 @@ dependencies = [
 
 [[package]]
 name = "glifo"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "bytemuck",
  "core_maths",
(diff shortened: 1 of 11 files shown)


nicoburns (approved):
Seems fine.

Any chance we can get a Vello Classic release too this time around? That would unblock me from using the recent releases in Blitz/AnyRender.

My understanding is that the PR Bruce (@waywardmonkeys) wants (https://github.com/linebender/vello/pull/1664 I believe) fixes a bug that was already present in the previous release, so I think it probably shouldn't be blocking. And we can always do a fast-follow release like this one once it's ready. It's also a non-breaking change, so could be released in a patch release.

LaurenzV:
Fine from my side. @waywardmonkeys?

LaurenzV:
I forgot glifo depends on vello_common as well, so gotta bump that too.

PR #1805, opened 2026-08-07
vello_hybrid: Respect the `WEBGL_provoking_vertex` flag

LaurenzV:
See the comments, which should explain the rationale in enough detail. Below you can see data from a benchmark that compares the current `set_paint` against drawing with external texture rects.

Before: iPhone 12 struggles to keep 60 FPS with more than 10 images.
image

After: iPhone 12 can handle more thousands of external texture rects without performance tanking.
image


grebmeg (approved):
LGTM! Nice finding! 🎉

PR #1806, opened 2026-08-07
vello_hybrid: Set the `failIfMajorPerformanceCaveat` flag

LaurenzV:
In case the device is expected to be super slow on GPU (for example because it uses a software rasterizer behind the scenes), we would rather not run Vello Hybrid at all. In the future, maybe this should become a configurable option in case some users do desire this?

See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#failifmajorperformancecaveat


grebmeg (approved):
LGTM

I think we’ll need to add a `webglcontextcreationerror` listener either here or, ,maybe preferably, in the downstream consumer.

PR #1808, opened 2026-08-10
feat: allow disabling vello_hybrid depth buffers

grebmeg:
#### Summary

Allow applications to reduce GPU memory usage or operate without a reliable 24-bit depth buffer by disabling Vello Hybrid’s opaque-strip depth optimization.

Adds `RenderSettings::use_depth_buffer`, enabled by default to preserve existing behavior. When disabled:
- wgpu skips depth-texture allocation.
- WebGL requests a context without depth.
- All strips render in painter order with blending instead of early-Z occlusion.

#### Commits
1. Propagates the setting through scheduling and implements depth-disabled rendering in the wgpu and webgl backends.
2. Adds reusable no-depth test variants and regression coverage for painter ordering, transparency, AA edges, external textures, layers, filters, and blending.


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -586,11 +597,11 @@ impl WebGlRenderer {
         // See: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_invalidateframebuffer
         // We want to indicate to the GPU driver that we won't read the depth buffer again
         // until the next clear. This enables the GPU to avoid storing depth tiles back to VRAM.
```
LaurenzV: This comment should probably move down, no?

sparse_strips/vello_hybrid/src/render/wgpu/mod.rs
```diff
@@ -919,9 +924,9 @@ struct Programs {
     /// Root opaque-strip pipeline.
     opaque_strip_pipeline: RenderPipeline,
     /// Depth texture for early-z rejection on the Output target.
-    depth_texture: Texture,
+    depth_texture: Option<Texture>,
```
LaurenzV: By the way, i believe we don't even need to store the `depth_texture` here, just storing `depth_texture_view` should be enough (related to the proposal I want to make).

sparse_strips/vello_hybrid/src/schedule/test_support.rs
```diff
@@ -59,9 +59,30 @@ impl SceneCase {
         root_target: RootTarget,
         texture_size: SizeU16,
         max_textures: usize,
+    ) -> Result<ScheduledCase, RenderError> {
+        self.schedule_with_depth(root_target, texture_size, max_textures, true)
```
LaurenzV: Hmm... Maybe let's just change `schedule` to force the caller to pass in `use_depth_buffer` instead of creating `schedule_with_depth`? Yes it will require updating a few tests, but I think this is better, because it's not immediately obvious why `schedule` would always pass true here. Better to have each test be explicit, I think.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -111,6 +111,10 @@ impl Drawable for RecordedDraw {
 pub struct RenderSettings {
     /// The SIMD level that should be used for rendering operations.
     pub level: Level,
+    /// Whether to allocate and use a depth buffer for opaque-strip occlusion.
+    ///
+    /// Disabling this reduces GPU memory usage, but can increase overdraw.
+    pub use_depth_buffer: bool,
```
LaurenzV, line 117: In general, I agree that this would be the "cleanest" way to expose this to the user, so that the user just has to toggle a flag and it works across backends. However, since we are changing the behavior now, I would like to propose a different approach to handling depth buffers.

Remember that there is this issue https://github.com/linebender/vello/pull/1738. Basically, currently Vello Hybrid has the responsibility for managing the depth texture, but this makes it hard in cases as mentioned in the PR where the target ouput changes frequently. I would like to propose moving the responsibility of providing the depth texture to the user, such that Vello Hybrid doesn't have to deal with that, and the user can choose whatever policy they prefer for managing those. In particular, I think we should:

1) Remove the `use_depth_buffer` attribute from `RenderSettings`.
2) In the WebGL backend, simply add another `use_depth_buffer` boolean to the `render` method. Since the depth buffer is tied together with the main framebuffer, there isn't much to manage from our side. All the user can tell us is whether they'd like to enable it or not. The rest stays the same
3) In the `wgpu` backend, the `render` method instead takes an `Option<TextureView>`, representing the depth buffer. We just need to document that it needs to have the same size as the root output target and fulfill the other criteria of a depth buffer. In case it's `None`, we basically treat it as `use_depth_buffer = false`, in case it's some we simply use the texture view as the depth buffer. While this does put some more responsibility onto clients, it allows us not having to deal with the question of whether and how to cache depth textures, and therefore would allow us to also close https://github.com/linebender/vello/pull/1738, since the client can now do whatever they want.

What do you think?

sparse_strips/vello_sparse_tests/tests/hybrid_depth.rs
LaurenzV, line 1: Nit, but since we called it `hybrid_schedule` maybe we should call this `hybrid_depth`?

LaurenzV: On the other hand, we didn't do it for external textures, so I guess it doesn't really matter. 😄

LaurenzV (approved):
Nice, I agree this should be configurable! I do have a suggestion for what I think would be the better approach to implement this for `wgpu`, but if you disagree or would like to defer this to a follow-up, it's up to you!

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Agreed

sparse_strips/vello_hybrid/src/render/wgpu/mod.rs
grebmeg: Agreed

sparse_strips/vello_hybrid/src/schedule/test_support.rs
grebmeg: Agreed

sparse_strips/vello_sparse_tests/tests/hybrid_depth.rs
grebmeg: Good point, I renamed the module to `hybrid_depth`. I think we'll just need to rename `external_texture` later as well.

sparse_strips/vello_hybrid/src/scene.rs
grebmeg: I think it's a good idea! but I'd prefer to create it as a follow-up, as it would also require your review.

PR #1811, opened 2026-08-11
vello_hybrid: Add more descriptive message if WebGl2 context creation fails

LaurenzV:


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -238,7 +238,7 @@ impl WebGlRenderer {
         let gl = canvas
             .get_context_with_context_options("webgl2", &context_options)
             .expect("WebGL2 context to be available")
-            .unwrap()
+            .expect("WebGL2 context creation failed: This is possibly due to `failIfMajorPerformanceCaveat` being triggered")
```
grebmeg, line 241: Thanks for adding this! [as a follow-up] I think we have plenty of other places where we're unwrapping, and those places would also benefit from having error messages so we can monitor those failures through telemetry. On the other hand, we don't want to increase the WASM binary size. What do you think about using compact error codes here, and in those other places, instead of adding unique `expect` strings? That would let us distinguish failures in production while keeping verbose messages out of the WASM binary. JS or telemetry pipeline could then map each error code to a human-readable description if needed.

grebmeg (approved):

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Sure! But once again, I think that's better done in a refactor that improves error handling across the whole WebGL backend instead of just introducing it for some cases now. Will put this on my TODO list!

grebmeg: Sounds good. I agree it's better to do this as part of a broader refactor. That said, I think I'd like to have it done in a more consistent way across the whole hybrid.

PR #1812, opened 2026-08-11
vello_hybrid: Add support for setting a transparency hint for external textures

LaurenzV:
As per description. This requires making the opaque pass aware of external textures. An unfortunate consequence of this is that opaque external images that are not aligned to integer coordinates will need to be drawn twice: Once during the opaque pass, and another time for the fractional edges during the alpha pass. But at least they can now be drawn with depth occlusion, which should be a huge win.

### DIFF
--- a/sparse_strips/vello_hybrid/src/sampling.rs
+++ b/sparse_strips/vello_hybrid/src/sampling.rs
@@ -28,3 +28,21 @@ pub struct SampleRect {
     /// [`Self::source_region`].
     pub transform: Affine,
 }
+
+impl SampleRect {
+    /// Create a new [`SampleRect`].
+    pub fn new(source_region: RectU16, transform: Affine) -> Self {
+        Self {
+            source_region,
+            may_have_transparency: true,
+            transform,
+        }
+    }
+
+    /// Indicate that the sample rect only contains opaque pixels.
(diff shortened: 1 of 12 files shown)


sparse_strips/vello_hybrid/src/sampling.rs
```diff
@@ -13,6 +13,15 @@ pub struct SampleRect {
...
     pub source_region: RectU16,
 
+    /// Whether the sampled source region may contain non-opaque pixels.
+    ///
+    /// Only set this to `false` if every pixel is guaranteed to be opaque (e.g. in a texture
+    /// generated from a JPEG image). If you set this to `false` even though there are non-opaque
+    /// pixels, you will get wrong rendering.
+    ///
+    /// If unsure, always set this to `true`.
+    pub may_have_transparency: bool,
```
grebmeg, line 23: Should we add a constructor that defaults `may_have_transparency` to `true`, maybe with an explicit method for marking a region opaque?

grebmeg (approved):

sparse_strips/vello_hybrid/src/sampling.rs
LaurenzV: Sure!

PR #1813, opened 2026-08-11
vello_hybrid: Support extend modes for external texture rects

LaurenzV:
Based on #1812.

# Context
The main reason for opening this PR is that currently, the external texture rect API has no way of supporting other extend modes. This PR proposes tweaking the existing API for drawing external texture rects such that this use case can be accommodated.

# Implementation
It took me a while to figure out what the best way of implementing this is. Ultimately, I arrived at the conclusion that we should change the existing `draw_texture_rects` API to instead have a single `draw_texture_rect` API. I presume the original motivation for having the ability to submit multiple texture rectangles at once is performance. However, if you look at the actual implementation, the bulk of the work happens inside of the loop that iterates over each rectangles. Only very little work is lifted outside of the loop, so I don’t see any performance benefit in having a batched method for this. On the other hand, by making the API only take a single texture rect, we can make it more consistent with how other methods work and more easily add support for the missing features. For example, the `Scene::paint_transform` field now actually serves a purpose for external textures.

In order to get what we need, I also changed the API to taking two different rectangles: A source and a destination rectangle. The source rectangle defines what the area is we want to sample, and the destination (as well as path/paint transform) how it’s mapped into the actual scene. Any pixel that would fall outside of the source region but still be visible in the destination region is sampled according to the extend mode.

Also CC @tomcur since he designed the original API.


sparse_strips/vello_hybrid/src/sampling.rs
```diff
@@ -3,16 +3,29 @@
...
-pub struct SampleRect {
+pub struct ExternalTextureRect {
+    /// The external texture to sample.
+    pub texture_id: TextureId,
+
     /// Source region in texel coordinates.
     pub source_region: RectU16,
 
+    /// Destination rectangle in local scene coordinates.
+    pub destination_rect: Rect,
```
grebmeg: Could we rename these to `source_rect` and `destination_rect` for consistency, since both values are rectangles? `src_rect` and `dst_rect` would also be concise.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -307,19 +307,16 @@ impl Scene {
...
         let idx = self.encoded_paints.len();
-        let sampler = ImageSampler {
-            x_extend,
-            y_extend,
-            quality,
-            alpha: 1.0,
-        };
         let has_opacity = self
             .render_state
             .tint
```
grebmeg, line 322: I think we should support `ImageSampler::alpha` for external textures too, perhaps as a follow-up?

grebmeg (approved):

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Sure, but they currently aren’t supported for normal images either. So probably worth doing together.

sparse_strips/vello_hybrid/src/sampling.rs
LaurenzV: Yep!

LaurenzV: Ended up doing `dest` instead.

PR #1817, opened 2026-08-13
Clamp the discarded select arm's index in prefix-sum lookups

### LINKED ISSUE #1816: Device lost on Mali (Pixel 9) for any scene containing a path
Rendering any scene that contains at least one fill or stroke loses the Vulkan
device on a Pixel 9 Pro (Immortalis-G715, stock Android 16, driver v1.r54p2).
The device lost callback fires with "Device is lost" about 1.8s after the first
`render_to_texture`, which then takes the whole wgpu device down. An empty
scene renders fine, and so does `use_cpu: true`.

This reproduces headless: no surface, single thread, a 54x54 offscreen target,
vello 0.9 / wgpu 29, run as a plain executable from `adb shell`. A trivial
hand-written compute dispatch on the same device works, so it's not compute in
general.

I bisected it by compiling one stage at a time with
`ShaderRuntimeChecks::unchecked()` while keeping the rest checked: `tile_alloc`,
`backdrop_dyn` and `coarse` each fault on their own. Enabling only
`bounds_checks` fixes it while enabling only `force_loop_bounding` does not, so
it's a real out-of-bounds access rather than a loop miscompilation.

The access is this pattern, which appears in all three failing shaders (and
twice in `fine.wgsl`):

```wgsl
let tile_subix = select(0u, sh_tile_count[local_id.x - 1u], local_id.x > 0u);
```

Both arms of `select` are evaluated, so lane 0 computes
`sh_tile_count[0u - 1u]` and discards it. That's presumably harmless on
hardware with dedicated shared memory, but Mali lowers workgroup arrays to
ordinary memory, so the underflowed index turns into a wild address and the
driver kills the context. Nothing catches it since the shaders run unchecked
(#1093).

Clamping the index in the discarded arm (`sh_tile_count[max(local_id.x, 1u) -
1u]`) makes all my previously-failing scenes render correctly with fully
unchecked shaders. I'll open a PR with that change.

Possibly the same root cause as #1465, which also went away once bounds checks
were enabled.

lexoliu:
Fixes #1816.

Several prefix-sum lookups have the shape
`select(default, sh_array[ix - 1u], ix > 0u)`. Both arms of `select` are
evaluated, so lanes with `ix == 0` compute `sh_array[0u - 1u]` before
discarding it. Mali lowers workgroup arrays to ordinary memory, so the
underflowed index becomes a wild address and the driver reports a device loss;
on a Pixel 9 Pro (Immortalis-G715, Android 16) any scene containing a path
lost the device on the first frame. The shaders run unchecked since #1093, so
there is no bounds check to mask it.

This clamps the index in the discarded arm, `sh_array[max(ix, 1u) - 1u]`,
which changes nothing when `ix > 0` and keeps the dead access in bounds when
`ix == 0`. Six sites: one each in `tile_alloc` and `backdrop_dyn`, two in
`coarse` (these four fault on the device), and two in `fine` which have the
same pattern but didn't fault in my scenes.

Verified on the Pixel 9 Pro with fully unchecked shaders: the scenes that
previously lost the device render correctly. `vello_tests` passes on Metal, so
the selected values are unchanged. An explicit `if` would avoid the dual
evaluation altogether; I kept the clamp since it's the smallest diff and it's
what I tested on hardware, but happy to change the shape.

### DIFF
--- a/vello_shaders/shader/backdrop_dyn.wgsl
+++ b/vello_shaders/shader/backdrop_dyn.wgsl
@@ -73,7 +73,7 @@ fn main(
         }
         let width = sh_row_width[el_ix];
         if width > 0u {
-            var seq_ix = row - select(0u, sh_row_count[el_ix - 1u], el_ix > 0u);
+            var seq_ix = row - select(0u, sh_row_count[max(el_ix, 1u) - 1u], el_ix > 0u);
             var tile_ix = sh_offset[el_ix] + seq_ix * width;
             var sum = tiles[tile_ix].backdrop;
             for (var x = 1u; x < width; x += 1u) {
--- a/vello_shaders/shader/coarse.wgsl
+++ b/vello_shaders/shader/coarse.wgsl
@@ -248,7 +248,7 @@ fn main(
                         part_ix = probe;
                     }
                 }
-                ix -= select(part_start_ix, sh_part_count[part_ix - 1u], part_ix > 0u);
+                ix -= select(part_start_ix, sh_part_count[max(part_ix, 1u) - 1u], part_ix > 0u);
                 let offset = config.bin_data_start + sh_part_offsets[part_ix];
                 sh_drawobj_ix[local_id.x] = info_bin_data[offset + ix];
             }
@@ -312,7 +312,7 @@ fn main(
             }
             drawobj_ix = sh_drawobj_ix[el_ix];
             tag = scene[config.drawtag_base + drawobj_ix];
-            let seq_ix = ix - select(0u, sh_tile_count[el_ix - 1u], el_ix > 0u);
+            let seq_ix = ix - select(0u, sh_tile_count[max(el_ix, 1u) - 1u], el_ix > 0u);
             let width = sh_tile_width[el_ix];
             let x0y0 = sh_tile_x0y0[el_ix];
             let x = (x0y0 & 0xffffu) + seq_ix % width;
--- a/vello_shaders/shader/fine.wgsl
+++ b/vello_shaders/shader/fine.wgsl
@@ -220,7 +220,7 @@ fn fill_path_ms(fill: CmdFill, local_id: vec2<u32>, result: ptr<function, array<
             }
             let el_ix = lo;
             let last_pixel = i + 1u == sh_count[el_ix];
-            let sub_ix = i - select(0u, sh_count[el_ix - 1u], el_ix > 0u);
+            let sub_ix = i - select(0u, sh_count[max(el_ix, 1u) - 1u], el_ix > 0u);
             let seg_off = fill.seg_data + batch * WG_SIZE + el_ix;
             let segment = segments[seg_off];
             // Coordinates are relative to tile origin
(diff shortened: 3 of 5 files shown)


vello_shaders/shader/backdrop_dyn.wgsl
```diff
@@ -73,7 +73,7 @@ fn main(
         }
         let width = sh_row_width[el_ix];
         if width > 0u {
-            var seq_ix = row - select(0u, sh_row_count[el_ix - 1u], el_ix > 0u);
+            var seq_ix = row - select(0u, sh_row_count[max(el_ix, 1u) - 1u], el_ix > 0u);
```
b0nes164, line 76: Nit: It is redundant to both `max` and `select` here. The issue is that, on the Vulkan backend, the code is compiling to an  `OpSelect` which evaluates both sides of the ternary and discards the false result, so the `select` is not providing the intended protection. 

Instead you should branch with an `if`, which compiles to an `OpBranchConditional`. This should be less depth than `select` + `max`
```
var seq_ix = 0u;
if (el_ix > 0u) {
   . . .
}
```

vello_shaders/shader/coarse.wgsl
```diff
@@ -248,7 +248,7 @@ fn main(
                         part_ix = probe;
                     }
                 }
-                ix -= select(part_start_ix, sh_part_count[part_ix - 1u], part_ix > 0u);
+                ix -= select(part_start_ix, sh_part_count[max(part_ix, 1u) - 1u], part_ix > 0u);
```
b0nes164, line 251: Here too

vello_shaders/shader/coarse.wgsl
```diff
@@ -312,7 +312,7 @@ fn main(
             }
             drawobj_ix = sh_drawobj_ix[el_ix];
             tag = scene[config.drawtag_base + drawobj_ix];
-            let seq_ix = ix - select(0u, sh_tile_count[el_ix - 1u], el_ix > 0u);
+            let seq_ix = ix - select(0u, sh_tile_count[max(el_ix, 1u) - 1u], el_ix > 0u);
```
b0nes164, line 315: Here too

vello_shaders/shader/fine.wgsl
```diff
@@ -220,7 +220,7 @@ fn fill_path_ms(fill: CmdFill, local_id: vec2<u32>, result: ptr<function, array<
             }
             let el_ix = lo;
             let last_pixel = i + 1u == sh_count[el_ix];
-            let sub_ix = i - select(0u, sh_count[el_ix - 1u], el_ix > 0u);
+            let sub_ix = i - select(0u, sh_count[max(el_ix, 1u) - 1u], el_ix > 0u);
```
b0nes164, line 223: Here too

vello_shaders/shader/fine.wgsl
```diff
@@ -576,7 +576,7 @@ fn fill_path_ms_evenodd(fill: CmdFill, local_id: vec2<u32>, result: ptr<function
...
```
b0nes164, line 579: Here too

vello_shaders/shader/tile_alloc.wgsl
```diff
@@ -104,7 +104,7 @@ fn main(
     let tile_offset = paths[drawobj_ix | (WG_SIZE - 1u)].tiles;
     storageBarrier();
     if drawobj_ix < config.n_drawobj {
-        let tile_subix = select(0u, sh_tile_count[local_id.x - 1u], local_id.x > 0u);
+        let tile_subix = select(0u, sh_tile_count[max(local_id.x, 1u) - 1u], local_id.x > 0u);
```
b0nes164, line 107: Here too

b0nes164 (changes requested):

b0nes164:
Hi @lexoliu, just checking in, are you still planning on making this change? I can take over it if you are too busy.

PR #1818, opened 2026-08-13
refactor: split WESL shaders into helper modules

grebmeg:
Extract reusable WESL code for blending, gradients, images, alpha handling, and utilities into `helpers` submodules. Update root shaders to import these helpers, reducing duplication while preserving rendering behavior.

The `helpers` name avoids WGSL reserved words (`shared` and `common`) and supports `wgsl-analyzer` import resolution.


sparse_strips/vello_sparse_shaders/shaders/helpers/image.wesl
```diff
@@ -0,0 +1,188 @@
+// Copyright 2026 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Shared encoded image accessors and atlas sampling (bilinear/bicubic) helpers.
+
+/// Tint mode constants.
+const TINT_MODE_ALPHA_MASK: u32 = 0u;
+const TINT_MODE_MULTIPLY: u32 = 1u;
+
+// Encoded image layout. Must match `GpuEncodedImage` in `vello_hybrid/src/render/common.rs`.
```
LaurenzV, line 19: Not for this PR, but in the future I'd also like to propose for us to define bit layouts only in one location (either on the Rust side or on the shader side). I think it's a bit annoying having to keep the two up-to-date.

sparse_strips/vello_sparse_shaders/shaders/blend.wesl
LaurenzV: I think that computation is used everywhere, is that also something that could be extracted?

LaurenzV: Something like

```
fn quad_corner(vertex_index: u32) -> vec2<f32> {
    return vec2<f32>(
        f32(vertex_index & 1u),
        f32(vertex_index >> 1u),
    );
}

fn pixel_to_ndc(
    pixel: vec2<f32>,
    target_size: vec2<f32>,
) -> vec2<f32> {
    return vec2<f32>(
        pixel.x * 2.0 / target_size.x - 1.0,
        1.0 - pixel.y * 2.0 / target_size.y,
    );
}
```

maybe?

sparse_strips/vello_sparse_shaders/shaders/blend.wesl
```diff
@@ -143,13 +109,7 @@ fn fs_main(
             vec2<u32>(texel_index % texture_width, texel_index / texture_width),
```
LaurenzV: How about a 

```
fn flat_index_to_texture_coord(index: u32, width: u32) -> vec2<u32> {
    return vec2(index % width, index / width);
}
```

method? Same pattern is used for gradients anfd filters I think.

sparse_strips/vello_sparse_shaders/shaders/filter.wesl
LaurenzV, line 1: Doesn't have to be now, but do you agree it might be nice to create a new `filters` directory with one filter pass per file? If so, maybe let's add a TODO to track this?

sparse_strips/vello_sparse_shaders/shaders/render.wesl
LaurenzV, line 1: There's some more places we can use `unpack_u16_pair`:

Image

sparse_strips/vello_sparse_shaders/shaders/render.wesl
```diff
@@ -559,129 +570,8 @@ fn load_encoded_paint_texel(paint_tex_idx: u32, texel_offset: u32) -> vec4<u32>
...
-    let b = textureLoad(tex, vec2<i32>(uv_quad.xw), atlas_idx, 0);
-    let c = textureLoad(tex, vec2<i32>(uv_quad.zy), atlas_idx, 0);
-    let d = textureLoad(tex, vec2<i32>(uv_quad.zw), atlas_idx, 0);
-    return mix(mix(a, b, uv_frac.y), mix(c, d, uv_frac.y), uv_frac.x);
-}
-
-// This is the same as `bilinear_sample` above, but for external textures instead of the atlas texture array.
+// This is the same as `helpers::image::bilinear_sample`, but for the external texture binding
+// instead of the atlas texture array.
 fn external_bilinear_sample(
```
LaurenzV: Why not move all the external texture stuff into a  `external_texture` helper module?

LaurenzV (approved):
Love it!!!! Thanks a lot!

sparse_strips/vello_sparse_shaders/shaders/helpers/image.wesl
grebmeg: Do you mean defining the layout in a shared schema and generating both the Rust packing code and WESL accessors from it at build time? I don’t think Rust and WESL can otherwise share the same constants directly.

sparse_strips/vello_sparse_shaders/shaders/blend.wesl
grebmeg: Good idea! I extracted `quad_corner` and `pixel_to_ndc` into a shared `geometry` helper and updated all five shaders that duplicated these calculations.

sparse_strips/vello_sparse_shaders/shaders/blend.wesl
grebmeg: Done.

sparse_strips/vello_sparse_shaders/shaders/filter.wesl
grebmeg: Good suggestion! I split the substantial filter logic into focused `scale`, `blur`, and `drop_shadow` modules. I kept the tiny copy, `flood`, and `offset` operations in the root dispatch to avoid creating one-line modules.

sparse_strips/vello_sparse_shaders/shaders/render.wesl
grebmeg: Done. Thanks!

sparse_strips/vello_sparse_shaders/shaders/render.wesl
grebmeg: Done! I moved the external texture sampling logic into a dedicated `external_texture` helper module.

grebmeg (commented):
Thanks for the review, @LaurenzV! ❤️ Everything looks much better now!!

sparse_strips/vello_sparse_shaders/shaders/helpers/image.wesl
LaurenzV: Nono, I just mean not having comments that document the layout in two places.

PR #1819, opened 2026-08-13
[DO NOT MERGE] vello_hybrid: Prohibit using integers directly in vertex shader inputs

LaurenzV:
**Note that this PR should not be merged because I don't think it's fully sound. However, just opening it for visibility.**

Another day, another round of ugly workarounds we need to apply to work around bugs. From my testing, this seems to affect various MacBooks with Intel HD Graphics 4000 GPUs, on a specific but considerable subset of older Chrome browsers. The latest version I was able to reproduce this was on macOS Catalina with Chrome 128, but earlier OS versions are also affected. 

It seems very likely that this was somehow caused by https://github.com/chromium/chromium/commit/1e1b3403de1fd2ea302469ab402719166bbb9cad, because it matches my observation on Chrome versions affected on the different MacOS versions. But who exactly is at fault here is not known to me.

A similar StackOverflow post where this is happening is in https://stackoverflow.com/questions/6578026/cant-get-integer-vertex-attributes-working-in-glsl-1-5.

# Problem

The bug can be reproduced with the following example:

```html
<!doctype html>
<canvas style="width:100vw;height:100vh;display:block"></canvas>
<script>

precision highp int;

layout(location = 0) in uint input_value;
flat out vec4 color;

void main() {
    vec2 positions[3] = vec2[3](
        vec2(-1.0, -1.0),
        vec2( 3.0, -1.0),
        vec2(-1.0,  3.0)
    );

    if (input_value == 0xffffffffu) {
        color = vec4(0.0, 1.0, 0.0, 1.0);
    } else if (input_value == 0x4f800000u) {
        color = vec4(1.0, 0.0, 0.0, 1.0);
    } else {
        color = vec4(0.0, 0.0, 1.0, 1.0);
    }

    gl_Position = vec4(positions[gl_VertexID], 0.0, 1.0);
}
`;

precision highp float;

flat in vec4 color;
layout(location = 0) out vec4 out_color;

void main() {
    out_color = color;
}
`;

  function shader(type, source) {
    gl.compileShader(shader);
    return shader;
  }

  gl.attachShader(program, shader(gl.VERTEX_SHADER, vertexSource));
  gl.attachShader(program, shader(gl.FRAGMENT_SHADER, fragmentSource));
  gl.linkProgram(program);
  gl.useProgram(program);

  gl.bindBuffer(gl.ARRAY_BUFFER, gl.createBuffer());
  // Working alternatives:
  // - Disable the array and use vertexAttribI4ui(0, 0xffffffff, 0, 0, 0).
  // - Upload the bits as FLOAT, use vertexAttribPointer, and recover them in
  //   the shader with floatBitsToUint.
  // - Store the uint in a std140 UNIFORM_BUFFER.
  gl.bufferData(
    gl.ARRAY_BUFFER,
    new Uint32Array([0xffffffff]),
    gl.STATIC_DRAW
  );
  gl.enableVertexAttribArray(0);
  gl.vertexAttribIPointer(0, 1, gl.UNSIGNED_INT, 0, 0);
  gl.vertexAttribDivisor(0, 1);

  canvas.width = innerWidth;
  canvas.height = innerHeight;
  gl.drawArraysInstanced(gl.TRIANGLES, 0, 3, 1);
</script>
```

On affected devices, this will output red instead of green. Basically, when uploading integers via buffers for vertex attributes, it seems like on the affected devices, it goes through a int -> float conversion and then just returns the bit pattern of the resulting float. Therefore, we basically always get complete garbage numbers. 

# Solution

What does seem to work is:
- Setting vertex attributes as constants.
- Bitcasting into float and then bitcasting back into integer in the shader.
- Passing integers from the vertex shader to the fragment shader.

For our use case, it seemed to me like the second is the best approach, since it should essentially be free (we are just bitcasting, not converting) and also is easily enforceable with a lint. I also confirmed that with this fix, Vello Hybrid runs successfully on those devices.

# Caveat
Unfortunately, the biggest problem with this PR is that it assumes that floating point values are always preserved with bit-exactness, so for example subnormals must not be flushed to 0 and NaNs must always be preserved. Unfortunately, I don't think there is any guarantee that this is the case on all GPUs. Since corrupting the bit pattern would have fatal consequences, I don't think we can merge this PR in its current form, unfortunately.

CC @grebmeg


LaurenzV:
**Update**: It turns out that the bug is a bit more specific. It only breaks when using `drawArraysIntanced`, but it works with `drawElementsInstanced`. The following diff make everything work on those devices: https://github.com/linebender/vello/compare/main...laurenz/elements


When I open https://registry.khronos.org/webgl/sdk/tests/conformance2/rendering/attrib-type-match.html on the device, I also only get 2 test failures, which seems to confirm this:
image


However, I'm not sure whether there is a performance penalty when rendering this way.

grebmeg:
> **Update**: It turns out that the bug is a bit more specific. It only breaks when using `drawArraysIntanced`, but it works with `drawElementsInstanced`.

I'd expect that switch to be performance-neutral. The cost of the draw call is mostly dominated by fragment shading and, on webgl, the cpu-side driver overhead for each draw call. I also expect reads from the `ELEMENT_ARRAY_BUFFER` to be fully cache-friendly in this case. One nice benefit of using `drawElementsInstanced` is that it makes it explicit that the indices are `UNSIGNED_SHORT`, so we "avoid" 🤞 any implicit unnecessary floating-point conversions.

If we want to be extra confident, we could run a before/after benchmark on a low-end device. If everything looks good, I'd be happy to go with `drawElementsInstanced`. What do you think?

LaurenzV:
Sounds good, will run some experiments.

LaurenzV:
Will be superseded by #1827.

PR #1821, opened 2026-08-13
vello_hybrid: Prohibit dynamic indexing

LaurenzV:
On macOS Monterey on Safari 15.6, filters currently render like this:

https://github.com/user-attachments/assets/1c2fc6cf-77cf-41ec-a3b5-a0d894e78614

After experimenting, I reduced it to the following code snippet, which shows up as red on the given device (though depending on how you change it, you can also produce a non-deterministic version, which is the reason why the artifacts seem to change in each frame).

```html
<!doctype html>
<meta name="viewport" content="width=device-width, initial-scale=1">
<style>
  html, body, canvas {
    width: 100%;
    height: 100%;
    margin: 0;
    display: block;
  }
</style>

<canvas></canvas>

<script>

void main() {
    vec2 position = vec2((gl_VertexID << 1) & 2, gl_VertexID & 2);
    gl_Position = vec4(position * 2.0 - 1.0, 0.0, 1.0);
}
`;

precision highp float;
precision highp int;

uniform int index;
out vec4 color;

void main() {
    // Working alternatives:
    // - Store the values in a vec3 and index that dynamically.
    // - Select values[0], values[1], or values[2] with explicit branches.
    // - Perform the same runtime array lookup in the vertex shader and pass
    //   the result to this shader as a flat varying.
    float values[3] = float[3](0.2, 0.5, 1.0);
    bool passed = values[index] == 1.0;
    color = passed ? vec4(0.0, 1.0, 0.0, 1.0)
                   : vec4(1.0, 0.0, 0.0, 1.0);
}
`;

function compile(type, source) {
  gl.compileShader(shader);
  return shader;
}

gl.attachShader(program, compile(gl.VERTEX_SHADER, vertexSource));
gl.attachShader(program, compile(gl.FRAGMENT_SHADER, fragmentSource));
gl.linkProgram(program);
gl.useProgram(program);
gl.uniform1i(gl.getUniformLocation(program, "index"), 2);

canvas.width = innerWidth;
canvas.height = innerHeight;
</script>

```

I didn't look into this even further, but clearly for some reason dynamic indexing doesn't work correctly. In this case the index is passed as a uniform, but it seems to more generally happen with any "dynamic" index (like in our for loop). Since dynamic indexing has caused other problems before (see the existing comments), let's simply unroll the loop. With this, filter render correctly on that device.


LaurenzV:
Let me know if the lint is too strict and if I should remove it. I'm not sure if it's worth completely banning them, as I'm sure many kinds of loops will work fine (for example, we could restrict it to only dynamic loops). But I think after all these bugs, it's becoming clear that we need to keep our shaders as simple as possible, and it does seem like loops (or at least things like dynamic indexing) are clearly not considered "simple" features. 😄

grebmeg:
> On macOS Monterey on Safari 15.6, filters currently render like this:

> After experimenting, I reduced it to the following code snippet, which shows up as red on the given device (though depending on how you change it, you can also produce a non-deterministic version, which is the reason why the artifacts seem to change in each frame).

~~Hmm, I'm trying to reproduce this, but I haven't been able to. Is there anything else I need to do?~~

~~I'm using macOS Monterey 12.7.5 with Safari 15.6.1, and it's showing green in the snippet you shared above.~~

I was able to repro

grebmeg:
> Let me know if the lint is too strict and if I should remove it. I'm not sure if it's worth completely banning them, as I'm sure many kinds of loops will work fine (for example, we could restrict it to only dynamic loops). But I think after all these bugs, it's becoming clear that we need to keep our shaders as simple as possible, and it does seem like loops (or at least things like dynamic indexing) are clearly not considered "simple" features. 😄

Would it be possible to write a lint rule that detects dynamic indexing instead? That's the actual issue, and it can occur even without a loop.

This looks like a Safari/WebKit shader compiler bug. I found several related reports:

- [WebKit #234926](https://bugs.webkit.org/show_bug.cgi?id=234926): Safari 15 on macOS 12 fails when dynamically indexing fixed-size arrays with the Metal backend.
- [WebKit #214393](https://bugs.webkit.org/show_bug.cgi?id=214393): dynamic vector indexing produced an incorrect result — red instead of green.
- [WebKit #223349](https://bugs.webkit.org/show_bug.cgi?id=223349): dynamic matrix indexing produces incorrect results on macOS.

Using `values[2]` works because the compiler knows the index while compiling. Using `values[index]` requires a runtime lookup, which Safari appears to compile incorrectly?

grebmeg (approved):
I don’t mind unrolling this loop, since texture sampling is the main bottleneck here. However, I’m concerned that we’re prematurely banning all loops. Could we make the lint less strict, perhaps emitting a warning instead? Ideally, I’d prefer to lint dynamic indexing directly, if possible.

LaurenzV:
Yep, will change the lint to check for dynamic access instead.

PR #1822, opened 2026-08-13
Don't enable WGPU in std

taj-p:
Today, it's impossible to enable `std` and `webgl` without seeing this error: "Both WebGL and wgpu with the "webgl" feature are enabled" because WGPU is enabled via `std`.

### DIFF
--- a/sparse_strips/vello_hybrid/Cargo.toml
+++ b/sparse_strips/vello_hybrid/Cargo.toml
@@ -48,7 +48,7 @@ roxmltree = "0.20.0"
 
 [features]
 default = ["wgpu", "wgpu_default", "text"]
-std = ["wgpu", "vello_common/std", "glifo?/std"]
+std = ["vello_common/std", "glifo?/std"]
 libm = ["vello_common/libm", "glifo?/libm"]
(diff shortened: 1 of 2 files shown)


LaurenzV:
Unfortunately it’s rather annoying to fix. :( I think I’ve opened a Zulip thread about this before.

taj-p:
\

> Unfortunately it’s rather annoying to fix. :( I think I’ve opened a Zulip thread about this before.

Does the current impl work? I'm not trying to get `no_std` to work, but instead trying to compile vello_hybrid with webgl and std enabled without wgpu enabled at the same time. i.e., I want to avoid the below warning:

https://github.com/linebender/vello/blob/08576de9df7d4b8f4108fe4fd5a436bd69146c43/sparse_strips/vello_hybrid/src/render/common.rs#L625-L638

LaurenzV:
Sorry haha I think my brain was too fried already 😆 I’ll check in the morning!

grebmeg (commented):
I didn't look into this in much detail, overall LGTM if LGT @LaurenzV. I just wanted to mention that I had the same thought yesterday while commuting home, that we should remove this message reporting.

LaurenzV (approved):
Thanks!!

PR #1823, opened 2026-08-14
feat: add non-blocking WebGL shader compilation

grebmeg:
Adds pollable WebGL renderer initialization using `KHR_parallel_shader_compile`, allowing shader compilation and program linking to proceed without blocking the browser’s main thread.

Existing synchronous constructors remain compatible, and browsers without extension support automatically fall back to blocking initialization.

Extension adoption ~73% overall support: [Web3D Survey — KHR_parallel_shader_compile](https://web3dsurvey.com/webgl2/extensions/KHR_parallel_shader_compile)

A video comparing blocking and non-blocking renderer initialization:

https://github.com/user-attachments/assets/94b078f2-dca9-4ad2-a834-6deeb7126770

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -1249,9 +1249,6 @@ impl WebGlIntermediateTexture {
     }
 }
 
-/// `COMPLETION_STATUS_KHR` from the `KHR_parallel_shader_compile` specification.
-const COMPLETION_STATUS_KHR: u32 = 0x91B1;
-
 #[derive(Debug)]
 struct PendingShaderProgram {
     vertex_shader: VertexShader,
@@ -1281,7 +1278,12 @@ impl PendingShaderProgram {
         }
     }
 
-    fn is_complete(&self, gl: &WebGl2RenderingContext) -> bool {
+    fn is_complete(&self, gl: &WebGl2RenderingContext, parallel_shader_compile: bool) -> bool {
+        /// `COMPLETION_STATUS_KHR` from the `KHR_parallel_shader_compile` specification.
+        const COMPLETION_STATUS_KHR: u32 = 0x91B1;
+        if !parallel_shader_compile {
+            return true;
+        }
         gl.get_program_parameter(&self.program, COMPLETION_STATUS_KHR)
             .as_bool()
             .unwrap_or(false)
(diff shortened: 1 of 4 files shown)


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1166,28 +1249,119 @@ impl WebGlIntermediateTexture {
     }
 }
 
-impl WebGlPrograms {
-    /// Creates programs and initializes resources.
+/// `COMPLETION_STATUS_KHR` from the `KHR_parallel_shader_compile` specification.
+const COMPLETION_STATUS_KHR: u32 = 0x91B1;
```
LaurenzV: Can we put this in the method where it's used?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1166,28 +1249,119 @@ impl WebGlIntermediateTexture {
...
+
+        Self {
+            vertex_shader,
+            fragment_shader,
+            program,
+        }
+    }
+
+    fn is_complete(&self, gl: &WebGl2RenderingContext) -> bool {
+        gl.get_program_parameter(&self.program, COMPLETION_STATUS_KHR)
```
LaurenzV, line 1287: nit: This method must not be called if the extension is not available, right? Could we maybe change it so that the method takes the `parallel_shader_compile` as a boolean and the short-circuiting happens there? Yes this means we have 4 boolean checks instead of just one, but it's basically nothing and it makes it harder to accidentally misuse the method if the extension is not available. WDYT?

LaurenzV (approved):
LGTM! Unfortunately, I tried this and it doesn't seem like the extension is available on many Android devices (I haven't found any at all). :( But it's still good to have for devices that do support it.

Also, the fact that we are know compiling + linking all shaders first before checking the status also seems to improve the times for the blocking case (and also is the recommended way of doing it, it seems).

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Done

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
grebmeg: Done

grebmeg:
> LGTM! Unfortunately, I tried this and it doesn't seem like the extension is available on many Android devices (I haven't found any at all). :( But it's still good to have for devices that do support it.

Yeah, unfortunately all Android devices (+ Firefox browser), including most Chromebooks, are out of scope for this feature:

image


image

PR #1824, opened 2026-08-14
vello_hybrid: Integrate external textures into the existing `set_paint` API

LaurenzV:
After #1813, I realized there really isn't any reason anymore to keep external textures as a separate API limited to rectangles. Instead, we can just introduce a new external texture paint type, and then just paint any shape with this. This reduces the API surface and makes it more consistent with the existing API. This also makes it easier to switch between the two image upload modes at runtime, which is good for us because we want to do more experimentation.

I feel like there is room for more simplification by the external image ID and this new paint type, but for now this is already a step in the right direction, I think.


sparse_strips/vello_sparse_tests/tests/external_texture.rs
```diff
@@ -46,46 +46,56 @@ mod tests {
...
+    fn external_texture_image(
+        texture_id: TextureId,
+        source_region: RectU16,
+        sampler: ImageSampler,
+        may_have_transparency: bool,
+    ) -> Image {
+        Image {
+            image: ImageSource::external_texture(texture_id, source_region, may_have_transparency),
+            sampler,
+        }
```
grebmeg: Could we inline this instead? It's a really small helper, and we already inline similar `Image` usage elsewhere.

sparse_strips/vello_hybrid/src/scene.rs
```diff
@@ -490,39 +453,6 @@ impl Scene {
         });
     }
```
grebmeg, line 455: Would it be better not to add the external texture paint to `encoded_paints` when the geometry is a no-op?

```rust
 #[test]
fn zero_area_external_texture_rect_is_a_noop() {
    let mut scene = Scene::new(100, 100);
    scene.set_paint(Image {
        image: ImageSource::external_texture(TextureId(7), RectU16::new(0, 0, 8, 8), false),
        sampler: ImageSampler::default(),
    });

    scene.fill_rect(&Rect::new(10.0, 10.0, 10.0, 20.0));

    assert!(scene.encoded_paints.is_empty());
    assert!(scene.strip_storage.borrow().strips.is_empty());
    assert!(scene.recorder.draws.is_empty());
    assert!(scene.recorder.nodes.is_empty());
}
```

grebmeg (approved):
This is amazing! It looks so much better now!

sparse_strips/vello_sparse_tests/tests/external_texture.rs
LaurenzV: Yep, will inline.

sparse_strips/vello_hybrid/src/scene.rs
LaurenzV: Makes sense!

LaurenzV: Though I guess this is not about external textures now, but really all rectangles that are drawn with zero-area, right? Regardless of which complex paint. So something like this:

```rs
    pub fn fill_rect(&mut self, rect: &Rect) {
        if !self.paint_visible || rect.is_zero_area() {
            return;
        }
```

instead of just 

```rs
    pub fn fill_rect(&mut self, rect: &Rect) {
        if !self.paint_visible {
            return;
        }
```

LaurenzV: Will add a TODO instead since it's not really related to the change of this PR.

LaurenzV: Oh, I see now, this did exist before for external textures, hence why you pointed it out. Will add it.

grebmeg: Thank you! 🙏

PR #1825, opened 2026-08-14
vello_common: Remove the `allocate_excluding` API

LaurenzV:
This was needed in the past when filter layers were rendered into the image atlas. Now that we have our new layer scheduler, I don't think we are going to need this anymore. So I propose removing it to simplify the code.


grebmeg (approved):
Nice! Thanks for cleaning that up.

PR #1826, opened 2026-08-14
vello_hybrid: Use `u16` consistently throughout the codebase

LaurenzV:
Currently, Vello Hybrid is very inconsistent in its use of data types for storing sizes and various rectangles. Because of this, there's a lot of code that converts between the two, and we essentially have to duplicate each data structure for both, u16 and u32.

The current sparse strips architecture (in `vello_common`) does not support targets larger than u16, due to coordinates being stored this way to improve performance and reduce memory consumption. In addition to that, most GPUs don't support render targets larger than 16k anyway. Therefore, I propose that we use u16 consistently throughout the code and only switch to u32 on demand 1) in the WebGL/WebGPU backend because it's required by them and 2) at the boundary between Vello and guillotierre, since the crate also uses u32 coordinates. In all other parts of the codebase, we stick to u16.

This PR was done with assistence of GPT 5.6 Sol.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -178,9 +178,9 @@ impl WebGlTextureBindings {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct AtlasTextureInfo {
     /// Width of the currently bound texture array.
-    pub width: u32,
+    pub width: u16,
     /// Height of the currently bound texture array.
-    pub height: u32,
+    pub height: u16,
     /// Number of real atlas layers available to the image cache.
     pub layer_count: u32,
 }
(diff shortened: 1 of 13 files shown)


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -967,14 +968,12 @@ impl WebGlRenderer {
...
-    ];
-    let width = u32::from(rect.width) + padding * 2;
-    let height = u32::from(rect.height) + padding * 2;
-    renderer.clear_atlas_region(AtlasId::new(rect.page_index), offset, width, height);
+    renderer.clear_atlas_region(
+        AtlasId::new(rect.page_index),
+        [rect.x, rect.y],
+        rect.width,
+        rect.height,
+    );
```
LaurenzV, line 976: Turns out that this was actually a bug on current main. The pending clear rect already contains glyph padding:
https://github.com/linebender/vello/blob/ea474c2014f1ace419b1f18792f9a70865925147/glifo/src/atlas/cache.rs#L439-L447

So we were essentially applying it twice here. The wgpu backend doesn't have this issue.

grebmeg (approved):
I think this makes sense for now: use `u16` for atlas dimensions internally and widen to `u32` only at API or arithmetic boundaries. LGTM!

PR #1827, opened 2026-08-14
vello_hybrid: Use `drawElementsInstanced` instead of `drawArraysInstanced` for drawing strips

LaurenzV:
See https://github.com/linebender/vello/pull/1819 for more info.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -1145,6 +1145,8 @@ struct StripUniforms {
 /// Contains all WebGL resources needed for rendering.
 #[derive(Debug)]
 pub(crate) struct WebGlResources {
+    /// Shared indices for four-vertex triangle-strip draws.
+    quad_index_buffer: Buffer,
     /// VAO for strip rendering.
     strip_vao: VertexArray,
     /// Buffer for [`GpuStrip`] data.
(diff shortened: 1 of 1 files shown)


LaurenzV:
@grebmeg Tried this via vello bench on various devices by drawing many small, differently-sized rectangles, and dynamically toggling between the two modes. Didn't notice any difference in FPS whatsoever. I think this should be safe to land! 🙂

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2786,6 +2807,25 @@ struct WebGlRendererContext<'a> {
 }
 
 impl WebGlRendererContext<'_> {
+    /// Draw `instance_count` quad instances using the currently bound VAO.
+    fn draw_instanced_quads(&self, instance_count: i32) {
+        self.gl.bind_buffer(
+            WebGl2RenderingContext::ELEMENT_ARRAY_BUFFER,
+            Some(&self.programs.resources.quad_index_buffer),
+        );
```
grebmeg, line 2815: Could we bind `quad_index_buffer` once in each VAO initializer instead? Since `ELEMENT_ARRAY_BUFFER` is stored as VAO state, this would avoid rebinding it before every draw.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2567,6 +2569,24 @@ fn create_webgl_resources(
...
+    let quad_indices = js_sys::Uint16Array::from(&[0_u16, 1, 2, 3][..]);
+    gl.bind_buffer(
+        WebGl2RenderingContext::ELEMENT_ARRAY_BUFFER,
+        Some(&quad_index_buffer),
+    );
+    gl.buffer_data_with_array_buffer_view(
+        WebGl2RenderingContext::ELEMENT_ARRAY_BUFFER,
+        &quad_indices,
+        WebGl2RenderingContext::STATIC_DRAW,
+    );
```
grebmeg, line 2588: Could this overwrite the `ELEMENT_ARRAY_BUFFER` of a VAO already bound by the caller? Since this binding is stored as VAO state and we haven’t bound one of Vello’s VAOs yet, perhaps we should bind our own VAO first or preserve and restore the previous state?

grebmeg (approved):
Nice!! Thanks for the fix! LGTM, though it would be great to address the two comments in a follow-up.

PR #1829, opened 2026-08-17
vello_hybrid: Use external textures for the probe

LaurenzV:
Makes the code easier to follow I think!

Verified that probing still works with this change on a couple of different devices.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/probe.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/probe.rs
@@ -1,21 +1,21 @@
 // Copyright 2026 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use crate::render::common::IMAGE_PADDING;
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render/webgl/probe.rs
```diff
@@ -177,19 +168,6 @@ impl WebGlRenderer {
...
-            &mut self.programs.resources.atlas_size,
-            &mut probe_atlas_size,
-        );
-        core::mem::swap(
-            &mut self.programs.resources.atlas_layer_count,
-            &mut probe_atlas_layer_count,
-        );
-
         // We do this here instead of above such that in case the render result is not
         // valid, we still properly restore the state (e.g. the old atlas texture array).
```
grebmeg:
```suggestion
// Propagate render failures only after restoring the framebuffer override.
```

grebmeg (approved):
Looks simpler now! LGTM, thanks!

grebmeg (approved):

PR #1830, opened 2026-08-17
vello_hybrid: Limit the texture sizes of data textures to 4096

LaurenzV:
The issue I've been encountering is encapsulated in the following simplified example:

```html
<!doctype html>

<canvas></canvas>

<style>
  html,
  body,
  canvas {
    width: 100%;
    height: 100%;
    margin: 0;
    display: block;
  }
</style>

<script>
  const gl = document.querySelector("canvas").getContext("webgl2");

  // The exact cutoff doesn't seem fully reproducible, but this width seems to
  // fail consistently.
  const width = 16365;

  function compile(type, source) {

    gl.compileShader(shader);

    return shader;
  }

  gl.attachShader(
    program,
    compile(
      gl.VERTEX_SHADER,
      `#version 300 es

void main() {
  vec2 p = vec2(
    gl_VertexID == 1 ? 3.0 : -1.0,
    gl_VertexID == 2 ? 3.0 : -1.0
  );
  gl_Position = vec4(p, 0.0, 1.0);
}
`,
    ),
  );

  gl.attachShader(
    program,
    compile(
      gl.FRAGMENT_SHADER,
      `#version 300 es

precision highp float;
precision highp int;

uniform highp usampler2D texture_;
uniform int frame_;

out vec4 color;

void main() {
  float value = float(texelFetch(texture_, ivec2(0), 0).r & 255u) / 255.0;

  color = frame_ == 0
    ? vec4(value, 0.0, 0.0, 1.0)
    : vec4(0.0, value, 0.0, 1.0);
}
`,
    ),
  );

  gl.linkProgram(program);
  gl.useProgram(program);

  const frameUniform = gl.getUniformLocation(program, "frame_");

  gl.bindTexture(gl.TEXTURE_2D, gl.createTexture());

  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);

  gl.texImage2D(
    gl.TEXTURE_2D,
    0,
    gl.RGBA32UI,
    width,
    1,
    0,
    gl.RGBA_INTEGER,
    gl.UNSIGNED_INT,
    null,
  );

  let frames = 0;

  function frame() {
    const pixels = new Uint32Array(width * 4);
    pixels[0] = 255;

    gl.texSubImage2D(
      gl.TEXTURE_2D,
      0,
      0,
      0,
      width,
      1,
      gl.RGBA_INTEGER,
      gl.UNSIGNED_INT,
      pixels,
    );

    gl.uniform1i(frameUniform, frames);

    frames++;

    if (frames < 2) {
      requestAnimationFrame(frame);
    }
  }

  requestAnimationFrame(frame);
</script>
```

On working machines, this should show red briefly but then go to green. However, on the affected machines, this will show red **and then the browser will completely freeze for ~30 seconds**, after which the WebGL context becomes lost. I have tested this on a Big Sur machine with Chrome 138 and a Monterey machine with Chrome 151. However, it's more likely that it has to do with the GPU and the driver version instead of the OS version. 

Basically, it seems that for unknown reasons, using textures close to the 16k limit (even with just a height and uploading to them repeatedly over at least 2 frames causes this hang. This issue does not seem to be about the overall amount of data uploaded (for example uploading a 4kx4k texture works fine), but instead really about the maximum width of the texture. If I set the width to 8k or even lower, it seems to work fine (even 13k or 14k seem so to have worked most of the time). I know it would be interesting to do more digging, but this is very frustrating to debug (you always need to wait until the freeze is over, you need to restart regularly because at some point Chrome just falls back to the software rasterizer, etc.), so I hope we can just leave it at that for now in the short term.

Setting the maximum limit to 4k gives us storage for 268MB of alpha values, which should be more than sufficient for any sane scene, let alone any of the other data textures. We could do 8k instead since at least the affected machines, but I feel like it's better to stick to a lower number for now. I confirmed that with rhese changes, vello bench works fine on those devices!

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -1732,17 +1732,17 @@ impl WebGlPrograms {
     }
 
     /// Update the alpha texture size if needed.
-    fn maybe_resize_alphas_tex(&mut self, max_texture_dimension_2d: u32, alphas_len: usize) {
+    fn maybe_resize_alphas_tex(&mut self, resource_texture_dimension_2d: u32, alphas_len: usize) {
         let required_alpha_height = (alphas_len as u32)
             // There are 16 1-byte alpha values per texel.
-            .div_ceil(max_texture_dimension_2d << 4);
+            .div_ceil(resource_texture_dimension_2d << 4);
 
         let current_alpha_height = self.resources.alpha_texture_height;
         if required_alpha_height > current_alpha_height {
             // We need to resize the alpha texture to fit the new alpha data.
             assert!(
-                required_alpha_height <= max_texture_dimension_2d,
-                "Alpha texture height exceeds max texture dimensions"
+                required_alpha_height <= resource_texture_dimension_2d,
+                "Alpha texture height exceeds resource texture dimensions"
             );
 
             // Track the new height.
@@ -1753,20 +1753,21 @@ impl WebGlPrograms {
     /// Update the encoded paints texture size if needed.
     fn maybe_resize_encoded_paints_tex(
         &mut self,
-        max_texture_dimension_2d: u32,
+        resource_texture_dimension_2d: u32,
         paint_idxs: &[u32],
     ) {
         let required_texels = paint_idxs.last().unwrap();
-        let required_encoded_paints_height = required_texels.div_ceil(max_texture_dimension_2d);
+        let required_encoded_paints_height =
+            required_texels.div_ceil(resource_texture_dimension_2d);
         let current_encoded_paints_height = self.resources.encoded_paints_texture_height;
         if required_encoded_paints_height > current_encoded_paints_height {
             assert!(
-                required_encoded_paints_height <= max_texture_dimension_2d,
-                "Encoded paints texture height exceeds max texture dimensions"
+                required_encoded_paints_height <= resource_texture_dimension_2d,
+                "Encoded paints texture height exceeds resource texture dimensions"
             );
 
             let required_encoded_paints_size =
-                (max_texture_dimension_2d * required_encoded_paints_height) << 4;
+                (resource_texture_dimension_2d * required_encoded_paints_height) << 4;
             self.encoded_paints_data
                 .resize(required_encoded_paints_size as usize, 0);
             self.resources.encoded_paints_texture_height = required_encoded_paints_height;
@@ -1777,7 +1778,7 @@ impl WebGlPrograms {
     fn maybe_resize_gradient_tex(
         &mut self,
         _gl: &WebGl2RenderingContext,
-        max_texture_dimension_2d: u32,
+        resource_texture_dimension_2d: u32,
         gradient_cache: &GradientRampCache,
     ) {
         if gradient_cache.is_empty() {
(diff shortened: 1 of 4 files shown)


sparse_strips/vello_hybrid/src/render/wgpu/mod.rs
```diff
@@ -1024,6 +1025,8 @@ struct GpuResources {
     strips_buffer: Buffer,
     /// Alpha texture.
     alphas_texture: Texture,
+    /// Maximum width or height of packed resource textures.
+    resource_texture_dimension_2d: u32,
```
grebmeg, line 1029: Since this workaround targets a webgl driver issue, could we apply the 4K cap only to the webgl backend?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -379,16 +379,21 @@ impl WebGlRenderer {
             );
         }
         let resources = Resources::new(settings.memory_settings.image_atlas_config);
+        let resource_texture_dimension_2d = device_limits.resource_texture_dimension_2d();
```
grebmeg, line 382: What do you think about moving the main texture limits into `MemorySettings`? Consumers could configure each limit explicitly or choose a preset that requests the maximum supported allocations, with each backend normalizing those values against its device limits. WebGL could additionally cap affected resource textures at `4096`.

LaurenzV:
@grebmeg After thinking about this a bit more, I think I'm still inclined to leave it as is. I'd like to avoid the WebGL and WGPU backends from diverging too much, so I think it's better if it applies to both backends consistently. If it turns out to be too little, in the future we can still raise the limit to 8k, which from my experiments the devices can handle just fine. Setting it to 4k was just really more of a cautionary measure, since in my view we aren't going to need more than that for any sane scene. And as we discussed, I think for now it's better to not expose this to the user, but open to changing this in the future if we figure out how. 

Let me know if that works for you or if you have other comments!

grebmeg (approved):
@LaurenzV I think it would be really interesting to debug this further, as the fix seems quite unexpected.

More broadly, I don't think the best way to address GPU bugs is to lower our limits to the common denominator of a buggy device, especially when that also affects the wgpu backend.

That said, I think the current fix is fine for now. I agree with you that 4K should be more than sufficient for almost all cases, and we can reconsider it later if needed. My concern is that revisiting this in the future could be painful, as we'd need to rebuild the same context and investigate the issue again (for example to relax the limit or remove it for wgpu).

Another thing we should probably start doing is reporting issues like this upstream, whether that's to ANGLE, browser vendors, or other relevant projects. That way, we can at least get some triage and better understand the root cause. Some of these issues can potentially be fixed in ANGLE itself. For example, you can look at [gl_features.json](https://github.com/google/angle/blob/main/include/platform/gl_features.json) and [gpu_driver_bug_list.json](https://github.com/chromium/chromium/blob/main/gpu/config/gpu_driver_bug_list.json), which lists many browser shader-related issues that have been addressed in angle and chromium.

LaurenzV:
> My concern is that revisiting this in the future could be painful, as we'd need to rebuild the same context and investigate the issue again (for example to relax the limit or remove it for wgpu).

Makes sense, but I don't think it should be too bad either. I've provided a minimal example above as well as the machines where I can reproduce it, so hopefully it shouldn't be too hard to get up to speed if we need to revisit this in the future!

I agree regarding bug reporting, since this issue is reproducible on newest Chrome on some machines, it's probably something worth reporting, will do so when I find some time! Thanks again!

PR #1831, opened 2026-08-18
vello_hybrid: Compact shaders

LaurenzV:
Currently, functions are duplicated across both, vertex and fragment shader in the GLSL files, even if they are only needed in one stage. We can easily prune them by calling `naga::compact::compact`. This reduces the file size of `compiled_shaders.rs` from 209KB to 164KB.

I manually checked the resulting diff from the GLSL files and they look reasonable.

### DIFF
--- a/sparse_strips/vello_sparse_shaders/src/compile.rs
+++ b/sparse_strips/vello_sparse_shaders/src/compile.rs
@@ -2,10 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use naga::{
-    ShaderStage,
+    Module, ShaderStage,
     back::glsl::{self, PipelineOptions, Version},
+    compact::KeepUnused,
     front::wgsl,
-    valid::{Capabilities, ValidationFlags, Validator},
+    valid::{Capabilities, ModuleInfo, ValidationFlags, Validator},
 };
 
 use crate::lint::lint;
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_sparse_shaders/src/compile.rs
```diff
@@ -40,45 +37,36 @@ pub(crate) fn compile_wgsl_shader(
...
-    let reflection_vs = {
-        let pipeline_options = PipelineOptions {
-            entry_point: vertex_entry.into(),
-            shader_stage: ShaderStage::Vertex,
-            multiview: None,
-        };
+    CompiledGlsl {
+        vertex: compile_stage(&module, vertex_entry, ShaderStage::Vertex, &options),
+        fragment: compile_stage(&module, fragment_entry, ShaderStage::Fragment, &options),
+    }
```
grebmeg, line 43: Could we enable [WESL stripping](https://github.com/linebender/vello/blob/354661ade0a3263d906c52f0cf4838c3997a15a0/sparse_strips/vello_sparse_shaders/build.rs#L50-L51) as well? It would remove declarations unused by all entry points and avoid loading unused imports earlier, although it wouldn’t replace the per-stage naga compaction because dependencies used by the other stage would still be retained.

grebmeg (approved):
Nice! Thank you so much! ❤️

sparse_strips/vello_sparse_shaders/src/compile.rs
LaurenzV: Yep will enable it. Seems like it does remove some constants.

PR #1833, opened 2026-08-19
perf: reduce generated WebGL shader size

grebmeg:
On top of #1831 

## Problem
Splitting shaders into WESL modules made generated identifiers significantly longer because imported declarations include their full module paths.

For example, the generated GLSL previously contained:

```glsl
vec4 package_helpers_1external_texture_2external_bicubic_sample(...) {
    vec4 weights = package_helpers_image_1cubic_weights(...);
}
```

After this change:

```glsl
vec4 w4Eev8DamF9z(...) {
    vec4 weights = weCTPm97SrWh(...);
}
```

## Root Cause
WESL’s default `EscapeMangler` produces reversible names by encoding the complete module path. These names appear in declarations and every call site, increasing the embedded shader size.

## Solution
Use a custom mangler that:

- Generates deterministic 64-bit base62 identifiers.
- Detects collisions and fails the build.
- Retains reverse mappings for diagnostics.
- Keeps root declarations and shader entry points unchanged.
- Ends identifiers with letters to avoid additional suffixes from Naga.

## Size Impact
The baseline is commit after shader compaction from #1831.

Both revisions were built with fresh `CARGO_TARGET_DIR` directories, then measured with:

```sh
cargo build -p vello_sparse_shaders --features glsl

wc -c \
    "$CARGO_TARGET_DIR"/debug/build/vello_sparse_shaders-*/out/compiled_shaders.rs
```

Results:

- Before: **164,800 bytes** (160.94 KiB)
- After: **141,713 bytes** (138.39 KiB)
- Saved: **23,087 bytes** (22.55 KiB)
- Reduction: **14.01%**

### DIFF
--- a/sparse_strips/vello_sparse_shaders/src/mangle.rs
+++ b/sparse_strips/vello_sparse_shaders/src/mangle.rs
@@ -10,8 +10,8 @@ use wesl::{Mangler, ModulePath};
 
 const FNV_OFFSET_BASIS: u64 = 0xcbf29ce484222325;
 const FNV_PRIME: u64 = 0x100000001b3;
-const BASE62: &[u8; 62] = b"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
-const LETTERS: &[u8; 52] = b"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+const BASE62: &[u8] = b"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+const LETTERS: &[u8] = b"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 
 /// Mangler producing compact names from stable hashes of fully-qualified declarations.
 #[derive(Default)]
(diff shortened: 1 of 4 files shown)


sparse_strips/vello_sparse_shaders/src/mangle.rs
```diff
@@ -0,0 +1,135 @@
...
+    let mut buffer = [0; 11];
+    let mut start = buffer.len() - 1;
+
+    // Naga appends `_` to names ending in digits so that its own numeric suffixes remain
+    // unambiguous. Use a letter for the final digit while retaining base62 for the rest.
+    buffer[start] = LETTERS[(value % 52) as usize];
+    value /= 52;
+    while value != 0 {
+        start -= 1;
+        buffer[start] = BASE62[(value % 62) as usize];
```
LaurenzV: Could we use `.len()` instead of hardcoding the length as numbers?

LaurenzV (approved):
Can't really judge how well this hash works, but since we detect collisions at built-time anyway, it's fine. LGTM!

sparse_strips/vello_sparse_shaders/src/mangle.rs
grebmeg: Updated — both radices are now derived from the corresponding alphabet lengths, and the alphabet constants no longer duplicate those lengths in their types.

PR #1834, opened 2026-08-19
vello_common: Unify and improve constructor for pixmaps

LaurenzV:
This PR makes two changes:
- It combines all pixmap constructors into a single constructor that always takes Vec<u8> and makes the specific alpha options an additional configuration. This way we have only a single entry point and avoid duplicate documentation.
- It provides a way of submitting unpremultiplied data directly, so that clients don't have to reimplement it. It drastically improves the performance of the premultiplication step (compared to what is currently used for PNGs) by only relying on u8/u16 and using SIMD. A small downside is that the precision can be slightly worse, hence why some test images had to be regenerated. In the normal case there was no delta larger than 1, except for the blending test cases where the rounding errors while performing blending. If this is a problem, I'm happy to add another parameter for defining whether to do premultiplication using exact f32 arithmetic or not and try to SIMDify it as well, but I would like to avoid this.

## Neon

On NEON, I'm getting the following results compared to what we had before (in a make-shift benchmark I set up, comparing the existing premultiplication code of the from_png method to the new one). As you can see, it basically gives us a 3x speedup!
```
premultiply_rgba8/800x500/old_scalar
                        time:   [156.82 µs 157.15 µs 157.48 µs]
                        thrpt:  [9.4621 GiB/s 9.4821 GiB/s 9.5023 GiB/s]
premultiply_rgba8/800x500/new_simd_interleaved_64_with_transparency
                        time:   [46.522 µs 46.593 µs 46.664 µs]
                        thrpt:  [31.933 GiB/s 31.981 GiB/s 32.030 GiB/s]

premultiply_rgba8/1920x1080/old_scalar
                        time:   [827.20 µs 828.84 µs 830.58 µs]
                        thrpt:  [9.3005 GiB/s 9.3199 GiB/s 9.3385 GiB/s]

premultiply_rgba8/1920x1080/new_simd_interleaved_64_with_transparency
                        time:   [243.32 µs 243.69 µs 244.03 µs]
                        thrpt:  [31.655 GiB/s 31.699 GiB/s 31.748 GiB/s]

premultiply_rgba8/3840x2160/old_scalar
                        time:   [3.3393 ms 3.3459 ms 3.3529 ms]
                        thrpt:  [9.2155 GiB/s 9.2349 GiB/s 9.2532 GiB/s]

premultiply_rgba8/3840x2160/new_simd_interleaved_64_with_transparency
                        time:   [980.54 µs 981.64 µs 982.72 µs]
                        thrpt:  [31.442 GiB/s 31.477 GiB/s 31.512 GiB/s]
```

## AVX2

4x speedup!

```
premultiply_rgba8/800x500/old_scalar
                        time:   [830.42 µs 836.86 µs 843.27 µs]
                        thrpt:  [1.7671 GiB/s 1.7806 GiB/s 1.7944 GiB/s]
premultiply_rgba8/800x500/new_simd_interleaved_64_with_transparency
                        time:   [194.31 µs 196.88 µs 199.57 µs]
                        thrpt:  [7.4668 GiB/s 7.5687 GiB/s 7.6687 GiB/s]

premultiply_rgba8/1920x1080/old_scalar
                        time:   [4.0676 ms 4.1098 ms 4.1722 ms]
                        thrpt:  [1.8515 GiB/s 1.8796 GiB/s 1.8991 GiB/s]
premultiply_rgba8/1920x1080/new_simd_interleaved_64_with_transparency
                        time:   [962.16 µs 964.40 µs 966.92 µs]
                        thrpt:  [7.9891 GiB/s 8.0099 GiB/s 8.0285 GiB/s]

premultiply_rgba8/3840x2160/old_scalar
                        time:   [16.177 ms 16.183 ms 16.190 ms]
                        thrpt:  [1.9085 GiB/s 1.9094 GiB/s 1.9101 GiB/s]
premultiply_rgba8/3840x2160/new_simd_interleaved_64_with_transparency
                        time:   [3.8700 ms 3.8794 ms 3.8898 ms]
                        thrpt:  [7.9437 GiB/s 7.9649 GiB/s 7.9842 GiB/s]
``` 

## WASM
I also tested this on WASM on some devices on BrowserStack, and was seeing similar results. Very weirdly, on an iPhone 12 I was even seeing an 8x speedup!


sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -88,46 +93,42 @@ impl Pixmap {
...
+            // If not, we can mark it as fully opaque. This only adds little overhead (around
+            // 5-10% from my benchmarks), but is useful in cases where the user conservatively
+            // passed `true` because they received their image from a PNG or similar, but in
+            // reality the image is fully opaque.
+            premultiply_rgba8(&mut data)
+        } else {
+            pixel_metadata.may_have_transparency
+        };
+
+        let data: Vec<PremulRgba8> = bytemuck::try_cast_vec(data)
```
LaurenzV, line 123: This is a bit unfortunate. But really, this problem already exists on current main for any client that works on `Vec<u8>` instead of peniko-native `PremulRgba` type, so the problem existed before and at least now it's centralized in one location.

Open for suggestions on how to improve this, though.

sparse_strips/vello_common/src/paint.rs
```diff
@@ -174,40 +172,26 @@ impl ImageSource {
...
-                } else {
-                    PremulRgba8 {
-                        r: rgba[0],
-                        g: rgba[1],
-                        b: rgba[2],
-                        a: rgba[3],
-                    }
+        // Unfortunately, we have to create a new allocation, because pixmap requires
+        // a real vector.
+        let mut rgba = image.data.data().to_vec();
```
grebmeg, line 178: Would it be possible to remove this allocation? I see two ways to avoid this allocation: normalize the pixel `Vec` in place before creating `ImageData`, or consume `ImageData`, recover its uniquely owned buffer (`take` ?), and normalize that in place and then put it back. The first is simpler but moves conversion responsibility upstream; the second preserves this API boundary but requires adding something in `Blob`.

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -88,46 +93,42 @@ impl Pixmap {
...
+            && pixel_metadata.alpha_type == ImageAlphaType::Alpha
+        {
+            // If there might be transparency and the data is not premultiplied yet, we need to
+            // iterate over all pixels anyway to premultiply. Because of this, we can use the
+            // opportunity to recheck the pixels whether there really is any transparent pixel.
+            // If not, we can mark it as fully opaque. This only adds little overhead (around
+            // 5-10% from my benchmarks), but is useful in cases where the user conservatively
+            // passed `true` because they received their image from a PNG or similar, but in
+            // reality the image is fully opaque.
+            premultiply_rgba8(&mut data)
```
grebmeg, line 115: Should we move this comment onto `premultiply_rgba8` itself? The function returns `may_have_transparency` but has no doc explaining that, so anyone reading the definition has to jump to the call site to understand the return value. Documenting it on the function would make the return meaning self-contained, and we could keep just the caller-specific rationale (downgrading a conservative true to opaque) here.

sparse_strips/vello_common/src/paint.rs
```diff
@@ -174,40 +172,26 @@ impl ImageSource {
...
-        let pixmap = Pixmap::from_parts_with_opacity(pixels, width, height, may_have_transparency);
+            }
+            format => unimplemented!("Unsupported image format: {format:?}"),
+        }
+
+        let pixmap = Pixmap::from_parts(
+            rgba,
+            width,
+            height,
+            PixelMetadata::new(image.alpha_type, true),
```
grebmeg, line 194: For premultiplied images, we skip `premultiply_rgba8`, so we also skip checking whether every alpha value is 255. Since `from_peniko_image_data` always passes `may_have_transparency: true`, fully opaque premultiplied images remain marked as potentially transparent. Is that intentional, or should we somehow preserve the previous opacity detection for this case?

sparse_strips/vello_common/src/pixmap.rs
grebmeg: Could we store `Vec<u8>` on `Pixmap` internally instead? The constructor and SIMD path already operate on bytes, and typed access can remain zero-cost through `bytemuck::cast_slice`, without the fragile capacity requirement of `try_cast_vec`.

sparse_strips/vello_common/src/paint.rs
LaurenzV: The problem is the second won't work because internally a blob just stores a `Arc<dyn AsRef<[T]> + Send + Sync>`, so it's not guaranteed to be mutable (and I believe this is intentional).

And for the first one, yes, the user can manually convert it to RGBA8 themselves, but then from our side there isn't much we can do other than document that BGRA will allocate internally, right? In any case, this is not a regression compared to what we had previously, so I think it's safe to ignore for this PR, right?

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: Updated this.

sparse_strips/vello_common/src/paint.rs
LaurenzV: Yes, this is intentional, I added another comment for this. As was briefly mentioned in the original one, I'm not sure it's worth loading all pixels from memory once just to perform this check. If we need to premultiply then we need to load them anyway so we might as well piggy-back onto that to check the transparency, but in case we don't, it seems to me to be better to just not bother in the first place? 

But if you think it would be good to have, we could do another linear scan, I don't feel very strongly about it.

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: We could try it, but it will require some code changes because other code assumes to be able to sample `PremulRgba` directly. It also requires more careful benchmarking again, because instead of simply sampling a single value in memory, we always need to slice it now [0..4]. So overall, I think it would be better for a follow-up. In our codebase we are already using `cast_vec` when creating the pixmap, so it's not a regression for us.

sparse_strips/vello_common/src/paint.rs
grebmeg: Yeah, you’re right — we already allocated there through `collect()`. Still, avoiding the copy would be valuable on wasm: a 4096×4096 RGBA8 pixmap is 64 MiB, so retaining the original blob while copying it requires about 128 MiB just for pixel data. It may be worth finding another approach, even if that means using something other than `ImageData` here.

grebmeg (approved):
Thanks for answering my comments! I understand this better now and agree with merging this, with my comments/ideas potentially addressed in future follow-ups.

sparse_strips/vello_common/src/paint.rs
LaurenzV: Sure, let me turn this into a TODO then.

PR #1835, opened 2026-08-21
perf: use immutable storage for WebGL data textures

grebmeg:
## Problem

The WebGL renderer calls `texImage2D` every frame for alpha, encoded-paint, gradient, and filter-data textures. This uploads the texture’s full peak allocation even when the current frame only uses a few rows.

After one large frame grows a texture, subsequent small frames continue paying the peak-sized upload cost.

## Root Cause

`texImage2D` does more than copy pixels: it **specifies the texture storage** for a mip level, including its dimensions and internal format.

**respecification**: the call tells WebGL that the existing storage may be replaced with a differently sized or formatted allocation. Even when the arguments remain unchanged, the API permits a storage change, so the browser and driver may need to:

- validate the dimensions, format, and texture completeness again;
- allocate, replace, or orphan the backing GPU storage;
- update internal texture metadata;
- synchronize or defer allocation until the texture is used.

By contrast, `texSubImage2D` cannot change the allocation. It only writes pixels into storage whose size and format are already fixed, giving the implementation stronger guarantees.

The renderer’s CPU buffers and texture heights only grow, so the old `texImage2D` calls also upload the largest allocation observed rather than the rows required by the current frame.

## Solution

Allocate these data textures with immutable `texStorage2D` storage and update their contents using `texSubImage2D`.

- Recreate immutable storage only when a texture must grow.
- Upload only the full-width rows containing current-frame data.
- Leave unused rows untouched because current-frame shader indices cannot reference them.
- Allocate one mip level because these data textures do not use mipmapping.

## Expected Performance Impact

The improvement should primarily appear in CPU time inside `WebGlRenderer::render`, specifically during per-frame texture uploads:

- less browser/driver allocation and validation work;
- fewer opportunities for allocation-related stalls or pipeline flushes;
- lower CPU-to-GPU transfer volume when the current frame is smaller than the historical peak;
- more stable frame times after temporary workload spikes grow the textures.

The largest improvement should occur when a heavy frame is followed by smaller frames. Previously, every smaller frame continued re-specifying and uploading the peak-sized texture; now it uploads only its active rows.

Frames that grow a texture still require a new allocation, but growth is infrequent because capacity only increases. Frames that genuinely use the full peak allocation will transfer approximately the same amount of data as before.

## References

- [MDN: WebGL best practices — Use `texStorage` to create textures](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_texstorage_to_create_textures) explains that independently specified `texImage` mip levels prevent drivers from preparing texture memory until draw time. It recommends `texStorage` + `texSubImage` for WebGL 2 and notes that some drivers may otherwise allocate an unnecessary full mip chain, adding approximately 30% memory.
- [MDN: `texStorage2D`](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/texStorage2D) documents allocating texture levels with fixed dimensions and format.

- [Khronos: Texture Storage](https://wikis.khronos.org/opengl/Texture_Storage) explains the fundamental difference between mutable and immutable allocation. Mutable storage is defined one mip level at a time with `glTexImage*`, and those levels can later be replaced with different dimensions or formats. This leaves the application responsible for maintaining texture completeness. Immutable storage allocates all requested levels together with fixed dimensions and formats. Importantly, “immutable” applies only to the allocation — not its pixel contents. The contents can still be replaced repeatedly with `glTexSubImage*`. For Vello, this means the backing allocation remains stable between frames while frame-specific data can still be uploaded normally.
- [Khronos: Pixel Transfer](https://wikis.khronos.org/opengl/Pixel_Transfer) explicitly separates allocation from updating:
  - `glTexImage*` **allocates mutable storage** for a mip level and optionally writes pixel data.
  - `glTexSubImage*` only **writes pixel data into existing storage**.

  This distinction is directly relevant to Vello. The old code used the allocation-capable `texImage2D` operation for every frame update, even when the texture dimensions and format had not changed. The new code uses `texStorage2D` only when creating or growing a texture, then uses the update-only `texSubImage2D` path during normal frames. That communicates to the driver that no allocation or format change is possible during routine uploads.

These Khronos pages describe OpenGL, but the same storage-versus-content model underlies the equivalent WebGL 2 APIs. They establish the semantic reason for using `texStorage2D`; they do not by themselves guarantee a particular performance improvement.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -1467,13 +1467,13 @@ impl WebGlPrograms {
     ) {
         let resource_texture_dimension_2d = self.resources.resource_texture_dimension_2d;
 
-        self.maybe_resize_alphas_tex(resource_texture_dimension_2d, alphas.len());
-        self.maybe_resize_encoded_paints_tex(resource_texture_dimension_2d, paint_idxs);
-        self.maybe_resize_filter_data_tex(filter_context);
+        self.maybe_resize_alphas_tex(gl, resource_texture_dimension_2d, alphas.len());
+        self.maybe_resize_encoded_paints_tex(gl, resource_texture_dimension_2d, paint_idxs);
+        self.maybe_resize_filter_data_tex(gl, filter_context);
         self.maybe_update_config_buffer(gl, resource_texture_dimension_2d, render_size);
 
         self.upload_alpha_texture(gl, alphas);
-        self.upload_encoded_paints_texture(gl, encoded_paints);
+        self.upload_encoded_paints_texture(gl, encoded_paints, paint_idxs);
         self.upload_filter_data_texture(gl, filter_context);
 
         if gradient_cache.has_changed() {
@@ -1644,7 +1644,14 @@ impl WebGlPrograms {
         }
     }
 
-    fn maybe_resize_filter_data_tex(&mut self, filter_context: &FilterContext) {
+    /// Grow the filter data texture if needed.
+    ///
+    /// See [`maybe_resize_alphas_tex`](Self::maybe_resize_alphas_tex) for the storage strategy.
+    fn maybe_resize_filter_data_tex(
+        &mut self,
+        gl: &WebGl2RenderingContext,
+        filter_context: &FilterContext,
+    ) {
         let resource_texture_dimension_2d = self.resources.resource_texture_dimension_2d;
 
         let Some(required_height) =
@@ -1656,10 +1663,20 @@ impl WebGlPrograms {
         if required_height > self.resources.filter_data_texture_height {
             let required_size = (resource_texture_dimension_2d * required_height) << 4;
             self.filter_data.resize(required_size as usize, 0);
+            self.resources.filter_data_texture = create_data_texture_storage(
+                gl,
+                WebGl2RenderingContext::RGBA32UI,
+                resource_texture_dimension_2d,
+                required_height,
+            );
             self.resources.filter_data_texture_height = required_height;
         }
     }
 
+    /// Upload filter data to the texture.
+    ///
+    /// Only the rows covering this frame's filter data are uploaded; see
+    /// [`upload_alpha_texture`](Self::upload_alpha_texture).
     fn upload_filter_data_texture(
         &mut self,
         gl: &WebGl2RenderingContext,
@@ -1670,27 +1687,18 @@ impl WebGlPrograms {
         }
 
         let width = self.resources.resource_texture_dimension_2d;
-        let height = self.resources.filter_data_texture_height;
+        let used_height = filter_context
+            .required_filter_data_height(width)
+            .expect("non-empty filter context must have a data height");
+        let used_size = ((width * used_height) << 4) as usize;
         filter_context.serialize_to_buffer(&mut self.filter_data);
-        gl.active_texture(WebGl2RenderingContext::TEXTURE0);
-        gl.bind_texture(
-            WebGl2RenderingContext::TEXTURE_2D,
-            Some(&self.resources.filter_data_texture),
+        upload_rgba32ui_rows(
+            gl,
+            &self.resources.filter_data_texture,
+            bytemuck::cast_slice::<u8, u32>(&self.filter_data[..used_size]),
+            width,
+            used_height,
         );
-        let data_as_u32 = bytemuck::cast_slice::<u8, u32>(&self.filter_data);
-        let packed_array = js_sys::Uint32Array::from(data_as_u32);
-        gl.tex_image_2d_with_i32_and_i32_and_i32_and_format_and_type_and_opt_array_buffer_view(
-            WebGl2RenderingContext::TEXTURE_2D,
-            0,
-            WebGl2RenderingContext::RGBA32UI as i32,
-            width as i32,
-            height as i32,
-            0,
-            WebGl2RenderingContext::RGBA_INTEGER,
-            WebGl2RenderingContext::UNSIGNED_INT,
-            Some(&packed_array),
-        )
-        .unwrap();
     }
 
     fn upload_filter_instances(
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -3627,23 +3691,33 @@ fn copy_to_texture_array_layer(
...
 }
 
-// Upload the data to the currently bound texture assuming a RGBA32UI format.
-fn upload_data_to_rgba32_texture(
+/// Upload `rows` full rows of RGBA32UI texel data into the given texture's existing
+/// (immutable) storage via `texSubImage2D`.
+///
+/// `data` must contain exactly `width * rows` RGBA32UI texels (4 `u32`s per texel). Only the
+/// uploaded rows are modified; rows above `rows` keep their previous contents, which is fine
+/// because shaders only read texels referenced by the current frame's data.
```
LaurenzV: Up to you, but once again, I don't think we need to put so much "internal" details into doc comments, so we can make them shorter. What do you think about something like this?

```
/// Upload full rows of RGBA32UI data into immutable texture storage.
///
/// `data` must contain exactly `width * rows * 4` `u32` values.
/// Texels outside the uploaded rows remain unchanged.
```

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2497,44 +2552,53 @@ fn create_texture_inner(
...
+}
+
 /// Create a 1x1 RGBA8 placeholder texture.
 ///
 /// A texture with no allocated level is incomplete, and binding an incomplete texture to a sampler
 /// makes the draw call non-renderable: the driver substitutes a black texture for every draw and
 /// browsers log a render warning for each one. Data textures are only uploaded when a scene has
 /// data for them, so allocate a minimal level 0 up front to keep them complete for scenes that
-/// don't.
+/// don't. WebGL zero-initializes the storage, so these sample as transparent black.
```
LaurenzV: Since it's just a placeholder and we don't sample from it, is it relevant to explicitly state that it samples as transparent black?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2497,44 +2552,53 @@ fn create_texture_inner(
...
+        internal_format,
+        width as i32,
+        height as i32,
+    );
     texture
 }
 
+/// Create a 1x1 RGBA32UI placeholder texture.
+fn create_placeholder_rgba32ui_texture(gl: &WebGl2RenderingContext) -> Texture {
+    // See `create_placeholder_rgba8_texture` for why the initial allocation matters.
```
LaurenzV: What initial allocation? Just from looking at the function call and the comment it's not obvious to me what this is referring to.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2497,44 +2552,53 @@ fn create_texture_inner(
...
+/// Create a data texture backed by immutable storage (`texStorage2D`).
+///
+/// Immutable storage is preferred over `texImage2D` for the per-frame data textures: the driver
+/// allocates and validates the texture once at creation instead of re-specifying (and
+/// re-validating) the backing store on every upload, and contents can then be streamed with
+/// `texSubImage2D` covering only the region that changed. See
+/// <https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_texstorage_to_create_textures>.
+///
+/// The size and format are fixed for the lifetime of the texture; growing requires creating a
+/// new texture. WebGL guarantees the storage is zero-initialized.
```
LaurenzV: Sorry for being nitpicky again. 😦 But can we once again reduce the text here a bit?

```suggestion
/// Create a zero-initialized data texture backed by immutable storage (`texStorage2D`).
```

And then maybe above the `text_storage_2d` call something like:
```
We prefer using `textStorage2D` over `texImage2D` due to potentially better performance properties, see <https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_texstorage_to_create_textures> for more information.
```

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1844,50 +1886,61 @@ impl WebGlPrograms {
...
         );
 
         // Truncate back to the original size.
         alphas.truncate(original_len);
     }
 
     /// Upload encoded paints to the texture.
+    ///
+    /// Only the rows covering this frame's paints are uploaded; see
+    /// [`upload_alpha_texture`](Self::upload_alpha_texture).
```
LaurenzV: Once again, is this something that really needs to be explicitly documented?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -1844,50 +1886,61 @@ impl WebGlPrograms {
     }
 
     /// Upload alpha data to the texture.
+    ///
+    /// Only the rows used by this frame's alpha data are uploaded, not the full (peak-sized)
+    /// texture. Rows above keep stale contents from previous frames, which is fine because the
+    /// shader only reads alpha indices produced by the current frame.
```
LaurenzV: Same here and all other `maybe_resize` methods, and `upload_filter_data_texture`.

LaurenzV (approved):
LGTM! I'm sorry for being so nitpicky about the comments. 😦 If you find it annoying, feel free to disregard my proposed changes. 

The reason why I personally think it's worth paying attention to is simply because having those lengthy comments makes future code reviews harder, and unfortunately AI has a tendency to add comments for _everything_. Because in case anything ever changes, not only do we need to re-review all of the code, but we also need to reread and verify all the comments, so it adds additional churn. 

So for example instead of referring to the storage strategy from all the different callers, I think it's better to just specify it once where it really happens. And in case users want to read more about the specific advantages of `texStorage2D` they can simply open the link or do their own research, I don't think we need to document all of this in the code itself, we only need to document the minimum so the rationale for _why_ we wrote the code the way it is now can be understood later on in case anyone reads the code. 

But if you don't mind this let me know and I won't point it out in the future anymore. 😄

grebmeg:
Simplified the doc comments as suggested.

PR #1836, opened 2026-08-21
vello_hybrid:  Improvements to bindings of element array buffer

LaurenzV:
Addresses two post-review comments from #1827.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -2383,6 +2383,10 @@ const FILTER_INSTANCE_STRIDE: i32 = size_of::<FilterInstanceData>() as i32;
 fn initialize_filter_vao(gl: &WebGl2RenderingContext, resources: &WebGlResources) {
     gl.bind_vertex_array(Some(&resources.filter_vao));
     gl.bind_buffer(
+        WebGl2RenderingContext::ELEMENT_ARRAY_BUFFER,
+        Some(&resources.quad_index_buffer),
+    );
+    gl.bind_buffer(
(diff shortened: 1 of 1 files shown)


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2577,6 +2589,12 @@ fn create_webgl_resources(
     // In Vulkan u8 only was added in later versions: https://docs.vulkan.org/refpages/latest/refpages/source/VkIndexType.html
     // ANGLE also seems to special-case on Metal, as additional evidence: https://chromium.googlesource.com/angle/angle/+/bfc764c553fa0613f858315bc6c0cc1ecee469a1/src/libANGLE/renderer/metal/VertexArrayMtl.mm#63
     let quad_indices = js_sys::Uint16Array::from(&[0_u16, 1, 2, 3][..]);
+
+    let temp_vao = VertexArray::new(gl);
```
grebmeg: Could we avoid the throwaway VAO by creating and binding one of Vello’s persistent VAOs, such as `strip_vao`, before uploading the index buffer? Alternatively, we could keep a dedicated upload VAO as a persistent resource, though reusing `strip_vao` seems simpler.

grebmeg (approved):

grebmeg (approved):
LGTM with the comment addressed about reusing the temp vao instead of creating a new one in `create_webgl_resources`.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Done.

PR #1839, opened 2026-08-22
vello_common: Optimize pixmap unpremultiplication

LaurenzV:
Builds on top of #1838. This PR adds SIMD acceleration for the pixmap unpremultiplication method. Unfortunately, unpremultiplying is fundamentally harder than premutiplying because we need to divide by (a non-constant) integer, which isn't well supported in SIMD. I experimented with various different approaches, such as using SIMD but with f32 instead, but this one seemed to be by far the fastest.

I also tested this in my PDF library and didn't see any regressions.

## NEON
Before:
``` 
pixmap/take_unpremultiplied
                        time:   [2.2905 ms 2.3092 ms 2.3296 ms]
                        change: [+156.03% +161.53% +166.71%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  1 (1.00%) high severe
``` 

After:
``` 
pixmap/take_unpremultiplied
                        time:   [920.59 µs 944.89 µs 969.06 µs]
                        change: [-61.151% -60.308% -59.391%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
``` 

It's still 2-3 times slower than premultiplying, but at least it's faster than before! The gather for the reciprocals is unfortunately pretty expensive.

## AVX2
I can't check currently, unfortunately, but will test this as soon as I have access to my device again.

### DIFF
--- a/sparse_strips/vello_common/src/pixmap.rs
+++ b/sparse_strips/vello_common/src/pixmap.rs
@@ -398,7 +398,7 @@ impl Pixmap {
     /// The pixels are in row-major order.
     pub fn take_unpremultiplied(mut self) -> Vec<Rgba8> {
         unpremultiply_rgba8(bytemuck::cast_slice_mut(&mut self.buf));
-        
+
         bytemuck::cast_vec(self.buf)
     }
 }
@@ -448,7 +448,7 @@ fn premultiply_rgba8(data: &mut [u8]) -> bool {
 /// Unpremultiplies each RGBA8 pixel in `data`.
 fn unpremultiply_rgba8(data: &mut [u8]) {
     let level = Level::try_detect().unwrap_or(Level::baseline());
-    
+
     dispatch!(level, simd => unpremultiply_rgba8_impl(simd, data));
 }
 
--- a/sparse_strips/vello_common/src/pixmap.rs
+++ b/sparse_strips/vello_common/src/pixmap.rs
@@ -400,7 +400,9 @@ impl Pixmap {
     ///
     /// The pixels are in row-major order.
     pub fn take_unpremultiplied(mut self) -> Vec<Rgba8> {
-        unpremultiply_rgba8(bytemuck::cast_slice_mut(&mut self.buf));
+        if self.may_have_transparency {
+            unpremultiply_rgba8(bytemuck::cast_slice_mut(&mut self.buf));
+        }
 
         bytemuck::cast_vec(self.buf)
     }
--- a/sparse_strips/vello_common/src/pixmap.rs
+++ b/sparse_strips/vello_common/src/pixmap.rs
@@ -95,6 +95,9 @@ impl Pixmap {
 
     /// Create a new pixmap from the given buffer of bytes, representing pixel data.
     ///
+    /// When passing premultiplied pixels, the data must be correctly premultiplied, i.e. each RGB
+    /// component must be less than or equal to its pixel's alpha component.
+    ///
     /// # Panics
     ///
     /// - Panics if `data` is not exactly `width * height * 4` bytes long.
--- a/sparse_strips/vello_common/src/pixmap.rs
+++ b/sparse_strips/vello_common/src/pixmap.rs
@@ -8,12 +8,12 @@ use alloc::vec::Vec;
 #[cfg(feature = "png")]
 use std::io::{BufRead, Seek};
 
-use crate::fearless_simd::{Level, Simd, SimdBase, SimdInt, SimdMask, dispatch, mask8x16};
+use crate::fearless_simd::{Level, Simd, SimdBase, SimdInt, SimdMask, dispatch, mask8x16, u16x16};
 use crate::peniko::{
     ImageAlphaType,
     color::{PremulRgba8, Rgba8},
 };
-use crate::util::Div255Ext;
+use crate::util::{Div255Ext, unpremultiply};
 
 #[cfg(feature = "png")]
 extern crate std;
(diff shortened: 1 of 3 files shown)


sparse_strips/vello_common/src/util.rs
```diff
@@ -369,6 +369,139 @@ pub fn strip_bbox(strips: &[Strip]) -> Option<RectU16> {
...
+    const RECIPROCALS: [u16; 256] = reciprocals();
+
+    #[inline(always)]
+    pub(crate) const fn reciprocal(alpha: u8) -> u16 {
+        RECIPROCALS[alpha as usize]
+    }
+
+    #[inline(always)]
+    pub(crate) fn scalar(component: u8, reciprocal: u16) -> u8 {
+        (u16::from(component) * reciprocal).div_256() as u8
```
grebmeg: Could we handle non-canonical premultiplied input here? A consumer can construct a `Pixmap` containing `[2, 0, 0, 1]`, causing `take_unpremultiplied()` to overflow and panic in the scalar debug path. What do you think about either saturating these values or documenting and enforcing the RGB ≤ alpha requirement?

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -395,28 +395,11 @@ impl Pixmap {
...
-                        b: unpremultiply(b),
-                        a,
-                    }
-                } else {
-                    Rgba8 { r, g, b, a }
-                }
-            })
-            .collect()
+    pub fn take_unpremultiplied(mut self) -> Vec<Rgba8> {
+        unpremultiply_rgba8(bytemuck::cast_slice_mut(&mut self.buf));
```
grebmeg: Could we skip `unpremultiply_rgba8` when `may_have_transparency` is `false`?

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -462,6 +445,39 @@ fn premultiply_rgba8(data: &mut [u8]) -> bool {
...
+    let level = Level::try_detect().unwrap_or(Level::baseline());
+
+    dispatch!(level, simd => unpremultiply_rgba8_impl(simd, data));
+}
+
+#[inline(always)]
+fn unpremultiply_rgba8_impl<S: Simd>(simd: S, data: &mut [u8]) {
+    let (body, tail) = data.as_chunks_mut::<64>();
+
+    for chunk in body {
```
grebmeg, line 464: Could we add an opaque fast path per SIMD chunk? Many images have large opaque regions, so checking whether all alpha lanes are 255 could avoid the reciprocal gather and multiplications for those chunks. It would also be useful to benchmark opaque and spatially clustered transparency separately.

sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -462,6 +445,39 @@ fn premultiply_rgba8(data: &mut [u8]) -> bool {
     dispatch!(level, simd => premultiply_rgba8_impl(simd, data))
 }
 
+/// Unpremultiplies each RGBA8 pixel in `data`.
+fn unpremultiply_rgba8(data: &mut [u8]) {
+    let level = Level::try_detect().unwrap_or(Level::baseline());
```
grebmeg, line 455: Could the detected SIMD level be cached or supplied by the caller? `Level::try_detect()` currently runs for every conversion, so avoiding repeated detection may help small pixmaps where fixed overhead is significant.

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: Yep! It won't help much for now because when rendering we always set this to true, but in a future where we detect opaque backgrounds it will hopefully be helpful.

sparse_strips/vello_common/src/util.rs
LaurenzV: I agree that we should document the invariant on the pixmap constructor! If it's just about the panic, then happy to change it to saturating (or rather `wrapping_mul`, since this is what's used in fearless_simd). I'm a bit hesitant to try to handle this correctly for the SIMD path, because there aren't dedicated SIMD instructions for saturating multiplication I believe, so we would have to clamp it manually, which would be a bit unfortunate.

So my proposal would be to use `wrapping_mul` to get rid of the panic, but not make any correctness guarantees for this case? What do you think? Internally, we should always be upholding the invariant already.

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: Yes we can, but I did some testing, and while overall it seems to be worth it for fully opaque images, the advantage is much less visible with mixed images, likely due to the effect of the branch predictor. Therefore, I'd like to add this in a follow-up instead so I can evaluate it more carefully.

sparse_strips/vello_common/src/pixmap.rs
LaurenzV: I would suggest we just upgrade to new fearless_simd soon so we can profit off https://github.com/linebender/fearless_simd/pull/278. Alternatively, we could have the user provide the level, but for consistency the same should exist for `premultiply`, so maybe better in a follow-up?

grebmeg: Sure!

sparse_strips/vello_common/src/pixmap.rs
grebmeg: Makes sense! Thanks!

sparse_strips/vello_common/src/util.rs
grebmeg: Okay, let's go with your proposed approach.

grebmeg (approved):

PR #1842, opened 2026-08-22
vello_bench: Add more varied pixmap benchmarks

LaurenzV:
This PR extend the existing benchmarks to cover three different cases: A fully opaque image, a fully translucent image and an interleaved image, where the two alternate. This is in preparation for a follow-up PR which adds some more optimizations to show that they are worth it.

### DIFF
--- a/sparse_strips/vello_bench/src/pixmap.rs
+++ b/sparse_strips/vello_bench/src/pixmap.rs
@@ -16,6 +16,7 @@ pub fn pixmap(c: &mut Criterion) {
         ("opaque", OPAQUE_BLUE.repeat(pixel_count)),
         ("translucent", TRANSLUCENT_BLUE.repeat(pixel_count)),
         ("interleaved", interleaved_pixels(pixel_count)),
+        ("mixed_lanes", mixed_lane_pixels(pixel_count)),
     ];
 
     let mut group = c.benchmark_group("pixmap/premultiply");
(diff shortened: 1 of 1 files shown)


LaurenzV:
Hmm okay, I'm not completely sure whether the optimization is going to be worth it. But I think it's useful regardless to have some more varied benchmarks for this!

sparse_strips/vello_bench/src/pixmap.rs
```diff
@@ -7,43 +7,70 @@ use vello_common::pixmap::{PixelMetadata, Pixmap};
...
+            );
+        });
+    }
     group.finish();
 }
+
+fn interleaved_pixels(pixel_count: usize) -> Vec<u8> {
+    let mut rgba = Vec::with_capacity(pixel_count * 4);
+    for index in 0..pixel_count {
+        let pixel = if (index / 16).is_multiple_of(2) {
```
grebmeg, line 69: The SIMD kernel processes 16 pixels per chunk, so this makes every vector uniformly opaque or translucent. Could we alternate every eight pixels, or add a separate mixed-lanes case? That would preserve the previous benchmark’s coverage of mixed alpha values within one SIMD iteration.

grebmeg (approved):

sparse_strips/vello_bench/src/pixmap.rs
LaurenzV: Added a new benchmark with mixed alpha values in a single SIMD block.

PR #1845, opened 2026-08-25
sparse_strips: Fix issues on big-endian targets

LaurenzV:
First commit adds a CI step that runs (only scalar) tests from `vello_sparse_tests` on a big-endian target. As you can see, it's completetly broken.

Second commit fixes this by making the `u8_to_f32` and `f32_to_u8` methods aware of endianness

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -346,40 +346,6 @@ jobs:
       - name: cargo test --doc
         run: cargo test --doc --workspace --locked --all-features --no-fail-fast
 
-  test-stable-big-endian:
-    name: cargo test (big-endian, powerpc64)
-    needs: prime-lfs-cache
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-
-      - name: Restore lfs cache
-        id: lfs-cache
-        uses: actions/cache/restore@v4
-        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -346,6 +346,40 @@ jobs:
...
+
+      - name: install cross
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cross
+
+      - name: cargo test
+        run: cross test -p vello_sparse_tests --test tests --locked --release --target powerpc64-unknown-linux-gnu _scalar
+        env:
+          CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS: "true"
```
grebmeg: Given how rare big-endian targets are and our already long, capacity-constrained CI, does this justify running the full scalar suite on every PR? Also, I'm cursiour, how did you discover this issue originally?

LaurenzV: Sure, as mentioned in the Zulip thread, happy to remove the CI part. To be honest I don't really remember anymore. 😅 I think it was when reviewing a fearless_simd PR where we discovered an issue with endianness and that's how I thought of taking a look at whether vello has similar issues.

grebmeg (approved):

PR #1846, opened 2026-08-26
perf: minify generated sparse shaders

grebmeg:
Minify linked WGSL and generated GLSL by shortening Naga IR identifiers and safely removing redundant whitespace.

For example, generated code changes from:

```wgsl
fn verbose_helper(value: u32) -> u32 {
    let incremented = value + 1u;
    return incremented;
}
```

to approximately:

```wgsl
fn A(B:u32)->u32{let C=B+1u;return C;}
```

Renderer-facing entry points such as `vs_main` and `fs_main` remain unchanged. Authored resource names are retained in reflection metadata. This also removes the redundant WESL mangler and updates generated-output tooling and documentation.

#### Size impact
The generated `compiled_shaders.rs` module decreases by 62,488 bytes (44.6%):

```text
main:      140,260 bytes
this PR:    77,772 bytes
```

Measured on `main` from Cargo’s generated output after:

```sh
cargo build -p vello_sparse_shaders --features glsl
```

The PR result was measured from `generated_glsl/compiled_shaders.rs` after generation.


LaurenzV (approved):
Nice!! To be honest, I have not tried to understand the full reparsing/minimization logic myself, but I tested it on some different browsers and they seem to work fine. It does feel pretty hacky to parse it like that. 😄 But it doesn't seem like naga provides anything better here.

However, I think it would be good to still have a way to create non-minified shaders via the CLI, so that we can still inspect the raw GLSL output in a readable way. Would it be possible to add something like this with another feature flag in the generator? Doesn't have to be in this PR, could be a follow-up too.

taj-p:
> main:      140,260 bytes
> this PR:    77,772 bytes

So good!!

grebmeg:
Agreed — the serialize/reparse step is a little awkward, but Naga doesn’t currently expose identifier mangling or minified output. I'll add a CLI option to generate non-minified shaders.

PR #1847, opened 2026-08-26
Re-export vello_encoding from the low_level module

### LINKED ISSUE #1290: `vello_encoding` is not re-exported from `vello`
I'm just starting to play around with vello, and I noticed that it diligently re-exports `peniko` and `kurbo`, but not `vello_encoding`.   
For example, `Scene::encoding()` is still usable, but its return type can't be spelled out without depending on `vello_encoding` as well.  
Sorry if this is intended behavior, but it struck me as odd.

ChrisJr404:
Closes #1290. This re-exports the `vello_encoding` crate from `vello::low_level` so consumers building an alternative renderer can reach the encoding types using the same version vello was built against, as suggested in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 This release has an [MSRV][] of 1.88.
 
+### Added
+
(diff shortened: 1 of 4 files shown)


DJMcNab (approved):
Thanks. I've pushed a few tweaks

PR #1849, opened 2026-08-27
vello_bench: Reduce the number of benchmarks run by default

LaurenzV:
A small annoyance for me has been that we have a lot of benchmarks, therefore, running the whole benchmark suite takes really long. But in reality, there are only a few benchmarks that we really care about _in the usual case_ when trying to figure out whether there are any major regressions (in my case, as part of updating `fearless_simd`). Therefore, this PR proposes running only a small subset of benchmarks by default, and requiring the user to opt-in explicitly to run the whole benchmark suite.

### DIFF
--- a/sparse_strips/vello_bench/src/lib.rs
+++ b/sparse_strips/vello_bench/src/lib.rs
@@ -19,6 +19,6 @@ pub mod strip;
 pub mod tile;
 
 pub(crate) const SEED: [u8; 32] = [0; 32];
-pub(crate) static EXTENDED: LazyLock<bool> = LazyLock::new(|| std::env::var("EXTENDED").is_ok());
+pub(crate) const EXTENDED: bool = cfg!(feature = "extended");
 pub static DATA_PATH: LazyLock<PathBuf> =
     LazyLock::new(|| PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("data"));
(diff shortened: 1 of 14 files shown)


sparse_strips/vello_bench/src/lib.rs
```diff
@@ -19,5 +19,6 @@ pub mod strip;
 pub mod tile;
 
 pub(crate) const SEED: [u8; 32] = [0; 32];
+pub(crate) static EXTENDED: LazyLock<bool> = LazyLock::new(|| std::env::var("EXTENDED").is_ok());
```
grebmeg: Could we model the extended suite as a separate Cargo benchmark target instead of checking an environment variable throughout the benchmark implementations?

```toml
[features]
extended = []

[[bench]]
name = "extended"
harness = false
required-features = ["extended"]
```

`benches/main.rs` would register core benchmarks, while `benches/extended.rs` would register the additional ones. Then `cargo bench` runs the core suite and `cargo bench --features extended` runs both, keeping the selection centralized and avoiding ambiguous behavior such as `EXTENDED=0` enabling everything.

LaurenzV: I've changed it now so that it's a feature flag. However, I'd like to avoid having two entry points because then we need to duplicate the method calls and it also makes it harder to only run parts of a single method in extended mode. Is the current version okay?

grebmeg: Yes, the current version looks good to me. Thanks so much!

grebmeg (approved):
LGTM

PR #1850, opened 2026-08-27
vello_cpu: Detect opaque backgrounds and handle them more efficiently

LaurenzV:
# Problem

The main motivation for this PR is that currently, when running `take_unpremultiplied` on a pixmap that we rendered into with vello_cpu, we always unpremultiply all pixels. However, this is completely unnecessary in case all pixels are already opaque. If we had that information, we could skip unpremultiplication completely in the common case where the whole background has a specific opaque color. 

# Solution

This PR aims to address this by detecting backgrounds that are drawn with `fill_rect` as viewport-sized rects before anything else is drawn. A question I spent quite some time thinking about is whether we should have an explicit `set_background` method instead of trying to detect this. While semantically it seems cleaner to treat the background separately, in the end I decided for most users it's much more convenient if this just happens automatically, because the background often isn't stored explicitly. 

In order to achieve this PR introduces a new `RootBackground` struct that contains the main logic for storing and updating the background and wires it up into `vello_cpu`. I've put this into vello_common because I think there is a chance that we can reuse this in vello_hybrid, but it's a bit more tricky (and also depends on what we do with https://github.com/linebender/vello/pull/1783). 

In addition, with this information, when rasterizing into a pixmap we now also have the possibility of updating its `may_have_transparency` flag. This is what allows us to skip premultiplication when the user fetches the RGB bytes.

I've also added a bunch of tests because there is a relatively high risk of missing some edge case in the background detection, but I hope that I've considered everything in this implementation.

This PR was done with assistance of GPT 5.6 Sol.e

### DIFF
--- a/sparse_strips/vello_cpu/src/render.rs
+++ b/sparse_strips/vello_cpu/src/render.rs
@@ -30,6 +30,7 @@ use vello_common::paint::{ImageId, ImageResolver, Paint, PaintType, Tint};
 use vello_common::peniko::color::palette::css::BLACK;
 use vello_common::peniko::{BlendMode, Fill};
 use vello_common::pixmap::{Pixmap, PixmapMut};
+use vello_common::record::RootBackground;
 use vello_common::render_state::RenderState;
 use vello_common::transforms::{RootTransforms, Transforms};
 use vello_common::util::is_axis_aligned;
@@ -170,6 +171,7 @@ pub struct RenderContext {
     pub(crate) aliasing_threshold: Option<u8>,
     pub(crate) encoded_paints: Vec<EncodedPaint>,
     pub(crate) filter: Option<Filter>,
+    root_background: RootBackground,
     #[cfg_attr(
         not(feature = "text"),
         allow(dead_code, reason = "used when the `text` feature is enabled")
@@ -231,6 +233,7 @@ impl RenderContext {
         let encoded_paints = vec![];
         let temp_path = BezPath::new();
         let aliasing_threshold = None;
+        let scene_viewport = Rect::new(0.0, 0.0, f64::from(width), f64::from(height));
 
         Self {
             width,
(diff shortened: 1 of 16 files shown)


sparse_strips/vello_cpu/src/render.rs
```diff
@@ -321,13 +348,19 @@ impl RenderContext {
...
             // Fast path: Use optimized rect filling if we have no skew in the path transform
             // and anti-aliasing is enabled.
             // TODO: Maybe also support no anti-aliasing in the fast path
             if is_axis_aligned(&transform) && ctx.aliasing_threshold.is_none() {
                 // Transform the rect to screen coordinates.
                 let transformed_rect = transform.transform_rect_bbox(*rect);
+
+                if ctx.try_extract_background(transformed_rect, &paint) {
+                    return;
+                }
```
grebmeg, line 362: Would a separate `clear` or `set_background` method be simpler than inferring the background from drawing calls? The current approach adds blend tracking to every draw and eligibility checks to each fast-path `fill_rect`, even when no background is extracted. An explicit API, similar to Skia’s `SkCanvas::clear`, could avoid that overhead and make the intent unambiguous.

grebmeg: My concern with inferring the background is that it adds blend tracking to every draw and eligibility checks to each fast-path `fill_rect`, even when the optimization is unused. It is also dependent on draw order and API shape, so an equivalent `fill_path` or a later full-screen fill will not benefit. This introduces substantial state and plumbing for behavior that an explicit `clear` or `set_background` method could express more predictably.

grebmeg: The detection is order-sensitive. For example, an opaque full-viewport `fill_rect` after a small draw replaces all earlier content, but `is_empty()` prevents it from becoming the background. A dedicated background operation `clear` or `set_background` could explicitly define that replacement and discard earlier commands without relying on draw-order inference.

LaurenzV: > My concern with inferring the background is that it adds blend tracking to every draw and eligibility checks to each fast-path fill_rect, even when the optimization is unused.

In which way? The method is marked as inline and it's just a few boolean checks, so I don't think this is going to matter much. Or are you referring to something else?

> It is also dependent on draw order and API shape, so an equivalent fill_path or a later full-screen fill will not benefit.

I get the point about `fill_path`, but not about the draw order: If you have a background, then it's always going to be the first thing you draw anyway, no? 🤔 When would you want to update the background mid-render? I don't see this as being an actual problem. 😄 

> This introduces substantial state and plumbing for behavior that an explicit clear or set_background method could express more predictably.

I agree the expliciteness is better! However, you still need to keep track of state: You need to keep track of the background color and also whether there was a blending operation against the background. So essentially, you only avoid storing `viewport` and `only_opaque`. Which is less, but I don't think the main source of complexity is coming from those two.

LaurenzV:
We decided that an explicit approach is better. So this PR is superseded by #1866.

PR #1852, opened 2026-08-28
feat: add GPU timing to WebGL browser examples

grebmeg:
Add asynchronous GPU execution timing to the native WebGL2 and wgpu-WebGL browser examples using disjoint timer queries.
 
- Vello Hybrid WebGL backend - Tiger scene 1

image

- Vello Hybrid WebGL backend - Filters scene 7

image


LaurenzV (approved):
Only tested on my MacBook but seems to work fine!

PR #1855, opened 2026-08-28
vello_common: Add support for requesting RGB bytes from a pixmap

LaurenzV:
As the title says, this PR refactors `Pixmap` to have a single `take` API that allows users to configure 1) whether to return premultiplied or unpremultiplied bytes and 2) whether to return the data as RGBA or RGB (when possible). We also refactor the existing `into_png` implementation to make use of this so it's also implictly coverted in the tests.

### DIFF
--- a/sparse_strips/vello_common/src/pixmap.rs
+++ b/sparse_strips/vello_common/src/pixmap.rs
@@ -8,11 +8,8 @@ use alloc::vec::Vec;
 #[cfg(feature = "png")]
 use std::io::{BufRead, Seek};
 
-use crate::fearless_simd::{Level, dispatch, mask8x16, prelude::*, u8x16, u16x16};
-use crate::peniko::{
-    ImageAlphaType,
-    color::{PremulRgba8, Rgba8},
-};
+use crate::fearless_simd::{Level, dispatch, mask8x16, prelude::*, u8x16, u8x32, u8x64, u16x16};
+use crate::peniko::{ImageAlphaType, color::PremulRgba8};
 use crate::util::{Div255Ext, narrow, unpremultiply, widen};
 
 #[cfg(feature = "png")]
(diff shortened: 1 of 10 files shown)


sparse_strips/vello_common/src/pixmap.rs
```diff
@@ -310,12 +307,19 @@ impl Pixmap {
     /// Return the current content of the pixmap as a PNG.
     #[cfg(feature = "png")]
     pub fn into_png(self) -> Result<Vec<u8>, png::EncodingError> {
+        let width = u32::from(self.width);
+        let height = u32::from(self.height);
+        let pixels = self.take(PixelBufferOptions::RGB8);
```
grebmeg: I’m concerned that requesting RGB8 may silently return RGBA8 depending on the pixmap’s transparency. Could the format request instead be strict: RGB8 always returns RGB8 after unpremultiplying and removing alpha, while transparency information is reported separately? An enum would distinguish an exact result from an unverified hint:

```rust
pub enum SourceTransparency {
    /// Every source pixel is known to be fully opaque.
    Opaque,
    /// At least one source pixel was found to be non-opaque.
    HasTransparency,
    /// The source may contain non-opaque pixels, but its pixels were not inspected.
    MayHaveTransparency,
}

pub struct PixelBuffer {
    pub data: Vec<u8>,
    pub format: PixelBufferFormat,
    pub source_transparency: SourceTransparency,
}
```

This ensures consumers always receive the requested format while still knowing whether transparency existed or was merely possible in the source pixmap.

LaurenzV: But what if the original pixmap had transparency but you request RGB? Then the alpha channel was (invalidly) discarded, and you cannot recover it anymore.

The only option I see to do one more pass across the whole pixmap, once for redetermining the transparency flag and then another time for doing unpremultiplication and/or RGB conversion, which would be a bit unfortunate in my opinion.

And I don't think it's _silently_ returning RGBA8, it's the reason why the `PixelBuffer` also returns a pixel format and why this is documented explicitly. Is this not visible enough in your opinion?

LaurenzV:
@grebmeg I have changed this now so that

1) There is an explicit `take_rgba8` method that always returns just the raw RGBA data.
2) The other method is called `try_take_rgb8`, and returns an explicit enum for the data. This way, the caller is forced toe explicitly match on the enum and it should become "impossible" to accidentally miss this. Let me know what you think.

grebmeg (approved):
LGTM! Thanks for looking into this again — I think it’s much clearer now. Thanks!

PR #1858, opened 2026-08-31
vello_hybrid: Allow binding multiple external textures per draw call

LaurenzV:
This PR adds support for binding up to 4 external textures in a single draw call. This is achieved by simply making use of more shader slots (which WebGL2 guarantees to support at least 16, so we are still well below the limit) and changing the CPU batching algorithm to support multiple external images in a single pass. This reduces the number of draw calls emitted, but does make the shader more branchy. Once we move to specialized shaders, we should carefully re-evaluate whether the branching makes image rendering slower, but for our current uber shader it does not seem to make a difference. And our previous experiments have shown that the CPU-side overhead of draw calls can also be considerable, so in this regard this PR is an improvement.

The main motivation for this PR however is to enable the follow-up PR that removes 2D texure arrays to keep the draw count minimal.

### DIFF
--- a/sparse_strips/vello_hybrid/src/draw.rs
+++ b/sparse_strips/vello_hybrid/src/draw.rs
@@ -468,18 +468,20 @@ impl ExternalTextureBindings {
     /// Returns `None` when all four slots are occupied by other textures.
     #[inline]
     fn get_or_insert(&mut self, texture_id: TextureId) -> Option<u8> {
-        if let Some(slot) = self
-            .texture_ids
-            .iter()
-            .position(|candidate| *candidate == Some(texture_id))
-        {
-            return Some(u8::try_from(slot).unwrap());
+        // This iteration order assumes slots are assigned without leaving "holes" in-between,
+        // i.e. if we hit `None`, any later slot is also guaranteed to be `None`.
+        for (slot, candidate) in self.texture_ids.iter_mut().enumerate() {
+            match *candidate {
+                Some(id) if id == texture_id => return Some(u8::try_from(slot).unwrap()),
+                None => {
+                    *candidate = Some(texture_id);
+                    return Some(u8::try_from(slot).unwrap());
+                }
+                _ => {}
+            }
         }
 
-        let slot = self.texture_ids.iter().position(Option::is_none)?;
-        self.texture_ids[slot] = Some(texture_id);
-
-        Some(u8::try_from(slot).unwrap())
+        None
     }
 
     #[inline]
--- a/sparse_strips/vello_hybrid/src/draw.rs
+++ b/sparse_strips/vello_hybrid/src/draw.rs
@@ -4,7 +4,9 @@
 //! Draw construction for strip render passes.
 
 use crate::GpuStrip;
-use crate::paint::{COLOR_SOURCE_LAYER, COLOR_SOURCE_SHIFT, PaintResolver};
+use crate::paint::{
+    COLOR_SOURCE_LAYER, COLOR_SOURCE_SHIFT, EXTERNAL_TEXTURE_SLOT_SHIFT, PaintResolver,
+};
 use crate::rect::{RectPart, split_rect};
 use crate::scene::{RecordedDraw, RecordedPath};
 use crate::target::{DrawTarget, LayerTextureRegion};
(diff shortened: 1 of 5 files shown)


sparse_strips/vello_sparse_shaders/shaders/render.wesl
```diff
@@ -223,7 +225,16 @@ var<uniform> config: Config;
...
+var external_texture_0: texture_2d<f32>;
+
+@group(1) @binding(2)
+var external_texture_1: texture_2d<f32>;
+
+@group(1) @binding(3)
+var external_texture_2: texture_2d<f32>;
+
+@group(1) @binding(4)
+var external_texture_3: texture_2d<f32>;
```
grebmeg, line 237: I expected this to be build-time-specialized from a single `MAX_EXTERNAL_TEXTURES` value rather than hardcoded to 4. With 1, we should generate one binding and sample it directly; with N > 1, generate N bindings and the matching static selection logic. The CPU batching and backend layouts should derive from the same value, so each build still contains only one matching shader variant. What do you think?

LaurenzV: Hmm but how exactly? Is there some WESL feature for that? If not, how do you want to make shader code dependent on a constant in Rust?

grebmeg: My main wish here is to have some form of shader parameterization, close to the planned specialized-shader work. It could generate N bindings and disable unused features such as blurred rounded rectangles, while also letting us experiment with simpler cases like one versus multiple bindings. That said, this would make shader generation significantly more complex, so I agree that we shouldn’t do it in this PR.

sparse_strips/vello_sparse_shaders/shaders/render.wesl
```diff
@@ -223,7 +225,16 @@ var<uniform> config: Config;
 var atlas_texture_array: texture_2d_array<f32>;
```
grebmeg, line 225: The paint index is now limited to bits 0–23, but above `StripInstance` still document bits 0–25. Should we update lines there?

sparse_strips/vello_hybrid/src/draw.rs
```diff
@@ -444,32 +450,77 @@ impl LayerTextureRegion {
...
+            .position(|candidate| *candidate == Some(texture_id))
+        {
+            return Some(u8::try_from(slot).unwrap());
+        }
+
+        let slot = self.texture_ids.iter().position(Option::is_none)?;
+        self.texture_ids[slot] = Some(texture_id);
+
+        Some(u8::try_from(slot).unwrap())
+    }
```
grebmeg, line 485: Could this use a single pass? Since occupied slots always form a packed prefix, we can return an existing match or insert at the first `None`:

```rust
fn get_or_insert(&mut self, texture_id: TextureId) -> Option<u8> {
    for (slot, candidate) in self.texture_ids.iter_mut().enumerate() {
        match *candidate {
            Some(id) if id == texture_id => {
                return Some(u8::try_from(slot).unwrap());
            }
            None => {
                *candidate = Some(texture_id);
                return Some(u8::try_from(slot).unwrap());
            }
            _ => {}
        }
    }

    None
}
```

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -2340,7 +2340,12 @@ fn get_strip_uniforms(gl: &WebGl2RenderingContext, program: &Program) -> StripUn
     let encoded_paints_texture_fs_name = render::fragment::ENCODED_PAINTS_TEXTURE;
     let encoded_paints_texture_vs_name = render::vertex::ENCODED_PAINTS_TEXTURE;
     let gradient_texture_name = render::fragment::GRADIENT_TEXTURE;
-    let external_texture_name = render::fragment::EXTERNAL_TEXTURE;
+    let external_texture_names = [
+        render::fragment::EXTERNAL_TEXTURE_0,
+        render::fragment::EXTERNAL_TEXTURE_1,
+        render::fragment::EXTERNAL_TEXTURE_2,
+        render::fragment::EXTERNAL_TEXTURE_3,
+    ];
```
grebmeg, line 2349: It'd be great to crete this based on `EXTERNAL_TEXTURE_SLOT_COUNT` or something similar, so that if you change that value to 1, everything still "works" correctly.

grebmeg (approved):
LGTM

sparse_strips/vello_sparse_shaders/shaders/render.wesl
LaurenzV: Ah yes, sorry I wrote that comment before we had the conversation.

sparse_strips/vello_hybrid/src/draw.rs
LaurenzV: Good point!

sparse_strips/vello_sparse_shaders/shaders/render.wesl
LaurenzV: Yep!

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: I don't think it's super straight-forward because the variables are generated in `vello_sparse_shaders`. Will add a TODO for now.

PR #1859, opened 2026-08-31
vello_hybrid: Replace the 2D texture array with individual 2D textures

LaurenzV:
As part of our rollout attempts, we have determined that dynamic use and mutation of 2D texture arrays are apparently too "obscure" and therefore not well-supported on many devices. One some GPUs they can cause complete hangs, on Adreno GPUs copying between the textures is completely broken for 2+ textures, and they also aren't great for the memory footprint.

Therefore, this PR:
- Proposes to remove the 2D texture array for our image atlas.
- Instead, image atlas textures are created and treated as separate 2D textures, and just reuse the existing external texture mechanism. Because of this, we are also able to remove a lot of duplicated code.
- Thanks to the previous PR, we can still have up to 4 atlas layers while still ensuring that everything is drawn in a single call. However, if we end up with more than that, we do still need multiple draw calls. This is a downside compared to the previous approach, but for things like glyph caching, this should hopefully be enough.
- A complication is that we need to ensure the atlas textures (which are managed internally by Vello) don't conflict with user-generated IDs for external textures (currently assigned arbitrarily by the user). Because of this, I changed `TextureId` to rely on a static global counter to hand out unique image IDs, so that accidental conflicts are not possible. Open for other suggestions, but this seemed like the best way to ensure that atlas textures can just be treated as a special kind of external textures.

In the long term, I think the responsibility for handling atlases should be completely moved out of Vello Hybrid, so that there is only one consistent way of binding images in Vello Hybrid (via the external texture API). Users who still want to have atlas-like functionality can simply construct and manage those atlases themselves and use them with the existing external texture API. However, in order to keep the code diff minimal, we don't do this in this PR.

I have tested this PR on various devices with both image APIs and didn't encounter any issues.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -33,7 +33,7 @@ use crate::{
     copy::GpuCopyInstance,
     filter::{FilterContext, FilterInstanceData, FilterPassPlan},
     gradient_cache::GradientRampCache,
-    paint::PaintResolver,
+    paint::{PaintResolver, TextureSourceId},
     render::{
         Config,
         common::{
@@ -91,18 +91,11 @@ const GPU_PAINT_PLACEHOLDER: GpuEncodedPaint = GpuEncodedPaint::LinearGradient(G
 });
 
 /// First texture unit from which the strip program samples external textures.
-const EXTERNAL_TEXTURE_UNIT_START: u32 = 5;
+const EXTERNAL_TEXTURE_UNIT_START: u32 = 4;
 
 /// Query the WebGL context for the max texture size.
 fn get_max_texture_dimension_2d(gl: &WebGl2RenderingContext) -> u32 {
     gl.get_parameter(WebGl2RenderingContext::MAX_TEXTURE_SIZE)
-        .unwrap()
-        .as_f64()
-        .unwrap() as u32
-}
-
-fn get_max_texture_array_layers(gl: &WebGl2RenderingContext) -> u32 {
-    gl.get_parameter(WebGl2RenderingContext::MAX_ARRAY_TEXTURE_LAYERS)
         .unwrap()
         .as_f64()
         .unwrap() as u32
@@ -201,15 +194,15 @@ impl WebGlTextureBindings {
     }
 }
 
-/// Current allocation state of the WebGL image/glyph atlas texture.
+/// Current allocation state of the WebGL image/glyph atlas textures.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct AtlasTextureInfo {
-    /// Width of the currently bound texture array.
+    /// Width of each atlas texture.
     pub width: u16,
-    /// Height of the currently bound texture array.
+    /// Height of each atlas texture.
     pub height: u16,
-    /// Number of real atlas layers available to the image cache.
-    pub layer_count: u32,
+    /// Number of atlas textures available to the image cache.
+    pub texture_count: u32,
 }
 
 impl WebGlRenderer {
@@ -364,8 +357,6 @@ impl WebGlRenderer {
         let device_limits = DeviceLimits {
             max_texture_dimension_2d: u16::try_from(get_max_texture_dimension_2d(&gl))
                 .unwrap_or(u16::MAX),
-            max_texture_array_layers: u16::try_from(get_max_texture_array_layers(&gl))
-                .unwrap_or(u16::MAX),
         };
         settings.memory_settings.normalize(&device_limits);
         if use_depth_buffer {
@@ -461,6 +452,7 @@ impl WebGlRenderer {
             true,
             RootTarget::UserSurface,
             texture_bindings,
+            None,
         )?;
 
         #[cfg(feature = "text")]
@@ -503,13 +495,14 @@ impl WebGlRenderer {
         texture_bindings: &WebGlTextureBindings,
     ) -> Result<(), RenderError> {
         self.programs
-            .maybe_resize_atlas_texture_array(&self.gl, atlas_count);
+            .maybe_create_atlas_textures(&self.gl, atlas_count);
 
         let (atlas_width, atlas_height) = atlas_config.atlas_size;
         let atlas_render_size = RenderSize {
             width: u32::from(atlas_width),
             height: u32::from(atlas_height),
         };
+        let render_target_texture = self.atlas_texture(atlas_id).clone();
 
         let atlas_framebuffer = self
             .programs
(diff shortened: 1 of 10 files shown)


sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -521,25 +512,18 @@ impl WebGlRenderer {
...
-        self.gl.framebuffer_texture_layer(
+        self.gl.framebuffer_texture_2d(
             WebGl2RenderingContext::FRAMEBUFFER,
             WebGl2RenderingContext::COLOR_ATTACHMENT0,
-            Some(&self.programs.resources.atlas_texture_array.texture),
+            WebGl2RenderingContext::TEXTURE_2D,
+            Some(&self.programs.resources.atlas_textures[atlas_id.as_u32() as usize]),
             0,
-            atlas_id.as_u32() as i32,
         );
```
grebmeg, line 523: Could we guard against framebuffer–sampler feedback somewhere? A caller can now pass the destination atlas texture through `WebGlTextureBindings`, causing WebGL `INVALID_OPERATION` while `render_to_atlas` still returns `Ok(())`. This test reproduces it:

```rust
#[wasm_bindgen_test]
fn render_to_atlas_does_not_sample_from_its_destination() {
    let canvas = document
        .create_element("canvas")
        .unwrap()
        .dyn_into::<HtmlCanvasElement>()
        .unwrap();
    let atlas_config = AtlasConfig {
        initial_atlas_count: 1,
        max_atlases: 1,
        atlas_size: (1, 1),
        ..AtlasConfig::default()
    };
    let settings = RenderSettings {
        memory_settings: MemorySettings {
            image_atlas_config: atlas_config,
            ..MemorySettings::default()
        },
        ..RenderSettings::default()
    };
    let (mut renderer, _resources) = WebGlRenderer::new_with(&canvas, settings, true);
    let texture_id = TextureId::new();
    let mut bindings = WebGlTextureBindings::new();
    bindings.insert(
        texture_id,
        renderer.atlas_texture(AtlasId::new(0)).clone(),
    );

    let mut scene = Scene::new(1, 1);
    scene.set_paint(Image {
        image: ImageSource::external_texture(
            texture_id,
            RectU16::new(0, 0, 1, 1),
            false,
        ),
        sampler: ImageSampler::default(),
    });
    scene.fill_rect(&Rect::new(0.0, 0.0, 1.0, 1.0));

    renderer
        .render_to_atlas(&scene, 1, atlas_config, AtlasId::new(0), &bindings)
        .unwrap();

    assert_eq!(
        renderer.gl_context().get_error(),
        WebGl2RenderingContext::NO_ERROR
    );
}
```

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
```diff
@@ -3603,71 +3422,32 @@ impl WebGlAtlasWriter for WebGlTextureWithDimensions {
...
-    dest_layer: u32,
+    dest_texture: &WebGlTexture,
     dest_offset: [u32; 2],
     copy_size: [u32; 2],
 ) {
     let _state_guard = WebGlStateGuard::for_texture_copy(gl);
     let read_framebuffer = Framebuffer::new(gl);
 
-    // Bind destination texture array
     gl.active_texture(WebGl2RenderingContext::TEXTURE0);
```
grebmeg, line 3455: Could we preserve texture unit 0’s original binding here? The guard saves the binding from whichever unit is currently active, but this code switches to unit 0 and replaces its binding. Unit 7 is restored correctly in this example, while unit 0 remains bound to the atlas texture.

This regression test reproduces it:

```rust
#[wasm_bindgen_test]
fn texture_copy_restores_texture_zero_binding_when_another_unit_is_active() {
    let settings = RenderSettings {
        memory_settings: MemorySettings {
            image_atlas_config: AtlasConfig {
                initial_atlas_count: 1,
                max_atlases: 1,
                atlas_size: (1, 1),
                ..AtlasConfig::default()
            },
            ..MemorySettings::default()
        },
        ..RenderSettings::default()
    };
    let (mut renderer, mut resources) =
        WebGlRenderer::new_with(&canvas, settings, true);
    let gl = renderer.gl_context().clone();

    let source = solid_texture(&gl, [255, 0, 0, 255]);
    let texture_zero = solid_texture(&gl, [0, 255, 0, 255]);
    let texture_seven = solid_texture(&gl, [0, 0, 255, 255]);

    gl.active_texture(WebGl2RenderingContext::TEXTURE0);
    gl.bind_texture(WebGl2RenderingContext::TEXTURE_2D, Some(&texture_zero));
    gl.active_texture(WebGl2RenderingContext::TEXTURE7);
    gl.bind_texture(WebGl2RenderingContext::TEXTURE_2D, Some(&texture_seven));

    renderer.upload_image(
        &mut resources,
        &WebGlTextureWithDimensions {
            texture: source,
            width: 1,
            height: 1,
        },
    );

    assert_texture_bound(&gl, &texture_seven);
    gl.active_texture(WebGl2RenderingContext::TEXTURE0);
    assert_texture_bound(&gl, &texture_zero);
}
```

sparse_strips/vello_common/src/lib.rs
```diff
@@ -63,6 +63,8 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
+use core::sync::atomic::{AtomicU64, Ordering};
```
grebmeg: Could we use `AtomicU32`? Realistic Vello workloads should never approach 2^32 texture IDs, and this would preserve support for common 32-bit no_std targets where `AtomicU64` is unavailable.

sparse_strips/vello_common/src/lib.rs
```diff
@@ -104,4 +106,27 @@ pub use peniko::kurbo;
...
+    #[expect(clippy::new_without_default, reason = "better to be explicit here")]
+    pub fn new() -> Self {
+        static NEXT_TEXTURE_ID: AtomicU64 = AtomicU64::new(0);
+
+        let id = NEXT_TEXTURE_ID
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |id| id.checked_add(1))
+            .expect("texture ID space exhausted");
+        Self(id)
+    }
+}
```
grebmeg: Could we avoid global atomics by making IDs binding-scoped?

```rust
let texture_id = bindings.insert(texture);
```

Scheduling could then distinguish internal atlases from external textures directly:

```rust
enum TextureSourceId {
    Atlas(AtlasId),
    External(TextureId),
}
```

This models ownership more directly and prevents internal/external ID collisions without a global counter. What do you think?

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: I agree it's unfortunate, the reason I "ignored" this here is 1) For now, we don't intend `render_to_atlas` to be used publically (hence why it's marked as `#[doc(hidden)]`), and for our glyph caching use cases we don't ever reference images, so it can't happen there.

But it's probably good to catch it, so let me look into it.

sparse_strips/vello_common/src/lib.rs
LaurenzV: hmm... could we feature-gate it? So use `AtomicU32` if it's not available, otherwise use `AtomicU64`? Yes realistically we shouldn't ever hit this limit for single users, but what if someone runs Vello Hybrid in a long-running process on a server? Not unreasonable in that case, IMO.

sparse_strips/vello_common/src/lib.rs
LaurenzV: That's what I used originally, but I kind of liked the fact that atlases are really just treated the same way as external textures. But if this is your preference I will change it back.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: I can do it for now if you want, but I do want to point out that there are plenty of other places that modify `Texture0` right now without restoring what was there before, so we aren't very consistent with this in the first place. 😄

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Added such a check now.

sparse_strips/vello_common/src/lib.rs
LaurenzV: Not relevant anymore since I removed this now.

sparse_strips/vello_common/src/lib.rs
LaurenzV: Changed it to this.

sparse_strips/vello_hybrid/src/render/webgl/mod.rs
LaurenzV: Implemented, though it's a bit ugly in my opinion. 😅

grebmeg (approved):
Nice! Thank you! LGTM

PR #1864, opened 2026-08-31
vello_hybrid: Use immutable WebGL texture storage everywhere

LaurenzV:
On Apple Silicon machines with Chrome 127, probing currently fails and Vello Hybrid renders layers incorrectly. The root cause is [a bug in ANGLE](https://issues.chromium.org/issues/355605685).

The bug requires the combination of:

1. Defining mutable texture storage with `texImage2D`.
2. Setting `TEXTURE_MAX_LEVEL`.
3. Rendering into the texture.
4. Sampling from it afterward.

One workaround is to stop setting `TEXTURE_MAX_LEVEL`. However, keeping the valid mip range explicit is useful defensively, as some drivers have failed to clamp the effective maximum level correctly for immutable textures.

Instead, this switches all Vello-owned WebGL textures to immutable storage using `texStorage2D`, which is also the generally recommended allocation path. 

With this fix, Vello Hybrid works correctly on the affected browser versions.

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -2393,24 +2393,27 @@ fn create_texture_inner(
         WebGl2RenderingContext::TEXTURE_WRAP_T,
         WebGl2RenderingContext::CLAMP_TO_EDGE as i32,
     );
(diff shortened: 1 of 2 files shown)


grebmeg (approved):
LGTM

PR #1865, opened 2026-09-01
vello_cpu: Isolate root draw blend modes

LaurenzV:
This PR implements two changes:

Each draw also holds information about blending, and we use this to determine a more accurate `root_is_blend_target` flag for Vello CPU, where paths themselves can also have blend modes. For Vello Hybrid this doesn't change anything because paths themselves can't have blend modes themselves, only layers.

When rendering, we ensure that the background (if it's not just transparent) is isolated if the root is a blend target. What this means is that, in case our new scene has a non-default blend at the root, it cannot modify the background that was loaded from the previous pixmap contents. 

It's not necessarily the case that the previous behavior was "incorrect", one could also argue the usefulness of the previous behavior, but I believe this new approach matches more precisely how a background should behave (especially for the follow-up PRs where you can set a background color) and it aligns us with what Vello Hybrid does: In Vello Hybrid, when rendering onto a framebuffer with already drawn contents, we use the GPU to load them as the backdrop for whatever our new scene contains. However, a non-default blend mode in our new scene can never modify the existing backdrop, while this was possible in Vello CPU up until now. It's not even possible to implement that same behavior in Vello Hybrid even if we wanted to, because we cannot sample from the main drawing buffer.

We only perform the isolation if there really is a non-default blend operation against the root and there actually is a non-default background, so the normal case will not have anyperformance penalty associated with it. There is no test for this now, but in a later PR I will add one that ensures Vello Hybrid and Vello CPU are consistent here (test `root_blend_target_is_isolated`).

### DIFF
--- a/sparse_strips/vello_common/src/record.rs
+++ b/sparse_strips/vello_common/src/record.rs
@@ -47,6 +47,9 @@ pub trait Drawable {
     /// Return the **tile-aligned** bounding box of the given object, if it
     /// has one.
     fn bbox(&self, strips: &[Strip]) -> Option<RectU16>;
+
+    /// Return the blend mode applied directly by this draw, if any.
+    fn blend_mode(&self) -> Option<&BlendMode>;
 }
 
 /// A node in the recorded render graph.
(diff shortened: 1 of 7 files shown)


grebmeg (approved):
From a memory perspective, the current CPU path is cheaper because it uses only the existing row buffer, while isolation requires an additional temporary row buffer and Hybrid similarly needs a scratch texture? Still, since this only affects non-default blend modes and makes CPU semantics consistent with Hybrid, if you think the extra memory is worth that consistency, I’m fine with this approach.

LaurenzV:
Yep, your understanding is right. For our internal use, we should never hit this as our only usage of blending is in COLR glyphs, which we always isolate themselves.

PR #1866, opened 2026-09-01
vello_cpu: Support configurable target clears

LaurenzV:
This PR replaces `CompositeMode` with a more generic `TargetInit` enum that also allows specifying the exact clear color, and is generic so that it can be used by the Vello Hybrid backend as well. As a consequence, users now have a way of setting a background color when rendering scenes with Vello CPU.

### DIFF
--- a/vello_cpu/src/fine/mod.rs
+++ b/vello_cpu/src/fine/mod.rs
@@ -28,6 +28,8 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 use core::iter;
+use vello_common::TargetInit;
+use vello_common::color::AlphaColor;
 use vello_common::encode::{
     EncodedBlurredRoundedRectangle, EncodedGradient, EncodedImage, EncodedKind, EncodedPaint,
 };
@@ -469,7 +471,7 @@ pub(crate) fn rasterize_region<S: Simd, T: FineKernel<S>>(
     region: &mut Region<'_>,
     bucketer: &CommandBucketer,
     resources: FineResources<'_>,
-    mut unpack_dest: bool,
+    mut target_init: TargetInit<PremulColor>,
     root_is_blend_target: bool,
 ) {
     let scene_y = region.row_idx as u16 * Tile::HEIGHT;
@@ -480,12 +482,17 @@ pub(crate) fn rasterize_region<S: Simd, T: FineKernel<S>>(
     depth.clear();
 
     let has_cmds = !row.depth_cmds.is_empty() || !row.render_cmds.is_empty();
-    let isolate_bg = unpack_dest && root_is_blend_target && has_cmds;
+    let isolate_bg = root_is_blend_target
+        && has_cmds
+        && match target_init {
+            TargetInit::SrcOver => true,
+            TargetInit::Clear(color) => !color.is_transparent(),
+        };
     if isolate_bg {
-        fine.init_uncovered_range(span, region, unpack_dest, depth);
+        fine.init_uncovered_range(span, region, target_init, depth);
         fine.push_buf(None);
 
-        unpack_dest = false;
+        target_init = TargetInit::Clear(PremulColor::from_alpha_color(AlphaColor::TRANSPARENT));
     }
 
     // Render depth-buffer commands front-to-back, with depth-buffer read and write.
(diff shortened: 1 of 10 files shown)


vello_cpu/src/fine/mod.rs
```diff
@@ -480,12 +482,17 @@ pub(crate) fn rasterize_region<S: Simd, T: FineKernel<S>>(
     depth.clear();
 
     let has_cmds = !row.depth_cmds.is_empty() || !row.render_cmds.is_empty();
-    let isolate_bg = unpack_dest && root_is_blend_target && has_cmds;
+    let isolate_bg = root_is_blend_target
+        && has_cmds
+        && match target_init {
+            TargetInit::SrcOver => true,
+            TargetInit::Clear(color) => !color.is_transparent(),
```
grebmeg, line 489: Should `TargetInit::Clear(color)` initialize the drawing backdrop directly rather than trigger isolation? With `Clear(BLUE)`, a root `Normal + Clear` operation should clear the covered pixel to transparent, but the current code clears a temporary transparent buffer and then composites it over blue, leaving the pixel blue.

```rust
#[test]
fn clear_color_is_root_compose_backdrop() {
    let mut ctx = RenderContext::new(1, 1);
    ctx.set_blend_mode(BlendMode::new(Mix::Normal, Compose::Clear));
    ctx.set_paint(RED);
    ctx.fill_rect(&Rect::new(0.0, 0.0, 1.0, 1.0));
    ctx.flush();

    let mut pixmap = solid_pixmap(1, 1, GRAY);
    ctx.render_with(
        &mut pixmap,
        &mut Resources::new(),
        RasterizerSettings {
            target_init: TargetInit::Clear(BLUE),
            ..Default::default()
        },
    );

    assert_eq!(pixmap.sample(0, 0), transparent_pixel());
}
```

LaurenzV: It depends on what behavior you expect of a background, but I think the current behavior makes more sense, for two reasons:

1) As outlined in the previous PR, if we want to have consistency with Vello Hybrid, we need to do it this way. In the WebGL backend, clearing to a specific color will be implemented using scissor rect + clear color, since this is the most effective way of doing it. By doing so, we cannot implement the semantics of a destructive blending operation affecting the background, because the clear has already been applied to the main drawing buffer and we don't sample from it anymore. Therefore, if we want consistency in Vello CPU, we need to implement it in the same way (as was done in the last PR for plain `SrcOver`).

2) I think it is _usually_ more desirable to have the background be isolated from the actual scene. hayro (my PDF renderer) [currently does the same](https://github.com/LaurenzV/hayro/blob/5a5f0e247c970df948505ee0bb36e2df2504bf86/hayro/src/lib.rs#L269-L285), and it's necessary for correctness reasons. Hayro uses a white background to emulate the fact that the PDF is composed on top of a white page of paper. Even if the PDF itself contains (destructive) blends, the real order of operation is to first render the whole scene, and then composite the whole result on top of this white page, like a real printer would do. 

There might be cases where the other behavior is more desirable, but overall I think the current behavior is the better default. For our internal use cases, I think that logic for page backgrounds is also going to make more sense.

grebmeg: Thanks for the detailed explanation — I now understand why you want these semantics, especially for backend consistency and page-background use cases. My remaining concern is the API contract: `TargetInit::Clear` says the surface is cleared before drawing, while this behaves like compositing the completed scene over a background afterward. Could we document that distinction explicitly somehwere and add this case as a test?

grebmeg (approved):
LGTM

vello_cpu/src/fine/mod.rs
LaurenzV: Makes sense, will see whether I can tweak the documentation.

Regarding tests, I have quite a few in the last PR: https://github.com/linebender/vello/pull/1869. Is there anything specific missing? The reason I put them in the last one is because otherwise it was a bit hard to write the tests themselves since they only work in Vello CPU up to this PR.

LaurenzV: Will add the following comment:

```
    /// The clear color is treated as an isolated background on top of which the fully
    /// rendered scene is composited. In particular, it will not be used as the backdrop
    /// for blending operations in the main scene.
```

If you want more tests, happy to add them to the final commit!

PR #1867, opened 2026-09-01
vello_cpu: Update target transparency hints

LaurenzV:
This PR implements an optimization that changes the `may_have_transparency` flag of a pixmap that is a render target. In particular, if the background is opaque, we know that there cannot be any transparency in the final pixmap (as long as there is no non-default blend against the root). As a consequence, we can detect that the pixmap is opaque and do not need to pay the cost of unnecessarily unpremultiplying its pixels!


grebmeg (approved):
LGTM

PR #1868, opened 2026-09-01
vello_sparse_tests: Separate rendering from snapshots

LaurenzV:
This PR introduces a small change to the test suite: It splits `render_to_pixmap` into two separate methods, one for just rendering and one for actually getting a snapshot of the current contents as a pixmap. This is useful for the next PR, where we will have tests that draw more than once.

### DIFF
--- a/vello_tests/tests/renderer.rs
+++ b/vello_tests/tests/renderer.rs
@@ -83,9 +83,8 @@ pub(crate) trait Renderer: Sized {
     fn set_filter_effect(&mut self, filter: Filter);
     fn reset_filter_effect(&mut self);
     fn reset(&mut self);
-    fn render_to_pixmap(&mut self, pixmap: &mut Pixmap);
-    fn width(&self) -> u16;
-    fn height(&self) -> u16;
+    fn render(&mut self);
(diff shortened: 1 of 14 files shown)


vello_tests/tests/renderer.rs
```diff
@@ -356,6 +357,13 @@ impl HybridRenderer {
...
             next_external_texture_id: 1,
+            gpu_test_guard: None,
+        }
+    }
+
+    fn lock_gpu_test(&mut self) {
+        if self.gpu_test_guard.is_none() {
+            self.gpu_test_guard = Some(WGPU_TEST_MUTEX.lock().unwrap());
         }
     }
```
grebmeg, line 368: Is the wgpu crash that requires this mutex still reproducible? If so, do we have an upstream issue or minimal reproducer we could use to fix it in wgpu and eventually remove this workaround?

grebmeg (approved):
LGTM

vello_tests/tests/renderer.rs
LaurenzV: I'm not sure, i think @tomcur was the one who encountered it. I haven't had any problems myself.

PR #1869, opened 2026-09-01
vello_hybrid: Support configurable target clears

LaurenzV:
This PR builds on top of the previous PRs that have introduced a way of specifying how the target should be initialized. In this PR, we implement this for Vello Hybrid. Unlike for Vello CPU, we also add an operation that allows clearing multiple rectangular regions with the provided clear color, since this is an operation that can be implemented efficiently on the GPU and is useful for dirty rect rendering. I think the code is mostly self-explanatory, so I won’t describe this in much more detail here.

However, there is one thing worth mentioning: As you can see, the `SrcOver` and `ClearSettings::Rects` options are currently only exposed by wgpu when rendering to the root target. **In the WebGL backend, we currently only support clearing the whole viewport** (when rendering to the root viewport). The reason for this is that, for the default drawing buffer, if the option `preserveDrawingBuffer` is set to false (which is the default and also highly recommended for good performance), there actually isn’t any guarantee that the existing contents are still preserved once the frame has been presented. Therefore, I think it’s misleading if we offer an option to only clear parts of the framebuffer, or not at all. Perhaps we can consider allowing this to be set to `true` in the future, but for now I think it’s the better option if we simply force a clear operation so that users cannot get this wrong accidentally.

### DIFF
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -1456,12 +1456,8 @@ impl WebGlPrograms {
         self.maybe_resize_alphas_tex(gl, resource_texture_dimension_2d, alphas.len());
         self.maybe_resize_encoded_paints_tex(gl, resource_texture_dimension_2d, paint_idxs);
         self.maybe_resize_filter_data_tex(gl, filter_context);
-        self.maybe_update_config_buffer(
-            gl,
-            resource_texture_dimension_2d,
-            render_size,
-            DrawPassTarget::Root(root_target),
-        );
+        let negate_ndc = DrawPassTarget::Root(root_target).negate_ndc();
+        self.maybe_update_config_buffer(gl, resource_texture_dimension_2d, render_size, negate_ndc);
 
         self.upload_alpha_texture(gl, alphas);
         self.upload_encoded_paints_texture(gl, encoded_paints, paint_idxs);
@@ -1739,10 +1735,8 @@ impl WebGlPrograms {
         gl: &WebGl2RenderingContext,
         resource_texture_dimension_2d: u32,
         new_render_size: &RenderSize,
-        target: DrawPassTarget,
+        negate_ndc: bool,
     ) {
-        let negate_ndc = target.negate_ndc();
-
         // TODO: Collect all attributes that influence the config buffer into a
         // single struct and compare that, such that we cannot forget to update the
         // condition in case we add new fields in the future.
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -3439,6 +3439,7 @@ fn upload_rgba32ui_rows(
     )
     .unwrap();
 }
+
 impl DrawPassTarget {
     fn negate_ndc(self) -> bool {
         // Only negate if we are rendering to the main frame buffer.
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -3080,7 +3080,6 @@ impl WebGlRendererContext<'_> {
             }
         };
 
-        self.gl.viewport(0, 0, i32::from(width), i32::from(height));
         let [r, g, b, a] = color.premultiply().components;
         self.gl.clear_color(r, g, b, a);
 
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -28,7 +28,8 @@ use crate::draw::{EXTERNAL_TEXTURE_SLOT_COUNT, ExternalTextureBindings, External
 use crate::render::common::IMAGE_PADDING;
 use crate::util::RangedSlice;
 use crate::{
-    GpuStrip, LayersConfig, RenderError, RenderSettings, RenderSize, Resources,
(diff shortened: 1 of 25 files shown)


vello_tests/tests/target.rs
```diff
@@ -0,0 +1,136 @@
...
+    );
+    ctx.fill_rect(&rect);
+    ctx.pop_layer();
+}
+
+#[vello_test]
+fn clear_area_viewport_uses_requested_color(ctx: &mut impl Renderer) {
+    prepare_clear_test(ctx);
+    ctx.set_target_init(TargetInit::Clear(ClearSettings::Viewport {
+        color: AlphaColor::from_rgba8(18, 52, 86, 120),
```
grebmeg, line 51: Just checking my understanding: is the slight difference between the requested `rgba(18, 52, 86, 120)` and the resulting `rgba(17, 51, 85, 120)` caused by rounding during premultiplication? If so, I assume this is unavoidable with an 8-bit target?

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -409,6 +411,11 @@ impl WebGlRenderer {
         resources: &mut Resources,
         render_size: &RenderSize,
         texture_bindings: &WebGlTextureBindings,
+        // Note that in WebGL, the contents of drawing buffers are not guaranteed to be preserved
+        // unless `preserveDrawingBuffer` has been set to `true` (which is highly discouraged).
+        // Therefore, for the WebGL backend we currently don't expose more configurable clear
+        // options.
+        clear_color: AlphaColor<Srgb>,
```
grebmeg, line 422: Could WebGL support persistent target contents by rendering into a caller-owned texture/framebuffer, updating only dirty regions there, and then blitting the complete texture to the canvas each frame? That would provide behavior similar to wgpu’s persistent `TextureView` without requiring the slower `preserveDrawingBuffer` option.

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -1724,9 +1734,9 @@ impl WebGlPrograms {
         gl: &WebGl2RenderingContext,
         resource_texture_dimension_2d: u32,
         new_render_size: &RenderSize,
+        target: DrawPassTarget,
     ) {
-        // Only negate if we are rendering to the main frame buffer.
-        let negate_ndc = self.resources.view_framebuffer_override.is_none();
+        let negate_ndc = target.negate_ndc();
```
grebmeg: Would it be simpler to calculate `negate_ndc` before this call and pass the boolean directly? `maybe_update_config_buffer` does not otherwise use `DrawPassTarget`, so accepting only the value it needs would make the dependency clearer.

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -3430,3 +3471,9 @@ fn upload_rgba32ui_rows(
     )
     .unwrap();
 }
+impl DrawPassTarget {
```
grebmeg, line 3442: nit: a new line

```suggestion

impl DrawPassTarget {
```

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -3098,32 +3096,69 @@ impl WebGlRendererContext<'_> {
...
+            DrawPassTarget::Layer(target) => {
+                let size = self.texture_size();
+                let framebuffer = self.programs.resources.layer_framebuffer(target);
+                self.gl
+                    .bind_framebuffer(WebGl2RenderingContext::FRAMEBUFFER, Some(framebuffer));
+                (size.width(), size.height())
+            }
+        };
+
+        self.gl.viewport(0, 0, i32::from(width), i32::from(height));
```
grebmeg: `gl.viewport` only controls where subsequent drawing happens; it does not affect `gl.clear`, which uses the framebuffer and scissor state. Since the first drawing pass immediately sets the same viewport again, could we skip this call and avoid an extra WebGL boundary crossing each frame (maybe at least for `ClearSettings::Viewport`?

grebmeg (approved):
Overall LGTM. It would be great to eventually expose multi-rect clearing when rendering to preserved offscreen framebuffers (or textures), even if the visible canvas still requires a full clear and redraw each frame.

vello_tests/tests/target.rs
LaurenzV: Lemme double-check.

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Probably could, yes. Happy to add that if the need arises!

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Makes sense!

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Ah, yeah you are right. Seems like `clear_atlas_region` has the same issue as well.

vello_tests/tests/target.rs
LaurenzV: Yeah seems to be the rounding. For the first one, we get `18 * 120 / 255 = 8.47`, in the drawing buffer it's going to be stored as 8, and when unpremultiplying you get `8 * 255 / 120 = 17`. Probably only avoidable by supporting f16/f32 render targets.

Issue #1871, opened 2026-09-02
Proposal: reorganize Vello renderer structure


grebmeg:
## Proposal: reorganize Vello renderer structure

Today, we reached consensus on the target repository structure, and this issue will track the incremental movement of its parts.

Proposed structure:

```text
README.md                     # project overview, quick start, status, and navigation
ARCHITECTURE.md               # project-wide renderer architecture guide

vello_common/                 # shared Sparse Strips data structures and utilities
vello_cpu/                    # package `vello_cpu` — CPU renderer
vello_gpu/                    # package `vello_gpu` — CPU/GPU renderer
vello_gpu_shaders/            # shaders used by the GPU renderer
vello_bench/                  # Sparse Strips benchmarks

vello_tests/                  # shared CPU and GPU integration/snapshot tests
├── vello_dev_macros/         # test-only procedural macros
├── vello_example_scenes/     # scenes shared by tests and examples
└── web/                      # browser test and size-check tooling

vello_toy/                    # temporary placement; final role/name remains open

research/                     # compute-centric research implementation
├── README.md                 # research status, development guide, and package map
├── CHANGELOG.md              # former `vello` 0.10.x release history
├── vello_research/           # package `vello_research`
├── vello_encoding/           # compute encoding support crate
├── vello_shaders/            # compute shaders and CPU reference kernels
├── vello_research_tests/     # research integration and snapshot tests
├── xtask/                    # research snapshot/comparison tooling
├── examples/
│   ├── headless/
│   ├── scenes/
│   ├── simple/
│   ├── simple_sdl2/
│   └── with_winit/
└── doc/                      # remaining research-specific design and history docs

assets/                       # resources shared across renderer families
tools/
└── run_wasm/                 # renderer-independent WASM runner
```

Agreed direction:

- Use a shallow repository structure without `renderers/` or `sparse/`.
- Keep `vello_cpu` at the repository root.
- Rename `vello_hybrid` to `vello_gpu` and move it to the repository root.
- Move the current compute-centric renderer to `research/vello_research`.
- Rename `vello_sparse_shaders` to `vello_gpu_shaders`.
- Rename `vello_sparse_tests` to the root-level `vello_tests`.
- Move the existing compute-renderer tests to `research/vello_research_tests`.
- Place `vello_dev_macros`, `vello_example_scenes`, and web tooling under `vello_tests/`.
- Move `ARCHITECTURE.md` to the repository root.
- Add a future `vello` abstraction only when a concrete API exists.
[...]

PR #1872, opened 2026-09-02
Move sparse renderers under renderers

grebmeg:
#### Summary
Reorganizes the repository to make vello sparse strips renderers are our default ones:

- Moves Sparse Strips renderers and support crates to the root.
- Groups the compute-centric implementation under `research/`.
- Moves shared assets and tooling to stable root locations.
- Repairs workspace, CI, LFS, runtime, and documentation paths.

Package names and public APIs remain unchanged. The first commit contains only `git mv` operations; the second contains repairs.

Part of #1871.


LaurenzV:
Gonna wait with the review until we have full consensus (see [Olivier's](https://github.com/linebender/vello/issues/1871#issuecomment-5523716693) comment).

LaurenzV (approved):
Overall structure looks good to me, with the understanding that README's are still pending to be updated and the folder renames have intentionally not been done yet.

I have not tried to verify myself that all the stale links are updated now, but it seems like Codex found some more stuff:

Image

PR #1873, opened 2026-09-03
fix: avoid enabling both Vello Hybrid web renderers

grebmeg:
Prevent native WebGL builds from also enabling wgpu by explicitly selecting Vello Hybrid features per consumer.

### DIFF
--- a/sparse_strips/vello_sparse_tests/Cargo.toml
+++ b/sparse_strips/vello_sparse_tests/Cargo.toml
@@ -22,7 +22,7 @@ path = "src/regenerate_probe_reference.rs"
 glifo = { workspace = true }
 vello_common = { workspace = true, features = ["std", "probe"] }
 vello_cpu = { workspace = true, features = ["multithreading", "std", "f32_pipeline"] }
-vello_hybrid = { workspace = true }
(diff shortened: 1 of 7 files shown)


sparse_strips/vello_sparse_tests/Cargo.toml
```diff
@@ -22,7 +22,7 @@ path = "src/regenerate_probe_reference.rs"
 glifo = { workspace = true }
 vello_common = { workspace = true, features = ["std", "probe"] }
 vello_cpu = { workspace = true, features = ["multithreading", "std", "f32_pipeline"] }
-vello_hybrid = { workspace = true }
+vello_hybrid = { workspace = true, features = ["text", "wgpu", "wgpu_default"] }
```
LaurenzV, line 25: Why do we not need to enable webgl here? 🤔

LaurenzV (approved):

sparse_strips/vello_sparse_tests/Cargo.toml
grebmeg: It’s enabled conditionally via the `vello_sparse_tests webgl` feature below, which forwards to `vello_hybrid/webgl`. CI passes `--features webgl`. Adding it here would enable both renderers unconditionally.

PR #1874, opened 2026-09-03
vello_hybrid: Avoid preserving draw state in WebGL state guard

LaurenzV:
After looking into why the initial synchronous part of the probe runs so slowly on some devices, I noticed that a specific `getParameter` call is the issue, forcing a flush and therefore blocking the main thread for a very long time (more than 1 second):

image

Because of these footguns, we want to get rid of all `getParameter` calls in the long term, but in order to keep the code simple this PR only removes part of them for now. The contract is now that each method that performs draw calls must set all of these parameters explicitly themself, and we don't guarantee to restore the original values after vello is done. This currently does has the consequence that we often emit state-change calls unnecessarily (by changing the state to a value that was already set), but optimizing this is for the future.

With this change in place, the synchronous probe part only takes a few milliseconds now on that device!

Take this analysis with a grain of salt, but it seems like `ACTIVE_TEXTURE` is the main hazard remaining right now:
image

### DIFF
--- a/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
+++ b/sparse_strips/vello_hybrid/src/render/webgl/mod.rs
@@ -1863,6 +1863,7 @@ impl WebGlPrograms {
             WebGl2RenderingContext::FRAMEBUFFER,
             self.resources.view_framebuffer_override.as_deref(),
         );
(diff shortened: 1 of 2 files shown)


grebmeg (approved):
LGTM! Definitely in favour of this. I did some similar prototyping and agree with minimizing anything that requires synchronous stalls/flushing.

PR #1878, opened 2026-09-07
vello_hybrid: Do not apply a magic offset for image sampling

LaurenzV:
I think we should avoid complicating the shaders by adding arbitrary offsets just to make specific tests work in CI. Vello CPU itself does not apply any such offset, so it produces an inconsistency between Vello CPU and Vello Hybrid. In addition to that, this offset doesn't actually fix anything, it just moves the problem to somewhere else: While this does remove the artifacts for the specific case of axis-aligned images that are skewed by 45%, if we were to shift the image slightly we would get the same artifacts.

The fact that NN images can cause artifacts is (in my view) inevitable, and if you want to avoid that you should just use bilinear image sampling instead.

I think I would like to do the same for gradients, but in a different PR.

(Open for options on how to handle the test images instead! Maybe I should convert them to use bilinear sampling instead?)

### DIFF
--- a/sparse_strips/vello_sparse_tests/tests/image.rs
+++ b/sparse_strips/vello_sparse_tests/tests/image.rs
@@ -202,7 +202,7 @@ fn image_with_transform_scaling_non_uniform(ctx: &mut impl Renderer) {
     );
 }
 
-#[vello_test]
+#[vello_test(diff_pixels = 64)]
 fn image_with_transform_skew_x_1(ctx: &mut impl Renderer) {
     transform(
         ctx,
@@ -214,7 +214,7 @@ fn image_with_transform_skew_x_1(ctx: &mut impl Renderer) {
     );
 }
 
-#[vello_test]
+#[vello_test(diff_pixels = 64)]
 fn image_with_transform_skew_x_2(ctx: &mut impl Renderer) {
     transform(
         ctx,
@@ -226,7 +226,7 @@ fn image_with_transform_skew_x_2(ctx: &mut impl Renderer) {
     );
 }
 
-#[vello_test]
+#[vello_test(diff_pixels = 64)]
 fn image_with_transform_skew_y_1(ctx: &mut impl Renderer) {
     transform(
         ctx,
@@ -238,7 +238,7 @@ fn image_with_transform_skew_y_1(ctx: &mut impl Renderer) {
     );
 }
 
-#[vello_test]
+#[vello_test(diff_pixels = 64)]
 fn image_with_transform_skew_y_2(ctx: &mut impl Renderer) {
     transform(
         ctx,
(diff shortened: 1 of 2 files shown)


sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -202,7 +202,7 @@ fn image_with_transform_scaling_non_uniform(ctx: &mut impl Renderer) {
     );
 }
 
-#[vello_test]
+#[vello_test(diff_pixels = 64)]
```
grebmeg, line 205: > (Open for options on how to handle the test images instead! Maybe I should convert them to use bilinear sampling instead?)

I’d prefer more coverage here rather than converting the existing tests. It would be great to keep the NN skew tests and add equivalent bilinear tests as well.

sparse_strips/vello_sparse_tests/tests/image.rs
```diff
@@ -214,7 +214,7 @@ fn image_with_transform_skew_x_1(ctx: &mut impl Renderer) {
...
```
grebmeg, line 217: Before merging this, could we extend the test framework so comparison parameters, including `diff_pixels`, tolerances, and similar options, can be configured per backend? The expected differences here are Hybrid-specific, so unaffected backends should retain stricter checks without requiring one-off parameters for every setting.

LaurenzV: Hmm sure!

sparse_strips/vello_sparse_tests/tests/image.rs
LaurenzV: Hmm, what do you think about instead just changing the existing skew/rotate tests to not use a precise 45 degree angle? I think there is just little point in having a test just for aliasing artifacts. Back then, the original reason I added the test (I think it was me!) was just to ensure that skewed/rotated paints themselves work correctly.

grebmeg: I think we can have both? I think the 45-degree angle is a very interesting edge case that's still worth having.

sparse_strips/vello_sparse_tests/tests/image.rs
laurenz-canva: After thinking about this, I'm not so sure anymore. I can still make `diff_pixels` configurable per backend, but CPUs might actually suffer from the same floating point issues as GPUs, we just haven't observed this on our devices. So I think it might be better to have the tolerance for both?

grebmeg: Yes, but it’d be great to identify those devices and understand which ones are affected, rather than hiding the differences behind a single tolerance for both. I’d much prefer to find the problematic cases first and apply a tolerance afterward, rather than doing it preemptively.

laurenz-canva: Ok!

LaurenzV: Changed it so it's only applied on GPU!

grebmeg (approved):

PR #1879, opened 2026-09-07
vello_hybrid: Don't store active texture and texture bindings in state guard

LaurenzV:
Follow up to https://github.com/linebender/vello/pull/1874. As was shown there, it seems like in Safari and Firefox `ACTIVE_TEXTURE` queries can still be expensive and aren't cached on the CPU side.

We were already not being fully consistent about this before, anyway (for example, `write_to_atlas` would update the texture 0 binding without restoring it later on). Since we want to work towards getting rid of all `getParameter` queries, I think the best thing to do here is to simply not make any promises about what textures are active and which bindings are preserved when running Vello Hybrid.

### DIFF
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -1926,8 +1926,6 @@ pub(crate) struct WebGlStateGuard {
     config: WebGlStateConfig,
     original_framebuffer: Option<WebGlFramebuffer>,
     original_read_framebuffer: Option<WebGlFramebuffer>,
-    original_active_texture: Option<u32>,
-    original_texture_2d: Option<WebGlTexture>,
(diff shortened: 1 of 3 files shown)


vello_gpu/src/render/webgl/mod.rs
grebmeg, line 1: Could we document the new contract somewhere? In particular, renderer operations may change the active texture unit and texture bindings, so callers sharing the context need to re-establish any state they depend on afterward.

grebmeg (approved):
LGTM

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Added this to the `WebGlRenderer` struct, as I don't think we should pollute the documentation of each method with this information. Hope that's good enough!

PR #1880, opened 2026-09-07
vello_common: Support non-opaque image samplers

LaurenzV:
Adapted from #1736, unfortunately I didn't have push access to that branch, so I had to create a new one to push some cleanups.

### DIFF
--- a/vello_common/src/encode.rs
+++ b/vello_common/src/encode.rs
@@ -8,7 +8,7 @@ use crate::color::palette::css::BLACK;
 use crate::color::{ColorSpaceTag, HueDirection, Srgb, gradient};
 use crate::kurbo::{Affine, Point, Vec2};
 use crate::math::{FloatExt, compute_erf7};
-use crate::paint::{Image, ImageSource, IndexedPaint, Paint, PremulColor, Tint};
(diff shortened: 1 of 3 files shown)


vello_common/src/encode.rs
```diff
@@ -483,15 +483,19 @@ impl EncodeExt for Image {
...
         let mut sampler = self.sampler;
 
         if sampler.alpha != 1.0 {
-            // If the sampler alpha is not 1.0, we need to force alpha compositing.
-            unimplemented!("Applying opacity to image commands");
+            let tint = tint.get_or_insert(Tint {
+                color: peniko::Color::WHITE,
+                mode: TintMode::Multiply,
+            });
+            tint.color = tint.color.multiply_alpha(sampler.alpha);
```
grebmeg: `ImageSampler::with_alpha` accepts values above one, but folding 2.0 into a white tint leaves its alpha at 2.0. Should we clamp the combined alpha?

grebmeg (approved):
LGTM

vello_common/src/encode.rs
LaurenzV: Makes sense!

PR #1883, opened 2026-09-08
refactor: rename renderer package families

grebmeg:
Rename renderer directories and package identities to clarify their roles:

- `vello_hybrid` → `vello_gpu`
- `vello_sparse_shaders` → `vello_gpu_shaders`
- `vello_sparse_tests` → `vello_tests`
- Research directories gain explicit `vello_research` naming while preserving the published Vello Classic package names.

Also repairs Cargo, source, CI, scripts, LFS, and documentation references.

Part of #1871.


LaurenzV (approved):
So just to double check, the package name of Vello is still intentionall `vello` instead of `vello_research`, right?

Otherwise LGTM!

grebmeg:
> So just to double check, the package name of Vello is still intentionall vello instead of vello_research, right?

Thanks for double-checking! Yes, the `vello` repo is intentionally preserved. I know it’s a bit confusing, but all the options have their pros and cons. After this PR, I’ll work on hopefully the last one, where I’ll update all the documentation and clarify this as well.

PR #1885, opened 2026-09-08
vello_common: Add depth buffer test case to the probe

LaurenzV:
The Vello Hybrid probe has so far proven to be very effective at detecting various GPU defects. One thing that I think has been missing so far is an explicit test that checks the correctness of the depth buffer (to ensure that, even though all opaque shapes are drawn before the translucent ones, they still end up on top). This PR fills this gap by adding a new probe case. The important part here is that the rectangles are not anti-aliased, otherwise they will not hit the opaque path since there is a minimum threshold for such rectangles.


vello_common/src/probe.rs
```diff
@@ -89,6 +90,8 @@ pub enum ProbeFeature {
     Blending = 7,
     /// Drawing with a non-identity transform.
     Transformed = 8,
+    /// Layering opaque draws and a transparent foreground to exercise depth buffering.
+    DepthBuffer = 9,
```
grebmeg, line 94: Does this actually exercise depth buffering in WebGL? `probe_inner` renders the scene as `RootTarget::AtlasLayer`, but `enable_depth()` only returns `true` for `UserSurface`. This means opaque rectangles fall back to the regular strip list and `strip_pass_inner` disables `DEPTH_TEST`, so the expected painter-order output can be produced without using the depth buffer.

LaurenzV: 🤦 Thanks, yeah let me think how to fix this.

LaurenzV:
Will open a new PR for this.

PR #1888, opened 2026-09-08
vello_tests: Unignore probe test

LaurenzV:
It seems like whatever caused the issue before, as part of our refactors this test doesn't seem to be flaky anymore. Tried this in my fork and it passed twice in a row.


grebmeg (approved):
Nice!

PR #1889, opened 2026-09-08
vello_hybrid: Make render backend callbacks fallible

LaurenzV:
This PR makes the various render methods fallible. This is necessary because in theory, the WebGL backend can fail if the buffer for instances exceeds i32. Admittedly that's very much an edge case, but I don't think there is any real downside to making those fallible.

### DIFF
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -648,7 +648,8 @@ impl WebGlRenderer {
             &mut self.schedule_storage,
             schedule,
             root_output_target,
-        );
+        )
+        .unwrap_or_else(|error| match error {});
 
         self.gl.bind_framebuffer(
             WebGl2RenderingContext::FRAMEBUFFER,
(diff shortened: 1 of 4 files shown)


vello_gpu/src/render/webgl/mod.rs
```diff
@@ -648,7 +648,8 @@ impl WebGlRenderer {
             &mut self.schedule_storage,
             schedule,
             root_output_target,
-        );
+        )
+        .unwrap_or_else(|error| match error {});
```
HigherOrderLogic, line 652: I think this looks cleaner.

```suggestion
	let _ = crate::schedule::execute(
		&mut ctx,
		&mut self.schedule_storage,
		schedule,
		root_output_target,
	)
```

grebmeg (approved):

vello_gpu/src/render/webgl/mod.rs
LaurenzV: I agree it's ugly, but it makes it more explicit that the error is infallible. It will also go away anyway in the follow-up PR that makes rendering fallible.

PR #1892, opened 2026-09-08
vello_gpu: Make WebGL operations fallible

LaurenzV:
Right now, we are very inconsistent in how we handle errors. In some cases we do have dedicated error types, but in other cases we just unwrap or have random assertions in the code, making it hard for the user to properly catch those. This applies to both, the wgpu and WebGL backend, but this PR only focuses on the WebGL backend for now. This PR *improves* the situation by introducing more error types and using them. I'm not sure whether this current state is already what I'd consider ideal, but I think it's certainly better than before!

While it looks like a simple change, this actually turned out a bit harder than I had initially anticipated because it's hard to draw a boundary between what should be an error or not. On the one hand, ideally the user should be able to catch any panics without relying on panic unwind, but on the other hand we also don't want to expose different error kinds for every single thing. There also was some discussion about reducing the number of strings in the program, but I decided to leave this to a follow-up since this PR is already large enough. However, I've designed the error hierarchy in such a way that it should support this more easily.

There are two things worth pointing out as tricky:

- The first problem is context loss. Unfortunately, a WebGL context loss can basically happen at any time between any two calls. The unfortunate thing is that there is no consistent way how they surface: Certain WebGL calls (like draw calls) don't report anything at all and just fail silently, while other methods will instead return `null`. While we could simply check whether the WebGL context has been lost at the start and end of each method, I feel like this would be way too invasive. Because of this, I've decided to not add direct recognition for this. In case a method returns null, we do explicitly return a failure via the `OperationFailed` error, but it's not guaranteed to be hit in every frame. Therefore, I've decided to instead document that clients should add an event listener to the WebGL context lost event and handle it accordingly.
- The second, as mentioned, is the error boundary. I've decided to mostly follow the approach of adding dedicated error types for situations that can cause real failure (for example, the number of depth bits not being large enough), but keep unwraps/expect in any situation that either should never occur (for example MAX_TEXTURE_SIZE not returning a number) or is really an internal logic error (for example a missing layer texture). The number of different errors is already a bit too large for my liking, so I hope this provides a good balance.

I hope this represents a good improvement.


LaurenzV:
(Also, if you consider the errors to be detailed, definitely open to revisiting the exact granularity!)

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -2275,7 +2385,7 @@ pub(crate) fn create_texture_storage(
         width as i32,
         height as i32,
```
grebmeg, line 2476: `texStorage2D` reports allocation and dimension failures through the WebGL error flag rather than this JS Result, so this function can return `Ok` for an unusable texture. Could we restore the removed resource-dimension guards as a typed error? Using `getError()` could detect allocation failures here, but I don’t think it is practical because it can introduce an expensive synchronous round trip. I’d prefer explicit validation for predictable limits and document that failures such as OOM remain best-effort.

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -754,8 +792,10 @@ impl WebGlRenderer {
                 ],
                 image_resource.width + padding * 2,
                 image_resource.height + padding * 2,
-            );
+            )?;
         }
+
+        Ok(())
```
grebmeg, line 809: Could we clear the atlas region before committing the cache deallocation? If framebuffer creation fails here, the image ID is already gone, preventing a retry and potentially leaving stale padding in a subsequently reused region.

vello_gpu/src/render/webgl/mod.rs
grebmeg: I don’t yet have a strong understanding of where we should add `getError()` checks around image uploads and other void WebGL operations. My understanding is that errors such as `OUT_OF_MEMORY` remain pending until `getError()` is called, and each call clears one error; unless we check immediately after the relevant operation with no earlier errors pending, we cannot reliably identify which operation failed. Should we check every high-risk operation, or define broader checkpoints to avoid the performance cost? (not in this PR if we even do)

vello_gpu/src/render/webgl/probe.rs
```diff
@@ -258,8 +258,11 @@ fn launch_probe(
     framebuffer: &Framebuffer,
     width: u16,
     height: u16,
-) -> WebGlPendingProbe {
-    let pixel_pack_buffer = gl.create_buffer().unwrap();
+) -> Result<WebGlPendingProbe, WebGlError> {
+    let pixel_pack_buffer = gl.create_buffer().ok_or(WebGlError::OperationFailed {
+        operation: WebGlOperation::Probe(WebGlProbeOperation::BufferCreation),
+        message: None,
+    })?;
```
grebmeg, line 295: Could we guard `pixel_pack_buffer` immediately after creation? Both `readPixels` and `fenceSync` can return before ownership reaches `WebGlPendingProbe`, so `deleteBuffer` is never called. For example:

```rust
let mut pending = WebGlPendingProbe {
    gl: gl.clone(),
    sync: None,
    buffer: Some(pixel_pack_buffer),
    width,
    height,
};

// Perform the fallible readback setup.

pending.sync = Some(
    gl.fence_sync(WebGl2RenderingContext::SYNC_GPU_COMMANDS_COMPLETE, 0)
        .ok_or(WebGlError::OperationFailed {
            operation: WebGlOperation::Probe(
                WebGlProbeOperation::Synchronization,
            ),
            message: None,
        })?,
);

Ok(pending)
```

This makes the existing `Drop` implementation clean up the buffer on every early return.

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -2028,24 +2121,42 @@ pub(crate) struct WebGlStateConfig {
...
+fn required_uniform_location(
+    gl: &WebGl2RenderingContext,
+    program: &Program,
+    name: &'static str,
+) -> Result<WebGlUniformLocation, WebGlError> {
+    gl.get_uniform_location(program, name)
+        .ok_or(WebGlError::OperationFailed {
+            operation: WebGlOperation::ShaderInterface(WebGlShaderInterfaceOperation::Uniform),
+            message: Some(name.into()),
+        })
```
grebmeg, line 2223: Could this use `ok_or_else`? `ok_or` eagerly allocates the uniform-name `String` even when the lookup succeeds, causing 16 unnecessary allocations during renderer initialization.

```rust
gl.get_uniform_location(program, name)
    .ok_or_else(|| WebGlError::OperationFailed {
        operation: WebGlOperation::ShaderInterface(
            WebGlShaderInterfaceOperation::Uniform,
        ),
        message: Some(name.into()),
    })
```

grebmeg (approved):
LGTM if some of the comments are addressed.

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Done.

vello_gpu/src/render/webgl/probe.rs
LaurenzV: Moved the construction of the pending probe up to the very top. I think the long-term solution would be to simply create RAII wrappers for those two as well.

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Done!

vello_gpu/src/render/webgl/mod.rs
LaurenzV: Thanks, yes, I will restore the resource-dimension guard!

Also unsure how to handle `getError`. Since image uploads happen pretty frequently, it would be a bit to make a `getError` call every time. Might be worth to experiment with this at some point.

PR #1894, opened 2026-09-08
vello_bench: Add CPU allocation regression benchmarks

grebmeg:
Add deterministic allocation regression benchmarks. Added for warmed-up Tiger scene building, rasterization, and full-frame rendering. The benchmarks support configurable frame counts, enforce allocation limits, report regressions, and run in CI.

### DIFF
--- a/vello_bench/src/allocations/counter.rs
+++ b/vello_bench/src/allocations/counter.rs
@@ -1,0 +1,150 @@
+// Copyright 2026 the Vello Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use std::alloc::{GlobalAlloc, Layout, System};
+use std::cmp::Ordering;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering::Relaxed};
+
+static ACTIVE: AtomicBool = AtomicBool::new(false);
+static ALLOCATIONS: AtomicUsize = AtomicUsize::new(0);
+static REALLOCATIONS: AtomicUsize = AtomicUsize::new(0);
+static ALLOCATED_BYTES: AtomicUsize = AtomicUsize::new(0);
+static LIVE_BYTES: AtomicUsize = AtomicUsize::new(0);
+static PEAK_LIVE_BYTES: AtomicUsize = AtomicUsize::new(0);
+
+/// A system allocator wrapper that records allocations inside [`measure`].
+#[derive(Debug)]
+pub struct CountingAllocator;
+
+// SAFETY: Every operation delegates to `System` with the original arguments. The additional
+// bookkeeping does not inspect or modify allocated memory.
+unsafe impl GlobalAlloc for CountingAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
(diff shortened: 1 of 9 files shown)


LaurenzV:
I agree it would be good to have some tests for memory usage, but I'm not sure if vello_bench is the right place, especially since I'm rewriting it currently so we can run benchmarks in WASM, etc... But my gut feeling is that this should probably live in `vello_sparse_tests` instead (this would also save us a CI job), but the downside is we currently don't have a way of rendering the ghostscript tiger there easily.

What do you think about revisiting this when I'm done with that work? Or do you think it's urgent?

grebmeg:
> What do you think about revisiting this when I'm done with that work? Or do you think it's urgent?

I think it’s better to have this than not, and we can always move it somewhere else later. I don’t have a strong preference for where it should live, but I agree that it belongs in the tests.

vello_bench/src/allocations/counter.rs
```diff
@@ -0,0 +1,150 @@
...
+        if !ptr.is_null() {
+            record_allocation(layout.size());
+        }
+        ptr
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+        LIVE_BYTES.fetch_sub(layout.size(), Relaxed);
+        // SAFETY: The caller upholds `GlobalAlloc::dealloc`'s contract.
+        unsafe { System.dealloc(ptr, layout) };
```
LaurenzV, line 43: Just wondering if there's a reason why here the system call happens after the recording while in the other cases before?

vello_bench/src/allocations/counter.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
+        // SAFETY: The caller upholds `GlobalAlloc::realloc`'s contract.
+        let new_ptr = unsafe { System.realloc(ptr, layout, new_size) };
+        if !new_ptr.is_null() {
+            let old_size = layout.size();
+            if new_size >= old_size {
+                let added = new_size - old_size;
+                let live = LIVE_BYTES.fetch_add(added, Relaxed) + added;
+                record_peak(live);
```
LaurenzV, line 54: Just for my understanding, in theory, this probably isn't fully accurate, because in reality the system needs to create a new allocation for the full size, copy the contents over, and only then free the old allocation, right?

Not relevant for us since we only need an estimate, but just wondering. 😄

LaurenzV (approved):
If it's only a "temporary" measure until we have better infrastructure, LGTM!

vello_bench/src/allocations/counter.rs
grebmeg: No particular reason — good catch. I moved the bookkeeping after the system deallocation.

vello_bench/src/allocations/counter.rs
grebmeg: Yes, if `realloc` has to move the allocation, the allocator may transiently hold both blocks while copying. It can also grow in place, and this wrapper can’t observe allocator-internal behavior, so the metric intentionally tracks live requested bytes before and after the call rather than true allocator peak memory.

PR #1896, opened 2026-09-09
vello_gpu: upload only used rows of data texture

HigherOrderLogic:
Only upload the used rows of strip data texture.

This is similar to how it's done in the WebGl renderer.


vello_gpu/src/render/wgpu/mod.rs
```diff
@@ -2415,28 +2419,44 @@ impl Programs {
...
+            .sum::<usize>()
+            / 16;
+        let rows = used_texels
+            .div_ceil(encoded_paints_texture_width as usize)
+            .min(encoded_paints_texture_height as usize);
+
+        GpuEncodedPaint::serialize_to_buffer(
+            encoded_paints,
+            &mut self.encoded_paints_data[..rows * encoded_paints_texture_width as usize * 16],
+        );
```
grebmeg: Could we pass `paint_idxs` here and use `*paint_idxs.last().unwrap()`, as the WebGL implementation does? [The final entry is already assigned `current_idx`](https://github.com/linebender/vello/blob/a6fe010150b20bb33c454e2e6c857dfcc9a54581/vello_gpu/src/render/wgpu/mod.rs#L745), so this would keep the upload size consistent with texture resizing and shader addressing while avoiding a second pass over the paints.

grebmeg (commented):
Thanks for putting this PR together!

grebmeg (commented):
Amazing! Thank you!

grebmeg (approved):

PR #1897, opened 2026-09-09
vello_gpu: store strips data in arena

HigherOrderLogic:
Clear the TODO.


vello_gpu/src/render/wgpu/mod.rs
```diff
@@ -960,11 +974,52 @@ struct Programs {
     filter_data: Vec<u8>,
 }
 
+#[derive(Debug)]
+struct StripsArena {
```
grebmeg: Maybe `StripBufferArena`?

vello_gpu/src/render/wgpu/mod.rs
```diff
@@ -2510,37 +2556,40 @@ impl Programs {
...
+        if total_len == 0 {
+            return 0..0;
+        }
+
+        let arena = &mut self.strips_arena;
+        let offset = arena.alloc(total_len);
+        let size = NonZeroU64::new(total_len).expect("total length is non-zero");
+        let mut view = queue
+            .write_buffer_with(&arena.buffer, offset, size)
             .expect("Capacity handled in creation");
```
grebmeg, line 2571: This removes per-pass destination-buffer creation, but `write_buffer_with` still allocates temporary staging storage on native wgpu for every pass. Could we retain the staging-allocation `TODO`?

vello_gpu/src/render/wgpu/mod.rs
grebmeg: Would you be able to look into applying the same frame-arena approach to the WebGL backend as a follow-up? WebGL currently reallocates the strip buffer for each pass, so it may benefit from the same reuse strategy.

grebmeg (approved):
LGTM if LGT @LaurenzV

vello_gpu/src/render/wgpu/mod.rs
HigherOrderLogic: > Would you be able to look into applying the same frame-arena approach to the WebGL backend as a follow-up?

I'll do it at a later date. I should have one more arena optimization for the Wgpu backend first, then I'll consider if these can be applied to the WebGL one.

LaurenzV:
Thanks! Looking at the profile I think removing the temporary CPU-side staging buffers from wgpu will give us the biggest wins, but this is a good improvement nevertheless!

image

vello_gpu/src/render/wgpu/mod.rs
grebmeg: Thank you!

PR #1899, opened 2026-09-09
vello_gpu: Refactor WebGL root framebuffer selection

LaurenzV:
This PR

- Replaces our previous `framebuffer_override` option with a more elaborate `ViewFramebuffer` enum that also stores information about depth buffers.
- Adds support for using depth buffers in a non-default framebuffer, but doesn't actually use it anywhere yet.


LaurenzV:
Need to revisit this again, don't review yet please!

grebmeg (approved):

PR #1900, opened 2026-09-09
vello_gpu: Add optional support for a depth buffer to the WebGL probe

LaurenzV:
This PR enables the depth buffer in the Vello Hybrid probe, depending on whether the user enabled the depth buffer originally.

I've decided to not add the new render buffer to the WebGL state guard since we are trying to move away from this, anyway. Let me know if I should change this.

### DIFF
--- a/vello_gpu/src/render/webgl/probe.rs
+++ b/vello_gpu/src/render/webgl/probe.rs
@@ -1,7 +1,7 @@
 // Copyright 2026 the Vello Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use crate::render::webgl::resource::Framebuffer;
+use crate::render::webgl::resource::{Framebuffer, Renderbuffer};
 use crate::render::webgl::{
     ViewFramebuffer, WebGlStateConfig, WebGlStateGuard, WebGlTextureBindings,
     create_framebuffer_for_texture, create_texture_storage,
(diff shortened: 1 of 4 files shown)


vello_gpu/src/render/webgl/resource.rs
```diff
@@ -110,6 +112,19 @@ impl GlResource for WebGlFramebuffer {
     }
 }
 
+#[cfg(feature = "probe")]
```
LaurenzV, line 115: Bit annoying to feature gate, but necessary for clippy.

LaurenzV:
Sorry, not ready for review yet.

vello_gpu/src/render/webgl/probe.rs
```diff
@@ -72,9 +72,6 @@ impl WebGlRenderer {
     }
 
     fn probe_inner(&mut self) -> Result<WebGlPendingProbe, RenderError> {
-        // IMPORTANT NOTE: When making any changes to the probe, make sure to
-        // unignore and rerun the "webgl_probe_succeeds" test locally.
-
```
LaurenzV, line 78: This is an outdated comment from when we ignored the probe in CI, which was fixed in #1888.

grebmeg (approved):

LaurenzV:
Changed it now so the pixels are flipped after readin the pixels of the probe.

PR #1901, opened 2026-09-09
vello_common: Add depth buffer test case

LaurenzV:
Another attempt at #1885. This time, I've double-checked that the depth buffer is actually activated at runtime. 😄

Apart from that:
- I've also run this whole stack of three PRs on some mobile devices and worked fine. Except for an issue on older MacOS devices where using a depth buffer makes the probe take longer?
- It would be nice to have a unit test that checks that the depth buffer test case actually exeercises it, but it's a bit hard to add since it's all abstracted away behind the scheduler.


vello_common/src/probe.rs
```diff
@@ -40,6 +40,7 @@ pub const PROBE_ELEMENTS: [ProbeFeature; 8] = [
     ProbeFeature::OpacityLayer,
     ProbeFeature::Blending,
     ProbeFeature::Transformed,
+    ProbeFeature::DepthBuffer,
```
grebmeg, line 43: Could we make the set of probe features configurable, so callers can enable them independently? I’d mostly use this to enable `DepthBuffer` selectively, since I’m concerned depth buffer may be broken on some older GPUs.

LaurenzV: If it's okay with you, I would defer this to the follow-up PR where we make the whole probe configurable? Since we wanted this anyway for the filter probes. And isn't the point exactly to find those GPUs where the depth buffer is broken? 😄 Or am I missing something?

grebmeg: Yes, that is the point. My concern is that enabling the depth buffer by default may itself cause issues on older devices or drivers, and we don’t yet know how broadly stable it is.

vello_common/src/probe.rs
```diff
@@ -471,6 +476,38 @@ fn draw_opacity_layer_rect(ctx: &mut impl ProbeRenderer, rect: Rect) {
     ctx.pop_layer();
 }
 
+fn draw_depth_buffer_rects(ctx: &mut impl ProbeRenderer, cell: Rect) {
```
grebmeg, line 479: As I understand it, `draw_depth_buffer_rects` produces the same expected output whether depth buffering is enabled or disabled. Although the front-to-back reordering should expose a broken depth test, should we validate more explicitly that the depth buffer path was actually used?

grebmeg (approved):

PR #1903, opened 2026-09-10
vello_gpu: Decouple WebGL NDC negation from root target

LaurenzV:
This PR makes `root_negate_ndc` an explicit argument to the WebGL `render_scene` implementation instead of deriving it from `RootTarget`. 

Previously, we assumed that rendering to the root target = we _always_ want to undo the NAGA flip. However, in the next PRs we will change the probe so that it's also treated as a normal user surface, so that we can attach a depth buffer to it. However, for the probe, we do not want to undo the naga y-flip. Therefore, we need to decouple these two properties.

### DIFF
--- a/vello_gpu/src/render/webgl/mod.rs
+++ b/vello_gpu/src/render/webgl/mod.rs
@@ -460,6 +460,7 @@ impl WebGlRenderer {
             render_size,
             TargetInit::Clear(ClearSettings::Viewport { color: clear_color }),
             RootTarget::UserSurface,
+            true,
             texture_bindings,
             None,
         )?;
@@ -549,6 +550,7 @@ impl WebGlRenderer {
             &atlas_render_size,
             TargetInit::SrcOver,
             RootTarget::AtlasLayer,
+            false,
             texture_bindings,
             Some(&render_target_texture),
         );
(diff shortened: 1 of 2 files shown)


vello_gpu/src/render/webgl/mod.rs
```diff
@@ -588,6 +590,7 @@ impl WebGlRenderer {
         render_size: &RenderSize,
         target_init: TargetInit<'_>,
         root_output_target: RootTarget,
+        root_negate_ndc: bool,
```
grebmeg, line 593: Could we avoid passing this boolean through `render_scene`? Since the probe is only 51×51 and already copies the result into CPU memory, we could render it like a normal `UserSurface` and flip the rows before comparing them. This would also ensure the probe tests the same orientation used for canvas rendering.

If other render targets need to control orientation independently, could that setting live on `ViewFramebuffer` instead? That would keep the framebuffer and its coordinate convention together.

LaurenzV: Ok sure, that's another option!

LaurenzV:
Since we are gonna apply the flip in the probe instead, this PR becomes irrelevant, so will close it.

PR #1907, opened 2026-09-14
vello_sparse_tests: Increase timeout

LaurenzV:
I don't know why, but it seems like since I added the second probe test CI is getting a bit flaky again:

https://github.com/linebender/vello/actions/runs/34816344888/job/103887749239

Hopefully this fixes it, but not sure. Otherwise, we might have to reduce them to just one test again.


LaurenzV:
Seems to have worked fine for the last three runs at least.

nicoburns (approved):

nicoburns:
Did we ever land running tests in release mode?

LaurenzV:
Yes, it's already running in release mode. But that's not what's making CI so slow, it's the GPU tests (which take 1 second each because we don't reuse the GPU context) + cache thrashing (just take a look at https://github.com/linebender/vello/actions/caches 😅)

PR #1908, opened 2026-09-14
sparse_strips: Make probe more configurable

LaurenzV:
This PR contains two changes, split into two commits:
- It creates a single global cell size and ensures that all probe elements fit nicely into it. It also makes the probe grid itself more compact, allowing us to reduce the size of the elements.
- It allows the user to customize the probe elements that are used at runtime. How this works is that we create a single image containing all probe elements and embed it into the binary (as is already the case on main), and then at runtime we generate a new reference probe image depending on the constellation of probe elements chosen, and compare the actual result against that instead. Not super pretty, but not sure there are many other ways of achieving what we want.


vello_common/assets/probe_3_elements.png
LaurenzV, line 1: We technically don't need to check those into git, but I think it's nice to also visually see how a dynamically generate reference image looks like, just as a sanity check.

vello_common/assets/probe.png
LaurenzV, line 1: I also changed it so that only the first rect is anti-aliased, since the other ones are really about testing images + opacity layers instead.

LaurenzV:
Another option might be to instead store each probe element as a separate PNG/RGBA buffer, and then compare the cells in the actual image directly against the singular probe reference images. Then, we could for example also return per-probe-element statistics about the number of differing pixels etc. instead of just one globally. But it would probably make the code a bit more complicated. WDYT @grebmeg

vello_common/src/probe.rs
```diff
@@ -232,91 +239,97 @@ pub trait ProbeRenderer {
...
+    // is white as well.
+    let mut data = alloc::vec![255; usize::from(width) * usize::from(height) * 4];
+    let reference_layout = GridLayout::from_elements(REFERENCE_ELEMENTS);
+    let reference_width = usize::from(reference_layout.canvas_size().0);
+
+    for (index, element) in elements.iter().copied().enumerate() {
+        copy_cell(
+            REFERENCE_RGBA,
+            reference_width,
+            reference_layout.cell_origin(element as usize),
```
grebmeg: This lookup relies on every `ALL_PROBE_ELEMENTS` index matching its enum discriminant. Could we either enforce that invariant at compile time:

```rust
const _: () = {
    let mut index = 0;
    while index < ALL_PROBE_ELEMENTS.len() {
        assert!(ALL_PROBE_ELEMENTS[index] as usize == index);
        index += 1;
    }
};
```

or define an exhaustive atlas mapping:

```rust
impl ProbeFeature {
    const fn atlas_index(self) -> usize {
        match self {
            Self::SolidRect => 0,
            // ...
            Self::DepthBuffer => 9,
        }
    }
}
```

I slightly prefer the explicit mapping because adding a variant forces the atlas location to be considered. Maybe there're some other options.

vello_common/src/probe.rs
```diff
@@ -178,13 +186,8 @@ impl<E> Probe<E> {
...
     /// Construct a new probe result by inspecting the provided pixmap and comparing it
     /// against the reference output.
-    pub fn from_actual(actual: Pixmap) -> Self {
-        let (width, height) = canvas_size();
-        let expected = ProbeImage {
-            width,
-            height,
-            data: REFERENCE_RGBA.to_vec(),
-        };
+    pub fn from_actual(actual: Pixmap, elements: &[ProbeFeature]) -> Self {
```
grebmeg, line 126: Could we compare the actual cells directly against the corresponding cells in `REFERENCE_RGBA`, without constructing an intermediate expected image? Or at minimum, it would be good to short-circuit `ALL_PROBE_ELEMENTS` and use the complete reference image directly. I think I’d prefer the first option, but I’ll leave it to your judgment.

grebmeg (approved):
LGTM

LaurenzV:
@grebmeg What do you think of this new approach?

grebmeg:
> @grebmeg What do you think of this new approach?

I think this is even better, as we can get more data per cell and therefore reason more precisely about what might be causing mismatches in that specific cell.

vello_common/src/probe.rs
```diff
@@ -232,91 +191,98 @@ pub trait ProbeRenderer {
...
+                for (max_discrepancy, (expected, actual)) in statistics
+                    .max_channel_discrepancy
+                    .iter_mut()
+                    .zip(expected.iter().zip(actual))
+                {
+                    *max_discrepancy = (*max_discrepancy).max(expected.abs_diff(*actual));
+                }
+            }
+
+            if !pixels_within_tolerance(expected, actual, CHANNEL_TOLERANCE) {
```
grebmeg: nit: `cell_statistics` computes each channel’s `abs_diff` once for `max_channel_discrepancy` and again inside `pixels_within_tolerance`. Could these loops be fused?

grebmeg (approved):
LGTM

vello_common/src/probe.rs
LaurenzV: Yep!

PR #1915, opened 2026-09-15
vello_gpu: reuse staging belt for strips upload

HigherOrderLogic:
Clear more TODO.

This remove the queue allocation and use an arena-like staging belt instead.


LaurenzV:
Ah, sorry, I forgot to mention that there was a reason why we haven't used the wgpu-native staging belt, see this PR: https://github.com/linebender/vello/pull/1532

Or does this problem not exist in this implementation?

HigherOrderLogic:
> Or does this problem not exist in this implementation?

I think it should eliminate the overlapped memory problem, since the old belt is dropped every time a new one is created. However, there would still be overlapped memory spike, for example when frames are submitted too fast, but this should only last for a short duration.

PR #1916, opened 2026-09-15
sparse_strips: Add a `push_clip_rect` method

LaurenzV:
This PR adds a new `push_clip_rect` method. It therefore implements a part of what was done in https://github.com/linebender/vello/pull/1786 (though this PR was written from scratch), but is less "ambitious" because it doesn't attempt to add an optimization for axis-aligned rectangles yet. 

Right now, the only advantage of this PR is that by using `push_clip_rect`, we can avoid the generic strip generation mechanism and instead use the CPU-side rectangle fast path. This gives a good speed boost if there are many rectangle clips, but for Vello GPU, there's still the disadvantage that the existence of any kind of clip prevents using the **GPU-side** fast-rect path. This PR makes a first step towards changing that by introducing a semantic distinction between rectangle clips and normal clips. In the future, we can add an optimization such that, if the currently active clip path is a rectangle and we draw another rectangle, we can just intersect them directly and send the resulting rectangle through the GPU fast path, instead of going through the slower CPU-side path. But this is left as a future optimization.


vello_common/src/clip.rs
```diff
@@ -87,13 +87,38 @@ impl ClipContext {
...
+                aliasing_threshold,
+                storage,
+                existing_clip,
+            );
+        });
+    }
+
+    /// Push a rectangular clip path.
+    #[inline]
+    pub fn push_clip_rect(&mut self, rect: &Rect, strip_generator: &mut StripGenerator) {
```
grebmeg, line 112: Could we benchmark `push_clip_rect` against `push_clip_path` end to end? I’m curious about the numbers. Or is it more like the difference between `generate_filled_path` and `generate_filled_rect_fast`?

grebmeg (approved):
LGTM

vello_common/src/clip.rs
LaurenzV: Yep, it's more the difference between those two. Will add some more benchmarks in a follow-up.

LaurenzV: Added here: https://github.com/linebender/vello/pull/1922

PR #1917, opened 2026-09-15
vello_tests: Refactor test infrastructure

LaurenzV:
This PR refactors the test infrastructure a bit. In particular, it gets rid of a lot of duplicated code by changing how the various test variants are represented and then constructed.

Apart from that, it also fixes two bugs:

- WASM with `simd128` now uses the SIMD tolerance instead of the scalar one.
- Cached glyph tests now also run in WebGL, which wasn't the case before (probably accidentally?).

Can be reviewed commit by commit.


grebmeg (approved):
Nice! Love it! ❤️

PR #1918, opened 2026-09-15
vello_gpu: Rename remaining "hybrid" occurrences

LaurenzV:
This PR gets rid of the hybrid branding in some more places and replaces it with the new GPU branding instead.


vello_gpu/src/scene.rs
```diff
@@ -201,7 +201,7 @@ impl Default for RenderSettings {
     }
 }
 
-/// A render context for hybrid CPU/GPU rendering.
+/// A render context for GPU rendering.
```
LaurenzV, line 204: Should we keep the hybrid branding here?

grebmeg: I think it’s okay to get rid of "hybrid" here.

grebmeg (approved):

PR #1921, opened 2026-09-16
 vello_gpu: Report actual depth-buffer precision

grebmeg:
Report the actual and required depth-buffer bit counts when a WebGL context has insufficient depth precision, improving compatibility diagnostics.


vello_gpu/src/render/webgl/mod.rs
```diff
@@ -399,9 +402,12 @@ impl WebGlRenderer {
                 .as_f64()
                 .unwrap() as u32;
 
-            if actual_bits < 24 {
+            if actual_bits < REQUIRED_DEPTH_BITS {
```
grebmeg, line 405: Actually, could we use `actual_bits` in the shader instead of requiring 24 bits? That might let us retain the early-Z optimization on devices with 16-bit depth buffers. @LaurenzV WDYT?

LaurenzV: We could, but I think it would be interesting to get some data on how often such devices actually appear. I think in the majority of the cases the number of depth bits will just be 0, but I might be wrong!

And while it's still very extreme, it doesn't seem completely unlikely to hit the 65k object limit, so to me it seems better to wait a bit, but up to you!

LaurenzV (approved):

PR #1923, opened 2026-09-16
Make CI faster

LaurenzV:
This PR tries to address two of the main reasons why CI in the Vello repository has become both slow and flaky:

- We have too many CI jobs. In practice, there appears to be a repository-wide concurrency limit of around 20 jobs. With multiple PRs or pushes in flight, we exhaust those slots very quickly and every run has to wait for unrelated jobs to finish.
- We only have 10 GB of GitHub Actions cache storage, and the individual Rust caches can be close to 1 GB each. This causes constant cache thrashing and can even evict the [Git LFS cache](https://github.com/linebender/vello/actions/caches), after which tests may fail because their PNG reference images are unavailable. And it also causes us to often not be able to profit form the build caches, making CI times very unpredictable.

This PR makes the following changes:

- Rust build caches are now only written by successful runs on `main`. Pull requests can still restore caches from `main`, which is what matters for most code changes. The trade-off is that a PR which updates dependencies cannot populate a reusable cache across multiple pushes until it is merged.
- Less frequently useful checks now run once every 24 hours against `main`, or manually through `workflow_dispatch`. This includes Clippy on macOS and Windows, Android checks, all native and Wasm MSRV checks, and the AVX-512 tests. We also disable the cache for those, ensuring that we save what space we have available for the CI jobs that run on every push.
- The Web Demo is now updated once per day instead of after every push to `main`.
- The existing formatting checks and the very fast typo check now share a single `static checks` job.
- The WebGL example tests are folded into the default WASM browser-test job. They still run for every PR, but no longer require another runner and a separate Rust cache.

I know it's a bit unfortunate to have some of these checks (like the msrv ones) not run on every push, but I think it's a trade-off we have to make to accelerate the common case and make the best use of the few resources we currently have available. This does mean it might sometimes happen that we only notice something broke after the PR was already merged and need a separate PR to fix it, but I've purposefull only chosen the jobs that I consider less important or less likely to break. For example, we have very little MacOS/Windows specific code in the repo, so it's unlikely that we accidentally break this. The MSRV one also is mostly only a concern when updating to new dependencies, and only sometimes when accidentally using new features that aren't available yet, but it hasn't really happened to me in practice so far. My suggstion would simply be to run the manual workflows once before making new releases of the crates.

The Wasm renderer-size check and allocation benchmark also still have their own jobs and caches, but I think we can fold these into the other jobs as well in a follow-up. I am also still wondering whether the non-SIMD WASM test run needs to execute for every PR, since much of its coverage should overlap with `fearless_simd`, but this PR leaves it unchanged for now.

With these changes applied:
- CI cache is stable at around 8GB (tested this on my private branch), so no more thrashing unless we actually bump dependencies.
- CI times are now much more predictable, taking at most **~12 minutes** (instead of 20-30), see [here](https://github.com/laurenz-canva/vello/actions/runs/35088917416/job/104770349066) for an example run. The bottleneck is now Windows, where just building the workspace crates themselves takes around 6 minutes, and running the tests themselves another 4 minutes.

Further TODOs, which I'm going to keep for a follow-up for simplicity:
- The build cache seems to still not properly work for SIMD, gotta look into that.
- As mentioned, look into folding wasm renderer size and allocation benchmark job into existing jobs, as they also unnecessarily consume CI cache space.
- Look into shortening Windows CI. For example, since we already run `cargo test --doc` on Linux, we likely don't need to run it again on Windows. Depending on what exactly is taking up all the time when compiling on windows, we might for example try to not compile all SSE4/AVX2 variants on Windows.
- GPU tests take suspiciously long on all platforms, maybe there is more stuff that can be done to optimize this.


nicoburns:
Could we just make MSRV linux only? That would catch the common case hopefully without adding too much overhead.

The other big win would likely be to not test the power set of feature flags. From other repos, I find that just testing each feature individually catches 99% of issues.

.github/workflows/ci.yml
```diff
@@ -140,6 +149,8 @@ jobs:
 
       - name: restore cache
         uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
HigherOrderLogic, line 153: Wont this make CI on PRs longer?

LaurenzV: Only if the PR changes something in the dependencies, I think.

LaurenzV:
> Could we just make MSRV linux only? That would catch the common case hopefully without adding too much overhead.

Perhaps, but we are already scratching the 10GB limit so I'd like to avoid adding more. :(

> The other big win would likely be to not test the power set of feature flags. From other repos, I find that just testing each feature individually catches 99% of issues.

This might help with runtimes, but probably not with the size of the caches, so it's a bit orthogonal to this PR specifically. But I agree it's worth investigating improvements here!

LaurenzV:
I tested the Zulip thing and it's very easy, just need to add the bot API key to the repo secrets once we decide to merge this.

LaurenzV:
I'm wondering whether this is worth landing in isolation or whether I should first find all the wins and then do a single PR with all the improvements I was able to find?

LaurenzV:
I think it's probably better if I first try to push it as hard as I can and then make a single PR.

grebmeg:
I removed myself as a reviewer since this is now in draft. Please reassign me when it’s ready for review again.

grebmeg:
Wanted to share my main high-level concern early:

I don’t think running these checks nightly is an equivalent replacement for running them before merge. With this change, a PR can be green and merge even though the Android, native and Wasm MSRV, AVX-512, and Windows/macOS Clippy checks have not run. If one fails later, the nightly build only tells us that `main` is already broken rather than preventing the regression. That introduces avoidable friction: someone must investigate the delayed failure, recover the original context, prepare another PR, and potentially work around a broken `main` in the meantime. It may also interrupt unrelated contributors whose work depends on a healthy main branch.

Could we preserve the full pre-merge coverage and reduce resource usage in other ways? For example, we could keep the main-only cache writes, run the full suite only for non-draft PRs and merge groups, combine related checks into the existing platform jobs, serialize some checks to use fewer runners, and skip checks based on changed paths only where we can prove they are irrelevant.

I’m comfortable with the final validation taking a little longer if that reduces runner usage and cache pressure, but I don’t think we should trade away merge confidence. The state I would like us to preserve is that anything landing on `main` has already been fully tested and validated against every check that forms part of the project’s supported contract. A green, mergeable PR should give us strong confidence that the code complies with all those checks, rather than relying on problems being discovered after it has merged.

PR #1925, opened 2026-09-17
Bump upper bound of `skrifa` to 0.47

akx:
To allow transitively pulling in read-fonts 0.44.0.


LaurenzV:
I don't really understand the purpose here, why not just limit it on your side?

akx:
Similar idea, context and discussion here: https://github.com/pop-os/cosmic-text/pull/513

I'm looking at upgrading the deps for [egui](https://github.com/emilk/egui) without having multiple copies of crate versions linked in.

The `crate.toml` there says things like

```
harfrust = "0.12.0" # Can't update to 0.13: it needs read-fonts 0.43, but fontique 0.11 pins read-fonts 0.41, so bumping splits read-fonts into two versions
skrifa = { version = "0.44.0" ... # Can't update to 0.45+: `glifo` (via vello_cpu 0.2) pins skrifa 0.44, so bumping splits skrifa into two versions
```

so this PR is the part that would address the egui -> vello_cpu -> glifo -> skrifa chain; skrifa (`main` version anyhow) already requires read-fonts 0.44: https://github.com/googlefonts/fontations/blob/9f3cd3c7f86e9db9a86bff934fd9682e2e765798/Cargo.toml#L55

My https://github.com/harfbuzz/harfrust/pull/478 is the part that would address harfrust's  read-fonts 0.44 bump.

https://github.com/dfrg/swash/pull/144 is another one in the same universe.

---

If there is a way to force this downstream, I'm all ears, but it looks like ranges are the way when a dep is < 1.0?

nicoburns:
@akx Are you an egui maintainer? I've been trying to do synchronised dep bumps of Fontations deps across all of vello, parley, resvg, swash. Would be happy to include egui in that (that would also be pretty helpful for Servo).

But I suggest we wait for a corresponding HarfRust release to be available before attempting a synchronised upgrade.

This will likely need to be backported to stable Vello branches (don't exist yet) for release.

akx:
@nicoburns Not a maintainer, just an user and contributor. Been bumping dependencies there and in surrounding crates for a couple personal projects. (And I contributed to https://github.com/linebender/resvg/pull/922 too! :) )

LaurenzV:
Yeah I think I'd rather we just push for updating to newest skrifa in an organized manner everywhere it's needed, instead of trying to limit the version!

PR #1926, opened 2026-09-17
sparse_strips: optimize drawing rectangles with active rectangular clips in place

LaurenzV:
This PR adds an optimization for drawing rectangles when there is an active rectangular clip in the clip stack. Inside of the clip stack, we now keep track whether the currently active clip path is in fact just a rectangle or some arbitary path.

For vello_cpu, when drawing a rectangle we know avoid intersecting it using the strip-intersection algorithm, instead intersecting the rectangle mathematically before processing further. In Vello Hybrid, we can go a step further and now accept the GPU fast-rect path in case the currently active clip path is a rectangle!

Something worth noting is that it changes the semantics of some anti-aliasing calculations, but they are now actually more correct than before for this specific case. The `clip_non_isolated_with_aa_with_rect_aa` test now correctly calculates 50% coverage instead of 25%, because we calculate the resulting rectangle by mathematically intersecting it with the clip path (therefore avoiding conflation artifacts) instead of calculating it by multiplying the conflated coverages. It does produce some inconsistency (for example, if you were to run the same test but with push_clip_path instead you would get 25% again), but this new version is technically more correct, so I think it's fine to accept?

Note that whenever we call `push_clip_rect`, we still eagerly calculate the strip representation of the rectangle, even if we might not actually need it in case only rectangle are drawn. It might be possible to compute the strips lazily instead, but this sounds complicated enough that I'm going to leave this for a follow-up, if at all.

As a result, when drawing thing while constraining the viewport once in the beginning (e.g. when doing dirty rectangle rendering), we are now much faster when drawing many quads!

https://github.com/user-attachments/assets/1b897436-a603-42e3-bfaa-d7fa533e4050

https://github.com/user-attachments/assets/2a8be835-b6bf-4a5f-a92a-14d732ccd2b1


LaurenzV:
Gonna test this on my PDF library as well when I'm home, but overall this should be ready for review.

nicoburns:
Very excited for this optimization, because this is an exceedingly common case for web content.

grebmeg (approved):
Brilliant idea! Nice one!

PR #1928, opened 2026-09-18
feat: deploy Vello GPU web demos

grebmeg:
Expand the Vello web deployment with native WebGL and wgpu WebGL demos, move the research renderer under `/vello/research/`, and add a selector for switching between renderers.


LaurenzV (approved):
Overall fine from my side, but could you try running it on your fork to make sure it works (and maybe post the link so we can see it live, in case it's not too much work)?

.github/workflows/web-demo.yml
```diff
@@ -29,27 +29,140 @@ jobs:
...
+      - name: build research demo (wasm)
         run: cargo build -p with_winit --bin with_winit_bin --release --target wasm32-unknown-unknown
 
-      - name: package wasm
+      - name: package native WebGL demo
         run: |
           mkdir public
-          wasm-bindgen --target web --out-dir public target/wasm32-unknown-unknown/release/with_winit_bin.wasm --no-typescript
+          wasm-bindgen --target web --out-dir public target/wasm32-unknown-unknown/release/native_webgl.wasm --no-typescript
           cat << EOF > public/index.html
```
djmcnab-canva: I think this file is probably now big enough that it warrants living as a 'real' HTML file?

djmcnab-canva (commented):
Some thoughts:

1) Having three inline embedded HTML files in the yml isn't great.
2) I would still like us to move this away from https://linebender.org/vello, but that needn't block this PR.
3) I presume that we don't need/want a WebGPU GPU or Vello CPU demo

Otherwise, conceptually this seems reasonable enough.

.github/workflows/web-demo.yml
```diff
@@ -29,27 +29,140 @@ jobs:
           # If updating, also update in research/examples/with_winit/Cargo.toml
           tool: wasm-bindgen@0.2.115
 
-      - name: build (wasm)
+      - name: build native WebGL demo (wasm)
+        run: cargo build -p native_webgl --bin native_webgl --release --target wasm32-unknown-unknown
+
+      - name: build wgpu WebGL demo (wasm)
+        run: cargo build -p wgpu_webgl --bin wgpu_webgl --release --target wasm32-unknown-unknown
```
DJMcNab: The terms "native WebGL" and "wgpu WebGL" are very cryptic. I have enough context to parse them, but I don't think they are especially self-evident.

What do you think about "direct" or perhaps "web_sys" instead of "native"?

DJMcNab (commented):
Some thoughts:

1) Having three inline embedded HTML files in the yml isn't great. I think they probably should be their own files?
2) I would still like us to move this away from https://linebender.org/vello, but that needn't block this PR.
3) I presume that we don't need/want a WebGPU GPU or Vello CPU demo

Otherwise, conceptually this seems reasonable enough.

grebmeg:
> Overall fine from my side, but could you try running it on your fork to make sure it works (and maybe post the link so we can see it live, in case it's not too much work)?

Thanks! I ran the workflow on my fork and verified that all three demos load and the selector works between them: https://grebmeg.github.io/vello/
Workflow run: https://github.com/grebmeg/vello/actions/runs/35547378353

.github/workflows/web-demo.yml
grebmeg: Agreed — “direct” describes the distinction better than “native”, while `web_sys` feels too implementation-specific. I’ll use “Direct WebGL2” and “WebGL2 via wgpu” for the human-facing labels and leave the existing package names unchanged.

grebmeg:
> 1. Having three inline embedded HTML files in the yml isn't great. I think they probably should be their own files?

Good point. I extracted the three HTML documents into standalone files under `.github/web-demo/`; the workflow now copies them into the Pages artifact instead of embedding.

> 2. I would still like us to move this away from https://linebender.org/vello, but that needn't block this PR.

Yeah, agreed — let’s handle this separately after this PR. What hosting location or setup would you like to see us move toward?

> 3. I presume that we don't need/want a WebGPU GPU or Vello CPU demo

I think we’d actually like to include both. Vello CPU already has a ready WASM demo, so I added it here since it was inexpensive to include. I’ve left Vello GPU WebGPU out for now; that path still needs some additional work, so it can follow separately.

grebmeg:
I’ve redeployed the updated demos on my fork and verified that all four load correctly: https://grebmeg.github.io/vello/. This now includes Direct WebGL2, WebGL2 via wgpu, Vello CPU, and the research renderer. The workflow run is available at https://github.com/grebmeg/vello/actions/runs/35549734010.

PR #1932, opened 2026-09-18
vello_common: Consider alpha interpolation space in gradient cache key

LaurenzV:
Should be self-explanatory.


vello_common/src/encode.rs
```diff
@@ -814,6 +818,7 @@ impl BitEq for GradientCacheKey {
     fn bit_eq(&self, other: &Self) -> bool {
         self.stops.bit_eq(&other.stops)
             && self.interpolation_cs == other.interpolation_cs
+            && self.interpolation_alpha_space == other.interpolation_alpha_space
```
grebmeg, line 820: Would it make sense comapring the cheap enum fields first and move `stops.bit_eq()` to the end? That allows mismatched interpolation settings to short-circuit before comparing the variable-length stop list.

grebmeg (approved):

vello_common/src/encode.rs
LaurenzV: Hmm sure!

PR #1934, opened 2026-09-19
vello_tests: Allow specifying pixel diffs for CPU and GPU separately

LaurenzV:
As requested in https://github.com/linebender/vello/pull/1878#discussion_r3954312453.


grebmeg (approved):
I think it’s fine for now, but what I initially proposed was refactoring our testing framework to make it more flexible, so you can configure all tolerances per target. As far as I remember, the Skia test suite even has separate golden images per target, rather than using a single golden image across multiple targets like we do. I’m not proposing that we do the same as Skia here, at least not for now.

LaurenzV:
Per-target, you mean per operating system? Or you mean per SIMD-pipeline combination? Or per hardware configuration?

PR #1936, opened 2026-09-20
Tweak offset hack for Apple Color Emoji

LaurenzV:
I had an LLM do a deep-dive into the Apple Color Emoji shift again ([if you want to read the whole slop "report"](https://github.com/user-attachments/files/32431049/apple-emoji-placement-report.html)) , and also checked some other sources online. Based on these, I made some adjustments on how we handle this:
- This shift seems to specifically apply only to Apple Color Emoji, not any sbix font in general.
- The shift also seems to be more generally 1/8 upem, instead of always being 100 font units (Apple Color Emoji happens to have a upem of 800, hence why this matches what we had previously).
- The shift also seems to be independent of the existing glyph bearings, so I removed that part.

It seems like the shift is slightly different on iOS and there might be more stuff going on with small glyphs specifically, but I think this is the best we can do.

We already have a visual regression test for Apple Emoji Color and it still works fine, so this also doesn't change the existing behavior. It just tweaks the heuristic for determining this, to not affect other sbix fonts.


grebmeg (approved):
LGTM if LGT @DJMcNab

DJMcNab:
I never had a mac at the time I last touched this code. I can't remember the provenance, but I don't think I'm the person who can gate this.

grebmeg:
> I never had a mac at the time I last touched this code. I can't remember the provenance, but I don't think I'm the person who can gate this.

@DJMcNab I added you here as the most relevant person, in my opinion, since you’re currently working on text, so your expertise could help if anything should be approached differently. You also worked on and are familiar with the vello research. Let me know who else would be a good person to co-review this.

DJMcNab (commented):
I see. I thought I was being brought in to review from the open source side, as it were.

I don't have much to say on this, but it seems reasonably low risk. I do note that some of the linked discussion mention that the method used to get the font matters.

LaurenzV:
Yes, that's true. But back then when this topic was discussed for Typst, not applying this offset made the emojis look artifically high, so applying the offset by itself is definitely the right thing to do, I think. You can also see it in the harfbuzz thread, I think the second version undoubtedly looks better than the first one.

image

PR #1939, opened 2026-09-20
sparse_strips: Bug fix in sbix transformation handling

LaurenzV:
Not sure how that happened, but somehow we managed to mess up the signs in the code that handles sbix glyphs. The Skia code. The [Skia code](https://github.com/google/skia/blob/61ac357e8e3338b90fb84983100d90768230797f/src/ports/SkTypeface_fontations.cpp#L669-L678) does this correctly, so we probably messed that up when porting.

In order to have better testing, the first commit adds an (AI-generated) test sbix font that tries to exercise the various different cases of a sbix font. As you can see, in that first commit, many letters in that reference image are all over the place, and doesn't match CoreText at all. Here is also a video:

https://github.com/user-attachments/assets/2461d232-869b-4350-b9dc-6df476375a98

With the second commit that fixes the sign, the image looks much better. 

image

There are two remaining failures:
- The Q and R are still shifted, apparently due to a bug in skrifa. Haven't looked into it more deeply yet.
- Another set of letters at the end fails because we don't support the `flip` flag yet, which is also pending upstream: https://github.com/googlefonts/fontations/pull/1631


grebmeg (approved):
LGTM if LGT @DJMcNab

LaurenzV:
Let me know if the Python script is too sloppy to keep in the repo, but I think it's good if we can tweak the font in the future.

LaurenzV:
Confirmed that this also fixes the offset issue we observed in production.

PR #1941, opened 2026-09-21
feat: add WebGPU browser demo

grebmeg:
Adds a standalone Vello GPU browser example using WebGPU through wgpu. Integrates it into the GitHub Pages workflow and demo selector alongside the existing WebGL and CPU examples.


grebmeg:
I ran the updated workflow on my fork and verified that all five demos load and the selector works between them: https://grebmeg.github.io/vello/

LaurenzV:
It's probably not related to this PR, but just wanted to remark that panning and zooming seem to not work correctly compared to WebGL:


https://github.com/user-attachments/assets/0b71f35c-29f7-4f07-b466-9215ecd39553

vello_gpu/examples/wgpu_webgpu/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+        use wgpu_webgpu::run_interactive;
+
+        console_error_panic_hook::set_once();
+        console_log::init_with_level(log::Level::Debug).unwrap();
+
+        let window = web_sys::window().unwrap();
+        let dpr = window.device_pixel_ratio();
+
+        let width = window.inner_width().unwrap().as_f64().unwrap() as u16 * dpr as u16;
+        let height = window.inner_height().unwrap().as_f64().unwrap() as u16 * dpr as u16;
```
LaurenzV: I think the DPR can also be fractional, so we should first multiply in f32/f64 and then cast to u16.

LaurenzV (approved):

vello_gpu/examples/wgpu_webgpu/src/main.rs
grebmeg: Good catch!

grebmeg:
> It's probably not related to this PR, but just wanted to remark that panning and zooming seem to not work correctly compared to WebGL:

Thanks for flagging this — it was related to a CSS-pixel versus canvas-pixel mismatch in the new WebGPU demo. Fixed now.

PR #1943, opened 2026-09-21
vello_hybrid: Add more timing information to the probe and renderer initialization

LaurenzV:
This adds some more timing information to both, the initialization of the renderer as well as the probe, so we can get some more insight into where the time goes. Unsure whether I've captured all of the main things we care about, but I think this is a good start, and if we notice that we need more information somewhere, we can add it in the future. I'd like to avoid having too many `performance` calls, so hopefully this is a good compromise.

And if not, this PR at least establishes the basis so we can add more timings more easily in the future!


vello_gpu/src/render/webgl/probe.rs
```diff
@@ -184,7 +265,15 @@ impl WebGlRenderer {
         // Propagate render failures only after restoring the previous framebuffer.
         render_result?;
 
-        let pending = launch_probe(&self.gl, &probe_framebuffer, width, height, elements)?;
+        let pending = launch_probe(
```
grebmeg, line 212: Could we separate the `render_scene` timing from `launch_probe` so we can distinguish CPU-side render submission from readback setup? It would also be useful to report whether `EXT_disjoint_timer_query_webgl2` is available and, when supported, include the GPU timing around `render_scene`, wdyt? Could we also optionally time the GPU-side readPixels operation with EXT_disjoint_timer_query_webgl2? This would help distinguish slow GPU framebuffer transfers from the existing CPU readback and polling timings, especially across different devices. Essentially we'd have something like `render_submission`, `gpu_render`, `readback_submission`, `gpu_readback`.

vello_gpu/src/render/webgl/mod.rs
```diff
@@ -282,6 +328,7 @@ impl WebGlRenderer {
                 "`WebGlRenderer` can only be constructed when targeting `wasm32`",
             );
         }
+        let initialization_started_at_ms = now_ms();
```
grebmeg: Could we expose a more detailed initialization breakdown using `performance.now()`? For example, `context_creation` would cover obtaining/configuring WebGL2, `shader_submission` the compile/link calls, `resource_initialization` textures/buffers/framebuffers/VAOs, `shader_completion` the wait until parallel compilation finishes, and `finalization` link validation and interface lookup. These measurements should be inexpensive and could help explain startup differences between devices without requiring GPU timer queries. WDYT?

PR #1944, opened 2026-09-22
glifo: Make glyph-drawing methods fallible

LaurenzV:
This PR changes it so that `fill_glyphs` and `stroke_glyphs` return an `Option`, allowing users to inspect whether there was any glyph that didn't have a rendereable representation. Users who care about this can just unwrap, users who would rather keep rendering and just ignore those glyphs don't unwrap.

The motivation for this is to handle the case where we have a bitmap font but the `png` feature is disabled. We want to be able to detect this and handle this properly instead of silently rendering nothing.

I hope I didn't miss any edge case here, but it seems like as long as there is a glyf/CFF table, even empty outlines (like for the space glyph) are registered by skrifa, so this should really only trigger if 1) the Glyph ID is out-of-bounds or 2) The glyph only has color tables, but we didn't manage to choose an appropriate representation (e.g. because the PNG feature was disabled).


LaurenzV:
Unfortunately this doesn't fix our original issue since the font in question actually has a `glyf` table but with placeholder outlines. 🫠 But overall I still think this is a good addition to the API.

nicoburns (changes requested):
This definitely feels like it ought to be a `Result` rather than an `Option`. You could then encode the failure reason(s) in the error type.

LaurenzV:
Not sure about this one, because it's not a really a "failure". It might be on purpose, so this is more of a hint that this happened. But I agree an Option isn't great either, so open to other suggestions.

glifo/src/glyph.rs
```diff
@@ -307,12 +307,20 @@ pub trait GlyphRunBackend<'a>: Sized {
     fn atlas_cache(self, enabled: bool) -> Self;
 
     /// Fill the given glyph sequence using the configured builder state.
-    fn fill_glyphs<Glyphs>(self, run: GlyphRun<'a>, glyphs: Glyphs)
+    ///
+    /// Returns `None` if at least one of the glyphs had no renderable representation,
+    /// thus rendering as blank. Even if `None` is returned, all unaffected glyphs will still be
+    /// rendered successfully.
+    fn fill_glyphs<Glyphs>(self, run: GlyphRun<'a>, glyphs: Glyphs) -> Option<()>
```
grebmeg, line 314: What do you think about returning a `Result` with information about every glyph that was skipped? `Result` is already `#[must_use]`, and this would let callers distinguish cases such as an invalid glyph ID from missing PNG support while unaffected glyphs continue rendering. Something like:

```rust
pub struct GlyphRenderError {
    pub skipped: Vec<SkippedGlyph>,
}

pub struct SkippedGlyph {
    pub index: usize,
    pub glyph: Glyph,
    pub reason: GlyphSkipReason,
}

pub enum GlyphSkipReason {
    InvalidGlyphId,
    PngSupportDisabled,
    UnsupportedBitmapFormat,
    BitmapDecodeFailed,
    NoRenderableRepresentation,
}
```

The methods could then return `Result<(), GlyphRenderError>`, with documentation clarifying that `Err` means rendering was only partially successful rather than rolled back?

LaurenzV: I agree it's nicer in principle, but this means that we have a new allocation for every glyph run where this happens. :( Do you think this is really worth it?

grebmeg: That’s fair — although these cases should be rare across normal runs, an unsupported font could trigger this on every frame. We could avoid per-run allocations by: returning an allocation-free summary with the first skipped glyph, count, and reason flags; or by reporting each skip through a callback; or by accepting a caller-owned reusable buffer; or by using fixed-capacity inline storage and counting any overflow. I think there're some options. The simplest might be something like:

```rust
pub struct GlyphRenderError {
    pub first_skipped: SkippedGlyph,
    pub skipped_count: usize,
}
```

LaurenzV: Yes I think I like that more! This should be enough info for now, I don't think we are planning on really collecting detailed info about this.

LaurenzV: I updated this now. I did not include the `InvalidGlyphId` variant because it would require us to always re-query the maximum number of glyphs in the font, which I think is a bit unnecessary, instead, it simply returns `NoRenderableRepresentation`.

PR #1945, opened 2026-09-22
sparse_strips: Reset the aliasing threshold

LaurenzV:


nicoburns (approved):
Consider moving the `reset` method immediately below the struct definition to make it harder to forget to keep it updated.

LaurenzV:
Will leave this to a follow-up to keep the diff clean.

PR #1946, opened 2026-09-22
vello_cpu: make `Repeat` and `Reflect` more robust

LaurenzV:
While trying out some changes to image sampling in Vello GPU and comparing against Vello CPU, I noticed that there is a rather "serious" failure mode in our current extend algorithm on CPU when combined with bilinear/bicubic filtering.

Fundamentally, the problem is that we perform two different calculations during filtering: On the one hand, we use the exact fraction of our current position to determine fractional weights for the 4 (or 16) surrounding samples:

https://github.com/linebender/vello/blob/3b08372ad1d11a4008d724e9be69470e604aa290/vello_cpu/src/fine/common/image.rs#L247-L248

On the other hand, our extension logic has a failure mode where the positions might be snapped incorrectly due to rounding while dividing and then multiplying again:
https://github.com/linebender/vello/blob/3b08372ad1d11a4008d724e9be69470e604aa290/vello_cpu/src/fine/common/image.rs#L443-L448

Consider the test added in the first commit, where our image has a width of 3 and we are sampling the position one less to 15 (i.e. something like 14.9999). The correct position for this would be index 2, and the weight for the sample is calculated as if it were. However, due to precision loss we end up calculating index 0 instead, so we sample a neighboring pixel but with the exact same weight, resulting in a completely different color!

This PR fixes this by... Copying the simpler implementation we already use [for GPU](https://github.com/linebender/vello/blob/main/vello_gpu_shaders/shaders/helpers/extend.wesl)! Basically, we normalize the number to be between 0 and 1, do the extend application in that range and then scale back. I'm not even sure why our CPU implementation was this complicated in the first place, either an accidental relict from the past, or because we wanted to avoid using floating point computations. However, this new implementation is more robust and also seems to be faster, in fact:

TODO


LaurenzV:
Hmpf, requires a bit more thinking.

vello_cpu/src/fine/common/image.rs
```diff
@@ -428,43 +410,32 @@ pub(crate) fn sample<S: Simd>(
...
-    inv_max: f32x4<S>,
-) -> f32x4<S> {
+pub(crate) fn extend_mode<S, V>(val: V, extend: crate::peniko::Extend, max: V) -> V
+where
+    S: Simd,
+    V: SimdFloat<S, Element = f32>,
+{
+    if matches!(extend, crate::peniko::Extend::Pad) {
+        return val.min(max - 1.0).max(0.0);
+    }
```
LaurenzV, line 420: For some reason LLVM started pessimizing the `Pad` case even though the `floor + 0.5` was only applied for Reflect and Repeat, not Pad. Not sure why. 😅 But changing it to this representation seemed to fix it.

LaurenzV:
@grebmeg While I still need to benchmark this properly, since I presume this change might warrant some discussion, thoughts or ideas for a better implementation would be appreciated while I work on the benchmarking side of things!

vello_cpu/src/fine/common/image.rs
```diff
@@ -428,43 +410,32 @@ pub(crate) fn sample<S: Simd>(
...
 
 #[inline(always)]
-pub(crate) fn extend<S: Simd>(
-    simd: S,
-    val: f32x4<S>,
-    extend: crate::peniko::Extend,
-    max: f32x4<S>,
-    inv_max: f32x4<S>,
-) -> f32x4<S> {
+pub(crate) fn extend_mode<S, V>(val: V, extend: crate::peniko::Extend, max: V) -> V
```
grebmeg, line 413: Nice fix — snapping the tap before wrapping is exactly the right idea. I spent some time digging into this with an LLM and we landed on a slightly different formulation; curious what you think.

The concern with `floor() + 0.5` is that it reintroduces the same class of problem at large coordinates: half-integers stop being representable at 2^23 while integers are fine up to 2^24. For example, with width 2, `8_388_609.0` (an exact `f32`) becomes `8_388_610.0` after the `+ 0.5`, and we pick texel 0 instead of 1; `Reflect` has the same issue at `8_388_610.0` with width 3. The normalized `t / max` form also throws away low bits for large `t`, which is why the `.min(max - 1.0)` guard is needed.

The alternative is to stay in the integer domain: wrap `val.floor()` directly with a Euclidean mod, and don't trust the reciprocal multiply to give an exact quotient — just fix the sign afterwards. This is the whole thing (it brings back `width_inv`/`height_inv` on `ImagePainterData` and passes them through at the call sites):

```rust
#[inline(always)]
pub(crate) fn extend_mode<S: Simd>(
    val: f32x4<S>,
    extend: crate::peniko::Extend,
    max: f32x4<S>,
    inv_max: f32x4<S>,
) -> f32x4<S> {
    match extend {
        crate::peniko::Extend::Pad => val.min(max - 1.0).max(0.0),
        // Snap to the texel first; everything below is integer-valued and exact.
        crate::peniko::Extend::Repeat => euclid_mod(val.floor(), max, inv_max),
        crate::peniko::Extend::Reflect => {
            // Wrap over a period of two images, then mirror the second half back:
            // width 3: 0 1 2 3 4 5 -> 0 1 2 2 1 0
            let period = max + max;
            let r = euclid_mod(val.floor(), period, inv_max * 0.5);
            r.simd_ge(max).select(period - 1.0 - r, r)
        }
    }
}

/// Euclidean `t mod m` for integer-valued `t` and positive integer `m`; exact for `|t| < 2^24`.
#[inline(always)]
fn euclid_mod<S: Simd>(t: f32x4<S>, m: f32x4<S>, inv_m: f32x4<S>) -> f32x4<S> {
    // Nearest quotient: `|t - q*m| <= m/2 + eps < m`, so only the sign can be wrong.
    let q = (t * inv_m).round_ties_even();
    // Fused, so the remainder is exact whenever it is representable (always: |r| < m).
    let r = q.mul_add(-m, t);
    r.simd_lt(0.0).select(r + m, r)
}
```

Everything here is an integer in `f32`, so it's exact for all `|t| < 2^24` with no clamp; I checked it against an `i64` oracle (30M random cases plus an exhaustive sweep of every tile seam for several widths) with zero mismatches, and it passes the existing snapshot tests. It also drops the per-tap division: on my M-series machine the mode-specific overhead vs. Pad came out ~4–9% lower than this PR for `Repeat` and ~5–16% lower for `Reflect` across NN/bilinear/bicubic. `period` and `inv_max * 0.5` could also be precomputed per axis to shave two more ops off `Reflect`.

LaurenzV: Thanks! I like it, updated to your proposal, benchmarks on NEON are looking good.

The only issue is that `mul_add` is not guaranteed to have only a single rounding step (there is `mul_add_precise` but we need to update to fearless_simd 1.0 first), but it only causes issues for even larger numbers, so I suggest we disregard this for now.

